PR #23222, opened 2024-08-26
[improve] [pip] PIP-375 Expose the Admin client configs: readTimeout, requestTimeout, and connectionTimeout

poorbarcode:
### Motivation

- [DIscussion](https://lists.apache.org/thread/nwvc1jc29k0hnvtn6qtpdyb8b5xp3ch7)
- [Vote](https://lists.apache.org/thread/2bmzvv7hyfnb2kf6jsgmgjtrjz23qhj2)

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


pip/pip-375.md
```diff
@@ -0,0 +1,32 @@
+# PIP-375: Expose the Admin client configs: readTimeout, requestTimeout, connectionTimeout
+
+# Background knowledge
+
+- `pulsar-admin sinks create` will upload files to brokers, but if the file is large and the network is slow, it may cost a long time.
+- The default value of `requestTimeout` of Admin API is `60s`, after the fix https://github.com/apache/pulsar/pull/23128, it was changed to `300s`.
```
lhotari, line 6: regarding "it was changed to `300s`", where did that exactly happen?

liangyepianzhou (approved):
LGTM!
These configurations is very helpful to me.

pip/pip-375.md
poorbarcode: The default value of `requestTimeout` is `300s` and the default value of `readTimeout` is `60s`, before #23128 , the admin client used an in-correct variable (`readTimeout = 60s`), after that it used a correct variable (`requestTimeout=300s`)

pip/pip-375.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+### CLI
+
+**client.conf**
+```properties
+adminReadTimeoutInSeconds=300
+adminRequestTimeoutInSeconds=300
+adminConnectionTimeoutInSeconds=30
+```
```
lhotari, line 37: Are these values intended to be good sample values? 
In other sources, I think that read timeout is typically set to a rather short timeout.

pip/pip-375.md
```diff
@@ -0,0 +1,41 @@
...
+
+- `pulsar-admin sinks create` will upload files to brokers, but if the file is large and the network is slow, it may cost a long time.
+- The default value of `requestTimeout` of Admin API is `60s`, after the fix https://github.com/apache/pulsar/pull/23128, it was changed to `300s`.
+
+# Motivation
+
+We'd better to expose the following configs to users:
+- `readTimeout`
+- `requestTimeout`
+- `connectionTimeout`
```
lhotari, line 13: Please add a short definition of these configuration parameters.

One reason is that there has been misconceptions about readTimeout and requestTimeout as described in #23128.

There's also a possibility for confusion about "connection timeout". One has to think whether it means "connect timeout" or "connection idle timeout", or something else. Based on the javadoc it means a "connect timeout".

pip/pip-375.md
poorbarcode: Here it is: https://github.com/apache/pulsar/blob/master/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/PulsarAdminImpl.java#L76

pip/pip-375.md
shibd: +1

pip/pip-375.md
poorbarcode: Restored to the value the same as the current default configuration.

pip/pip-375.md
poorbarcode: Added description

pip/pip-375.md
lhotari: Do you mean to say that the default *was* 60s before the change in PR 23128, which made the default value 300s?
Why was this increase in the value a problem? Or was it a problem?

pip/pip-375.md
```diff
@@ -0,0 +1,58 @@
+# PIP-375: Expose the Admin client configs: readTimeout, requestTimeout, connectionTimeout
+
+# Background knowledge
+
+- `pulsar-admin sinks create` will upload files to brokers, but if the file is large and the network is slow, it may cost a long time.
+- The default value of `requestTimeout` of Admin API is `60s`, after the fix https://github.com/apache/pulsar/pull/23128, it was changed to `300s`.
+
+# Motivation
+
+We'd better to expose the following configs to users:
```
lhotari, line 10: Why is it better?

pip/pip-375.md
```diff
@@ -0,0 +1,58 @@
...
+
+```properties
+# The value is the timeout to build a connection to the server. Value 0 represents infinity. Negative values are not allowed.
+connectionTimeout
+
+# The value is the timeout to read a response. If the server doesn't respond within the defined timeframe, ProcessingException is thrown with TimeoutException as a cause.
```
lhotari, line 16: Is it possible that the server responds, but the client just doesn't know about this? 🤔
What if the server responds, but the response isn't fully received by the client? 
Is "response timeout" different from "read timeout"?
Could we improve the definition to reduce the confusion?

pip/pip-375.md
poorbarcode: See https://github.com/apache/pulsar/pull/23222/files#diff-be84b11fa7ada15d3ef3a329b5ec3b0cd798c9d50d70d2d8c73e77714592d547R5

lhotari: It would be good to update this information in the PIP document.

pip/pip-375.md
lhotari: It's better to have clear descriptions for the settings. You could use GitHub Copilot to revisit the comment based on the feedback to assist in coming up with an improvement.

crossoverJie (approved):

Issue #23205, opened 2024-08-20
[feat] Add topic stats and metrics for observing message replay behavior and Key_Shared filtering/blocking behavior


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Currently, it's very challenging to investigate issues related to message replay ("message redelivery controller"). Some examples of this include:

- The "repeated Read-and-discard when using Key_Shared mode" issue mitigated by:
  - https://github.com/apache/pulsar/pull/22245
  - https://github.com/apache/pulsar/pull/21739
- An older mitigation: #7105

### Solution

Add topic stats and metrics for observing message replay and related Key_Shared filtering (hash blocking) behavior.

### Specific Metrics to Consider
1. Number of messages in redelivery (replay)
2. For Key_Shared subscriptions: Ways to observe internal state related to blocked hashes
3. Counter for delayed delivery messages being added to delivery (replay)

### Implementation Requirements
- It should be possible to detect replays in topic stats (or internal stats) and also in aggregated metrics
- The aggregated metrics should be usable in monitoring tools (e.g., Grafana dashboards)
- The specific types of metrics (counters, gauges) to be used will be determined in the detailed design phase

## Expected Benefits
- Improved observability for message replay and Key_Shared behavior
- Easier troubleshooting of related issues
- Enhanced monitoring capabilities for Pulsar clusters

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
It seems that PIP-282 added some metrics in https://github.com/apache/pulsar/pull/21953 that improve observability of Key_Shared.

lhotari:
There's already a counter for message redelivery: https://github.com/apache/pulsar/blob/77b6378ae8b9ac83962f71063ad44d6ac57f8e32/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java#L959-L961
However, this isn't currently exposed in the subscription stats. 
This counter was added as part of Otel changes in https://github.com/apache/pulsar/pull/22693 . 
There's also an ack counter that was added:
https://github.com/apache/pulsar/blob/77b6378ae8b9ac83962f71063ad44d6ac57f8e32/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java#L955-L957

I think that it would be a non-breaking change to expose these in stats which wouldn't necessarily require a PIP.

[...]

Issue #23200, opened 2024-08-19
[Bug] Key_Shared subscription could deliver messages late and nonContiguousDeletedMessagesRange could exceed managedLedgerMaxUnackedRangesToPersist


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

This issue is present in master, branch-3.0 and branch-3.3 branch. 
It seems to be present in all released Pulsar versions since #7105.

### Minimal reproduce step

Start Pulsar 3.3.1 in Docker:
```
docker run --name pulsar-standalone --rm -it -e PULSAR_STANDALONE_USE_ZOOKEEPER=1 -p 8080:8080 -p 6650:6650 apachepulsar/pulsar:3.3.1 /pulsar/bin/pulsar standalone -nss -nfw
```

Clone, compile & run the test app ([source code](https://github.com/lhotari/pulsar-playground/blob/key_shared_issue-2024-08-19/src/main/java/com/github/lhotari/pulsar/playground/TestScenarioIssueKeyShared.java)) that reproduces the issue:
```
git clone -b key_shared_issue-2024-08-19 --depth=1 https://github.com/lhotari/pulsar-playground
cd pulsar-playground
./gradlew build
java -cp build/libs/pulsar-playground-all.jar com.github.lhotari.pulsar.playground.TestScenarioIssueKeyShared
```

The test takes over 7 minutes to run and the final log lines show the results:
```
2024-08-19T15:31:35,983+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Done receiving. Remaining: 0 duplicates: 0 unique: 1000000
max latency difference of subsequent messages: 53071 ms
max ack holes: 27103
2024-08-19T15:31:35,983+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Consumer consumer1 received 259642 unique messages 0 duplicates in 445 s, max latency difference of subsequent messages 52959 ms
2024-08-19T15:31:35,983+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Consumer consumer2 received 233963 unique messages 0 duplicates in 417 s, max latency difference of subsequent messages 50930 ms
[...]

PR #23224, opened 2024-08-26
[improve][broker] Add msgInReplay subscription stat and metric to improve Key_Shared observability

lhotari:
Main Issue: #23205

### Motivation

Currently there aren't ways to observe the number of messages that are "in replay" with Key_Shared or Shared subscriptions. Having these metrics would make it easier to understand the behavior of issues such as #23200 .
Please check #23205 for more reasons.

### Modifications

- add `msgInReplay` in the similar way as the existing `msgDelayed` subscription stat and metric.

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageRedeliveryController.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageRedeliveryController.java
@@ -149,4 +149,13 @@ public class MessageRedeliveryController {
     public NavigableSet<Position> getMessagesToReplayNow(int maxMessagesToRead) {
         return messagesToRedeliver.items(maxMessagesToRead, PositionFactory::create);
     }
+
+    /**
+     * Get the number of messages registered for replay in the redelivery controller.
+     *
+     * @return number of messages
+     */
+    public int size() {
+        return messagesToRedeliver.size();
+    }
 }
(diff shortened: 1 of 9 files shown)


poorbarcode (approved):
Could you add a test?

dao-jun:
LGTM, but maybe the change needs a PIP?

liudezhi2098 (approved):

nlu90 (approved):

codelipenghui (commented):
+1 for having a PIP for this change since it actually changed the Public API output and metrics.

Another point is can we only have new the field in the topic stats? Since we have a lot of metrics on the topic level. Finally, we can't add everything to the prometheus metrics. The stats, the logs are also the part of observability.

dlg99 (approved):

lhotari:
> +1 for having a PIP for this change since it actually changed the Public API output and metrics.
> 
> Another point is can we only have new the field in the topic stats? Since we have a lot of metrics on the topic level. Finally, we can't add everything to the prometheus metrics. The stats, the logs are also the part of observability.

@codelipenghui This is not a breaking change in metrics and this change isn't adding significantly more metrics than what currently exists. It's currently a gap in observability of all released versions of Pulsar when msgInReplay is missing. That's why everyone needs this and making it configurable with a feature flag wouldn't be sensible.

PR #18315, opened 2022-11-03
[fix][broker] fix consume stuck of shared streaming dispatcher

coderzc:
Fixes #18314

### Motivation

In #18289, we set `sendInProgress` to false only after the last entry in a batch of messages has been delivered, but this will make consume stuck, because `StreamingEntryReader.pendingReads` may is not empty when without available entries from the ledger and lead to `ctx.isLast()` always be false.

https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/streamingdispatch/StreamingEntryReader.java#L172-L177

### Modifications

* Abstract `acquireSendInProgress()`/`releaseSendInProgress()`/`isSendInProgress()`
* In subclass PersistentStreamingDispatcherMultipleConsumers, use `sendingTaskCounter` instead of `sendInProgress` to avoid `sendInProgress` to be set to true and false multiple times.
* Clean up useless code

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/coderzc/pulsar/pull/26

PR #10762, opened 2021-06-01
[broker] Fix issue that message ordering could be broken when redelivering messages on Key_Shared subscription

massakam:
### Motivation

Messages with the same key can be out of order if message redelivery occurs on a Key_Shared subscription.

1. Suppose `PersistentDispatcherMultipleConsumers#messagesToRedeliver` contains message-1 and message-2. Message-1 will be delivered to consumer-a and message-2 will be delivered to consumer-b.
2. The dispatcher tried to send message-1 to consumer-a, but the consumer was too slow to send it.
3. Consumer-a is added to `stuckConsumers`.
https://github.com/apache/pulsar/blob/894d92b2be3bee334e7ce32760c4d2e7978603aa/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L263-L266
4. The next time `readMoreEntries()` is run, `getMessagesToReplayNow()` will return an empty Set because `isDispatcherStuckOnReplays` is true.
https://github.com/apache/pulsar/blob/894d92b2be3bee334e7ce32760c4d2e7978603aa/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L368-L374
5. The dispatcher reads newer messages instead of the messages contained in `messagesToRedeliver`.
https://github.com/apache/pulsar/blob/894d92b2be3bee334e7ce32760c4d2e7978603aa/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L233-L267
6. A new message (message-3) is delivered to consumer-b.
7. Message-2 contained in messagesToRedeliver is delivered to consumer-b.
8. As a result, the order of message-2 and message-3 is reversed.

### Modifications

When adding a message to be redeliver to `messagesToRedeliver`, save the hash of the key that the message has. If the dispatcher attempts to send newer messages to the consumer that have a key corresponding to any one of the saved hash values, they will be added to `messagesToRedeliver` instead of being sent. This prevents messages with the same key from being out of order.

PR #7105, opened 2020-05-29
Fixed KeyShared consumers getting stuck on delivery

merlimat:
~Note: this is based on top of #6791 & #7104. Once these are merged, I'll rebase here. For the sake of this review, check commit 23d6dcb7~

### Motivation

If one consumer is slowly processing messages, this can prevent other consumers from making progress on the topic. Instead we're in a loop of keep trying to replay messages without being able to dispatch any message. 

The basic idea here is that we can make progress by keep going through the topic and dispatch these messages to the consumers that are free, at least the keys that belong to them.

PR #23226, opened 2024-08-26
[improve][broker] Reschedule reads with increasing backoff when no messages are dispatched

lhotari:
Main Issue: #23200 

### Motivation

There's currently a clear problem with Key_Shared that in normal operations, it causes a lot of "ack holes" which result in several problems. One of the problems is the latency issues that are explained in #23200. Another problem is that the large number of "ack holes" exceed managedLedgerMaxUnackedRangesToPersist (10000) in usual cases such as in the demonstration in #23200. 

There are multiple other issues where there has been a large number of "ack holes" when Pulsar users have experienced problems. One of the previous mitigations is [PIP-299: Stop dispatch messages if the individual acks will be lost in the persistent storage.](https://github.com/apache/pulsar/blob/master/pip/pip-299.md). The need for PIP-299 proves that the large number of "ack holes" is a fairly common problem.

### Modifications

While experimenting on #23200, it was determined that #7105 changes were related to the cause of the issue.
I also noticed that #18315 contained some impactful changes (https://github.com/apache/pulsar/pull/18315/files#diff-c48d5c94842ac8c9a0c9314b207298069f38c8dcfeda4a9886fb3bb1f77843f2). For Key_Shared subscriptions, the change in #10762 makes the case of not dispatching any messages a common case.

Based on this information, I decided to implement a solution where there would be a backoff when no messages are dispatched.

This PR contains a change that reschedules a call to `readMoreEntries` where the delay is exponentially increasing as long as no entries are dispatched. The backoff delay starts at 100ms and is limited to 1000ms. These values are configurable in broker.conf. 

### Additional context

While testing this change, I happened to notice that this change mitigates the problem in the reproducer of of #23200.

With the changes of this PR, these are the results:
```
2024-08-26T16:09:42,328+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Done receiving. Remaining: 0 duplicates: 0 unique: 1000000
max latency difference of subsequent messages: 974 ms
max ack holes: 668
2024-08-26T16:09:42,329+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Consumer consumer1 received 259642 unique messages 0 duplicates in 456 s, max latency difference of subsequent messages 763 ms
2024-08-26T16:09:42,329+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Consumer consumer2 received 233963 unique messages 0 duplicates in 456 s, max latency difference of subsequent messages 974 ms
2024-08-26T16:09:42,329+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Consumer consumer3 received 244279 unique messages 0 duplicates in 457 s, max latency difference of subsequent messages 898 ms
2024-08-26T16:09:42,329+0300 [main] INFO  playground.TestScenarioIssueKeyShared - Consumer consumer4 received 262116 unique messages 0 duplicates in 456 s, max latency difference of subsequent messages 657 ms
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -85,6 +85,10 @@ import org.slf4j.LoggerFactory;
  */
 public class PersistentDispatcherMultipleConsumers extends AbstractDispatcherMultipleConsumers
         implements Dispatcher, ReadEntriesCallback {
+    // rescheduling a read after no entries are dispatches will be delayed by this duration using a backoff
+    private static final int RESCHEDULE_READ_INITIAL_DELAY_MS = 100;
+    // maximum delay for rescheduling a read after no entries are dispatched
+    private static final int RESCHEDULE_READ_INITIAL_MAX_DELAY_MS = 5000;
 
     protected final PersistentTopic topic;
     protected final ManagedCursor cursor;
@@ -134,7 +138,10 @@ public class PersistentDispatcherMultipleConsumers extends AbstractDispatcherMul
     private AtomicBoolean isRescheduleReadInProgress = new AtomicBoolean(false);
     protected final ExecutorService dispatchMessagesThread;
     private final SharedConsumerAssignor assignor;
-
+    protected int lastNumberOfEntriesDispatched;
+    private final Backoff rescheduleReadBackoff = new Backoff(RESCHEDULE_READ_INITIAL_DELAY_MS, TimeUnit.MILLISECONDS,
+            RESCHEDULE_READ_INITIAL_MAX_DELAY_MS, TimeUnit.MILLISECONDS, 0,
+            TimeUnit.MILLISECONDS);
     protected enum ReadType {
         Normal, Replay
     }
@@ -437,16 +444,20 @@ public class PersistentDispatcherMultipleConsumers extends AbstractDispatcherMul
 
     @Override
     protected void reScheduleRead() {
+        reScheduleReadInMs(MESSAGE_RATE_BACKOFF_MS);
+    }
+
+    protected void reScheduleReadInMs(long readAfterMs) {
         if (isRescheduleReadInProgress.compareAndSet(false, true)) {
             if (log.isDebugEnabled()) {
-                log.debug("[{}] [{}] Reschedule message read in {} ms", topic.getName(), name, MESSAGE_RATE_BACKOFF_MS);
+                log.debug("[{}] [{}] Reschedule message read in {} ms", topic.getName(), name, readAfterMs);
             }
             topic.getBrokerService().executor().schedule(
                     () -> {
                         isRescheduleReadInProgress.set(false);
                         readMoreEntries();
                         },
-                    MESSAGE_RATE_BACKOFF_MS, TimeUnit.MILLISECONDS);
+                    readAfterMs, TimeUnit.MILLISECONDS);
         }
     }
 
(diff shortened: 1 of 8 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -85,6 +85,10 @@
  */
 public class PersistentDispatcherMultipleConsumers extends AbstractDispatcherMultipleConsumers
         implements Dispatcher, ReadEntriesCallback {
+    // rescheduling a read after no entries are dispatches will be delayed by this duration using a backoff
+    private static final int RESCHEDULE_READ_INITIAL_DELAY_MS = 100;
+    // maximum delay for rescheduling a read after no entries are dispatched
+    private static final int RESCHEDULE_READ_INITIAL_MAX_DELAY_MS = 5000;
```
dlg99, line 91: I think we should have these parameters configurable

dlg99 (approved):
LGTM but I think the backoff should be configurable

dao-jun (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -671,19 +682,28 @@ public final synchronized void readEntriesComplete(List<Entry> entries, Object c
...
+        updatePendingBytesToDispatch(-totalBytesSize);
+        if (triggerReadingMore) {
+            if (entriesDispatched > 0) {
+                // Reset the backoff when we successfully dispatched messages
+                rescheduleReadBackoff.reset();
+                // Call readMoreEntries in the same thread to trigger the next read
+                readMoreEntries();
+            } else if (entriesDispatched == 0) {
+                // If no messages were dispatched, we need to reschedule a new read with an increasing backoff delay
+                reScheduleReadInMs(rescheduleReadBackoff.next());
```
equanz, line 706: Just a question, the issue is that the `isDispatcherStuckOnReplays` feature\[1\] is causing multiple skips, correct?

* \[1\] https://github.com/apache/pulsar/blob/9a97c843a46e23a0811e2172991cd00a3af642c0/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L555-L561

lhotari: @equanz Exactly. This feature was added in #7105. 

There's a separate PR #23231 to address the problem that this "look ahead" keeps on reading the complete back log in certain cases. To address the complete issue it requires the backoff for retries (this current PR #23226) and the limit for look ahead reads (PR #23231).

equanz: @lhotari I see, thank you.

equanz (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
lhotari: It's possible to reproduce the issue with the instructions in #23200. I also assume that any test application where there's a producer and multiple consumers on a key_shared subscription with a random delay such as this in the processing will reproduce the problem.

```java
                // sleep for a random time with 3% probability
                if (random.nextInt(100) < 3) {
                    try {
                        Thread.sleep(random.nextInt(100) + 1);
                    }
                }
```

@equanz  Without the backoff, the readMoreEntries method call gets called immediately in the code that you pointed out. When the consumers are without permits, this will happen in a tight loop multiple times and in many cases all messages in the backlog get moved to the redelivery controller to replay. That's the problem that this PR and the separate limit PR #23231 address.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
lhotari: This is now addressed. I also made the max delay 1000 instead of 5000.

lhotari:
This introduces a regression that is fixed by #23284

PR #22977, opened 2024-06-25
[fix][broker] Ensure that PulsarService is ready for serving incoming requests

lhotari:
Fixes #22975

### Motivation

Pulsar broker will start serving requests while the broker is starting. This can cause issues and bugs which are hard to reproduce. Serving requests should be delayed until the broker is ready.
The `brokerId` bug #22975 is just one example of problems that this PR will address.

### Modifications

- start Pulsar binary protocol connections with autoread set to false so that request handling can be postponed
- enable autoread after PulsarService has been started
  - add solution to PulsarService for running a function when the service has been started
- add a filter to REST implementation which blocks until PulsarService is started
  - add solution to PulsarService to wait until the service has been started

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23230, opened 2024-08-27
[improve][broker] Add retry for start service unit state channel (ExtensibleLoadManagerImpl only)

Demogorgon314:
### Motivation

When the service unit state channel fails to start, it will throw an exception, and after the PR https://github.com/apache/pulsar/pull/22977 , it will fail forever, because the pulsar broker cannot catch the exception, and it won't have any retry mechanism

### Modifications

Add retry for start service unit state channel

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -420,7 +420,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                         log.warn("The broker:{} failed to start service unit state channel. Retrying {} th ...",
                                 pulsar.getBrokerId(), ++retry, e);
                         try {
-                            TimeUnit.SECONDS.sleep(backoff.next());
+                            Thread.sleep(backoff.next());
                         } catch (InterruptedException ex) {
                             log.warn("Interrupted while sleeping.");
                             // preserve thread's interrupt status
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -125,6 +125,8 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     public static final long COMPACTION_THRESHOLD = 5 * 1024 * 1024;
 
+    public static final int STARTUP_TIMEOUT_SECONDS = 30;
+
     public static final int MAX_RETRY = 5;
 
     private static final String ELECTION_ROOT = "/loadbalance/extension/leader";
@@ -408,7 +410,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
             pulsar.runWhenReadyForIncomingRequests(() -> {
                 Backoff backoff = new BackoffBuilder()
                         .setInitialTime(100, TimeUnit.MILLISECONDS)
-                        .setMax(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS)
+                        .setMax(STARTUP_TIMEOUT_SECONDS, TimeUnit.SECONDS)
                         .create();
                 int retry = 0;
                 while (!Thread.currentThread().isInterrupted()) {
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -100,6 +100,7 @@ import org.apache.pulsar.common.naming.TopicDomain;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.stats.Metrics;
 import org.apache.pulsar.common.util.FutureUtil;
+import org.apache.pulsar.metadata.api.MetadataStoreException;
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/ModularLoadManagerImpl.java
```diff
@@ -879,6 +879,7 @@ private void preallocateBundle(String bundle, String broker) {
                                 k -> ConcurrentOpenHashMap.<String,
                                         ConcurrentOpenHashSet<String>>newBuilder()
                                         .build());
+        Set<String> myConcurrentSet = ConcurrentHashMap.newKeySet();
```
heesung-sn, line 882: Is this change required?

Demogorgon314: Removed, thanks

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -401,10 +404,33 @@ public void start() throws PulsarServerException {
...
+                while (!Thread.currentThread().isInterrupted()) {
+                    try {
+                        brokerRegistry.register();
+                        this.serviceUnitStateChannel.start();
+                        break;
+                    } catch (Exception e) {
+                        log.warn("The broker:{} failed to start service unit state channel. Retrying {} th ...",
+                                pulsar.getBrokerId(), ++retry, e);
+                        try {
+                            TimeUnit.SECONDS.sleep(Math.min(retry * 10L, MAX_ROLE_CHANGE_RETRY_DELAY_IN_MILLIS));
```
lhotari, line 417: There's currently the org.apache.pulsar.common.util.Backoff class for doing exponential backoff with jitter. I think it would be better to use that solution instead of adding another way to do backoffs unless there's a specific reason for this logic.

Demogorgon314: I want to set the max retry count, but since like the `Backoff` doesn't support it.

lhotari: @Demogorgon314 Could you use the Backoff class for the delay calculation and keep the existing logic for max retry count?

Demogorgon314: Sure, it was updated.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -498,8 +530,15 @@ private void failStarting(Exception ex) {
                 this.brokerRegistry, ex);
         if (this.brokerRegistry != null) {
             try {
-                brokerRegistry.close();
-            } catch (PulsarServerException e) {
+                brokerRegistry.unregister();
```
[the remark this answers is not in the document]
Demogorgon314: I use `unregister` here so we can `register` it back when retried.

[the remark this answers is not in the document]
lhotari: > It will be closed after `pulsar.close()`.

where?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -401,10 +406,37 @@ public void start() throws PulsarServerException {
             this.serviceUnitStateChannel.listen(splitManager);
             this.leaderElectionService.start();
             pulsar.runWhenReadyForIncomingRequests(() -> {
-                try {
-                    this.serviceUnitStateChannel.start();
-                } catch (Exception e) {
-                    failStarting(e);
+                Backoff backoff = new BackoffBuilder()
+                        .setInitialTime(100, TimeUnit.MILLISECONDS)
+                        .setMax(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS)
```
RobertIndie, line 411: What's the reason to use `metadataStoreOperationTimeoutSeconds` as the max backoff time?

Demogorgon314: I added a new field `STARTUP_TIMEOUT_SECONDS` as the timeout.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -401,10 +408,37 @@ public void start() throws PulsarServerException {
...
+                        break;
+                    } catch (Exception e) {
+                        log.warn("The broker:{} failed to start service unit state channel. Retrying {} th ...",
+                                pulsar.getBrokerId(), ++retry, e);
+                        try {
+                            Thread.sleep(backoff.next());
+                        } catch (InterruptedException ex) {
+                            log.warn("Interrupted while sleeping.");
+                            // preserve thread's interrupt status
+                            Thread.currentThread().interrupt();
```
gaoran10, line 429: Maybe we need to close the pulsar service if the thread is interrupted.

gaoran10 (approved):

BewareMyPower:
Let me merge it since there is no new response from @lhotari. Please open a new PR if you found something wrong.

PR #21953, opened 2024-01-23
[fix][broker] Introduce the last sent position to fix message ordering issues in Key_Shared (PIP-282)

equanz:
PIP: #20776 

### Motivation

Key_Shared has a mechanism called the "recently joined consumers" to keep message ordering.
However, currently, it doesn't care about some corner cases.
More specifically, we found two out-of-order issues cased by:

1. [issue-1] The race condition in the "recently joined consumers", where consumers can be added before finishing reading and dispatching messages from ledgers.
2. [issue-2] Messages could be added to messagesToRedeliver without consumer-side operations such as unacknowledgement.

We should care about these cases in Key_Shared subscription.
(For more details, please see the PIP document.)

### Modifications

1. Introduce the new position (the lastSentPosition and individuallySentPositions) and use them in PersistentStickyKeyDispatcherMultipleConsumers.
2. Rename the consumer stats `readPositionWhenJoining` to `lastSentPositionWhenJoining`.
3. Modify the subscription stats (consumersAfterMarkDeletePosition) of the definition.

(For more details, please see the PIP document.)

### Verifying this change

- [x] Make sure that the change passes the CI checks.

This change added tests and can be verified as follows:

- Unit tests for the lastSentPosition and individuallySentPositions
- Test for guaranteed message ordering in corner cases

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics
  - https://github.com/apache/pulsar/blob/43a2dc3f9fd9ab68e7429f2345899a6b0cb0bcbe/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L1281-L1288
  - https://github.com/apache/pulsar/blob/43a2dc3f9fd9ab68e7429f2345899a6b0cb0bcbe/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/policies/data/ConsumerStats.java#L72-L73
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/equanz/pulsar/pull/6

PR #21118, opened 2023-09-04
[improve] [pip] PIP-299: Stop dispatch messages if the individual acks will be lost in the persistent storage

poorbarcode:
- [DISCUSS](https://lists.apache.org/thread/2fzo1gnlyd1t5o80g1polbvbx4699r09)
- [VOTE](https://lists.apache.org/thread/j3k97599b3qq0lhmgyp5f93o6ny99cdt)

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #23052, opened 2024-07-18
[improve] [broker] Improve CPU resources usege of TopicName Cache

poorbarcode:
### Motivation

`LoadableCache.get` cost too many CPU resources

[broker_cpu.html.txt](https://github.com/user-attachments/files/16278265/broker_cpu.html.txt)

```
pulsar-io-4-20" #135 prio=5 os_prio=0 cpu=79326083.91ms elapsed=1465985.25s tid=0x00007f5bc40740b0 nid=0xe9 runnable  [0x00007f5ba51fd000]
   java.lang.Thread.State: RUNNABLE
	at java.util.concurrent.ConcurrentLinkedQueue.offer(java.base@17.0.10/ConcurrentLinkedQueue.java:380)
	at java.util.concurrent.ConcurrentLinkedQueue.add(java.base@17.0.10/ConcurrentLinkedQueue.java:283)
	at com.google.common.cache.LocalCache$Segment.recordRead(LocalCache.java:2546)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2068)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4012)
	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:4035)
	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:5011)
	at org.apache.pulsar.common.naming.TopicName.get(TopicName.java:81)
	at org.apache.pulsar.common.naming.SystemTopicNames.isEventSystemTopic(SystemTopicNames.java:62)
	at org.apache.pulsar.common.naming.SystemTopicNames.isSystemTopic(SystemTopicNames.java:86)
	at org.apache.pulsar.common.topics.TopicList.lambda$filterSystemTopic$1(TopicList.java:65)
	at org.apache.pulsar.common.topics.TopicList$$Lambda$2132/0x00007f5d049a4800.test(Unknown Source)
	at java.util.stream.ReferencePipeline$2$1.accept(java.base@17.0.10/ReferencePipeline.java:178)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(java.base@17.0.10/ArrayList.java:1625)
	at java.util.stream.AbstractPipeline.copyInto(java.base@17.0.10/AbstractPipeline.java:509)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(java.base@17.0.10/AbstractPipeline.java:499)
[...]

PR #23231, opened 2024-08-27
[improve][broker] Add limits for Key_Shared Subscription look ahead in dispatching

### LINKED ISSUE #23200: [Bug] Key_Shared subscription could deliver messages late and nonContiguousDeletedMessagesRange could exceed managedLedgerMaxUnackedRangesToPersist

lhotari:
Fixes #23200 

### Motivation

This PR introduces enhancements to the Key_Shared subscription dispatching logic in the Pulsar broker. The changes aim to optimize message dispatching, improve the handling of slow consumers, and address several identified issues related to message replay and message "look ahead" when consumers are blocked in delivery due to key ordering restrictions.

This PR adds limits to the Key_Shared Subscription look-ahead feature, which was introduced in PR #7105. Additionally, there are several improvements to the previous logic.

Since there wasn't any named concept for the feature added in PR #7105, it has been named "key shared look ahead" in this PR. This references the feature added in PR #7105, where the dispatcher will skip replaying messages and read more messages from the backlog. The changes in PR #7105 didn't add a limit for reading more messages. The comments in the PR mention that it will be limited by the unacked message limits. However, this isn't the case since the read messages aren't considered unacked messages until they have been dispatched to consumers.

Reading ahead would be triggered in multiple cases:
* When consumers are backpressured due to slow processing or message delivery
* When consumers are blocked due to key ordering restrictions
  * The previous logic for blocking a hash was unoptimal. The impact is discussed in a [Pulsar Q&A discussion](https://github.com/apache/pulsar/discussions/22912#discussioncomment-9808993). It seemed that a hash range got blocked, but that wasn't the case.
     * The replay queue had a head-of-line blocking issue, which caused many hashes being blocked unnecessarily
     * In reading more entries, all entries got dropped when any of the read entries was blocked by a hash. 
       * This is unnecessary for retaining ordering. This is addressed in this PR.
         * It seems that addressing this issues depends on PIP-282 changes, #21953.

Since consumers were blocked unnecessarily, reading ahead would often just continue reading the complete backlog into the replay queue. The impact is described in #23200 
* Reading the entries to the replay queue causes unnecessary network traffic, disk IO & CPU on bookies and CPU on brokers. The read entries will be dropped in most cases.
  * Issue #22245 has been a previous attempt to mitigate this symptom
* Reading ahead would cause a lot of "ack holes".
  * PIP-299 #21118 has been a previous attempt to mitigate this symptom

In addition to fix the infinite key shared look ahead issue, this PR seems to fix a problem in the existing implementation where this code blocks progress of a consumer:
https://github.com/apache/pulsar/blob/d7e8ea16e6682df9a9354cda25cf4f1f9cb54429/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L453-L458
What this means is that when the dispatcher has read a batch, it will discard all messages in the read messages if any hash is present in the replay queue (redelivery messages). This is an unoptimal solution for ensuring that ordering is preserved. This PR addresses the issue while attempting to keep backwards compatibility from a single consumer's perspective.

The "look ahead" limit introduced in this PR will add backpressure to how the dispatcher works. Eventually consumers will be slower than the dispatcher sending messages to consumers and that's why it's necessary to have the limit, also in cases where the hashes wouldn't be blocked or "allowOutOfOrderDelivery" mode is used. 

There's also another related problem that this change addresses. The head of the replay queue might get blocked in the release Pulsar versions.
https://github.com/apache/pulsar/blob/ed14f21de94c6af2bfae5318a014c56fac8a1a21/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L348-L350

In this PR, the solution to this is to take as many entries that match the criteria from the replay queue. This required changing the getMessagesToReplayNow to accept a Predicate for filtering:
```java
    public NavigableSet<Position> getMessagesToReplayNow(int maxMessagesToRead, Predicate<Position> filter) {
        NavigableSet<Position> items = new TreeSet<>();
        messagesToRedeliver.processItems(PositionFactory::create, item -> {
            if (filter.test(item)) {
                items.add(item);
            }
            return items.size() < maxMessagesToRead;
        });
        return items;
    }
```

### Modifications

1. **Configuration Properties:**
    - Added new configuration properties to control the limits of Key_Shared subscriptions look ahead in dispatching:
      - `keySharedLookAheadMsgInReplayThresholdPerConsumer`
      - `keySharedLookAheadMsgInReplayThresholdPerSubscription`
      - `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist`

2. **Broker Configuration:**
    - Updated `broker.conf` to include descriptions and default values for the new configuration properties.

3. **Key_Shared Subscription Logic:**
    - Implement the key shared look-ahead limits that are configured with the above configuration properties.
    - Enhance the dispatching to ensure efficient dispatching:
      - Instead of filtering the list of entries pulled from the message redelivery controller, refactor the solution to take a function that can filter the entries until there's a sufficient amount of messages for dispatching.
        - This refactoring eliminates the previous method `filterOutEntriesWillBeDiscarded` while preserving the previous logic of filtering.
      - Remove the unnecessary thread-local fields `localGroupedEntries` and the unused `localGroupedPositions`.
        - It is unnecessary to eliminate the creation of short-lived objects in modern JVMs. It just clutters the code without any performance benefit. "No garbage" style is relevant mainly for long-lived objects stored in caches or large array allocations.
    - Replace `getMessagesToReplayNow(1)` with `getFirstPositionInReplay()`.
      - This check will no longer have the side effect of pulling messages from delayed delivery when it's called.
    - Add a new method `canReplayMessages` to make the look-ahead logic more explicit. The previous solution of returning an empty set from the `getMessagesToReplayNow` method made it hard to understand how the look-ahead behavior is implemented.
    - Rename the `isDispatcherStuckOnReplays` field to `skipNextReplayToTriggerLookAhead` to make the meaning of the field explicit and self-descriptive.
    - Fix issue where the `!havePendingReplayRead` status wasn't checked before starting a new read.
      - For shared subscription, it's fine to just check `!havePendingPendingRead`, but for Key_Shared it isn't.
      - extract doesntHavePendingRead for handling the logic
    - Rename the `keyNumbers` variable to `remainingConsumersToFinishSending` so that the meaning of the variable is explicit and self-descriptive.
    - Rename the `groupedEntries` variable to `entriesByConsumerForDispatching` so that the meaning of the variable is explicit and self-descriptive.
    - Rename the `entriesWithSameKey` variable to `entriesForConsumer` so that the variable is self-descriptive.
    - Fix calculations based on `maxUnackedMessages` and `unackedMessages`. The unit is in individual messages and permits are in entries. These 2 cannot be mixed.
    - Addressing the consumer blocked issue: the logic has been changed to evaluate each message hash one-by-one.

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
This PR is ready for initial review feedback. I'm looking into ways of adding tests for the key_shared look ahead limits.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -436,36 +370,159 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
+        Position maxLastSentPosition = resolveMaxLastSentPositionForRecentlyJoinedConsumer(consumer, readType);
+        if (maxLastSentPosition != null && entry.getPosition().compareTo(maxLastSentPosition) > 0) {
+            return false;
+        }
+
+        // If redeliveryMessages contains messages that correspond to the same hash as the entry to be dispatched
+        // do not send those messages for order guarantee
+        if (readType == ReadType.Normal && redeliveryMessages.containsStickyKeyHash(stickyKeyHash)) {
+            return false;
         }
```
lhotari, line 443: This is a solution to another problem that exists. All progress got blocked when any of the messages that were read were in the replay queue:
https://github.com/apache/pulsar/blob/d7e8ea16e6682df9a9354cda25cf4f1f9cb54429/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L453-L458
This defeats the purpose of the look ahead feature added in #7105 and makes the problem worse.
This problem is also fixed in this PR.

lhotari:
Tomorrow I'll continue and add tests and fix broken tests. I'd appreciate any reviews along the way.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -949,6 +949,48 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    @FieldContext(
+            category = CATEGORY_POLICIES,
+            doc = "For Key_Shared subscriptions, if messages cannot be dispatched to consumers due to a slow consumer"
+                    + " or a blocked key hash (because of ordering constraints), the broker will continue reading more"
+                    + " messages from the backlog and attempt to dispatch them to consumers until the number of replay"
+                    + " messages reaches the calculated threshold.\n"
+                    + "This setting controls whether look ahead is enabled when recently joined consumers are present.",
+            dynamic = true
+    )
+    private boolean keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist = false;
```
merlimat, line 992: Is this going to change existing behavior, with the default setting?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -439,36 +373,159 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
+            // if out of order delivery is allowed, then any position will be replayed
+            if (isAllowOutOfOrderDelivery()) {
+                return true;
+            }
+            // lookup the sticky key hash for the entry at the replay position
+            Long stickyKeyHash = redeliveryMessages.getHash(position.getLedgerId(), position.getEntryId());
+            if (stickyKeyHash == null) {
+                // the sticky key hash is missing for delayed messages, the filtering will happen at the time of
+                // dispatch after reading the entry from the ledger
+                log.debug("[{}] replay of entry at position {} doesn't contain sticky key hash.", name, position);
```
merlimat, line 481: nit: 

```suggestion
                if (log.isDebugEnabled()) {
                    log.debug("[{}] replay of entry at position {} doesn't contain sticky key hash.", name, position);
                }
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -439,36 +373,159 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
+            if (consumer == null) {
+                return false;
+            }
+            // lookup the available permits for the consumer
+            MutableInt availablePermits =
+                    availablePermitsMap.computeIfAbsent(consumer,
+                            k -> new MutableInt(getAvailablePermits(consumer)));
+            // skip replaying the message position if the consumer has no available permits
+            if (availablePermits.intValue() <= 0) {
+                return false;
```
merlimat, line 496: If we stop here because 1 consumer has no permit, wouldn't it stop the replay for everyone? Or is the context for a single consumer?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -439,36 +373,159 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
+    protected Predicate<Position> createFilterForReplay() {
+        return new ReplayPositionFilter();
+    }
+
+    /**
+     * Filter for replaying messages. The filter is stateful for a single invocation and shouldn't be cached, shared
+     * or reused. This is a short-lived object, and optimizing it for the "no garbage" coding style of Pulsar is
+     * unnecessary since the JVM can optimize allocations for short-lived objects.
+     */
+    private class ReplayPositionFilter implements Predicate<Position> {
```
merlimat, line 465: Maybe we could export this to a separate file and have an explicit variable pointing back to the dispatcher.

merlimat (approved):
Very nice. LGTM

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari: no, that's what exists in the current released versions that it doesn't "look ahead" when recently joined consumers exist. I guess we need to experiment whether this could be changed to do the look ahead also in that case. I was cautious about changing the existing behavior and that's why I added this flag to control the behavior. We'd need to experiment and analyze whether the default value could be changed without risks of regressions. I'm trying to minimize the risks of regressions.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: I tried that but there are a few private methods and making them accessible would have a negative impact.

lhotari:
There are some test failures. Trying to understand what org.apache.pulsar.client.api.KeySharedSubscriptionTest#testCheckBetweenSkippingAndRecentlyJoinedConsumers is testing.
https://github.com/apache/pulsar/blob/dccc06bf50bb5ca510b39167908c02d2b4602ca5/pulsar-broker/src/test/java/org/apache/pulsar/client/api/KeySharedSubscriptionTest.java#L1235-L1333
"Test for guaranteed message ordering in corner cases" says the PR.
@equanz Do you remember which corner case this test method is for?

lhotari:
Found the cause of SimpleProducerConsumerTest#testMaxUnAckMessagesLowerThanPermits failure. It's about available permits handling. The previous implementation before this PR ignored the batch index count so I skipped the logic. There's a lot of duplication required at the moment, so it becomes a broader refactoring to address this properly.

lhotari:
The remaining part is to add tests that reproduce the issues that this PR is addressing.

dlg99 (approved):
LGTM

dao-jun (approved):
Great work! LGTM

nodece (approved):

equanz:
https://github.com/apache/pulsar/pull/23231#issuecomment-2321233767

This test verifies that the issue referred to in PIP-282 as `[issue-2]` is not reproduced.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -439,36 +373,159 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
         return false;
     }
 
-    private int getRestrictedMaxEntriesForConsumer(Consumer consumer, List<? extends Position> entries,
-           int availablePermits, ReadType readType, Set<Integer> stickyKeyHashes) {
-        int maxMessages = Math.min(entries.size(), availablePermits);
-        if (maxMessages == 0) {
-            return 0;
+    private boolean isLookAheadAllowed() {
+        if (serviceConfig.isKeySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist()
```
equanz, line 390: In my understanding, the selector will select also the consumer that recently joined. Messages that will be selected to the new consumer are not dispatched by the recentlyJoinedConsumers feature. It may make large ack-hole(I haven't verified this, so it could be different...).

But I don't think it will be a major problem because this PR introduces the limit, and this feature is disabled by default.

equanz (approved):
Great work!
(I wrote some memos for review, but please ignore them.)

lhotari:
One of the details that I'll have to address is the fact that the Pulsar client consumer sends permits for entries/batches and not individual messages. This detail isn't accurately described in the [Pulsar binary protocol specification](https://pulsar.apache.org/docs/next/developing-binary-protocol/#consumer).

lhotari:
> One of the details that I'll have to address is the fact that the Pulsar client consumer sends permits for entries/batches and not individual messages. This detail isn't accurately described in the [Pulsar binary protocol specification](https://pulsar.apache.org/docs/next/developing-binary-protocol/#consumer).

I noticed that the calculations based on available permits and available unacknowledged messages were wrong for both Shared and Key_Shared subscriptions. The fixes are included in this PR for both subscription types. Tests break unless the fixes are made for Key_Shared subscriptions and since the fix is made for Key_Shared, it should also be applied for Shared subscriptions. This does show a lack of proper test coverage to catch regressions in this area so it's not the most optimal solution to roll the changes in this PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: I've removed the limitation in the most recent commits.

liudezhi2098 (approved):
Great work!

lhotari:
> The remaining part is to add tests that reproduce the issues that this PR is addressing.

I had the goal to add some system level tests, but that seems hard. I'll try to add some tests to verify the behavior by adding tests to the existing Key_Shared test suite which uses partial mocks.

poorbarcode (changes requested):
Please do not review this PR fast, I will finish my review tomorrow

lhotari:
> Please do not merge this PR fast, I will finish my review tomorrow

Thanks @poorbarcode. The tests are also pending. I'm planning to finish them tomorrow and hopefully your feedback will be available for addressing at that time too.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -439,36 +372,162 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
+        Map<Consumer, List<Entry>> entriesGroupedByConsumer = new HashMap<>();
+        // permits for consumer, permits are for entries/batches
+        Map<Consumer, MutableInt> permitsForConsumer = new HashMap<>();
+
+        for (Entry entry : entries) {
+            int stickyKeyHash = getStickyKeyHash(entry);
+            Consumer consumer = selector.select(stickyKeyHash);
+            boolean dispatchEntry = false;
+            // a consumer was found for the sticky key hash and the entry can be dispatched
+            if (consumer != null && canDispatchEntry(consumer, entry, readType, stickyKeyHash)) {
```
poorbarcode, line 413: > Addressing the consumer blocked issue: the logic has been changed to evaluate each message hash one-by-one. this is the code:

- This change has a positive effect on a normal read.
- But it has a negative effect on replay read, it also checks entries one by one.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Two things
- Even if `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist` is `false`, the broker will still look ahead and read entries, because it only affects when `messagesSent == 0`, see https://github.com/apache/pulsar/pull/23231/files#diff-c48d5c94842ac8c9a0c9314b207298069f38c8dcfeda4a9886fb3bb1f77843f2R362. 
- Since you set `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist` to `false`, all tests will run with `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist = true`, so we have no tests cover the cases that `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist = false`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -183,22 +180,6 @@ public synchronized void removeConsumer(Consumer consumer) throws BrokerServiceE
         }
     }
 
-    private static final FastThreadLocal<Map<Consumer, List<Entry>>> localGroupedEntries =
```
poorbarcode, line 186: > - Remove the unnecessary thread-local fields localGroupedEntries and the unused localGroupedPositions.
>   - It is unnecessary to eliminate the creation of short-lived objects in modern JVMs. It just clutters the code without any performance benefit. "No garbage" style is relevant mainly for long-lived objects stored in caches or large array allocations.

I am thinking about whether the "No garbage" design of Netty is unnecessary.

poorbarcode (commented):
Since we added new configs, could you add tests to cover the new configs?

poorbarcode:
And a new comment here: https://github.com/apache/pulsar/pull/23231#discussion_r1746465968

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: > But it has a negative effect on replay read, it also checks entries one by one.

@poorbarcode what is the negative effect? Checking entries one-by-one by calling a separate method isn't inefficient in this case. The JVM's JIT compiler can optimize calls and run very efficiently. It would be a different matter if IO or locks would be involved. In those cases there's a benefit of batching.

poorbarcode: > The JVM's JIT compiler can optimize calls and run very efficiently. It would be a different matter if IO or locks would be involved. In those cases there's a benefit of batching.

I think JIT can not optimize it very efficiently.

### 1. The method `canDispatchEntry` does a lot of things, such as follows:
- Check whether the param-position is lower than `recentlyJoinedConsumers`
- Call `removeConsumersFromRecentJoinedConsumers`

### 2. We made many improvements that save CPU resources by removing unnecessary loop
- https://github.com/apache/pulsar/pull/20607
- https://github.com/apache/pulsar/pull/19167
- https://github.com/apache/pulsar/pull/23052

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
poorbarcode: Since we had this optimization before, we should leave them there. There are many cases using `recycle objects` in our project, if this place can be removed, do you think that others can also be removed?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: thanks for the context @poorbarcode . I see now what I need to change to optimize it.
I'll cache the result of the resolveMaxLastSentPositionForRecentlyJoinedConsumer call.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: > if this place can be removed, do you think that others can also be removed?

Recycling is a waste for short-lived objects in modern JVMs. In those cases, the JVM can optimize the allocations and memory management so that it doesn't really add essential GC overhead. Recycling has higher overhead in those cases. For long living objects, there's a potential benefit. There are other benchmarks from a few years ago in the context of Netty and the performance benefit of recycling over not recycling was very marginal (<10%). Generational ZGC is even better so it might be an unnecessary overhead to do the recycling in most cases. We'd have to confirm with benchmarking in some cases.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: @poorbarcode Thanks for catching the possible performance regression before it came one! Please check recent changes in 88256d6a4c8659cee759ae90851cab1ee4a1ef21, this concern should be addressed.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari: > Even if keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist is false, the broker will still look ahead and read entries, because it only affects when messagesSent == 0

Yes, I'll cover this in the documentation. That's the reason why I'm calling it "look ahead". I'll find a way to improve this.

lhotari: > * Since you set the default value of `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist` to `false`, all tests will run with `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist = false`, so we have no tests cover the cases that `keySharedLookAheadEnabledWhenRecentlyJoinedConsumersExist = true`.

Valid point. I'm planning to remove this and make look ahead enabled when recently joined consumers exist if the reason why a message couldn't be dispatched was caused by a blocked hash. This will make the solution more robust too.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: I'll also address this for replays. I'm working on a larger refactoring, will push soon.

lhotari:
Noticing more challenges to cover while getting into the tests. Setting this to draft mode until the concerns are addressed and tests have been added. Some of the concerns are related to #23264.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: @poorbarcode The performance regression concerns about canDispatchEntry are now addressed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
lhotari: I'm considering this part resolved.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java
```diff
@@ -274,96 +251,58 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
...
 
-                TOTAL_AVAILABLE_PERMITS_UPDATER.getAndAdd(this,
-                        -(sendMessageInfo.getTotalMessages() - batchIndexesAcks.getTotalAckedIndexCount()));
-                totalMessagesSent += sendMessageInfo.getTotalMessages();
-                totalBytesSent += sendMessageInfo.getTotalBytes();
-            } else {
-                currentThreadKeyNumber = keyNumbers.decrementAndGet();
-            }
+            TOTAL_AVAILABLE_PERMITS_UPDATER.getAndAdd(this,
+                    -(sendMessageInfo.getTotalMessages() - batchIndexesAcks.getTotalAckedIndexCount()));
```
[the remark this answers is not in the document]
lhotari: reported as separate issue #23263

poorbarcode:
@lhotari 

- `Two weeks ago`: I submitted a Request Change

- `Two weeks ago`, you left a comment: Noticing more challenges to cover while getting into the tests. Setting this to draft mode until the concerns are addressed and tests have been added. Some of the concerns are related to https://github.com/apache/pulsar/issues/23264.

- `5 days ago`: you dismissed the Request Change, and left a comment: "Outdated comment"

Between `step-2` and `step-3`, you never marked it as available to review and never ask to review again, then you merged the PR :joy:

lhotari:
I'm sorry @poorbarcode. Please let me explain the misunderstanding.

> 5 days ago: you dismissed the Request Change, and left a comment: "Outdated comment"

For me, the "request for changes" comment was "Please do not merge this PR fast, I will finish my review tomorrow" (https://github.com/apache/pulsar/pull/23231#pullrequestreview-2283712399). That's why I dismissed it with "Outdated comment". I had already addressed also the other review comments you had made.

> Between `step-2` and `step-3`, you never marked it as available to review and never ask to review again, then you merged the PR 😂

I'm sorry for that. I did request a review, but you didn't respond. You can check the timeline. The PR was in draft mode for some time, but the reason for that was that the PR was already approved and I didn't want anyone to merge it. I didn't realize that you wouldn't be reviewing it because of the draft status.

The changes in this PR aren't final, if you have review feedback, it's possible also address that after the PR has been merged. Please add comments to this PR of any concerns you may have.

poorbarcode:
@lhotari 

Sure, thanks for replying this comment. ❤️

PR #23235, opened 2024-08-29
[improve][pip] PIP-374: Visibility of messages in receiverQueue for the consumers

touchkey:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pip/pip-374.md
+++ b/pip/pip-374.md
@@ -1,0 +1,71 @@
+# PIP-374: Visibility of messages in receiverQueue for the consumers
+
+# Background knowledge
+
+When a consumer connects to the Broker, the broker starts dispatching the messages based on receiverQueueSize configured. 
+There is no observability for the messages arrived on the consumer side if the user didn't call the receive method. It leads to ambiguities at times as
+the consumer application does not know whether the message was actually sent by the broker or is it lost in the network or is it lost in the receiver queue.
+
+ConsumerInterceptors is a plugin interface that  intercept and possibly mutate messages received by the consumer.
+
+
+# Motivation
+
+* We need to receive queue filling of the event as the particular message is already on particular consumer's receiver queue and waiting for the consumer to pickup and process. It may wait in the recieverQueue longer if the consumer processing takes more time. It's very important to provide the visibility of the messages that are waiting in receiverQueue for processing.
+
+* Availability of a consumer application w.r.t any messaging system depends on the number of messages dispatched from the server/broker against the number of messages acknowledged from the consumer app. This metric defines the processing rate of a consumer.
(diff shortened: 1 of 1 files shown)


coderzc:
@touchkey Please reference https://github.com/apache/pulsar/blob/master/pip/README.md, email the [dev@pulsar.apache.org](mailto:dev@pulsar.apache.org) mailing list to kick off a discussion.

pip/pip-374.md
```diff
@@ -0,0 +1,71 @@
...
+
+* ConsumerInterceptor.java
+```
+default Message<T> onArrival()(Consumer<T> consumer, Message<T> message){
+ return message;
+}
+
+```
+
+* Add hook in ConsumerImpl.messageReceived which calls onArrival method which calculates the the number of message received.
```
shibd, line 54: There is a lot of business logic in the `ConsumerImpl.messageReceived` method. If we need to pass a `Message` object to `onArrival`, we must ensure that the message has been correctly parsed.

However, before correctly parsing the message object, there is a lot of logic that may skip this message. 
https://github.com/apache/pulsar/blob/d4bbf10f58771e2d43e576dc3422e502834b1de4/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1349-L1423

I want to say that this might not align with the expected behavior defined by this method(onArrival).

Maybe we can modify the method to `onArrival()(Consumer<T> consumer, MessageId messageId)`. When the message just arrives at this method, we can get the messageId and then inform the user.

coderzc: cc @touchkey

shibd: I checked the code again, and our internal method operates at the message level, so the above suggestion might not be appropriate. However, we may need to clarify this in the method comments. The `onArrival` method is called after the message has been parsed.

shibd (approved):

coderzc (approved):

pip/pip-374.md
```diff
@@ -0,0 +1,71 @@
...
+
+Since we added a default method onArrival() in interface, one who has provided the implementations for ConsumerInterceptor will not get any compile time error as it has default implementation. If user wants to give implementation from his side, he can override and provide implementation.
+
+# Links
+
+<!--
+Updated afterwards
+-->
+* Mailing List discussion thread:
+* Mailing List voting thread:
```
coderzc, line 71: @touchkey Please add email link

shibd: I will add it to code PR.

PR #23248, opened 2024-09-03
[improve][pip] PIP-376: Make topic policies service pluggable

BewareMyPower:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


BewareMyPower:
Here is the draft implementation in my fork: https://github.com/BewareMyPower/pulsar/pull/35

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+## Topic policies service and system topics
+
+[PIP-39](https://github.com/apache/pulsar/wiki/PIP-39%3A-Namespace-Change-Events) introduces system topics and the topic level policies. However, the topic policies service (`TopicPoliciesService`) only has one implementation (`SystemTopicBasedTopicPoliciesService`) that depends on the system topics. So the following configs are both required (though they're all enabled by default now):
+
+```properties
+systemTopicEnabled=true
+topicLevelPoliciesEnabled=true
+```
+
+However, if the Pulsar storage is switched to a S3-based solution (by modifying the `managedLedgerStorageClassName` config), using system topics to manage topic policies could have low performance (due to the S3 write and read latency) and higher cost (due to redundant S3 API calls).
```
eolivelli, line 14: >  if the Pulsar storage is switched to a S3-based solution
is this a custom thing ?
I know that `managedLedgerStorageClassName` is configurable, but it is not properly a place where it is possible to code custom implementations because the API boundaries are not clear (as for Protocol Handlers, you end up in using internal classes)

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Badly designed TopicPoliciesService interface
```
eolivelli, line 16: I agree, this is a common problem across the various classnames that we made configurable

eolivelli (commented):
I generally support this proposal, we should have clearer APIs for the components we want to allow users to customize

dao-jun:
I support

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+
+`getTopicPoliciesBypassCacheAsync` is only used in tests. This method just creates a reader to replay the `__change_events` topic and construct the topic policies map.
+
+2. Confusing and inconsistent `getTopicPolicies` family
+
+There are two overrides of `getTopicPolicies`:
+
+```java
+TopicPolicies getTopicPolicies(TopicName topicName, boolean isGlobal) throws TopicPoliciesCacheNotInitException;
+TopicPolicies getTopicPolicies(TopicName topicName) throws TopicPoliciesCacheNotInitException;
```
codelipenghui, line 32: It looks like this one is only used by tests, we can remove if from the interface

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+`addOwnedNamespaceBundleAsync` and `removeOwnedNamespaceBundleAsync` are only used internally in `SystemTopicBasedTopicPoliciesService`.
+
+`getTopicPoliciesBypassCacheAsync` is only used in tests. This method just creates a reader to replay the `__change_events` topic and construct the topic policies map.
+
+2. Confusing and inconsistent `getTopicPolicies` family
+
+There are two overrides of `getTopicPolicies`:
+
+```java
+TopicPolicies getTopicPolicies(TopicName topicName, boolean isGlobal) throws TopicPoliciesCacheNotInitException;
```
codelipenghui, line 31: It looks like this one is only used by `SystemTopicBasedTopicPoliciesService` internally and tests, we can remove it from the interface.

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+The 2nd method is equivalent to `getTopicPolicies(topicName, false)`.
+
+The semantics of these two methods are very intuitive. First, they are not synchronous methods that are blocked by waiting a future. They just start an asynchronous policies initialization (creating a reader to replay the `__change_events` topic), and then try to get the policies from the cache. If the asynchronous policies initialization didn't start, just throw `TopicPoliciesCacheNotInitException`.
+
+As you can see, these two methods are hard to use. And they are also only used in tests except for the `getTopicPoliciesAsyncWithRetry` method, which uses a user-provided executor and backoff policy to call `getTopicPolicies` until `TopicPoliciesCacheNotInitException` is not thrown:
+
+```java
+    default CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsyncWithRetry(TopicName topicName,
+              final Backoff backoff, ScheduledExecutorService scheduledExecutorService, boolean isGlobal) {
+```
```
codelipenghui, line 44: Can we remove this one from the interface? Since this is added for fixing the bug from the SystemTopic based solution. And finally, we should also improve it with a pure async way for the SystemTopic based solution, but it's not required in this proposal.

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+
+https://github.com/apache/pulsar/pull/21231 adds two asynchronous overrides that are much more friendly to users:
+
+```java
+CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsync(@Nonnull TopicName topicName, boolean isGlobal);
+CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsync(@Nonnull TopicName topicName);
+```
+
+Now we have **5** asynchronous get methods. What's worse, unlike `getTopicPolicies`, `getTopicPoliciesAsync(topic)` is not equivalent to `getTopicPoliciesAsync(topic, false)`, instead,
+- `getTopicPoliciesAsync(topic)` will try getting local policies first, if absent, then try getting global policies
```
codelipenghui, line 56: Is it better to not expose this method? Users should only implement 

```
CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsync(@Nonnull TopicName topicName, boolean isGlobal);
```

Let's the broker decide when to get local topic policy or global policy. At lease for now, I don't think users should customize this behavior

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+Currently,
+- `getTopicPoliciesAsync(TopicName)` is used in `BrokerService#getTopicPoliciesBypassSystemTopic`, which is called when initializing the topic policies of `PersistentTopic` objects. So it uses the "local-first" semantics in case the global policies or local policies is deleted.
+- `getTopicPoliciesAsyncWithRetry` is used in `AdminResource#getTopicPoliciesAsyncWithRetry`, which is called for all topic policies admin APIs. Since these admin APIs all have a `isGlobal` field to indicate whether to get the global policies, it uses the "local only" or "global only" semantics.
+- Other methods are never called directly other than tests.
+
+Actually there is the 6th method `getTopicPoliciesIfExists`, which just tries to get the local topic policies from the cache. This method is the most clear and simple in all these stuffs.
+
+```java
+    TopicPolicies getTopicPoliciesIfExists(TopicName topicName);
+```
```
codelipenghui, line 71: This one is also confusing for plugin. If there is no policies for a topic, user can also implement by 

```
CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsync(@Nonnull TopicName topicName, boolean isGlobal);
```

I remember it's also added for resolving the issues from SystemTopic based solution which don't want to wait for the message replay from the system topic.

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+1. Add a unified method to get topic policies.
+
+```java
+    enum GetType {
+        LOCAL_FIRST, // try getting the local topic policies, if not present, then get the global policies
+        GLOBAL_ONLY, // only get the global policies
+        LOCAL_ONLY,  // only get the local policies
+    }
+    CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsync(TopicName topicName, GetType type);
+```
```
codelipenghui, line 100: Do we really need to expose LOCAL_FIRST? The priority of the policies defined in Pulsar

local_topic > global_topic > namespace > dymamic_config > broker.conf

Do we want users to customize the priority? If no, we can only let users to implement the method to get local policies and global policies.

codelipenghui (commented):
@BewareMyPower Thanks for driving the proposal
Could you please also add a section to describe what the TopicPoliciesService finally looks like?

pip/pip-376.md
BewareMyPower: Yeah, this section just talks about the existing interface.

pip/pip-376.md
BewareMyPower: Agreed. I will change the specific case here.

pip/pip-376.md
BewareMyPower: Let me recheck the references of this method.

pip/pip-376.md
BewareMyPower: The issue is actually related to where this API could be called. See the documents before:

> - `getTopicPoliciesAsync(TopicName)` is used in `BrokerService#getTopicPoliciesBypassSystemTopic`, which is called when initializing the topic policies of `PersistentTopic` objects. So it uses the "local-first" semantics in case the global policies or local policies is deleted.
> - `getTopicPoliciesAsyncWithRetry` is used in `AdminResource#getTopicPoliciesAsyncWithRetry`, which is called for all topic policies admin APIs. Since these admin APIs all have a `isGlobal` field to indicate whether to get the global policies, it uses the "local only" or "global only" semantics.

In `BrokerService` needs to read the policies according to the priority: `local > global`. However, for admin APIs, the current behavior is to read either local or global policies.

> Do we want users to customize the priority? 

The answer is no. But users need to support these three types. For example, if the customized topic policies do not support global policies, this method should return a failed future when the type is `GLOBAL_ONLY`.

pip/pip-376.md
BewareMyPower: > Let's the broker decide

That's why I design the `getTopicPoliciesAsync` interface like

```java
CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesAsync(TopicName topicName, GetType type);
```

- If broker side just want to get policies according to the priority, it will call `getTopicPoliciesAsync(topic, LOCAL_FIRST)`.
- If broker side only want to get local policies or global policies, it will call `getTopicPoliciesAsync(topic, LOCAL_ONLY)` or `getTopicPoliciesAsync(topic, GLOBAL_ONLY)`.

See
- https://github.com/BewareMyPower/pulsar/blob/02bffa26332ee48212c50eb985aeb683825f9055/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java#L367-L371
- https://github.com/BewareMyPower/pulsar/blob/02bffa26332ee48212c50eb985aeb683825f9055/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1176-L1179

pip/pip-376.md
```diff
@@ -0,0 +1,138 @@
...
+```properties
+systemTopicEnabled=true
+topicLevelPoliciesEnabled=true
+```
+
+However, if the Pulsar storage is switched to a S3-based solution (by modifying the `managedLedgerStorageClassName` config), using system topics to manage topic policies could have low performance (due to the S3 write and read latency) and higher cost (due to redundant S3 API calls).
+
+## Badly designed TopicPoliciesService interface
+
+The `TopicPoliciesService` interface is a terrible abstraction because it's never designed for 3rd party implementations.
```
lhotari, line 18: For the PIP document, it's good to remember that it remains as part of the Pulsar project documentation. Adapting the style to a professional documentation style would be helpful here.

BewareMyPower:
Now I've refactored the proposal, as well as the demo implementation with the following changes:
1. Do not mention other plugins. Just mention the default implementation might not always be best.
2. Change `LOCAL_FIRST` to `DEFAULT`.
3. List the whole interface after this change.
4. Refactored the document with the help of GitHub Copilot.

@eolivelli @codelipenghui @lhotari PTAL again

lhotari (approved):
LGTM. Good work @BewareMyPower!

pip/pip-376.md
```diff
@@ -0,0 +1,222 @@
...
+@InterfaceStability.Stable
+public interface TopicPolicyListener {
+
+  void onUpdate(TopicPolicies data);
+}
+```
+
+### Configuration
+
+Add a new configuration `topicLevelPoliciesEnabled`.
```
RobertIndie, line 205:
```suggestion
Add a new configuration `topicPoliciesServiceClassName `.
```

This new configuration should be `topicPoliciesServiceClassName `, right?

Issue #23250, opened 2024-09-03
[Bug] Broker seeing high latency, high CPU and high GC Pauses


rdhabalia:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

> 2.10

### Minimal reproduce step

1. Create global topic with unreachable remote replication cluster
2. Topic tries to start replicator producer which keeps failing to start and start logging large amount of data by creating large number of Exception objects

```
final String namespace = BrokerTestUtil.newUniqueName("pulsar/concurrent");
updateTenantInfo("pulsar",
        new TenantInfoImpl(Sets.newHashSet("appid1", "appid2", "appid3"),
                Sets.newHashSet("r1", "r2", "r3", "r10")));
admin1.namespaces().createNamespace(namespace);
admin1.namespaces().setNamespaceReplicationClusters(namespace, Sets.newHashSet("r1", "r10"));
final TopicName topicName = TopicName.get(BrokerTestUtil.newUniqueName("persistent://" + namespace + "/topic"));

@Cleanup
PulsarClient client1 = PulsarClient.builder().serviceUrl(url1.toString()).statsInterval(0, TimeUnit.SECONDS)
        .build();
Producer<byte[]> producer = client1.newProducer().topic(topicName.toString()).enableBatching(false)
        .messageRoutingMode(MessageRoutingMode.SinglePartition).create();
```

3. You can also reproduce by creating producer with unreachable connection url
```
@Cleanup
PulsarClient newPulsarClient = PulsarClient.builder()
    .serviceUrl(lookupUrl.toString())
    .lookupTimeout(1, TimeUnit.MINUTES)
    .build();

ProducerBuilder<byte[]> producerBuilder = newPulsarClient.newProducer()
        .topic("persistent://my-property/my-ns/my-topic1");


Producer<byte[]> producer = producerBuilder.create();
stopBroker();
for (int i = 0; i < 10; i++) {
    String message = "my-message-" + i;
    producer.sendAsync(message.getBytes());
}
```

This will cause below issues in Broker

1. Broker is busy in generating Exception toString() message
```
[...]

PR #11627, opened 2021-08-10
PIP-91: Separate lookup timeout from operation timeout

ivankelly:
This patch contains a number of changes.

TooManyRequests is retried for partition metadata and lookups

Lookup timeout configuration has been added. By default it matches
operation timeout.

Partition metadata timeout calculation has been fixed to calculate
the elapsed time correctly.

Small refactor on broker construction to allow a mocked ServerCnx
implementation for testing. Unfortunately, the test takes over 50
seconds, but this is unavoidable due to the fact that we're working
with timeouts here.

PulsarClientExceptions have been reworked to contain more
context (remote/local/reqid) and any previous exceptions which may
have occurred triggering retries. The previous exceptions must be
manually recorded, so this only applies to lookups on the consumer
side for now.

PR #23251, opened 2024-09-03
[fix][client] Fix broker/Client CPU reaching 100% during retriable connection failure

rdhabalia:
Fixes : https://github.com/apache/pulsar/issues/23250

### Motivation

[PIP-91](https://github.com/apache/pulsar/pull/11627/) had added a change to accumulate list of Exceptions and attempt for not any good reasons. On connection failure, client lib always logs the exception and error  so, there is always a way to find out any previous exception for the connection issues. and there is no any good reason or standard to keep large number of  exception-list in memory and additionally generate huge huge logs which creates unnecessary string objects and consumes high CPU that eventually makes broker unavailable due to high CPU and high gc pauses.

### Modifications

Client lib keeps the previous exception list during connection failure which is unnecessary because of two reasons: 
1) in almost all cases, all previous Exceptions are with same type and same error message as intermittent connection failure and retry happens due to same the issue. 
2) Client lib always logs the exception so, no need to keep list of previous exception list in memory and user can always track all previous exceptions from the log. 
however, client lib can still keep the counter for previous-attempt in the exception context to see number of failures/retry and how far user should check in the logs. So, Client lib should remove the in-memory unnecessary previous exception objects to avoid huge logging data-print but still can keep the exception-counter to help users to do debugging in logs.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
@@ -19,9 +19,9 @@
 package org.apache.pulsar.client.api;
 
 import java.io.IOException;
-import java.util.Collection;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
 import lombok.Getter;
 import org.apache.pulsar.common.classification.InterfaceAudience;
 import org.apache.pulsar.common.classification.InterfaceStability;
@@ -34,7 +34,7 @@ import org.apache.pulsar.common.classification.InterfaceStability;
 @SuppressWarnings("serial")
 public class PulsarClientException extends IOException {
     private long sequenceId = -1;
-    private Collection<Throwable> previous;
+    private AtomicInteger previousExceptionAttempt;
 
     /**
      * Constructs an {@code PulsarClientException} with the specified detail message.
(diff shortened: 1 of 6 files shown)


nodece (approved):

lhotari:
Thank you @rdhabalia. Good work!
One of the remaining challenges in addressing the problem is about retaining binary compatibility with previous versions of Pulsar clients. @rdhabalia do you think that there would be a way to make a change where this is achieved?

rdhabalia:
@lhotari 
`PulsarClientException::toString()` returns a string message; ideally, there is no such contract to maintain the format. in fact, none of the exception message string should have that contract.
another solution is to add client-configuration to restrict the stacktrace by ignoring storing and printing any previous exception but then it will be really bad design to add a client-configuration which will be used to control Exception toString message format.
I guess, motivation of introducing previous-Exception to find out number of attempts happened with such exception, and this PR still keeps attempts-value and prints into the toString() method.
Also, this change should not break any application logic as it prints less information, and any application must handle such information.
So, I guess this change should not break backward compatibility.

rdhabalia:
@lhotari please let me know if we have any additional concerns, or else we can merge this PR.

lhotari:
> @lhotari please let me know if we have any additional concerns, or else we can merge this PR.

No other concerns that notifying the dev mailing about changing the public methods of a published class and removing a PIP-91 feature. You could make it a lazy consensus decision from my point and merge it in 24hrs unless someone stops you and has a good reason to block it. Makes sense?

rdhabalia:
done. I have sent email and will wait for 1-2 days before merging this PR.

PR #23258, opened 2024-09-05
[fix][broker] fix pulsar-admin topics stats-internal caused a BK client thread a deadlock

poorbarcode:
### Motivation

`pulsar-admin topics stats-internal` caused a BK client thread deadlock, leading to other ledgers not being opened successfully, and topics not being loaded up.

A race condition that will trigger the issue
- Calling `topics stats-internal`
- Compaction is running at the last step(opening a new ledger to write)

Screenshot 2024-09-05 at 12 13 55

### Modifications
- Remove sync call

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
@@ -32,6 +32,8 @@ import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import org.apache.bookkeeper.client.BKException;
@@ -304,8 +306,10 @@ public class CompactedTopicImpl implements CompactedTopic {
      * Getter for CompactedTopicContext.
      * @return CompactedTopicContext
      */
-    public Optional<CompactedTopicContext> getCompactedTopicContext() throws ExecutionException, InterruptedException {
-        return compactedTopicContext == null ? Optional.empty() : Optional.of(compactedTopicContext.get());
+    public Optional<CompactedTopicContext> getCompactedTopicContext() throws ExecutionException, InterruptedException,
+            TimeoutException {
+        return compactedTopicContext == null ? Optional.empty() :
+                Optional.of(compactedTopicContext.get(30, TimeUnit.SECONDS));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
```diff
@@ -304,8 +306,10 @@ static CompletableFuture<List<Entry>> readEntries(LedgerHandle lh, long from, lo
      * Getter for CompactedTopicContext.
      * @return CompactedTopicContext
      */
-    public Optional<CompactedTopicContext> getCompactedTopicContext() throws ExecutionException, InterruptedException {
-        return compactedTopicContext == null ? Optional.empty() : Optional.of(compactedTopicContext.get());
+    public Optional<CompactedTopicContext> getCompactedTopicContext() throws ExecutionException, InterruptedException,
```
nodece, line 309: Please return a `CompletableFuture`.

DO NOT use `CompletableFuture#get`, which will block the thread.

nodece (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
Technoboy-, line 309: Why do we fix here?

poorbarcode: I added a new async method `getCompactedTopicContextAsync`, so there are two methods now
- `getCompactedTopicContext`
- `getCompactedTopicContextAsync`

There are `1` places that are calling the old method `getCompactedTopicContext`
- `topic internal stats`
- `HTTP requests: /metrics/`

I have changed the first use-case to `getCompactedTopicContextAsync` in the current PR, and the second one is hard to modify, we can separate a PR to improve the second case. And I added a timeout `30s` for the second case, which is better than calling it without a timeout.

poorbarcode: See the answer here https://github.com/apache/pulsar/pull/23258/files#r1749751058

nodece: You can remove `getCompactedTopicContext` in the PR.

I checked the usages of `getCompactedTopicContext`, one is here, and the remaining calls are in the test package.

Technoboy- (approved):

coderzc (approved):

pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
poorbarcode: see follows:
- https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/OpenTelemetryTopicStats.java#L472
- https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/NamespaceStatsAggregator.java#L328

nodece (approved):

PR #23266, opened 2024-09-06
[fix][broker] Execute the pending callbacks in order before ready for incoming requests

BewareMyPower:
### Background Knowledge

CompletableFuture has an intuitive behavior

```java
        final var future = new CompletableFuture<Integer>();
        future.thenRun(() -> System.out.println("A"));
        future.thenRun(() -> System.out.println("B"));
        future.thenRun(() -> System.out.println("C"));
        future.complete(0);
```

The outputs of the code above are:

```
C
B
A
```

That's because it maintains callbacks in the LIFO stack, not FIFO queue.

### Motivation

https://github.com/apache/pulsar/pull/22977 breaks the order of some events during extensible load manager's start by adding the runnable objects via `thenRun()`. The previous events order of `ExtensibleLoadManagerImpl`:
1. `playLeader()` or `playFollower()`
2. `serviceUnitStateChannel.start()`
3. Schedule some tasks (`brokerLoadDataReportTask`, `topBundlesLoadDataReportTask`, etc.) and set `started` with true.

Now, since they will be executed in reverse order, `started` will first be set true and then event 1 and 2 happened. It might cause unexpected issues 

### Modifications

Add a synchronized list to queue the pending tasks and execute them in order before the future is complete.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
@@ -298,6 +298,7 @@ public class PulsarService implements AutoCloseable, ShutdownService {
     private final DefaultMonotonicSnapshotClock monotonicSnapshotClock;
     private String brokerId;
     private final CompletableFuture<Void> readyForIncomingRequestsFuture = new CompletableFuture<>();
+    private final List<Runnable> pendingTasksBeforeReadyForIncomingRequests = new ArrayList<>();
 
     public enum State {
         Init, Started, Closing, Closed
@@ -1023,7 +1024,11 @@ public class PulsarService implements AutoCloseable, ShutdownService {
             this.metricsGenerator = new MetricsGenerator(this);
 
             // the broker is ready to accept incoming requests by Pulsar binary protocol and http/https
-            readyForIncomingRequestsFuture.complete(null);
+            synchronized (pendingTasksBeforeReadyForIncomingRequests) {
+                pendingTasksBeforeReadyForIncomingRequests.forEach(Runnable::run);
+                pendingTasksBeforeReadyForIncomingRequests.clear();
+                readyForIncomingRequestsFuture.complete(null);
+            }
 
             // Initialize the message protocol handlers.
             // start the protocol handlers only after the broker is ready,
@@ -1082,7 +1087,16 @@ public class PulsarService implements AutoCloseable, ShutdownService {
     }
 
     public void runWhenReadyForIncomingRequests(Runnable runnable) {
-        readyForIncomingRequestsFuture.thenRun(runnable);
+        // Here we don't call the thenRun() methods because CompletableFuture maintains a stack for pending callbacks,
+        // not a queue. Once the future is complete, the pending callbacks will be executed in reverse order of
+        // when they were added.
+        synchronized (pendingTasksBeforeReadyForIncomingRequests) {
+            if (readyForIncomingRequestsFuture.isDone()) {
+                runnable.run();
+            } else {
+                pendingTasksBeforeReadyForIncomingRequests.add(runnable);
+            }
+        }
     }
 
     public void waitUntilReadyForIncomingRequests() throws ExecutionException, InterruptedException {
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
@@ -1090,12 +1090,17 @@ public class PulsarService implements AutoCloseable, ShutdownService {
         // Here we don't call the thenRun() methods because CompletableFuture maintains a stack for pending callbacks,
         // not a queue. Once the future is complete, the pending callbacks will be executed in reverse order of
         // when they were added.
+        final boolean addedToPendingTasks;
         synchronized (pendingTasksBeforeReadyForIncomingRequests) {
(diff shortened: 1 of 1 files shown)


heesung-sn (approved):
Nice catch. ~~I think we also have other pulsar logics, e.g. FutureUtil.sequencer that uses future chain for ordering. I wonder if we need to revisit that too.~~

Nvm, I see that the sequencer uses thenCompose.

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
```diff
@@ -1082,7 +1087,16 @@ public void start() throws PulsarServerException {
...
 
     public void runWhenReadyForIncomingRequests(Runnable runnable) {
-        readyForIncomingRequestsFuture.thenRun(runnable);
+        // Here we don't call the thenRun() methods because CompletableFuture maintains a stack for pending callbacks,
+        // not a queue. Once the future is complete, the pending callbacks will be executed in reverse order of
+        // when they were added.
+        synchronized (pendingTasksBeforeReadyForIncomingRequests) {
+            if (readyForIncomingRequestsFuture.isDone()) {
+                runnable.run();
+            } else {
```
lhotari, line 1096: This requires a slight change so that it doesn't synchronize the running of the runnable needlessly after the broker is initialized. One possible solution is to set a flag and if the flag is set, to run the runnable outside of the synchronized block.

lhotari (changes requested):

lhotari:
> Nice catch. I think we also have other pulsar logics, e.g. FutureUtil.sequencer that uses future chain for ordering. I wonder if we need to revisit that too.

Yes, nice catch @BewareMyPower. 

It's useful to consider replacing FutureUtil.sequencer usage with `com.spotify.futures.ConcurrencyReducer` with the concurrency of 1. [ConcurrencyReducer](https://spotify.github.io/completable-futures/apidocs/com/spotify/futures/ConcurrencyReducer.html) is already included as a dependency in Pulsar as part of com.spotify:completable-futures.

heesung-sn:
Actually, I see that the sequencer uses thenCompose, which I think the order should be fifo.

lhotari:
> Actually, I see that the sequencer uses thenCompose, which I think the order should be fifo.

I think it's different. 
https://github.com/apache/pulsar/blob/9626e7e090e9481e12441a47cf7e89f209aadd03/pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java#L234-L242
the instances are chained and `.thenCompose` isn't called multiple times for the same instance.

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
BewareMyPower: Good suggestion. I will add a flag

BewareMyPower:
The sequencer is different because it's a chained futures (like linked list) that each future has only 1 callback in its internal stack while the existing design is a single future with N callbacks. However, the sequencer might be unnecessarily complicated for this simple case so I won't adopt it.

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
BewareMyPower: Addressed, PTAL again.

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
```diff
@@ -1023,7 +1024,11 @@ public void start() throws PulsarServerException {
             this.metricsGenerator = new MetricsGenerator(this);
 
             // the broker is ready to accept incoming requests by Pulsar binary protocol and http/https
-            readyForIncomingRequestsFuture.complete(null);
+            synchronized (pendingTasksBeforeReadyForIncomingRequests) {
+                pendingTasksBeforeReadyForIncomingRequests.forEach(Runnable::run);
+                pendingTasksBeforeReadyForIncomingRequests.clear();
+                readyForIncomingRequestsFuture.complete(null);
```
lhotari, line 1030: Since the CompletableFuture features aren't needed anymore, this could be simply a boolean field.

lhotari (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
lhotari: Looks good

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
BewareMyPower: No, we need to wait the future being completed in `waitUntilReadyForIncomingRequests`.

```java
            public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
                    throws IOException, ServletException {
                try {
                    // Wait until the PulsarService is ready to serve incoming requests
                    pulsarService.waitUntilReadyForIncomingRequests();
```

BewareMyPower:
I found a deadlock in `ClusterMigrationTest`. Mark this PR as drafted.

lhotari (approved):

codelipenghui (approved):

Issue #21958, opened 2024-01-23
[Bug] Msg backlog & unack msg remains when using acknowledgeAsync


pqab:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Version

3.0.0

### Minimal reproduce step

1. Publish 600k messages

2. Start 2 consumers with different subscription name and subscribe from Earliest

one with async ack
```
Mono.fromCompletionStage(() -> consumer.acknowledgeAsync(message))
```

another one with sync ack
```
Mono.fromRunnable(() -> this.acknowledge(message))
```

3. Wait until it finished### 

### What did you expect to see?

msg backlog & unack message should be 0 for both `acknowledge` & `acknowledgeAsync` subscription

### What did you see instead?

There are few messages in the backlog & unack message left even we received the ack callback when using `acknowledgeAsync`, `acknowledge` is working fine

Topic stats for the `acknowledgeAsync` subscription for reference
```
{
    "msgRateOut" : 382.4166635563445,
    "msgThroughputOut" : 1143831.4573635042,
    "bytesOutCounter" : 1590183426,
    "msgOutCounter" : 531437,
    "msgRateRedeliver" : 0.0,
    "messageAckRate" : 397.2166640450367,
    "chunkedMessageRate" : 0,
    "msgBacklog" : 4,
    "backlogSize" : 572603343,
    "earliestMsgPublishTimeInBacklog" : 0,
    "msgBacklogNoDelayed" : 4,
    "blockedSubscriptionOnUnackedMsgs" : false,
    "msgDelayed" : 0,
    "unackedMessages" : 7,
    "type" : "Key_Shared",
    "msgRateExpired" : 0.0,
    "totalMsgExpired" : 0,
    "lastExpireTimestamp" : 1706002395690,
    "lastConsumedFlowTimestamp" : 1706002431797,
    "lastConsumedTimestamp" : 1706002428241,
    "lastAckedTimestamp" : 1706002436543,
    "lastMarkDeleteAdvancedTimestamp" : 1706002000095,
    "consumers" : [ {
        "msgRateOut" : 382.4166635563445,
        "msgThroughputOut" : 1143831.4573635042,
        "bytesOutCounter" : 598198012,
        "msgOutCounter" : 200000,
        "msgRateRedeliver" : 0.0,
        "messageAckRate" : 397.2166640450367,
        "chunkedMessageRate" : 0.0,
        "consumerName" : "consumer",
[...]

PR #23267, opened 2024-09-07
[improve][pip] PIP-377: Automatic retry for failed acknowledgements

lhotari:
### Motivation

In the pip-377.md file

### Modifications

In the pip-377.md file

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-377.md
+++ b/pip/pip-377.md
@@ -79,5 +79,5 @@ Comprehensive testing will include:
 
 ## Links
 
-* Mailing List discussion thread:
+* Mailing List discussion thread: https://lists.apache.org/thread/7sg7hfv9dyxto36dr8kotghtksy1j0kr
 * Mailing List voting thread:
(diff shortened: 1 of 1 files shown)


aloyszhang:
Is there a precise step to reproduce the acknowledgment lost?

lhotari:
> Is there a precise step to reproduce the acknowledgment lost?

@aloyszhang not yet, there's #22709 and #21958 where it seems that acks get lost. 

One possible way that it could happen is due to #23200 where the number of ack holes exceeds the default limit of 10000 and the consumer gets reconnected at this time (client connection breaks or due to load balancing event). 
There was a previous attempt to limit the number of ack holes with PIP-299. The problem is that it won't prevent exceeding the limit. The problem described in #23200 is addressed with #23231 (which I'm currently working on).

lhotari:
#25525 makes this obsolete and addresses the actual issue

PR #23269, opened 2024-09-09
[improve][broker] Make cluster metadata init command support metadata config path

Demogorgon314:
### Motivation

The pulsar configuration supported `metadataStoreConfigPath` and `configurationStoreConfigPath`, but the cluster metadata init command does not have a place to set the config path.

### Modifications

Make cluster metadata init command support metadata config path

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
@@ -35,6 +35,7 @@ import org.apache.pulsar.broker.resources.PulsarResources;
 import org.apache.pulsar.broker.resources.TenantResources;
 import org.apache.pulsar.client.api.ProxyProtocol;
 import org.apache.pulsar.common.conf.InternalConfigurationData;
+import org.apache.pulsar.common.configuration.FieldContext;
 import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.SystemTopicNames;
 import org.apache.pulsar.common.naming.TopicName;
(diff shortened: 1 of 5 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
```diff
@@ -464,9 +476,12 @@ static void createPartitionedTopic(MetadataStore configStore, TopicName topicNam
         }
     }
 
-    public static MetadataStoreExtended initLocalMetadataStore(String connection, int sessionTimeout) throws Exception {
```
lhotari, line 467: Is it useful to keep backwards compatibility by keeping this method signature and delegating to the new method signature with `configPath` set to `null`? Just thinking of possible cases where this class is used by some external code.

pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
```diff
@@ -475,10 +490,13 @@ public static MetadataStoreExtended initLocalMetadataStore(String connection, in
         return store;
     }
 
-    public static MetadataStoreExtended initConfigMetadataStore(String connection, int sessionTimeout)
```
lhotari, line 478: Is it useful to keep backwards compatibility by keeping this method signature and delegating to the new method signature with `configPath` set to `null`? Just thinking of possible cases where this class is used by some external code.

lhotari (approved):
LGTM, left comments about keeping binary compatibility

pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
Demogorgon314: Good point, updated.

lhotari (approved):
LGTM

Coselding (approved):

Issue #23273, opened 2024-09-09
[Bug] Broker Fails to Restart Due to Incomplete NAR File Extraction in `/tmp` Directory


nikhilerigila09:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

Master branch

### Minimal reproduce step

1. Configure the Pulsar broker with a filter NAR file, using the following parameters:
    * `entryFilterNames=<your-filter-name>`
    * `entryFiltersDirectory=<path-to-filter-directory>`
2. Attempt to start the broker. Ensure the NAR file is available in the filters directory.
3. While the broker is attempting to unpack the NAR file, stop the broker mid-process.
4. Restart the broker.
5. Observe that the broker fails to restart due to a `NoSuchFileException` related to the `/tmp` directory.

### What did you expect to see?

The Pulsar broker should restart normally without any manual intervention or deletion of files in the `/tmp` directory.

### What did you see instead?

The broker fails to start, throwing a `NoSuchFileException` for missing files in the `/tmp` directory, specifically related to the filter NAR file. Manually deleting the `/tmp` directory allows the broker to start normally, but the issue reappears on subsequent restarts.

```
java.nio.file.NoSuchFileException: /tmp/pulsar-nar/pulsar-jms-5.0.4-nar.nar-unpacked/_Cvs_KLip3kCfKeErKASDO/META-INF/services/entry_filter.yml
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:92) ~[?:?]
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106) ~[?:?]
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111) ~[?:?]
	at sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:218) ~[?:?]
	at java.nio.file.Files.newByteChannel(Files.java:380) ~[?:?]
	at java.nio.file.Files.newByteChannel(Files.java:432) ~[?:?]
	at java.nio.file.Files.readAllBytes(Files.java:3288) ~[?:?]
[...]

PR #23274, opened 2024-09-09
[fix][broker] Fix incomplete NAR file extraction which prevents broker from starting

nikhilerigila09:
Fixes [23273](https://github.com/apache/pulsar/issues/23273)

### Motivation
This PR addresses an issue where the Pulsar broker fails to restart due to incomplete NAR file extraction. The broker attempts to reuse an incomplete directory in the `/tmp` folder, leading to a `NoSuchFileException`. This occurs when the broker is stopped during the extraction process, causing the NAR files to remain in an inconsistent state, which affects future restarts.

### Modifications

* extract to a temporary directory which is renamed to the target directory in the last step. This ensures the consistency of the extracted files.

### Verifying this change

* Adding new tests that simulate an incomplete NAR extraction, followed by a broker restart to ensure it handles incomplete extractions.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/cognitree/pulsar/pull/15

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/nar/NarUnpacker.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/nar/NarUnpacker.java
@@ -32,7 +32,9 @@ import java.io.InputStream;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
@@ -86,19 +88,32 @@ public class NarUnpacker {
             try (FileChannel channel = new RandomAccessFile(lockFile, "rw").getChannel();
                  FileLock lock = channel.lock()) {
                 File narWorkingDirectory = new File(parentDirectory, md5Sum);
-                if (narWorkingDirectory.mkdir()) {
+                if (!narWorkingDirectory.exists()) {
+                    File narExtractionTempDirectory = new File(parentDirectory, md5Sum + ".tmp");
+                    if (narExtractionTempDirectory.exists()) {
+                        FileUtils.deleteFile(narExtractionTempDirectory, true);
+                    }
+                    if (!narExtractionTempDirectory.mkdir()) {
+                        throw new IOException("Cannot create " + narExtractionTempDirectory);
+                    }
                     try {
-                        log.info("Extracting {} to {}", nar, narWorkingDirectory);
+                        log.info("Extracting {} to {}", nar, narExtractionTempDirectory);
                         if (extractCallback != null) {
                             extractCallback.run();
                         }
-                        unpack(nar, narWorkingDirectory);
+                        unpack(nar, narExtractionTempDirectory);
                     } catch (IOException e) {
                         log.error("There was a problem extracting the nar file. Deleting {} to clean up state.",
-                                narWorkingDirectory, e);
-                        FileUtils.deleteFile(narWorkingDirectory, true);
+                                narExtractionTempDirectory, e);
+                        try {
+                            FileUtils.deleteFile(narExtractionTempDirectory, true);
+                        } catch (IOException e2) {
+                            log.error("Failed to delete temporary directory {}", narExtractionTempDirectory, e2);
+                        }
                         throw e;
                     }
+                    Files.move(narExtractionTempDirectory.toPath(), narWorkingDirectory.toPath(),
+                            StandardCopyOption.ATOMIC_MOVE);
                 }
                 return narWorkingDirectory;
             }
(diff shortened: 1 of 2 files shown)


lhotari (changes requested):
Please follow a different approach where a temp directory is used which is then renamed instead of using a "success file". Example: https://github.com/lhotari/pulsar/commit/07b2151fd34b8d8d490e38020c0363e3c0526231

pulsar-common/src/main/java/org/apache/pulsar/common/nar/NarUnpacker.java
```diff
@@ -83,22 +83,31 @@ static File doUnpackNar(final File nar, final File baseWorkingDirectory, Runnabl
...
+                if (narWorkingDirectory.mkdir() || !successMarkerFile.exists()) {
                     try {
                         log.info("Extracting {} to {}", nar, narWorkingDirectory);
                         if (extractCallback != null) {
                             extractCallback.run();
                         }
                         unpack(nar, narWorkingDirectory);
+                        boolean successMarkerFileCreated = successMarkerFile.createNewFile();
+                        log.info("Successfully extracted nar file, status for creating success marker at {} is {}",
+                                successMarkerFile.getAbsolutePath(), successMarkerFileCreated);
```
nikhilerigila09, line 99: if there is a restart after creating the narWorkingDirectory then we would end up in the same situation

lhotari:
@nikhilerigila09 Are there any challenges in following the advice to use file renaming instead of using a "success file"?

pulsar-common/src/test/java/org/apache/pulsar/common/nar/NarUnpackerTest.java
```diff
@@ -118,6 +118,30 @@ public static void main(String[] args) {
...
+        AtomicInteger extractCounter = new AtomicInteger();
+
+        new Thread(() -> {
+            try {
+                File narWorkingDirectory = NarUnpacker.doUnpackNar(sampleZipFile, extractDirectory, extractCounter::incrementAndGet);
+                FileUtils.deleteFile(narWorkingDirectory, true);
+                NarUnpacker.doUnpackNar(sampleZipFile, extractDirectory, extractCounter::incrementAndGet);
+            } catch (Exception e) {
+                log.error("Unpacking failed", e);
+                exceptionCounter.incrementAndGet();
```
lhotari, line 134: This might introduce a flaky test since multiple threads are deleting the directory concurrently. It seems that a concurrency test doesn't make sense for testing that the unpacking happens when the working directory is missing.

lhotari: Running the code in a separate thread isn't necessary.

nikhilerigila09: Removed thread

pulsar-common/src/test/java/org/apache/pulsar/common/nar/NarUnpackerTest.java
```diff
@@ -118,6 +118,28 @@ public static void main(String[] args) {
         }
     }
 
+    @Test
+    void shouldReExtractWhenUnpackedDirectoryIsMissing() throws InterruptedException {
+        CountDownLatch countDownLatch = new CountDownLatch(1);
+        AtomicInteger exceptionCounter = new AtomicInteger();
```
lhotari, line 124: these could be removed since there aren't threads running in the background

pulsar-common/src/test/java/org/apache/pulsar/common/nar/NarUnpackerTest.java
```diff
@@ -118,6 +118,28 @@ public static void main(String[] args) {
...
+        try {
+            File narWorkingDirectory = NarUnpacker.doUnpackNar(sampleZipFile, extractDirectory, extractCounter::incrementAndGet);
+            FileUtils.deleteFile(narWorkingDirectory, true);
+            NarUnpacker.doUnpackNar(sampleZipFile, extractDirectory, extractCounter::incrementAndGet);
+        } catch (Exception e) {
+            log.error("Unpacking failed", e);
+            exceptionCounter.incrementAndGet();
+        } finally {
+            countDownLatch.countDown();
+        }
```
lhotari, line 136: remove this. there's no need to catch the exception since this isn't running in a background thread. if an exception is thrown, the test will fail.

lhotari (changes requested):

pulsar-common/src/test/java/org/apache/pulsar/common/nar/NarUnpackerTest.java
lhotari: Please clean up the remaining parts that were no longer needed. I added specific comments.

lhotari (approved):
LGTM, good work @nikhilerigila09

lhotari:
This could also impact Pulsar Functions or `localrun`. I've seen that once in the past where `localrun` would fail due to a incomplete NAR file extraction and the way to resolve the issue was to manually delete the `/tmp/pulsar-nar` directory.

PR #23285, opened 2024-09-11
[improve][broker] Reducing the parse of MessageMetadata in compaction

aloyszhang:
### Motivation

There are many redundant parses of MessageMetadata in topic compaction, including phase one and phase two

### Modifications

reuse the MessageMetadata object  in topic compaction

### Verifying this change

### Documentation

- [x] `doc-not-needed` 

PR in forked repository:  https://github.com/aloyszhang/pulsar/pull/23

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
@@ -59,6 +59,8 @@ public class RawBatchConverter {
         ByteBuf payload = msg.getHeadersAndPayload();
         if (metadata == null) {
             metadata = Commands.parseMessageMetadata(payload);
+        } else {
+            Commands.skipMessageMetadata(payload);
         }
         int batchSize = metadata.getNumMessagesInBatch();
 
@@ -137,6 +139,8 @@ public class RawBatchConverter {
         }
         if (metadata == null) {
             metadata = Commands.parseMessageMetadata(payload);
+        } else {
+            Commands.skipMessageMetadata(payload);
         }
         ByteBuf batchBuffer = PulsarByteBufAllocator.DEFAULT.buffer(payload.capacity());
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
@@ -52,12 +52,14 @@ public class RawBatchConverter {
         return metadata.hasNumMessagesInBatch() && metadata.getEncryptionKeysCount() == 0;
     }
 
-    public static List<MessageCompactionData> extractMessageCompactionData(RawMessage msg)
+    public static List<MessageCompactionData> extractMessageCompactionData(RawMessage msg, MessageMetadata metadata)
         throws IOException {
         checkArgument(msg.getMessageIdData().getBatchIndex() == -1);
 
         ByteBuf payload = msg.getHeadersAndPayload();
-        MessageMetadata metadata = Commands.parseMessageMetadata(payload);
+        if (metadata == null) {
+            metadata = Commands.parseMessageMetadata(payload);
+        }
         int batchSize = metadata.getNumMessagesInBatch();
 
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
```diff
@@ -52,12 +52,14 @@ public static boolean isReadableBatch(MessageMetadata metadata) {
...
 
-    public static List<MessageCompactionData> extractMessageCompactionData(RawMessage msg)
+    public static List<MessageCompactionData> extractMessageCompactionData(RawMessage msg, MessageMetadata metadata)
         throws IOException {
         checkArgument(msg.getMessageIdData().getBatchIndex() == -1);
 
         ByteBuf payload = msg.getHeadersAndPayload();
-        MessageMetadata metadata = Commands.parseMessageMetadata(payload);
+        if (metadata == null) {
+            metadata = Commands.parseMessageMetadata(payload);
```
lhotari, line 61: is there a need to skip the message metadata if it's not parsed?
Doesn't `parseMessageMetadata` move the read position of the ByteBuf forward?

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
```diff
@@ -123,7 +135,9 @@ public static Optional<RawMessage> rebatchMessage(RawMessage msg,
             payload.readerIndex(readerIndex);
             brokerMeta = payload.readSlice(brokerEntryMetadataSize + Short.BYTES + Integer.BYTES);
         }
-        MessageMetadata metadata = Commands.parseMessageMetadata(payload);
+        if (metadata == null) {
+            metadata = Commands.parseMessageMetadata(payload);
```
lhotari, line 139: is there a need to skip the message metadata if it's not parsed?
Doesn't `parseMessageMetadata` move the read position of the ByteBuf forward?

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
aloyszhang: Yes, you're right. I missed it before.

aloyszhang: fixed

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
aloyszhang: fixed

lhotari (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawBatchConverter.java
```diff
@@ -52,12 +52,16 @@ public static boolean isReadableBatch(MessageMetadata metadata) {
...
 
-    public static List<MessageCompactionData> extractMessageCompactionData(RawMessage msg)
+    public static List<MessageCompactionData> extractMessageCompactionData(RawMessage msg, MessageMetadata metadata)
         throws IOException {
         checkArgument(msg.getMessageIdData().getBatchIndex() == -1);
 
         ByteBuf payload = msg.getHeadersAndPayload();
-        MessageMetadata metadata = Commands.parseMessageMetadata(payload);
+        if (metadata == null) {
+            metadata = Commands.parseMessageMetadata(payload);
```
lhotari, line 61: One more concern. `Commands.parseMessageMetadata(payload)` will use a shared thread local instance. Is that a problem here?
`Commands.parseMessageMetadata(payload, new MessageMetadata())` would be a way to resolve it if there's a need to have multiple instances in flight.

aloyszhang: I think it's not a problem.
Metadata is parsed previously when doing compaction in both phase one and phase two.
The null value for metadata only happens in unit tests.

lordcheng10:
LGTM

PR #22055, opened 2024-02-15
[fix] Fixed implicit conversions from long -> int

merlimat:
### Motivation

Fix `Implicit narrowing conversion in compound assignment` issues reported by CodeQL scanner

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23286, opened 2024-09-11
[improve][broker] Fix non-persistent system topic schema compatibility

Demogorgon314:
### Motivation
When upgrading broker version from `3.0.x` to `3.3.x` with `ExtensibleLoadManagerImpl` enabled, it will have an `Unable to read schema` exception. And the broker will fail to start. This issue is caused by https://github.com/apache/pulsar/pull/22055 .

```
***************
*** 23,29 ****
      "default" : null
    }, {
      "name" : "bundleCount",
!     "type" : "int"
    }, {
      "name" : "cpu",
      "type" : [ "null", "org.apache.pulsar.policies.data.loadbalancer.ResourceUsage" ],
--- 23,29 ----
      "default" : null
    }, {
      "name" : "bundleCount",
!     "type" : "long"
    }, {
      "name" : "cpu",
      "type" : [ "null", "org.apache.pulsar.policies.data.loadbalancer.ResourceUsage" ],
***************
*** 59,65 ****
      "type" : "long"
    }, {
      "name" : "topics",
!     "type" : "int"
    }, {
      "name" : "updatedAt",
      "type" : "long"
--- 59,65 ----
      "type" : "long"
    }, {
      "name" : "topics",
!     "type" : "long"
    }, {
      "name" : "updatedAt",
      "type" : "long"
```

### Modifications

Add a new class `NonPersistentSystemTopic`, and it will use for system non-persistent topic.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


heesung-sn (approved):

codelipenghui (commented):
The fix looks good. But I still have concern about the compatibility.

- We have added `systemTopicSchemaCompatibilityStrategy=ALWAYS_COMPATIBLE` which is not by original design, because we don't have another way to resolve the compatibility issue.
- Even if we use `ALWAYS_COMPATIBLE`, that doesn't means we will ok for any data format change

So, I would like to suggest to add a real compatibility test for different version of data. For the code part, we should not remove the old format data struct directly. Instead, we can create v1, v2 ..., then we can test that we can publish data from v1, v2 and consume by v(latest), and vice versa.

Demogorgon314:
@codelipenghui Did you mean to set the `SchemaCompatibilityStrategy` to `BACKWARD`?

pulsar-broker/src/test/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImplTest.java
```diff
@@ -1897,6 +1902,43 @@ public void unloadTimeoutCheckTest()
                         TimeUnit.SECONDS).get(2, TimeUnit.SECONDS);
     }
 
+    @Test
+    public void testSystemNonPersistentTopicSchemaCompatibility() throws Exception {
+        String topicName = ExtensibleLoadManagerImpl.BROKER_LOAD_DATA_STORE_TOPIC;
+        NonPersistentSystemTopic topic = new NonPersistentSystemTopic(topicName, pulsar.getBrokerService());
+        Assert.assertEquals(SchemaCompatibilityStrategy.ALWAYS_COMPATIBLE, topic.getSchemaCompatibilityStrategy());
+
+        var brokerLoadDataStore = LoadDataStoreFactory.create(pulsar, topicName, BrokerLoadDataV1.class);
```
Demogorgon314, line 1911: @codelipenghui I have added a test to cover the `BrokerLoadData` format change.

shibd (approved):

PR #23297, opened 2024-09-12
[fix][broker] Fail fast if the extensible load manager failed to start

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/23230 tries to fix the issue that "the pulsar broker cannot catch the exception" caused by https://github.com/apache/pulsar/pull/22977. However, the fix is incorrect because even before https://github.com/apache/pulsar/pull/22977, the broker also wouldn't fail if the extensible load manager failed to start.

It's destructive to have broker running with a failed load manager. Since `BrokerRegistry#unregister` will be called for failures, the issue broker will unregister itself from the metadata store and it could not be selected as the owner broker. Besides, all lookup requests sent to this broker will fail because the load manager is not started.

### Modifications

- Revert https://github.com/apache/pulsar/pull/23230
- Implement `failStarting` correctly:
  - If it has registered, just unregister the broker from the metadata store via `BrokerRegistry#close`  and swallow the exception.
  - Complete `initWaiter` with false to tell background threads to exit directly. Otherwise, methods like `playFollower` will swallow the exception from `initWaiter` and continue the loop.
  - Propagate the exception by wrapping checked exception `PulsarServerException` into a unchecked exception `CompletionException` and unwrap it in `PulsarService#start`'s catch block.
- Add `LoadManagerFailFastTest` to verify the fail-fast behaviors and the broker will unregister itself from the metadata store.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -908,8 +908,6 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                 topBundlesLoadDataStore.close();
                 topBundlesLoadDataStore.startProducer();
                 break;
-            } catch (InterruptedException ignored) {
-                return;
             } catch (Throwable e) {
                 log.warn("The broker:{} failed to set the role. Retrying {} th ...",
                         pulsar.getBrokerId(), ++retry, e);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -36,6 +36,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -521,38 +471,33 @@ public void start() throws PulsarServerException {
...
-        if (this.serviceUnitStateChannel != null) {
-            try {
-                serviceUnitStateChannel.close();
-            } catch (IOException e) {
-                // ignore
+                brokerRegistry.close();
+            } catch (PulsarServerException e) {
+                // If close failed, this broker might still exist in the metadata store. Then it could be found by other
+                // brokers as an available broker. Hence, print a warning log for it.
+                log.warn("Failed to close the broker registry: {}", e.getMessage());
```
codelipenghui, line 493: If the broker can't unregister itself from failure, the lock should be released automatically after the connection get closed from broker to metadata store, right? So that the broker can register itself to the metadata store.

codelipenghui (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
heesung-sn: Yes, that should be the logic here.

heesung-sn (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
BewareMyPower: > the lock should be released automatically after the connection get closed 

Unfortunately not. The lock is released by deleting the metadata store node with the version stored internally. Closing the connection won't release the lock.

heesung-sn: I thought we are using ephemeral lock for this "ephemeral" ownership. No?

RobertIndie (approved):

lhotari:
There are issues in backporting to branch-3.0. I created a separate PR #23297

PR #23298, opened 2024-09-12
[improve][broker] Register the broker to metadata store without version id compare

BewareMyPower:
### Motivation

It's observed in our test environment many times after restarting the only broker with the extensible load manager's broker registry service.

```
2024-09-12T14:00:20,557+0000 [main] ERROR org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImpl - Failed to start the extensible load balance and close broker registry org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryImpl@6d6c4775.
org.apache.pulsar.broker.PulsarServerException: org.apache.pulsar.metadata.api.MetadataStoreException$LockBusyException: Resource at /loadbalance/brokers/<broker-id>:8080 is already locked
        at org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryImpl.start(BrokerRegistryImpl.java:112) ~
        at org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImpl.start(ExtensibleLoadManagerImpl.java:370) 
        at org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerWrapper.start(ExtensibleLoadManagerWrapper.java:50) 
        at org.apache.pulsar.broker.PulsarService.startLoadManagementService(PulsarService.java:1426) 
        at org.apache.pulsar.broker.PulsarService.start(PulsarService.java:986) 
        ...
Caused by: org.apache.pulsar.metadata.api.MetadataStoreException$LockBusyException: Resource at /loadbalance/brokers/<broker-id>:8080 is already locked
        at org.apache.pulsar.metadata.api.MetadataStoreException.unwrap(MetadataStoreException.java:186) ~
        at org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryImpl.register(BrokerRegistryImpl.java:131) ~
        at org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryImpl.start(BrokerRegistryImpl.java:110) ~
        ... 16 more
```

The "Resource xxx is already locked" is usually caused by a `BadVersionException`. This issue can happen when:
1. Broker somehow crashed (e.g. killed by `kill -9` or the container crashed) so it did not delete the node from the metadata store.
2. Broker restarted, but the node still existed so it failed to acquire the lock.
3. **This error can only be recovered by restarting the broker after the metadata store deleted that node (manually by operators or automatically by session timeout detected)**

Actually, it's not a good use case of  `LockManager` here. The broker registry only:
1. Put it's own broker's lookup data under the of `/loadbalance/brokers/<broker-id>` when starting (in the main thread)
2. Delete its own broker id from when closing.
3. Query any broker's lookup data from the `/loadbalance/brokers/<broker-id>`
4. List all children nodes of `/loadbalance/brokers`.

Currently, step 1 and 2 are performed by the distribution lock implemented by `LockManager`. However, there should be no race condition because the broker id is the unique identifier of a broker. There is no case that a broker starts or closes concurrently. In addition, it's meaningless to pass a version explicitly to the metadata store.

Regarding step 3 and 4, they are simply wrapping the `MetadataCache`'s methods. They are actually never related to something like "lock" though they're named `listLocks` and `readLock`.

### Modifications

- Add a `put` method to `MetadataCache`, which just calls `MetadataStore#put` with `Optional.empty()` underlying.
- Replace `LockManager` with `MetadataCache` in `BrokerRegistryImpl`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.RejectedExecutionException;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
@@ -21,6 +21,7 @@ package org.apache.pulsar.broker.loadbalance.extensions;
 import static org.apache.pulsar.broker.loadbalance.LoadManager.LOADBALANCE_BROKERS_ROOT;
 import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -43,6 +44,7 @@ import org.apache.pulsar.metadata.api.MetadataCache;
 import org.apache.pulsar.metadata.api.MetadataStoreException;
 import org.apache.pulsar.metadata.api.Notification;
 import org.apache.pulsar.metadata.api.NotificationType;
+import org.apache.pulsar.metadata.api.extended.CreateOption;
 
 /**
  * The broker registry impl, base on the LockManager.
@@ -119,7 +121,7 @@ public class BrokerRegistryImpl implements BrokerRegistry {
     public synchronized void register() throws MetadataStoreException {
         if (this.state == State.Started) {
             try {
-                brokerLookupDataMetadataCache.put(brokerIdKeyPath, brokerLookupData)
+                brokerLookupDataMetadataCache.put(brokerIdKeyPath, brokerLookupData, EnumSet.of(CreateOption.Ephemeral))
                         .get(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS);
                 this.state = State.Registered;
             } catch (InterruptedException | ExecutionException | TimeoutException e) {
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
@@ -136,9 +136,16 @@ public class BrokerRegistryImpl implements BrokerRegistry {
             try {
                 brokerLookupDataMetadataCache.delete(brokerIdKeyPath)
                         .get(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS);
-                this.state = State.Started;
-            } catch (CompletionException | InterruptedException | ExecutionException | TimeoutException e) {
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof MetadataStoreException.NotFoundException) {
+                    log.warn("{} has already been unregistered", brokerIdKeyPath);
+                } else {
+                    throw MetadataStoreException.unwrap(e);
+                }
+            } catch (InterruptedException | TimeoutException e) {
                 throw MetadataStoreException.unwrap(e);
+            } finally {
+                this.state = State.Started;
             }
         }
     }
@@ -192,11 +199,7 @@ public class BrokerRegistryImpl implements BrokerRegistry {
             this.listeners.clear();
             this.unregister();
         } catch (Exception ex) {
-            if (ex.getCause() instanceof MetadataStoreException.NotFoundException) {
-                throw new PulsarServerException.NotFoundException(MetadataStoreException.unwrap(ex));
-            } else {
-                throw new PulsarServerException(MetadataStoreException.unwrap(ex));
-            }
+            log.error("Unexpected error when unregistering the broker registry", ex);
         } finally {
             this.state = State.Closed;
         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
@@ -39,11 +39,10 @@ import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.ServiceConfiguration;
 import org.apache.pulsar.broker.loadbalance.extensions.data.BrokerLookupData;
 import org.apache.pulsar.common.util.FutureUtil;
+import org.apache.pulsar.metadata.api.MetadataCache;
 import org.apache.pulsar.metadata.api.MetadataStoreException;
 import org.apache.pulsar.metadata.api.Notification;
 import org.apache.pulsar.metadata.api.NotificationType;
-import org.apache.pulsar.metadata.api.coordination.LockManager;
-import org.apache.pulsar.metadata.api.coordination.ResourceLock;
 
 /**
  * The broker registry impl, base on the LockManager.
@@ -57,15 +56,13 @@ public class BrokerRegistryImpl implements BrokerRegistry {
 
     private final BrokerLookupData brokerLookupData;
 
-    private final LockManager<BrokerLookupData> brokerLookupDataLockManager;
+    private final MetadataCache<BrokerLookupData> brokerLookupDataMetadataCache;
 
-    private final String brokerId;
+    private final String brokerIdKeyPath;
 
     private final ScheduledExecutorService scheduler;
 
     private final List<BiConsumer<String, NotificationType>> listeners;
-
-    private volatile ResourceLock<BrokerLookupData> brokerLookupDataLock;
 
     protected enum State {
         Init,
@@ -79,10 +76,10 @@ public class BrokerRegistryImpl implements BrokerRegistry {
     public BrokerRegistryImpl(PulsarService pulsar) {
         this.pulsar = pulsar;
         this.conf = pulsar.getConfiguration();
-        this.brokerLookupDataLockManager = pulsar.getCoordinationService().getLockManager(BrokerLookupData.class);
+        this.brokerLookupDataMetadataCache = pulsar.getLocalMetadataStore().getMetadataCache(BrokerLookupData.class);
         this.scheduler = pulsar.getLoadManagerExecutor();
         this.listeners = new ArrayList<>();
-        this.brokerId = pulsar.getBrokerId();
+        this.brokerIdKeyPath = keyPath(pulsar.getBrokerId());
         this.brokerLookupData = new BrokerLookupData(
                 pulsar.getWebServiceAddress(),
                 pulsar.getWebServiceAddressTls(),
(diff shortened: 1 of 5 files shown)


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
```diff
@@ -243,6 +243,17 @@ public CompletableFuture<Void> create(String path, T value) {
...
 
+    @Override
+    public CompletableFuture<Void> put(String path, T value) {
+        final byte[] bytes;
+        try {
+            bytes = serde.serialize(path, value);
+        } catch (IOException e) {
+            return CompletableFuture.failedFuture(e);
+        }
+        return store.put(path, bytes, Optional.empty()).thenAccept(__ -> {});
```
lhotari, line 254: It seems that the possible already cached value should be expired or updated too?

lhotari: `.thenAccept(__ -> refresh(path))` would be aligned how the other operations handle cache invalidation and refreshing.

codelipenghui:
> Currently, step 1 and 2 are performed by the distribution lock implemented by LockManager. However, there should be no race condition because the broker id is the unique identifier of a broker. There is no case that a broker starts or closes concurrently. In addition, it's meaningless to pass a version explicitly to the metadata store.

@BewareMyPower But it could happen from user's incorrect configuration that more than 1 brokers use the same broker id especially for a on-prem environment. And the comment https://github.com/apache/pulsar/pull/23297#discussion_r1757265763 is also the same concern about the lock issue. It looks like something cause the lock didn't get released even if the broker unregister itself from the metadata server.

lhotari:
> @BewareMyPower But it could happen from user's incorrect configuration that more than 1 brokers use the same broker id especially for a on-prem environment. 

If multiple brokers have the same broker id, the configuration would have multiple other problems. Wouldn't it be a valid requirement that the admin of the Pulsar system ensures that broker ids are unique?

> And the comment [#23297 (comment)](https://github.com/apache/pulsar/pull/23297#discussion_r1757265763) is also the same concern about the lock issue. It looks like something cause the lock didn't get released even if the broker unregister itself from the metadata server.

Yes, that seems to be a real issue that would need to be addressed in any case.

BewareMyPower:
> But it could happen from user's incorrect configuration that more than 1 brokers use the same broker id especially for a on-prem environment. 

I agree with Lari that we should not guarantee a certain behavior with wrong configuration. If you assumed invalid configuration could happen, what about two brokers configured different metadata stores? How can you prevent the broker with wrong metadata URL from starting?

BewareMyPower:
From another perspective, this PR resolves the case the broker exited unexpectedly (e.g. by a crash). Just like the incident we met online:

```
Sep 12, 2024 @ 07:00:10.089	#  SIGSEGV (0xb) at pc=0x00000000000049c6, pid=1, tid=454
```

The JVM process exited by a SIGSEGV signal and no cleanup was done, even the metadata store client was not closed. Then when the broker restarted, it would fail with `LockBusyException`. After some time, the metadata store server detected the client's session timeout and then it deleted the node. Then restarting the broker would succeed.

When programming, we should handle exceptions rather than errors.
- Unexpected crash of some other unexpected logics that skip the unregister are "exceptions": it can be recovered by ignoring the version compare, like what this PR does.
- Misconfiguration at the beginning are "errors": it cannot be recovered until the operators realized this error and updated the correct configuration.

There is another solution that we can add retry logic to `BrokerRegistryImpl#register`, here is a simple example (without using `Backoff`):

```java
            final var deadline = System.currentTimeMillis() + pulsar.getConfig().getMetadataStoreSessionTimeoutMillis();
            Throwable throwable = null;
            while (true) {
                if (System.currentTimeMillis() > deadline) {
                    throw MetadataStoreException.unwrap(throwable == null ? new TimeoutException("Failed to register")
                            : throwable);
                }
                try {
                    this.brokerLookupDataLock = brokerLookupDataLockManager.acquireLock(keyPath(brokerId), brokerLookupData)
                            .get(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS);
                    this.state = State.Registered;
                    break;
                } catch (InterruptedException | ExecutionException | TimeoutException e) {
                    throwable = e;
                    try {
                        Thread.sleep(100);
                    } catch (InterruptedException ex) {
                        throw MetadataStoreException.unwrap(ex);
                    }
                }
            }
```

BewareMyPower:
I just thought it again. Retry is not a good solution. If the broker crashed, the registered data could still exist in the metadata store so that this "dead" broker could be selected as the owner. However, even after it restarted, it could not serve any request until the metadata store session timeout. So many topics could be unavailable for a long time.

Therefore, it does not make sense to keep the current flaky logic just to fail fast for a misconfiguration case. @codelipenghui

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
```diff
@@ -122,7 +119,7 @@ public boolean isStarted() {
     public synchronized void register() throws MetadataStoreException {
         if (this.state == State.Started) {
             try {
-                this.brokerLookupDataLock = brokerLookupDataLockManager.acquireLock(keyPath(brokerId), brokerLookupData)
+                brokerLookupDataMetadataCache.put(brokerIdKeyPath, brokerLookupData)
```
lhotari, line 122: One detail here is that the lock is an ephemeral entry in the metadata store which should get expired when the session dies or disconnects. when switching from acquireLock to put, it becomes a persistent entry in the metadata store. That might be a problem.

BewareMyPower: Good point! I will change the `put` API to accept `CreateOption`

BewareMyPower: It's still confusing to me for why did we introduce `MetadataStoreExtended` and in which case `MetadataStore` is not a `MetadataStoreExtended`. Currently `MetadataCache` just maintains a `MetadataStore` internally. Let me read more details first.

BewareMyPower: I added the `CreateOption` parameter now. PTAL again @lhotari 

BTW, I just read the code about `MetadataStoreExtended`. This API seems not designed well. Currently, built-in implementation all implement this interface. If it does not force the 3rd party library to implement the extra interfaces, it should be documented that where are these methods be used and what will happen if they are not implemented.

codelipenghui:
I'm not going to say we should handle the incorrect configuration from the user side, but what is the expected behavior if user made an incorrect broker ID (same broker ID)? As I understand, only the first broker should register successfully, all the other brokers with the same broker ID should get failed?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataCache.java
```diff
@@ -128,6 +130,24 @@ public interface MetadataCache<T> {
...
+     * already exists, the new value will override the old value.
+     * </p>
+     * @param path the path of the object in the metadata store
+     * @param value the object to put in the metadata store
+     * @param options the create options if the path does not in the metadata store
+     * @return the future that indicates if this operation failed, it could fail with
+     *   {@link java.io.IOException} if the value failed to be serialized
+     *   {@link MetadataStoreException} if the metadata store operation failed
+     */
+    CompletableFuture<Void> put(String path, T value, EnumSet<CreateOption> options);
```
lhotari, line 149:
```suggestion
    CompletableFuture<Void> put(String path, T value, Set<CreateOption> options);
```
It's better to depend on the `Set` interface instead of a particular implementation (`EnumSet`)

lhotari: Oh it looks like `EnumSet` is already in `MetadataStoreExtended`. Not really addressable.

lhotari (approved):
LGTM

lhotari:
> I'm not going to say we should handle the incorrect configuration from the user side, but what is the expected behavior if user made an incorrect broker ID (same broker ID)? As I understand, only the first broker should register successfully, all the other brokers with the same broker ID should get failed?

In this case, one of the benefits of this change is that in the case of broker crash, that the broker can recover immediately. The session of the crashed instance will block starting of a new instance if we don't make this change. I think that it's a different matter to detect if someone has duplicate broker ids in their Pulsar cluster configuration.

Demogorgon314 (approved):
Do we need to apply the same change for `ModularLoadManager`?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataCache.java
BewareMyPower: It makes sense to me since `MetadataCache` is implemented to hide some details to process when using `MetadataStore` directly.

lhotari: > It makes sense to me since `MetadataCache` is implemented to hide some details to process when using `MetadataStore` directly.

My comment was about using `EnumSet` in interface methods. It's not a good practice in interface design to require a specific implementation. Using `EnumSet` is similar to using `ArrayList` in interface methods. 
In this case it's not addressable since the `MetadataStoreExtended` already did that mistake and we don't want to start changing that as part of this PR.

BewareMyPower:
> Perhaps the description could be updated to include the actual reason why we want to do this change,

@lhotari Updated.

>  but what is the expected behavior if user made an incorrect broker ID (same broker ID)? As I understand, only the first broker should register successfully, all the other brokers with the same broker ID should get failed?

@codelipenghui I'd like to say, [undefined behavior](https://en.wikipedia.org/wiki/Undefined_behavior). Kafka's behavior is also that only one broker will succeed if multiple brokers configure the same [broker.id](https://kafka.apache.org/documentation/#brokerconfigs_broker.id). However, this behavior is never documented. It only suggests generating broker id automatically by ZooKeeper:

> To avoid conflicts between ZooKeeper generated broker id's and user configured broker id's, generated broker ids start from reserved.broker.max.id + 1.

Actually, in real world, only if two brokers were configured with the same `advertisedAddress` and `webServicePort` would they have the same broker id. Even if they both don't fail after the change of this PR, users can still easily found the service didn't work because the advertised address is wrong.

It would be good if users can quickly know they've configured the same broker id for different brokers. However, IMO, this issue caused by misconfiguration can hardly exist in practical rather than in theory.

BewareMyPower:
I swallowed the exception in `BrokerRegistryImpl#unregister()`.

BTW, I think all synchronous close operations for components should not throw any exception. Otherwise, the following close operations will be skipped. Just add error logs for unexpected exceptions and warning logs for expected exceptions.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java
```diff
@@ -122,7 +120,7 @@ public boolean isStarted() {
     public synchronized void register() throws MetadataStoreException {
         if (this.state == State.Started) {
             try {
-                this.brokerLookupDataLock = brokerLookupDataLockManager.acquireLock(keyPath(brokerId), brokerLookupData)
+                brokerLookupDataMetadataCache.put(brokerIdKeyPath, brokerLookupData, EnumSet.of(CreateOption.Ephemeral))
```
lhotari, line 123: In https://github.com/apache/pulsar/issues/23889#issuecomment-2614188629 @Joforde shared an observation that the `CreateOption.Ephemeral` will be ignored and a persistent node will get created. @BewareMyPower @heesung-sn Please check this.

lhotari: Will be addressed by PR #23902

PR #23300, opened 2024-09-12
[improve][pip] PIP-378 Add ServiceUnitStateTableView abstraction (ExtensibleLoadMangerImpl only)

heesung-sn:
This is a PIP. Please check the doc.

PR #23301, opened 2024-09-12
[improve][broker] Add ServiceUnitStateTableView (ExtensibleLoadManagerImpl only)

heesung-sn:
PIP: https://github.com/apache/pulsar/pull/23300 

### Motivation

Implements the PIP: https://github.com/apache/pulsar/pull/23300

### Modifications
- Add ServiceUnitStateTableView interface
- Add ServiceUnitStateTableViewImpl implementation that uses Pulsar System topic (compatible with existing behavior)
- Add ServiceUnitStateMetadataStoreTableViewImpl implementation that uses Pulsar Metadata Store (new behavior)
- Refactor ExtensibleLoadMangerImpl and ServiceUnitStateChannelImpl to accept ServiceUnitStateTableView.
- Add time-based producer restart logic in LoadDataTableView and improved its shutdown logic.
- Improve ownership release logic to be more graceful upon broker shutdown
- Refactor related tests code
- ServiceUnitStateTableViewSyncer to sync the two table views during migration.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/heesung-sn/pulsar/pull/81

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -2912,6 +2912,14 @@ public class ServiceConfiguration implements PulsarConfiguration {
     )
     private boolean loadBalancerMultiPhaseBundleUnload = true;
 
+    @FieldContext(
+            dynamic = false,
+            category = CATEGORY_LOAD_BALANCER,
+            doc = "Name of ServiceUnitStateTableView implementation class to use"
+    )
+    private String loadManagerServiceUnitStateTableViewClassName =
+            "org.apache.pulsar.broker.loadbalance.extensions.channel.ServiceUnitStateTableViewImpl";
+
     /**** --- Replication. --- ****/
     @FieldContext(
         category = CATEGORY_REPLICATION,
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -2927,9 +2927,9 @@ public class ServiceConfiguration implements PulsarConfiguration {
                     + "system topic table views during migration from one to the other. One could enable this"
                     + " syncer before migration and disable it after the migration finishes. "
                     + "It accepts `MetadataStoreToSystemTopicSyncer` or `SystemTopicToMetadataStoreSyncer` to "
-                    + "enable it. Null value disables it."
+                    + "enable it. It accepts `None` to disable it."
     )
-    private ServiceUnitTableViewSyncerType loadBalancerServiceUnitTableViewSyncer = null;
+    private ServiceUnitTableViewSyncerType loadBalancerServiceUnitTableViewSyncer = ServiceUnitTableViewSyncerType.None;
 
     /**** --- Replication. --- ****/
     @FieldContext(
@@ -3831,10 +3831,11 @@ public class ServiceConfiguration implements PulsarConfiguration {
     }
 
     public boolean isLoadBalancerServiceUnitTableViewSyncerEnabled() {
-        return loadBalancerServiceUnitTableViewSyncer != null;
+        return loadBalancerServiceUnitTableViewSyncer != ServiceUnitTableViewSyncerType.None;
     }
 
     public enum ServiceUnitTableViewSyncerType {
+        None,
         MetadataStoreToSystemTopicSyncer,
         SystemTopicToMetadataStoreSyncer;
     }
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -2920,6 +2920,15 @@ public class ServiceConfiguration implements PulsarConfiguration {
     private String loadManagerServiceUnitStateTableViewClassName =
             "org.apache.pulsar.broker.loadbalance.extensions.channel.ServiceUnitStateTableViewImpl";
 
+    @FieldContext(
+            dynamic = true,
+            category = CATEGORY_LOAD_BALANCER,
+            doc = "Enable ServiceUnitTableViewSyncer to sync service unit(bundle) states between metadata store and "
+                    + "system topic table views during migration from one to the other. One could enable this"
+                    + " syncer before migration and disable it after the migration finishes."
+    )
+    private boolean loadBalancerServiceUnitTableViewSyncerEnabled = false;
+
     /**** --- Replication. --- ****/
     @FieldContext(
         category = CATEGORY_REPLICATION,
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -2923,11 +2923,13 @@ public class ServiceConfiguration implements PulsarConfiguration {
     @FieldContext(
             dynamic = true,
             category = CATEGORY_LOAD_BALANCER,
-            doc = "Enable ServiceUnitTableViewSyncer to sync service unit(bundle) states between metadata store and "
+            doc = "Specify ServiceUnitTableViewSyncer to sync service unit(bundle) states between metadata store and "
                     + "system topic table views during migration from one to the other. One could enable this"
-                    + " syncer before migration and disable it after the migration finishes."
+                    + " syncer before migration and disable it after the migration finishes. "
+                    + "It accepts `MetadataStoreToSystemTopicSyncer` or `SystemTopicToMetadataStoreSyncer` to "
+                    + "enable it. Null value disables it."
     )
-    private boolean loadBalancerServiceUnitTableViewSyncerEnabled = false;
+    private ServiceUnitTableViewSyncerType loadBalancerServiceUnitTableViewSyncer = null;
 
     /**** --- Replication. --- ****/
     @FieldContext(
@@ -3826,5 +3828,14 @@ public class ServiceConfiguration implements PulsarConfiguration {
             }
         });
         return map;
+    }
+
+    public boolean isLoadBalancerServiceUnitTableViewSyncerEnabled() {
+        return loadBalancerServiceUnitTableViewSyncer != null;
+    }
+
+    public enum ServiceUnitTableViewSyncerType {
+        MetadataStoreToSystemTopicSyncer,
+        SystemTopicToMetadataStoreSyncer;
     }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
@@ -108,7 +108,7 @@ public class ServiceUnitStateMetadataStoreTableViewImpl extends ServiceUnitState
     @Override
     public ServiceUnitStateData get(String key) {
         if (!isValidState()) {
-            new IllegalStateException(INVALID_STATE_ERROR_MSG);
+            throw new IllegalStateException(INVALID_STATE_ERROR_MSG);
         }
         return tableview.get(key);
     }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
@@ -1,0 +1,155 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pulsar.broker.loadbalance.extensions.channel;
+
+import static org.apache.pulsar.broker.loadbalance.extensions.channel.ServiceUnitState.StorageType.MetadataStore;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.broker.PulsarService;
+import org.apache.pulsar.common.util.FutureUtil;
+import org.apache.pulsar.metadata.api.MetadataStoreException;
+import org.apache.pulsar.metadata.api.MetadataStoreTableView;
+import org.apache.pulsar.metadata.tableview.impl.MetadataStoreTableViewImpl;
+
+@Slf4j
+public class ServiceUnitStateMetadataStoreTableViewImpl extends ServiceUnitStateTableViewBase {
+    public static final String PATH_PREFIX = "/service_unit_state";
+    private static final String VALID_PATH_REG_EX = "^\\/service_unit_state\\/.*\\/0x[0-9a-fA-F]{8}_0x[0-9a-fA-F]{8}$";
+    private static final Pattern VALID_PATH_PATTERN;
+
+    static {
+        try {
+            VALID_PATH_PATTERN = Pattern.compile(VALID_PATH_REG_EX);
+        } catch (PatternSyntaxException error) {
+            log.error("Invalid regular expression {}", VALID_PATH_REG_EX, error);
+            throw new IllegalArgumentException(error);
+        }
+    }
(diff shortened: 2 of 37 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.pulsar.common.util.FutureUtil;
+import org.apache.pulsar.common.util.ObjectMapperFactory;
+
+/**
+ * Defines ServiceUnitTableViewSyncer.
+ * It syncs service unit(bundle) states between metadata store and system topic table views.
+ * One could enable this syncer before migration from one to the other and disable it after the migration finishes.
+ */
+@Slf4j
+public class ServiceUnitStateTableViewSyncer implements Cloneable {
```
Demogorgon314, line 43:
```suggestion
public class ServiceUnitStateTableViewSyncer implements Closeable {
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -0,0 +1,184 @@
...
+    private static final int MAX_CONCURRENT_SYNC_COUNT = 100;
+    private volatile ServiceUnitStateTableView systemTopicTableView;
+    private volatile ServiceUnitStateTableView metadataStoreTableView;
+    private volatile boolean isActive = false;
+
+    public void start(PulsarService pulsar)
+            throws IOException, TimeoutException, InterruptedException, ExecutionException {
+        if (!pulsar.getConfiguration().isLoadBalancerServiceUnitTableViewSyncerEnabled()) {
```
Demogorgon314, line 51: Why check it again on this method?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
heesung-sn: Nice catch. Thanks.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
heesung-sn: Just to make sure we only want to start it when the flag is on.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -0,0 +1,186 @@
...
+            log.error("Failed to start ServiceUnitStateTableViewSyncer", e);
+            throw e;
+        }
+    }
+
+    private CompletableFuture<Void> syncToSystemTopic(String key, ServiceUnitStateData data) {
+        return systemTopicTableView.put(key, data);
+    }
+
+    private CompletableFuture<Void>  syncToMetadataStore(String key, ServiceUnitStateData data) {
```
Demogorgon314, line 147:
```suggestion
    private CompletableFuture<Void> syncToMetadataStore(String key, ServiceUnitStateData data) {
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -0,0 +1,186 @@
...
+                if (futures.size() == MAX_CONCURRENT_SYNC_COUNT) {
+                    FutureUtil.waitForAll(futures).get(opTimeout, TimeUnit.SECONDS);
+                    futures.clear();
+                }
+            }
+            FutureUtil.waitForAll(futures).get(opTimeout, TimeUnit.SECONDS);
+            futures.clear();
+
+            int size = merged.size();
+            int syncTimeout = opTimeout * size;
```
Demogorgon314, line 119: Is the sync timeout too big here? If `opTimeout` is `30` and the size is `10`, then the `syncTimeout` will be `300` seconds.

heesung-sn: sure. let me add some cap here, for 5 mins here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -0,0 +1,186 @@
...
+            systemTopicTableView.start(
+                    pulsar,
+                    this::syncToMetadataStore,
+                    (k, v) -> {}
+            );
+            log.info("Started SystemTopicTableView");
+
+            Map<String, ServiceUnitStateData> merged = new HashMap<>();
+            metadataStoreTableView.entrySet().forEach(e -> merged.put(e.getKey(), e.getValue()));
+            systemTopicTableView.entrySet().forEach(e -> merged.put(e.getKey(), e.getValue()));
```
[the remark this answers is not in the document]
heesung-sn: This is valid concern. I think we can merge the state data based on the timestamp.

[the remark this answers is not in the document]
heesung-sn: I think this is valid concern.

In fact, I think we can change this config `loadBalancerServiceUnitTableViewSyncerEnabled`  to `loadBalancerServiceUnitTableViewSyncer` and make it have "MetadataStoreToSystemTopicSyncer" or "SystemTopicToMetadataStoreSyncer" config vals.

Based on it, we can override the destination view only copied from the source view.

Let me know if you have any concern with this direction.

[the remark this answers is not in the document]
heesung-sn: Updated the code accordingly.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/tableview/impl/MetadataStoreTableViewImpl.java
```diff
@@ -0,0 +1,329 @@
...
+                            } else {
+                                for (var child : children) {
+                                    q.add(path + "/" + child);
+                                }
+                                return CompletableFuture.completedFuture(null);
+                            }
+                        }));
+            }
+            try {
+                FutureUtil.waitForAll(futures).get(timeoutInMillis * size, TimeUnit.MILLISECONDS);
```
BewareMyPower, line 259: This timeout might be too long if the size of children is too large (at most 50*30s = 25 minutes).

Should we just use `timeoutInMillis` instead? Also, I'm wondering if we can ignore the exception (with just a warning log) and just skip the loop in this case. Because `fill()` should not be required. It's just used to establish an initial cache view.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStoreTableView.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    void start() throws MetadataStoreException;
+
+    /**
+     * Reads whether a specific key exists in the local tableview.
+     *
+     * @param key the key to check
+     * @return true if exists. Otherwise, false.
+     */
+    boolean exists(String key);
```
BewareMyPower, line 53: This method is never used except in tests. Please remove it

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
```diff
@@ -0,0 +1,155 @@
...
+        if (tableview == null) {
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public ServiceUnitStateData get(String key) {
+        if (!isValidState()) {
+            new IllegalStateException(INVALID_STATE_ERROR_MSG);
```
BewareMyPower, line 111:
```suggestion
            throw new IllegalStateException(INVALID_STATE_ERROR_MSG);
```

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStoreTableView.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.function.BiConsumer;
+
+/**
+ * Defines metadata store tableview.
+ * MetadataStoreTableView initially fills existing items to its local tableview and eventually
+ * synchronize remote updates to its local tableview from the remote metadata store.
+ * This abstraction can help replicate metadata in memory from metadata store.
+ */
+public interface MetadataStoreTableView<T> {
```
BewareMyPower, line 34: Regarding this interface, I found the following methods are only used in tests.

These methods are also never used except in tests.

```java
    T get(String key);
    int size();
    boolean isEmpty();
    Set<String> keySet();
    Collection<T> values();
    void forEach(BiConsumer<String, T> action);
```

Please remove them. Since this interface is mainly designed to be used internally by `ServiceUnitStateMetadataStoreTableViewImpl`, we'd better keep the interface as simple as possible so that we won't need to write many unnecessary tests because they're never used.

If you're going to verify some internal states in tests, you can just access the Impl's methods or fields directly.

The `org.apache.pulsar.broker.transaction.buffer.impl.TableView` written by me is a good example that it only retains necessary methods that are used internally.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -2874,6 +2874,25 @@ public double getLoadBalancerBandwidthOutResourceWeight() {
...
+    @FieldContext(
+            dynamic = true,
+            category = CATEGORY_LOAD_BALANCER,
+            doc = "Specify ServiceUnitTableViewSyncer to sync service unit(bundle) states between metadata store and "
+                    + "system topic table views during migration from one to the other. One could enable this"
+                    + " syncer before migration and disable it after the migration finishes. "
+                    + "It accepts `MetadataStoreToSystemTopicSyncer` or `SystemTopicToMetadataStoreSyncer` to "
+                    + "enable it. Null value disables it."
+    )
+    private ServiceUnitTableViewSyncerType loadBalancerServiceUnitTableViewSyncer = null;
```
BewareMyPower, line 2894: This config is inconsistent with the PIP-378, see https://github.com/apache/pulsar/blob/master/pip/pip-378.md#configuration

> Add a loadBalancerServiceUnitTableViewSyncerEnabled configuration to to enable ServiceUnitTableViewSyncer to sync metadata store and system topic ServiceUnitStateTableView during migration.

It makes sense because we have two syncer implementations, but it's not friendly to users for how to configure it in the `broker.conf`. Could you use an `enum` instead of `null` for it?

```java
    private ServiceUnitTableViewSyncerType loadBalancerServiceUnitTableViewSyncer = ServiceUnitTableViewSyncerType.None;

    public enum ServiceUnitTableViewSyncerType {
        None,
        MetadataStoreToSystemTopicSyncer,
        SystemTopicToMetadataStoreSyncer
    }
```

Then we can update the `broker.conf` again to include this config.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/tableview/impl/MetadataStoreTableViewImpl.java
heesung-sn: I think we can give 5 mins at most.

I am worried about the worst case with really large table view (although this table view should be relatively small)

Broker would behave wrong if any listener logic fails during this fill.

It is also important to start with complete view for orphan ownership cleanup - the leader needs to scan all current ownerships and reassign orphan ones asap upon any unexpected broker crash.

BewareMyPower: > with really large table view

It should be noted that this timeout is applied in **each loop**. 5 minutes are still large for at most 50 concurrent metadata store operations.

BewareMyPower: Should we use 5 minutes as the total timeout in the loop?

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
BewareMyPower: Besides, you should update the PIP document in this PR to keep implementation and documentation consistent. I think it's acceptable because it's not a big change.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/tableview/impl/MetadataStoreTableViewImpl.java
heesung-sn: ok total 5mins timeout make sense.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStoreTableView.java
heesung-sn: `T get(String key); ` is still required here. I will remove the others.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
heesung-sn: sure. let me update this.

heesung-sn:
@BewareMyPower @Demogorgon314 Updated the code. PTAL thanks.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -0,0 +1,282 @@
...
+
+    private CompletableFuture<Void> syncToSystemTopic(String key, ServiceUnitStateData data) {
+        return systemTopicTableView.put(key, data);
+    }
+
+    private CompletableFuture<Void> syncToMetadataStore(String key, ServiceUnitStateData data) {
+        return metadataStoreTableView.put(key, data);
+    }
+
+    private CompletableFuture<Void> dummy(String key, ServiceUnitStateData data) {
```
Demogorgon314, line 92: Why return a `CompletableFuture<Void>` here? I see the return value has never been used.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/store/TableViewLoadDataStoreImpl.java
```diff
@@ -65,48 +62,84 @@ public TableViewLoadDataStoreImpl(PulsarService pulsar, String topic, Class<T> c
...
+        }
         return Optional.ofNullable(tableView.get(key));
     }
 
     @Override
     public synchronized void forEach(BiConsumer<String, T> action) {
-        validateTableView();
+        String msg = validateTableView();
+        if (StringUtils.isNotBlank(msg)) {
+            throw new IllegalStateException(msg);
```
Demogorgon314, line 104: Can we visibly mark this method will throw an exception, otherwise the caller may miss the catch block?

heesung-sn: I think this can be a runtime exception. I dont think we need a checked one here and make the caller explicitly catch this.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
heesung-sn: This returns future in case it needs to be chained in async.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/tableview/impl/MetadataStoreTableViewImpl.java
```diff
@@ -0,0 +1,339 @@
...
+                            } else {
+                                for (var child : children) {
+                                    q.add(path + "/" + child);
+                                }
+                                return CompletableFuture.completedFuture(null);
+                            }
+                        }));
+            }
+            try {
+                FutureUtil.waitForAll(futures).get(timeoutInMillis, TimeUnit.MILLISECONDS);
```
BewareMyPower, line 269: Not a requested change. But it would be better to use a smaller value of left time and `timeoutInMillis` as the wait timeout.

```java
        final var deadline = System.currentTimeMillis() + FILL_TIMEOUT_IN_MILLIS;
        while (!q.isEmpty()) {
            // ...
            try {
                FutureUtil.waitForAll(futures).get(Math.min(timeoutInMillis, deadline - System.currentTimeMillis()),
                        TimeUnit.MILLISECONDS);
```

BewareMyPower (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
heesung-sn: updated it to return void.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/tableview/impl/MetadataStoreTableViewImpl.java
heesung-sn: updated.

Demogorgon314 (approved):

PR #23120, opened 2024-08-02
[improve][admin] PIP-369 Introduce `unload` flag in `ns-isolation-policy set` call

grssam:
Implementation PR for the PIP 369 "Flag based selective unload on changing ns-isolation-policy" #23116 

### Modifications

Added an `unload-scope` flag in the `ns-isolation-policy set` call to control the behavior of which namespaces to unload as part of the set call in a more granular manner.

As this PR is targetted for 3.0 LTS as well, keeping the changes backward compatible. i.e. the flag `unload-scope` is optional. Default value is `all_matching` which means all namespaces matching the namespaces regex being set in the policy set call.

There will be a followup PR which will be adding the following breaking changes:

- Making the default value of the flag to be `changed` i.e. only the newly added or now-removed namespace regex would be used to unload the namespaces
- The meaning of `all_matching` will change to cover all namespaces matching both the previous and the new namespace regex values. The idea is that if immediate placement correction is needed, we should also unload the namespaces which do not belong to the isolation group anymore.

### Verifying this change

This change added tests and can be verified as follows:

  - Added and made changes to integration tests covering this feature. All three unload scopes are tried and validation is done that only the relevant namespaces are unloaded

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [x] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repo - https://github.com/grssam/pulsar/pull/1

PR #23304, opened 2024-09-13
[improve][pip] PIP-380: Support setting up specific namespaces to skipping the load-shedding

Demogorgon314:


lhotari:
It seems fine. Just thinking that there could be similar requirements for user defined namespaces and therefore having a property in the namespace policy to control this would be useful. Instead of a single binary value, perhaps it could be some type of value that would be used for ordering the namespace bundles that are picked as candidates for load shedding. 
A system namespace could use a value that would result in all other namespace bundles being picked before it becomes a candidate for load shedding.

Demogorgon314:
@lhotari I'm a little worried that adding a new namespace policy may make the concept too complex, and if we truly want a namespace policy we can make another PIP for that future, right?

heesung-sn:
I wonder if we can limit this blacklisting to the TransferShedder under `loadBalanceTransferShedderSheddingExcludedNamespaces` config .

Demogorgon314:
@heesung-sn I would like to use the `loadBalancerSheddingExcludedNamespaces` config to support `ExtensibleLoadManager` and `ModularLoadManager`

heesung-sn:
I think we probably want to explore namespace policy like this work.

https://github.com/apache/pulsar/pull/23120

Demogorgon314:
Adding a namespace policy is a good idea, let me update the PIP.

Demogorgon314:
@lhotari Are you suggesting adding a `loadSheddingPriority` to the namespace police? I'm wondering which case will use the priority for load shedding. And when using priority for each namespace will make the load-shedding strategy let bit complex, since when we select the hot bundle to unload we have to consider the namespace priority. I prefer to use  `loadSheddingEnabled`, then for the protocol's system topic and some user's benchmark topic we can make the ownership stable.

Demogorgon314:
ping @lhotari

pip/pip_379.md
```diff
@@ -0,0 +1,80 @@
+# PIP-379: Support setting up specific namespaces to skipping the load-shedding
```
lhotari, line 1:
```suggestion
# PIP-380: Support setting up specific namespaces to skipping the load-shedding
```

lhotari:
PIP-379 is already taken, I increased the number to PIP-380. The way to find the next available number is to check the dev mailing list and open PRs. The first one to open a discussion for a specific PIP number takes it. This is the PIP-379 discussion: https://lists.apache.org/thread/l5zjq0fb2dscys3rsn6kfl7505tbndlx

lhotari (approved):
LGTM

lhotari:
> @lhotari Are you suggesting adding a `loadSheddingPriority` to the namespace police? I'm wondering which case will use the priority for load shedding. And when using priority for each namespace will make the load-shedding strategy let bit complex, since when we select the hot bundle to unload we have to consider the namespace priority. I prefer to use `loadSheddingEnabled`, then for the protocol's system topic and some user's benchmark topic we can make the ownership stable.

@Demogorgon314 It was something to consider for the future.

lhotari:
@Demogorgon314  Please update the links to the PIP discussion and voting threads.

RobertIndie (approved):

heesung-sn (approved):

PR #4331, opened 2019-05-21
Feature - reset cursor on Reader to current position

lovelle:
*Motivation*

There are some cases in which is it useful to be able to include current
position of message when reset of cursor was made.

This was reported by a `vvy` on slack, no issue has been created to track this.

*Modifications*

  - Add startMessageIdInclusive() to support include current position of
    reset on ReaderBuilder.
  - Add resetIncludeHead field for Reader and Consumer Configuration Data
  - Fix position of cursor for non durable consumer.
  - Improve discard if statement for batch enable mode.
  - Add discard if statement for batch disable mode.
  - Add test case to assert the start of specific message id at the expected
    position with data provider scenarios:
      A. Batch enable and start inclusive enable.
      B. Batch enable and start inclusive disable.
      C. Batch disable and start inclusive enable.
      D. Batch disable and start inclusive disable.

PR #23305, opened 2024-09-13
[fix][broker] Fix NonDurable Subscription msgBackLog incorrect after topic unload

### LINKED ISSUE #23239: [Bug] Non durable subscription backlog is wrong after topic unload
### Search before asking

### Read release policy

### Version

3.3.1
3.0.5

### Minimal reproduce step

1. Execute the following java program:

```java
package org.example;

import java.util.concurrent.atomic.AtomicBoolean;
import org.apache.pulsar.client.api.Consumer;
import org.apache.pulsar.client.api.Message;
import org.apache.pulsar.client.api.Producer;
import org.apache.pulsar.client.api.PulsarClient;
import org.apache.pulsar.client.api.PulsarClientException;
import org.apache.pulsar.client.api.SubscriptionInitialPosition;
import org.apache.pulsar.client.api.SubscriptionMode;

public class Main {

  public static void main(String[] args) throws PulsarClientException, InterruptedException {
    AtomicBoolean running = new AtomicBoolean(true);

    String topic = "tbce0";

    PulsarClient client = PulsarClient.builder()
        .serviceUrl("pulsar://localhost:61723")
        .build();

    Producer<byte[]> producer = client.newProducer()
        .topic(topic)
        .create();

    String message = "Hello, Pulsar!";
    for (int i = 0; i < 1000; i++) {
      producer.send(message.getBytes());
    }
    System.out.println("Sent messages: " + message);
    producer.close();

    Thread.ofVirtual().start(() -> {
      try {
        Consumer<byte[]> consumer = client.newConsumer()
            .topic(topic)
            .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
            .subscriptionName("bce-subscription-0")
            .subscriptionMode(SubscriptionMode.NonDurable)
            //.readCompacted(true)
            .subscribe();

        int counter = 1;
        while (running.get()) {
          Message<byte[]> receivedMessage = consumer.receive();
          System.out.println(
              "Received " + counter++ + " message: " + new String(receivedMessage.getData()));
          consumer.acknowledge(receivedMessage);
        }
        consumer.close();
      } catch (Exception e) {
      }
    });

    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
      System.out.println("Exiting...");
      running.set(false);
      try {
        client.close();
      } catch (PulsarClientException e) {
      }
    }));

    Thread.sleep(600_000);
  }
}
```

2. Wait for the results:

```log
...
Received 993 message: Hello, Pulsar!
Received 994 message: Hello, Pulsar!
Received 995 message: Hello, Pulsar!
Received 996 message: Hello, Pulsar!
Received 997 message: Hello, Pulsar!
Received 998 message: Hello, Pulsar!
Received 999 message: Hello, Pulsar!
Received 1000 message: Hello, Pulsar!

```

3. Run the following script (if on Kubernetes):

```bash
TOPIC=tbce0
kubectl exec -n kaap pulsar-broker-0 -- ./bin/pulsar-admin topics stats $TOPIC | grep msgBacklog
kubectl exec -n kaap pulsar-broker-0 -- ./bin/pulsar-admin topics unload $TOPIC
kubectl exec -n kaap pulsar-broker-0 -- ./bin/pulsar-admin topics stats $TOPIC | grep msgBacklog
```

The cursor should be set to the end of the topic, backlog should be 0.

```
Before unloading
      "msgBacklog" : 0,
      "msgBacklogNoDelayed" : 0,
After unloading
      "msgBacklog" : 0,
      "msgBacklogNoDelayed" : 0,

```

### What did you see instead?

Backlog is not empty:

Before unloading
      "msgBacklog" : 0,
      "msgBacklogNoDelayed" : 0,
After unloading
      "msgBacklog" : 1,
      "msgBacklogNoDelayed" : 1,

### Anything else?

1. You may want to increase the messages retention to not loose it between check
2. Unloading is just the easiest way to reproduce, broker restart may cause the same

summeriiii:
Fixes https://github.com/apache/pulsar/issues/23239

#### Briefly describe the issue: 
Consumer consumed all messages and then the msgBacklog is 0, but after topic unload, the msgBacklog turned to be 1 instead of 0 which is not what we expected. The detail is in test NonDurableSubscriptionTest#testNonDurableSubscriptionBackLogAfterTopicUnload.
    
#### Root cause:
I analysed the related code, and found this issue was introduced by https://github.com/apache/pulsar/pull/4331. In the https://github.com/apache/pulsar/pull/4331, 
`if (((BatchMessageIdImpl) msgId).getBatchIndex() >= 0) {`  was deleted directly, which cause none batch message will also execute the step `entryId = msgId.getEntryId() - 1;`.  Because of this, the entryId reduced by 1, then the msgBacklog turned to be 1 instead of 0.
 
In the https://github.com/apache/pulsar/pull/4331, we added `ReaderBuilder#startMessageIdInclusive` interface to allow create Reader which read containes startMessageId, but it has not been implemented correctly. We should add the field **resetIncludeHead** in CommandSubscribe to implement it.
  
### Modifications
- CommandSubscribe add the field **resetIncludeHead**, and when use the ReaderBuilder#startMessageIdInclusive, this param is true while other is false.
- Persist#getNonDurableSubscription method add the judge condition `(msgId.getBatchIndex() >= 0 || resetIncludeHead)`, entryId -1 will execute Only **when msg is batch or the resetIncludeHead is true.**
    
```java
               if (ledgerId >= 0 && entryId >= 0
                        && msgId instanceof BatchMessageIdImpl
                        && (msgId.getBatchIndex() >= 0 || resetIncludeHead)) {
                    // When the start message is relative to a batch, we need to take one step back on the previous
                    // message,
                    // because the "batch" might not have been consumed in its entirety.
                    // The client will then be able to discard the first messages if needed.
                    entryId = msgId.getEntryId() - 1;
                }
```  
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -889,7 +889,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                 option.getInitialPosition(), option.getStartMessageRollbackDurationSec(),
                 option.isReplicatedSubscriptionStateArg(), option.getKeySharedMeta(),
                 option.getSubscriptionProperties().orElse(Collections.emptyMap()),
-                option.getConsumerEpoch(), option.getSchemaType());
+                option.getConsumerEpoch(), option.getSchemaType(), option.isResetIncludeHead());
     }
 
     private CompletableFuture<Consumer> internalSubscribe(final TransportCnx cnx, String subscriptionName,
@@ -903,7 +903,8 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                                                           KeySharedMeta keySharedMeta,
                                                           Map<String, String> subscriptionProperties,
                                                           long consumerEpoch,
-                                                          SchemaType schemaType) {
+                                                          SchemaType schemaType,
+                                                          boolean resetIncludeHead) {
         if (readCompacted && !(subType == SubType.Failover || subType == SubType.Exclusive)) {
             return FutureUtil.failedFuture(new NotAllowedException(
                     "readCompacted only allowed on failover or exclusive subscriptions"));
@@ -986,7 +987,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                     ? getDurableSubscription(subscriptionName, initialPosition, startMessageRollbackDurationSec,
                             replicatedSubscriptionState, subscriptionProperties)
                     : getNonDurableSubscription(subscriptionName, startMessageId, initialPosition,
-                    startMessageRollbackDurationSec, readCompacted, subscriptionProperties);
+                    startMessageRollbackDurationSec, readCompacted, subscriptionProperties, resetIncludeHead);
 
             CompletableFuture<Consumer> future = subscriptionFuture.thenCompose(subscription -> {
                 Consumer consumer = new Consumer(subscription, subType, topic, consumerId, priorityLevel,
@@ -1070,7 +1071,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                                                  KeySharedMeta keySharedMeta) {
         return internalSubscribe(cnx, subscriptionName, consumerId, subType, priorityLevel, consumerName,
                 isDurable, startMessageId, metadata, readCompacted, initialPosition, startMessageRollbackDurationSec,
-                replicatedSubscriptionStateArg, keySharedMeta, null, DEFAULT_CONSUMER_EPOCH, null);
+                replicatedSubscriptionStateArg, keySharedMeta, null, DEFAULT_CONSUMER_EPOCH, null, false);
     }
 
     private CompletableFuture<Subscription> getDurableSubscription(String subscriptionName,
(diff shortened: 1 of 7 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonDurableSubscriptionTest.java
```diff
@@ -256,6 +256,57 @@ public void testNonDurableSubscriptionRecovery(SubscriptionType subscriptionType
...
+                Message<byte[]> msg;
+                try {
+                    msg = consumer.receive();
+                    consumer.acknowledge(msg);
+                } catch (PulsarClientException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        });
+        t.start();
```
Technoboy-, line 293: we don't need this thread, just consume it directly.

Technoboy-:
re-run the failed test

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1157,7 +1158,8 @@ private CompletableFuture<? extends Subscription> getNonDurableSubscription(Stri
                 long entryId = msgId.getEntryId();
                 // Ensure that the start message id starts from a valid entry.
                 if (ledgerId >= 0 && entryId >= 0
-                        && msgId instanceof BatchMessageIdImpl) {
+                        && msgId instanceof BatchMessageIdImpl
+                        && (msgId.getBatchIndex() >= 0 || resetIncludeHead)) {
```
Technoboy-, line 1162: Adding `resetIncludeHead` should start a PIP. right now, we can only fix by adding `msgId.getBatchIndex() >= 0`

Issue #23307, opened 2024-09-13
[Bug] Message ordering isn't retained in Key_Shared AUTO_SPLIT mode in a rolling restart type of test scenario


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

> [!WARNING]
> It's possible that the test setup contains bugs. More work is needed to validate that the test case is valid.

> [!NOTE]
> It's possible that multiple bugs are involved. The comments describe some observations which might be client side bugs.

This applies to both master branch (which includes PIP-282 changes) as well as in branch-3.3

### Minimal reproduce step

the test case has invocationCount=50 to repeat it 50 times. It usually fails about 25% of the test runs.

This is the test case:
[org.apache.pulsar.client.api.KeySharedSubscriptionTest#testOrderingAfterReconnects](https://github.com/lhotari/pulsar/blob/lh-key_shared-testing-2024-09-13-220512/pulsar-broker/src/test/java/org/apache/pulsar/client/api/KeySharedSubscriptionTest.java#L2311-L2473)
The test code depends on some [new test utility methods](https://github.com/lhotari/pulsar/blob/lh-key_shared-testing-2024-09-13-220512/pulsar-broker/src/test/java/org/apache/pulsar/broker/BrokerTestUtil.java#L120-L218).

reproducing based on master branch (with PIP-282 changes)
```
git clone -b lh-key_shared-testing-2024-09-13-220512 --depth=2 --single-branch https://github.com/lhotari/pulsar
cd pulsar
mvn -Pcore-modules,-main -T 1C clean install -DskipTests -Dspotbugs.skip=true -Dcheckstyle.skip=true -Dlicense.skip=true -DnarPluginPhase=none
mvn -DredirectTestOutputToFile=false -DtestRetryCount=0 test -pl pulsar-broker "-Dtest=org.apache.pulsar.client.api.KeySharedSubscriptionTest#testOrderingAfterReconnects"
```

based on branch-3.3 (without PIP-282)
```
git clone -b lh-key_shared-testing-branch-3.3-2024-09-13-220747 --depth=2 --single-branch https://github.com/lhotari/pulsar
cd pulsar
[...]

PR #23309, opened 2024-09-14
[improve][pip] PIP-379: Key_Shared Draining Hashes for Improved Message Ordering

lhotari:
### Motivation

Read the [pip-379.md document](https://github.com/lhotari/pulsar/blob/lh-pip-379/pip/pip-379.md) in this PR.

[Mailing list discussion thread](https://lists.apache.org/thread/l5zjq0fb2dscys3rsn6kfl7505tbndlx)

implementation PR: #23352 

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


pip/pip-379.md
```diff
@@ -0,0 +1,115 @@
...
+
+### High-Level implementation plan
+
+The proposed solution introduces a "draining hashes" concept to efficiently manage message ordering in Key_Shared subscriptions:
+
+1. When consumer hash ranges change (e.g., a consumer joins or leaves), affected hashes of pending messages are added to a "draining hashes" set.
+2. Following messages with hashes in the "draining hashes" set are blocked from further delivery until pending messages are processed.
+3. A reference counter tracks pending messages for each hash in the "draining hashes" set.
+4. As messages are acknowledged or consumers disconnect and therefore get removed from pending messages, the reference counter is decremented.
```
equanz, line 79: Can this method be used when there are some messages with the same **key hash** but different **keys**?
(I think it does because the draining hash data structure holds the number of messages.)

equanz (commented):
Awesome.

I'd like to ask you some questions as I still don't understand some parts.

pip/pip-379.md
```diff
@@ -0,0 +1,115 @@
...
+
+### High-Level implementation plan
+
+The proposed solution introduces a "draining hashes" concept to efficiently manage message ordering in Key_Shared subscriptions:
+
+1. When consumer hash ranges change (e.g., a consumer joins or leaves), affected hashes of pending messages are added to a "draining hashes" set.
```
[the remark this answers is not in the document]
lhotari: They are the hashes where the assignment changes from an existing consumer to another customer, which could be a new or existing consumer.
Previously there has been an assumption that hash assignments would only happen from existing consumers to new consumers. However #23315 shows that this assumption doesn't hold. 
In PIP-379 design, there's no assumption that hash assignments couldn't change for existing consumers.

pip/pip-379.md
```diff
@@ -0,0 +1,115 @@
...
+2. Following messages with hashes in the "draining hashes" set are blocked from further delivery until pending messages are processed.
+3. A reference counter tracks pending messages for each hash in the "draining hashes" set.
+4. As messages are acknowledged or consumers disconnect and therefore get removed from pending messages, the reference counter is decremented.
+5. When the reference counter reaches zero, the hash is removed from the set, allowing new message delivery.
+6. Consumer hash assignments may change multiple times, and a draining hash might be reassigned to the original consumer. The draining hash data structure contains information about the draining consumer. When a message is attempted for delivery, the system can check if the target consumer is the same as the draining consumer. If they match, there's no need to block the hash. The implementation should also remove such hashes from the draining hashes set. This "lazy" approach reduces the need for actively scanning all draining hashes whenever hash assignments change.
```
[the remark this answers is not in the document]
lhotari: When a consumer disconnects, the messages return to the redelivery queue. There are currently some race conditions related to the timing of when these messages are delivered. This was discussed in https://github.com/apache/pulsar/issues/23307#issuecomment-2360183793 . 

In the current version of PIP-379, the "Updated contract of preserving ordering" is simply "In Key_Shared subscriptions, messages with the same key are delivered and allowed to be in unacknowledged state to only one consumer at a time.". In the implementation of PIP-379, this is an [invariant](https://en.wikipedia.org/wiki/Invariant_(mathematics)#Invariants_in_computer_science) which must hold at all times. Since the invariant is simple, there's much better chances to succeed in holding this invariant at all times in the system. I believe that taking this approach will help detect and resolve possible race conditions that would break this invariant.

pip/pip-379.md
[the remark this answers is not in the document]
lhotari: It could be simply `Map<Integer, DrainingHashEntry>`.
where key is `Integer hash` (16 bytes object overhead + 4 bytes for `int`) and the value is
```java
// 16 bytes overhead
class DrainingHashEntry {
    // AtomicIntegerFieldUpdater could be used to update refCount without an additional wrapper object
    // 4 bytes
    volatile int refCount;
    // 8 bytes
    final long consumerId;
    ...
}
```

A rough byte size estimate of memory usage per entry is 80 bytes (key: 16+16+4 value:16+16+4+8). I shared a [memory usage estimate on the mailing list](https://lists.apache.org/thread/8vrpms6j9m3c4h8bvddzhp6yky3w3lpb):
> Memory usage estimates:
> - Worst case (all 64k hashes draining for a subscription): about 5MB
> - Practical case (less than 1000 hashes draining): less than 80 kilobytes
> - For 10,000 draining hashes: about 800 kB

pip/pip-379.md
lhotari: The different keys wouldn't matter. This solution is for tracking at hash level. The invariant of "In Key_Shared subscriptions, messages with the same key are delivered and allowed to be in unacknowledged state to only one consumer at a time." would still hold. 
It can be considered an implementation detail that the tracking is at hash level. This would be explained to users since the observability (stats/metrics) is at the hash level.

The broker doesn't unpack batch messages. It's the responsibility of the producer to produce batch messages where all messages in the batch contain the same key and therefore have the same hash.

lhotari: The main consequence of hash collisions is that a single key is coupled to any other key with the same hash and would get blocked together.

pip/pip-379.md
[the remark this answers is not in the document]
lhotari: One of the challenges in getting a snapshot of the current hash assignments: "ConsistentHashingStickyKeyConsumerSelector's getConsumerKeyHashRanges doesn't return correct results" #23321

pip/pip-379.md
```diff
@@ -0,0 +1,115 @@
...
+
+This approach will meet the updated contract of preserving ordering while minimizing the impact on performance and memory usage. The tracking only comes into play during transition states. When consumers have been connected for a longer duration and all draining hashes have been removed, there won't be a need to check any special rules or maintain any extra state.
+
+## Public-facing Changes
+
+### Topic Stats Changes & Observability
+
+Topic stats for the removed PIP-282 "recently joined consumers"/"last sent position" solution are removed:
+- `lastSentPositionWhenJoining` field for each consumer
+- `consumersAfterMarkDeletePosition` field for each Key_Shared subscription
```
codelipenghui, line 91: This is a break change for the existing users for how they troubleshoot Key_Shared subscription issues today.

lhotari: There's a replacement. It's not a breaking change in that sense. That's also the reason why there's a benefit that this goes into 4.0 since in major versions breaking changes are expected. I'm sure that nobody will miss using these fields! If they do, they should remain on Pulsar 3.x.

There won't even be a need to troubleshoot consumer stuck issues in PIP-379 design in the same way. There's a replacement for cases where a consumer really doesn't acknowledge a message. There's also an optional REST API for finding out the actual message IDs of the unacked messages that could be used to assist the troubleshooting.

codelipenghui: That is a break change, we changed the REST API without any smooth transition. The field `consumersAfterMarkDeletePosition` will be delete, some users might build metrics or something on this field. We need to keep the default behavior consistent for all the APIs exposed to users. We can add more fields, but should not remove the existing one or change the meaning of the existing one.

3.0 to 4.0 should not have break change, we promised user that they upgrade from 3.0 to 4.0 directly. https://pulsar.apache.org/contribute/release-policy/#compatibility-between-releases.

lhotari: > That is a break change, we changed the REST API without any smooth transition. The field `consumersAfterMarkDeletePosition` will be delete, some users might build metrics or something on this field. We need to keep the default behavior consistent for all the APIs exposed to users. We can add more fields, but should not remove the existing one or change the meaning of the existing one.
> 
> 3.0 to 4.0 should not have break change, we promised user that they can upgrade from 3.0 to 4.0 directly without any compatibility issue. https://pulsar.apache.org/contribute/release-policy/#compatibility-between-releases. 

It's change, but not a breaking change. This is not a compatibility in the sense that our release policy expresses. That's about bookkeeper storage formats and such major details.

codelipenghui: I can't agree more. If we follow this way, we can delete or modify any output for the REST API.

pip/pip-379.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Future work in needed for supporting key-based ordering with negative acknowledgements
+
+The updated contract explicitly states that it is not possible to retain key-based ordering of messages when negative acknowledgements are used. Changing this is out of scope for PIP-379. A potential future solution for handling this would be to modify the client so that when a message is negatively acknowledged, it would also reject all further messages with the same key until the original message gets redelivered. It's already possible to attempt to implement this in client-side code. However, a proper solution would require support on the broker side to block further delivery of the specific key when there are pending negatively acknowledged messages until all negatively acknowledged messages for that particular key have been acknowledged by the consumer. This solution is out of scope for PIP-379. A future implementation to address these problems could build upon PIP-379 concepts such as "draining hashes" and extend that to cover the negative acknowledgement scenarios.
+
+### High-Level implementation plan
```
codelipenghui, line 72: I tried to understand how the new mechanism works. But I don't really understand how can it "Unnecessary Message Blocking". @lhotari Could you please give a simple example like what happened when the first consumer joined, and then the second consumer, third consumer? So that we can understand the approach in a clear manner.

And I think @equanz also has most of the questions about this part.

pip/pip-379.md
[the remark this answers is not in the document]
codelipenghui: It should be defined in the proposal, not comment. Otherwise, the reviewers might not understand how it works in details.

pip/pip-379.md
lhotari: @codelipenghui [PIP-282](https://github.com/apache/pulsar/blob/master/pip/pip-282.md) already changed the fields and the meaning of the fields. Have you considered that?

pip/pip-379.md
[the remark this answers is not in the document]
lhotari: @codelipenghui Thanks. I'll make sure to include this in the document.

pip/pip-379.md
lhotari: @codelipenghui 

I'll improve the document later to include a better explanation.

I'll try to describe it also in this comment. So if we think about it, there's "necessary message blocking" that is absolutely necessary. The invariant is "In Key_Shared subscriptions, messages with the same key are delivered and allowed to be in unacknowledged state to only one consumer at a time." .
All other message blocking outside of this is unnecessary.

Since we only block messages to ensure that the invariant holds, it's reducing all unnecessary blocking.

In our invariant, we mention that we need to ensure that a same key cannot be in unacked state to multiple consumers. In the implementation, the implementation detail is to do the tracking at hash level instead of key level. The invariant will still hold. If all "unnecessary blocking" would be eliminated, the tracking would be at key level. However we want to make a tradeoff between memory usage and minimal blocking and that's why using a hash makes sense in the implementation. 

Does this explanation of "unnecessary" and "necessary" blocking make sense?

I could add the simple example to the documentation to explain how the solution would work.

pip/pip-379.md
codelipenghui: Yes, the proposal [PIP-282](https://github.com/apache/pulsar/blob/master/pip/pip-282.md) should provide an option for user to move smoothly. Instead of changing meaning or rename the output of public API directly.

pip/pip-379.md
codelipenghui: @lhotari I understand what is the "unnecessary blocking" and what is the "necessary message blocking". My question is the document only show the very high level of Drained Hash, if there is an example, step by step for consumer join, leaving, then we can more easier to understand the approach.

Now I guess the solution is for each delivered message, we will record a hash for it hash -> (consumer Id + counter) and increase the counter if the hash already exist, and decrease the counter if the message get acked?

pip/pip-379.md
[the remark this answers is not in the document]
lhotari: @equanz I have implemented a PoC where these details are clarified in code. I shared the update on the dev mailing list, https://lists.apache.org/thread/vrsjzkk3lnzq2dqhg380740hr9ftdn4t.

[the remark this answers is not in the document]
lhotari: I shared the update on the dev mailing list, https://lists.apache.org/thread/vrsjzkk3lnzq2dqhg380740hr9ftdn4t. It also explains the data structure and points to the PoC implementation. 
The update also explains that the PIP document will be updated asap.

pip/pip-379.md
lhotari: @codelipenghui  I've added a comment this in the update to the mailing list, https://lists.apache.org/thread/vrsjzkk3lnzq2dqhg380740hr9ftdn4t.

pip/pip-379.md
lhotari: > Now I guess the solution is for each delivered message, we will record a hash for it hash -> (consumer Id + counter) and increase the counter if the hash already exist, and decrease the counter if the message get acked?

The existing "pending acks" is used for tracking. The draining hashes counter is only needed when a hash should switch to "draining" state. That happens when the hash gets assigned to a new consumer, but there are outstanding unacked messages for that hash. The PoC contains exact details of how this is implemented and covered. I also found a specific race condition that I solved in the PoC. The reference to the PoC PR is in the mailing list update, https://lists.apache.org/thread/vrsjzkk3lnzq2dqhg380740hr9ftdn4t.

pip/pip-379.md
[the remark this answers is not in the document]
equanz: I understand it better now. Thank you.

pip/pip-379.md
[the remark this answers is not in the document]
equanz: Okay. I think I understand your point.
But my question is which consumer can receive redelivery messages that were held by the disconnected consumer?

However, I began to think that it would simply be the consumer that the selector returns.

pip/pip-379.md
```diff
@@ -0,0 +1,390 @@
...
+
+This is handled in the `DrainingHashesTracker`
+```java
+    public synchronized boolean shouldBlockStickyKeyHash(Consumer consumer, int stickyKeyHash) {
+        DrainingHashEntry entry = drainingHashes.get(stickyKeyHash);
+        if (entry == null) {
+            return false;
+        }
+        // hash has been reassigned to the original consumer, remove the entry
+        if (entry.getConsumerId() == consumer.consumerId()) {
```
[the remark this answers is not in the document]
lhotari: > The consumerId is unique within PulsarClient. Is this spec affected to your proposal?

@equanz Thanks for pointing this out. I'll update this detail in the implementation PR and also in the PIP document.
Instead of using the consumerId, I'll use the Consumer instance and do instance comparison.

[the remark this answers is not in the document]
lhotari: @equanz I updated the implementation PR with commit https://github.com/apache/pulsar/pull/23352/commits/e9fa63f2f77d300e899c6b4faa4d411893dcdbdd and the PIP document PR with commit https://github.com/apache/pulsar/pull/23309/commits/57705d551b392cfe5c2ca0af5582552764ff7b54 to address this. PTAL

pip/pip-379.md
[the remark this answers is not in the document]
lhotari: > Okay. I think I understand your point. But my question is which consumer can receive redelivery messages that were held by the disconnected consumer?
> 
> However, I began to think that it would simply be the consumer that the selector returns.

@equanz Yes, messages are always delivered to the consumer that the selector returns.

equanz (approved):
LGTM
I look forward to achieving this goal.

PR #23311, opened 2024-09-16
[improve][broker] Decouple ManagedLedger interfaces from the current implementation

lhotari:
## Motivation

This PR is part of an initiative to clean up and decouple the ManagedLedger interfaces from their current implementation in preparation for Pulsar 4.0. The primary goals are:

1. To allow for alternative ManagedLedger implementations in Pulsar 4.0.
2. To improve the overall architecture by reducing coupling between core Pulsar components and specific ManagedLedger implementations.

This work stems from a community discussion on the Apache Pulsar mailing list: [Preparing for Pulsar 4.0: cleaning up the Managed Ledger interfaces](https://lists.apache.org/thread/b2f4vkql693x3frdxm75tndk686crh9k)

Since the ManagedLedger interface is not exposed externally and this PR doesn't introduce implementation changes, a formal PIP (Pulsar Improvement Proposal) is not required. The mailing list discussion and the review process for this PR are considered sufficient for proceeding with these changes.

## Scope

This PR aims to make minimal, focused changes to achieve the decoupling goal. It does not include:
- Adding comprehensive JavaDocs to the interfaces (this will be addressed in future PRs)
- Changing any external APIs or behaviors

## Modifications

1. **Decouple from `ManagedLedgerImpl`:**
   - Add required methods from `ManagedLedgerImpl` to the `ManagedLedger` interface
   - Update relevant code to use the interface instead of the implementation

2. **Decouple from `ManagedLedgerFactoryImpl`:**
   - Add necessary methods from `ManagedLedgerFactoryImpl` to the `ManagedLedgerFactory` interface
   - Refactor code to depend on the interface rather than the concrete implementation

3. **Decouple from `ManagedCursorImpl`:**
   - Enhance the `ManagedCursor` interface with required methods from `ManagedCursorImpl`
   - Update code to use the interface methods instead of implementation-specific ones

4. **Introduce `ReadOnlyManagedLedger` interface:**
   - Extract this interface to decouple from `ReadOnlyManagedLedgerImpl`
   - Adjust code to utilize the new interface where appropriate

5. **Decouple `OpAddEntry` from `ManagedLedgerInterceptor`:**
   - add interface `AddEntryOperation` to be used by the `beforeAddEntry` method
   - revisit the method return type since it should be `void`
   - Adjust code to utilize the new interface where appropriate
   - Eliminate the use of `AddEntryOperation` in `processPayloadBeforeLedgerWrite` since mutation should happen with the `PayloadProcessorHandle` in that case.

6. **Decouple `LedgerHandle` from `ManagedLedgerInterceptor`:**
   - add interface `LastEntryHandle` to be used by the `onManagedLedgerLastLedgerInitialize` method
   - Adjust code to utilize the new interface where appropriate

7. **Additional minor changes:**
   - Adjust method signatures and parameter types as needed
   - Refactor any code that directly referenced implementation classes

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/PositionBound.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/PositionBound.java
@@ -1,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.mledger;
+
+public enum PositionBound {
+    // define boundaries for position based seeks and searches
+        startIncluded, startExcluded
+}
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/PositionBound.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/PositionBound.java
@@ -20,5 +20,5 @@ package org.apache.bookkeeper.mledger;
 
 public enum PositionBound {
     // define boundaries for position based seeks and searches
-        startIncluded, startExcluded
+    startIncluded, startExcluded
 }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerOfflineBacklog.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerOfflineBacklog.java
@@ -20,29 +20,16 @@ package org.apache.bookkeeper.mledger.impl;
 
 import com.google.common.collect.BoundType;
 import com.google.common.collect.Range;
-import com.google.protobuf.InvalidProtocolBufferException;
-import java.util.Enumeration;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.NavigableMap;
-import java.util.concurrent.ConcurrentSkipListMap;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-import org.apache.bookkeeper.client.AsyncCallback;
-import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.client.BookKeeper;
-import org.apache.bookkeeper.client.BookKeeperAdmin;
-import org.apache.bookkeeper.client.LedgerEntry;
-import org.apache.bookkeeper.client.LedgerHandle;
 import org.apache.bookkeeper.client.api.DigestType;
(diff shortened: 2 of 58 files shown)


lhotari:
A remaining challenge is ManagedLedgerInterceptor which couples to implementation details, OpAddEntry.
That will be handled in a separate PR.
There's also a need to make some methods asynchronous and that would be handled separately.

lhotari:
ManagedLedgerInterceptor's onManagedLedgerLastLedgerInitialize couples directly to Bookkeeper client's LedgerHandle. That needs to be decoupled too.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerOfflineBacklog.java
```diff
@@ -102,369 +89,28 @@ private long getNumberOfEntries(Range<Position> range,
...
+        if (factory instanceof ManagedLedgerFactoryImpl) {
+            List<Object> ctx = new ArrayList<>();
+            ctx.add(digestType);
+            ctx.add(password);
+            factory.estimateUnloadedTopicBacklog(offlineTopicStats, topicName, accurate, ctx);
         } else {
-            mlMetaCounter.await(META_READ_TIMEOUT_SECONDS, TimeUnit.SECONDS);
-
+            Object ctx = null;
+            factory.estimateUnloadedTopicBacklog(offlineTopicStats, topicName, accurate, ctx);
```
BewareMyPower, line 109: This `estimateUnloadedTopicBacklog` API is not intuitive.
1. Why it needs an `accurate` parameter? (actually it's always false) 
2. If the factory is not a `ManagedLedgerFactoryImpl`, the `ctx` field is always null.

image

It's not easy for the author of a 3rd party ML to know how to implement this method correctly without enough API docs. The API design is bad. It should be something like:

```java
   PersistentOfflineTopicStats estimateUnloadedTopicBacklog(TopicName topicName);
```

BTW, IMHO, this API should be moved to an internal method of `ManagedLedgerFactoryImpl`, but it might need another proposal to discuss about the API design.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/PositionBound.java
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.mledger;
+
+public enum PositionBound {
+    // define boundaries for position based seeks and searches
+        startIncluded, startExcluded
```
BewareMyPower, line 23:
```suggestion
    startIncluded, startExcluded
```

Adjust indent

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerOfflineBacklog.java
lhotari: > BTW, IMHO, this API should be moved to an internal method of `ManagedLedgerFactoryImpl`, but it might need another proposal to discuss about the API design.

Yes, this is a mess that should be untangled. The usage is here: https://github.com/apache/pulsar/blob/faed541bc64ed112eb3c61abd9cdda10ecaeed1e/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java#L3144-L3184

BewareMyPower: Since this PR's motivation is mainly to decouple the ML interface from the Impls, this comment can be improved in a future PR

BewareMyPower (approved):

PR #23313, opened 2024-09-17
[improve][broker][PIP-384] Decouple Bookkeeper client from ManagedLedgerStorage and enable multiple ManagedLedgerFactory instances

lhotari:
This PR is partial implementation for "PIP-384: ManagedLedger interface decoupling", #23363.

## Motivation

This PR is part of an initiative to clean up and decouple the ManagedLedger interfaces from their current implementation in preparation for Pulsar 4.0. The primary goals are:

1. To allow for alternative ManagedLedger implementations in Pulsar 4.0.
2. To improve the overall architecture by reducing coupling between core Pulsar components and specific ManagedLedger implementations.

This work stems from a community discussion on the Apache Pulsar mailing list: [Preparing for Pulsar 4.0: cleaning up the Managed Ledger interfaces](https://lists.apache.org/thread/b2f4vkql693x3frdxm75tndk686crh9k)

This PR adds supports for multiple ManagedLedger implementations when using a custom ManagedLedgerStorage implementation. The benefit of this is that a Bookkeeper implementation can co-exist with a custom ManagedLedger implementation in the same Pulsar cluster. 

## Scope

This PR aims to make minimal, focused changes to achieve the decoupling goal. It does not include:
- Adding comprehensive JavaDocs to the interfaces (this will be addressed in future PRs)
- Changing any external APIs or behaviors

### Modifications

- add `managedLedgerStorageClassName` to `PersistencePolicies` to control the `ManagedLedgerFactory` instance that is used for a particular namespace or topic.
- add `storageClassName` to `ManagedLedgerConfig`, this field is read from `PersistencePolicies` (namespace or topic policies)
- modify `ManagedLedgerStorage` interface to support getting a `ManagedLedgerFactory` with the `storageClassName`
- change topics to use this for selecting the `ManagedLedgerFactory` instance
- for existing code that is not aware of multiple choices, use the default storage class and the default ManagedLedgerFactory instance.

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/TransactionMetadataStoreService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/TransactionMetadataStoreService.java
@@ -227,7 +227,9 @@ public class TransactionMetadataStoreService {
                 .setBatchedWriteMaxDelayInMillis(serviceConfiguration.getTransactionLogBatchedWriteMaxDelayInMillis());
 
         return pulsarService.getBrokerService().getManagedLedgerConfig(getMLTransactionLogName(tcId)).thenCompose(
-                v -> transactionMetadataStoreProvider.openStore(tcId, pulsarService.getManagedLedgerFactory(), v,
+                v -> transactionMetadataStoreProvider.openStore(tcId,
+                        pulsarService.getManagedLedgerStorage().getManagedLedgerStorageClass(v.getStorageClassName())
+                                .get().getManagedLedgerFactory(), v,
                         timeoutTracker, recoverTracker,
                         pulsarService.getConfig().getMaxActiveTransactionsPerCoordinator(), txnLogBufferedWriterConfig,
                         brokerClientSharedTimer));
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/storage/ManagedLedgerStorage.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/storage/ManagedLedgerStorage.java
@@ -21,9 +21,8 @@ package org.apache.pulsar.broker.storage;
 import io.netty.channel.EventLoopGroup;
 import io.opentelemetry.api.OpenTelemetry;
 import java.io.IOException;
-import org.apache.bookkeeper.client.BookKeeper;
-import org.apache.bookkeeper.mledger.ManagedLedgerFactory;
-import org.apache.bookkeeper.stats.StatsProvider;
+import java.util.Collection;
+import java.util.Optional;
 import org.apache.pulsar.broker.BookKeeperClientFactory;
 import org.apache.pulsar.broker.ServiceConfiguration;
 import org.apache.pulsar.common.classification.InterfaceAudience.Private;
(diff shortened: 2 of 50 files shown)


merlimat (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/storage/ManagedLedgerStorage.java
```diff
@@ -52,25 +57,25 @@ void initialize(ServiceConfiguration conf,
...
     /**
-     * Return the default bookkeeper client.
-     *
-     * @return the default bookkeeper client.
+     * Lookup a storage class by name.
+     * @param name storage class name
+     * @return storage class instance, or empty if not found
      */
-    BookKeeper getBookKeeperClient();
+    Optional<ManagedLedgerStorageClass> getManagedLedgerStorageClass(String name);
```
BewareMyPower, line 78: This method is hard to understand. From the perspective of customizing a managed ledger storage, what I want to know is that when could this method be called and how could the `name` be.

It seems that the name comes from `ManagedLedgerConfig#storageClassName`. But how could this config be changed?

BewareMyPower: It seems that it can only be set via the topic policies.

pulsar-broker/src/main/java/org/apache/pulsar/broker/TransactionMetadataStoreService.java
```diff
@@ -227,7 +227,9 @@ public CompletableFuture<Void> handleTcClientConnect(TransactionCoordinatorID tc
                 .setBatchedWriteMaxDelayInMillis(serviceConfiguration.getTransactionLogBatchedWriteMaxDelayInMillis());
 
         return pulsarService.getBrokerService().getManagedLedgerConfig(getMLTransactionLogName(tcId)).thenCompose(
-                v -> transactionMetadataStoreProvider.openStore(tcId, pulsarService.getManagedLedgerFactory(), v,
+                v -> transactionMetadataStoreProvider.openStore(tcId,
+                        pulsarService.getManagedLedgerStorage().getManagedLedgerStorageClass(v.getStorageClassName())
+                                .get().getManagedLedgerFactory(), v,
```
BewareMyPower, line 232: NPE will happen if `getManagedLedgerStorageClass` returns empty. This case is not handled well

pulsar-broker/src/main/java/org/apache/pulsar/broker/storage/ManagedLedgerStorage.java
lhotari: This PR doesn't consider how the `ManagedLedgerStorage` implementation configures the default value. It's possible to override the storage class name at namespace or topic level in persistence policies.

pulsar-broker/src/main/java/org/apache/pulsar/broker/TransactionMetadataStoreService.java
lhotari: That's correct. This could be improved.

PR #23319, opened 2024-09-19
[improve][broker] PIP-376: Make topic policies service pluggable

BewareMyPower:
PIP: https://github.com/apache/pulsar/pull/23248

### Modifications

Adapting to the new interface according to the PIP documents.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/35

PR #23326, opened 2024-09-20
[fix][broker] Fix topic policies cannot be queried with extensible load manager

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/23319 brings a regression that topic policies cannot be queried with extensible load manager.  `compactionScheduleTest` could not succeed now.

In `SystemTopicBasedTopicPoliciesService#getTopicPoliciesAsync`, it skips getting events from `__change_events` because the reader will fail to create if the load manager does not start.

https://github.com/apache/pulsar/blob/4b3b273c1c57741f9f9da2118eb4ec5dfeee2220/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java#L253-L259

However, the state validation in `ServiceUnitStateChannelImpl` here is wrong: https://github.com/apache/pulsar/blob/4b3b273c1c57741f9f9da2118eb4ec5dfeee2220/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java#L270

The method above returns true when the state is `Closed`, `Constructed` or `LeaderElectionServiceStarted`, which is the opposite of the expected result (`Started`).

### Modifications

Fix the `ServiceUnitStateChannelImpl#started()` method with the correct state validation.

However, if only with this simple fix, the extensible load manager would fail to start. It's because after the table view creation for `loadbalancer-broker-load-data` and `loadbalancer-top-bundles-load-data` topics happen after the load manager start. In `isAllowAutoSubscriptionCreationAsync`, it will wait until the future of `getTopicPoliciesAsync` on these topics is done, while the reader creation on `__change_events` will fail.

Therefore, to fix this start failure, just return true for internal topics in `isAllowAutoSubscriptionCreationAsync`.

### Verifying this change

`ExtensibleLoadManagerImplTest#compactionScheduleTest` will pass now.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
@@ -267,7 +267,7 @@ public class ServiceUnitStateChannelImpl implements ServiceUnitStateChannel {
 
     @Override
     public synchronized boolean started() {
-        return validateChannelState(LeaderElectionServiceStarted, false);
+        return validateChannelState(Started, true);
     }
 
     public synchronized void start() throws PulsarServerException {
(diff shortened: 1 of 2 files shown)


heesung-sn (approved):

lhotari (approved):

BewareMyPower:
`ExtensibleLoadManagerCloseTest` failed now. Let me take a look

BewareMyPower:
I skipped `getTopicPoliciesAsync` calls for extensible LM's internal topics in `BrokerService`'s topic initialization, let's see if all tests can pass now.

PR #23018, opened 2024-07-09
[fix] [broker] Internal reader of __change_events can not started after metadata store session rebuilt

poorbarcode:
#### 1. Fix the issue below
- settings: enable `systemTopic & topicLevelPolicies`.
- Own topics
  - `broker-2` loads a `topic-a` up.
  - `broker-2` loads system topic(`{ns}/__change_events`) up.
- The ZK session expires.
- Bundle shedding 
  - `topic-a` was transferred to `broker-0`
  - the system topic(`{ns}/__change_events`) was transferred to `broker-0`
- **(Highlight)** Issue occurs on `broker-2`
  -  The internal reader for `{ns}/__change_events` lookup the topic's owner
  - `broker-2` responds `broker-2` due to its `OwnershipCache` contains the old value.
  - The internal reader for `{ns}/__change_events` can not be created successfully, all topics under this namespace can not work.
  - The consumers/producers who call lookup to `broker-2` will get a wrong value, and keep getting a "Please redo the lookup" error.

#### 2. Add a test to confirm the topic will be unloaded after the metadata store session is rebuilt.
- Broker load a topic up.
- The ZK session expires.
- The ZK session rebuild.
- Verify: The topics that the broker owned will be unloaded

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #23334, opened 2024-09-23
[Draft] [fix] [broker] Do not record a bundle-unloading into the topic load failed metrics

poorbarcode:

Pulsar has a metric that indicates load topic failed: `topic_load_failed_total`, it reflects something is not expected, such as 
- Not enough bookies
- Failed to connect to the Metadata store

But the topic loaded failed due to a bundle in unloading is expected, we should not record it into this metrics.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerServiceTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerServiceTest.java
@@ -67,7 +67,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.Cleanup;
@@ -81,7 +80,6 @@ import org.apache.bookkeeper.mledger.impl.ManagedLedgerFactoryImpl;
 import org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl;
 import org.apache.bookkeeper.mledger.impl.NullLedgerOffloader;
 import org.apache.bookkeeper.mledger.impl.NonAppendableLedgerOffloader;
-import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.CloseableHttpClient;
@@ -1679,8 +1677,8 @@ public class BrokerServiceTest extends BrokerTestBase {
         admin.topics().unload(topic);
 
         // Get original counter.
-        AtomicLong failedLoadTopic1 = new AtomicLong(0);
-        AtomicLong concurrencyLoadTopicAndUnloadBundle1 = new AtomicLong(0);
+        final AtomicLong failedLoadTopic1 = new AtomicLong(0);
+        final AtomicLong concurrencyLoadTopicAndUnloadBundle1 = new AtomicLong(0);
         JerseyClient httpClient = JerseyClientBuilder.createClient();
         String response = httpClient.target(pulsar.getWebServiceAddress()).path("/metrics/")
                 .request().get(String.class);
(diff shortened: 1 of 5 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerServiceTest.java
```diff
@@ -1674,6 +1676,18 @@ public void testMetricsPersistentTopicLoadFails() throws Exception {
         admin.topics().createNonPartitionedTopic(topic);
         admin.topics().unload(topic);
 
+        // Get original counter.
+        MutableInt failedLoadTopic1 = new MutableInt(0);
+        MutableInt concurrencyLoadTopicAndUnloadBundle1 = new MutableInt(0);
```
lhotari, line 1681: since multiple threads are involved, shouldn't these be AtomicIntegers?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerServiceTest.java
```diff
@@ -1710,6 +1719,74 @@ public void testMetricsPersistentTopicLoadFails() throws Exception {
...
+    public void testMetricsPersistentTopicLoadFailsDueToBundleUnloading() throws Exception {
+        final String namespace = "prop/" + UUID.randomUUID().toString().replaceAll("-", "");
+        String topic = "persistent://" + namespace + "/topic1_" + UUID.randomUUID();
+        admin.namespaces().createNamespace(namespace);
+        admin.topics().createNonPartitionedTopic(topic);
+        admin.namespaces().unload(namespace);
+
+        // Get original counter.
+        MutableInt failedLoadTopic1 = new MutableInt(0);
+        MutableInt concurrencyLoadTopicAndUnloadBundle1 = new MutableInt(0);
```
lhotari, line 1732: Use AtomicIntegers since multiple threads are involved.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerServiceTest.java
lhotari: I don't currently understand the reason to use `MutableInt`s. 
If the reason is to that a final variable is needed for an inner lambda, it might be better to use a final variable.
One trick that could be useful is to isolate variable scopes with a `{....}` block. That would allow to introduce the same variable name multiple times in a single unit test without variable name collisions since the variables would e in different scopes.

heesung-sn:
> But the topic loaded failed due to a bundle in unloading is expected, we should not record it into this metrics.

Don't we need to count this case as failure as well? To me if the topic load failed on the broker and client lookup failed, it should be counted as a failure.

I think this is one area that Extensible Load Manager can shine as it handles lookups more gracefully during unloading -- the topic lookups are deferred until the unloading completes in the service unit state channel. Hence, the client won't see failure, and the lookup response should be slightly delayed instead of returning error code(no retry as well).

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerServiceTest.java
poorbarcode: Fixed

poorbarcode:
@heesung-sn 

> Don't we need to count this case as failure as well? To me if the topic load failed on the broker and client lookup failed, it should be counted as a failure.

It is useful to check if there is a bug in `LoadBalancer`, such as: if the variable keep increasing fastly for a long time, maybe the a broker cached a wrong bundle owner or DNS issue

heesung-sn:
I still think this is a still failure case. Can we add a "reason" label instead of adding a new metric?

poorbarcode:
@heesung-sn 

> I still think this is a still failure case. Can we add a "reason" label instead of adding a new metric?

I will start a PIP for this metric, it is useful to detect the following issue:
- https://github.com/apache/pulsar/pull/23018

poorbarcode:
Created a new PR https://github.com/apache/pulsar/pull/26264

poorbarcode:
Created https://github.com/apache/pulsar/pull/26264 instead

PR #23336, opened 2024-09-23
[fix] FIx ArrayIndexOutOfBoundsException when using SameAuthParamsLookupAutoClusterFailover

poorbarcode:
### Motivation

There is a [mistake](https://github.com/apache/pulsar/pull/23336/files#diff-d1cfc9f5db8cc65b6f7ec1b04a3bfda01230a2f16610069c776714dc67134eefR133) in `SameAuthParamsLookupAutoClusterFailover.class`, which may cause a `ArrayIndexOutOfBoundsException`, but it affects nothing because it only happens when no server is available.

### Modifications

- Fix the mistake
- Add an log

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/SameAuthParamsLookupAutoClusterFailover.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/SameAuthParamsLookupAutoClusterFailover.java
@@ -71,29 +71,36 @@ public class SameAuthParamsLookupAutoClusterFailover implements ServiceUrlProvid
         this.executor = EventLoopUtil.newEventLoopGroup(1, false,
                 new ExecutorProvider.ExtendedThreadFactory("broker-service-url-check"));
         scheduledCheckTask = executor.scheduleAtFixedRate(() -> {
-            if (closed) {
-                return;
-            }
-            checkPulsarServices();
-            int firstHealthyPulsarService = firstHealthyPulsarService();
-            if (firstHealthyPulsarService == currentPulsarServiceIndex) {
-                return;
-            }
-            if (firstHealthyPulsarService < 0) {
-                int failoverTo = findFailoverTo();
-                if (failoverTo < 0) {
-                    // No healthy pulsar service to connect.
-                    log.error("Failed to choose a pulsar service to connect, no one pulsar service is healthy. Current"
-                            + " pulsar service: [{}] {}. States: {}, Counters: {}", currentPulsarServiceIndex,
-                            pulsarServiceUrlArray[currentPulsarServiceIndex], Arrays.toString(pulsarServiceStateArray),
-                            Arrays.toString(checkCounterArray));
-                } else {
-                    // Failover to low priority pulsar service.
-                    updateServiceUrl(failoverTo);
+            try {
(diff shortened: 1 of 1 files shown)


lhotari:
Any chance to add a test case?

Technoboy- (approved):

lhotari:
> Any chance to add a test case?

@poorbarcode @Technoboy- any chance to handle this?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/SameAuthParamsLookupAutoClusterFailover.java
```diff
@@ -123,7 +130,7 @@ private int firstHealthyPulsarService() {
     }
 
     private int findFailoverTo() {
-        for (int i = currentPulsarServiceIndex + 1; i <= pulsarServiceUrlArray.length; i++) {
+        for (int i = currentPulsarServiceIndex + 1; i < pulsarServiceUrlArray.length; i++) {
```
lhotari, line 133: Just wondering if this logic is correct in the first place. Shouldn't this wrap around the array?

the for loop in `checkPulsarServices` looks strange too.

lhotari:
@poorbarcode Do you have a chance to follow up on the review comment?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/SameAuthParamsLookupAutoClusterFailover.java
poorbarcode: The logic of fallover is follows
- if the current cluster is the highest priority cluster and it is healthy, do nothing.
- else if there is a healthy and high priority cluster, recover to the higher priority one.
- else if both current cluster and higher priority clusters are not healthy, find a healthy cluster to fallover to.

So the two method `checkPulsarServices` and `findFailoverTo` are for the different cases to same CPU resouces.
- `checkPulsarServices` is for checking whether their is a healthy and high priority cluster.
- `findFailoverTo` is for checking whether their has a cluster is healthy.

poorbarcode:
> @poorbarcode Do you have a chance to follow up on the review comment?

Answered here: https://github.com/apache/pulsar/pull/23336#discussion_r1865112171

poorbarcode:
Rebase master

dao-jun (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/SameAuthParamsLookupAutoClusterFailover.java
poorbarcode: > what happens in the case where currentPulsarServiceIndex is already the last one? How can it find a failover cluster in that case?

It will transfer to the first healthy cluster. Once encounters the `ArrayIndexOutOfBoundsException` error, it means no cluster is healthy.

The issue the correct PR solved affects nothing

lhotari:
Seems that #24662 was related.

PR #23340, opened 2024-09-23
[fix] Key_Shared mode consumption latency when low traffic

poorbarcode:
### Motivation

Without #23226 and #23284, the consumption latency is low, but the consumption latency increases to `1s` when the traffic is low.

Consumption latency after an upgrade(3 topics).

`histogram_quantile(0.98, sum-by (le, topic) (rate (Message_duration_eventstate_ inbound_latencymin_bucket {kubernetes_namespace-"$(env)"} [$__rate_interval])))`

Screenshot 2024-09-24 at 00 27 11

Root cause

Screenshot 2024-09-24 at 02 30 27

### Modifications

Change the feature's introduced by #23226 default values to 1ms initial / 10ms max.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -1231,14 +1231,14 @@ public class ServiceConfiguration implements PulsarConfiguration {
             doc = "On Shared and KeyShared subscriptions, if all available messages in the subscription are filtered "
                     + "out and not dispatched to any consumer, message dispatching will be rescheduled with a backoff "
                     + "delay. This parameter sets the initial backoff delay in milliseconds.")
-    private int dispatcherRetryBackoffInitialTimeInMs = 100;
+    private int dispatcherRetryBackoffInitialTimeInMs = 0;
 
     @FieldContext(
             category = CATEGORY_POLICIES,
             doc = "On Shared and KeyShared subscriptions, if all available messages in the subscription are filtered "
                     + "out and not dispatched to any consumer, message dispatching will be rescheduled with a backoff "
                     + "delay. This parameter sets the maximum backoff delay in milliseconds.")
-    private int dispatcherRetryBackoffMaxTimeInMs = 1000;
+    private int dispatcherRetryBackoffMaxTimeInMs = 0;
 
     @FieldContext(
             dynamic = true,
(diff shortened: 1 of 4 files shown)


pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -1231,14 +1231,14 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
             doc = "On Shared and KeyShared subscriptions, if all available messages in the subscription are filtered "
                     + "out and not dispatched to any consumer, message dispatching will be rescheduled with a backoff "
                     + "delay. This parameter sets the initial backoff delay in milliseconds.")
-    private int dispatcherRetryBackoffInitialTimeInMs = 100;
+    private int dispatcherRetryBackoffInitialTimeInMs = 0;
```
lhotari, line 1234:
```suggestion
    private int dispatcherRetryBackoffInitialTimeInMs = 10;
```

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -1231,14 +1231,14 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
 
     @FieldContext(
             category = CATEGORY_POLICIES,
             doc = "On Shared and KeyShared subscriptions, if all available messages in the subscription are filtered "
                     + "out and not dispatched to any consumer, message dispatching will be rescheduled with a backoff "
                     + "delay. This parameter sets the maximum backoff delay in milliseconds.")
-    private int dispatcherRetryBackoffMaxTimeInMs = 1000;
+    private int dispatcherRetryBackoffMaxTimeInMs = 0;
```
lhotari, line 1241:
```suggestion
    private int dispatcherRetryBackoffMaxTimeInMs = 100;
```

lhotari (changes requested):
The previous solution was burning CPU in a tight loop. Instead of disabling this, the values can be made 100x smaller.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari: We shouldn't disable the limit. Setting it to a very low value like 10 ms or even 1 ms is better for saving CPU and preventing a spinning loop that burns a lot of CPU.

lhotari:
One additional detail here is that another merged PR contains a small change to eliminate the extra latency when Key_Shared is reading ahead when keys are blocked: https://github.com/apache/pulsar/pull/23231/files#diff-c48d5c94842ac8c9a0c9314b207298069f38c8dcfeda4a9886fb3bb1f77843f2R369-R378
(This PR should also be backported so that we'd have a fix for #23200 in branch-3.0 and branch-3.3.)

codelipenghui (approved):

PR #23343, opened 2024-09-24
[fix][broker] Avoid introducing delay when there are delayed messages or marker messages

lhotari:
### Motivation

PR #23226 introduced a change to introduce a delay before calling `readMoreEntries` in PersistentDispatcherMultipleConsumers classes when no messages were dispatched.
This logic didn't take into account the case when all messages are delayed messages, marker messages or messages from an aborted transaction. In those cases additional delays would be introduced, causing performance regressions.

### Modifications

- add a failing test case that simulates dispatching a delayed message that gets filtered
  - no delay should be introduced before the next call to readMoreEntries
- Fix the regression by tracking how many entries were "processed" by consumers in the last round instead of tracking how many entries were dispatched.
  -  `lastNumberOfEntriesProcessed` tracks how many entries were processed by consumers in the last `trySendMessagesToConsumers` call
    - the number includes also delayed messages, marker messages, aborted txn messages and filtered messages
    - When no messages were processed, the value is 0. This is also an indication that the dispatcher didn't make progress in the last `trySendMessagesToConsumers` call. Only in those cases should the retry happen after a delay with increasing backoff.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


codelipenghui (approved):

dao-jun (approved):
lgtm

PR #23347, opened 2024-09-24
[fix][cli] Fix Pulsar-Client CLI to print metadata of message including encryption metadata

rdhabalia:
### Motivation

Currently, client-cli doesn't print encryption related metadata which requires in my cases by users to troubleshoot encryption/decryption related issues and validate key related metadata.

### Modifications

Add `--print-metadata` param in client-cli for consumer to allow to print additional metadata of the message.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/CmdRead.java
+++ b/pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/CmdRead.java
@@ -105,6 +105,9 @@ public class CmdRead extends AbstractCmdConsume {
     @Option(names = { "-ca", "--crypto-failure-action" }, description = "Crypto Failure Action")
     private ConsumerCryptoFailureAction cryptoFailureAction = ConsumerCryptoFailureAction.FAIL;
 
+    @Option(names = { "-mp", "--print-metadata" }, description = "Message metadata")
+    private boolean printMetadata = false;
+
     public CmdRead() {
         // Do nothing
         super();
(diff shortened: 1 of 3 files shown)


pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/CmdRead.java
```diff
@@ -101,6 +102,9 @@ public class CmdRead extends AbstractCmdConsume {
     @Option(names = { "-pm", "--pool-messages" }, description = "Use the pooled message", arity = "1")
     private boolean poolMessages = true;
 
+    @Option(names = { "-mp", "--print-metadata" }, description = "Message metadata")
```
lhotari, line 105: `-mp` isn't very intuitive. perhaps there doesn't need to be a short option and `--print-metadata` would be sufficient? (no strong opinion on this)

lhotari (commented):
LGTM

lhotari (approved):

crossoverJie (approved):

pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/CmdRead.java
rdhabalia: long option is mostly useful for documentation but the short option is useful when any user would like to use such tool on-demand and it can be used without dealing with typos. also, for consistency with other params that have a combination of short and long options, maybe it's better to keep both options for this new param as well.

PR #23349, opened 2024-09-25
[fix][broker] Fix unloadNamespaceBundlesGracefully can be stuck with extensible load manager 

BewareMyPower:
### Motivation

I observed an issue that broker was stuck at close for a long time. It's stuck at `BrokerService#unloadNamespaceBundlesGracefully`, which calls `disableBroker` once and `unloadNamespaceBundleAsync` for all owned namespace bundles **synchronously**. Most issues happen when the broker is the last broker.

#### Issue 1: Free events won't be sent in `overrideOwnership`

In `overrideOwnership`, if no broker is available, a `Free` event will be created.

https://github.com/apache/pulsar/blob/4ce0c752cc4b2d6dccb818ab0ffa854e82e42b85/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java#L1363

However, since the `dstBroker` and `sourceBroker` fields are null in the `Free` event, an exception will be thrown so that the `Free` event won't be created and sent.

https://github.com/apache/pulsar/blob/4ce0c752cc4b2d6dccb818ab0ffa854e82e42b85/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateData.java#L37

#### Issue 2: Free events could be skipped due to the same version id

The `Free` event is created in `overrideOwnership` based on a previous event on the same bundle from the table view. However, there might be inflight events that are not in the table view yet. In `ServiceUnitStateDataConflictResolver#shouldKeepLeft`, if the version id is the same, the `Free` event will be skipped. Then, if the last event is the `Owned` event whose target broker is the current broker in close, `waitForCleanups` will wait until the timeout exceeds.

https://github.com/apache/pulsar/blob/4ce0c752cc4b2d6dccb818ab0ffa854e82e42b85/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java#L1394-L1396

#### Issue 3: `__change_events` topic prevents `waitForCleanups` from finishing

The `__change_events` topic's reader, which is managed by the system topic based topic policies service, will try acquire the ownership. So that in `waitForCleanups`, there will always be a `Owned` event for this topic's bundle. If the target broker is the broker itself, `waitForCleanups` will never have a chance to exit until the timeout exceeds.

#### Issue 4: unloadNamespaceBundleAsync will be stuck at getOwnershipAsync if there is no available broker

Broker unregisters itself in `disableBroker`, if it's the last broker, then no brokers will be available after that. However, `unloadNamespaceBundleAsync` needs to publish a `Unloaded` event in `unloadAsync`.

https://github.com/apache/pulsar/blob/4ce0c752cc4b2d6dccb818ab0ffa854e82e42b85/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java#L700-L701

### Modifications

- issue 1: change the constructor of `ServiceUnitStateData` to all null `dstBroker` and `sourceBroker` for `Free` events.
- issue 2: call `TableView#refreshAsync` to refresh the entry set in `ServiceUnitStateTableViewImpl#flush` and call it  before `overrideOwnership`.
- issue 3: close the topic policies service at the beginning of `PulsarService#closeAsync`.
- issue 4: close the extensible load manager if there is no available broker after `unregister`

In addition, in `disableBroker`, cancel the load data report tasks and shutdown the `LoadDataStore` objects to avoid being affected by the producers and readers on these two non-persistent topics.

Since `LoadDataStore#get` is still used in `LeastResourceUsageWithWeight#select`, don't throw an exception and return an empty in `get`. And handle the case that `select` might throw an exception in `ExtensibleLoadManagerImpl#selectAsync`.

To handle the specific case when the broker is the last broker, close the broker in advance if there is no available broker in the metadata store. Then any namespace bundle's unload will also be skipped because the state was restored to INIT.

Add `testLookup` to cover the changes above.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -1038,7 +1038,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     private void closeInternalTopics() {
         List<CompletableFuture<Void>> futures = new ArrayList<>();
-        for (String name : Set.of(BROKER_LOAD_DATA_STORE_TOPIC, TOP_BUNDLES_LOAD_DATA_STORE_TOPIC)) {
+        for (String name : INTERNAL_TOPICS) {
             pulsar.getBrokerService()
                     .getTopicReference(name)
                     .ifPresent(topic -> futures.add(topic.close(true)
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -181,8 +181,14 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     private SplitManager splitManager;
 
-    volatile boolean started = false;
-    boolean disabling = false;
+    enum State {
+        INIT,
+        RUNNING,
+        // It's removing visibility of the current broker from other brokers. In this state, it cannot play as a leader
+        // or follower.
+        DISABLED,
+    }
+    private final AtomicReference<State> state = new AtomicReference<>(State.INIT);
 
     private boolean configuredSystemTopics = false;
 
@@ -215,7 +221,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
     }
 
     public Set<NamespaceBundle> getOwnedServiceUnits() {
-        if (!started) {
+        if (!running()) {
             log.warn("Failed to get owned service units, load manager is not started.");
             return Collections.emptySet();
         }
@@ -345,7 +351,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     @Override
     public void start() throws PulsarServerException {
-        if (this.started) {
+        if (state.get() != State.INIT) {
             return;
         }
         try {
@@ -444,7 +450,9 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
                     this.splitScheduler.start();
                     this.initWaiter.complete(true);
-                    this.started = true;
+                    if (!state.compareAndSet(State.INIT, State.RUNNING)) {
+                        failForUnexpectedState("start");
+                    }
                     log.info("Started load manager.");
                 } catch (Throwable e) {
                     failStarting(e);
@@ -756,7 +764,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     @Override
     public void close() throws PulsarServerException {
-        if (!this.started) {
+        if (state.get() == State.INIT) {
             return;
         }
         try {
@@ -792,7 +800,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                     } catch (Exception e) {
                         throw new PulsarServerException(e);
                     } finally {
-                        this.started = false;
+                        state.set(State.INIT);
                     }
                 }
 
@@ -809,21 +817,21 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     @VisibleForTesting
     synchronized void playLeader() {
-        if (disabling) {
-            return;
-        }
         log.info("This broker:{} is setting the role from {} to {}",
                 pulsar.getBrokerId(), role, Leader);
         int retry = 0;
         boolean becameFollower = false;
         while (!Thread.currentThread().isInterrupted()) {
             try {
-                if (!initWaiter.get()) {
+                if (!initWaiter.get() || disabled()) {
                     return;
                 }
                 if (!serviceUnitStateChannel.isChannelOwner()) {
                     becameFollower = true;
                     break;
+                }
+                if (disabled()) {
+                    return;
                 }
                 // Confirm the system topics have been created or create them if they do not exist.
                 // If the leader has changed, the new leader need to reset
@@ -839,8 +847,9 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                 }
                 break;
             } catch (Throwable e) {
-                if (disabling) {
-                    log.warn("The broker:{} failed to playLeader, exit because it's disabled", pulsar.getBrokerId());
+                if (disabled()) {
+                    log.warn("The broker:{} failed to set the role but exit because it's disabled",
+                            pulsar.getBrokerId(), e);
                     return;
                 }
                 log.warn("The broker:{} failed to set the role. Retrying {} th ...",
@@ -854,7 +863,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                 }
             }
         }
-        if (disabling) {
+        if (disabled()) {
             return;
         }
 
@@ -874,21 +883,21 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
 
     @VisibleForTesting
     synchronized void playFollower() {
-        if (disabling) {
-            return;
-        }
         log.info("This broker:{} is setting the role from {} to {}",
                 pulsar.getBrokerId(), role, Follower);
         int retry = 0;
         boolean becameLeader = false;
         while (!Thread.currentThread().isInterrupted()) {
             try {
-                if (!initWaiter.get()) {
+                if (!initWaiter.get() || disabled()) {
                     return;
                 }
                 if (serviceUnitStateChannel.isChannelOwner()) {
                     becameLeader = true;
                     break;
+                }
+                if (disabled()) {
+                    return;
                 }
                 unloadScheduler.close();
                 serviceUnitStateChannel.cancelOwnershipMonitor();
@@ -899,8 +908,9 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                 serviceUnitStateTableViewSyncer.close();
                 break;
             } catch (Throwable e) {
-                if (disabling) {
-                    log.warn("The broker:{} failed to playFollower, exit because it's disabled", pulsar.getBrokerId());
+                if (disabled()) {
+                    log.warn("The broker:{} failed to set the role but exit because it's disabled",
+                            pulsar.getBrokerId(), e);
                     return;
                 }
                 log.warn("The broker:{} failed to set the role. Retrying {} th ...",
@@ -914,7 +924,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                 }
             }
         }
-        if (disabling) {
+        if (disabled()) {
             return;
         }
 
@@ -1005,8 +1015,8 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
     public void disableBroker() throws Exception {
         // TopicDoesNotExistException might be thrown and it's not recoverable. Enable this flag to exit playFollower()
         // or playLeader() quickly.
-        synchronized (this) {
-            disabling = true;
+        if (!state.compareAndSet(State.RUNNING, State.DISABLED)) {
+            failForUnexpectedState("disableBroker");
         }
         serviceUnitStateChannel.cleanOwnerships();
         leaderElectionService.close();
@@ -1048,5 +1058,17 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
     @VisibleForTesting
     protected ServiceUnitStateChannel createServiceUnitStateChannel(PulsarService pulsar) {
         return new ServiceUnitStateChannelImpl(pulsar);
+    }
+
+    private void failForUnexpectedState(String msg) {
+        throw new IllegalStateException("Failed to " + msg + ", state: " + state.get());
+    }
+
+    boolean running() {
+        return state.get() == State.RUNNING;
+    }
+
+    private boolean disabled() {
+        return state.get() == State.DISABLED;
     }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -221,7 +221,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
     }
 
     public Set<NamespaceBundle> getOwnedServiceUnits() {
-        if (!running()) {
+        if (state.get() == State.INIT) {
             log.warn("Failed to get owned service units, load manager is not started.");
             return Collections.emptySet();
         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -624,7 +624,11 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
                                 filter.filterAsync(availableBrokerCandidates, bundle, context);
                         futures.add(future);
                     }
-                    return FutureUtil.waitForAll(futures).thenApply(__ -> {
+                    return FutureUtil.waitForAll(futures).exceptionally(e -> {
+                        // TODO: We may need to revisit this error case.
+                        log.error("Failed to filter out brokers when select bundle: {}", bundle, e);
+                        return null;
+                    }).thenApply(__ -> {
                         if (availableBrokerCandidates.isEmpty()) {
                             return Optional.empty();
                         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
@@ -808,7 +808,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
         try {
             topBundlesLoadDataStore.shutdown();
         } catch (IOException e) {
-            log.warn("Failed to shutdown brokerLoadDataStore", e);
+            log.warn("Failed to shutdown topBundlesLoadDataStore", e);
         }
     }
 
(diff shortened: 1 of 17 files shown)


lhotari (approved):
LGTM

heesung-sn:
Thank you for taking a loot at this issue.


For the last broker shutdown issue,

I think we have two options:

1. set all ownerships to Free and shutdown

This can be a little tricky because the clients could trigger lookups and new assignments might happen during shutdown. We can introduce a state to fail assignment during shutdown.

2. skip the gracefulShutdown and just shutdown loadManager.

Since there is no broker available to transfer ownerships, we could just simply shutdown the last broker -- since the load balancer is shutdown, no new assignment will happen during shutdown.

When the fist broker(leader) starts, it will fix any orphan ones. 

Seems like this pr is moving to the option 1.

Regarding the skip message issue,
I think the current skip logic can return lookups too soon, and I dont see a good reason to keep it. For example, when there are concurrent Assign events, it could return deferred lookups too soon by the skip msg logic, before Own event. I think it can just wait for the final Own event.

heesung-sn:
Thank you for finding bugs in the ServiceUnitStateData constructor that does not consider the Free state case and that waitForCleanup should handle Free case properly too.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -182,6 +182,7 @@ public class ExtensibleLoadManagerImpl implements ExtensibleLoadManager, BrokerS
     private SplitManager splitManager;
 
     volatile boolean started = false;
```
heesung-sn, line 184: Can we introduce an enum state?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -141,6 +141,7 @@ public class ServiceUnitStateChannelImpl implements ServiceUnitStateChannel {
     private volatile long lastOwnEventHandledAt = 0;
     private long lastOwnedServiceUnitCountAt = 0;
     private int totalOwnedServiceUnitCnt = 0;
+    private volatile boolean disablePubOwnedEvent = false;
```
heesung-sn, line 144: Can we add this case in the ChannelState enum below? I think a better name could be "CleaningOwnerships"

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -307,7 +309,7 @@ public synchronized void start() throws PulsarServerException {
                             pulsar.getConfiguration().getDefaultNumberOfNamespaceBundles());
 
             tableview = createServiceUnitStateTableView();
-            tableview.start(pulsar, this::handleEvent, this::handleExisting);
+            tableview.start(pulsar, this::handleEvent, this::handleExisting, this::handleSkippedEvent);
```
heesung-sn, line 312: I am actually not sure if this skip logic is required as this can return deferred lookups too soon.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -774,7 +774,11 @@ brokerId, getLogEventTag(data), serviceUnit,
         }
     }
 
-    private void handleSkippedEvent(String serviceUnit) {
+    private void handleSkippedEvent(String serviceUnit, ServiceUnitStateData skippedData) {
+        if (skippedData.state() == Free) {
```
heesung-sn, line 778: This is risky as the skipped message impacts the current states.

The channel design principle is that only non-skipped messages should impact the states in this channel.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -1381,8 +1400,10 @@ private void overrideOwnership(String serviceUnit, ServiceUnitStateData orphanDa
 
     private void waitForCleanups(String broker, boolean excludeSystemTopics, int maxWaitTimeInMillis) {
         long started = System.currentTimeMillis();
```
heesung-sn, line 1402: I think we can skip this waiting logic if the broker to cleanup is the last broker. 

I think this waiting logic is optional, and the intention is to give some time to the cluster to converge cleaned ownership view, if ever. 

However, because of the nature of the current ownership changes, there might be a chance that the ownership view might not converge within the timeout -- eventually the leader should fix any orphan or too long inflight states, though).

If there is no broker waiting, we can just skip this waiting.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateData.java
```diff
@@ -34,7 +34,7 @@ public record ServiceUnitStateData(
 
     public ServiceUnitStateData {
         Objects.requireNonNull(state);
-        if (StringUtils.isBlank(dstBroker) && StringUtils.isBlank(sourceBroker)) {
```
heesung-sn, line 37: The intention is that we want to tag a source or destination broker for all events to trace where the event comes from.

For example, the logic tries to set "source broker" when setting Free state, to indicate who was the previous owner. In what cases, both source and dst brokers can be empty here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -815,6 +819,15 @@ private void handleOwnEvent(String serviceUnit, ServiceUnitStateData data) {
 
     private void handleAssignEvent(String serviceUnit, ServiceUnitStateData data) {
         if (isTargetBroker(data.dstBroker())) {
+            if (disablePubOwnedEvent) {
+                log.info("Skip assigning self({}) as the owner after cleanOwnerships", serviceUnit);
+                final var getOwnerRequest = getOwnerRequests.remove(serviceUnit);
```
heesung-sn, line 824: This will make other brokers to wait for the Owned event. I think whatever ongoing state change should follow the protocol without interruption.

I think better place to check this "disablePubOwnedEvent" state in before publishing the state change command.

- publishAssignEventAsync
- publishOverrideEventAsync
- publishUnloadEventAsync
- publishSplitEventAsync

We already have `validateChannelState` func to check the channel state, and I think we can additionally consider "disablePubOwnedEvent" state there.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: Actually I found there is a `Disabled` state introduced in https://github.com/apache/pulsar/pull/23301. But it's never set other than tests.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: Let me revert this change.

This skip logic is not required. I restored this logic just because I want to handle the skipped Free events in the same method. But from your explanation [here](https://github.com/apache/pulsar/pull/23349#discussion_r1775957883), handling the skipped Free events might be dangerous so I'm going to think of another solution

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: >  if the broker to cleanup is the last broker.

I found sometimes the cleanup did not fail for the last broker but for the 2nd broker in `testCloseAfterLoadingBundles`.

```
2024-09-26T15:35:56,001 - INFO  - [main:BrokerService] - Disable broker in load manager completed in 0.081 seconds
2024-09-26T15:36:31,287 - INFO  - [main:BrokerService] - Disable broker in load manager completed in 35.099 seconds
2024-09-26T15:36:31,644 - INFO  - [main:BrokerService] - Disable broker in load manager completed in 0.214 seconds
2024-09-26T15:36:31,783 - INFO  - [main:ExtensibleLoadManagerCloseTest] - Brokers close time: [275, 35242, 354]
```

Let me look it deeper

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateData.java
BewareMyPower: https://github.com/apache/pulsar/blob/b1c5d96ab480ba1b8ad8cbe2077cbe4c467dfc61/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java#L1364

If the state is not `Owned`, there is still a case that `orphanData.sourceBroker()` is null. Here is the log I debugged

```
ServiceUnitStateData[state=Assigning, dstBroker=localhost:50930, sourceBroker=null,
```

BewareMyPower:
I reverted the changes to `Free` events and disabled all events except the override events (whose `force` field is true) during `cleanOwnerships()`, PTAL again @heesung-sn

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: I used this state in the latest code now

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: The cause when the 2nd broker failed to close is the issue 5 I mentioned in the latest PR description.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: I used the `Disabled` state instead and disabled all events except the override events (whose force field is true)

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -791,14 +782,36 @@ public void close() throws PulsarServerException {
...
+        }
+        try {
+            brokerLoadDataStore.shutdown();
+        } catch (IOException e) {
+            log.warn("Failed to shutdown brokerLoadDataStore", e);
+        }
+        try {
+            topBundlesLoadDataStore.shutdown();
+        } catch (IOException e) {
+            log.warn("Failed to shutdown brokerLoadDataStore", e);
```
heesung-sn, line 811: topBundlesLoadDataStore

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
```diff
@@ -982,17 +1017,28 @@ protected void monitor() {
...
+        }
         // Close the internal topics (if owned any) after giving up the possible leader role,
         // so that the subsequent lookups could hit the next leader.
         closeInternalTopics();
     }
 
     private void closeInternalTopics() {
         List<CompletableFuture<Void>> futures = new ArrayList<>();
-        for (String name : INTERNAL_TOPICS) {
+        for (String name : Set.of(BROKER_LOAD_DATA_STORE_TOPIC, TOP_BUNDLES_LOAD_DATA_STORE_TOPIC)) {
```
heesung-sn, line 1041: Could you explain why this change is required?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -1391,7 +1402,9 @@ private void waitForCleanups(String broker, boolean excludeSystemTopics, int max
                     continue;
                 }
 
-                if (data.state() == Owned && broker.equals(data.dstBroker())) {
+                if (data.state() == Free) {
+                    futures.put(serviceUnit, handleFreeEvent(serviceUnit, data));
```
heesung-sn, line 1406: why do we need to additionally handleFreeEvent here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -679,11 +678,15 @@ private void handleEvent(String serviceUnit, ServiceUnitStateData data) {
                     brokerId, serviceUnit, data, totalHandledRequests);
         }
 
-        if (channelState == Disabled) {
+        ServiceUnitState state = state(data);
+        if (channelState == Disabled && (data == null || !data.force())) {
```
heesung-sn, line 682: why do we need `(data == null || !data.force()` logic here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/store/TableViewLoadDataStoreImpl.java
```diff
@@ -92,7 +92,11 @@ public synchronized CompletableFuture<Void> removeAsync(String key) {
     public synchronized Optional<T> get(String key) {
         String msg = validateTableView();
         if (StringUtils.isNotBlank(msg)) {
-            throw new IllegalStateException(msg);
+            if (msg.equals(SHUTDOWN_ERR_MSG)) {
+                return Optional.empty();
```
heesung-sn, line 96: can you explain why we need to return empty here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
BewareMyPower: See the issue 5 that I added in the PR description.

If the `loadbalancer-service-unit-state` topic's owner happens to be the current broker, the producers could be disconnected. Then in `unloadNamespaceBundleAsync` it would fail to send unload related events.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: It might not be needed here now. Let me double confirm it

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/ExtensibleLoadManagerImpl.java
BewareMyPower: It's not required. I reverted this change first.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: Reverted

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/store/TableViewLoadDataStoreImpl.java
BewareMyPower: Without this change, `testCloseAfterLoadingBundles` would always fail.

```
2024-09-26T22:29:35,236 - INFO  - [main:ServiceUnitStateChannelImpl] - Started ownership cleanup for the inactive broker:localhost:58804
2024-09-26T22:29:35,244 - ERROR - [main:ServiceUnitStateChannelImpl] - Failed to override inactiveBroker:localhost:58804 ownership serviceUnit:public/default/0x50000000_0x60000000 orphanData:ServiceUnitStateData[state=Owned, dstBroker=localhost:58804, sourceBroker=localhost:58783, splitServiceUnitToDestBroker=null, force=true, timestamp=1727360975026, versionId=3]. totalCleanupErrorCnt:1
java.util.concurrent.ExecutionException: java.lang.IllegalStateException: This load store tableview has been shutdown
...
```

heesung-sn (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
```diff
@@ -364,8 +364,8 @@ private void readAllExistingMessages(Reader<T> reader, CompletableFuture<Void> f
                                   }
                                }).exceptionally(ex -> {
                                    if (ex.getCause() instanceof PulsarClientException.AlreadyClosedException) {
-                                       log.error("Reader {} was closed while reading existing messages.",
-                                               reader.getTopic(), ex);
+                                       log.info("Reader {} was closed while reading existing messages.",
```
Demogorgon314, line 367: Why change the log level from error to info?

Demogorgon314 (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
BewareMyPower: To optimize the error logs because the `AlreadyClosedException` is an expected exception but the full stack will be printed each time

BewareMyPower:
It seems there are some failed tests in `ExtensibleLoadManagerImplTest`, please don't merge and let me take a look

BewareMyPower:
image

`ExtensibleLoadManagerImplTest` passed locally now. Let's wait if it would fail.

PR #23353, opened 2024-09-26
[fix] [broker] Fix incorrect stat sub.msgBacklogNoDelayed if enabled delayedDeliveryFixedDelayDetectionLookahead

poorbarcode:
### Motivation

**Background**
- When we notice the `sub.msgBacklogNoDelayed` is a large number, we will increase the consumer count to consume quickly.
- After enabling `delayedDeliveryFixedDelayDetectionLookahead` and the messages encountered the feature, `sub.msgDelayed` is always `50k` and `sub.msgBacklogNoDelayed` keeps increasing, but the consumers will receive nothing.

### Modifications

- Correct the value of `sub.msgBacklogNoDelayed` if enabled `delayedDeliveryFixedDelayDetectionLookahead`
- Write a test.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -1335,7 +1335,7 @@ public class PersistentDispatcherMultipleConsumers extends AbstractDispatcherMul
     }
 
     @Override
-    public boolean isAllMessagesAreFixedDelayed() {
+    public boolean isAllWaitingReadMessagesAreFixedDelayed() {
         return shouldPauseDeliveryForDelayTracker();
     }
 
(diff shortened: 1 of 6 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -1334,6 +1334,11 @@ protected synchronized boolean shouldPauseDeliveryForDelayTracker() {
         return delayedDeliveryTracker.isPresent() && delayedDeliveryTracker.get().shouldPauseAllDeliveries();
     }
 
+    @Override
+    public boolean isAllWaitingReadMessagesAreFixedDelayed() {
+        return shouldPauseDeliveryForDelayTracker();
+    }
```
Technoboy-, line 1340: `pendingReadMessageAreFixedDelayed`?

PR #23359, opened 2024-09-26
[fix][broker] Fix the broker registery cannot recover from the metadata node deletion

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/23298 introduce a regression that once the metadata node of this broker was deleted (e.g. by session timeout), the broker register would never have a chance to recover. In this case, the clients whose owner is this broker would never be able to produce or consume.

### Modifications

Add a default listener to `BrokerRegisterImpl` that will register itself again in an asynchronous way if the deleted node is the current broker. Add a new state `Unregistering` to prevent the broker from registering itself again after `unregister()` is called.

Add `BrokerRegistryIntegrationTest` to verify this fix and the behavior introduced from https://github.com/apache/pulsar/pull/23298

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -1234,6 +1234,9 @@ protected void handleBrokerRegistrationEvent(String broker, NotificationType typ
             handleBrokerCreationEvent(broker);
         } else if (type == NotificationType.Deleted) {
             log.info("BrokerRegistry detected the broker:{} registry has been deleted.", broker);
+            if (brokerRegistry.getBrokerId().equals(broker)) {
+                brokerRegistry.registerAsync();
```
heesung-sn, line 1238: Nit: It is counter-intuitive that this broker is still alive but receives deletion events(maybe network issue?).

Can you add a comment for this case?

heesung-sn (approved):

heesung-sn:
I wonder if we need to periodically check this broker registration lock in the monitor thread as well for better fault tolerance.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
BewareMyPower: Added.

BewareMyPower:
> periodically check this broker registration lock in the monitor thread as well for better fault tolerance.

Yeah it can handle the case when there is a bug with the metadata store client. But even if it has a bug, we can easily find it from the alerts and fix it by manually restarting the broker or creating the metadata node. Therefore, I neither support nor oppose this idea.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateChannelImpl.java
```diff
@@ -1234,6 +1234,11 @@ protected void handleBrokerRegistrationEvent(String broker, NotificationType typ
             handleBrokerCreationEvent(broker);
         } else if (type == NotificationType.Deleted) {
             log.info("BrokerRegistry detected the broker:{} registry has been deleted.", broker);
+            // The registered node is an ephemeral node that could be deleted when the metadata store client's session
+            // is expired. In this case, we should register again.
+            if (brokerRegistry.getBrokerId().equals(broker)) {
+                brokerRegistry.registerAsync();
```
Demogorgon314, line 1240: Should we move this logic to `BrokerRegistry`?

BewareMyPower: It makes sense. Moved now.

pulsar-broker/src/test/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryIntegrationTest.java
```diff
@@ -0,0 +1,124 @@
...
+        Awaitility.await().atMost(Duration.ofSeconds(3)).untilAsserted(() -> Assert.assertEquals(
+                brokerRegistry.getAvailableBrokersAsync().join(), List.of(pulsar.getBrokerId())));
+        pulsar.getLocalMetadataStore().delete(brokerMetadataPath, Optional.empty());
+        Awaitility.await().atMost(Duration.ofSeconds(3)).untilAsserted(() -> Assert.assertEquals(
+                brokerRegistry.getAvailableBrokersAsync().join(), List.of(pulsar.getBrokerId())));
+
+        // If the node is deleted by unregister(), it should not recreate the path
+        brokerRegistry.unregister();
+        Thread.sleep(3000);
+        Assert.assertTrue(brokerRegistry.getAvailableBrokersAsync().get().isEmpty());
```
Demogorgon314, line 84:
```suggestion
        Awaitility.await().atMost(Duration.ofSeconds(5)).untilAsserted(() ->
                Assert.assertTrue(brokerRegistry.getAvailableBrokersAsync().get().isEmpty()));
```

Demogorgon314 (approved):

pulsar-broker/src/test/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryIntegrationTest.java
BewareMyPower: This suggested change is wrong. Here it's verifying that the node won't be created after 3 seconds, not that the node is deleted in 3 seconds.

Demogorgon314: Oh, I see.

lhotari:
This PR introduced a flaky test #23365, @BewareMyPower do you have a chance to fix it?

lhotari:
@BewareMyPower Is the current solution sufficient for register for the deletion event?

In LockManagerImpl, there's handling for session events:
https://github.com/lhotari/pulsar/blob/83b86abcb74595d7e8aa31b238a7dbb19a04dde2/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/LockManagerImpl.java#L118-L141

Before #23984 & #23988 changes, Zookeeper would be always creating persistent nodes even when ephemeral node creation was requested, in the case of using the API that is used in this case. That's why the session disconnection might not have been a visible issue before.

lhotari:
I created #24057 which is a similar issue.

Issue #23348, opened 2024-09-25
[Bug] Zookeeper OutOfMemoryError After Upgrading to Pulsar 3.3.0 with Zookeeper 3.9.2


jamesvsshark:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

Pulsar version: 3.3.0
Zookeeper version: 3.9.2
Kubernetes environment: Helm chart deployment
Zookeeper resource configuration:
Request/limit: 6GB memory, 2 CPU
Heap settings: -Xms5632m -Xmx5632m
GC settings:

```
-XX:+UseG1GC
-XX:MaxGCPauseMillis=10
-XX:+ParallelRefProcEnabled
-XX:+UnlockExperimentalVMOptions
-XX:+DoEscapeAnalysis
-XX:+DisableExplicitGC
-XX:+ExitOnOutOfMemoryError
-XX:+PerfDisableSharedMem
```

### Minimal reproduce step

1. Upgrade Pulsar to version 3.3.0 and Zookeeper to 3.9.2.
2. Deploy the Zookeeper quorum using default GC and memory settings from the Helm chart.
3. Observe memory consumption and monitor for crashes after a few days of running.

### What did you expect to see?

Zookeeper should run without constantly increasing memory usage or exhausting resources.

### What did you see instead?

Error observed:

`java.lang.OutOfMemoryError: unable to create a native thread: possibly out of memory or process/resource limits reached.
`

### Anything else?

Additional Context:

After downgrading Zookeeper to version 3.2.2, the OOM issue stopped, and no pod restarts occurred.
Autorecovery pods running 3.3.0 are also encountering Java heap memory issues.
Reviewing the Pulsar 3.3.0 release notes and PIP-324 ([#22054](https://github.com/apache/pulsar/pull/22054)), I suspect changes to the Alpine base image could be affecting thread creation and memory management.

Possible Solution:

It may be necessary to modify the Dockerfile to increase the stack size by setting the PTHREAD_STACK_MIN environment variable:

`ENV PTHREAD_STACK_MIN 2097152`

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

lhotari:
[...]

Issue #23306, opened 2024-09-13
[Bug] Alpine-based images quit with fatal error on aarch64


ozangunalp:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

Official Pulsar images with 3.3.0 and 3.3.1

### Minimal reproduce step

Running alpine-based container images on aarch64 machine.
We could reproduce it on RHEL 8 and raspberrypi but not not M1.

### What did you expect to see?

Pulsar server continue to run

### What did you see instead?

Here is the complete log of the container:

Here is the log: [pulsar.txt](https://github.com/user-attachments/files/16946424/pulsar.txt)

Last lines of log before fatal error : 

```
2024-09-10T14:02:58,193+0000 [pulsar-io-18-4] INFO  org.apache.pulsar.broker.service.ServerCnx - [[id: 0xa1215d54, L:/127.0.0.1:6650 - R:/127.0.0.1:34536] [SR:127.0.0.1, state:Connected]] Subscribing on topic persistent://public/default/__change_events / reader-936c229a0f. consumerId: 0
2024-09-10T14:02:58,269+0000 [pulsar-io-18-4] INFO  org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl - Opening managed ledger public/default/persistent/__change_events
2024-09-10T14:02:58,271+0000 [bookkeeper-ml-scheduler-OrderedScheduler-2-0] INFO  org.apache.bookkeeper.mledger.impl.MetaStoreImpl - Creating '/managed-ledgers/public/default/persistent/__change_events'
2024-09-10T14:02:58,340+0000 [bookkeeper-ml-scheduler-OrderedScheduler-2-0] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [192.168.144.2:46605] for ledger: 1
2024-09-10T14:02:58,344+0000 [BookKeeperClientWorker-OrderedExecutor-18-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl - [public/default/persistent/__change_events] Created ledger 1 after closed null
[...]

PR #23366, opened 2024-09-28
[fix] EXPERIMENT: Switch docker base image to avoid mixing musl & glibc libraries at runtime

lhotari:
### Motivation

While testing Pulsar 4.0.0-preview.1, some compatibility issues with the Alpine base image have shown up.
It appears that the solution in the base image is something that will cause further compatibility issues. Mixing musl and glibc at runtime is known to cause compatibility and stability issues.

https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/24647#note_176723 states it clearly:
_"Combining glibc and musl runtimes is basically all but guaranteed to create an unstable environment, unless the system is appropriately configured (glibc side uses glibc binaries only, and vice versa)."_

https://github.com/sgerrand/alpine-pkg-glibc/issues/194#issuecomment-1510140972 mentions:
_"You really need everything pure musl, or everything musl (but works with gcompat or libc6-compat shim-like tools...) or everything glibc using this package)"_ 

In our case, we are mixing pure musl libraries and glibc compiled libraries with our Alpine base image configuration which should be avoided.

Another reason to switch to a base image that is optimized for running Java JVM applications. The current Alpine image crashes the JVM in unexpected ways. #23348/#23361 is an example of this.

```
java.lang.OutOfMemoryError: unable to create a native thread: possibly out of memory or process/resource limits reached.
```

There's also one more release blocker which could be due to the mixed musl & glibc environment:  #23306

### Why mixing musl & glibc at runtime should be avoided

* blog post by Ariadne Conill, [there is no such thing as a "glibc based alpine image"](https://ariadne.space/2021/08/26/there-is-no-such-thing-as-a-glibc-based-alpine-image/)
* comments explaining reasons:
  * https://github.com/docker-library/official-images/pull/10779#issuecomment-906095981
  * https://github.com/docker-library/official-images/pull/10779#issuecomment-906240756
  * https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/24647#note_176723

### Modifications

Use https://hub.docker.com/r/bellsoft/liberica-runtime-container as the base image. ([Apache NiFi uses a bellsoft/liberica-openjdk-debian](https://github.com/apache/nifi/blob/master/nifi-docker/dockerhub/Dockerfile#L20), so Liberica is already used in ASF.)
It's the Liberica OpenJDK's official base image provided by BellSoft. It uses Alpaquita Linux as the OS, which is similar to Alpine and uses `apk` as the package manager. Alpaquita Linux is optimized for running JVM applications.

The image size is very similar as before.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


docker/pulsar/Dockerfile
```diff
@@ -116,31 +93,16 @@ prometheus_client\n\
...
-
-# Install GLibc compatibility library
-COPY --from=glibc /root/packages /root/packages
-RUN apk add --allow-untrusted --force-overwrite /root/packages/glibc-*.apk
-
-COPY --from=jvm /opt/jvm /opt/jvm
-ENV JAVA_HOME=/opt/jvm
-
-COPY --from=snappy-java /tmp/libsnappyjava.so /usr/lib/libsnappyjava.so
+" > /tmp/requirements.txt && RUN pip3 install --break-system-packages --no-cache-dir --only-binary grpcio -r /tmp/requirements.txt && rm /tmp/requirements.txt
```
nodece, line 96:
```suggestion
" > /tmp/requirements.txt && pip3 install --break-system-packages --no-cache-dir --only-binary grpcio -r /tmp/requirements.txt && rm /tmp/requirements.txt
```

docker/pulsar/Dockerfile
```diff
@@ -116,31 +93,16 @@ prometheus_client\n\
...
-RUN rm /requirements.txt
-
-# Install GLibc compatibility library
-COPY --from=glibc /root/packages /root/packages
-RUN apk add --allow-untrusted --force-overwrite /root/packages/glibc-*.apk
-
-COPY --from=jvm /opt/jvm /opt/jvm
-ENV JAVA_HOME=/opt/jvm
-
-COPY --from=snappy-java /tmp/libsnappyjava.so /usr/lib/libsnappyjava.so
```
nodece, line 131: Why not remove the `libsnappyjava`? Is built-in snappy work fine?

lhotari: Yes

lhotari: The reason why snappy-java had to be built separately was due to the musl & glibc mixing issue.

lhotari:
It looks like test image building fails due to missing supervisor package. It is available also with pip install: http://supervisord.org/installing.html

lhotari:
> It looks like test image building fails due to missing supervisor package. It is available also with pip install: http://supervisord.org/installing.html

supervisor has been installed, but there is some remaining issue.

PR #23369, opened 2024-09-29
[fix][test] Fix ReplicatorRateLimiterTest

Demogorgon314:
### Motivation

The `ReplicatorRateLimiterTest` will always fail with `StackOverflowError `, because in https://github.com/apache/pulsar/pull/23322 the `getRateLimiter` always recursive call to itself.

```
ReplicatorRateLimiterTest.testReplicatorRateLimiterMessageNotReceivedAllMessages
java.lang.StackOverflowError
at org.apache.pulsar.broker.service.ReplicatorRateLimiterTest.getRateLimiter(ReplicatorRateLimiterTest.java:614)
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorRateLimiterTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorRateLimiterTest.java
@@ -611,10 +611,7 @@ public class ReplicatorRateLimiterTest extends ReplicatorTestBase {
     }
 
     private static Optional<DispatchRateLimiter> getRateLimiter(PersistentTopic topic) {
-        if (topic.getReplicators().isEmpty()) {
-            return Optional.empty();
-        }
-        return topic.getReplicators().values().iterator().next().getRateLimiter();
+        return topic.getReplicators().values().stream().findFirst().map(Replicator::getRateLimiter).orElseThrow();
     }
 
     private static final Logger log = LoggerFactory.getLogger(ReplicatorRateLimiterTest.class);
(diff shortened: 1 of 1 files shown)


BewareMyPower:
It seems that this failed test was not detected by CI. At least it was not exposed in the latest PR:

image

We'd better ensure this test failure can be detected by CI, even if it's not required.

@lhotari Do you have any idea?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorRateLimiterTest.java
```diff
@@ -611,7 +611,10 @@ public void testReplicatorRateLimiterByBytes() throws Exception {
     }
 
     private static Optional<DispatchRateLimiter> getRateLimiter(PersistentTopic topic) {
-        return getRateLimiter(topic);
+        if (topic.getReplicators().isEmpty()) {
+            return Optional.empty();
+        }
+        return topic.getReplicators().values().iterator().next().getRateLimiter();
```
BewareMyPower, line 617:
```java
return topic.getReplicators().values().stream().findFirst().map(Replicator::getRateLimiter).orElseThrow();
```

A one-line fix

BewareMyPower (approved):

lhotari (approved):

lhotari:
> It seems that this failed test was not detected by CI.

@BewareMyPower the test is in the quarantine group. The reporting is very unnoticeable at the moment. It's another solution for flaky tests. Tests get run but don't block the build. 
The reason for quarantine is that the tests are so flaky that they fail almost on every run. There should be a issue about fixing it. The tests should get fixed, but there aren't many who care about addressing the problem.

BewareMyPower:
> the test is in the quarantine group. The reporting is very unnoticeable at the moment. 

Where do these tests run? Could you point out a workflow for example?

PR #23370, opened 2024-09-29
[fix][cli] Remove deprecated "-client" JVM arg

lhotari:
### Motivation

`-client` and `-server` JVM args have been deprecated in Java for a very long time. Pulsar CLI tools currently add `-client` argument. This causes some OpenJDK distributions to fail:
```
❯ docker run --rm -it bellsoft/liberica-runtime-container:jdk-17-cds-stream-glibc java -client
Unrecognized option: -client
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

### Modifications

- remove the unnecessary `-client` argument.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/conf/pulsar_tools_env.sh
+++ b/conf/pulsar_tools_env.sh
@@ -57,9 +57,6 @@ if [ -n "$PULSAR_MEM" ]; then
 fi
 PULSAR_MEM=${PULSAR_MEM:-"-Xmx128m -XX:MaxDirectMemorySize=128m"}
 
-# Garbage collection options
-PULSAR_GC=${PULSAR_GC:-" -client "}
-
 # Extra options to be passed to the jvm
 PULSAR_EXTRA_OPTS="${PULSAR_MEM} ${PULSAR_GC} ${PULSAR_GC_LOG} -Dio.netty.leakDetectionLevel=disabled ${PULSAR_EXTRA_OPTS}"
 


lhotari:
Claude AI response to question "what is the difference in starting java with "java -client" or "java -server" in Java 17 ?":

> # Difference between "java -client" and "java -server" in Java 17
> 
> In Java 17, there is no functional difference between starting Java with `java -client` or `java -server`. These flags are essentially deprecated and ignored. Here's a breakdown of the historical context and current situation:
> 
> ## Historical Context
> 
> - In earlier Java versions (pre-Java 8):
>   - `-client`: Used a JIT compiler optimized for faster startup and lower memory usage
>   - `-server`: Used a JIT compiler optimized for better long-term performance
> 
> ## Changes in Java 8 and Later
> 
> - JVM automatically chooses the most appropriate compiler mode
> - Introduction of Tiered Compilation, combining benefits of both client and server compilers
> 
> ## Java 17 Behavior
> 
> - `-client` and `-server` flags are ignored
> - Tiered Compilation is used by default, balancing startup time and long-term performance
> 
> ## Current Best Practices
> 
> - Avoid using `-client` or `-server` flags in Java 17
> - Let the JVM optimize performance automatically
> - For performance tuning, explore other JVM flags and options specific to Java 17
> 
> In conclusion, for Java 17, starting with `java -client` or `java -server` produces identical results, as both flags are treated the same way by the JVM and are effectively deprecated.

BewareMyPower (approved):

nodece (approved):
> In Oracle Java Runtime Environment (JRE) 8 and earlier, different implementations of the JVM, (the client VM, server VM, and minimal VM) were supported for configurations commonly used as clients, as servers, and for embedded systems. Because most systems can now take advantage of the server VM, only that VM implementation is provided in later versions.

https://docs.oracle.com/en/java/javase/21/vm/java-virtual-machine-technology-overview.html

Technoboy- (approved):

Issue #23365, opened 2024-09-28
Flaky-test: BrokerRegistryIntegrationTest.testRecoverFromNodeDeletion


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/apache/pulsar/actions/runs/11074101345/job/30796634094?pr=23362#step:11:1686

### Exception stacktrace

```
  Error:  org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryIntegrationTest.testRecoverFromNodeDeletion  Time elapsed: 5.037 s  <<< FAILURE!
  org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryIntegrationTest lists don't have the same size expected [1] but found [0] within 3 seconds.
  	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
  	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
  	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
  	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:985)
  	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:769)
  	at org.apache.pulsar.broker.loadbalance.extensions.BrokerRegistryIntegrationTest.testRecoverFromNodeDeletion(BrokerRegistryIntegrationTest.java:78)
  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
  	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
  	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
  	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
  	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
[...]

PR #23371, opened 2024-09-29
[fix][broker] Fix the broker registering might be blocked for long time

BewareMyPower:
Fixes https://github.com/apache/pulsar/issues/23365

### Motivation

https://github.com/apache/pulsar/pull/23359 would register the broker again if the registered metadata node was somehow deleted. However, the listeners in `BrokerRegistryImpl` are triggered in the load manager thread (via `Pulsar#Service#loadManagerExecutor`).

https://github.com/apache/pulsar/blob/5e832a1cc1441eaf8d64fe72c1a2af8829030d3d/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java#L229-L233

However, this thread executes many blocking tasks, including `monitor()`, `playLeader()` and `playFollower()` of `ExtensibleLoadManagerImpl`. If there were no available brokers, the blocking tasks might never complete until timeout exceeds. Here is an example:

```
2024-09-29T15:24:15,675 - WARN  - [pulsar-io-24-2:ConnectionHandler] - [non-persistent://pulsar/system/loadbalancer-broker-load-data] [reader-40952cbf05] Could not get connection to broker: org.apache.pulsar.client.api.PulsarClientException$BrokerMetadataException: {"errorMsg":"No available broker found.","reqId":175112655758658573, "remote":"localhost/127.0.0.1:56266", "local":"/127.0.0.1:56281"} -- Will try again in 3.146 s
2024-09-29T15:24:17,686 - WARN  - [pulsar-load-manager-22-1:ExtensibleLoadManagerImpl] - The broker:localhost:56270 failed to set the role. Retrying 1 th ...
org.apache.pulsar.broker.loadbalance.extensions.store.LoadDataStoreException: java.util.concurrent.TimeoutException
	at org.apache.pulsar.broker.loadbalance.extensions.store.TableViewLoadDataStoreImpl.startTableView(TableViewLoadDataStoreImpl.java:179) ~[classes/:?]
	at org.apache.pulsar.broker.loadbalance.extensions.store.TableViewLoadDataStoreImpl.start(TableViewLoadDataStoreImpl.java:157) ~[classes/:?]
	at org.apache.pulsar.broker.loadbalance.extensions.store.TableViewLoadDataStoreImpl.init(TableViewLoadDataStoreImpl.java:141) ~[classes/:?]
	at org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImpl.playLeader(ExtensibleLoadManagerImpl.java:849) ~[classes/:?]
```

In this case, `registerAsync` would have to wait until these methods are done, while they could have been recovered by a `registerAsync` call.

While the direct cause of failed `BrokerRegistryIntegrationTest` is that `TableViewImpl#flush` would not complete if there is no message in the topic. See https://github.com/apache/pulsar/issues/23365#issuecomment-2381090760 

### Modifications

Two fixes:
1. Fix the `TableViewImpl#flush` by filtering  the empty topics via checking if the last message id's entry id is non-negative
2. Call `registerAsync` directly in the metadata store callback thread instead of the load manager thread. It's an asynchronous metadata store operation so it won't block the thread.

Tests enhancements:
- Enable `BrokerRegistryIntegrationTest` again to verify fix 1 works
- Check the broker close time to verify fix 2 works
- Add `BrokerRegistryMetadataStoreIntegrationTest` to verify fix 2 works

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
@@ -36,6 +36,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.client.api.CryptoKeyReader;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessageId;
+import org.apache.pulsar.client.api.MessageIdAdv;
 import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.client.api.Reader;
 import org.apache.pulsar.client.api.ReaderBuilder;
@@ -260,6 +261,10 @@ public class TableViewImpl<T> implements TableView<T> {
     public CompletableFuture<Void> refreshAsync() {
         CompletableFuture<Void> completableFuture = new CompletableFuture<>();
         reader.thenCompose(reader -> getLastMessageIds(reader).thenAccept(lastMessageIds -> {
+            if (lastMessageIds.isEmpty()) {
+                completableFuture.complete(null);
+                return;
+            }
             // After get the response of lastMessageIds, put the future and result into `refreshMap`
             // and then filter out partitions that has been read to the lastMessageID.
             pendingRefreshRequests.put(completableFuture, lastMessageIds);
@@ -292,6 +297,10 @@ public class TableViewImpl<T> implements TableView<T> {
 
         CompletableFuture<Void> future = new CompletableFuture<>();
         getLastMessageIds(reader).thenAccept(maxMessageIds -> {
+            if (maxMessageIds.isEmpty()) {
+                future.complete(null);
+                return;
+            }
             readAllExistingMessages(reader, future, startTime, messagesRead, maxMessageIds);
         }).exceptionally(ex -> {
             future.completeExceptionally(ex);
@@ -304,7 +313,9 @@ public class TableViewImpl<T> implements TableView<T> {
         return reader.getLastMessageIdsAsync().thenApply(lastMessageIds -> {
             Map<String, TopicMessageId> maxMessageIds = new ConcurrentHashMap<>();
             lastMessageIds.forEach(topicMessageId -> {
-                maxMessageIds.put(topicMessageId.getOwnerTopic(), topicMessageId);
+                if (((MessageIdAdv) topicMessageId).getEntryId() >= 0) {
+                    maxMessageIds.put(topicMessageId.getOwnerTopic(), topicMessageId);
+                }
             });
             return maxMessageIds;
         });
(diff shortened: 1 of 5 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
```diff
@@ -292,6 +297,10 @@ private CompletableFuture<Void> readAllExistingMessages(Reader<T> reader) {
 
         CompletableFuture<Void> future = new CompletableFuture<>();
         getLastMessageIds(reader).thenAccept(maxMessageIds -> {
```
heesung-sn, line 299: nit: maxMessageIds -> lastMessageIds

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
```diff
@@ -304,7 +313,9 @@ private CompletableFuture<Map<String, TopicMessageId>> getLastMessageIds(Reader<
         return reader.getLastMessageIdsAsync().thenApply(lastMessageIds -> {
             Map<String, TopicMessageId> maxMessageIds = new ConcurrentHashMap<>();
             lastMessageIds.forEach(topicMessageId -> {
-                maxMessageIds.put(topicMessageId.getOwnerTopic(), topicMessageId);
+                if (((MessageIdAdv) topicMessageId).getEntryId() >= 0) {
```
heesung-sn, line 316: I think this is counter-intuitive that the the tableview needs to handle such invalid messageId in the downstream logic. Don't we need to add this logic in the upstream? For example, we could skip this message either in broker or consumer if the downstream shouldn't see such invalid messageIds?

BewareMyPower: > we could skip this message either in broker or consumer if the downstream shouldn't see such invalid messageIds?

This message id actually has the same meaning with `MessageId.Earliest`, which does not represent the position of a message. Instead, it means all messages' ids should be greater than this message id, i.e. `x.compareTo(y)` will return false if `x` is such a message id and `y` is an id of a message.

For example, given a consumer that subscribes 2 topics ("tp1" with 2 partitions and "tp2" with 1 partition), the `getLastMessageIds` method will return a list

```
[0] TopicMessageId("tp1", {ledgerId: -1, entryId: -1, partition: 0})
[1] TopicMessageId("tp1", {ledgerId: -1, entryId: -1, partition: 1})
[2] TopicMessageId("tp2", {ledgerId: -1, entryId: -1, partition: 0})
```

Then the caller will know these two topics are subscribed by the current consumer. If you're going to skip these two message ids, the client side won't be able to see the subscribed topics from the result of `getLastMessageIds`. This would be a breaking change that could affect the downstream applications.

IMO, it would be better to add a method `isEarliest()` or `isLatest()` to `MessageId` (but it's beyond the scope of this PR). Actually `getLastMessageIds` does not return `MessageId.earliest`, whose partition index is -1. So I have to just check if `entryid` is negative here. You can also see the similar check internally in `ConsumerImpl#hasMessageAvailableAsync`.

BewareMyPower: >  the tableview needs to handle such invalid messageId

1. When the message id represents "earliest", the future of this topic should be completed immediately.
2. Otherwise, the future of this topic will be completed after the reader reads the message whose message is greater than the message id.

It's intuitive because if the topic is empty, no messages will be read.

BewareMyPower: I changed the method name and added comments for it now

heesung-sn (approved):

Demogorgon314 (approved):

PR #23355, opened 2024-09-26
[improve][PIP] PIP-383: Support granting/revoking permissions for multiple topics

Technoboy-:
### Motivation

In AuthorizationProvider, the authorization interface `grantPermissionAsync(TopicName topicName, Set<AuthAction> actions, String role, String authDataJson)` currently only supports granting permissions to a single topic. 
If multiple topics need to be authorized under a namespace, the client needs to call the authorization interface concurrently. 
Since the permissions information are stored in the namespace-level policies, and multiple topics may be on different brokers, concurrent authorization will cause concurrent modification exceptions. 
Therefore, supporting granting/revoking permissions for multiple topics is very friendly.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23372, opened 2024-09-29
[improve][broker] PIP-383: Support granting/revoking permissions for multiple topics

Technoboy-:
### Motivation

PIP-383:  https://github.com/apache/pulsar/pull/23355

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
@@ -24,14 +24,18 @@ import com.google.common.collect.Sets;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Stream;
 import javax.ws.rs.core.Response;
 import org.apache.pulsar.broker.PulsarServerException;
 import org.apache.pulsar.broker.ServiceConfiguration;
 import org.apache.pulsar.broker.authentication.AuthenticationDataSource;
 import org.apache.pulsar.broker.resources.PulsarResources;
+import org.apache.pulsar.client.admin.GrantTopicPermissionOptions;
+import org.apache.pulsar.client.admin.RevokeTopicPermissionOptions;
 import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.policies.data.AuthAction;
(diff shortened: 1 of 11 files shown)


pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
```diff
@@ -251,6 +254,68 @@ public CompletableFuture<Void> grantPermissionAsync(TopicName topicName, Set<Aut
...
+        return getPoliciesReadOnlyAsync().thenCompose(readonly -> {
+            if (readonly) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Policies are read-only. Broker cannot do read-write operations");
+                }
+                throw new IllegalStateException("policies are in readonly mode");
+            }
+            TopicName topicName = TopicName.get(options.get(0).getTopic());
+            return pulsarResources.getNamespaceResources()
+                    .setPoliciesAsync(topicName.getNamespaceObject(), policies -> {
```
lhotari, line 267: This code seems to assume that all topics are in the same namespace. That seems like a problematic assumption. Is there a goal to support `GrantTopicPermissionOptions` that apply to topics that could be in different namespaces?

Although I can see that there's a check in the different layer. The validation should happen here again unless support for multiple namespaces is added. The reason for this is that the caller shouldn't need to know implementation details. If it needs to know the implementation details, it's a leaky or broken abstraction.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
```diff
@@ -251,6 +254,68 @@ public CompletableFuture<Void> grantPermissionAsync(TopicName topicName, Set<Aut
...
```
lhotari, line 296: This code seems to assume that all topics are in the same namespace. Similar comment as in the previous case.

lhotari (changes requested):

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
Technoboy-: yes, Lari, should be in the same namespace.
I have checked it in the Rest Part

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -613,6 +616,83 @@ protected CompletableFuture<Void> internalGrantPermissionOnNamespaceAsync(String
...
+                        throw new RestException(Status.CONFLICT, "Concurrent modification");
+                    } else {
+                        log.error("[{}] Failed to revoke permissions for namespace {}",
+                                clientAppId(), namespaceName, ex);
+                        throw new RestException(realCause);
+                    }
+                });
+    }
+
+    private CompletableFuture<Boolean> checkNamespace(Stream<String> namespaces) {
```
Technoboy-, line 686: I have checked it here

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
Technoboy-: https://github.com/apache/pulsar/pull/23372/files#r1794452230
@lhotari

Technoboy-: Thanks @lhotari , I have addressed your comment. add the validation here

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java
Technoboy-: addressed

lhotari (approved):
LGTM, good work @Technoboy-

PR #22054, opened 2024-02-14
[improve][pip] PIP-324: Alpine image

merlimat:
PIP:  [PIP-324](https://github.com/apache/pulsar/pull/21716)

### Modifications

Changed the base Docker image from Ubuntu to Alpine.

Example Docker image: `merlimat/pulsar:3.3.0-SNAPSHOT-f2a91a1`

https://hub.docker.com/layers/merlimat/pulsar/3.3.0-SNAPSHOT-f2a91a1/images/sha256-723cda334135f93b366bd51920795b911e5412a55b0f4436fb81762a747dd6ec?context=explore

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [x] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [X] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/merlimat/pulsar/pull/7

PR #23373, opened 2024-09-29
[improve][ci] Switch to Java 21 as default JVM version for CI

lhotari:
### Motivation

Java 21 has been switched in the Docker image in 3.3.0 with "PIP-324: Alpine image (#22054)":
https://github.com/apache/pulsar/blob/branch-3.3/docker/pulsar/Dockerfile#L61

Pulsar CI should also run on Java 21 by default.

### Modifications

- Use Java 21 by default in CI
- Run a scheduled build on Java 17 with docker image using Java 17
- Add `-XX:+ZGenerational` to `PULSAR_GC` parameters by default when running on Java 21+
- Cleanup remaining Ubuntu related properties from the Maven build.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/conf/bkenv.sh
+++ b/conf/bkenv.sh
@@ -37,9 +37,6 @@ BOOKIE_LOG_DIR=${BOOKIE_LOG_DIR:-"${PULSAR_LOG_DIR}"}
 # Memory size options
 BOOKIE_MEM=${BOOKIE_MEM:-${PULSAR_MEM:-"-Xms2g -Xmx2g -XX:MaxDirectMemorySize=2g"}}
 
-# Garbage collection options
-BOOKIE_GC=${BOOKIE_GC:-${PULSAR_GC:-"-XX:+UseZGC -XX:+PerfDisableSharedMem -XX:+AlwaysPreTouch"}}
-
 if [ -z "$JAVA_HOME" ]; then
   JAVA_BIN=java
 else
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -84,9 +84,7 @@ flexible messaging model and an intuitive client API.</description>
 
     <pulsar.client.python.version>3.4.0</pulsar.client.python.version>
 
-    <UBUNTU_MIRROR>http://archive.ubuntu.com/ubuntu/</UBUNTU_MIRROR>
```
heesung-sn, line 87: nit: why do we need to remove UBUNTU_MIRROR?

conf/bkenv.sh
```diff
@@ -60,6 +57,17 @@ for token in $("$JAVA_BIN" -version 2>&1 | grep 'version "'); do
     fi
 done
 
+# Garbage collection options
+BOOKIE_GC="${BOOKIE_GC:-${PULSAR_GC}}"
+if [ -z "$BOOKIE_GC" ]; then
+  BOOKIE_GC="-XX:+PerfDisableSharedMem -XX:+AlwaysPreTouch"
+  if [[ $JAVA_MAJOR_VERSION -ge 21 ]]; then
+    BOOKIE_GC="-XX:+UseZGC -XX:+ZGenerational ${BOOKIE_GC}"
```
heesung-sn, line 65: Do we know ZGenerational is a better option for Pulsar?

pom.xml
lhotari: That was removed in this PR as a cleanup. It was needed before 3.3.x with the Ubuntu  Image.

conf/bkenv.sh
lhotari: One reason why it's better since it doesn't have the double memory mapping issue. 

We don't have benchmark results yet, but I don't see a reason why we couldn't promote it as default. Most deployment methods override the defaults so it is very unlikely to cause regressions.

lhotari: https://netflixtechblog.com/bending-pause-times-to-your-will-with-generational-zgc-256629c9386b

heesung-sn: It will be interesting to check this benchmark, but I guess that could be a separate task.

heesung-sn (approved):

nodece (approved):

PR #23364, opened 2024-09-28
[fix][misc] Log Conscrypt security provider initialization warnings at debug level

lhotari:
### Motivation

- `pulsar-admin` output parsing breaks in 4.0.0-preview.1 testing because of the warnings

[Example from Pulsar Helm chart testing](https://github.com/apache/pulsar-helm-chart/actions/runs/11076453216/job/30779614305?pr=530#step:6:5223)
```
Creating subscription for output topic
2024-09-27T19:25:20,336+0000 [main] WARN  org.apache.pulsar.common.util.SecurityUtility - Conscrypt isn't available for Linux amd64. Using JDK default security provider.
Waiting for function to be ready
parse error: Invalid numeric literal at line 1, column 14
```

### Modifications

- switch logging level from warn to debug

### Additional context

- previous change in this area:  #20371

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23374, opened 2024-09-30
[fix] Remove gcompat from the Alpine based image since real glibc is installed

lhotari:
### Motivation

In 3.3.1 and 4.0.0-preview.1 images, Conscrypt cannot be loaded. Jetty uses it for native OpenSSL based TLS.
This caused a problem described in #23364 where the warning log message "2024-09-27T19:25:20,336+0000 [main] WARN  org.apache.pulsar.common.util.SecurityUtility - Conscrypt isn't available for Linux amd64. Using JDK default security provider." broke pulsar-admin output parsing in other scripts. 
The problem was addressed by setting log level to `debug`.

While investigating the issue, it appeared that musl + gcompat and glibc in Alpine shouldn't be mixed at runtime in dynamic linking. It's not recommended to mix musl and glibc either.

However, for the Conscrypt case, the loading of the library works when `gcompat` is removed. There shouldn't be a reason to have `gcompat` exist with the current real `glibc` solution in Alpine.

### Modifications

- remove `gcompat` package from Alpine.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docker/pulsar/Dockerfile
+++ b/docker/pulsar/Dockerfile
@@ -89,7 +89,6 @@ RUN apk add --no-cache \
             py3-pip \
             py3-grpcio \
             py3-yaml \
-            gcompat \
             ca-certificates \
             procps \
             curl \


lhotari:
IIRC, the reason to add real glibc support was due to the fact that the Pulsar IO Kinesis connector didn't work with the Alpine image. The Kinesis connector uses Amazon Kinesis Producer which contains an embedded native application that has been compiled for glibc.
https://github.com/awslabs/amazon-kinesis-producer/blob/3be1d3d6e77ffa80eb173fd7e06ac962db6ec6b1/java/amazon-kinesis-producer/src/main/java/com/amazonaws/services/kinesis/producer/Daemon.java#L429-L461

There's a script in the repository to compile it for other runtime environments: https://github.com/awslabs/amazon-kinesis-producer/blob/master/bootstrap.sh

For using amazon-kinesis-producer, the correct approach would be to have build a musl compatible binary or have a solution to run the process in an isolated glibc environment. It would be possible to build a minimal environment with the minimal set of libraries and then pass LD_LIBRARY_PATH environment variable to use it for running the `kinesis_producer` binary. The current library already supports overriding the binary and passing environment variables.

PR #23380, opened 2024-09-30
[improve][client]: add --streaming option to topics partitioned-stats-internal

dlg99:
### Motivation

Streaming the stats output directly to the standard output without parsing the response in memory.

### Modifications

added `--streaming` flag for  `topics stats-internal/partitioned-stats-internal` commands to stream the stats.

### Verifying this change

This change added unit tests.

**added `--streaming` flag for  `topics stats-internal/partitioned-stats-internal` commands**

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [x] **The admin CLI options**
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/dlg99/pulsar/pull/21

### DIFF
--- a/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
+++ b/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
@@ -18,6 +18,7 @@
  */
 package org.apache.pulsar.client.admin;
 
+import java.io.InputStream;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


lhotari (changes requested):
Instead of an `InputStream`, a `Reader` would be more appropriate for the interface.

The reason for this is in this case, the payload is most likely UTF-8 encoded which uses multi-byte characters for most non-ascii characters.
When reading in 2048 chunks, a multi-byte character might get split at the boundary which will corrupt the output.

Using a `Reader` would address this without any additional logic.
A long time ago I worked on such a fix for Gradle, https://issues.gradle.org/browse/GRADLE-3329. In that case there was custom buffering and fixing the issue wasn't trivial. I ended up using `java.nio.charset.CharsetDecoder`  and `java.nio.charset.CoderResult#isUnderflow` to detect when a byte is missing.

A `Reader` will buffer the underlying bytes when there's a multi-byte character case and the remaining byte is not yet readable. That's why it's better to rely on a `Reader` in the interface.

lhotari:
@dlg99 please address the review feedback, https://github.com/apache/pulsar/pull/23380#pullrequestreview-2339343482.

PR #23382, opened 2024-10-01
[fix][broker] timeout when broker registry hangs and monitor broker registry (ExtensibleLoadManagerImpl only)

heesung-sn:
### Motivation

- metadata store could fail the broker lock registration. In this case, we should retry and also monitor the registry periodically and try to register it if not registered yet.
- also we need to verify if the broker is healthy or not before starting the orphan bundle cleanups, as the broker might receive "out-dated" notifications.
### Modifications
- add monitor logic to periodically check if the broker is registered and to try to register it.
- add timeout logic when the broker re-registration hangs.
- add retry logic when the broker re-registration fails.
- add health check logic before starting the orphan bundle cleanup jobs and canceling the cleanup jobs.
- add brokerId in broker health admin api

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/heesung-sn/pulsar/pull/82

PR #23390, opened 2024-10-02
[improve][broker] Add error logs and retry if the broker failed to register itself for metadata node deletion

BewareMyPower:
### Motivation

In my chaos test environment, I found some `registerAsync` call failed. However, there is no error log for it.

### Modifications

Add the error logs for failure of `registerAsync` and retry in `handleMetadataStoreNotification`. For `BrokerRegistryImpl#start`, the error info is wrapped into the `PulsarServerException` propagated to the `PulsarService#start`'s catch block and it will fail fast so we don't need to log errors here.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


heesung-sn:
This change conflicts with this one, https://github.com/apache/pulsar/pull/23382. I also added a retry logic in this PR, and I think we need to run the listeners after registering the lock. Otherwise, there can be a racing condition between orphan bundle cleanup and the lock recreation.

BewareMyPower:
Let's close this PR for now

Issue #20622, opened 2023-06-20
Provide consistent concept around `partitioned topic` and `topic partition`/`internal topic`


JooHyukKim:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

(follow up of question raised from [slack community](https://apache-pulsar.slack.com/archives/C5Z4T36F7/p1686636952934049) and discussion with @lhotari)

# Motivation

Concepts should include whatever is commonly used. But currently, there is no good definition of what "partitioned topic" and "non-partitioned topic" mean and some places `topic partition` and `internal topic` interchangeably. And there are currently two locations where we can improve on

### 1# **Topic Name API**

Let’s take look at [one of questions raised in slack community](https://apache-pulsar.slack.com/archives/C5Z4T36F7/p1686636952934049)

```java
TopicName topicName = TopicName.get("persistent://public/default/example-partition-0");
System.out.println(topicName.isPartitioned()); // true

String partitionedTopicName = topicName.getPartitionedTopicName();
TopicName.get(partitionedTopicName).isPartitioned()); // false, so it's a `PartitionedTopicName` but isn't partitioned, why? I can't understand.
```

Even though we know those who are used to Pulsar can distinguish the difference, not those without Pulsar concept. The term `partition` is somewhat overloaded with concepts. 

Just to be clear,

- "**partitioned topic**" is `persistent://public/default/example` and the
- "**topic partition**" is `persistent://public/default/example-partition-0`

### **2# Documentation & Glossary**

At [terminology page of Pulsar-site](https://pulsar.apache.org/docs/3.0.x/reference-terminology/#partitioned-topic) says..

> **Partitioned Topic**
> A topic that is served by multiple brokers, which enables higher throughput.

…. which isn't consistent with the other parts of the docs and the explanations aren't sufficient. And the “other parts” refer to docs about ["topics"](https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#topics) and ["partitioned topics"](https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#partitioned-topics)

# Solution

### 1# Improve naming for **TopicName API**

- Deprecate `ToicName.isPartitioned()` and add new
    - `TopicName.isInternallyPartitioned()` or
    - `TopicName.hasInternalPartitions()` or
    - `TopicName.isPartitionedType()`
- Deprecate `TopicName.getPartitionedTopicName()`  and add new
[...]

PR #23407, opened 2024-10-07
[improve][functions]Adding boolean in contextImpl.java to check partitioned topic present or not.

nikam14:
### Motivation
In `contextImpl.java` the `getConsumer()` - When trying to get partitioned topic consumer it throws error consumer not found.
Instead show error message that partitioned topic not present.
Also reduces computation to search for consumer.

### Modifications
Added a boolean var to check partitioned topic present or not.
when try to get consumer check whether partitioned topic present then proceed.

### Verifying this change

This change is already covered by existing tests, such as `ContextImplTest.`

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/ContextImpl.java
+++ b/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/ContextImpl.java
@@ -763,7 +763,7 @@ class ContextImpl implements Context, SinkContext, SourceContext, AutoCloseable 
             throw new PulsarClientException("Getting consumer is not supported");
         }
 
-        if ((partition != 0) && (partionedTopicPresent == false)) {
+        if ((partition != 0) && (!partionedTopicPresent)) {
             throw new PulsarClientException("No Partioned topic present");
         }
 
--- a/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/ContextImpl.java
+++ b/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/ContextImpl.java
@@ -721,7 +721,8 @@ class ContextImpl implements Context, SinkContext, SourceContext, AutoCloseable 
                         topicConsumers.putIfAbsent(TopicName.get(consumer.getTopic()), consumer);
                         if (consumer.getTopic().contains("-partition-")) {
                             partionedTopicPresent = true;
-                        }});
+                        }
+                });
     }
 
     private void reloadConsumersFromMultiTopicsConsumers() {
@@ -738,7 +739,8 @@ class ContextImpl implements Context, SinkContext, SourceContext, AutoCloseable 
                          topicConsumers.putIfAbsent(TopicName.get(c.getTopic()), c);
                          if (c.getTopic().contains("-partition-")) {
                             partionedTopicPresent = true;
-                         }});
+                         }
+                });
     }
 
     // returns null if consumer not found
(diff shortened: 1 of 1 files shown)


pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/ContextImpl.java
```diff
@@ -716,7 +717,12 @@ public void setInputConsumers(List<Consumer<?>> inputConsumers) {
                         consumer instanceof MultiTopicsConsumerImpl
                                 ? ((MultiTopicsConsumerImpl<?>) consumer).getConsumers().stream()
                                 : Stream.of(consumer))
-                .forEach(consumer -> topicConsumers.putIfAbsent(TopicName.get(consumer.getTopic()), consumer));
+                .forEach(consumer -> {
+                        topicConsumers.putIfAbsent(TopicName.get(consumer.getTopic()), consumer);
+                        if (consumer.getTopic().contains("-partition-")) {
```
lhotari, line 722: Please use the `isPartitioned()` in `TopicName` class to determine whether the topic is partitioned.

lhotari (changes requested):
Thanks for the contribution @nikam14. Please check the review comments. In addition, there would have to be a test case to cover this. One possible location for the test case could be one of the classes in https://github.com/apache/pulsar/tree/master/pulsar-broker/src/test/java/org/apache/pulsar/io directory so that there would be a end-to-end test to cover the use case. In addition, a unit test could be added to https://github.com/apache/pulsar/blob/master/pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/ContextImplTest.java if it is feasible.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/ContextImpl.java
nikam14: Thanks for review

lhotari:
> In `contextImpl.java` the `getConsumer()` - When trying to get partitioned topic consumer it throws error consumer not found.
> Instead show error message that partitioned topic not present.
> Also reduces computation to search for consumer.

@nikam14 Could you please provide more context about the specific use case where you encountered this issue? It would be helpful to understand:

1. What were you trying to accomplish?
2. What did you expect to happen?
3. What actually happened instead?

This additional information will help us better understand the impact on end-users and ensure our solution addresses the root cause effectively.

lhotari:
> 1 - With these changes when try to get `partitonedTopicConsumer` it will first check whether partitionedTopicConsumer exist or 
>       not. If exist then return else throw error 'No Partitioned topic present' .
>       with these changes error message will be more specific.
> 2 - A more specific error message .
> 3 - If Consumer is not found in the Map then it returns consumer not found.

Please explain the context of the end user use case? As a pulsar end user, what is the use case?

lhotari:
> changes are covered in `testGetConsumer()-ContextImplTest`. In `testGetConsumer()` test function we can see if user is trying to get a consumer of partitioned topic it will get error message that " no partitioned topic present ", which means there is no consumer with partitioned topic.

Ok. What is the current error message without this change?

lhotari:
> changes are covered in `testGetConsumer()-ContextImplTest`. In `testGetConsumer()` test function we can see if user is trying to get a consumer of partitioned topic it will get error message that " no partitioned topic present ", which means there is no consumer with partitioned topic.

this is not a end user test case. It's possible that you are making incorrect assumptions about the behavior as well as the reviewers. In Pulsar, the partitioned topics concepts can easily mislead anyone. Please see https://github.com/apache/pulsar/issues/20622 for more details.

Issue #23164, opened 2024-08-13
Flaky-test: ManagedLedgerTest.testDeleteCurrentLedgerWhenItIsClosed


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failures

- [2024-08-07T22:14:33.5197682Z](https://github.com/apache/pulsar/actions/runs/10291342248/job/28486280269#step:11:90) 
- [2024-08-07T21:07:35.1097748Z](https://github.com/apache/pulsar/actions/runs/10291342248/job/28483793177#step:11:90) 

### Exception stacktrace

```
org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a org.apache.bookkeeper.mledger.impl.ManagedLedgerTest expected [1] but found [2] within 10 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:985)
	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:769)
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerTest.testDeleteCurrentLedgerWhenItIsClosed(ManagedLedgerTest.java:4368)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:11)
[...]

PR #23437, opened 2024-10-11
[fix][test] Fix flaky test ManagedLedgerTest.testDeleteCurrentLedgerWhenItIsClosed

summeriiii:
Fixes https://github.com/apache/pulsar/issues/23164

### Motivation
The previous current ledger can only be deleted if the new ledger is created.
When param closeLedgerByAddEntry is true, there may exist the situation: when call internalTrimEdgers to delete previous current ledger, new ledger has not been created, and causing the test failed.
 

                         
### Modifications
Whether closeLedgerByAddEntry is true or false, we both need to add `Awaitility.await().untilAsserted(() -> assertEquals(ml.ledgers.size(), 2));` to wait the new ledger create.
             
 
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


lhotari (approved):
LGTM

PR #23438, opened 2024-10-11
[fix][broker] normalize path

nodece:
### Motivation

In the broker configuration, we're using relative paths for parameters such as `brokerInterceptorsDirectory`, `protocolHandlerDirectory`, `additionalServletDirectory`, and `offloadersDirectory`. These paths include the `./` notation, and since the broker doesn't normalize them, the absolute paths generated are incorrect, for example: `/app/pulsar/20241011165422044/./interceptors/pulsar-xxx.nar`. As a result, the broker is unable to resolve the correct file locations.

Test cases: 
```java
    @Test
    public void testPathNormalize() {
        Path path = Paths.get("./interceptors").toAbsolutePath().normalize();
        assertEquals(path.toString().indexOf("./"), -1);
    }

    @Test
    public void testPathWithoutNormalize() {
        Path path = Paths.get("./interceptors").toAbsolutePath();
        assertNotEquals(path.toString().indexOf("./"), -1);
    }
```

### Modifications

-  Append the `java.nio.file.Path#normalize` call after the `java.nio.file.Path#toAbsolutePath` call to ensure proper path resolution.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/offload/OffloaderUtils.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/offload/OffloaderUtils.java
@@ -125,7 +125,7 @@ public class OffloaderUtils {
 
     public static Offloaders searchForOffloaders(String offloadersPath, String narExtractionDirectory)
             throws IOException {
-        Path path = Paths.get(offloadersPath).toAbsolutePath();
+        Path path = Paths.get(offloadersPath).toAbsolutePath().normalize();
         log.info("Searching for offloaders in {}", path);
 
         Offloaders offloaders = new Offloaders();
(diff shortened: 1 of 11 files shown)


lhotari:
> These paths include the `./` notation, and since the broker doesn't normalize them, the absolute paths generated are incorrect, for example: `/app/pulsar/20241011165422044/./interceptors/pulsar-xxx.nar`. As a result, the broker is unable to resolve the correct file locations.

I'm not against making this change. Just one question: Technically a path like /app/pulsar/20241011165422044/./interceptors/pulsar-xxx.nar is correct. What is the error message in this case?

lhotari (approved):
LGTM

nodece:
59778ea0-ce0c-47ee-aa1d-4a0efb710c53

I'm sure the `message-trace.nar` exists.

Changed the `brokerInterceptorsDirectory` value from `. /interceptors` to `interceptors`, it works fine.

lhotari:
>  nodece added this to the 4.0.0 milestone 3 minutes ago

@nodece Please don't modify the milestone. It will mess up tracking since the code freeze has already started. I can cherry-pick this to branch-4.0. Please read https://lists.apache.org/thread/xtnd51oy8ytmbtrctpmlpg2cqpjgf957

PR #23441, opened 2024-10-11
[improve][broker][PIP-379] Don't replace a consumer when there's a collision

### LINKED ISSUE #23439: [Bug][PIP-379] Hash ranges should not move from one consumer to another when a consumer joins
> What happens if a hash is moved between existing consumers? Is it handled?

Yes. All removed ranges are processed in both when a consumer is added and when it's removed. In some experimental testing I noticed that it actually could happen also when a consumer is added, due to how hash range conflicts are resolved. There's a need to make an improvement that when there are conflicts that the first assigned consumer is preserved instead of replacing it. I'll make a PR for that since hash range conflicts are fairly likely to happen with the reduced hash range of 65535 and 100 hash points when there's a large amount of consumers. I'll address that.

_Originally posted by @lhotari in https://github.com/apache/pulsar/pull/23352#discussion_r1797026999_

lhotari:
Fixes #23439

### Motivation

When a consumer joins or leaves, the hash ranges from existing consumers shouldn't move from one to another.

### Modifications

- Revisit adding the to hash ring so that existing consumers aren't replaced when there are collisions.
- Add a test case which ensures this

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ConsistentHashingStickyKeyConsumerSelector.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ConsistentHashingStickyKeyConsumerSelector.java
@@ -27,12 +27,14 @@ import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.client.api.Range;
 
 /**
  * This is a consumer selector using consistent hashing to evenly split
  * the number of keys assigned to each consumer.
  */
+@Slf4j
 public class ConsistentHashingStickyKeyConsumerSelector implements StickyKeyConsumerSelector {
     // use NUL character as field separator for hash key calculation
     private static final String KEY_SEPARATOR = "\0";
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ConsistentHashingStickyKeyConsumerSelector.java
```diff
@@ -76,18 +78,36 @@ public CompletableFuture<Optional<ImpactedConsumersResult>> addConsumer(Consumer
...
+                    hashPointsAdded++;
                 }
             }
+            if (hashPointsAdded == 0) {
+                log.error("Failed to add consumer '{}' to the hash ring. There were {} collisions. Consider increasing "
+                                + "the number of points ({}) per consumer by setting "
+                                + "subscriptionKeySharedConsistentHashingReplicaPoints={}",
+                        consumer, hashPointCollisions, numberOfPoints,
+                        Math.max((int) (numberOfPoints * 1.5d), numberOfPoints + 1));
+            }
```
equanz, line 106: It can be done outside of scope, but I think it would be better if this state could be sent to consumers.
(Consumers can't know there are no hash points, can they?)

lhotari: One problem is that the hash points don't get restored if an earlier consumer leaves. A solution where there's a list and the first entry is used would be a solution to this. That's one reason to restore the list in the hash ring entry. I'll create an issue to track it.

merlimat (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ConsistentHashingStickyKeyConsumerSelector.java
lhotari: @equanz @merlimat please review #23447, it addresses the remaining issue.

PR #23449, opened 2024-10-13
[fix][broker] Key_Shared subscription: Reject consumers with incompatible policy

### LINKED ISSUE #23272: [Bug] Dispatcher will be closing and reopening when Key_shared consumers have different policies (AUTO_SPLIT/STICKY, allowOutOfOrderDelivery true/false)
### Search before asking

### Read release policy

### Version

all released versions including master branch

### Minimal reproduce step

The conclusion is currently based on the source code:
https://github.com/apache/pulsar/blob/766d2a407196533832184447c25498c6a82f7a86/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L287-L309

When multiple consumers are using different policies, this should be properly handled.
One possibility is to keep the policy of the connected consumers and reject any other consumers and return a proper error message.

### What did you see instead?

Based on the source code, it looks like the solution cannot work:
- a new dispatcher is created while the existing dispatcher is active
- the existing dispatcher is closed asynchronously
- no proper errors messages are logged or logged that could help resolving the situation

### Anything else?

_No response_

pdolif:
Fixes #23272

### Motivation

When a consumer with a different Key_Shared policy connects, the current behavior is that the existing consumer will be closed and replaced with the consumer with the different policy. Since the replaced consumer gets disconnected, it will connect again and swap with the other consumer, and this replacement loop keeps going on.

Taken from #23272 by @lhotari:

When multiple consumers are using different policies, this should be properly handled.
One possibility is to keep the policy of the connected consumers and reject any other consumers and return a proper error message.

### Modifications

- Updated PersistentSubscription and NonPersistentSubscription behavior to reject new Key_Shared consumers (added to a Key_Shared subscription) when the new consumer's policy is incompatible with the dispatcher's policy. This is the case if the Key_Shared mode (AUTO_SPLIT/STICKY) or allowOutOfOrderDelivery (true/false) differs.

### Verifying this change

- Added unit test to PersistentSubscriptionTest that verifies that incompatible Key_Shared consumers are rejected by a persistent subscription.
- Added new unit test NonPersistentSubscriptionTest that verifies the same but for nonpersistent subscriptions.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/pdolif/pulsar/pull/4


lhotari (approved):
LGTM. Great work, @pdolif!

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -310,6 +310,19 @@ private CompletableFuture<Void> addConsumerInternal(Consumer consumer) {
...
+                        KeySharedMeta consumerKsm = consumer.getKeySharedMeta();
+                        if (dispatcherKsm.getKeySharedMode() != consumerKsm.getKeySharedMode()) {
+                            return FutureUtil.failedFuture(
+                                    new SubscriptionBusyException("Subscription is of different key_shared mode"));
+                        }
+                        if (dispatcherKsm.isAllowOutOfOrderDelivery() != consumerKsm.isAllowOutOfOrderDelivery()) {
+                            return FutureUtil.failedFuture(
+                                    new SubscriptionBusyException(dispatcherKsm.isAllowOutOfOrderDelivery()
+                                            ? "Subscription allows out of order delivery" :
+                                            "Subscription does not allow out of order delivery"));
```
lhotari, line 324: @pdolif These error messages are a bit vague. It's already an issue with the existing "Subscription is of different type" exception. It's currently hard to determine which mode is active on the broker side and which is the mode on the client. 
Something like this for all error messages could be helpful:
`"Subscription is of different type"` -> `String.format("Subscription is of different type. Active subscription type of '%s' is different than the connecting consumer's type '%s'", dispatcher.getType(), consumer.subType())` etc.
It might be useful to keep "Subscription is of different type" as the prefix for all errors since there might be client applications that are looking for that String since there's no unique exception for this problem.

pdolif: @lhotari I agree and updated the error messages. What do you think of the new ones?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/nonpersistent/NonPersistentSubscription.java
```diff
@@ -160,7 +160,29 @@ public synchronized CompletableFuture<Void> addConsumer(Consumer consumer) {
...
+                if (dispatcherKsm.isAllowOutOfOrderDelivery() != consumerKsm.isAllowOutOfOrderDelivery()) {
+                    return FutureUtil.failedFuture(new SubscriptionBusyException(
+                            String.format("Subscription is of different type. %s",
+                                    dispatcherKsm.isAllowOutOfOrderDelivery()
+                                            ? "Active subscription allows out of order delivery while the "
+                                            + "connecting consumer does not allow it." :
+                                            "Active subscription does not allow out of order delivery while "
+                                                    + "the connecting consumer allows it.")));
+                }
             }
```
lhotari, line 186: extract this logic into a method and push it to `AbstractSubscription` so that code duplication could be eliminated since there's similar logic in `PersistentSubscription`. I guess the return type could be `Optional<Future<?>>` so that the conditionality could be resolved?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
lhotari: @pdolif great work. I added a suggestion for reducing the duplication.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/nonpersistent/NonPersistentSubscription.java
pdolif: That's an excellent idea. I moved it to AbstractSubscription. The return type is `Optional<CompletableFuture<Void>>` now. What do you think, would it make sense to add additional tests for the extracted method to `AbstractSubscriptionTest`?

lhotari: > That's an excellent idea. I moved it to AbstractSubscription. The return type is `Optional<CompletableFuture<Void>>` now. What do you think, would it make sense to add additional tests for the extracted method to `AbstractSubscriptionTest`?

I think it's sufficient to test it with the actual behavior that is tested. Testing private or protected methods is a "code smell" hinting that a new concept or a "unit" should be extracted which has it's own API that is tested in a unit test. We don't follow this guideline in the Pulsar code base so there's a lot of bad examples around. We can improve over time.

lhotari (approved):
LGTM, Good work!

lhotari:
there's some unrelated test flakiness such as #23486 which caused the first attempt to fail.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SubscriptionTestBase.java
```diff
@@ -0,0 +1,84 @@
...
+
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import org.apache.pulsar.broker.BrokerTestUtil;
+import org.apache.pulsar.common.api.proto.CommandSubscribe;
+import org.apache.pulsar.common.api.proto.KeySharedMeta;
+import org.apache.pulsar.common.api.proto.KeySharedMode;
+import org.testng.annotations.DataProvider;
+
+public abstract class SubscriptionTestBase {
```
lhotari, line 29: This is perhaps a slightly confusing base class. It could be better to have a single class for the compatibility tests and test both persistent and non-persistent type in the same test class. This would be isolated from the existing tests.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/nonpersistent/NonPersistentSubscriptionTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.pulsar.common.api.proto.KeySharedMeta;
+import org.testng.annotations.Test;
+
+@Test(groups = "broker")
+public class NonPersistentSubscriptionTest extends SubscriptionTestBase {
+
+    @Test(dataProvider = "incompatibleKeySharedPolicies")
+    public void testIncompatibleKeySharedPoliciesNotAllowed(KeySharedMeta consumer1Ksm, KeySharedMeta consumer2Ksm,
+                                                            String expectedErrorMessage) throws Exception {
+        PulsarTestContext context = PulsarTestContext.builderForNonStartableContext().build();
```
lhotari, line 39: Using Lombok's Cleanup annotation is recommended for closing test resources.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SubscriptionTestBase.java
pdolif: I moved them to a separate test class. 
I thought of combining the two test methods (for persistent/non-persistent subscription) into one because they are very similar except for the subscription used. But in my opinion, adding the subscription to the data provider could decrease the test readability. 
Maybe it makes sense to extract a method from this (or parts of it):
```java
try {
    // add second consumer with incompatible key_shared policy
    sub.addConsumer(keySharedConsumerMock2).get(5, TimeUnit.SECONDS);
    fail(BrokerServiceException.SubscriptionBusyException.class.getSimpleName() + " not thrown");
} catch (Exception e) {
    // subscription throws exception when consumer with incompatible key_shared policy is added
    Throwable cause = e.getCause();
    assertTrue(cause instanceof BrokerServiceException.SubscriptionBusyException);
    assertEquals(cause.getMessage(), expectedErrorMessage);
}
```
What do you think?

lhotari: The test looks great now.

> Maybe it makes sense to extract a method from this (or parts of it):

I think it's fine for a test, especially in the Pulsar codebase. 
By the way, not directly related to our topic of discussion, but for new tests, it would be great to enable using JUnit 5 instead of requiring the usage of TestNG for all tests. In BookKeeper, [Junit 5 has been enabled (Junit 4 was used before instead of TestNG)](https://github.com/apache/bookkeeper/pulls?q=is%3Apr+is%3Aclosed+%22junit+5%22). It would also be great to prefer AssertJ for assertions since the error messages are very useful in AssertJ when a test fails. In Pulsar tests, AssertJ is already used in some tests.

lhotari (approved):
Great work @pdolif, thanks for contributing!

PR #23452, opened 2024-10-14
[fix][client] Fix the javadoc for ConsumerBuilder.isAckReceiptEnabled

### LINKED ISSUE #23249: [Bug] The javadoc of ConsumerBuilder.isAckReceiptEnabled is not consistent, it references ack timeout and redelivery which isn't related to ack receipt
### Search before asking

### Read release policy

### Version

any released version, master branch

### Minimal reproduce step

The javadoc of ConsumerBuilder.isAckReceiptEnabled is not consistent, it references ack timeout and redelivery which isn't related to ack receipt at all:
https://github.com/apache/pulsar/blob/10f4e0248f0f985b1dc7ad38970c906b7fe629be/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java#L200-L208

There should be a proper explanation of why ack receipt is useful and what considerations are needed when using it.

One reason to enable ack receipt feature is to ensure that acknowledgements reach the server. The current assumption is that there isn't any automatic acknowledgement retry logic, but this is to be confirmed when documenting the feature.
When using this feature, it's recommended to use acknowledgeAsync. Using acknowledge would cause performance issues for most applications since it introduces a round trip to the server and would prevent pipelining (having multiple messages in-flight). In some usecases that might be desirable.

### What did you see instead?

Misleading description

### Anything else?

_No response_

lhotari:
Fixes #23249

### Motivation

The current javadoc for ConsumerBuilder.isAckReceiptEnabled doesn't make sense and it's wrong, see #23249 for details.

### Modifications

Fix the javadoc.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java
@@ -201,12 +201,13 @@ public interface ConsumerBuilder<T> extends Cloneable {
      * Enables or disables the acknowledgment receipt feature.
      *
      * <p>When this feature is enabled, the consumer ensures that acknowledgments are processed by the broker by
-     * waiting for a receipt from the broker. It is recommended to use the asynchronous
-     * {@link Consumer#acknowledgeAsync(Message)} method for acknowledgment when this feature is enabled. This is
-     * because using the synchronous {@link Consumer#acknowledge(Message)} method with acknowledgment receipt can cause
-     * performance issues due to the round trip to the server, which prevents pipelining (having multiple
-     * messages in-flight). With the asynchronous method, the consumer can continue consuming other messages while
-     * waiting for the acknowledgment receipts.
+     * waiting for a receipt from the broker. Even when the broker returns a receipt, it doesn't guarantee that the
+     * message won't be redelivered later due to certain implementation details.
+     * It is recommended to use the asynchronous {@link Consumer#acknowledgeAsync(Message)} method for acknowledgment
+     * when this feature is enabled. This is because using the synchronous {@link Consumer#acknowledge(Message)} method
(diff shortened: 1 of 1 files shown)


pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java
```diff
@@ -198,17 +198,20 @@ public interface ConsumerBuilder<T> extends Cloneable {
     ConsumerBuilder<T> ackTimeout(long ackTimeout, TimeUnit timeUnit);
 
     /**
-     * Acknowledgement returns receipt, but the message is not re-sent after getting receipt.
+     * Enables or disables the acknowledgment receipt feature.
```
eolivelli, line 201: I think that it is worth to note that even if the server acknowledged the operation, this method doesn't guarantee that the message won't be redelivered.
Because on the broker side acknowledgement are persisted periodically or by other strategies (no need to explain here)

eolivelli (approved):

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java
lhotari: Thanks, I'll mention that.

lhotari: @eolivelli added "Even when the broker returns a receipt, it doesn't guarantee that the message won't be redelivered later due to certain implementation details."

Technoboy- (approved):

nodece (approved):

PR #19236, opened 2023-01-15
[improve][broker] Add `topic_load_failed` metric

tjiuming:
PIP: https://github.com/apache/pulsar/issues/18979

### Motivation

Add `topic_load_failed` metric

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:  https://github.com/tjiuming/pulsar/pull/20

PR #23351, opened 2024-09-25
[improve] [pip] PIP-382: Add a label named reason for topic_load_failed_total

poorbarcode:
### Motivation

- [DIscussion](https://lists.apache.org/thread/f3xhmm342jor042n5ykkxoc32ffcn85s)
- [Vote](https://lists.apache.org/thread/ng6z0dssjh1hgp91f590wkcl2ymhvn48)

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #23458, opened 2024-10-14
[fix][broker] Do not record as a topic load failure for the requested broker is not the owner

codelipenghui:
### Motivation

The topic_load_failed metric is introduced by https://github.com/apache/pulsar/pull/19236 which is used to monitor the topic load failures such as timeout, metadata service issue, etc.

But the topic owner check exception is an expected behavior if the requested broker is not the owner of the topic. We should avoid to record as a topic load failure from topic owner check. 

For the non-persistent topic, we don't need this change. 

https://github.com/apache/pulsar/blob/d759a1e616614ab661c3f858292dcada8f733e63/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1351-L1364

### Verifying this change

- New unit tests were added

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -1320,7 +1320,9 @@ public class BrokerService implements Closeable {
         topicFuture.exceptionally(t -> {
             // If the broker is not the owner of the topic, the client will redirect to another broker.
             // In this case, we should not record as a topic load failure.
-            if (!(FutureUtil.unwrapCompletionException(t) instanceof ServiceUnitNotReadyException)) {
+            Throwable cause = FutureUtil.unwrapCompletionException(t);
+            if (!(cause instanceof ServiceUnitNotReadyException)
+                    && !(cause.getMessage().contains("not served by this instance"))) {
                 getPulsarStats().recordTopicLoadFailed();
             }
             pulsar.getExecutor().execute(() -> topics.remove(topic, topicFuture));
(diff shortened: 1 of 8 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -1318,8 +1318,9 @@ public void deleteTopicAuthenticationWithRetry(String topic, CompletableFuture<V
     private CompletableFuture<Optional<Topic>> createNonPersistentTopic(String topic) {
         CompletableFuture<Optional<Topic>> topicFuture = new CompletableFuture<>();
         topicFuture.exceptionally(t -> {
+            // If the broker is not the owner of the topic, the client will redirect to another broker.
+            // In this case, we should not record as a topic load failure.
             pulsarStats.recordTopicLoadFailed();
-            pulsar.getExecutor().execute(() -> topics.remove(topic, topicFuture));
```
lhotari, line 1322: Removing this line doesn't seem to be related to topic load failed metrics. Is this intentional?
It seems that in the original solution this would ensure that failed topics don't added to the topics map.

codelipenghui:
There is a PIP to improve this case https://github.com/apache/pulsar/pull/23351, which is better than just handle the ServiceNotReady exception. I will close this PR.

PR #23488, opened 2024-10-21
[fix][broker] Don't allow creating a non-persistent partitioned topic with '-partition-' in name

hanmz:
### Motivation

Currently, we can create a non-persistent partition topic contain `-partition-.` However, we cannot use it as a normal non-persistent partition topic. For example, when we use the `partitioned-stats` command, we will encounter errors.

_Partitioned Topic Name should not contain '-partition-'
Reason: Partitioned Topic Name should not contain '-partition-'_

### Modifications

We should disable the creation of non-persistent partition topics contain `-partition-`.

### Verifying this change

All unit tests passed, not have any behavior change.

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
Thanks for the contribution, @hanmz. Would you be able to add a test case to cover this change?

lhotari (approved):

hanmz:
> Thanks for the contribution, @hanmz. Would you be able to add a test case to cover this change?


OK, I've added a  test case.

lhotari:
Closing and reopening to trigger a new CI run

Jason918:
reopen to trigger ci

PR #22854, opened 2024-06-05
[fix][client] Fix pattern consumer create crash if a part of partitions of a topic have been deleted

poorbarcode:
#### Background

**1.Regarding the API getting topics by regexp pattern, there are two implementations:***
- `List<String> HttpLookupService.getPartitionedTopicMetadata(...)`
- `CommandGetTopicsOfNamespace BinaryProtoLookupService.getPartitionedTopicMetadata(...)`

Pulsar transferred both response types `List<String>` and `CommandGetTopicsOfNamespace` to a `GetTopicsResult` object. And discarded the partition information when doing transference. For example:
- Get a list `topic-1-partition-0`, `topic-1-partition-1`.
- The transferring operation will group them to `topic-1`.

**2.The behavior of Patten consumers**
- Before https://github.com/apache/pulsar/pull/5230, Patten consumers will try to create the missing partitions when it starts up.
- After https://github.com/apache/pulsar/pull/5230, Patten consumers only subscribe to the partitions that exist to lead the topics can be deleted automatically. In other words, it sets the variable `createTopicIfDoesNotExist` of Multi Topics Consumer to `false`.

---

#### Issue
1. When users are starting a Pattern consumer, the consumer will try to subscribe to all the partitions even if some partitions have been deleted before, and then the client crashes due to a Topic Not Exists Exception. You can reproduce the issue by the test [testConsumerAfterOnePartDeleted](https://github.com/poorbarcode/pulsar/pull/91/files#diff-c6bb7cf6a6d6f5c4673ec5837351d981c835024fbbb864f32113c05b8ab30b4eR142)
2. The Patten consumer that is started removes all partitions when one partition was deleted, even if there are still half of the partitions exist.

---

### Modifications

- Multi Topics Consumer only subscribes to the existing partitions if the config `createTopicIfDoesNotExist` is `false`.
- The Patten consumer that is started only removes the partitions that were deleted.
- Fix race condition issues of updating subscriptions

### Addational things
[flink-connector-pulsar](https://github.com/apache/flink-connector-pulsar) also has the same issue

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #23489, opened 2024-10-21
[fix] [proxy] Fix pattern consumer does not work when using Proxy

poorbarcode:
### Motivation

https://github.com/apache/pulsar/pull/22854 added a new method `List<Integer> getExistsPartitions(String topic)`, which depends on the new method `GetTopicsResult.nonPartitionedOrPartitionTopics`, but the author of the PR forgot the change the deserializer of Pulsar-Proxy, which may cause the following issue:
- Broker respond: `tp-partition-0`
- Proxy received `tp-partition-0` and recreate a new response that may remove the suffix `-partition-0` to client 
- Client received `tp`, which without `-partition-0`
  - Client assumed the parition `0` does not exists. 

### Modifications

Fix the bug

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


lhotari (approved):
LGTM

Issue #23494, opened 2024-10-21
[Bug] Consumer receive acknowledged messages from compacted topic after reconnection 


summeriiii:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

master

### Minimal reproduce step

1. producer send message 
2. trigger compact
3. consume the compacted topic and acknowledge message
4.  unload the topic and reconnect
5.  re-consume the compacted topic
``` @Test
    public void testIndividualAcknowledgeWithReconnection() throws Exception {
        final String topicName = "persistent://my-property/use/my-ns/testIndividualAcknowledge" + UUID.randomUUID();
        final String subName = "sub1";
        final int numMessages = 10;

        pulsarClient.newConsumer().topic(topicName).subscriptionName(subName)
                .receiverQueueSize(1).readCompacted(true).subscribe().close();

        Map<String, String> expected = new HashMap<>();

        @Cleanup
        Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
                .topic(topicName)
                .enableBatching(false)
                .messageRoutingMode(MessageRoutingMode.SinglePartition)
                .create();

        for (int i = 0; i < numMessages; i++) {
            String key = "key" + new Random().nextInt(4);
            String value = ("my-message-" + i);
            producer.newMessage().key(key).value(value).send();
            expected.put(key, value);
        }
        producer.flush();

        // compact the topic
        compact(topicName);

        @Cleanup
        Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING).topic(topicName)
                .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
                .subscriptionName(subName).receiverQueueSize(1).readCompacted(true).subscribe();

        Map<String, String> results = new HashMap<>();
        while (true) {
            Message<String> message = consumer.receive(3, TimeUnit.SECONDS);
            if (message == null) {
                break;
            }
            results.put(message.getKey(), message.getValue());
            consumer.acknowledge(message);
        }
        Awaitility.await().untilAsserted(() ->
                assertEquals(admin.topics().getStats(topicName, true).getSubscriptions().get(subName).getMsgBacklog(),
                        numMessages - expected.size()));

        // unload the topic
        admin.topics().unload(topicName);

[...]

PR #21187, opened 2023-09-15
[fix][broker] Avoid consumers receiving acknowledged messages from compacted topic after reconnection

coderzc:
Fixes #21074

### Motivation

In #11287, we use MessageId.earliest instead of the earliest available message to read on the topic when we first read after reconnection, this will lead to consumers receiving acknowledged messages from the compacted topic after reconnection, so I think for a durable cursor (exclude __compaction cursor), we should read messages from the mark-delete position.

### Modifications

* If a durable cursor (exclude `__compaction` cursor) has the mark-delete position, don't read messages from the earliest.
* To avoid skipping the remaining data in the compacted ledger after rewind, we only skip invalid md-position when recovering cursor and add `rewind(boolean readCompacted)` method in ManagedCursor and just rewind to the next message of the mark delete position instead of resetting the read position to the next valid position of the original topic.
* This change will also make `subscriptionInitialPosition=Latest` work normally since the mark-delete doesn't is invalid position (entryID==-1L).

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23495, opened 2024-10-21
[fix][broker] Fix consumer receive individual acknowledged messages from compacted topic after reconnection

summeriiii:
Fixes https://github.com/apache/pulsar/issues/23494

### Motivation
After consumed the compacted topic and individual acknowledged messages, if the topic unload and consumer reconnect, we will receive the individual acknowledged messages again.
Like the none-compacted topic, I think that consumer should not receive  acknowledge message after reconnection.
                                      
### Modifications
 - add `isMessageIndividualDeleted` to check if this message has been individual acknowledged
 - In `CompactedTopicUtils#asyncReadCompactedEntries`, use isMessageIndividualDeleted to filter out and skip unnecessary entry
 
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java
@@ -893,6 +893,8 @@ public interface ManagedCursor {
 
     boolean isMessageDeleted(Position position);
 
+    boolean isMessageIndividualDeleted(Position position);
+
     ManagedCursor duplicateNonDurableCursor(String nonDurableCursorName) throws ManagedLedgerException;
 
     long[] getBatchPositionAckSet(Position position);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicUtils.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicUtils.java
@@ -21,7 +21,9 @@ package org.apache.pulsar.compaction;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.pulsar.common.protocol.Commands.DEFAULT_CONSUMER_EPOCH;
 import com.google.common.annotations.Beta;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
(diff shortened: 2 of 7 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicUtils.java
```diff
@@ -88,15 +92,27 @@ public static void asyncReadCompactedEntries(TopicCompactionService topicCompact
...
+                                }
+                            }
                         }
-                        cursor.updateReadStats(entries.size(), entriesSize);
+                        cursor.updateReadStats(unAckedEntries.size(), entriesSize);
 
                         Entry lastEntry = entries.get(entries.size() - 1);
                         cursor.seek(lastEntry.getPosition().getNext(), true);
-                        callback.readEntriesComplete(entries, readEntriesCtx);
+                        callback.readEntriesComplete(unAckedEntries, readEntriesCtx);
```
coderzc, line 115: This will cause these messages to be lost the next time compaction is run, and other consumers will not be able to read these messages. so why use individual ack for Exclusive mode?

coderzc (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicUtils.java
coderzc: > for the compaction_cursor_name __compaction , we don't skip the individual acked position, this will not affect the compaction

OK， I see

> The individual ack is not recommended in Exclusive mode? I don't know about this, I always use this before😂

Yes, for Exclusive mode, suggest using `acknowledgeCumulative` acknowledge. actually, if every message is confirmed, then acknowledgment and acknowledgment have the same effect, for this cause, I think #21187 has fixed it. But if the acknowledged messages are discontinuous then we can receive acknowledged messages, for this case, I think is a misuse of individual acknowledge in Exclusive mode

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java
```diff
@@ -893,6 +893,8 @@ default ManagedCursorAttributes getManagedCursorAttributes() {
 
     boolean isMessageDeleted(Position position);
 
+    boolean isMessageIndividualDeleted(Position position);
```
coderzc, line 896: I think we can using `isMessageDeleted` instead

pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicUtils.java
coderzc: @summeriiii I think for the compacted topic, we cannot use individual acknowledgment, otherwise the entry will not be deleted because the messages deleted by compact will never be acknowledged. so I suggest using `acknowledgeCumulative` acknowledge message for compacted topic.

Issue #23501, opened 2024-10-22
[Bug] Reader.hasMessageAvailable return wrong value after seeking by timestamp with startMessageIdInclusive


summeriiii:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

master

### Minimal reproduce step

1. create new reader with startMessageIdInclusive
2. seek to the current time  `reader.seek(System.currentTimeMillis());`
3. call the `reader.hasMessageAvailable()`
```java
    @Test(dataProvider = "initializeLastMessageIdInBroker")
    public void testHasMessageAvailableAfterSeekTimestampWithMessageIdInclusive(boolean initializeLastMessageIdInBroker)
            throws Exception {
        final String topic = "persistent://my-property/my-ns/" +
                "testHasMessageAvailableAfterSeekTimestampWithMessageInclusive";

        @Cleanup
        Producer<String> producer = pulsarClient.newProducer(Schema.STRING).topic(topic).create();
        final long timestampBeforeSend = System.currentTimeMillis();
        final MessageId sentMsgId = producer.send("msg");

        final List<MessageId> messageIds = new ArrayList<>();
        messageIds.add(MessageId.earliest);
        messageIds.add(sentMsgId);
        messageIds.add(MessageId.latest);

        for (MessageId messageId : messageIds) {
            @Cleanup
            Reader<String> reader = pulsarClient.newReader(Schema.STRING).topic(topic).receiverQueueSize(1)
                    .startMessageIdInclusive()
                    .startMessageId(messageId).create();
            if (initializeLastMessageIdInBroker) {
                assertTrue(reader.hasMessageAvailable());
            }
            reader.seek(System.currentTimeMillis());
            assertFalse(reader.hasMessageAvailable());
            Message<String> message = reader.readNext(10, TimeUnit.SECONDS);
            assertNull(message);
        }

        for (MessageId messageId : messageIds) {
            @Cleanup
            Reader<String> reader = pulsarClient.newReader(Schema.STRING).topic(topic).receiverQueueSize(1)
                    .startMessageIdInclusive()
                    .startMessageId(messageId).create();
            if (initializeLastMessageIdInBroker) {
                assertTrue(reader.hasMessageAvailable());
            }
            reader.seek(timestampBeforeSend);
[...]

Issue #23474, opened 2024-10-17
Flaky-test: BadVersionExceptions in OneWayReplicatorUsingGlobalZKTest.cleanup 


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/apache/pulsar/actions/runs/11368523712/job/31653376302?pr=23468#step:11:2008

### Exception stacktrace

```
  Error:  Failures: 
  Error:    OneWayReplicatorUsingGlobalZKTest.cleanup » TestNGRuntime org.apache.pulsar.client.admin.PulsarAdminException$ServerSideErrorException: 
   --- An unexpected error occurred in the server ---
  
  Message: org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp_-70132750-44af-4d14-817c-219034d2b7be-partition-0/pulsar.repl.r2
  
  Stacktrace:
  
  org.apache.pulsar.broker.service.BrokerServiceException$PersistenceException: org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp_-70132750-44af-4d14-817c-219034d2b7be-partition-0/pulsar.repl.r2
  	at org.apache.pulsar.broker.service.persistent.PersistentTopic$6.deleteLedgerFailed(PersistentTopic.java:1546)
  	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.lambda$asyncDelete$35(ManagedLedgerImpl.java:2978)
  	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:863)
  	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:841)
  	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
  	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2194)
[...]

PR #23502, opened 2024-10-22
[fix][client] Fix Reader.hasMessageAvailable return wrong value after seeking by timestamp with startMessageIdInclusive

summeriiii:
Fixes https://github.com/apache/pulsar/issues/23501

### Motivation
When seek to the current time `reader.seek(System.currentTimeMillis());`,  `reader.hasMessageAvailable()` should return false cause there has no message to read, but right now it returns true.
                                      
### Modifications
- when hasSoughtByTimestamp, we only need to check  result < 0
```java 
else if (hasSoughtByTimestamp) {
      completehasMessageAvailableWithValue(booleanFuture, result < 0);
}
```
 
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ReaderTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ReaderTest.java
@@ -906,9 +906,8 @@ public class ReaderTest extends MockedPulsarServiceBaseTest {
         }
     }
 
-    @Test(dataProvider = "initializeLastMessageIdInBroker")
-    public void testHasMessageAvailableAfterSeekTimestampWithMessageIdInclusive(boolean initializeLastMessageIdInBroker)
-            throws Exception {
+    @Test
+    public void testHasMessageAvailableAfterSeekTimestampWithMessageIdInclusive() throws Exception {
         final String topic = "persistent://my-property/my-ns/" +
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ReaderTest.java
```diff
@@ -906,6 +906,49 @@ public void testHasMessageAvailableAfterSeekTimestamp(boolean initializeLastMess
         }
     }
 
+    @Test(dataProvider = "initializeLastMessageIdInBroker")
+    public void testHasMessageAvailableAfterSeekTimestampWithMessageIdInclusive(boolean initializeLastMessageIdInBroker)
```
lhotari, line 910: I don't really understand the purpose of `initializeLastMessageIdInBroker`. I tried looking at the existing tests that have it, but I don't see why it's used. Could it be removed from this test?

lhotari (approved):
LGTM, thanks for the contribution @summeriiii

BewareMyPower (approved):

BewareMyPower:
The failed flaky test is https://github.com/apache/pulsar/issues/23474, let's merge this PR first

PR #23515, opened 2024-10-25
[fix][broker] Fix ownership loss

nodece:
### Motivation

 `org.apache.pulsar.client.api.BrokerServiceLookupTest#testLookupConnectionNotCloseIfGetUnloadingExOrMetadataEx` test failed locally and `ascentstream/pulsar` ci:
```
2024-10-26T01:32:04,373 - INFO  - [pulsar-client-io-37-3:ConnectionHandler] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] [s1] Closed connection [id: 0x14fd4828, L:/127.0.0.1:49376 - R:localhost/127.0.0.1:49366] -- Will try again in 0.1 s, hostUrl: null
2024-10-26T01:32:04,378 - INFO  - [bookkeeper-ml-scheduler-OrderedScheduler-2-0:ManagedCursorImpl] - [public/default/persistent/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e][s1] Closed cursor at md-position=7:-1
2024-10-26T01:32:04,379 - INFO  - [bookkeeper-ml-scheduler-OrderedScheduler-2-0:PersistentTopic] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] Topic closed
2024-10-26T01:32:04,475 - INFO  - [pulsar-timer-75-1:ConnectionHandler] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] [test-0-0] Reconnecting after 0.1 s timeout, hostUrl: null
2024-10-26T01:32:04,475 - INFO  - [pulsar-timer-75-1:ConnectionHandler] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] [s1] Reconnecting after 0.1 s timeout, hostUrl: null
2024-10-26T01:32:04,476 - INFO  - [metadata-store-2-1:ResourceLockImpl] - Acquired resource lock on /namespace/public/default/0x00000000_0xffffffff
2024-10-26T01:32:04,476 - INFO  - [metadata-store-2-1:ResourceLockImpl] - Successfully re-acquired missing lock at /namespace/public/default/0x00000000_0xffffffff
2024-10-26T01:32:04,476 - INFO  - [metadata-store-2-1:ResourceLockImpl] - Successfully revalidated the lock on /namespace/public/default/0x00000000_0xffffffff
2024-10-26T01:32:04,488 - INFO  - [pulsar-client-io-37-3:ConsumerImpl] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e][s1] Subscribing to topic on cnx [id: 0x14fd4828, L:/127.0.0.1:49376 - R:localhost/127.0.0.1:49366], consumerId 0
2024-10-26T01:32:04,488 - INFO  - [pulsar-client-io-37-3:ProducerImpl] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] [test-0-0] Creating producer on cnx [id: 0x14fd4828, L:/127.0.0.1:49376 - R:localhost/127.0.0.1:49366]
2024-10-26T01:32:04,490 - INFO  - [pulsar-io-8-4:ServerCnx] - [[id: 0x4b4d1182, L:/127.0.0.1:49366 - R:/127.0.0.1:49376] [SR:127.0.0.1, state:Connected]] Subscribing on topic persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e / s1. consumerId: 0, role: null
2024-10-26T01:32:04,493 - WARN  - [pulsar-io-8-4:BrokerService] - Namespace bundle for topic (persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e) not served by this instance:localhost:49367. Please redo the lookup. Request is denied: namespace=public/default
2024-10-26T01:32:04,494 - WARN  - [pulsar-io-8-4:ServerCnx] - [/127.0.0.1:49376][persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e][s1] Failed to create consumer: consumerId=0, Namespace bundle for topic (persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e) not served by this instance:localhost:49367. Please redo the lookup. Request is denied: namespace=public/default
2024-10-26T01:32:04,499 - WARN  - [pulsar-io-8-4:BrokerService] - Namespace bundle for topic (persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e) not served by this instance:localhost:49367. Please redo the lookup. Request is denied: namespace=public/default
2024-10-26T01:32:04,514 - WARN  - [pulsar-client-io-37-3:ClientCnx] - [id: 0x14fd4828, L:/127.0.0.1:49376 - R:localhost/127.0.0.1:49366] Received error from server: Namespace bundle for topic (persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e) not served by this instance:localhost:49367. Please redo the lookup. Request is denied: namespace=public/default
2024-10-26T01:32:04,514 - WARN  - [pulsar-client-io-37-3:ConsumerImpl] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e][s1] Failed to subscribe to topic on localhost/127.0.0.1:49366
2024-10-26T01:32:04,514 - WARN  - [pulsar-client-io-37-3:ConnectionHandler] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] [s1] Error connecting to broker: org.apache.pulsar.client.api.PulsarClientException$LookupException: {"errorMsg":"Namespace bundle for topic (persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e) not served by this instance:localhost:49367. Please redo the lookup. Request is denied: namespace=public/default","reqId":850604948710385036, "remote":"localhost/127.0.0.1:49366", "local":"/127.0.0.1:49376"}
2024-10-26T01:32:04,515 - WARN  - [pulsar-client-io-37-3:ConnectionHandler] - [persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e] [s1] Could not get connection to broker: org.apache.pulsar.client.api.PulsarClientException$LookupException: {"errorMsg":"Namespace bundle for topic (persistent://public/default/tp-359a79af-7dbe-4aa6-9e2a-6a1af63c4d8e) not served by this instance:localhost:49367. Please redo the lookup. Request is denied: namespace=public/default","reqId":850604948710385036, "remote":"localhost/127.0.0.1:49366", "local":"/127.0.0.1:49376"} -- Will try again in 0.199 s
```

In the lookup operation, if the ownership cannot be found in the `ownedBundlesCache` but exists in the metadata store, the broker will directly read the ownership from the metadata store by the `LockManager`, please see `org.apache.pulsar.broker.namespace.OwnershipCache#getOwnerAsync`.

When the broker creates the producer, it will check if the topic ownership exists on the cache, if not found, print `Please redo the lookup. Request is denied: namespace=public/default`, and causes the client to loop through reconnecting to the current broker.

I'm not sure if the issue is only in the pulsar testing environment, because we directly delete the zk data and cache data in the case.

### Modifications

- If the ownership belongs to the current broker in the `getOwnerAsync`, the broker will try to acquire ownership by the `ownedBundlesCache` to avoid the ownership loss in the cache.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java
```diff
@@ -170,7 +170,17 @@ public CompletableFuture<Optional<NamespaceEphemeralData>> getOwnerAsync(Namespa
 
         // If we're not the owner, we need to check if anybody else is
         String path = ServiceUnitUtils.path(suName);
-        return lockManager.readLock(path);
+        return lockManager.readLock(path).thenCompose(owner -> {
+            //If the current broker is the owner, we must try to acquire ownership to avoid cache loss.
```
BewareMyPower, line 174: What will happen if it's removed from the cache? I don't get the benefit here

nodece: The client to loop through reconnecting to the current broker.

lhotari: Is it possible to reproduce this in a test? Why does it get into a loop?

BewareMyPower: I have the same question. Could you explain more?

nodece: “org.apache.pulsar.client.api.BrokerServiceLookupTest#testLookupConnectionNotCloseIfGetUnloadingExOrMetadataExalways” fail locally.

Could you try to run this test?

The root cause is the topic’s ownership not found in the ownedBundlesCache, and then the broker sends an error to the client, so you see the client is reconnecting. When the ownership is never found, the client keeps trying reconnect. Please check the WARN log from the PR description for details. I hope this log can help you.

poorbarcode: @nodece 

I can not find a test named `testLookupConnectionNotCloseIfGetUnloadingExOrMetadataExalways` in the test class named `org.apache.pulsar.client.api.BrokerServiceLookupTest`

nodece: @poorbarcode `testLookupConnectionNotCloseIfGetUnloadingExOrMetadataEx` is correct.

nodece:
`org.apache.pulsar.client.api.BrokerServiceLookupTest#testLookupConnectionNotCloseIfGetUnloadingExOrMetadataEx` should be a flaky test.

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java
heesung-sn: Seems like we are adding a caching logic in this getter func. Seems like we are doing more work here although the original intention of this func is to simply fetch the lock info from the metadata store.


I think we need to confirm 

1. if there is any side effect from adding this logic
2. Change the func name and doc that clarifies it also fills the cache.


Usually, we fetches any metadata info from the cache which refreshes and does async load if not already cached. I am wondering why we are not following this practice here for this ownership check, or we missed anything that the original author intended here.

heesung-sn: Can we explain when/how this ownership cache and metadata store are out of sync?

heesung-sn: Seems like we don't invalidate ownedBundlesCache when deleting the ownership. (Maybe I missed something here)

https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java#L223-L233

It appears that we probably need to make those 

locallyAcquiredLocks
ownedBundlesCache
metadataStore

synchronized better.

heesung-sn: I guess we might see a racing condition here, when the same lock has been re-acquired immediately after release.

- locallyAcquiredLocks.put(namespaceBundle, rl) (new lock creation)
- invalidateLocalOwnerCache(namespaceBundle) (by lock.release deletes the ownedBundlesCache later)

ref: https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java#L100-L107

heesung-sn: This PR looks ok to fix the possible cache and metadata inconsistency state.

I think we can also try to skip `expiredFuture.complete(null);` if the lock is recreated on the same broker.

However, I think we better make this ownership release protocol clearer not to cause such inconsistency states. 
In ExtensibleLoadBalancer, we release the topic resources earlier in the first state(Releasing state) and blocks(ignores by conflict resolver) further re-assignment until the current transfer protocol finishes. So, I think we should see less such racing conditions in ExtensibleLoadBalancer.

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java
```diff
@@ -170,7 +170,17 @@ public CompletableFuture<Optional<NamespaceEphemeralData>> getOwnerAsync(Namespa
...
+            if (owner.isPresent() && owner.get().equals(selfOwnerInfo)) {
+                try {
+                    return tryAcquiringOwnership(suName).thenApply(Optional::ofNullable);
```
heesung-sn, line 177: As this shouldn't happen in normal cases. Please help add a warning log here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java
nodece: @heesung-sn Thanks for your explanation. I think the ownership release protocol may work fine.

> This PR looks ok to fix the possible cache and metadata inconsistency state.

All right. I'm not sure if the issue is only in the pulsar testing environment.

Please notice that the `org.apache.pulsar.client.api.BrokerServiceLookupTest.BundleOfTopic` class in the test.

```
        private void releaseBundleLockAndMakeAcquireFail() throws Exception {
            ownedBundlesCache.synchronous().invalidateAll();
            mockZooKeeper.delete(ServiceUnitUtils.path(namespaceBundle), -1);
            mockZooKeeper.setAlwaysFail(KeeperException.Code.OPERATIONTIMEOUT);
        }
```

This method directly discards the cache and deletes zk data,  which may break the test case.

I will double-check the ownership mechanism to avoid the racing condition.

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java
nodece: Ok, good catch!

nodece:
@lhotari @BewareMyPower @poorbarcode @heesung-sn This PR description has been updated, could you have a chance to review this PR?

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java
```diff
@@ -170,7 +170,22 @@ public CompletableFuture<Optional<NamespaceEphemeralData>> getOwnerAsync(Namespa
...
-        return lockManager.readLock(path);
+        return lockManager.readLock(path).thenCompose(owner -> {
+            // If the current broker is the owner, attempt to reacquire ownership to avoid cache loss.
+            if (owner.isPresent() && owner.get().equals(selfOwnerInfo)) {
+                log.warn("Detected ownership loss for broker [{}] on namespace bundle [{}]. "
+                                + "Attempting to reacquire ownership to maintain cache consistency.",
+                        selfOwnerInfo, suName);
+                try {
+                    return tryAcquiringOwnership(suName).thenApply(Optional::ofNullable);
+                } catch (Exception e) {
```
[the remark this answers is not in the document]
heesung-sn: shouldn't we use `exceptionally` instead?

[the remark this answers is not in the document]
nodece: @heesung-sn `tryAcquiringOwnershipthrows` defines throwing the Exception, so we must use try...catch. There is a need to clean up.

heesung-sn:
> Please notice that the org.apache.pulsar.client.api.BrokerServiceLookupTest.BundleOfTopic class in the test.
> 
>         private void releaseBundleLockAndMakeAcquireFail() throws Exception {
>             ownedBundlesCache.synchronous().invalidateAll();
>             mockZooKeeper.delete(ServiceUnitUtils.path(namespaceBundle), -1);
>             mockZooKeeper.setAlwaysFail(KeeperException.Code.OPERATIONTIMEOUT);
>         }
> This method directly discards the cache and deletes zk data, which may break the test case.
> 
> I will double-check the ownership mechanism to avoid the racing condition.
> 


This sounds like this ownership metadata and cache inconsistency issue only occurs from the wrong test setup. Should we just fix the test code in this case?

nodece:
> This sounds like this ownership metadata and cache inconsistency issue only occurs from the wrong test setup.

@heesung-sn Sure, this is the root cause.

> Should we just fix the test code in this case?

In one case, the user may delete the ownership on the zk. This PR can fix this issue and help the user recover ownership.

What do you think that?

heesung-sn (approved):

heesung-sn:
> In one case, the user may delete the ownership on the zk. This PR can fix this issue and help the user recover ownership.

Please consider adding comments to explain this.

I think this PR is useful to fix the possible inconsistency state, although I think that we better fix the root cause that makes this inconsistency state, if possible.

nodece:
Ok, I'll try to fix the root cause.

PR #23520, opened 2024-10-28
[improve][broker] Make cluster metadata teardown command support metadata config path

Demogorgon314:
### Motivation

The pulsar configuration supported `metadataStoreConfigPath` and `configurationStoreConfigPath`, but the cluster metadata teardown command does not have a place to set the config path.

### Modifications

Make cluster metadata init command support metadata config path

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
@@ -37,6 +37,7 @@ import org.apache.pulsar.metadata.api.MetadataStore;
 import org.apache.pulsar.metadata.api.MetadataStoreConfig;
 import org.apache.pulsar.metadata.api.MetadataStoreFactory;
 import org.apache.pulsar.metadata.api.extended.MetadataStoreExtended;
+import org.apache.pulsar.metadata.impl.ZKMetadataStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import picocli.CommandLine;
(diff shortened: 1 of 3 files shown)


poorbarcode (approved):

pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
```diff
@@ -55,6 +56,14 @@ private static class Arguments {
                 "--zookeeper"}, description = "Local ZooKeeper quorum connection string", required = true)
```
nodece, line 56: Please remove `required = true`.

pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
```diff
@@ -99,11 +113,22 @@ public static void main(String[] args) throws Exception {
             throw e;
         }
 
+        if (arguments.metadataStoreUrl == null && arguments.zookeeper == null) {
+            System.err.println("Metadata store address argument is required (--metadata-store)");
+            commander.usage(commander.getOut());
+            System.exit(1);
```
nodece, line 119: DOT USE `System.exit(1)`. In the testing environment, it may cause the test to exit.

```java
        if (arguments.metadataStoreUrl == null && arguments.zookeeper == null) {
            commander.usage(commander.getOut());
            throw new IllegalArgumentException("Metadata store address argument is required (--metadata-store)");
        }
```

nodece (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
Demogorgon314: Updated.

pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
```diff
@@ -99,11 +113,22 @@ public static void main(String[] args) throws Exception {
             throw e;
         }
 
+        if (arguments.metadataStoreUrl == null && arguments.zookeeper == null) {
+            System.err.println("Metadata store address argument is required (--metadata-store)");
```
nodece, line 117:
```suggestion
```

nodece (approved):

PR #23523, opened 2024-10-28
[fix][broker] Correct the 'connected' status in topic stats of multi-partition topic for geo replication

Coselding:
### Motivation

The default value of the `connected` field in `org.apache.pulsar.common.policies.data.stats.ReplicatorStatsImpl` is false, and the logic for accumulating each partition (`org.apache.pulsar.common.policies.data.stats.ReplicatorStatsImpl#add`) is `this.connected &= stats.connected;`. This causes the status of a multi-partition topic to always be false after calculation, and it is impossible to truly feedback the correct status of the geo replication task.

### Modifications

Modify the `connected` field in `org.apache.pulsar.common.policies.data.stats.ReplicatorStatsImpl` from false to true.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
```diff
@@ -4354,6 +4354,7 @@ CompletableFuture<Message<byte[]>> examineMessageAsync(String topic, String init
      *
      * @param topic
      * @param applied
+     *            Whether to return the configuration value of the previous level if the current level's don't set
```
lhotari, line 4357: Is this update related to this pull request? If not, please remove from this PR and improve the javadoc in a separate PR.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
```diff
@@ -4364,6 +4365,7 @@ CompletableFuture<Message<byte[]>> examineMessageAsync(String topic, String init
...
```
lhotari, line 4368: Is this update related to this pull request? If not, please remove from this PR and improve the javadoc in a separate PR.

Coselding: I corrected the problem and submitted a new Pull Request. <https://github.com/apache/pulsar/pull/23695>

Issue #23506, opened 2024-10-23
[Bug] managedLedgerMaxReadsInFlightSizeInMB limit doesn't work properly if managedLedgerReadEntryTimeoutSeconds isn't set


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

all released versions

### Minimal reproduce step

Problem description:
When the system is loaded, an exception such as `Time-out elapsed while acquiring enough permits on the memory limiter to read from ledger [ledgerid], [topic], estimated read size [read size] bytes for [dispatcherMaxReadBatchSize] entries (check managedLedgerMaxReadsInFlightSizeInMB)` will happen when `managedLedgerReadEntryTimeoutSeconds` isn't set.
The solution expects that `managedLedgerReadEntryTimeoutSeconds` has been set.
In addition to this, the solution is inefficient since retries happen in a tight loop. It's also possible that ordering of reads gets mixed up since any next call will get the permits for doing the next read.

### What did you expect to see?

`managedLedgerMaxReadsInFlightSizeInMB` limit should work also with setting `managedLedgerReadEntryTimeoutSeconds`

### What did you see instead?

timeouts could happen when `managedLedgerReadEntryTimeoutSeconds` isn't set. "fairness" is missing for the waiting read requests. This could cause starvation type of issue when the system is overloaded.

### Anything else?

 InflightReadsLimiter should be refactored to return a `CompletableFuture` so that logic can be handled asynchronously and reactively. There should be a queue so that the next waiting acquire call will prevent other calls until there are sufficient amount of permits available.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

lhotari:
I noticed also multiple other issues in InflightReadsLimiter / PendingReadsManager:
* PendingReadsManager will acquire permits for partial reads in de-duplication which is wrong
* Permits aren't released until the buffer is deallocated
  * When the entry is cached, it would limit the reads which is not intended by managedLedgerMaxReadsInFlightSizeInMB

lhotari:
Possible issue with PendingReadsManager:
* when reads are de-duplicated, the same entries are returned to all callers in case of an exact match. This could be problematic since readerIndex is shared state of the ByteBuffer and that gets mutated in some cases.

Issue #23505, opened 2024-10-23
[Bug] Redelivering messages doesn't take dispatcherMaxReadSizeBytes into account in Shared and Key_Shared subscriptions


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

all released versions

### Minimal reproduce step

Problem description:
In the Shared subscription, messages get added to the replay queue when a consumer disconnects. In Key_Shared subscription, the replay queue is also used when messages cannot be dispatched to a target consumer due to insufficient permits or when the hash is blocked.
There's a problem in the current implementation since the `dispatcherMaxReadSizeBytes` (default 5MB) setting isn't taken into account in the reads. The impact of this is that consumers will receive a large batch messages at once if the reads succeed. The exact implication of this isn't fully known at this time. However, it's against the design to ignore the `dispatcherMaxReadSizeBytes` setting which is helpful in making smaller incremental progress on individual dispatchers and service all active dispatchers in the broker one-by-one as fairly as possible.

### What did you expect to see?

That `dispatcherMaxReadSizeBytes` is used for redelivering message.

### What did you see instead?

`dispatcherMaxReadSizeBytes` is ignored.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

ZhaoGuorui666:
Hi lhotari,

I have been looking into this issue today and noticed that you have self-assigned it. I would like to discuss the following points with you:

Where is the dispatcherMaxReadSizeBytes limit applied in the code?

* Before inserting into the replay queue?
![image](https://github.com/user-attachments/assets/f905b510-40c7-4acc-9ee4-6f77596cf3a5)

* Within the readMoreEntries() method? (However, I noticed that the calculateToRead() method already imposes a limit at the beginning)
![image](https://github.com/user-attachments/assets/a170da1f-57cf-4f3b-9eef-09ca81f45140)

Looking forward to your PR and learning from your approach to resolving this issue.

Thank you!

lhotari:
> Hi lhotari,
> 
> I have been looking into this issue today and noticed that you have self-assigned it. I would like to discuss the following points with you:
> 
> Where is the dispatcherMaxReadSizeBytes limit applied in the code?
> 
> * Before inserting into the replay queue?
[...]

Issue #23504, opened 2024-10-23
[Bug] Redelivery (replay) queue reads don't get deduplicated or limited by managedLedgerMaxReadsInFlightSizeInMB


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

any released version

### Minimal reproduce step

problem description:
This problem applies to Shared and Key_Shared subscriptions.
When a consumer disconnects and while there are other active consumers, the unacknowledged (pending) message ids get added to the redelivery (replay) queue in the dispatcher. When these messages get dispatched on the next round in the Shared or Key_Shared subscription dispatcher, the reads for the messages don't get deduplicated or limited by managedLedgerMaxReadsInFlightSizeInMB. This is a problem for high fanout scenarios for Shared subscriptions, when a large amount of consumers disconnect and reconnect while there's at least one active consumer so that the dispatcher doesn't discard the replay queue and "rewind" in between.

### Additional problem

When an entry gets added to the replay queue, it doesn't get added to the cache. The read for this entry will usually happen short after and this will result in a new read to bookkeeper. This issue will be reported separately so that it's possible to address issues individually.

### What did you expect to see?

That redelivery reads get deduplicate and limited by managedLedgerMaxReadsInFlightSizeInMB.

### What did you see instead?

PendingReadsManager isn't used, so all reads will be made multiple times to bookkeeper (in fanout with multiple consumers). The reads won't be limited  by managedLedgerMaxReadsInFlightSizeInMB.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

Issue #23482, opened 2024-10-18
[Enhancement] Adjust batch size in dispatcher so that a single batch doesn't exceed `managedLedgerMaxReadsInFlightSizeInMB`


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Current `managedLedgerMaxReadsInFlightSizeInMB` will have to be set to value that is higher than `dispatcherMaxReadBatchSize` * `maxMessageSize`. 

Otherwise it could result in error `Time-out elapsed while acquiring enough permits on the memory limiter to read from ledger [ledgerid], [topic], estimated read size [read size] bytes for [dispatcherMaxReadBatchSize] entries (check managedLedgerMaxReadsInFlightSizeInMB)`.
`dispatcherMaxReadBatchSize` defaults to `100` and `maxMessageSize` defaults to `5MB` in bytes.

This means that currently `managedLedgerMaxReadsInFlightSizeInMB` must be set to `500` or higher to prevent the issue from occurring in all cases.

### Solution

Adjust batch size in dispatcher so that a single batch doesn't exceed `managedLedgerMaxReadsInFlightSizeInMB`

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
There's another problem. The current solution expects that managedLedgerReadEntryTimeoutSeconds is set. It will fail immediately without waiting in that case, resulting in a similar exception.

lhotari:
`dispatcherMaxReadSizeBytes` is already for the complete batch so this is an invalid issue.

PR #23524, opened 2024-10-28
[fix][broker] Fix reading entries failed due to max in-flight reading

Technoboy-:
### Motivation

If the `estimatedReadSize` larger than the `managedLedgerMaxReadsInFlightSizeInMB`, reading entries will fail.

```
2024-10-10T14:31:46,939+0000 [BookKeeperClientWorker-OrderedExecutor-2-0] ERROR org.apache.bookkeeper.mledger.impl.cache.RangeEntryCacheImpl - Time-out elapsed while acquiring enough permits on the memory limiter to read from ledger 114226, public/default/persistent/test-partition-0, estimated read size 120875300 bytes for 100 entries (check managedLedgerMaxReadsInFlightSizeInMB)
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
@@ -23,6 +23,8 @@ import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.metrics.ObservableLongCounter;
 import io.prometheus.client.Gauge;
 import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.opentelemetry.Constants;
@@ -56,7 +58,9 @@ public class InflightReadsLimiter implements AutoCloseable {
             .help("Available space for inflight data read from storage or cache")
             .register();
 
-    private final long maxReadsInFlightSize;
+    @Setter
+    @Getter
+    private long maxReadsInFlightSize;
     private long remainingBytes;
 
     public InflightReadsLimiter(long maxReadsInFlightSize, OpenTelemetry openTelemetry) {
@@ -177,6 +181,5 @@ public class InflightReadsLimiter implements AutoCloseable {
     public boolean isDisabled() {
         return maxReadsInFlightSize <= 0;
     }
-
 
 }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
@@ -362,7 +362,12 @@ public class RangeEntryCacheImpl implements EntryCache {
         }
         long estimatedReadSize = (1 + lastEntry - firstEntry)
                 * (estimatedEntrySize + BOOKKEEPER_READ_OVERHEAD_PER_ENTRY);
-        if (estimatedReadSize > pendingReadsLimiter.getMaxReadsInFlightSize()) {
+        final long maxReadsInFlightSize = pendingReadsLimiter.getMaxReadsInFlightSize();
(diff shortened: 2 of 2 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
```diff
@@ -362,6 +362,9 @@ private AsyncCallbacks.ReadEntriesCallback handlePendingReadsLimits(ReadHandle l
         }
         long estimatedReadSize = (1 + lastEntry - firstEntry)
                 * (estimatedEntrySize + BOOKKEEPER_READ_OVERHEAD_PER_ENTRY);
+        if (estimatedReadSize > pendingReadsLimiter.getMaxReadsInFlightSize()) {
+            pendingReadsLimiter.setMaxReadsInFlightSize(estimatedReadSize);
```
eolivelli, line 366: Here we are mutating the configuration of a shared object.

We could log an error but in the end the administrator has to change the value

eolivelli (changes requested):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
Technoboy-: address the comment.
Add log here

eolivelli: I am sorry, I was not clear.
The point is not about adding only the logging.
But we cannot update the configuration dynamically here.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
```diff
@@ -56,7 +58,9 @@ public class InflightReadsLimiter implements AutoCloseable {
             .help("Available space for inflight data read from storage or cache")
             .register();
 
-    private final long maxReadsInFlightSize;
+    @Setter
+    @Getter
+    private long maxReadsInFlightSize;
```
eolivelli, line 63: This variable should stay final

eolivelli (changes requested):
We should not update a global variable this way. The update will affect all of the ledgers/topics without control.

Also, we should add a test that reproduces the problem and ensures that it is fixed

@lhotari wdyt?

lhotari:
@Technoboy- I'm working on a broader refactoring to address the issue. There are multiple challenges. I have reported multiple issues, #23482, #23504, #23505 and #23506. I'm working to resolve them soon.

lhotari:
One reason why the read size could exceed managedLedgerMaxReadsInFlightSizeInMB is #23482, that's what I'm also addressing in my changes that are WIP ([example commit](https://github.com/lhotari/pulsar/commit/769911dcf6f4f82fd8e9a96208b3ce1f16fe92d6), part of [the WIP changes](https://github.com/apache/pulsar/compare/master...lhotari:pulsar:lh-fix-individual-read-deduplication-and-limiting)).

Technoboy-:
> @Technoboy- I'm working on a broader refactoring to address the issue. There are multiple challenges. I have reported multiple issues, #23482, #23504, #23505 and #23506. I'm working to resolve them soon. While fixing the issues, I have discovered more. For example, when reads get de-duplicated by PendingReadsManager and there are partial matches, the permits will get acquired also for the partial reads. I have a large change set in progress, which I will split into smaller pull requests once I have addressed the issues with the refactoring and the improvements and I have tests passing. Since my main goal is to improve caching for Key_Shared subscriptions, this has revealed more gaps in addressing that. In the current solution replay reads aren't cached at all. I noticed that @eolivelli has reported a related issue #16421 about that for Shared subscriptions. The comment [#16421 (comment)](https://github.com/apache/pulsar/issues/16421#issuecomment-1177516197) is relevant. Messages in the replay queues shouldn't be discarded from the cache. I'm also trying to address that in my experiments. That's why the changes have expanded to also address broker cache short comings.

ok,  i will close this patch.

PR #23525, opened 2024-10-29
[improve][client][PIP-389] Add a producer config to improve compression performance

liangyepianzhou:
PIP: https://github.com/apache/pulsar/pull/23526
### Motivation

The motivation of this PIP is to provide a way to improve the compression performance by skipping the compression of small messages.
We want to add a new configuration compressMinMsgBodySize to the producer configuration.
This configuration will allow the user to set the minimum size of the message body that will be compressed.
If the message body size is less than the compressMinMsgBodySize, the message will not be compressed.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


lhotari:
Please add the PIP number to the PR title as we usually do.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ProducerConfigurationData.java
```diff
@@ -189,6 +189,8 @@ public class ProducerConfigurationData implements Serializable, Cloneable {
     )
     private CompressionType compressionType = CompressionType.NONE;
 
+    private int compressMinMsgBodySize = 4 * 1024; // 4kb
```
lhotari, line 192: What's the rationale for the default value?

liangyepianzhou: - It is default value for rocket mq.
https://github.com/apache/rocketmq/blob/dd62ed0f3b16919adec5d5eece21a1050dc9c5a0/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java#L117
- And in my testing (against Pulsar Producer), when the message body is in the 3k-5k range, the CPU usage of compressing the same data starts to drop significantly and the compression rate increases significantly.

liangyepianzhou: This data may be more convincing.
https://docs.google.com/spreadsheets/d/1pFOAgxVYsos38oraeva1_RHC9P4J3oN2Oj2fQ65L8OI/edit?gid=0#gid=0

lhotari: Thanks for sharing the reasoning.

lhotari:
@liangyepianzhou Regarding performance optimizations for compression in Pulsar, there's also work that should be done.
For example for gzip compression/decompression this is very inefficient: https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-common/src/main/java/org/apache/pulsar/common/compression/CompressionCodecZLib.java#L60-L85 . 
Another detail is that the current implementation isn't using "zero copy" approaches that are available.
For example in Snappy:
https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-common/src/test/java/org/apache/pulsar/common/compression/CompressionCodecSnappyJNI.java#L34-L64
In BookKeeper, I added zero-copy for calculating checksums in https://github.com/apache/bookkeeper/pull/4196. The ByteBufVisitor approach could be used to avoid copying source buffers to an extra nio buffer.
Calling Netty's io.netty.buffer.CompositeByteBuf#nioBuffer will allocate a new nio ByteBuffer in the heap and copy the content there. That's not very great from performance perspective, especially when we want to reduce allocations and garbage. With the ByteBufVisitor approach it's possible to read the source direct byte buffers without extra copies.
**Have you considered in addressing this performance issue in the Pulsar message compression solution?**

liangyepianzhou:
>Have you considered in addressing this performance issue in the Pulsar message compression solution?

Sounds good, maybe I can try optimizing it in other PRs

lhotari:
> > Have you considered in addressing this performance issue in the Pulsar message compression solution?
> 
> Sounds good, maybe I can try optimizing it in other PRs

+1, In the Pulsar code base, we have a special module called `microbench` for microbenchmarks with JMH, https://github.com/apache/pulsar/tree/master/microbench . Testing performance with JMH could be useful for such improvements.

liangyepianzhou:
> > > Have you considered in addressing this performance issue in the Pulsar message compression solution?
> > 
> > 
> > Sounds good, maybe I can try optimizing it in other PRs
> 
> +1, In the Pulsar code base, we have a special module called `microbench` for microbenchmarks with JMH, https://github.com/apache/pulsar/tree/master/microbench . Testing performance with JMH could be useful for such improvements.

Thanks for the reminder.

liangyepianzhou:
> @liangyepianzhou Regarding performance optimizations for compression in Pulsar, there's also work that should be done. For example for gzip compression/decompression this is very inefficient:
> 
> https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-common/src/main/java/org/apache/pulsar/common/compression/CompressionCodecZLib.java#L60-L85
> 
> .
> Another detail is that the current implementation isn't using "zero copy" approaches that are available.
> For example in Snappy:
> https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-common/src/test/java/org/apache/pulsar/common/compression/CompressionCodecSnappyJNI.java#L34-L64
> 
> 
> In BookKeeper, I added zero-copy for calculating checksums in [apache/bookkeeper#4196](https://github.com/apache/bookkeeper/pull/4196). The ByteBufVisitor approach could be used to avoid copying source buffers to an extra nio buffer.
> Calling Netty's io.netty.buffer.CompositeByteBuf#nioBuffer will allocate a new nio ByteBuffer in the heap and copy the content there. That's not very great from performance perspective, especially when we want to reduce allocations and garbage. With the ByteBufVisitor approach it's possible to read the source direct byte buffers without extra copies.
> **Have you considered in addressing this performance issue in the Pulsar message compression solution?**

@lhotari I try to optimize it yesterday.  but I found that Pulsar did not use `CompositeByteBuf` when sending messages. Single sending uses `io.netty.buffer.UnpooledHeapByteBufmemory`, while batch sending uses `io.netty.buffer.PooledUnsafeDirectByteBufmemory`. They all have memory addresses or
array, so the current implementation already uses zero copy.  It seems that there is no need to optimize?
And the example you gave, `CompressionCodecSnappyJNI.java`, is a compression class used in testing. It seems that there is no need to compress it.

liangyepianzhou:
@lhotari Do you have time to review the code changes for [pip-389](https://lists.apache.org/thread/xv7x3vmycxzsrhbdo7vmssh8lxxzyxd5) that have been approved on the mailing list?
Thank you!

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
```diff
@@ -38,11 +38,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import lombok.Getter;
-import org.apache.pulsar.client.api.Message;
-import org.apache.pulsar.client.api.MessageId;
-import org.apache.pulsar.client.api.MessageIdAdv;
-import org.apache.pulsar.client.api.Schema;
-import org.apache.pulsar.client.api.SchemaSerializationException;
```
lhotari, line 45: This change is against the Pulsar code style. IDE instructions: https://pulsar.apache.org/contribute/setup-ide/#configure-code-style

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerConsumerInternalTest.java
```diff
@@ -18,24 +18,21 @@
...
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.broker.BrokerTestUtil;
 import org.apache.pulsar.broker.service.ServerCnx;
-import org.apache.pulsar.client.api.BatcherBuilder;
-import org.apache.pulsar.client.api.Consumer;
-import org.apache.pulsar.client.api.Message;
-import org.apache.pulsar.client.api.MessageId;
-import org.apache.pulsar.client.api.Producer;
-import org.apache.pulsar.client.api.ProducerConsumerBase;
-import org.apache.pulsar.client.api.SubscriptionType;
```
lhotari, line 38: This change is against the Pulsar code style. IDE instructions: https://pulsar.apache.org/contribute/setup-ide/#configure-code-style

lhotari (changes requested):
Please use Pulsar code style. IDE instructions: https://pulsar.apache.org/contribute/setup-ide/#configure-code-style

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -505,22 +506,27 @@ public void sendAsync(Message<?> message, SendCallback callback) {
...
-                return;
+            } else {
+                compressedPayload = applyCompression(payload);
+                compressed = true;
+
+                // validate msg-size (For batching this will be check at the batch completion size)
+                int compressedSize = compressedPayload.readableBytes();
+                if (compressedSize > getMaxMessageSize() && !this.conf.isChunkingEnabled()) {
+                    compressedPayload.release();
+                    String compressedStr = conf.getCompressionType() != CompressionType.NONE ? "Compressed" : "";
```
lhotari, line 520: would it be useful to also show the compression type?
```suggestion
                    String compressedStr = conf.getCompressionType() != CompressionType.NONE ? ("compressed (" + conf.getCompressionType() + ")") : "uncompressed";
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
```diff
@@ -169,9 +169,20 @@ protected ByteBuf getCompressedBatchMetadataAndPayload() {
         }
 
         int uncompressedSize = batchedMessageMetadataAndPayload.readableBytes();
-        ByteBuf compressedPayload = compressor.encode(batchedMessageMetadataAndPayload);
-        batchedMessageMetadataAndPayload.release();
-        if (compressionType != CompressionType.NONE) {
+        ByteBuf compressedPayload;
+        boolean isCompressed = false;
+        if (!isBrokerTwoPhaseCompactor && producer != null){
+            if (uncompressedSize > producer.conf.getCompressMinMsgBodySize()) {
```
lhotari, line 175: This seems to miss the `compressionType != CompressionType.NONE` condition. It would be logical to have the condition here since the logic has been changed.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
```diff
@@ -169,9 +169,20 @@ protected ByteBuf getCompressedBatchMetadataAndPayload() {
...
+            } else {
+                compressedPayload = batchedMessageMetadataAndPayload;
+            }
+        } else {
+            compressedPayload = compressor.encode(batchedMessageMetadataAndPayload);
+            batchedMessageMetadataAndPayload.release();
+        }
+        if (compressionType != CompressionType.NONE && isCompressed) {
             messageMetadata.setCompression(compressionType);
             messageMetadata.setUncompressedSize(uncompressedSize);
```
lhotari, line 187: move these lines after the code where compression is applied (`producer.applyCompression`).

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
```diff
@@ -141,7 +141,7 @@ public boolean add(MessageImpl<?> msg, SendCallback callback) {
         return isBatchFull();
     }
 
-    protected ByteBuf getCompressedBatchMetadataAndPayload() {
+    protected ByteBuf getCompressedBatchMetadataAndPayload(boolean isBrokerTwoPhaseCompactor) {
```
lhotari, line 144: Leaking implementation details that compaction doesn't compress messages is not very great. it would be better to rename `isBrokerTwoPhaseCompactor` to `allowCompression` and inverse the value. Add an override that passes `true` as the default value for `allowCompression`.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
```diff
@@ -252,7 +263,8 @@ public OpSendMsg createOpSendMsg() throws IOException {
         if (messages.size() == 1) {
             messageMetadata.clear();
             messageMetadata.copyFrom(messages.get(0).getMessageBuilder());
-            ByteBuf encryptedPayload = producer.encryptMessage(messageMetadata, getCompressedBatchMetadataAndPayload());
+            ByteBuf encryptedPayload = producer.encryptMessage(messageMetadata,
+                    getCompressedBatchMetadataAndPayload(false));
```
lhotari, line 267: this change can be reverted once there's a `getCompressedBatchMetadataAndPayload()` method as explained in one of the previous comments.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
```diff
@@ -283,7 +295,8 @@ public OpSendMsg createOpSendMsg() throws IOException {
             lowestSequenceId = -1L;
             return op;
         }
-        ByteBuf encryptedPayload = producer.encryptMessage(messageMetadata, getCompressedBatchMetadataAndPayload());
+        ByteBuf encryptedPayload = producer.encryptMessage(messageMetadata,
+                getCompressedBatchMetadataAndPayload(false));
```
lhotari, line 299: this change can be reverted once there's a `getCompressedBatchMetadataAndPayload()` method as explained in one of the previous comments.

lhotari (changes requested):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
liangyepianzhou: done

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
liangyepianzhou: done

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
liangyepianzhou: done

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
liangyepianzhou: done

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageContainerImpl.java
liangyepianzhou: done

lhotari (approved):

PR #23530, opened 2024-10-30
[fix][broker] Fix currently client retries until operation timeout if the topic does not exist

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/23291 introduces a regression that when the authentication is enabled, if the topic does not exist and cannot be automatically created (e.g. namespace does not exist), broker will respond with a `ServerError.MetadataError` that is converted to a trivial `PulsarClientException`, which is not retriable. In this case, the client will loop in sending the `PartitionMetadata` request and failing until the operation timeout happens (30 seconds by default)

Besides, `PulsarClientImpl#checkPartitions` is incorrect. Even if `forceNoPartitioned` is false, `checkPartitions.complete(0)` will still be called.

```java
            if (forceNoPartitioned && actEx instanceof PulsarClientException.NotFoundException
                    || actEx instanceof PulsarClientException.TopicDoesNotExistException
                    || actEx instanceof PulsarAdminException.NotFoundException) {
                checkPartitions.complete(0);
```

Because `a && b || c || d` is true when `a` and `b` are false while `c` is true. Without this fix, the producer creation on a topic whose namespace does not exist will fail with `AuthorizationError` because it will go to the topic lookup phase.

This is not an expected result because the case that the error is not retriable and the timeout error is confusing to users.

### Modifications

- Return `ServerError.TopicNotFound` when `isTopicOperationAllowed` fails with HTTP 404 error.
- Fix the incorrect check in `PulsarClientImpl#checkPartitions`.
- Add `testTopicNotFound` to cover the fixes above.
- Fix the `testNamespaceNotExist`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -713,7 +713,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                 Throwable actEx = FutureUtil.unwrapCompletionException(ex);
                 if (actEx instanceof WebApplicationException restException) {
                     if (restException.getResponse().getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
-                        writeAndFlush(Commands.newPartitionMetadataResponse(ServerError.MetadataError,
+                        writeAndFlush(Commands.newPartitionMetadataResponse(ServerError.TopicNotFound,
                         "Tenant or namespace or topic does not exist: " + topicName.getNamespace() ,
                                 requestId));
                         lookupSemaphore.release();
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/GetPartitionMetadataTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/GetPartitionMetadataTest.java
@@ -599,8 +599,7 @@ public class GetPartitionMetadataTest extends TestRetrySupport {
                 fail("Expected a not found ex");
             } catch (Exception ex) {
                 Throwable unwrapEx = FutureUtil.unwrapCompletionException(ex);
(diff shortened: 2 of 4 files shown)


Technoboy- (approved):

lhotari:
@BewareMyPower There's a problem in cherry-picking to branch-3.3 and branch-3.0 since TokenAuthenticatedProducerConsumerTest fails.

reproducing issue:
```shell
mvn -DredirectTestOutputToFile=false -DtestRetryCount=0 test -pl pulsar-broker -Dtest=TokenAuthenticatedProducerConsumerTest
```

```
[ERROR] Tests run: 13, Failures: 1, Errors: 0, Skipped: 10, Time elapsed: 14.379 s <<< FAILURE! - in org.apache.pulsar.client.api.TokenAuthenticatedProducerConsumerTest
[ERROR] org.apache.pulsar.client.api.TokenAuthenticatedProducerConsumerTest.testTopicNotFound[true, false](4)  Time elapsed: 2.08 s  <<< FAILURE!
java.lang.AssertionError: expected [true] but found [false]
	at org.apache.pulsar.client.api.TokenAuthenticatedProducerConsumerTest.testTopicNotFound(TokenAuthenticatedProducerConsumerTest.java:211)
```

It seems to be some sort of isolation problem since the test passes when the single method `testTopicNotFound` is run
```shell
mvn -DredirectTestOutputToFile=false -DtestRetryCount=0 test -pl pulsar-broker -Dtest=TokenAuthenticatedProducerConsumerTest#testTopicNotFound
```

@BewareMyPower do you have a chance to fix the test in branch-3.3 and also cherry-pick this PR to branch-3.0 after addressing the problem? thanks

BewareMyPower:
@lhotari It's weird that this test should also fail with the master branch. But the CI is not broken

Issue #23550, opened 2024-11-04
[Bug][txn] txn committing stuck and never finish commit process


TakaHiR07:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

3.0.x

### Minimal reproduce step

run transaction.

### What did you expect to see?

transaction not stuck

### What did you see instead?

After running txn for a long time in a multi-node cluster, notice that some transactions are stuck in one broker and never finish. These transactions are stuck in the committing -> committed process. However, after dumping the object, finding that the requestCredits is 1000, but transaction commit still can not be executed. And the tb_client_pending_request keep growing.

![image](https://github.com/user-attachments/assets/6624c3bd-e748-48fe-8454-33dc241e0a33)

![image](https://github.com/user-attachments/assets/484424c2-6e86-4fbb-a7c5-c19f573a2542)

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #23551, opened 2024-11-04
[fix][txn] fix concurrent error cause txn stuck in TransactionBufferHandlerImpl#endTxn

TakaHiR07:
Fixes https://github.com/apache/pulsar/issues/23550

### Motivation

After diving into the code, finding that there is a concurrent error in TransactionBufferHandlerImpl#checkRequestCredits(),  checkPendingRequests(), which would cause the above issue.

Currently, we have config TransactionBufferClientMaxConcurrentRequests to control the concurrent request number. However, if the request and response is executed as follow, the request would permanently stuck in queue. 
(to simplify the case, let's set permit is 1)

| step      | request-1  | request-2 | response-1 |
| ----------- | ----------- | ----------- | ----------- | 
| 1      |    start do checkRequestCredits()    |   |  |
| 2      |    compareAndSet requestCredits to 0    |   |  |
| 3      |    execute endTxn    |   |  |
| 4      |         |  start do checkRequestCredits()  |  |
| 5      |         |  get currentPermit = 0  |  |
| 6      |         |    |  trigger onResponse(), set requestCredits to 1|
| 7      |         |    | trigger checkPendingRequests(), permit == 1 && pendingRequests is null, so break the while process |
| 8      |         |  currentPermits == 0 && pendingRequest is null, then add op to pendingRequest  |  |

Now we can find there is no response can trigger pendingRequest.remove, and then all the new requests just add to pendingRequest but permanently not execute. 

### Modifications

The root reason is currently only onResponse() can trigger pendingRequest.remove. But when we execute onResponse(), the requestOp may not have been added to pendingRequest.

- So one modification is to let it can check the pendingRequest queue in checkRequestCredits()
- And the while(true) in checkPendingRequests() is not necessary, 1 response come back, take 1 requestOp from pendingRequest is OK.

It is hard to add test for this concurrent case. 

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


TakaHiR07:
@codelipenghui @congbobo184  Can you help review this problem?

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionBufferHandlerImpl.java
```diff
@@ -237,26 +240,25 @@ public void onResponse(OpRequestSend op) {
...
+                        invalid.recycle();
                     }
+                    endTxn(polled);
+                } else {
+                    REQUEST_CREDITS_UPDATER.incrementAndGet(this);
+                    checkPendingRequests();
                 }
             } else {
-                break;
+                checkPendingRequests();
```
lhotari, line 261: just wondering if deeply nested recursive calls could become a problem. is there a specific reason to replace the while loop with recursion?

TakaHiR07: Because the while loop is somewhat confusing, it look like 1 response can trigger taking multiple requests from pendingRequest. But actually it is executed when its permit > 0 . And permit would +1 when 1 response return. 

Besides, if different responses come back at the same time, both of them would go into while loop and compete for acquire the permits. That is not neccessary. 

So I replace to 1 response trigger take 1 request from pendingRequestQueue.  And only two case we should do recursion to retry:
- If the request taking is null, we should retry checkPendingRequest() to take the next one. 
- If permit.compareAndSet is not ok, it means there is concurrent checkRequestCredits() or checkPendingRequest(), we should retry checkPendingRequest() again to make sure response can trigger taking 1  request from pendingRequestQueue

TakaHiR07: The recursive calls actually may be problem. I have updated the code, the new one is better.

lhotari (approved):
LGTM

congbobo184 (approved):
LGTM

dao-jun (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionBufferHandlerImpl.java
```diff
@@ -251,6 +254,7 @@ private void checkPendingRequests() {
                             invalid.recycle();
                         }
                         endTxn(polled);
+                        break;
```
lhotari, line 257: would this break other usages of `checkPendingRequests`?

PR #23560, opened 2024-11-05
[fix][client] Fix race-condition causing doReconsumeLater to hang when creating retryLetterProducer has failed

hanmz:
### Motivation

When `retryLetterProducer` creation fails, `retryLetterProducer` will be set to null. So `retryLetterProducer==null `may occur when race-condition is present, at this time, the results cannot be completed.

Only handle the situation when retryLetterProducer is not equal to null:
https://github.com/apache/pulsar/blob/3f12269d8d99e514cf48ef6d57fc3928d37b3646/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L678

### Modifications
When `retryLetterProducer==null`, mark the result as an completeExceptionally.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
Do you have a chance to add a test case?

hanmz:
> Do you have a chance to add a test case?

I've added a test case.

lhotari (approved):
LGTM, thanks for the contribution @hanmz

PR #23562, opened 2024-11-05
[improve][broker]Restrict the creation of topics with "-partition--x"

zjxxzjwang:
### Motivation
If you actively create a topic with "-partition--x" (x is an integer)(for example, "peisisten://tenant-name/ns-name/topic-name-partition--1"), the topic can be created successfully, and a node with partition 3 will be attached to the zk. But it is not a partition topic

### Modifications

By modifying the isAllowAutoTopicCreationAsync logic () method, when the topic name contains "- partiton - x" (x is an integer), trigger logic returns false
image

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -3529,6 +3529,12 @@ public class BrokerService implements Closeable {
             return CompletableFuture.completedFuture(true);
         }
 
+        // If this topic name contains "-partition--x" , not allow to be created automatically.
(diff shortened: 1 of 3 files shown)


lhotari:
Please add a test case for the new functionality

zjxxzjwang:
> 请为新功能添加测试用例
OK, I've added a test case.

zjxxzjwang:
> 请为新功能添加测试示例

OK, I've added a test case.

lhotari (approved):
LGTM

PR #23563, opened 2024-11-05
[improve][admin] Print error log if handle http response fails

poorbarcode:
### Motivation

When users are calling `pulsar-admin` API, and there is an internal error( such as a deserialize JSON error), there is no error log.

You can test it this way:
- add a method getMsgRateIn() for `TopicStatsImpl`, and return a `String` value(such as `abc`), which will make the `response.msgRateIn` as a String value `abc`.
- call `pulsar-admin topics stats`
  - You will get nothing, and no error log is printed. 

### Modifications

Print error log if handle HTTP response fails

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


shibd (approved):

lhotari (approved):
LGTM

Technoboy- (approved):

PR #23565, opened 2024-11-05
[improve][pip] Improve the reusability of Pulsar integration test code and best practices for unit and integration tests in Pulsar

heesung-sn:
This is a PIP

### DIFF
--- a/pip/pip-390.md
+++ b/pip/pip-390.md
@@ -1,0 +1,78 @@
+# PIP-390: Improve the reusability of Pulsar integration test code and best practices for unit and integration tests in Pulsar
+
+## Background
+
+- Integration tests for Apache Pulsar are located in the `tests` module.
+
+## Motivation
+
+Apache Pulsar has several reusable test code, such as topic message test cases, container setup, and other helper functions in the `tests` module. However, since this module is only accessible within the "test" scope, it is difficult to reuse these utilities in other projects.
+
+The Pulsar community has discussed the ideal structures for unit and integration tests in Pulsar (see discussion [here](https://lists.apache.org/thread/fn3rk1x7v9291qh3g6vf4jxhvq6zc4mm)). This PIP aims to document the results of these discussions for future reference.
+
+## Goals
+
+### In Scope
+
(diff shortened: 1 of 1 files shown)


pip/pip-390.md
```diff
@@ -0,0 +1,79 @@
+# PIP-390: Improve the reusability of Pulsar test code and best practices for unit and integration tests in Pulsar
+
+## Background
+
+- Pulsar integration tests are located in the `tests` module, which run on [testcontainers](https://github.com/testcontainers).
+- Pulsar unit tests are located in each module under the test scope, with many tests running on mocked (in-memory) Pulsar clusters. 
+- Currently, Pulsar tests run on the `testng` framework.
```
zzzk1, line 7:
```suggestion
- Currently, Pulsar tests run on the [TestNG](https://testng.org/#_welcome_to_testng).
```

zzzk1: I believe that directly providing the framework link may be clearer.

falser101: So is Pulsar's testing going to switch to using JUnit 5 and remove testng？

lhotari: > So is Pulsar's testing going to switch to using JUnit 5 and remove testng？

@falser101 Perhaps eventually, but that will take a very long time. Instead of having the goal of converting from TestNG to JUnit5, I believe that it would be more valuable to first provide ways to create JUnit5 tests. There's not much value in doing a 1-to-1 mapping from TestNG to JUnit5.

falser101: > > So is Pulsar's testing going to switch to using JUnit 5 and remove testng？
> 
> @falser101 Perhaps eventually, but that will take a very long time. Instead of having the goal of converting from TestNG to JUnit5, I believe that it would be more valuable to first provide ways to create JUnit5 tests. There's not much value in doing a 1-to-1 mapping from TestNG to JUnit5.

yes, maybe we can commit the first JUnit5 test case

heesung-sn: I realized this pip is too much ambitious, and I am not sure it is practical any more. I am thinking of closing this PIP.

PR #23577, opened 2024-11-07
[fix][client] Fix DLQ producer name conflicts when there are same name consumers

geniusjoe:
Main Issue:  https://github.com/apache/pulsar-client-go/issues/1297
Relative #21890 

### Motivation
When multiple consumers with same name are using dlq policies at the same time, prompt will throw below exceptions:
`
ERRO[0069] Failed to create DLQ producer  dlq-topic="persistent://<tenant>/<namespace>/<topic>-<subscription>-DLQ" error="server error: ProducerBusy: org.apache.pulsar.broker.service.BrokerServiceException$NamingException: Producer with name '-<subscription>-<consumerName>-DLQ' is already connected to topic"
`

### Modifications

I think maybe we need to add a random suffix in dlq producer name like: `opt.Name = fmt.Sprintf("%s-%s-%s-%s-DLQ", r.topicName, r.subscriptionName, r.consumerName, generateRandomName()) `

### Verifying this change

This change is already covered by existing tests, such as `DeadLetterTopicTest.testDeadLetterTopicWithProducerName`

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

After change golang, java and c++ sdks, we could update document like this pr https://github.com/apache/pulsar-site/pull/772/files.

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -98,6 +98,7 @@ import org.apache.pulsar.client.impl.metrics.UpDownCounter;
 import org.apache.pulsar.client.impl.schema.AutoConsumeSchema;
 import org.apache.pulsar.client.impl.transaction.TransactionImpl;
 import org.apache.pulsar.client.util.ExecutorProvider;
+import org.apache.pulsar.client.util.NameUtil;
 import org.apache.pulsar.client.util.RetryMessageUtil;
 import org.apache.pulsar.common.allocator.PulsarByteBufAllocator;
 import org.apache.pulsar.common.api.EncryptionContext;
@@ -2263,8 +2264,8 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
                             ((ProducerBuilderImpl<byte[]>) client.newProducer(Schema.AUTO_PRODUCE_BYTES(schema)))
                                     .initialSubscriptionName(this.deadLetterPolicy.getInitialSubscriptionName())
                                     .topic(this.deadLetterPolicy.getDeadLetterTopic())
-                                    .producerName(String.format("%s-%s-%s-DLQ", this.topicName, this.subscription,
-                                            this.consumerName))
+                                    .producerName(String.format("%s-%s-%s-%s-DLQ", this.topicName, this.subscription,
+                                            this.consumerName, NameUtil.generateRandomName()))
                                     .blockIfQueueFull(false)
                                     .enableBatching(false)
                                     .enableChunking(true)
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/util/NameUtil.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/util/NameUtil.java
@@ -21,7 +21,7 @@ package org.apache.pulsar.client.util;
 import java.util.UUID;
 import org.apache.commons.codec.digest.DigestUtils;
 
-public class ConsumerName {
+public class NameUtil {
     public static String generateRandomName() {
         return DigestUtils.sha1Hex(UUID.randomUUID().toString()).substring(0, 5);
     }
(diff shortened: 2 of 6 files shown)


hanmz:
Can you add a test to verify this case that the consumer name is specified?

zjxxzjwang (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2263,8 +2264,8 @@ private void initDeadLetterProducerIfNeeded() {
                             ((ProducerBuilderImpl<byte[]>) client.newProducer(Schema.AUTO_PRODUCE_BYTES(schema)))
                                     .initialSubscriptionName(this.deadLetterPolicy.getInitialSubscriptionName())
                                     .topic(this.deadLetterPolicy.getDeadLetterTopic())
-                                    .producerName(String.format("%s-%s-%s-DLQ", this.topicName, this.subscription,
-                                            this.consumerName))
+                                    .producerName(String.format("%s-%s-%s-%s-DLQ", this.topicName, this.subscription,
+                                            this.consumerName, ConsumerName.generateRandomName()))
```
RobertIndie, line 2268: Using the `ConsumerName` util class to generate a suffix for the producer name seems confusing.

geniusjoe: I rename util class from `pulsar.client.util.ConsumerName` to `pulsar.client.util.NameUtil`, please review again.

RobertIndie (approved):

Jason918 (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/util/NameUtil.java
```diff
@@ -21,7 +21,7 @@
 import java.util.UUID;
 import org.apache.commons.codec.digest.DigestUtils;
 
-public class ConsumerName {
+public class NameUtil {
     public static String generateRandomName() {
         return DigestUtils.sha1Hex(UUID.randomUUID().toString()).substring(0, 5);
     }
```
lhotari, line 27: why was this class renamed? This is not a generic method that we'd like to reuse more broadly. The original method `generateRandomName` is not even describing the name properly.

lhotari: Now I see the reason of renaming the class, https://github.com/apache/pulsar/pull/23577#discussion_r1846456815 .
I'd suggest removing this class completely in a follow up PR and using Commons Lang3's `org.apache.commons.lang3.RandomStringUtils#randomAscii(int)` method with the parameter `5` to create a random string of 5 characters. @geniusjoe Would you like to submit a PR for that?

geniusjoe: > Now I see the reason of renaming the class, [#23577 (comment)](https://github.com/apache/pulsar/pull/23577#discussion_r1846456815) . I'd suggest removing this class completely in a follow up PR and using Commons Lang3's `org.apache.commons.lang3.RandomStringUtils#randomAscii(int)` method with the parameter `5` to create a random string of 5 characters. @geniusjoe Would you like to submit a PR for that?

@lhotari  Thank you for reviewing my pr. This `generateRandomName()` only use in just few cases and I think it's better to use `randomAscii()` you mentioned to replace this class. I'm glad to submit a new PR ASAP.

geniusjoe: @lhotari 
Hello Lari I find that `randomAscii(int)` will return a string which also contains symbols and signs, example below:
```
~ # for ((i=32;i<127;i++)) do printf "\\$(printf %03o "$i")"; done;printf "\n"                      root@VM-152-173-centos
 !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~
```
The original `generateRandomName()` function uses lower-case UUID substring (i.e. uses 0-9 and a-f), so that it might be a break changes when we replace this function with `randomAscii(int)`. Shall we choose another random string generator function?

lhotari: @geniusjoe good point. instead of `randomAscii`, we could use `randomAlphanumeric` in `RandomStringUtils`.

lhotari:
This doesn't apply to 3.0.x since it doesn't include #21589

Issue #19864, opened 2023-03-20
[Bug] 


congbobo184:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Version

1

### Minimal reproduce step

1

### What did you expect to see?

1

### What did you see instead?

1

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

Issue #15189, opened 2022-04-17
Failover subscription sends first message to multiple consumers


gvolpe:
**Describe the bug**

I have two consumers connected using a `Failover` subscription to the same topic. I first run one consumer, and after a few minutes, I run the second consumer (no messages are being produced yet).

Once both consumers are up, I start publishing messages to the topic. When this happens, **both** consumers get the first message, and from there, only the selected *active consumer* gets the subsequent messages.

It looks like a bug to me, but if this is expected, I would appreciate if you could point me out to any kind of documentation in this regard.

**To Reproduce**

1. Create a `persistent` non-partitioned topic.
2. Create two consumers in `Failover` mode and run them.
3. Create a producer and start publishing messages to the topic.
4. See first message being received by both consumers.

**Expected behavior**

Only the active consumer should get **all** messages.

**Desktop (please complete the following information):**

 - OS: Linux (NixOS)
 
**Additional context**

Here are a few things that might be relevant. 

Pulsar version: `2.10.0`
Pulsar Java client: `2.10.0`

I have deduplication enabled on my broker configuration:

```
brokerDeduplicationEnabled=true
```

Messages published to the topic have an `orderingKey`, as I have other service running in `KeyShared` subscription mode reading the very same messages.

Technoboy-:
Hi, does the topic has partitions?
Below is the explanation of `failover`:
```
/**
* Multiple consumer will be able to use the same subscription name but only 1 consumer will receive the messages.
* If that consumer disconnects, one of the other connected consumers will start receiving messages.
*
* <p>In failover mode, the consumption ordering is guaranteed.
*
 * <p>In case of partitioned topics, the ordering is guaranteed on a per-partition basis.
* The partitions assignments will be split across the available consumers. On each partition,
* at most one consumer will be active at a given point in time.
*/
```

gvolpe:
@Technoboy- no, it's not a partitioned topic. I have updated the issue details to clarify that.

Technoboy-:
> @Technoboy- no, it's not a partitioned topic. I have updated the issue details to clarify that.

Could the below test match the reproduce steps ?
```
@Test
    public void testFailoverConsumer() throws Exception {
        TopicName topicName = TopicName.get("persistent://my-property/my-ns/testFailoverConsumer");
        @Cleanup
        Consumer<byte[]> consumer1 = pulsarClient.newConsumer()
                .topic(topicName.toString())
                .consumerName("aaa")
[...]

PR #23583, opened 2024-11-11
[improve][pip] PIP-392: Enable consistent hashing to select active consumer in partitioned topic for failover subscription

shibd:
### Motivation

- Discuss thread: https://lists.apache.org/thread/scfj13jh167lyx16g2tkwg1rtj8dt7nv
- Vote thread: https://lists.apache.org/thread/vvytpb2tlovmdrkryyg09flqtv1k5dbr
- Draft PR: https://github.com/apache/pulsar/pull/23584

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-392.md
+++ b/pip/pip-392.md
@@ -13,16 +13,65 @@ Consider a scenario with 100 topics named `public/default/topic-{0~100}`, each h
 If 10 consumers are created using a `regex` subscription with the `Failover type`, all topic will be assigned to the same consumer(the first connected consumer). This results in an imbalanced distribution of consumers.
 
 # Goals
-Introduce a configuration option to enable consistent hashing for selecting active consumers in partitioned topics with the Failover subscription type.
+
+## In Scope
+- Address the issue of imbalance for `failover` subscription type consumers in single-partition or few-partition topics.
+
+## Out of Scope
+- Excluding the `exclusive` subscription type.
 
 
 # High Level Design
 The solution involves adding a configuration setting that allows users to enable consistent hashing for partitioned topics.
-When enabled, the consumer selection process will use consistent hashing instead of the modulo operation, leading to a more even distribution of active consumers.
+When enabled, the consumer selection process will use consistent hashing instead of the modulo operation.
+
+The algorithm already exists through [#19502](https://github.com/apache/pulsar/pull/19502
+
+In simple terms, the hash algorithm includes the following steps:
+
+1. Hash Ring Creation: Traverse all consumers and use `consumer name` to calculate a hash ring with 100 virtual nodes.
+[Exist code](https://github.com/apache/pulsar/blob/1b1bd4b610dd768a6908964ef841a6790bb0f4f0/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L152-L162)
+```java
+    private NavigableMap<Integer, Integer> makeHashRing(int consumerSize) {
+        NavigableMap<Integer, Integer> hashRing = new TreeMap<>();
+        for (int i = 0; i < consumerSize; i++) {
+            for (int j = 0; j < CONSUMER_CONSISTENT_HASH_REPLICAS; j++) {
+                String key = consumers.get(i).consumerName() + j;
+                int hash = Murmur3_32Hash.getInstance().makeHash(key.getBytes());
+                hashRing.put(hash, i);
+            }
+        }
+        return Collections.unmodifiableNavigableMap(hashRing);
+    }
+```
+
+2. Consumer Selection: Use the hash of the topic name to select the matching consumer from the hash ring.
+[Exist code](https://github.com/apache/pulsar/blob/1b1bd4b610dd768a6908964ef841a6790bb0f4f0/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L146-L150)
+```java
+    private int peekConsumerIndexFromHashRing(NavigableMap<Integer, Integer> hashRing) {
+        int hash = Murmur3Hash32.getInstance().makeHash(topicName);
+        Map.Entry<Integer, Integer> ceilingEntry = hashRing.ceilingEntry(hash);
+        return ceilingEntry != null ? ceilingEntry.getValue() : hashRing.firstEntry().getValue();
+    }
+```
+
+This approach ensures a more even distribution of active consumers across topics, improving load balancing and resource utilization.
 
 # Detailed Design
 
 ## Design & Implementation Details
+Refer to implementation PR: https://github.com/apache/pulsar/pull/23584
+
+The implementation is simple. If this activeConsumerFailoverConsistentHashing is enabled, the consistent hashing algorithm is used regardless of whether the topic is partitioned.
+
+## Public-facing Changes
+
+If activeConsumerFailoverConsistentHashing is enabled, when users use the failover subscription model,
+the `first consumer` will not necessarily consume `P1`, and the `second consumer` will not necessarily consume `P2`. 
+
+As described in the documentation:: https://pulsar.apache.org/docs/4.0.x/concepts-messaging/#failover--partitioned-topics
+
+Instead, the hash algorithm will determine which consumer consumes which partition.
 
 ### Configuration
 
--- a/pip/pip-392.md
+++ b/pip/pip-392.md
@@ -1,0 +1,43 @@
+# PIP-392: Add configuration to enable consistent hashing to select active consumer for partitioned topic
+
+# Background knowledge
+
+After [#19502](https://github.com/apache/pulsar/pull/19502) will use consistent hashing to select active consumer for non-partitioned topic
+
+# Motivation
+
+Currently, for partitioned topics, the active consumer is selected using the formula [partitionedIndex % consumerSize](https://github.com/apache/pulsar/blob/137df29f85798b00de75460a1acb91c7bc25453f/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L129-L130). 
+This method can lead to uneven distribution of active consumers.
+
+Consider a scenario with 100 topics named `public/default/topic-{0~100}`, each having `one partition`. 
+If 10 consumers are created using a `regex` subscription with the `Failover type`, all topic will be assigned to the same consumer(the first connected consumer). This results in an imbalanced distribution of consumers.
+
+# Goals
+Introduce a configuration option to enable consistent hashing for selecting active consumers in partitioned topics with the Failover subscription type.
+
+
+# High Level Design
+The solution involves adding a configuration setting that allows users to enable consistent hashing for partitioned topics.
+When enabled, the consumer selection process will use consistent hashing instead of the modulo operation, leading to a more even distribution of active consumers.
+
+# Detailed Design
+
+## Design & Implementation Details
+
+### Configuration
+
+A new configuration field will be added:
+
+```java
+@FieldContext(
+    category = CATEGORY_POLICIES,
+    doc = "Enable consistent hashing for selecting the active consumer in partitioned "
+        + "topics with Failover subscription type. "
+        + "For non-partitioned topics, consistent hashing is used by default."
+)
+private boolean activeConsumerFailoverConsistentHashing = false;
+```
+
+
+# Backward & Forward Compatibility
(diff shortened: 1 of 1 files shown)


pip/pip-392.md
```diff
@@ -0,0 +1,43 @@
...
+Consider a scenario with 100 topics named `public/default/topic-{0~100}`, each having `one partition`. 
+If 10 consumers are created using a `regex` subscription with the `Failover type`, all topic will be assigned to the same consumer(the first connected consumer). This results in an imbalanced distribution of consumers.
+
+# Goals
+Introduce a configuration option to enable consistent hashing for selecting active consumers in partitioned topics with the Failover subscription type.
+
+
+# High Level Design
+The solution involves adding a configuration setting that allows users to enable consistent hashing for partitioned topics.
+When enabled, the consumer selection process will use consistent hashing instead of the modulo operation, leading to a more even distribution of active consumers.
```
lhotari, line 21: Please describe how the consistent hashing will be used to select the consumer.

pip/pip-392.md
```diff
@@ -0,0 +1,43 @@
...
+# Motivation
+
+Currently, for partitioned topics, the active consumer is selected using the formula [partitionedIndex % consumerSize](https://github.com/apache/pulsar/blob/137df29f85798b00de75460a1acb91c7bc25453f/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L129-L130). 
+This method can lead to uneven distribution of active consumers.
+
+Consider a scenario with 100 topics named `public/default/topic-{0~100}`, each having `one partition`. 
+If 10 consumers are created using a `regex` subscription with the `Failover type`, all topic will be assigned to the same consumer(the first connected consumer). This results in an imbalanced distribution of consumers.
+
+# Goals
+Introduce a configuration option to enable consistent hashing for selecting active consumers in partitioned topics with the Failover subscription type.
```
lhotari, line 16: Please describe the desired outcome. The solution isn't a goal. I guess the actual goal is to solve the imbalanced distribution of consumers for partitioned topics?

pip/pip-392.md
lhotari: Also describe why and how the leads to more even distribution of active consumers. 
What is the expected accuracy of the balancing? What factors impact it?

pip/pip-392.md
```diff
@@ -0,0 +1,94 @@
...
+Currently, for partitioned topics, the active consumer is selected using the formula [partitionedIndex % consumerSize](https://github.com/apache/pulsar/blob/137df29f85798b00de75460a1acb91c7bc25453f/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L129-L130). 
+This method can lead to uneven distribution of active consumers.
+
+Consider a scenario with 100 topics named `public/default/topic-{0~100}`, each having `one partition`. 
+If 10 consumers are created using a `regex` subscription with the `Failover type`, all topic will be assigned to the same consumer(the first connected consumer). This results in an imbalanced distribution of consumers.
+
+# Goals
+
+## In Scope
+- Address the issue of imbalance for `failover` subscription type consumers in single-partition or few-partition topics.
```
lhotari, line 18: This change will introduce a new problem: Let's say that there's a single connected consumer and another consumer connects, it could switch to use the new consumer. When this happens, there's an existing problem with failover and exclusive consumers where the outstanding messages aren't properly drained before switching. 
There's a stalled PIP, [PIP-260: Client consumer filter received messages](https://github.com/apache/pulsar/issues/19864), which is related to the issue, however it's not exactly describing the consumer switching scenario. For non-partitioned topics there's a reported issue #15189. After PIP-392 changes, a similar problem would also appear for partitioned topics.
I guess addressing #15189 could happen in a separate PIP, but I'd suggest documenting this detail also in the PIP-392 document that it's a known consequence of the change.
For consumer switching in failover subscriptions, there should be a way to ensure that pending messages are processed before switching the consumer. Otherwise it leads to duplicate message processing in clients which could be considered as out-of-order message processing from the client application perspective.

shibd: hi, @lhotari Thanks for bringing up this topic. 

Here's my conclusion: **Pulsar current cannot avoid duplicate messages on the client side during broker or active consumer transfer scenarios, and this PIP doesn't introduce new issues because both consistent hashing and modulo algorithms will cause active consumer transfer.**

Please help review my explanation.


### For broker ownership transfer case: https://github.com/apache/pulsar/issues/19864
This is a known case, so let's not discuss it too much here.


### For active consumer transfer case(Failover subscription)

#### 1. Use consistent hashing algorithm

Before this PIP, we only used the consistent hashing algorithm with `non-partition` topic.

Since the consistent hashing algorithm creates a hash ring based on the consumer's name, when a new consumer joins, new nodes are added to the hash ring, causing topics to move to this consumer. Thus, topics are transferred.

Refer to the image: The `topic-3` transfers from `c2` to `c4`.

image

This issue exists with non-partitioned topics because they use consistent hashing. I agree with your point that after this PIP, it will also occur with partitioned topics.

However, I want to emphasize that this issue can occur even without consistent hashing for `partitioned topic`. Please see the explanation below.


#### 2. Use mod(partitionedIndex % consumerSize)   - **current partitioned topic algorithm**

Assume a topic has 2 partition: **partitionedIndex [P0, P1]**

1. First consumer added(**consumer-0**)
  - **P0** belong **consumer-0**: (paritonIndex:0)%(consumerSize:1)=0, 
  - **P1** belong  **consumer-0**: (paritonIndex:1)%(consumerSize:1)=0, 
2. When secnd consumer is added(consumer-1) 
  - **P0** belong **consumer-0**: (paritonIndex:0)%(consumerSize:2)=0, 
  - **P1** belong  **consumer-1**: (paritonIndex:1)%(consumerSize:2)=1, 

P1 activeConsumer from consumer-0 transfer to consumer-1.

P1's active consumer transfers from consumer-0 to consumer-1.

Therefore, P1 undergoes an active consumer transfer.

### So go back this comment
> After PIP-392 changes, a similar problem would also appear for partitioned topics.
I guess addressing https://github.com/apache/pulsar/issues/15189 could happen in a separate PIP, but I'd suggest documenting this detail also in the PIP-392 document that it's a known consequence of the change.

I don't think it's necessary to add more explanation for this.

1. #15189 does not have a clear conclusion proving that the consistent hash algorithm for non-partitioned topics has issues.
2. Regarding potential duplicate messages and out-of-order behavior in failover, we have already explained this in the documentation: https://pulsar.apache.org/docs/4.0.x/concepts-messaging/#failover"

lhotari: > Here's my conclusion: **Pulsar current cannot avoid duplicate messages on the client side during broker or active consumer transfer scenarios, and this PIP doesn't introduce new issues because both consistent hashing and modulo algorithms will cause active consumer transfer.**

The way how this PR introduces more frequent problems is that the consumer will switch frequently when new consumers get added. A rolling restart is a common problematic case. This is why the PIP document should at least document that this is a known problem and consequence of this change.

I hope that @codelipenghui and @merlimat could share their view on this matter.

shibd: > The way how this PR introduces more frequent problems is that the consumer will switch frequently when new consumers get added. A rolling restart is a common problematic case. This is why the PIP document should at least document that this is a known problem and consequence of this change.

If I understand correctly, the modulo algorithm also has this situation.

Assuming there are 5 partitions with indices partitionIndex: [0, 1, 2, 3, 4]

#### 1. The first consumer-0 joins, partitions selected:
  - consumer-0: [0, 1, 2, 3, 4]
```
partitionIndex % consumerSize = consumerIndex
0 % 1 = 0
1 % 1 = 0
2 % 1 = 0
3 % 1 = 0
4 % 1 = 0
```

#### 2. When the second consumer-1 joins, partitions selected:
   - consumer-0: [0, 2, 4]  
   - consumer-1: [1, 3]
```
partitionIndex % consumerSize = consumerIndex
0 % 2 = 0
1 % 2 = 1
2 % 2 = 0
3 % 2 = 1
4 % 2 = 0
```
`!!`**!!P1 and P3** also get transferred

#### 3. When the third consumer-2 joins, partitions selected:
   - consumer-0: [0, 3]  
   - consumer-1: [1, 4]
   - consumer-2: [2]  
```
partitionIndex % consumerSize = consumerIndex
0 % 3 = 0
1 % 3 = 1
2 % 3 = 2
3 % 3 = 0
4 % 3 = 1
```
`!!`**P2, P3 and P4** also get transferred

Altogether transferred

1. P1 from c0 -> c1
2. P2 from c0 -> c2
3. P3 from c0 -> c1 -> c2
4. P4 from c0 -> c2

lhotari: @shibd Thanks for clarifying. Makes sense. PIP-392 won't make the problem worse in that case. One of my assumptions was that in the current solution there's "stickyness" in a way that the active consumer doesn't get switched unless it fails. I guess that's a false assumption from my side. 

Although it's out-of-scope for PIP-392, it would be great to start making progress in addressing the clear problem with Failover subscriptions which is documented as "In some cases, a partition may have an older active consumer processing messages while a newly switched over active consumer starts receiving new messages. This may lead to message duplication or out-of-order.". That goes against a lot of the expected guarantees for Failover subscriptions and causes problems. I found out that @grssam provided some feedback about this in the past in https://github.com/apache/pulsar-site/pull/633#issuecomment-1625416461 . It could be valuable if @grssam could also review PIP-392.

shibd: hi, @lhotari I added this to `Out of Scope` section. PTAL.

lhotari: Thanks, this resolves this discussion.

pip/pip-392.md
```diff
@@ -0,0 +1,97 @@
...
+
+# High Level Design
+The solution involves adding a configuration setting that allows users to enable consistent hashing for partitioned topics.
+When enabled, the consumer selection process will use consistent hashing instead of the modulo operation.
+
+The algorithm already exists through [#19502](https://github.com/apache/pulsar/pull/19502)
+
+In simple terms, the hash algorithm includes the following steps:
+
+1. Hash Ring Creation: Traverse all consumers and use `consumer name` to calculate a hash ring with 100 virtual nodes.
```
lhotari, line 35: Have we documented the recommendation to always use unique consumer names on all consumers?

For Key_Shared subscriptions, there was a challenge with duplicate consumer names. It has been solved in Pulsar 4.0 [with a specific solution for handling duplicates](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ConsumerNameIndexTracker.java). 
The referenced code for consistent hashing in AbstractDispatcherSingleActiveConsumer doesn't properly handle duplicate consumer names. 

Will this PIP handle duplicate consumer names to meet the goal when all consumers have the same consumer name?

pip/pip-392.md
grssam: yea, i don't see hash based having a bigger impact on the rebalance that happens during things like consumer pod rollout etc. maybe the situation would actually be better as previously it was skewed (for multiple partitioned topics in a single sub) so impact was higher, but now based on how evenly things are spread, not the same amount of partitions would get rebalanced.

pip/pip-392.md
shibd: When building the hash ring, we include the index, so even if there are identical consumers for a partitioned topic, they will be on different hash segments.

https://github.com/apache/pulsar/blob/1b1bd4b610dd768a6908964ef841a6790bb0f4f0/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L156

lhotari: > When building the hash ring, we include the index, so even if there are same consumer name for a partitioned topic, they will be on different hash segments.

I don't see that happening in the code. `j` is the index of the hash ring replica point. Consumers with the same name would collide and the last one to connect would replace all hash ring points.

btw. this code has also another problem: if one consumer has a name `consumer` and another one `consumer1`, they would also collide. A similar issue has been reported and solved in the Key_Shared consistent hashing solution.

shibd: oh, You are right. Therefore, it's best to use a unique `consumerName` with consistent hashing; otherwise, multiple partition topics may select the same consumer, causing imbalance.

I will mention this in the documentation.

Technoboy- (approved):
LGTM

shibd:
Vote passed, I will merge this PIP.

PR #23589, opened 2024-11-11
[fix] [broker] Do not take effect pub&sub rate-limit for system topics

poorbarcode:
### Motivation

The effect is huge if the `pub&sub` rate limiter takes effect for system topics, such as `__change_events`, transaction topics... Almost all topics can not work if the system topic encounters a throttling

### Modifications
- Do not take effect pub&sub rate-limit for system topics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -283,6 +283,10 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
         if (log.isDebugEnabled()) {
             log.debug("[{}]updateTopicPolicyByNamespacePolicy,data={}", topic, namespacePolicies);
         }
+        if (!isSystemTopic()) {
+            updateNamespacePublishRate(namespacePolicies, brokerService.getPulsar().getConfig().getClusterName());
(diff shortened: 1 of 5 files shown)


lhotari:
The only possible concern in a multi-tenant system is that a malicious user could circumvent rate-limits by using a topic name that gets considered as a system topic. I guess this is a matter of the definition of the security model in the multi-tenancy support that rate limits cannot be used to protect against malicious users. We don't currently document the security model so it's hard to resolve this as part of this PR. I guess we could create a separate issue for documenting the security model for multi-tenancy.

lhotari (approved):
LGTM

gaoran10 (approved):
LGTM

poorbarcode:
> The only possible concern in a multi-tenant system is that a malicious user could circumvent rate limits by using a topic name that gets considered as a system topic. I guess this is a matter of the definition of the security model in the multi-tenancy support that rate limits cannot be used to protect against malicious users. We don't currently document the security model so it's hard to resolve this as part of this PR. I guess we could create a separate issue for documenting the security model for multi-tenancy.

Seems we never focused on the isolation for multi-tenants before, that should be a huge PIP, which contains Transaction(all namespaces use the same Transaction metadata store) and other components. 😂 

Thanks for mentioning this, let me merge the PR first

PR #23592, opened 2024-11-12
[feat][pip] PIP-402: Role Anonymizer for Pulsar Logging

KannarFr:
PIP for https://github.com/apache/pulsar/pull/23386

### DIFF
--- a/pip/pip-402-role-anonymize.md
+++ b/pip/pip-402-role-anonymize.md
@@ -1,0 +1,120 @@
+# PIP-402: Role Anonymizer for Pulsar Logging
+
+# Background knowledge
+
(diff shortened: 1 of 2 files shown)


pip/pip-385-role-anonmyzer.md
```diff
@@ -0,0 +1,120 @@
+# PIP-385: Role Anonymizer for Pulsar Logging
```
lhotari, line 1: PIP-385 is already taken. 

Please find another available PIP. The current way to find an available PIP is to search on the dev mailing list discussions whether there's a conflict with the picked PIP number. The first one to open the discussion about the PIP "wins".

You should also make the PR title match the way that we use 
`[feat][pip] PIP-XXX: Role Anonymizer for Pulsar Logging` would be the way to name the PR. Start the thread on the dev mailing list asap after assigning the PR number so that you "win" the possible race condition in assigning the PIP number.

lhotari (approved):
Vote passed https://lists.apache.org/thread/bj8ncrl8k9p6o9bkh3fl4nvmvjgo5h18

PR #13344, opened 2021-12-15
PIP-120 : Enable client memory limit controller by default

HQebupt:
Fixes #13306  
### Motivation
See #13306 

### Modifications
Enable the `MemeoryLimitController`. The `memoryLimitBytes` default is 64M which is easier to control than the `maxPendingMessages`. 

The default 64M has a good performance in produce throughput in our perf test. I am going to post the performance result here.

### Verifying this change

- [x]  Make sure that the change passes the CI checks.

This change is a trivial rework / code cleanup without any test coverage.

### Does this pull request potentially affect one of the following parts:
If `yes` was chosen, please highlight the changes

- Dependencies (does it add or upgrade a dependency): (no)
- The public API: (no)
- The schema: (no)
- The default values of configurations: (no)
- The wire protocol: (no)
- The rest endpoints: (no)
- The admin cli options: (no)
- Anything that affects deployment: (no)

### Documentation
Check the box below and label this PR (if you have committer privilege).

Need to update docs? 

  
- [x] `doc` 
  
  (There is description in the configuration modification)

PR #23593, opened 2024-11-12
[fix][client] The partitionedProducer maxPendingMessages always is 0

shibd:
### Motivation
After #13344, the default value of maxPendingMessages and maxPendingMessagesAcrossPartitions change to **0**.

When creating a producer with the partitioned topic, setting one of the above configurations doesn't take effect; the actual value is always 0.

Because here have a minimum judgment.

https://github.com/apache/pulsar/blob/92448d5f3113d783d41de6bae718d34cd73ca934/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PartitionedProducerImpl.java#L86-L89

### Modifications
- When setting one of the values, use it directly. Only perform the `min` check when both values are not zero.

### Verifying this change
- Add PartitionedProducerImplTest.testMaxPendingQueueSize to cover this logic.
- Add testProducerQueueFullBlockingWithPartitionedTopic test to cover blockQueueFullWithPartitiond case.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


lhotari (approved):
LGTM, good catch

poorbarcode (approved):

PR #23594, opened 2024-11-12
[fix] [broker] No longer allow creating subscription that contains slash

poorbarcode:
### Motivation

**Topic name rule**
- old version: `{tenant}/{cluster}/{namespace}/{topic}/{subscription}`
- new version: `{tenant}/{namespace}/{topic}/{subscription}`

There are many HTTP APIs defined as `HTTP Method {topic name}/{subscription name}`

If a subscription contains `/`, the broker will assume it is a topic created with the old rule, then users will get a `cluster does not exist` error or a `topic not found error`

### Modifications

Use the util method https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/naming/NamedEntity.java to check new subscription name

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -202,6 +202,10 @@ allowAutoTopicCreationType=non-partitioned
 # the topic cannot be automatically created.
 allowAutoTopicCreationWithLegacyNamingScheme=true
 
+# If 'strictSubscriptionNameVerification' is true, the new subscription name can only contain (a-zA-Z_0-9) and these
+# special chars -=:.
+strictlyVerifySubscriptionName=false
+
 # Enable subscription auto creation if new consumer connected (disable auto creation with value false)
 allowAutoSubscriptionCreation=true
 
--- a/conf/standalone.conf
+++ b/conf/standalone.conf
@@ -1224,6 +1224,10 @@ allowAutoTopicCreationType=non-partitioned
 # the topic cannot be automatically created.
 allowAutoTopicCreationWithLegacyNamingScheme=true
 
+# If 'strictSubscriptionNameVerification' is true, the new subscription name can only contain (a-zA-Z_0-9) and these
+# special chars -=:.
+strictlyVerifySubscriptionName=false
+
 # Enable subscription auto creation if new consumer connected (disable auto creation with value false)
 allowAutoSubscriptionCreation=true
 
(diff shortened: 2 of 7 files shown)


Technoboy- (approved):

gaoran10:
Do we need to support using URL-encoded subscription names in the HTTP request? Thus users can use existing URL-encoded subscription names in HTTP requests.

hanmz (approved):

zjxxzjwang (commented):
1

zjxxzjwang (approved):

zjxxzjwang:
Can you put it into a method to facilitate the extension of more subscription name restrictions in the future

Shawyeok:
> HTTP Method {topic name}/{subscription name}

The subscription name should be URL-encoded. For example, to create a subscription `a/b` for the topic `persistent://public/default/topic1`, the request should be:  
`PUT /admin/v2/persistent/public/default/topic1/subscription/a%252Fb`.

I believe `pulsar-admin` and `pulsarctl` already handle this properly.

I appreciate the approach of restricting naming conventions for subscription and topic names. In my organization, subscription names are limited to the following characters: `a-zA-Z0-9.-_`.

lhotari:
Related PR: #23620

lhotari:
> I appreciate the approach of restricting naming conventions for subscription and topic names. In my organization, subscription names are limited to the following characters: `a-zA-Z0-9.-_`.

@Shawyeok There's an enhancement issue #23614 created by @alpreu to add validation to subscription names.

lhotari (changes requested):
It looks like PR #23620 could be an effective way to prevent the auto creation of topics with slashes.

I think that we should consider implementing #23614 to validate subscription names instead of preventing the creation of subscription names with slashes.

dao-jun (approved):

lhotari:
> Sent a discussion to do that
> 
> * https://lists.apache.org/thread/j8shr3wqr1xdv3rxpth2f53ctnzc1ny1

@poorbarcode I think it makes sense to reference https://github.com/apache/pulsar/issues/23614 and https://github.com/apache/pulsar/pull/23620 in the discussion since those describe useful solutions. They are addressing 2 separate issues.

poorbarcode:
@lhotari 

Used the util method https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/naming/NamedEntity.java to check new subscription name, please take a look

lhotari (approved):
LGTM

PR #23600, opened 2024-11-14
[improve][client] PIP-393: Improve performance of Negative Acknowledgement

thetumbled:
Implementation PR for PIP-393: https://github.com/apache/pulsar/pull/23601.
### Motivation

There are many issues with the current implementation of Negative Acknowledgement in Pulsar:
- the memory occupation is high.
- the code execution efficiency is low.
- the redelivery time is not accurate.
- multiple negative ack for messages in the same entry(batch) will interfere with each other.
All of these problem is severe and need to be solved.

#### Memory occupation is high
After the improvement of https://github.com/apache/pulsar/pull/23582, we have reduce half more memory occupation
of `NegativeAcksTracker` by replacing `HashMap` with `ConcurrentLongLongPairHashMap`. With 100w entry, the memory occupation decrease from 178Mb to 64Mb. With 1kw entry, the memory occupation decrease from 1132Mb to 512Mb.
The average memory occupation of each entry decrease from 1132MB/10000000=118byte to 512MB/10000000=53byte.

But it is not enough. Assuming that we negative ack message 1w/s, assigning 1h redelivery delay for each message,
the memory occupation of `NegativeAcksTracker` will be `3600*10000*53/1024/1024/1024=1.77GB`, if the delay is 5h,
the required memory is `3600*10000*53/1024/1024/1024*5=8.88GB`, which increase too fast.

#### Code execution efficiency is low
Currently, each time the timer task is triggered, it will iterate all the entries in `NegativeAcksTracker.nackedMessages`, 
which is unnecessary. We can sort entries by timestamp and only iterate the entries that need to be redelivered.

#### Redelivery time is not accurate
Currently, the redelivery time is controlled by the `timerIntervalNanos`, which is 1/3 of the `negativeAckRedeliveryDelay`.
That means, if the `negativeAckRedeliveryDelay` is 1h, the check interval time will be 20min, which is unacceptable.

#### Multiple negative ack for messages in the same entry(batch) will interfere with each other
Currently, `NegativeAcksTracker#nackedMessages` map `(ledgerId, entryId)` to `timestamp`, which means multiple nacks from messages in the same batch share single one timestamp. 
If we let msg1 redelivered 10s later, then let msg2 redelivered 20s later, these two messages are delivered 20s later together. msg1 will not be redelivered 10s later as the timestamp recorded in `NegativeAcksTracker#nackedMessages` is overrode by the second nack call.

we can reproduce this problem with test code below:
```
Consumer consumer = client.newConsumer()
                .topic("persistent://public/default/testNack")
                .subscriptionName("sub2")
                .negativeAckRedeliveryDelay(20, TimeUnit.SECONDS) // fixed delay with 20s.
                .subscribe();
        // receive first message and nack it.
        Message msg = consumer.receive();
        MessageIdAdv batchMessageId = (MessageIdAdv) msg.getMessageId();
        int batchIndex = batchMessageId.getBatchIndex();
        log.info("Message received, timestamp:{}, message id:{}, batch index:{}", getTime(), batchMessageId, batchIndex);
        consumer.negativeAcknowledge(msg);
        
        // receive the secode message and sleep for 10s, then nack it.
        msg = consumer.receive();
        batchMessageId = (MessageIdAdv) msg.getMessageId();
        batchIndex = batchMessageId.getBatchIndex();
        log.info("Message received, timestamp:{}, message id:{}, batch index:{}", getTime(), batchMessageId, batchIndex);
        Thread.sleep(10000);
        consumer.negativeAcknowledge(msg);
```
We expect the second message redelivered 10s later than the first message, as it call nack 10s later than the first one.
However, we will receive two messages together.
image

You can also reproduce this problem with the test code in this PR: org.apache.pulsar.client.impl.NegativeAcksTest#testNegativeAcksWithBatch

### Modifications

Refactor the `NegativeAcksTracker` to solve the above problems.

### Verifying this change

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/64

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2586,6 +2586,7 @@ flexible messaging model and an intuitive client API.</description>
         <module>pulsar-metadata</module>
         <module>jetcd-core-shaded</module>
         <module>jclouds-shaded</module>
+        <module>pulsar-client-dependencies-minimized</module>
 
         <!-- package management releated modules (begin) -->
         <module>pulsar-package-management</module>
@@ -2651,6 +2652,7 @@ flexible messaging model and an intuitive client API.</description>
         <module>distribution</module>
         <module>pulsar-metadata</module>
         <module>jetcd-core-shaded</module>
+        <module>pulsar-client-dependencies-minimized</module>
         <!-- package management releated modules (begin) -->
         <module>pulsar-package-management</module>
         <!-- package management releated modules (end) -->
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
@@ -19,6 +19,7 @@
 package org.apache.pulsar.client.impl;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertTrue;
 import java.util.HashSet;
@@ -311,7 +312,7 @@ public class NegativeAcksTest extends ProducerConsumerBase {
         // negative topic message id
         consumer.negativeAcknowledge(topicMessageId);
         NegativeAcksTracker negativeAcksTracker = consumer.getNegativeAcksTracker();
-        assertEquals(negativeAcksTracker.getNackedMessagesCount().orElse((long) -1).longValue(), 1L);
+        assertEquals(negativeAcksTracker.getNackedMessagesCount(), 1L);
         assertEquals(unAckedMessageTracker.size(), 0);
         negativeAcksTracker.close();
         // negative batch message id
@@ -319,9 +320,54 @@ public class NegativeAcksTest extends ProducerConsumerBase {
         consumer.negativeAcknowledge(batchMessageId);
         consumer.negativeAcknowledge(batchMessageId2);
         consumer.negativeAcknowledge(batchMessageId3);
-        assertEquals(negativeAcksTracker.getNackedMessagesCount().orElse((long) -1).longValue(), 1L);
+        assertEquals(negativeAcksTracker.getNackedMessagesCount(), 1L);
         assertEquals(unAckedMessageTracker.size(), 0);
         negativeAcksTracker.close();
+    }
+
+    /**
+     * If we nack multiple messages in the same batch with different redelivery delays, the messages should be redelivered
+     * with the correct delay. However, all messages are redelivered at the same time.
+     * @throws Exception
+     */
+    @Test
+    public void testNegativeAcksWithBatch() throws Exception {
+        cleanup();
+        conf.setAcknowledgmentAtBatchIndexLevelEnabled(true);
+        setup();
+        String topic = BrokerTestUtil.newUniqueName("testNegativeAcksWithBatch");
+
+        @Cleanup
+        Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING)
+                .topic(topic)
+                .subscriptionName("sub1")
+                .acknowledgmentGroupTime(0, TimeUnit.SECONDS)
+                .subscriptionType(SubscriptionType.Shared)
+                .enableBatchIndexAcknowledgment(true)
+                .negativeAckRedeliveryDelay(3, TimeUnit.SECONDS)
+                .subscribe();
+
+        @Cleanup
+        Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
+                .topic(topic)
+                .enableBatching(true)
+                .batchingMaxPublishDelay(1, TimeUnit.HOURS)
+                .batchingMaxMessages(2)
+                .create();
+        // send two messages in the same batch
+        producer.sendAsync("test-0");
+        producer.sendAsync("test-1");
+        producer.flush();
+
+        // negative ack the first message
+        consumer.negativeAcknowledge(consumer.receive());
+        // wait for 2s, negative ack the second message
+        Thread.sleep(2000);
+        consumer.negativeAcknowledge(consumer.receive());
+
+        // now 2s has passed, the first message should be redelivered 1s later.
+        Message<String> msg1 = consumer.receive(2, TimeUnit.SECONDS);
+        assertNotNull(msg1);
     }
 
     @Test
--- a/pulsar-client-all/pom.xml
+++ b/pulsar-client-all/pom.xml
@@ -212,6 +212,7 @@
                   <include>org.tukaani:xz</include>
                   <include>org.yaml:snakeyaml</include>
                   <include>org.apache.pulsar:pulsar-client-dependencies-minimized</include>
+                  <include>org.roaringbitmap:RoaringBitmap</include>
                 </includes>
                 <excludes>
                   <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
@@ -378,6 +379,11 @@
                   <rawString>true</rawString>
                 </relocation>
                 <relocation>
+                  <pattern>META-INF/versions/(\d+)/org/roaringbitmap/</pattern>
+                  <shadedPattern>META-INF/versions/$1/org/apache/pulsar/shade/org/roaringbitmap/</shadedPattern>
+                  <rawString>true</rawString>
+                </relocation>
+                <relocation>
                   <pattern>META-INF/versions/(\d+)/org/yaml/</pattern>
                   <shadedPattern>META-INF/versions/$1/org/apache/pulsar/shade/org/yaml/</shadedPattern>
                   <rawString>true</rawString>
@@ -454,6 +460,10 @@
                 <relocation>
                   <pattern>org.reactivestreams</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.reactivestreams</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>org.roaringbitmap</pattern>
+                  <shadedPattern>org.apache.pulsar.shade.org.roaringbitmap</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>org.tukaani</pattern>
--- a/pulsar-client-all/pom.xml
+++ b/pulsar-client-all/pom.xml
@@ -39,6 +39,17 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>pulsar-client-original</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>it.unimi.dsi</groupId>
+          <artifactId>fastutil</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>pulsar-client-dependencies-minimized</artifactId>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
@@ -200,7 +211,7 @@
                   <include>org.reactivestreams:reactive-streams</include>
                   <include>org.tukaani:xz</include>
                   <include>org.yaml:snakeyaml</include>
-                  <include>it.unimi.dsi:fastutil</include>
+                  <include>org.apache.pulsar:pulsar-client-dependencies-minimized</include>
                 </includes>
                 <excludes>
                   <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
@@ -317,6 +328,10 @@
                 <relocation>
                   <pattern>io.swagger</pattern>
                   <shadedPattern>org.apache.pulsar.shade.io.swagger</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>it.unimi.dsi.fastutil</pattern>
+                  <shadedPattern>org.apache.pulsar.shade.it.unimi.dsi.fastutil</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>javassist</pattern>
--- a/pulsar-client-all/pom.xml
+++ b/pulsar-client-all/pom.xml
@@ -200,6 +200,7 @@
                   <include>org.reactivestreams:reactive-streams</include>
                   <include>org.tukaani:xz</include>
                   <include>org.yaml:snakeyaml</include>
+                  <include>it.unimi.dsi:fastutil</include>
                 </includes>
                 <excludes>
                   <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
--- a/pulsar-client-shaded/pom.xml
+++ b/pulsar-client-shaded/pom.xml
@@ -176,6 +176,7 @@
                   <include>org.tukaani:xz</include>
                   <include>org.yaml:snakeyaml</include>
                   <include>org.apache.pulsar:pulsar-client-dependencies-minimized</include>
+                  <include>org.roaringbitmap:RoaringBitmap</include>
                 </includes>
                 <excludes>
                   <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
@@ -298,6 +299,11 @@
                   <rawString>true</rawString>
                 </relocation>
                 <relocation>
+                  <pattern>META-INF/versions/(\d+)/org/roaringbitmap/</pattern>
+                  <shadedPattern>META-INF/versions/$1/org/apache/pulsar/shade/org/roaringbitmap/</shadedPattern>
+                  <rawString>true</rawString>
+                </relocation>
+                <relocation>
                   <pattern>META-INF/versions/(\d+)/org/yaml/</pattern>
                   <shadedPattern>META-INF/versions/$1/org/apache/pulsar/shade/org/yaml/</shadedPattern>
                   <rawString>true</rawString>
@@ -358,6 +364,10 @@
                 <relocation>
                   <pattern>org.reactivestreams</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.reactivestreams</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>org.roaringbitmap</pattern>
+                  <shadedPattern>org.apache.pulsar.shade.org.roaringbitmap</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>org.tukaani</pattern>
--- a/pulsar-client-shaded/pom.xml
+++ b/pulsar-client-shaded/pom.xml
@@ -39,6 +39,17 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>pulsar-client-original</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>it.unimi.dsi</groupId>
+          <artifactId>fastutil</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>pulsar-client-dependencies-minimized</artifactId>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
@@ -164,7 +175,7 @@
                   <include>org.reactivestreams:reactive-streams</include>
                   <include>org.tukaani:xz</include>
                   <include>org.yaml:snakeyaml</include>
-                  <include>it.unimi.dsi:fastutil</include>
+                  <include>org.apache.pulsar:pulsar-client-dependencies-minimized</include>
                 </includes>
                 <excludes>
                   <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
@@ -263,6 +274,10 @@
                 <relocation>
                   <pattern>io.swagger</pattern>
                   <shadedPattern>org.apache.pulsar.shade.io.swagger</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>it.unimi.dsi.fastutil</pattern>
+                  <shadedPattern>org.apache.pulsar.shade.it.unimi.dsi.fastutil</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>javax.activation</pattern>
--- a/pulsar-client-shaded/pom.xml
+++ b/pulsar-client-shaded/pom.xml
@@ -164,6 +164,7 @@
                   <include>org.reactivestreams:reactive-streams</include>
                   <include>org.tukaani:xz</include>
                   <include>org.yaml:snakeyaml</include>
+                  <include>it.unimi.dsi:fastutil</include>
                 </includes>
                 <excludes>
                   <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
--- a/pulsar-client/pom.xml
+++ b/pulsar-client/pom.xml
@@ -207,6 +207,16 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.roaringbitmap</groupId>
+      <artifactId>RoaringBitmap</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>it.unimi.dsi</groupId>
+      <artifactId>fastutil</artifactId>
+    </dependency>
+
   </dependencies>
 
   <build>
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
@@ -22,16 +22,15 @@ import static org.apache.pulsar.client.impl.UnAckedMessageTracker.addChunkedMess
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.util.Timeout;
 import io.netty.util.Timer;
+import it.unimi.dsi.fastutil.longs.Long2ObjectAVLTreeMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectSortedMap;
+import it.unimi.dsi.fastutil.longs.LongBidirectionalIterator;
 import java.io.Closeable;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-
-import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
-import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
-import it.unimi.dsi.fastutil.longs.Long2ObjectRBTreeMap;
-import it.unimi.dsi.fastutil.longs.Long2ObjectSortedMap;
-import it.unimi.dsi.fastutil.longs.LongBidirectionalIterator;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageIdAdv;
@@ -145,7 +144,7 @@ class NegativeAcksTracker implements Closeable {
 
     private synchronized void add(MessageId messageId, int redeliveryCount) {
         if (nackedMessages == null) {
-            nackedMessages = new Long2ObjectRBTreeMap<>();
+            nackedMessages = new Long2ObjectAVLTreeMap<>();
         }
 
         long backoffMs;
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
@@ -24,48 +24,50 @@ import io.netty.util.Timeout;
 import io.netty.util.Timer;
 import java.io.Closeable;
 import java.util.HashSet;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+
+import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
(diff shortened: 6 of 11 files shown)


lhotari:
> I will implement a space efficient map structure for `ConcurrentTripleLong2LongHashMap` later, which for now use hashmap to ensure the logical correction.

@thetumbled Fastutil could be a better source of space efficient map data structures. I believe that there's a templating solution where it's possible to generate code for efficient implementations. In this case, is there a need for the data structure to be concurrent? Following a single writer principle could result in simpler and more performant designs. One way to address message passing from other threads to a single writer thread is to use message passing queues from JCTools which we already use in Pulsar. Just some food for thought.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentTripleLong2LongHashMap.java
```diff
@@ -0,0 +1,94 @@
...
+            if(obj instanceof TripleLong){
+                TripleLong other = (TripleLong) obj;
+                return first == other.first && second == other.second && third == other.third;
+            }
+            return false;
+        }
+    }
+
+    private HashMap<TripleLong, Long> map;
+    public ConcurrentTripleLong2LongHashMap(){
```
lhotari, line 43: Why does the name have the "Concurrent" prefix if this isn't supporting concurrent access?

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentTripleLong2LongHashMap.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+package org.apache.pulsar.common.util.collections;
+
+import java.util.HashMap;
+
+public class ConcurrentTripleLong2LongHashMap {
+    public class TripleLong{
+        public long first;
+        public long second;
+        public long third;
```
lhotari, line 27: Is this intentionally mutable?

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentTripleLong2LongHashMap.java
thetumbled: I use `HashMap` for now, which is not the final version.
I think we can implement a simillar concurrent map class like `ConcurrentLongLongPairHashMap`, each entry take up 4 long space and without any wrapper.

thetumbled:
> > I will implement a space efficient map structure for `ConcurrentTripleLong2LongHashMap` later, which for now use hashmap to ensure the logical correction.
> 
> @thetumbled Fastutil could be a better source of space efficient map data structures. I believe that there's a templating solution where it's possible to generate code for efficient implementations. In this case, is there a need for the data structure to be concurrent? Following a single writer principle could result in simpler and more performant designs. One way to address message passing from other threads to a single writer thread is to use message passing queues from JCTools which we already use in Pulsar. Just some food for thought.

If any solution from `Fastutil` prove to be more space efficient, i am glad to adopt it. Nack Tracker will consume enormous amount of memory, we have to choose the best one.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentTripleLong2LongHashMap.java
thetumbled: `TripleLong` will be removed after i implement the ConcurrentTripleLong2LongHashMap, as it use wrapper too.
Ideally, we will store the key and value in the array without any wrapper.

thetumbled: If FastUtil provide a TripleLong2LongMap, then it has the chance to be better than our implementation.
As the documentation of FastUtil point out, we should test the util class it provide before use it, there are high chance that it is not a good solution.

lhotari:
> If any solution from `Fastutil` prove to be more space efficient, i am glad to adopt it. Nack Tracker will consume enormous amount of memory, we have to choose the best one.

The main purpose of [Fastutil](https://fastutil.di.unimi.it/) is performance and space efficiency. 
Instead of adding very complex keys such as triple keys, there's a chance to use a datastructure that is hierarchical. That's the approach I took in the PendingAcksMap class:
https://github.com/apache/pulsar/blob/3d0625ba64294fb0fe7dafc27c7a34883b4be51b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PendingAcksMap.java#L101
There's a huge benefit when the keys are primitives and you can only achieve that with a hierarchical data structure.
In the long run, I'd rather get rid of our custom collection implementations instead of adding more of them.
It's not trivial to create bug free collections. Using existing libraries for that purpose is strongly preferred.

Have you considered a hierarchical data structure? (such as map of maps)

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentTripleLong2LongHashMap.java
lhotari: The triple key is most likely not needed and could be solved with a hierarchical datastructure (map of maps). Please see the other comment https://github.com/apache/pulsar/pull/23600#issuecomment-2475922541.

lhotari: There could always be a higher level abstraction to hide the implementation details.

thetumbled:
> > If any solution from `Fastutil` prove to be more space efficient, i am glad to adopt it. Nack Tracker will consume enormous amount of memory, we have to choose the best one.
> 
> The main purpose of [Fastutil](https://fastutil.di.unimi.it/) is performance and space efficiency. Instead of adding very complex keys such as triple keys, there's a chance to use a datastructure that is hierarchical. That's the approach I took in the PendingAcksMap class:
> 
> https://github.com/apache/pulsar/blob/3d0625ba64294fb0fe7dafc27c7a34883b4be51b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PendingAcksMap.java#L101
> 
> 
> There's a huge benefit when the keys are primitives and you can only achieve that with a hierarchical data structure.
> In the long run, I'd rather get rid of our custom collection implementations instead of adding more of them.
> It's not trivial to create bug free collections. Using existing libraries for that purpose is strongly preferred.
> @thetumbled Have you considered a hierarchical data structure? (such as map of maps)

It is a good point, i will test it.

lhotari:
> > > If any solution from `Fastutil` prove to be more space efficient, i am glad to adopt it. Nack Tracker will consume enormous amount of memory, we have to choose the best one.
> > 
> > 
> > The main purpose of [Fastutil](https://fastutil.di.unimi.it/) is performance and space efficiency. Instead of adding very complex keys such as triple keys, there's a chance to use a datastructure that is hierarchical. That's the approach I took in the PendingAcksMap class:
> > https://github.com/apache/pulsar/blob/3d0625ba64294fb0fe7dafc27c7a34883b4be51b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PendingAcksMap.java#L101
> > 
> > There's a huge benefit when the keys are primitives and you can only achieve that with a hierarchical data structure.
> > In the long run, I'd rather get rid of our custom collection implementations instead of adding more of them.
> > It's not trivial to create bug free collections. Using existing libraries for that purpose is strongly preferred.
> > @thetumbled Have you considered a hierarchical data structure? (such as map of maps)
> 
> It is a good point, i will test it.

@thetumbled In certain cases when tracking existence (true/false), it's worth considering to use space efficient bit maps. In Pulsar, we use the [RoaringBitmap](https://github.com/RoaringBitmap/RoaringBitmap) library.
I think that it should be used for storing nacks.

lhotari:
> I think that it should be used for storing nacks.

I guess it's not applicable in this case. 

@thetumbled I checked the NegativeAcksTracker class and it seems that the actual key is (ledgerId, entryId). 
The partitionIndex and timestamp are part of the value.
partitionIndex doesn't have to be a long value.

It's easy to implement (ledgerId, entryId) as map of maps.

lhotari:
This is a very poor solution in the current implementation:
https://github.com/apache/pulsar/blob/22cfa5428ef8a4fc6cfc0225182724012f0972c2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java#L80-L88

There should be a separate datastructure (a list or queue) which contains the entries in timestamp order. The benefit of that is that iterating could stop after the timestamp condition no longer holds.

lhotari:
@thetumbled It looks like there's no need for a map data structure in the first place. That's completely unnecessary for implementing NegativeAcksTracker

thetumbled:
> @thetumbled It looks like there's no need for a map data structure in the first place. That's completely unnecessary for implementing NegativeAcksTracker

You are right. We need to improve the code execution efficiency too. some kind of structures sorted by timestamp.

lhotari:
> You are right. We need to improve the code execution efficiency too. some kind of structures sorted by timestamp.

Fastutil contains multiple PriorityQueue implementations:  https://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/PriorityQueue.html.

thetumbled:
> > You are right. We need to improve the code execution efficiency too. some kind of structures sorted by timestamp.
> 
> Fastutil contains multiple PriorityQueue implementations: https://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/PriorityQueue.html. For example, this would work: https://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/objects/ObjectArrayPriorityQueue.html or this one: https://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/objects/ObjectHeapPriorityQueue.html

I propose a pip to fix several issues with nack tracker, with a new data structure :
```
Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> nackedMessages = new Long2ObjectAVLTreeMap<>();
```
This PR become the implementation PR for PIP-393: https://github.com/apache/pulsar/pull/23601.
I will implement PIP-393 soon.

pulsar-client/pom.xml
```diff
@@ -207,6 +207,16 @@
...
 
+    <dependency>
+      <groupId>org.roaringbitmap</groupId>
+      <artifactId>RoaringBitmap</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>it.unimi.dsi</groupId>
+      <artifactId>fastutil</artifactId>
+    </dependency>
```
lhotari, line 218: It will be necessary to update `pulsar-client-shaded` and `pulsar-client-all` shading configuration to shade these libraries since they haven't been used on the client by now. 

Avoiding client jar growth is something that we would need to address when adding new dependencies.

I realized that including the `fastuti` library will increase the client jar file size significantly. One detail about fastutil is that there's a smaller library alternative `fastutil-core` which includes a subset of the classes. `fastutil` is about 23MB and `fastutil-core` is about 6MB. 

Picking only the classes that are needed from fastutil is possible with `minimizeJar` with maven's shade plugin, but since `minimizeJar` is a global setting, it would require building a minimized version of fastutil for the client usage to pick only the classes are needed since applying `minimizeJar` to the complete Pulsar client jar would be a very large change which could break things.

pom.xml
```diff
@@ -322,6 +322,7 @@ flexible messaging model and an intuitive client API.</description>
     <extra-enforcer-rules.version>1.6.1</extra-enforcer-rules.version>
     <oshi.version>6.4.0</oshi.version>
     <checkerframework.version>3.33.0</checkerframework.version>
+    <it-unimi-dsi-fastutil.version>8.5.15</it-unimi-dsi-fastutil.version>
```
lhotari, line 325: there's already `fastutil.version` property for this purpose.

pulsar-client/pom.xml
```diff
@@ -214,7 +214,8 @@
 
     <dependency>
       <groupId>it.unimi.dsi</groupId>
-      <artifactId>fastutil</artifactId>
+      <artifactId>fastutil-core</artifactId>
+      <version>${it-unimi-dsi-fastutil.version}</version>
```
lhotari, line 218: The version could be omitted here by adding an entry to the root `pom.xml` `dependencyManagement`. You could add `fastutil-core` to follow after the existing `fastutil` dependency in dependencyManagement.

pulsar-client/pom.xml
thetumbled: > It will be necessary to update pulsar-client-shaded and pulsar-client-all shading configuration to shade these libraries since they haven't been used on the client by now.

I am not familiar with the shade part, hope this commit modify correctly.
https://github.com/apache/pulsar/pull/23600/commits/28251b2e9a3c434eec56224ea596bbfc297a9588

> I realized that including the fastuti library will increase the client jar file size significantly. One detail about fastutil is that there's a smaller library alternative fastutil-core which includes a subset of the classes. fastutil is about 23MB and fastutil-core is about 6MB.
Picking only the classes that are needed from fastutil is possible with minimizeJar with maven's shade plugin, but since minimizeJar is a global setting, it would require building a minimized version of fastutil for the client usage to pick only the classes are needed since applying minimizeJar to the complete Pulsar client jar would be a very large change which could break things.

I have update the dependency to `fastutil-core`. 
As for `minimizeJar`, it looks like we can minimize the size of jar to the minimum, but i think that we may improve pulsar with fastutil many places else, so some other data structures can be helpful too.

pom.xml
thetumbled: fixed.

pulsar-client/pom.xml
thetumbled: good idea, updated.

thetumbled:
There are some issues with the licence.
```
Run src/check-binary-license.sh ./distribution/server/target/apache-pulsar-*-bin.tar.gz
it.unimi.dsi-fastutil-core-8.5.14.jar unaccounted for in LICENSE

It looks like there are issues with the LICENSE/NOTICE.
```
Should we replace `fastutil-core` with `fastutil`? @lhotari

lhotari:
> Should we replace `fastutil-core` with `fastutil`?

@thetumbled That's fine to complete the experiment. For a proper solution there will need to be a way to avoid the shaded jar files growing significantly. The proper solution would require a separate module for fastutil which minimizes the number of classes. I tried to explain that in my previous comment, https://github.com/apache/pulsar/pull/23600#discussion_r1845971020 . Minimizing all classes would most likely be a breaking change. Since you cannot specify `minimizeJar` for a specific dependency in maven-shade-plugin configuration, it's necessary to have a separate intermediate module for minimizing just fastutil.

thetumbled:
> Run src/check-binary-license.sh ./distribution/server/target/apache-pulsar-*-bin.tar.gz
> it.unimi.dsi-fastutil-core-8.5.14.jar unaccounted for in LICENSE
> 
> It looks like there are issues with the LICENSE/NOTICE.

I wonder whether it is the problem with the `it.unimi.dsi-fastutil-core-8.5.14.jar`, or i need to configure something to fix this?

lhotari:
> I wonder whether it is the problem with the `it.unimi.dsi-fastutil-core-8.5.14.jar`, or i need to configure something to fix this?

@thetumbled When new dependencies are added, it's necessary to add them to LICENSE file(s). However, there's a bit of a problem in using `fastutil-core` together with `fastutil` since there would be duplication in the classpath. That's why it's better to simply use the `fastutil` dependency and add the minimizing solution in a later step.

thetumbled:
> > I wonder whether it is the problem with the `it.unimi.dsi-fastutil-core-8.5.14.jar`, or i need to configure something to fix this?
> 
> @thetumbled When new dependencies are added, it's necessary to add them to LICENSE file(s). However, there's a bit of a problem in using `fastutil-core` together with `fastutil` since there would be duplication in the classpath. That's why it's better to simply use the `fastutil` dependency and add the minimizing solution in a later step.

get it, i change it back to `fastutil` for now.

thetumbled:
The implementation is done, please help to review this PIP, thanks. @lhotari @BewareMyPower @nodece @poorbarcode @dao-jun

pulsar-client/pom.xml
lhotari: This comment cannot be marked resolved before this is really addressed.

It is necessary to add a new module that minimizes fastutil for the use of the shaded Pulsar client (in published pulsar-client and pulsar-client-all modules). The minimized fastutil module shouldn't be published to maven central at all, but it is necessary to make it a separate module since it's not possible to selectively minimize modules with the maven-shade-plugin.

pulsar-client-all/pom.xml
```diff
@@ -206,6 +206,7 @@
                   <include>org.tukaani:xz</include>
                   <!-- Issue #6834, Since Netty ByteBuf shaded, we need also shade this module -->
                   <include>org.apache.pulsar:pulsar-client-messagecrypto-bc</include>
+                  <include>it.unimi.dsi:*</include>
```
lhotari, line 209: This isn't the correct solution for shading fastutil. Please see https://github.com/apache/pulsar/pull/23600/files#r1848136065

pulsar-client-shaded/pom.xml
```diff
@@ -168,6 +168,7 @@
                   <include>org.tukaani:xz</include>
                   <!-- Issue #6834, Since Netty ByteBuf shaded, we need also shade this module -->
                   <include>org.apache.pulsar:pulsar-client-messagecrypto-bc</include>
+                  <include>it.unimi.dsi:*</include>
```
lhotari, line 171: This isn't the correct solution for shading fastutil. Please see https://github.com/apache/pulsar/pull/23600/files#r1848136065

pulsar-client/pom.xml
thetumbled: It is a later step that we minimize the fastutil, right?
Currently, we need to fix the shade problem, is it not enough with following code?
```
<include>it.unimi.dsi:*</include>
```
I will appreciate it if you can help to shade the fastutil dependency.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
```diff
@@ -311,19 +312,61 @@ public void testNegativeAcksDeleteFromUnackedTracker() throws Exception {
...
+                .subscriptionName("sub1")
+                .acknowledgmentGroupTime(0, TimeUnit.SECONDS)
+                .subscriptionType(SubscriptionType.Shared)
+                .enableBatchIndexAcknowledgment(true)
+                .negativeAckRedeliveryDelay(3, TimeUnit.SECONDS)
+                .subscribe();
+
+        @Cleanup
+        Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
+                .topic(topic)
```
nodece, line 352:
```suggestion
                .topic(topic)
                .enableBatching(true)
                .batchingMaxPublishDelay(1, TimeUnit.HOURS)
                .batchingMaxMessages(2)
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
```diff
@@ -108,39 +134,56 @@ public synchronized void add(Message<?> message) {
...
         if (nackedMessages == null) {
-            nackedMessages = ConcurrentLongLongPairHashMap.newBuilder()
-                    .autoShrink(true)
-                    .concurrencyLevel(1)
-                    .build();
+            nackedMessages = new Long2ObjectAVLTreeMap<>();
         }
 
-        long backoffNs;
+        long backoffMs;
```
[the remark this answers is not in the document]
thetumbled: - Nano is not convenient for calculating, as we need to estimate the accuracy of timestamp. For example, with `negativeAckPrecisionBitCnt = 10`, we know that the redelivery time may be earlier at most 2^10=1024ms~=1s. We just trim the lower 8 bit to bucket the time. But with nano(`1ms=1000000ns`).  we can't get a suitable conf value easily.
- It is unnecessary to use timestamp in `ns` unit. As the tick time of the timer is 1ms, and the latency of message delivery is at `ms` level.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ConsumerConfigurationData.java
```diff
@@ -155,6 +155,16 @@ public class ConsumerConfigurationData<T> implements Serializable, Cloneable {
...
 
+    @ApiModelProperty(
+            name = "negativeAckPrecisionBitCnt",
+            value = "The redelivery time precision bit count. The lower bits of the redelivery time will be"
+                    + "trimmed to reduce the memory occupation.\nThe default value is 8, which means the"
+                    + "redelivery time will be bucketed by 256ms, the redelivery time could be earlier(no later)"
+                    + "than the expected time, but no more than 256ms. \nIf set to k, the redelivery time will be"
+                    + "bucketed by 2^k ms.\nIf the value is 0, the redelivery time will be accurate to ms."
+    )
+    private int negativeAckPrecisionBitCnt = 8;
```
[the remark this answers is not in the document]
thetumbled: I suggest to configure it as the best the confguration value. Without this enhancement, it is unrealistic to use negative acknowledgement in production. The memory occupation will inflate very fast.
But i will listen to the voice from the community, if more and more people think that disabling this feature by default is better, i will update it.

lhotari:
@thetumbled Please don't include `[pip]` in the title of the implementation PR. It's only for the PIP document PR.

thetumbled:
> @thetumbled Please don't include `[pip]` in the title of the implementation PR. It's only for the PIP document PR.

Thanks for reminder.

thetumbled:
Hi, fellows in pulsar communtiy. Could you help to move this feature forward, thanks. 
@lhotari @nodece @Technoboy- @codelipenghui @dao-jun @poorbarcode @congbobo184

thetumbled:
> Failed to execute goal org.owasp:dependency-check-maven:10.0.2:aggregate (default) on project pulsar: 
Error:  
Error:  One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0': 
Error:  
Error:  alluxio-core-common-2.9.3.jar: CVE-2023-38889(9.8)
Error:  azure-identity-1.10.4.jar: CVE-2023-36415(8.8)
Error:  clickhouse-jdbc-0.4.6-all.jar/META-INF/maven/com.google.protobuf/protobuf-java/pom.xml: CVE-2024-7254(8.699999809265137)
Error:  clickhouse-jdbc-0.4.6-all.jar/META-INF/maven/com.squareup.okio/okio/pom.xml: CVE-2023-3635(7.5)
Error:  clickhouse-jdbc-0.4.6-all.jar: CVE-2022-44010(7.5)
Error:  grpc-core-1.37.0.jar: CVE-2023-44487(7.5), CVE-2023-4785(7.5), CVE-2023-33953(7.5)
Error:  grpc-core-1.56.1.jar: CVE-2023-44487(7.5), CVE-2023-33953(7.5)
Error:  grpc-protobuf-1.37.0.jar: CVE-2023-44487(7.5), CVE-2023-4785(7.5), CVE-2023-33953(7.5)
Error:  grpc-protobuf-1.56.1.jar: CVE-2023-44487(7.5), CVE-2023-33953(7.5)
Error:  json-20201115.jar: CVE-2023-5072(7.5), CVE-2022-45688(7.5)
Error:  kafka-clients-2.8.1.jar: CVE-2024-31141(7.099999904632568)
Error:  kafka-clients-3.4.0.jar: CVE-2024-31141(7.099999904632568)
Error:  kaml-0.20.0.jar: CVE-2023-28118(7.5)
Error:  lucene-core-9.11.1.jar: CVE-2024-45772(8.0)
Error:  mockserver-netty-no-dependencies-5.14.0.jar: CVE-2021-32827(9.6)
Error:  mysql-connector-java-8.0.30.jar: CVE-2023-22102(8.3)
Error:  nimbus-jose-jwt-9.30.2.jar: CVE-2023-52428(7.5)

Do you know the reason why this error throw out? @lhotari

lhotari:
> > Failed to execute goal org.owasp:dependency-check-maven:10.0.2:aggregate (default) on project pulsar:
> > Error:
> > Error:  One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0':
> > Error:
> > Error:  alluxio-core-common-2.9.3.jar: CVE-2023-38889(9.8)
> > Error:  azure-identity-1.10.4.jar: CVE-2023-36415(8.8)
> > Error:  clickhouse-jdbc-0.4.6-all.jar/META-INF/maven/com.google.protobuf/protobuf-java/pom.xml: CVE-2024-7254(8.699999809265137)
> > Error:  clickhouse-jdbc-0.4.6-all.jar/META-INF/maven/com.squareup.okio/okio/pom.xml: CVE-2023-3635(7.5)
> > Error:  clickhouse-jdbc-0.4.6-all.jar: CVE-2022-44010(7.5)
> > Error:  grpc-core-1.37.0.jar: CVE-2023-44487(7.5), CVE-2023-4785(7.5), CVE-2023-33953(7.5)
> > Error:  grpc-core-1.56.1.jar: CVE-2023-44487(7.5), CVE-2023-33953(7.5)
> > Error:  grpc-protobuf-1.37.0.jar: CVE-2023-44487(7.5), CVE-2023-4785(7.5), CVE-2023-33953(7.5)
> > Error:  grpc-protobuf-1.56.1.jar: CVE-2023-44487(7.5), CVE-2023-33953(7.5)
> > Error:  json-20201115.jar: CVE-2023-5072(7.5), CVE-2022-45688(7.5)
> > Error:  kafka-clients-2.8.1.jar: CVE-2024-31141(7.099999904632568)
> > Error:  kafka-clients-3.4.0.jar: CVE-2024-31141(7.099999904632568)
> > Error:  kaml-0.20.0.jar: CVE-2023-28118(7.5)
> > Error:  lucene-core-9.11.1.jar: CVE-2024-45772(8.0)
> > Error:  mockserver-netty-no-dependencies-5.14.0.jar: CVE-2021-32827(9.6)
> > Error:  mysql-connector-java-8.0.30.jar: CVE-2023-22102(8.3)
> > Error:  nimbus-jose-jwt-9.30.2.jar: CVE-2023-52428(7.5)
> 
> Do you know the reason why this error throw out? @lhotari

@thetumbled  We haven't addressed these issues. The OWASP dependency check has been failing for quite some time. Most of these issues are in Pulsar IO Connectors. For GRPC, there has been discussions on the dev mailing list.

thetumbled:
> > > Failed to execute goal org.owasp:dependency-check-maven:10.0.2:aggregate (default) on project pulsar:
> > > Error:
> > > Error:  One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0':
> > > Error:
> > > Error:  alluxio-core-common-2.9.3.jar: CVE-2023-38889(9.8)
> > > Error:  azure-identity-1.10.4.jar: CVE-2023-36415(8.8)
> > > Error:  clickhouse-jdbc-0.4.6-all.jar/META-INF/maven/com.google.protobuf/protobuf-java/pom.xml: CVE-2024-7254(8.699999809265137)
> > > Error:  clickhouse-jdbc-0.4.6-all.jar/META-INF/maven/com.squareup.okio/okio/pom.xml: CVE-2023-3635(7.5)
> > > Error:  clickhouse-jdbc-0.4.6-all.jar: CVE-2022-44010(7.5)
> > > Error:  grpc-core-1.37.0.jar: CVE-2023-44487(7.5), CVE-2023-4785(7.5), CVE-2023-33953(7.5)
> > > Error:  grpc-core-1.56.1.jar: CVE-2023-44487(7.5), CVE-2023-33953(7.5)
> > > Error:  grpc-protobuf-1.37.0.jar: CVE-2023-44487(7.5), CVE-2023-4785(7.5), CVE-2023-33953(7.5)
> > > Error:  grpc-protobuf-1.56.1.jar: CVE-2023-44487(7.5), CVE-2023-33953(7.5)
> > > Error:  json-20201115.jar: CVE-2023-5072(7.5), CVE-2022-45688(7.5)
> > > Error:  kafka-clients-2.8.1.jar: CVE-2024-31141(7.099999904632568)
> > > Error:  kafka-clients-3.4.0.jar: CVE-2024-31141(7.099999904632568)
> > > Error:  kaml-0.20.0.jar: CVE-2023-28118(7.5)
> > > Error:  lucene-core-9.11.1.jar: CVE-2024-45772(8.0)
> > > Error:  mockserver-netty-no-dependencies-5.14.0.jar: CVE-2021-32827(9.6)
> > > Error:  mysql-connector-java-8.0.30.jar: CVE-2023-22102(8.3)
> > > Error:  nimbus-jose-jwt-9.30.2.jar: CVE-2023-52428(7.5)
> > 
> > 
> > Do you know the reason why this error throw out? @lhotari
> 
> @thetumbled We haven't addressed these issues. The OWASP dependency check has been failing for quite some time. Most of these issues are in Pulsar IO Connectors. For GRPC, there has been discussions on the dev mailing list.

Can we proceed to review this pr without handing these issues? and one more approve is needed, could you help to review it? thanks.

lhotari:
> Can we proceed to review this pr without handing these issues? and one more approve is needed, could you help to review it? thanks.

Yes, Owasp check failure doesn't block merging of the PR.

lhotari:
@thetumbled I'll push the shading configuration changes directly to this PR.

lhotari:
@thetumbled  I added 2 commits to this PR to address the shading. The first commit abcfe5c adds a solution for minimizing the number of classes that are included from fastutil to the shaded jars. The second commit a03f415 addresses RoaringBitmap shading which was missing.

lhotari (approved):
LGTM

PR #23601, opened 2024-11-15
[improve][pip] PIP-393: Improve performance of Negative Acknowledgement

thetumbled:
PIP: 393
Implementation PR: https://github.com/apache/pulsar/pull/23600.

### Motivation

There are many issues with the current implementation of Negative Acknowledgement in Pulsar:
- the memory occupation is high.
- the code execution efficiency is low.
- the redelivery time is not accurate.
- multiple negative ack for messages in the same entry(batch) will interfere with each other.
All of these problem is severe and need to be solved.

### Modifications

Refactor the `NegativeAcksTracker` to solve the above problems.

### Space complexity of new data structure
I will show you how great the new data structure it is with theorectical space complexity analysis.

#### Space complexity of `ConcurrentLongLongPairHashMap`
Before analyzing the new data structure, we need to know how much space it take before this pip. We need to store 4 long field for `(ledgerId, entryId, partitionIndex, timestamp)` for each entry, which takes 4*8=32byte.
As `ConcurrentLongLongPairHashMap` use open hash addressing and linear probe to handle hash confliction, there are rebundunt spaces to avoid high confliction rate. There are two configurations that control how much rebundunt space to reserver: `fill factor` and `idle factor`. When the space utility rate soar high to `fill factor`, the size of backing array will be double, when the space utility rate reduce to `idle factor`,  the size of backing array will reduce by half.
The default value of `fill factor` is 0.66, `idle factor` is 0.15, which means the min space occupation of `ConcurrentLongLongPairHashMap` is `32/0.66N byte = 48N byte`, the max space occupation is `32/0.15N byte=213N byte`, where N is the number of entries.

List some test data to verify this:
image
There are 100w entries in the map, which take up `32*1000000/1024/1024byte=30MB`, the space utility rate is 30/64=0.46, in the range of `[0.15, 0.66]`.

#### Space complexity of new data structure
New data structure:
```
// timestamp -> ledgerId -> entryId
Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> map2 = new Long2ObjectAVLTreeMap<>();
```

The space used by new data structure is related to several factors: `message rate`, `the time deviation user accepted`, `the max entries written in one ledger`.
- Pulsar conf `managedLedgerMaxEntriesPerLedger=50000` determine the max entries can be wriitten into one ledger, we use the default value to analyze.
- `the time deviation user accepted`: when user accept 1024ms delivery time deviation, we can trim the lower 10 bit of the timestamp in ms, which can bucket 1024 timestamp.

We will analyze the space used by one bucket, and calculate the average space used by one entry.
Assuming that the message rate is `x msg/ms`, and we trim `y bit` of the timestamp, one bucket will contains `2**x` ms, `M=2**x*y` msgs in one bucket.
- For one single bucket, we only need to store one timestamp, which takes 8byte.
- Then, we need to store the ledgerId, when M is greater than 5w(`managedLedgerMaxEntriesPerLedger`), the ledger will switch. There are `L=ceil(M/50000)` ledgers, which take `8*L` byte.
- Further, we analyze how much space the entry id takes. As there are `L=ceil(M/50000)` ledgers, there will be `L` bitmap to store, which take L*size(bitmap). The total space consumed by new data structure is `8byte + 8L byte + L*size(bitmap)`. 

As the `size(bitmap)` is far more greater than `8byte`, we can ignore the first two items. Then we get the formular of space consumed **one bucket**: `D=L*size(bitmap)=ceil(M/50000)*size(bitmap)`.

Entry id is stored in a `Roaring64Bitmap`, for simplicity we can replace it with `RoaringBitmap`, as the max entry id is 49999, which is smaller than `4294967296 (2 * Integer.MAX_VALUE)`(the max value can be stored in `RoaringBitmap`). The space consume by `RoaringBitmap` depends on how many elements it contains, when the size of bitmap < 4096, the space is `4N btye`, when the size of bitmap > 4096, the consumed space is a fixed value 8KB.
Then we get the final result:
- when M>50000, `D = ceil(M/50000)*size(bitmap) ~= M/50000 * 8KB = M/50000 * 8 * 1024 byte = 0.163M byte`, each entry takes `0.163byte` by average.
-  when 4096<M<50000, `D = ceil(M/50000)*size(bitmap) = 1 * 8KB = 8KB`, each entry takes `8*1024/M=8192/M byte` by average.
- when M<4096, `D = ceil(M/50000)*size(bitmap) = 1 * 4Mbyte = 4Mbyte`, each entry take `4 byte` by average.

#### Conclusion
- The space complexity of `ConcurrentLongLongPairHashMap` is `48N` byte in best case, `213N byte` in worst case, where N is the number of entries.
- The space complexity of new data structure is determined by the total number of messages in one bucket M.
    - when M>50000, space complexity is `0.163N` byte.
    - when 4096<M<50000, space complexity is `8192/M * N byte` .
    - when M<4096, space complexity is `4N byte`.

### test data
List some experiment data to verify the analysis above.
Test code:
```
static long trimLowerBit(long timestamp, int bits) {
        return timestamp & (-1L << bits);
    }
    public static void main(String[] args) throws IOException {
        ConcurrentLongLongPairHashMap map1 = ConcurrentLongLongPairHashMap.newBuilder()
                .autoShrink(true)
                .concurrencyLevel(16)
                .build();
        // timestamp -> ledgerId -> entryId, no need to batch index, if different messages have
        // different timestamp, there will be multiple entries in the map
        // AVL Tree -> LongOpenHashMap -> Roaring64Bitmap
        // there are many timestamp, a few ledgerId, many entryId
        Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> map2 = new Long2ObjectAVLTreeMap<>();
        
        int trimLowerBits = 10;
        long numMessages = 1000000, entriesPerLedger = 1000, numLedgers = numMessages / entriesPerLedger;
        long ledgerId, entryId, timestamp=System.currentTimeMillis(), tmp=0;
        for (long i = 0; i < numLedgers; i++) {
            ledgerId = 10000+i;
            for (long j = 0; j < entriesPerLedger; j++) {
                entryId = j;
                // 1ms per message
                timestamp++;
                // queue.add(timestamp, ledgerId, entryId);
                map1.put(ledgerId, entryId, 0L, timestamp);
                
                tmp = trimLowerBit(timestamp, trimLowerBits);
                map2.computeIfAbsent(tmp, k -> new Long2ObjectOpenHashMap<>())
                    .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap())
                    .add(entryId);
            }
        }
    }
```

#### x=1, y=10
Let x=1, that is 1msg/ms, y=10, we will trim 10 bit of the timestamp. Then M=1*2**10=1024<4096. According to the reslut above, we predict that the space consume by 100w entries is `4*1000000/1024/1024=3.81MB`.

The actual space consumed is `3.35MB`, which is quite near to the theorectical value.

#### x=50, y=10
We  try to reach to the best space complexity. `M=50*2**10=51200>50000`, we predict that average space consume by one entry is 0.163 byte.
```
int trimLowerBits = 10, messagePerMs = 50, tick=0;
long numMessages = 1000000, entriesPerLedger = 50000, numLedgers = numMessages / entriesPerLedger;
```
image
But the experiment result is `0.33*1024*1024/1000000=0.34byte`, almost twice of the theorectal value `0.163`.

We can print the size of bitmap to know why.

There are still many bitmaps whose size is far more smaller than 5w, which result into the lower space utility rate.

#### x=500, y=10
```
int trimLowerBits = 10, messagePerMs = 500, tick=0;
long numMessages = 1000000, entriesPerLedger = 50000, numLedgers = numMessages / entriesPerLedger;
```

All bitmaps contains  almost 5w entries.
image
Each entry take `0.18*1024*1024/1000000=0.18byte`, which is quite near to the the theorectical value.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-393.md
+++ b/pip/pip-393.md
@@ -95,8 +95,26 @@ ledgerMap.put(ledgerId, entrySet);
 map.put(timestamp, ledgerMap);
 ```
 
-### Space complexity analysis
-#### Space complexity of `ConcurrentLongLongPairHashMap`
+### Configuration
+
+Add a new configuration `negativeAckPrecisionBitCnt` to control the precision of the redelivery time.
+```
+@ApiModelProperty(
+            name = "negativeAckPrecisionBitCnt",
+            value = "The redelivery time precision bit count. The lower bits of the redelivery time will be\n" + 
+                "trimmed to reduce the memory occupation. The default value is 8, which means the redelivery time\n" +
+                "will be bucketed by 256ms. In worst cases, the redelivery time will be 512ms earlier(no later)\n" +
+                "than the expected time. If the value is 0, the redelivery time will be accurate to ms.".
+    )
+    private long negativeAckPrecisionBitCnt = 8;
+```
+The higher the value, the more entries will be grouped into the same bucket, the less memory occupation, the less accurate the redelivery time.
+Default value is 8, which means the redelivery time will be bucketed by 256ms. In worst cases, the redelivery time will be 512ms earlier(no later)
+than the expected time.
+
+
+## Space complexity analysis
+### Space complexity of `ConcurrentLongLongPairHashMap`
 Before analyzing the new data structure, we need to know how much space it take before this pip.
 
 We need to store 4 long field for `(ledgerId, entryId, partitionIndex, timestamp)` for each entry, which takes `4*8=32byte`.
@@ -113,7 +131,7 @@ In the experiment showed in the PR, there are 100w entries in the map, taking up
 the space utility rate is 30/64=0.46, in the range of `[0.15, 0.66]`.
 
 
-#### Space complexity of the new data structure
+### Space complexity of the new data structure
 The space used by new data structure is related to several factors: `message rate`, `the time deviation user accepted`,
 `the max entries written in one ledger`.
 - Pulsar conf `managedLedgerMaxEntriesPerLedger=50000` determine the max entries can be written into one ledger,
@@ -145,7 +163,7 @@ each entry takes `0.163byte` by average.
 - when 4096<M<50000, `D = ceil(M/50000)*size(bitmap) = 1 * 8KB = 8KB`, each entry takes `8*1024/M=8192/M byte` by average.
 - when M<4096, `D = ceil(M/50000)*size(bitmap) = 1 * 4M byte = 4M byte`, each entry take `4 byte` by average.
 
-#### Conclusion
+### Conclusion
 Assuming N is the number of entries, M is the number of messages in one bucket.
 - `ConcurrentLongLongPairHashMap`: `48N` byte in best case, `213N byte` in worst case.
 - New data structure:
@@ -155,42 +173,38 @@ Assuming N is the number of entries, M is the number of messages in one bucket.
 
 Some experiment results are showed in the PR, we can fine tune the configuration to get the best performance.
 
-### Effect
+## Effect
 
-#### Memory occupation is high
+### Memory occupation is high
 With such kind of design, we can reduce the memory occupation of `NegativeAcksTracker` to 1% less than the current implementation.
-The detailed test result will be provided in the PR.
 
-#### Code execution efficiency is low
+### Code execution efficiency is low
 With the new design, we can avoid the iteration of all entries in `NegativeAcksTracker.nackedMessages`, and only iterate the entries
 that need to be redelivered.
 
-#### Redelivery time is not accurate
+### Redelivery time is not accurate
 With the new design, we avoid the fixed interval of the redelivery check time. We can control the precision of the redelivery time 
 by trimming the lower bits of the timestamp. If user can accept 1024ms deviation of the redelivery time, we can trim the lower
 10 bits of the timestamp, which can group a lot
 
-#### Multiple negative ack for messages in the same entry(batch) will interfere with each other
+### Multiple negative ack for messages in the same entry(batch) will interfere with each other
 With the new design, if we let msg1 redelivered 10s later, then let msg2 redelivered 20s later, these two nacks will not interfere
 with each other, as they are stored in different buckets.
 
 
-### Configuration
+## High-Level Design
+As this pip introduce new dependency `fastutil` into client, which is very large(23MB), while few classes are used, we need to
+reduce the size of the dependency. 
 
-Add a new configuration `negativeAckPrecisionBitCnt` to control the precision of the redelivery time.
-```
-@ApiModelProperty(
-            name = "negativeAckPrecisionBitCnt",
-            value = "The redelivery time precision bit count. The lower bits of the redelivery time will be\n" + 
-                "trimmed to reduce the memory occupation. The default value is 8, which means the redelivery time\n" +
-                "will be bucketed by 256ms. In worst cases, the redelivery time will be 512ms earlier(no later)\n" +
-                "than the expected time. If the value is 0, the redelivery time will be accurate to ms.".
-    )
-    private long negativeAckPrecisionBitCnt = 8;
-```
-The higher the value, the more entries will be grouped into the same bucket, the less memory occupation, the less accurate the redelivery time.
-Default value is 8, which means the redelivery time will be bucketed by 256ms. In worst cases, the redelivery time will be 512ms earlier(no later)
-than the expected time.
+Though there is alternative dependency `fastutil-core`, which is smaller(6MB), but it is also
+relatively large and using `fastutil-core` will introduce another problem on the broker side since there's already `fastutil` jar 
+which also includes `fastutil-core` jar classes.
+
+The optimal solution would be to include only the classes from fastutil into the shaded pulsar-client and pulsar-client-all 
+which are really used and needed. This could be achieved in many ways. One possible solution is to introduce an intermediate
+module for shaded pulsar-client and pulsar-client-all that isn't published to maven central at all. 
+It would be used to minimize and include only the classes from fastutil which are required by pulsar-client shading.
+
 
 
 # Backward & Forward Compatibility
--- a/pip/pip-393.md
+++ b/pip/pip-393.md
@@ -223,4 +223,4 @@ User can downgrade to the old version without any compatibility issue.
 Updated afterwards
 -->
 * Mailing List discussion thread: https://lists.apache.org/thread/yojl7ylk7cyjxktq3cn8849hvmyv0fg8
-* Mailing List voting thread:
+* Mailing List voting thread: https://lists.apache.org/thread/hyc1r2s9chowdhck53lq07tznopt50dy
--- a/pip/pip-393.md
+++ b/pip/pip-393.md
@@ -98,8 +98,24 @@ ledgerMap.put(ledgerId, entrySet);
 map.put(timestamp, ledgerMap);
 ```
 
+### Effect
+
+#### Memory occupation is high
 With such kind of design, we can reduce the memory occupation of `NegativeAcksTracker` to 1% less than the current implementation.
 The detailed test result will be provided in the PR.
+
+#### Code execution efficiency is low
+With the new design, we can avoid the iteration of all entries in `NegativeAcksTracker.nackedMessages`, and only iterate the entries
+that need to be redelivered.
+
+#### Redelivery time is not accurate
+With the new design, we avoid the fixed interval of the redelivery check time. We can control the precision of the redelivery time 
+by trimming the lower bits of the timestamp. If user can accept 1024ms deviation of the redelivery time, we can trim the lower
+10 bits of the timestamp, which can group a lot
+
+#### Multiple negative ack for messages in the same entry(batch) will interfere with each other
+With the new design, if we let msg1 redelivered 10s later, then let msg2 redelivered 20s later, these two nacks will not interfere
+with each other, as they are stored in different buckets.
 
 
 ### Configuration
--- a/pip/pip-393.md
+++ b/pip/pip-393.md
@@ -95,6 +95,66 @@ ledgerMap.put(ledgerId, entrySet);
 map.put(timestamp, ledgerMap);
 ```
 
+### Space complexity analysis
+#### Space complexity of `ConcurrentLongLongPairHashMap`
+Before analyzing the new data structure, we need to know how much space it take before this pip.
+
+We need to store 4 long field for `(ledgerId, entryId, partitionIndex, timestamp)` for each entry, which takes `4*8=32byte`.
+As `ConcurrentLongLongPairHashMap` use open hash addressing and linear probe to handle hash conflict, there are some 
+redundant spaces to avoid high conflict rate. There are two configurations that control how much redundant space to reserver: 
+`fill factor` and `idle factor`. When the space utility rate soar high to `fill factor`, the size of backing array will
+be double, when the space utility rate reduce to `idle factor`,  the size of backing array will reduce by half.
+
+The default value of `fill factor` is 0.66, `idle factor` is 0.15, which means the min space occupation of
+`ConcurrentLongLongPairHashMap` is `32/0.66N byte = 48N byte`, the max space occupation is `32/0.15N byte=213N byte`, 
+where N is the number of entries.
+
+In the experiment showed in the PR, there are 100w entries in the map, taking up `32*1000000/1024/1024byte=30MB`,
+the space utility rate is 30/64=0.46, in the range of `[0.15, 0.66]`.
+
+
+#### Space complexity of the new data structure
+The space used by new data structure is related to several factors: `message rate`, `the time deviation user accepted`,
+`the max entries written in one ledger`.
+- Pulsar conf `managedLedgerMaxEntriesPerLedger=50000` determine the max entries can be written into one ledger,
+we use the default value to analyze.
+- `the time deviation user accepted`: when user accept 1024ms delivery time deviation, we can trim the lower 10 bit
+of the timestamp in ms, which can bucket 1024 timestamp.
+
+Following we will analyze the space used by one bucket, and calculate the average space used by one entry.
+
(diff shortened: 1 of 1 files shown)


thetumbled:
List some experiment data:
```
    static long trimLowerBit(long timestamp, int bits) {
        return timestamp & (-1L << bits);
    }

    public static void main(String[] args) throws IOException {
        ConcurrentLongLongPairHashMap map1 = ConcurrentLongLongPairHashMap.newBuilder()
                .autoShrink(true)
                .concurrencyLevel(16)
                .build();
        // timestamp -> ledgerId -> entryId, no need to batch index, if different messages have
        // different timestamp, there will be multiple entries in the map
        // AVL Tree -> LongOpenHashMap -> Roaring64Bitmap
        // there are many timestamp, a few ledgerId, many entryId
        Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> map2 = new Long2ObjectAVLTreeMap<>();
        
        long numMessages = 1000000, numLedgers=100;
        long numEntries = numMessages/numLedgers;
        long ledgerId, entryId, partitionIndex, timestamp=System.currentTimeMillis();
        for(long i=0; i<numLedgers; i++) {
            ledgerId = 10000+i;
            for(long j=0; j<numEntries; j++) {
                entryId = 10000+j;
                partitionIndex = 0;
                // 1ms per message
                timestamp++;
                map1.put(ledgerId, entryId, partitionIndex, timestamp);
                
                timestamp = trimLowerBit(timestamp, 8);
                if (map2.containsKey(timestamp)) {
                    Long2ObjectMap<Roaring64Bitmap> ledgerMap = map2.get(timestamp);
                    if (ledgerMap.containsKey(ledgerId)) {
                        ledgerMap.get(ledgerId).add(entryId);
                    } else {
                        Roaring64Bitmap entrySet = new Roaring64Bitmap();
                        entrySet.add(entryId);
                        ledgerMap.put(ledgerId, entrySet);
                    }
                } else {
                    Long2ObjectMap<Roaring64Bitmap> ledgerMap = new Long2ObjectOpenHashMap<>();
                    Roaring64Bitmap entrySet = new Roaring64Bitmap();
                    entrySet.add(entryId);
                    ledgerMap.put(ledgerId, entrySet);
                    map2.put(timestamp, ledgerMap);
                }
            }
        }
        try {
            Thread.sleep(10000000);
            // TODO Auto-generated catch block
            e.printStackTrace();
        }
    }
```
image

pip/pip-393.md
```diff
@@ -0,0 +1,138 @@
...
+If we let msg1 redelivered 10s later, then let msg2 redelivered 20s later, these two messages are delivered 20s later together.
+msg1 will not be redelivered 10s later as the timestamp recorded in `NegativeAcksTracker#nackedMessages` is overrode by the second
+nack call.
+
+
+# Goals
+
+Refactor the `NegativeAcksTracker` to solve the above problems.
+
+To avoid interation of all entries in `NegativeAcksTracker.nackedMessages`, we use a sorted map to store the entries.
```
lhotari, line 54: A map datastructure isn't needed as we have discussed.

thetumbled: We need to key by timestamp. If you use  `PriorityQueue<Long2ObjectMap<Roaring64Bitmap>>`,  we need to wrap `timestamp` and `Long2ObjectMap<Roaring64Bitmap>` into a object, which consume space.
`Long2ObjectSortedMap` is better than `PriorityQueue`.

lhotari: > If you use `PriorityQueue<Long2ObjectMap<Roaring64Bitmap>>`, there are several problem:
> 
> * we can only access the first element and the last element of the queue, but **we need to update element that is at the middle of the queue.**
> * we need to wrap `timestamp` and `Long2ObjectMap<Roaring64Bitmap>` into a object, which consume space.

I'd try to avoid using a map completely. 

For scheduling, it would be possible to optimize the scheduling datastructure without the need to use a priority queue. This is possible since the scheduling is intended to group the nacks together with a certain resolution. I'd have a separate datastructure for that purpose. Something similar to Netty's HashedWheelTimer/HashedWheelBucket, but specialized for this use case. 

> multiple negative ack for messages in the same entry(batch) will interfere with each other.

The only reason why a nack would need to be updated seems to be related to batch messages?

Since a full batch will always be redelivered, I wonder what the correct and expected behavior even is from the user's point of view? 

There's a possibility to de-duplicate the nacks for batch messages with a `Long2ObjectMap<Roaring64Bitmap>` to keep state of current positions that are already scheduled to be delivered. If another redelivery is requested for the same position, it could be simply ignored. Would that be sufficient?

If that's not sufficient, I think it would be necessary to introduce a feature in NegativeAcksTracker to send nacks for batch message only after all batch message indices have been acknowledged or nacked. It seems that it would make more sense. 

For implementing this efficiently in the client, I think that BatchMessageIdImpl could be improved to avoid the need to use `org.apache.pulsar.client.impl.MessageIdAdvUtils#discardBatch`. Instead of creating new instances for the entry's message id, each `BatchMessageIdImpl` could hold a reference to a single BatchEntryMessageIdImpl (or whatever it's called) that would also hold the `BitSet` that is currently shared in `BatchMessageIdImpl` instances. When there's a shared instance for the message id for the entry of the batch messages, it could be used directly as a key in a map without causing more memory use. This change would reduce memory usage since the `.discardBatch` is used to create instances that are used in other trackers in the client at the moment. 
When a message id could be directly used as a key in a map, there would be any need for the current hacks that we just do to get better space efficiency.

thetumbled: >  I'd try to avoid using a map completely.

I don't know why should we avoid using map? The test result is good, i think it is good enough.

>> multiple negative ack for messages in the same entry(batch) will interfere with each other.

> The only reason why a nack would need to be updated seems to be related to batch messages?
Since a full batch will always be redelivered, I wonder what the correct and expected behavior even is from the user's point of view?
There's a possibility to de-duplicate the nacks for batch messages with a Long2ObjectMap<Roaring64Bitmap> to keep state of current positions that are already scheduled to be delivered. If another redelivery is requested for the same position, it could be simply ignored. Would that be sufficient?
If that's not sufficient, I think it would be necessary to introduce a feature in NegativeAcksTracker to send nacks for batch message only after all batch message indices have been acknowledged or nacked. It seems that it would make more sense.
For implementing this efficiently in the client, I think that BatchMessageIdImpl could be improved to avoid the need to use org.apache.pulsar.client.impl.MessageIdAdvUtils#discardBatch. Instead of creating new instances for the entry's message id, each BatchMessageIdImpl could hold a reference to a single BatchEntryMessageIdImpl (or whatever it's called) that would also hold the BitSet that is currently shared in BatchMessageIdImpl instances. When there's a shared instance for the message id for the entry of the batch messages, it could be used directly as a key in a map without causing more memory use. This change would reduce memory usage since the .discardBatch is used to create instances that are used in other trackers in the client at the moment.
When a message id could be directly used as a key in a map, there would be any need for the current hacks that we just do to get better space efficiency.

No, user actually don't need to know different messages come from the same batch. They need to be able assign different redelivery time for the different messages in the same batch.
For example, one batch contains msg1, msg2, msg3, msg4. Consumer can nack msg1 with delay 10s, nack msg2 with 20s, nack msg3 with 30s, nack msg4 with 20s. At such situation, we need to update the entry correponding to the timestamp for 20s, which is not the head of the queue, or the tail of the queue.

lhotari: > No, user actually don't need to know different messages come from the same batch. They need to be able assign different redelivery time for the different messages in the same batch.

I agree that the user doesn't need to know this, but there are technical limitations. The complete entry with all messages in the batch will get redelivered.

> For example, one batch contains msg1, msg2, msg3, msg4. Consumer can nack msg1 with delay 10s, nack msg2 with 20s, nack msg3 with 30s, nack msg4 with 20s. At such situation, we need to update the entry correponding to the timestamp for 20s, which is not the head of the queue, or the tail of the queue.
> 
> And we don't need to de-duplicate the nacks from messages in the same batch, as they are not-related. Consumer don't know different messages are from the same batch, batch is just a way to improve the performance.

I disagree. It doesn't make sense since each time the complete batch will get redelivered. The Pulsar client will filter out the already acknowledged messages from being processed again, however this filtering most likely has race conditions at least with message listeners. 

btw. If the nacks don't need to be de-duplicated, there wouldn't be any need to have a map data structure. You could simply add entries to a priority queue (or an optimized hashed wheel type of datastructure) each time a message is nacked.

thetumbled: > I agree that the user doesn't need to know this, but there are technical limitations. The complete entry with all messages in the batch will get redelivered.

It is ok that the complete entry with all messages in the batch will get redelivered. We should enable batch index ack when using negative ack to filter duplicate messages. 
It is unacceptable that the whole messages of one batch restrict to be redellivered at the same time. As consumer can negative different messages with different redelivery time. **Batch should be transparent to user.**

> I disagree. It doesn't make sense since each time the complete batch will get redelivered. The Pulsar client will filter out the already acknowledged messages from being processed again, however this filtering most likely has race conditions at least with message listeners.

There is guarantee for at-least-one semanticis, not exactly-once semantics. If we can enable batch index ack to de-dup message, which has became the default configuration.

> btw. If the nacks don't need to be de-duplicated, there wouldn't be any need to have a map data structure. You could simply add entries to a priority queue (or an optimized hashed wheel type of datastructure) each time a message is nacked.

We can't just store the entries in client memory within a priority queue because there will a greate number of entries. This is the reason why we need negative ack, we just retain a message id with tens of bytes.

thetumbled:
I add the space complexity analysis of the new data structure, please review it again, thanks. @lhotari @nodece @BewareMyPower @poorbarcode @codelipenghui

lhotari:
> I add the space complexity analysis of the new data structure, please review it again, thanks. @lhotari @nodece @BewareMyPower @poorbarcode @codelipenghui @dao-jun

Great analysis @thetumbled . Please move the analysis from the PR description to the PIP document itself. 

One small detail (which doesn't impact the analysis or solution): "Entry id is stored in a Roaring64Bitmap, for simplicity we can replace it with RoaringBitmap, as the max entry id is 49999, which is smaller than 65535."
Isn't the value `65535` irrelevant since RoaringBitmap supports storing 4294967296 (2 * Integer.MAX_VALUE) integers, explained in https://github.com/RoaringBitmap/RoaringBitmap/blob/cca90c986d5c0096bbeabb5f968833bf12c28c0e/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java#L46-L49 . Roaring64Bitmap can store up to 9223372036854775807 long integers (2 * Long.MAX_VALUE).

lhotari:
@thetumbled The title of any PR containing PIP documentation should include `[pip]` to distinguish it from other types of PRs. I made that change to the title.

lhotari (changes requested):
The PIP-393 document should include the high level plan of avoiding to increase the size of the Pulsar client by the size of fastutil jar file. The fastutil jar file is very large, 23MB. We use only a few classes of fastutil. There's fastutil-core library which is smaller, about ≅6MB. However, that is also relatively large and using fastutil-core will introduce another problem on the broker side since there's already fastutil jar which also includes fastutil-core jar classes. It's necessary to design a proper shading solution as part of this PIP design and implementation.
More details in the thread https://github.com/apache/pulsar/pull/23600#discussion_r1858069707

thetumbled:
> The PIP-393 document should include the high level plan of avoiding to increase the size of the Pulsar client by the size of fastutil jar file. The fastutil jar file is very large, 23MB. We use only a few classes of fastutil. There's fastutil-core library which is smaller, about ≅6MB. However, that is also relatively large and using fastutil-core will introduce another problem on the broker side since there's already fastutil jar which also includes fastutil-core jar classes. It's necessary to design a proper shading solution as part of this PIP design and implementation. More details in the thread [#23600 (comment)](https://github.com/apache/pulsar/pull/23600#discussion_r1858069707)

Thanks for review, i add it in high level design.

thetumbled:
The vote is completed, please review this pr again, thanks. @lhotari @nodece @eolivelli

pip/pip-393.md
```diff
@@ -0,0 +1,226 @@
...
+There are many issues with the current implementation of Negative Acknowledgement in Pulsar:
+- the memory occupation is high.
+- the code execution efficiency is low.
+- the redelivery time is not accurate.
+- multiple negative ack for messages in the same entry(batch) will interfere with each other.
+All of these problem is severe and need to be solved.
+
+## Memory occupation is high
+After the improvement of https://github.com/apache/pulsar/pull/23582, we have reduce half more memory occupation
+of `NegativeAcksTracker` by replacing `HashMap` with `ConcurrentLongLongPairHashMap`. With 100w entry, the memory
```
lhotari, line 23: what unit is `w` in "100w entry"? How about "1kw" entry?

pip/pip-393.md
```diff
@@ -0,0 +1,226 @@
...
+occupation decrease from 178Mb to 64Mb. With 1kw entry, the memory occupation decrease from 1132Mb to 512Mb.
```
lhotari, line 24: Use "MB" instead of "Mb" if you are referring to megabytes instead of megabits.

pip/pip-393.md
```diff
@@ -0,0 +1,226 @@
...
+The average memory occupation of each entry decrease from 1132MB/10000000=118byte to 512MB/10000000=53byte.
+
+But it is not enough. Assuming that we negative ack message 1w/s, assigning 1h redelivery delay for each message,
```
lhotari, line 27: What is "1w/s"?

lhotari (approved):
LGTM, thanks for driving the effort @thetumbled. Great work!

lhotari:
release labels shouldn't be added to PIP document PRs since we only maintain PIP documents in the master branch. release labels are used for cherry-picking.

PR #23602, opened 2024-11-15
[fix][broker] Fix failed TokenAuthenticatedProducerConsumerTest

BewareMyPower:
### Motivation

`TokenAuthenticatedProducerConsumerTest` cannot pass the test. The root cause is that when a namespace does not exist and the client carries a token without correct permission, the client would fail with `AuthorizationException` but not `TopicDoesNotExistException`.

It's because `PulsarAuthorizationProvider#validateTenantAdminAccess` will fail exceptionally by `NotFoundException` here: https://github.com/apache/pulsar/blob/0b0eef905c3a6ebe8a6fbb284743d98c8e97d5b8/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/PulsarAuthorizationProvider.java#L718

However, when the namespace does not exist, this exception won't be thrown. If `testTopicNotFound` was running after `testTokenProducerAndConsumer`, the "my-property" tenant would be created so it would fail.

However, the CI somehow passed and I see 1 test was skipped.

```
Warning:  Tests run: 6, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 22.034 s - in org.apache.pulsar.client.api.TokenAuthenticatedProducerConsumerTest
```

### Modifications

Change the test from `testTopicNotFound` to `testTenantNotExist` and use a tenant that does not exist.

Regarding the case when the namespace does not exist, we might need another PR for it.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/TokenAuthenticatedProducerConsumerTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/TokenAuthenticatedProducerConsumerTest.java
@@ -191,14 +191,14 @@ public class TokenAuthenticatedProducerConsumerTest extends ProducerConsumerBase
     }
 
     @Test(dataProvider = "provider")
-    public void testTopicNotFound(boolean useTcpServiceUrl, boolean useCorrectToken) throws Exception {
+    public void testTenantNotExist(boolean useTcpServiceUrl, boolean useCorrectToken) throws Exception {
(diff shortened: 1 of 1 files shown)


lhotari (approved):
LGTM

shibd (approved):

RobertIndie (approved):

PR #23611, opened 2024-11-19
[improve][broker] Reduce memory occupation of the delayed message queue

thetumbled:
### Motivation

Delayed queue consume a lot of memory to store the message id. We can reduce the memory occupation with new data structure simillar to https://github.com/apache/pulsar/pull/23601.

### Modifications

Use new data structure to reduce memory occupation of the delayed message queue, `InMemoryDelayedDeliveryTracker`.

List some experiment data with following test code:
```
    static long trimLowerBit(long timestamp, int bits) {
        return timestamp & (-1L << bits);
    }

    public static void main(String[] args) throws IOException {
        TripleLongPriorityQueue queue = new TripleLongPriorityQueue();
        Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> map2 = new Long2ObjectAVLTreeMap<>();

        long numMessages = 10000000, numLedgers=1000;
        long numEntries = numMessages/numLedgers;
        long ledgerId, entryId, partitionIndex, timestamp=System.currentTimeMillis();
        for(long i=0; i<numLedgers; i++) {
            ledgerId = 10000+i;
            for(long j=0; j<numEntries; j++) {
                entryId = 10000+j;
                partitionIndex = 0;
                // 1ms per message
                timestamp++;
                queue.add(timestamp, ledgerId, entryId);
                
                timestamp = trimLowerBit(timestamp, 8);
                if (map2.containsKey(timestamp)) {
                    Long2ObjectMap<Roaring64Bitmap> ledgerMap = map2.get(timestamp);
                    if (ledgerMap.containsKey(ledgerId)) {
                        ledgerMap.get(ledgerId).add(entryId);
                    } else {
                        Roaring64Bitmap entrySet = new Roaring64Bitmap();
                        entrySet.add(entryId);
                        ledgerMap.put(ledgerId, entrySet);
                    }
                } else {
                    Long2ObjectMap<Roaring64Bitmap> ledgerMap = new Long2ObjectOpenHashMap<>();
                    Roaring64Bitmap entrySet = new Roaring64Bitmap();
                    entrySet.add(entryId);
                    ledgerMap.put(ledgerId, entrySet);
                    map2.put(timestamp, ledgerMap);
                }
            }
        }
        
        System.out.println("queue size in MB: " + queue.bytesCapacity() / 1024.0 / 1024.0);
        try {
            Thread.sleep(10000000);
            // TODO Auto-generated catch block
            e.printStackTrace();
        }
    }
```
Produce 1kw entries with 1k ledgers, with the speed of 1 msg/ms.

- Original data structure: TripleLongPriorityQueue
It store the info in the direct memory, we can call `queue.bytesCapacity()` to get the size of memory it take.
image
So, TripleLongPriorityQueue take up 240MB.

- New data structure: `Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>>`
It store the data in heap, so we can use `jmap` to dump the memory to analyze.
image
So, it take up only 8MB, down to only 8/240=3.33% of the original size!

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/65

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -20,23 +20,21 @@ package org.apache.pulsar.broker.delayed;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.util.Timer;
-import java.time.Clock;
-import java.util.NavigableSet;
-import java.util.TreeSet;
-import java.util.concurrent.CompletableFuture;
-
 import it.unimi.dsi.fastutil.longs.Long2ObjectAVLTreeMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectSortedMap;
 import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import it.unimi.dsi.fastutil.longs.LongSet;
+import java.time.Clock;
+import java.util.NavigableSet;
+import java.util.TreeSet;
+import java.util.concurrent.CompletableFuture;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.Position;
 import org.apache.bookkeeper.mledger.PositionFactory;
 import org.apache.pulsar.broker.service.persistent.AbstractPersistentDispatcherMultipleConsumers;
-import org.apache.pulsar.common.util.collections.TripleLongPriorityQueue;
 import org.roaringbitmap.longlong.Roaring64Bitmap;
 
 @Slf4j
@@ -44,8 +42,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
 
     // timestamp -> ledgerId -> entryId
     // AVL tree -> OpenHashMap -> RoaringBitmap
-    protected final Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> priorityQueue
-            = new Long2ObjectAVLTreeMap<>();
+    protected final Long2ObjectSortedMap<Long2ObjectMap<Roaring64Bitmap>> priorityQueue = new Long2ObjectAVLTreeMap<>();
 
     // If we detect that all messages have fixed delay time, such that the delivery is
     // always going to be in FIFO order, then we can avoid pulling all the messages in
@@ -99,7 +96,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
             tickTimeMillis >>= 1;
             bitCnt++;
         }
-        return bitCnt-1;
+        return bitCnt - 1;
     }
 
     /**
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -61,7 +61,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     private boolean messagesHaveFixedDelay = true;
 
     // The bit count to trim to reduce memory occupation.
-    private int timestampPrecisionBitCnt = 0;
+    private final int timestampPrecisionBitCnt;
 
     InMemoryDelayedDeliveryTracker(AbstractPersistentDispatcherMultipleConsumers dispatcher, Timer timer,
                                    long tickTimeMillis,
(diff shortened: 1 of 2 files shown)


dao-jun:
Good work, and I'm just wandering how much memories can we reduce?

thetumbled:
> Good work, and I'm just wandering how much memories can we reduce?

I will do some experiment later.

lhotari (approved):
LGTM. Great improvement! Thanks for the contribution @thetumbled

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -52,12 +60,16 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     // Track whether we have seen all messages with fixed delay so far.
     private boolean messagesHaveFixedDelay = true;
 
+    // The bit count to trim to reduce memory occupation.
+    private int timestampPrecisionBitCnt = 0;
```
lhotari, line 64: could this be a final field?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -122,17 +166,40 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
...
+                long ledgerId = ledgerEntry.getLongKey();
+                Roaring64Bitmap entryIds = ledgerEntry.getValue();
+                if (entryIds.getLongCardinality() <= n) {
+                    entryIds.forEach(entryId -> {
+                        positions.add(PositionFactory.create(ledgerId, entryId));
+                    });
+                    n -= (int) entryIds.getLongCardinality();
+                    ledgerIdToDelete.add(ledgerId);
+                } else {
+                    long[] entryIdsArray = entryIds.toArray();
```
lhotari, line 186: calling `toArray()` isn't a great idea when using RoaringBitmap. It's better to use the iterator available with `getLongIterator()` method to minimize allocations.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -122,17 +166,40 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
...
+                    for (int i = 0; i < n; i++) {
+                        positions.add(PositionFactory.create(ledgerId, entryIdsArray[i]));
+                        entryIds.removeLong(entryIdsArray[i]);
+                    }
+                    n = 0;
```
lhotari, line 191: What if `entryIds().getLongCardinality() < n`. That case seems to be broken.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -157,18 +224,27 @@ public CompletableFuture<Void> clear() {
...
+     * overestimate the memory usage of the buffer a lot.
+     * @return the memory usage of the buffer
+     */
     @Override
     public long getBufferMemoryUsage() {
-        return priorityQueue.bytesCapacity();
+        return priorityQueue.values().stream().mapToLong(
+                ledgerMap -> ledgerMap.values().stream().mapToLong(
+                        Roaring64Bitmap::getLongSizeInBytes).sum()).sum();
     }
```
lhotari, line 243: Does this method get called in the same thread as other InMemoryDelayedDeliveryTracker methods? If not, there's a need to make the methods synchronized or add a ReadWriteLock to fix thread safety.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: updated.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: It is afraid that we can't update the bitmap while iterating it, as LongIterator don't provide remove method.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: `TripleLongPriorityQueue` is not thread safe too.
`getBufferMemoryUsage` is used to print the stat used by `InMemoryDelayedDeliveryTracker`, which don't require too much accuracy. 
What is more, the method `Roaring64Bitmap::getLongSizeInBytes` is not accurate too, as stated in the comment.

dao-jun (approved):

PR #23621, opened 2024-11-21
[improve][broker] TopicStats specify serialized field name

crossoverJie:
### Motivation

image
TopicStats API contains duplicate fields.

### Modifications

Use `@JsonProperty("isReplicated")` to specify serialization names.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/stats/SubscriptionStatsImpl.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/stats/SubscriptionStatsImpl.java
@@ -19,6 +19,7 @@
 package org.apache.pulsar.common.policies.data.stats;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -112,9 +113,11 @@ public class SubscriptionStatsImpl implements SubscriptionStats {
     public List<ConsumerStatsImpl> consumers;
 
     /** Tells whether this subscription is durable or ephemeral (eg.: from a reader). */
+    @JsonProperty("isDurable")
     public boolean isDurable;
 
     /** Mark that the subscription state is kept in sync across different regions. */
+    @JsonProperty("isReplicated")
     public boolean isReplicated;
 
     /** Whether out of order delivery is allowed on the Key_Shared subscription. */


lhotari:
@crossoverJie Which Pulsar client and broker version did you experience this with?

lhotari (changes requested):
Please add tests for this change. It's also possible that this is a breaking change which requires more thought to address it.

pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/stats/SubscriptionStatsImpl.java
```diff
@@ -115,9 +116,11 @@ public class SubscriptionStatsImpl implements SubscriptionStats {
     public List<ConsumerStatsImpl> consumers;
 
     /** Tells whether this subscription is durable or ephemeral (eg.: from a reader). */
+    @JsonProperty("isDurable")
     public boolean isDurable;
 
     /** Mark that the subscription state is kept in sync across different regions. */
+    @JsonProperty("isReplicated")
     public boolean isReplicated;
```
lhotari, line 124: I think that the properties with `is` prefix shouldn't ever end up in the JSON output. We should address that problem instead and keep `durable` and `replicated` in the output instead of making them `isDurable` and `isReplicated`.

dao-jun: +1 with lari

crossoverJie: Thanks for the suggestion, I has been modified and it is a breaking change.

dao-jun (approved):
lgtm

crossoverJie:
@lhotari please take a look

lhotari (changes requested):
https://github.com/apache/pulsar/pull/23621#discussion_r1851531766 . We don't want to introduce breaking changes or promote the fields starting with "is"?
Is the duplicate field the only issue?

crossoverJie:
> [#23621 (comment)](https://github.com/apache/pulsar/pull/23621#discussion_r1851531766) . We don't want to introduce breaking changes or promote the fields starting with "is"? Is the duplicate field the only issue?

Yes, just remove the `is` prefix.

Issue #23622, opened 2024-11-21
[Bug][client] Consumer incomingMessageSize and client memory usage is negative


TakaHiR07:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

master

### Minimal reproduce step

use a simple demo to consume. Found that client memory usage is negative, and ConsumerBase#incomingMessagesSize is also negative.

```
Consumer<byte[]> consumer = client.newConsumer(Schema.BYTES)
                .topic("persistent://test/normal/test")
                .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
                .subscriptionName("test")
                .subscriptionType(SubscriptionType.Shared)
                .autoScaledReceiverQueueSizeEnabled(true)
                .subscribe();

while (true) {
    Message msg = consumer.receive();
}
```

### What did you expect to see?

incomingMessageSize and client memory usage should not be negative

### What did you see instead?

incomingMessageSize and client memory usage is negative

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #23624, opened 2024-11-21
[fix][client] fix incomingMessageSize and client memory usage is negative

TakaHiR07:
Fixes https://github.com/apache/pulsar/issues/23622

### Motivation

After testing for autoScaledReceiverQueueSizeEnabled config, found that the client memory usage and incomingMessageSize is negative.

This error result in two problem:
1. autoScaledReceiverQueueSize feature is not correct. The receive queue size is expanded based on wrong memory usage.
2. batchReceivePolicy feature is not correct. The MaxNumBytes config in batchReceivePolicy is judged by wrong incomingMessageSize.

By the way, batchReceivePolicy is default used in multiTopicConsumer since pulsar-2.9.4, while autoScaledReceiverQueueSize is a  optional config in consumer.

The reason is because when enter ConsumerImpl#notifyPendingReceivedCallback, the message would directly pass to pendingReceive request. Then consumerImpl just decrease the incomingMessageSize, but not increase. So the value become negative

### Modifications

fix the error case that only do decrease the size, but not increase.

### Verifying this change

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/AutoScaledReceiverQueueSizeTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/AutoScaledReceiverQueueSizeTest.java
@@ -20,14 +20,22 @@ package org.apache.pulsar.client.impl;
 
 
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import lombok.Cleanup;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.broker.auth.MockedPulsarServiceBaseTest;
 import org.apache.pulsar.client.admin.PulsarAdminException;
 import org.apache.pulsar.client.api.BatchReceivePolicy;
+import org.apache.pulsar.client.api.Consumer;
+import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.Producer;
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM, great work @TakaHiR07

dao-jun (approved):
Nice catch, good work

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/AutoScaledReceiverQueueSizeTest.java
```diff
@@ -257,4 +266,61 @@ public void testMultiConsumerImplBatchReceive() throws PulsarClientException, Pu
...
+        Consumer<byte[]> consumer = pulsarClient.newConsumer()
+                .topic(topicName)
+                .subscriptionName(subName)
+                .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
+                .autoScaledReceiverQueueSizeEnabled(true)
+                .subscribe();
+
+        Field field = PulsarClientImpl.class.getDeclaredField("memoryLimitController");
+        field.setAccessible(true);
+        MemoryLimitController controller = (MemoryLimitController) field.get(pulsarClient);
```
nodece, line 302:
```suggestion
        MemoryLimitController controller = ((PulsarClientImpl) pulsarClient).getMemoryLimitController();
```

TakaHiR07: Thanks for checking

nodece (approved):

PR #17962, opened 2022-10-07
[feat] [broker] PIP-188 support blue-green cluster migration [part-1]

rdhabalia:
### Motivation
It fixes : https://github.com/apache/pulsar/issues/16551
Cluster migration or Blue-Green cluster deployment is one of the proven solutions to migrate live traffic from one cluster to another. One of the examples is applications running on Kubernetes sometimes require a Kubernetes cluster upgrade which can cause downtime for the entire application during a Kubernetes cluster upgrade. Blue-green deployment is an application release model that gradually transfers user traffic from a previous version of an app or microservice to a nearly identical new release—both of which are running in production.

The old version can be called the blue environment while the new version can be known as the green environment. Once production traffic is fully transferred from blue to green, blue can standby in case of rollback or be pulled from production and updated to become the template upon which the next update is made.

We need such capability in Apache pulsar to migrate live traffic from the blue cluster to the green cluster so, eventually, the entire traffic moves from the blue cluster to the green cluster without causing downtime for the topics.

### Modification
- Handle cluster migration for persistent/non-persistent topics
- Add CLI and REST API support to mark cluster migration

**NOTE**: This is the 1st part of the feature. 2nd PR will be submitted to handle Replicator changes which can handle cluster migration without compromising ordering and data persistent guarantee.

### Result
User can migrate Pulsar cluster without user traffic interruption.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The binary protocol
- [x] The REST endpoints
- [x] The admin CLI options
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [x] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23632, opened 2024-11-22
[fix][client] Make protobuf-java dependency optional in java client libraries

### LINKED ISSUE #23469: [Bug] protobuf-java should be an optional dependency for Pulsar Java client libraries
> IMO, making protobuf-java optional is beyond the scope of this PR. It’s something worth discussing separately at a later stage.

It was always meant to be optional. It's already optional in the unshaded module. It was included as required dependency by mistake and this is why it's getting included in the shaded jar. 

It don't think it's beyond the scope to fix the root cause of the issue.

_Originally posted by @merlimat in https://github.com/apache/pulsar/pull/23468#discussion_r1803518970_

Shawyeok:
Fixes #23469

### Motivation

In the Pulsar Java client libraries, the `protobuf-java` library is only used in `ProtobufSchema`, making it suitable as an optional dependency. However, it is mistakenly included in the `pulsar-client-admin` and `pulsar-client-all` shaded jar and declared as required dependency of `pulsar-client`, `pulsar-client-original` and `pulsar-client-admin-original`, as noted by @merlimat in [this discussion](https://github.com/apache/pulsar/pull/23468#discussion_r1803518970). 

PR #23468 moved `protobuf-java` from a shaded JAR to a declared dependency in the `pom.xml`, which helped avoid conflicts between `protobuf-java` and application dependencies. Since users only need `protobuf-java` when using `ProtobufSchema`, it makes sense to declare it as an optional dependency. This change reduces the number of dependencies in the Pulsar client libraries.

### Verifying this change

1. Execute the following Maven command:
  ```
  mvn clean install -am -pl pulsar-client-all,pulsar-client-shaded,pulsar-client-admin-shaded -DskipTests
  ```

2. Verify that the `protobuf-java` dependency is marked `optional` in the installed `pom.xml` files, e.g.
```
<dependency>
  <groupId>com.google.protobuf</groupId>
  <artifactId>protobuf-java</artifactId>
  <version>3.25.5</version>
  <scope>compile</scope>
  <optional>true</optional>
</dependency>
```

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/Shawyeok/pulsar/pull/19

### DIFF
--- a/distribution/shell/src/assemble/LICENSE.bin.txt
+++ b/distribution/shell/src/assemble/LICENSE.bin.txt
@@ -429,10 +429,6 @@ MIT License
  * The Checker Framework
     - checker-qual-3.33.0.jar
 
-Protocol Buffers License
- * Protocol Buffers
-   - protobuf-java-3.25.5.jar -- ../licenses/LICENSE-protobuf.txt
-
 CDDL-1.1 -- ../licenses/LICENSE-CDDL-1.1.txt
  * Java Annotations API
     - javax.annotation-api-1.3.2.jar
--- a/pulsar-client-admin-shaded/pom.xml
+++ b/pulsar-client-admin-shaded/pom.xml
@@ -44,6 +44,11 @@
       <version>${project.parent.version}</version>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <optional>true</optional>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/pulsar-client-admin-shaded/pom.xml
+++ b/pulsar-client-admin-shaded/pom.xml
@@ -44,11 +44,6 @@
       <version>${project.parent.version}</version>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>com.google.protobuf</groupId>
-      <artifactId>protobuf-java</artifactId>
-      <optional>true</optional>
-    </dependency>
   </dependencies>
   <build>
     <plugins>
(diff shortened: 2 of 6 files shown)


lhotari (approved):
LGTM. Thanks for addressing this. This could be a minor breaking change for some clients since they would now have to include the protobuf-java version that they prefer. We'll just have to add a release note that this is change is made. Since it's a minor breaking change and not many have upgraded to Pulsar 4.0.x, I think it makes sense to get this included in 4.0.x .

lhotari:
Build failure.

https://github.com/apache/pulsar/actions/runs/11977798414/job/3
3566154951?pr=23632#step:11:1

> Run src/check-binary-license.sh ./distribution/server/target/apache-pulsar-*-bin.tar.gz && src/check-binary-license.sh ./distribution/shell/target/apache-pulsar-shell-*-bin.tar.gz
> protobuf-java-3.25.5.jar mentioned in LICENSE, but not bundled

I'd guess that it's the `distribution/shell/src/assemble/LICENSE.bin.txt` file which should be updated or protobuf-java should be included as a dependency for the shell. I wonder if it's needed there? If not, the LICENSE.bin.txt file should be updated by removing the entry.

Shawyeok:
> I'd guess that it's the distribution/shell/src/assemble/LICENSE.bin.txt file which should be updated or protobuf-java should be included as a dependency for the shell. I wonder if it's needed there? If not, the LICENSE.bin.txt file should be updated by removing the entry.

@lhotari 

Yes, the `protobuf-java` dependency should be removed as well.

The `protobuf-java` dependency in the `pulsar-common` module was introduced in #17962. It seems unnecessary now, and removing it would be a cleaner approach. Please take a look.

PR #23634, opened 2024-11-24
[improvement][broker] If there is a deadlock in the service, the probe should return a failure because the service may be unavailable

yyj8:
Fixes https://github.com/apache/pulsar/issues/23635

Main Issue: #xyz

PIP: #xyz 

### Motivation
In some special scenarios, when the broker service has a deadlock, it needs to be able to automatically recover instead of requiring manual intervention. For example, when the service is deployed in a customer environment, we cannot directly manage it. If the service has a deadlock, the k8s probe should return a failure because the service may be unavailable. The probe failure triggers a broker pod restart to resolve the deadlock.

### Modifications
Add deadlock detection in the probe. If a deadlock exists, print the thread stack and return a service unavailable exception.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: 
https://github.com/yyj8/pulsar/pull/10


lhotari (changes requested):
There's already a deadlock check in the health check:
https://github.com/apache/pulsar/blob/c9de1bbefd410152c7da6f9e8d5a5d59ab9cbead/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/BrokersBase.java#L366-L408

It also contains an example of how to check deadlocks.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -52,7 +54,15 @@ public String checkStatus() {
         if (statusFilePath != null) {
             File statusFile = new File(statusFilePath);
             if (isReady && statusFile.exists() && statusFile.isFile()) {
-                return "OK";
+                // check deadlock
+                String diagnosticResult = ThreadDumpUtil.buildThreadDiagnosticString();
+                if (StringUtils.isBlank(diagnosticResult)) {
```
lhotari, line 59: `ThreadDumpUtil.buildThreadDiagnosticString` will alway return a non-blank String. 
There's an example of detecting a deadlock using JMX API in https://github.com/apache/pulsar/blob/c9de1bbefd410152c7da6f9e8d5a5d59ab9cbead/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/BrokersBase.java#L390-L408 .
If we were to add the deadlock check to the status check, it should be possible to configure the interval for running the check since there's a cost for running the deadlock check. It should also be possible to disable the deadlock check completely.

yyj8: > `ThreadDumpUtil.buildThreadDiagnosticString` will alway return a non-blank String. There's an example of detecting a deadlock using JMX API in
> 
> https://github.com/apache/pulsar/blob/c9de1bbefd410152c7da6f9e8d5a5d59ab9cbead/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/BrokersBase.java#L390-L408
> 
> .
> If we were to add the deadlock check to the status check, it should be possible to configure the interval for running the check since there's a cost for running the deadlock check. It should also be possible to disable the deadlock check completely.

@lhotari lhotari
Thank you very much for helping review the code. I have completed the modifications, please help me review again.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -55,13 +64,27 @@ public String checkStatus() {
...
+                            .collect(Collectors.joining(", "));
+                    if (System.currentTimeMillis() - threadDumpLoggedTimestamp
+                            > LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED) {
+                        String diagnosticResult = ThreadDumpUtil.buildThreadDiagnosticString();
+                        log.error("Deadlock detected, service may be unavailable, "
+                                + "thread stack details are as follows: {}.", diagnosticResult);
+                        threadDumpLoggedTimestamp = System.currentTimeMillis();
+                    } else {
+                        log.error("Deadlocked threads detected. {}", threadNames);
+                    }
```
lhotari, line 84: Since the status endpoint doesn't have authentication, it will be necessary to have a solution to prevent introducing a new DoS vulnerability where calling the status endpoint in a tight loop could introduce significant load to the system. One way would be to check that the deadlock check is executed only when there's more than 1 seconds from the previous check. If it's less than that, the previous result of the deadlock check would be reused.

yyj8: > Since the status endpoint doesn't have authentication, it will be necessary to have a solution to prevent introducing a new DoS vulnerability where calling the status endpoint in a tight loop could introduce significant load to the system. One way would be to check that the deadlock check is executed only when there's more than 1 seconds from the previous check. If it's less than that, the previous result of the deadlock check would be reused.

This is a very good proposal, and there is indeed a risk of being attacked by DoS. The code logic has been adjusted accordingly. Please help review the code again.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -47,13 +47,33 @@ public class VipStatus {
...
-    private static volatile long threadDumpLoggedTimestamp;
+    private static volatile long lastCheckStatusTimestamp;
+
+    // Since the status endpoint doesn't have authentication, it will be necessary to have a solution to prevent
+    // introducing a new DoS vulnerability where calling the status endpoint in a tight loop could introduce
+    // significant load to the system. One way would be to check that the deadlock check is executed only
+    // when there's more than 1 seconds from the previous check.
+    // If it's less than that, the previous result of the deadlock check would be reused.
+    private static final long DEADLOCK_DETECTED_INTERVAL = 1000L;
+    private static volatile boolean brokerIsHealthy = true;
```
lhotari, line 58: I'd suggest renaming this field so that it's more specific
```suggestion
    private static volatile boolean lastCheckStatusResult;
```

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -47,13 +47,33 @@ public class VipStatus {
...
     private static final long LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED = 600000L;
-    private static volatile long threadDumpLoggedTimestamp;
+    private static volatile long lastCheckStatusTimestamp;
+
+    // Since the status endpoint doesn't have authentication, it will be necessary to have a solution to prevent
+    // introducing a new DoS vulnerability where calling the status endpoint in a tight loop could introduce
+    // significant load to the system. One way would be to check that the deadlock check is executed only
+    // when there's more than 1 seconds from the previous check.
+    // If it's less than that, the previous result of the deadlock check would be reused.
+    private static final long DEADLOCK_DETECTED_INTERVAL = 1000L;
```
lhotari, line 57: Since the logic impacts the complete `checkStatus` result. I'd suggest renaming this field. It could also be considered to reduce the value from 1000 to 500.
```suggestion
    private static final long CHECK_STATUS_INTERVAL = 500L;
```

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -47,13 +47,33 @@ public class VipStatus {
...
     // to prevent excessive logging
     private static final long LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED = 600000L;
-    private static volatile long threadDumpLoggedTimestamp;
+    private static volatile long lastCheckStatusTimestamp;
+
+    // Since the status endpoint doesn't have authentication, it will be necessary to have a solution to prevent
+    // introducing a new DoS vulnerability where calling the status endpoint in a tight loop could introduce
+    // significant load to the system. One way would be to check that the deadlock check is executed only
+    // when there's more than 1 seconds from the previous check.
+    // If it's less than that, the previous result of the deadlock check would be reused.
```
lhotari, line 56:
```suggestion
    // Rate limit status checks to every 500ms to prevent DoS
```

lhotari (changes requested):
Good work @yyj8. Some suggestions for field naming and simplifying the code comment.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -55,7 +55,7 @@ public class VipStatus {
     // when there's more than 1 seconds from the previous check.
     // If it's less than that, the previous result of the deadlock check would be reused.
     private static final long DEADLOCK_DETECTED_INTERVAL = 1000L;
-    private static volatile boolean brokerIsHealthy = true;
+    private static volatile boolean lastCheckStatusResult = true;
```
lhotari, line 58: Please remove `= true`. There isn't a need to initialize this field to `true`.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -38,25 +44,75 @@ public class VipStatus {
     public static final String ATTRIBUTE_STATUS_FILE_PATH = "statusFilePath";
     public static final String ATTRIBUTE_IS_READY_PROBE = "isReadyProbe";
 
+    // log a full thread dump when a deadlock is detected in status check once every 10 minutes
+    // to prevent excessive logging
+    private static final long LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED = 600000L;
+    private static volatile long lastCheckStatusTimestamp;
```
lhotari, line 50: these shouldn't be static fields. This is a different case than the `BrokersBase` class where it was necessary.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -38,25 +44,75 @@ public class VipStatus {
...
     @Context
     protected ServletContext servletContext;
 
     @GET
     public String checkStatus() {
-        String statusFilePath = (String) servletContext.getAttribute(ATTRIBUTE_STATUS_FILE_PATH);
-        @SuppressWarnings("unchecked")
-        Supplier<Boolean> isReadyProbe = (Supplier<Boolean>) servletContext.getAttribute(ATTRIBUTE_IS_READY_PROBE);
+        // Locking classes to avoid deadlock detection in multi-thread concurrent requests.
+        synchronized (VipStatus.class) {
```
lhotari, line 66: since the fields should be instance fields, simply make the method `synchronized` or use a limited `synchronized(this)` block as there was before.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
lhotari: Make this an instance field by removing `static`.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
yyj8: > Since the logic impacts the complete `checkStatus` result. I'd suggest renaming this field. It could also be considered to reduce the value from 1000 to 500.

This is a great proposal, the meaning of the fields looks clearer and more specific and I will modify it.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
yyj8: > 

This is a great proposal,  and I will modify it.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
yyj8: I tested locally and printed the current object reference every time the `checkStatus () `method was called, and found that a new object was created each time. And it was found that if the `lastCheckStatusTimestamp ` field is not decorated with `static`, it is initialized to 0 every time the` checkStatus ()` method is called, which does not match our expected target of executing the detection at a specified interval.

The printed log is shown below:
```
org.apache.pulsar.common.configuration.VipStatus@65403bed
2024-11-27T11:17:15,947 - INFO  - [pulsar-web-41-3:Slf4jRequestLogWriter@62] - [0:0:0:0:0:0:0:1] - - [27/11月/2024:11:17:15 +0800] "GET /status.html HTTP/1.1" 200 2 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36" 267
org.apache.pulsar.common.configuration.VipStatus@d560619
2024-11-27T11:17:24,500 - INFO  - [pulsar-web-41-9:Slf4jRequestLogWriter@62] - [0:0:0:0:0:0:0:1] - - [27/11月/2024:11:17:24 +0800] "GET /status.html HTTP/1.1" 200 2 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36" 6
org.apache.pulsar.common.configuration.VipStatus@646ada8f
2024-11-27T11:17:57,562 - INFO  - [pulsar-web-41-8:Slf4jRequestLogWriter@62] - [0:0:0:0:0:0:0:1] - - [27/11月/2024:11:17:57 +0800] "GET /status.html HTTP/1.1" 200 2 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36" 3

```

So, I suggest using the `static ` keyword to decorated the `lastCheckStatusTimestamp` field. Use `synchronized (VipStatus.class)` for global synchronization to avoid DoS attacks caused by multi-threaded concurrent access.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
yyj8: Completed

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
yyj8: Completed

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -38,25 +44,71 @@ public class VipStatus {
...
+            if (System.currentTimeMillis() - lastCheckStatusTimestamp < CHECK_STATUS_INTERVAL) {
+                lastCheckStatusTimestamp = System.currentTimeMillis();
```
lhotari, line 64: I think that this line should be removed

lhotari (changes requested):

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
yyj8: > I think that this line should be removed since the purpose of the field value is to record the timestamp when an actual check was made and rate limit that.

Your suggestion is not to need synchronization, right? If there is no code synchronization, will there be a massive number of DoS attacks that execute deadlock detection logic simultaneously, causing a surge in node load.

lhotari: I hope this clarifies: https://github.com/apache/pulsar/pull/23634/files#r1860206850

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -38,25 +44,71 @@ public class VipStatus {
...
+                if (lastCheckStatusResult) {
+                    return "OK";
+                } else {
+                    throw new WebApplicationException(Status.SERVICE_UNAVAILABLE);
+                }
+            }
+            lastCheckStatusTimestamp = System.currentTimeMillis();
```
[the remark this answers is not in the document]
yyj8: The final global code logic content is as follows:
```java
public class VipStatus {

    public static final String ATTRIBUTE_STATUS_FILE_PATH = "statusFilePath";
    public static final String ATTRIBUTE_IS_READY_PROBE = "isReadyProbe";

    // log a full thread dump when a deadlock is detected in status check once every 10 minutes
    // to prevent excessive logging
    private static final long LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED = 600000L;
    private static volatile long lastCheckStatusTimestamp;

    // Rate limit status checks to every 500ms to prevent DoS
    private static final long CHECK_STATUS_INTERVAL = 500L;
    private static volatile boolean lastCheckStatusResult;

    @Context
    protected ServletContext servletContext;

    @GET
    public String checkStatus() {
        // Locking classes to avoid deadlock detection in multi-thread concurrent requests.
        synchronized (VipStatus.class) {
            if (System.currentTimeMillis() - lastCheckStatusTimestamp < CHECK_STATUS_INTERVAL) {
                if (lastCheckStatusResult) {
                    return "OK";
                } else {
                    throw new WebApplicationException(Status.SERVICE_UNAVAILABLE);
                }
            }
            lastCheckStatusTimestamp = System.currentTimeMillis();

            String statusFilePath = (String) servletContext.getAttribute(ATTRIBUTE_STATUS_FILE_PATH);
            @SuppressWarnings("unchecked")
            Supplier<Boolean> isReadyProbe = (Supplier<Boolean>) servletContext.getAttribute(ATTRIBUTE_IS_READY_PROBE);

            boolean isReady = isReadyProbe != null ? isReadyProbe.get() : true;

            if (statusFilePath != null) {
                File statusFile = new File(statusFilePath);
                if (isReady && statusFile.exists() && statusFile.isFile()) {
                    // check deadlock
                    ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
                    long[] threadIds = threadBean.findDeadlockedThreads();
                    if (threadIds != null && threadIds.length > 0) {
                        ThreadInfo[] threadInfos = threadBean.getThreadInfo(threadIds, false,
                                false);
                        String threadNames = Arrays.stream(threadInfos)
                                .map(threadInfo -> threadInfo.getThreadName()
                                        + "(tid=" + threadInfo.getThreadId() + ")")
                                .collect(Collectors.joining(", "));
                        if (System.currentTimeMillis() - lastCheckStatusTimestamp
                                > LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED) {
                            String diagnosticResult = ThreadDumpUtil.buildThreadDiagnosticString();
                            log.error("Deadlock detected, service may be unavailable, "
                                    + "thread stack details are as follows: {}.", diagnosticResult);
                        } else {
                            log.error("Deadlocked threads detected. {}", threadNames);
                        }
                        lastCheckStatusResult = false;
                        throw new WebApplicationException(Status.SERVICE_UNAVAILABLE);
                    } else {
                        lastCheckStatusResult = true;
                        return "OK";
                    }
                }
            }
            lastCheckStatusResult = false;
            log.warn("Failed to access \"status.html\". The service is not ready");
            throw new WebApplicationException(Status.NOT_FOUND);
        }
    }

}
```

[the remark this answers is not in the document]
lhotari: Adding tests will be useful since validating logic just by looking at the code isn't a good practice. You might have to modify some details to inject mocks. For example, there could be a separate constructor for tests where the `ThreadTXBean` is injected. In the default constructor used in production code, `ManagementFactory,getThreadMXBean()` method call could be used to pass the value. In test code, a mock could be passed. It's also useful to refactor the code to use a `java.time.Clock.millis()` instead of `System.currentTimeMillis()` and also have the `Clock` instance in the constructor so that test code could pass a mock implementation where it's possible to make time jump forward when validating the logic. `Clock.systemUTC()` could be used in the default constructor for production code.

lhotari:
@yyj8 btw. when you add commits to the PR, it's useful to make the commit title about the change and not copy the PR title into the follow up commits. When the PR is merged, all commits are squashed so they won't end up in the final merged commit. The benefit of the commit messages in the PR commits is that the reviewer will be able to follow the changes.

yyj8:
> @yyj8 btw. when you add commits to the PR, it's useful to make the commit title about the change and not copy the PR title into the follow up commits. When the PR is merged, all commits are squashed so they won't end up in the final merged commit. The benefit of the commit messages in the PR commits is that the reviewer will be able to follow the changes.

Very good suggestion, future code submissions will include detailed instructions for modifying the code.

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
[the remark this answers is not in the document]
yyj8: Unit testing code has been added, please help review it.

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/VipStatusTest.java
```diff
@@ -0,0 +1,139 @@
...
+        File file = new File(statusFilePath);
+        file.deleteOnExit();
+    }
+
+    public void testVipStatusCheckStatusWithoutDeadlock() {
+        assertEquals(vipStatus.checkStatus(), "OK");
+    }
+
+    public void testVipStatusCheckStatusWithDeadlock() {
+        MockDeadlock.startDeadlock();
```
lhotari, line 77: There should also be a way to interrupt the dead lock that is introduced since the threads would stay alive in CI for the duration of the test JVM lifetime.

yyj8: Completed

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/MockServletContext.java
lhotari, line 1: There shouldn't be a need to add this file since Mockito can be used to mock a ServletContext for the test.

yyj8: Very good suggestion, the code is much simpler after using `Mockito`.

It has been changed.

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/VipStatusTest.java
```diff
@@ -0,0 +1,152 @@
...
+@Slf4j
+public class VipStatusTest {
+
+    public static final String ATTRIBUTE_STATUS_FILE_PATH = "statusFilePath";
+    public static final String ATTRIBUTE_IS_READY_PROBE = "isReadyProbe";
+    private static final long LOG_THREADDUMP_INTERVAL_WHEN_DEADLOCK_DETECTED = 10000L;
+    // Rate limit status checks to every 500ms to prevent DoS
+    private static final long CHECK_STATUS_INTERVAL = 500L;
+
+    @Mock
```
lhotari, line 49: For TestNG tests, we don't currently have a solution that makes use of this annotation.

yyj8: At that time, the annotation `@Mock` was only used for annotation purposes, and the actual instantiation was done in: `mockServletContext = Mockito.mock(ServletContext.class);`

This annotation has now been deleted.

lhotari (approved):
LGTM, good work!

lhotari:
Checkstyle failed due to incorrect import order. I pushed a fix for that problem. @yyj8 If you are using IntelliJ/IDEA, please follow instructions at https://pulsar.apache.org/contribute/setup-ide/#configure-code-style to setup the IDE.

yyj8:
> Checkstyle failed due to incorrect import order. I pushed a fix for that problem. @yyj8 If you are using IntelliJ/IDEA, please follow instructions at https://pulsar.apache.org/contribute/setup-ide/#configure-code-style to setup the IDE.

Okay, thank you very much for your help.

lhotari:
Closing and reopening to run a fresh CI build with latest changes from master branch.

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/VipStatusTest.java
```diff
@@ -0,0 +1,150 @@
...
+    private static final long CHECK_STATUS_INTERVAL = 500L;
+
+    private ServletContext mockServletContext;
+    private VipStatus vipStatus;
+
+    @BeforeTest
+    public void setup() throws IOException {
+        String statusFilePath = "/tmp/status.html";
+        File file = new File(statusFilePath);
+        file.createNewFile();
```
lhotari, line 55: It's better to use something like `org.assertj.core.util.Files#newTemporaryFile` for creating the temporary file.

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/VipStatusTest.java
```diff
@@ -0,0 +1,150 @@
...
+        testVipStatusCheckStatusWithoutDeadlock();
+        // There is a deadlock
+        testVipStatusCheckStatusWithDeadlock();
+    }
+
+    @AfterTest
+    public void release() throws IOException {
+        String statusFilePath = "/tmp/status.html";
+        File file = new File(statusFilePath);
+        file.deleteOnExit();
```
lhotari, line 77: this doesn't make sense

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/VipStatusTest.java
yyj8: It has been changed.

pulsar-broker-common/src/test/java/org/apache/pulsar/common/configuration/VipStatusTest.java
yyj8: It has been changed.

lhotari (approved):

pulsar-broker-common/src/main/java/org/apache/pulsar/common/configuration/VipStatus.java
```diff
@@ -38,25 +46,84 @@ public class VipStatus {
...
+                        ThreadInfo[] threadInfos = threadBean.getThreadInfo(threadIds, false,
+                                false);
+                        String threadNames = Arrays.stream(threadInfos)
+                                .map(threadInfo -> threadInfo.getThreadName()
+                                        + "(tid=" + threadInfo.getThreadId() + ")")
+                                .collect(Collectors.joining(", "));
+                        if (clock.millis() - lastPrintThreadDumpTimestamp > printThreadDumpIntervalMs) {
+                            String diagnosticResult = ThreadDumpUtil.buildThreadDiagnosticString();
+                            log.error("Deadlock detected, service may be unavailable, "
+                                    + "thread stack details are as follows: {}.", diagnosticResult);
```
lhotari: It's better that the message is similar in both cases, and so that it includes the thread names. That will make it easier to find the information in logs. Some logging systems might cut off long log entries.
```suggestion
                            log.error("Deadlocked threads detected. {}. Service may be unavailable, "
                                    + "thread stack details are as follows:\n{}", threadNames, diagnosticResult);
```

PR #23636, opened 2024-11-25
[fix][perf] fix performanceConsumer log error throughput

TakaHiR07:
### Motivation

fix error throughput log in performanceConsumer. bytes -> Mb no need to *8.

### Modifications

fix error.

### Verifying this change

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-testclient/src/main/java/org/apache/pulsar/testclient/PerformanceConsumer.java
+++ b/pulsar-testclient/src/main/java/org/apache/pulsar/testclient/PerformanceConsumer.java
@@ -443,7 +443,7 @@ public class PerformanceConsumer extends PerformanceTopicListArguments{
             double elapsed = (now - oldTime) / 1e9;
             long total = totalMessagesReceived.sum();
             double rate = messagesReceived.sumThenReset() / elapsed;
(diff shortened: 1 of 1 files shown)


congbobo184 (commented):
please add a test for it

TakaHiR07:
> > fix error throughput log in performanceConsumer. bytes -> Mb no need to *8.
> 
> The previous calculation is intentional. The previous unit has been megabits instead of megabytes.
> 
> https://github.com/apache/pulsar/blob/002fa49692ff4b8b038bf2edeeb3b32ed218e64f/pulsar-testclient/src/main/java/org/apache/pulsar/testclient/PerformanceConsumer.java#L526

Ok. But I think MB is clearer

lhotari:
> Ok. But I think MB is clearer

@TakaHiR07 I agree that it's confusing to use Mbit/s here and MB/s in many other locations. 
Network bandwidth is commonly expressed as Mbit/s. Perhaps the reason to use Mbit/s here is to match the unit that is used for network bandwidth.

PR #23640, opened 2024-11-26
[improve][broker] Reduce memory occupation of InMemoryRedeliveryTracker.

thetumbled:
### Motivation

Reduce memory occupation of the `InMemoryRedeliveryTracker`.

### Modifications

Use `Long2ObjectMap<Long2IntMap>` to replace `ConcurrentLongLongPairHashMap`.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/66

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
@@ -19,48 +19,98 @@
 package org.apache.pulsar.broker.service;
 
 import java.util.List;
+import java.util.concurrent.locks.StampedLock;
+
+import it.unimi.dsi.fastutil.longs.Long2IntMap;
+import it.unimi.dsi.fastutil.longs.Long2IntOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2LongMap;
+import it.unimi.dsi.fastutil.longs.Long2LongOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import org.apache.bookkeeper.mledger.Position;
 import org.apache.pulsar.common.util.collections.ConcurrentLongLongPairHashMap;
 import org.apache.pulsar.common.util.collections.ConcurrentLongLongPairHashMap.LongPair;
 
-public class InMemoryRedeliveryTracker implements RedeliveryTracker {
-
-    private ConcurrentLongLongPairHashMap trackerCache = ConcurrentLongLongPairHashMap.newBuilder()
-            .concurrencyLevel(1)
-            .expectedItems(256)
-            .autoShrink(true)
-            .build();
+public class InMemoryRedeliveryTracker extends StampedLock implements RedeliveryTracker {
+    // ledgerId -> entryId -> count
+    private Long2ObjectMap<Long2IntMap> trackerCache = new Long2ObjectOpenHashMap<>();
 
     @Override
     public int incrementAndGetRedeliveryCount(Position position) {
-        Position positionImpl = position;
-        LongPair count = trackerCache.get(positionImpl.getLedgerId(), positionImpl.getEntryId());
-        int newCount = (int) (count != null ? count.first + 1 : 1);
-        trackerCache.put(positionImpl.getLedgerId(), positionImpl.getEntryId(), newCount, 0L);
+        long stamp = writeLock();
+        int newCount;
+        try {
+            Long2IntMap entryMap = trackerCache.computeIfAbsent(position.getLedgerId(),
+                    k -> new Long2IntOpenHashMap());
+            newCount = entryMap.getOrDefault(position.getEntryId(), 0) + 1;
+            entryMap.put(position.getEntryId(), newCount);
+        } finally {
+            unlockWrite(stamp);
+        }
         return newCount;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
```diff
@@ -18,49 +18,94 @@
...
-import org.apache.pulsar.common.util.collections.ConcurrentLongLongPairHashMap.LongPair;
 
-public class InMemoryRedeliveryTracker implements RedeliveryTracker {
-
-    private ConcurrentLongLongPairHashMap trackerCache = ConcurrentLongLongPairHashMap.newBuilder()
-            .concurrencyLevel(1)
-            .expectedItems(256)
-            .autoShrink(true)
-            .build();
+public class InMemoryRedeliveryTracker extends StampedLock implements RedeliveryTracker {
```
lhotari, line 29: It's always better to use composition instead of inheritance, especially in this case. I don't think it's a good idea to extend `StampedLock`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
```diff
@@ -18,49 +18,94 @@
...
         return newCount;
     }
 
     @Override
     public int getRedeliveryCount(long ledgerId, long entryId) {
-        LongPair count = trackerCache.get(ledgerId, entryId);
-        return (int) (count != null ? count.first : 0);
+        long stamp = tryOptimisticRead();
+        Long2IntMap entryMap = trackerCache.get(ledgerId);
+        int count = entryMap != null ? entryMap.get(entryId) : 0;
```
lhotari, line 52: Some data structures might not support an optimistic read. The concern is that a data race could lead to a infinite loop. For example, this used to be the case with JDK's HashMap very long time ago. It might be worth validating this with a concurrency test or by checking the documentation.

lhotari (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
thetumbled: good idea, updated.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
thetumbled: All classes from fastutil is not thread safe, not supporting concurrent modification, but support concurrent read. so we can use optimistic read on it.
https://fastutil.di.unimi.it/docs/#warnings-heading
```
All classes are not synchronized. If multiple threads access one of these classes concurrently, and at least one of the threads modifies it, it must be synchronized externally. Iterators will behave unpredictably in the presence of concurrent modifications. Reads, however, can be carried out concurrently.
```

lhotari: Thanks for checking, everything fine here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/InMemoryRedeliveryTracker.java
```diff
@@ -18,49 +18,95 @@
...
 import org.apache.bookkeeper.mledger.Position;
-import org.apache.pulsar.common.util.collections.ConcurrentLongLongPairHashMap;
-import org.apache.pulsar.common.util.collections.ConcurrentLongLongPairHashMap.LongPair;
 
 public class InMemoryRedeliveryTracker implements RedeliveryTracker {
-
-    private ConcurrentLongLongPairHashMap trackerCache = ConcurrentLongLongPairHashMap.newBuilder()
-            .concurrencyLevel(1)
-            .expectedItems(256)
-            .autoShrink(true)
```
lhotari, line 31: `Long2ObjectOpenHashMap` contains `trim()` and `trim(int)` methods to rehash the table if it's too large. I wonder if it would be useful to implement a feature similar to the `ConcurrentLongLongPairHashMap`'s `autoShrink` feature using the `trim(int)` method?

thetumbled: It seems that `Long2ObjectOpenHashMap` support auto shrink.
`If the table is emptied below one fourth of the load factor, it is halved in size` from the class document: 
`
The table is filled up to a specified load factor, and then doubled in size to accommodate new entries. If the table is emptied below one fourth of the load factor, it is halved in size; however, the table is never reduced to a size smaller than that at creation time: this approach makes it possible to create maps with a large capacity in which insertions and deletions do not cause immediately rehashing. Moreover, halving is not performed when deleting entries from an iterator, as it would interfere with the iteration process.
`

lhotari: Thanks for checking!

lhotari (approved):
LGTM, good work @thetumbled

Issue #21297, opened 2023-10-03
Upgrade to jakarta-annotation-api, javax.annotation-api has been deprecated


michalcukierman:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

It seems like Pulsar, (especially Pulsar Admin Client) still depends on javax.annotation-api 1.3.2, which has been archived, deprecated and replaced by jakarta-annotation-api. This can block the future libraries upgrade and make it complicated to integrate with new framworks/libraries, due to classloading issues.

Observed on using Pulsar Admin Client with Quarkus.

### Solution

Upgrade to jakarta-annotation-api, which is also used by transitive dependencies.
The source code can be found here: https://github.com/eclipse-ee4j/common-annotations-api

### Alternatives

Not a long-term solution, but we could stick with javax.annotation-api for a while. Sooner or later we need to upgrade.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
What is the impact of the deprecated annotations library?

In the Pulsar code base, there are some dependencies to javax libraries due to Jetty 9.4.x dependencies. It could be useful to first upgrade to Jetty 12 (#22939)  and then migrating to use Jakarta annotations and libraries (Servlet) during that.
The reason for this is that in Pulsar, all dependencies are in a flat classpath structure.

michalcukierman:
To use this with Quarkus, you’ll need to add the following dependency:

```xml
    <!-- Use an alternative package because Quarkus excludes javax.annotation-api after build -->
    <dependency>
      <groupId>org.apache.tomcat</groupId>
      <artifactId>tomcat-annotations-api</artifactId>
    </dependency>
```

This is necessary because `javax.annotation-api` is removed from the build artifact by Quarkus.
What's more - the annotation API is often available in recent versions (using Jakarta) in the project classpaths, so relying on the deprecated `javax.annotation` can be inconvenient. 

Although this isn’t a critical issue, I believe it may make it progressively more challenging to integrate admin-client with codebases over time.

lhotari:
> To use this with Quarkus, you’ll need to add the following dependency:
> 
> ```
>     
>     <dependency>
>       <groupId>org.apache.tomcat</groupId>
>       <artifactId>tomcat-annotations-api</artifactId>
>     </dependency>
> ```
> 
[...]

PR #23646, opened 2024-11-27
[fix][client] Remove javax.ws.rs-api dependency from client side

lhotari:
### Motivation

Current [pulsar-client 4.0.0 pom](https://repo1.maven.org/maven2/org/apache/pulsar/pulsar-client/4.0.0/pulsar-client-4.0.0.pom) includes multiple unwanted dependencies.
It contains the `javax.ws.rs-api` dependency which is a server-side dependency. It should be removed.
This is one of changes required to remove `javax` dependencies from the shaded Pulsar client, related to #21297.

### Modifications

- move RestException and PoliciesUtil to pulsar-broker-common since they are the classes that depend on javax.ws.rs-api and they aren't needed in client side code.
- move defaultBundle method to BundlesData
- remove unnecessary `javax.ws.rs-api` dependency from `pulsar-common` module
- remove unnecessary `javax.ws.rs-api` dependency from `pulsar-client-auth-sasl` module

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
Closing in favor of #23647 since integration tests fail when the dependency is removed from client side.

PR #23652, opened 2024-11-28
[improve][log] Print ZK path if write to ZK fails due to data being too large to persist

poorbarcode:

**Improvement 1**
```
java.io.IOException: Len error. A message from /127.0.0.6:43889 with advertised length of 17890775 is either a malformed message or too large to process (length is greater than jute.maxbuffer=10485760)
     at org.apache.zookeeper.server.NIOServerCnxn.readLength(NIOServerCnxn.java:567) ~[org.apache.zookeeper-zookeeper-3.9.2.jar:3.9.2]
     at org.apache.zookeeper.server.NIOServerCnxn.doIO(NIOServerCnxn.java:352) ~[org.apache.zookeeper-zookeeper-3.9.2.jar:3.9.2]
     at org.apache.zookeeper.server.NIOServerCnxnFactory$IOWorkRequest.doWork(NIOServerCnxnFactory.java:508) ~[org.apache.zookeeper-zookeeper-3.9.2.jar:3.9.2]
     at org.apache.zookeeper.server.WorkerService$ScheduledWorkRequest.run(WorkerService.java:153) ~[org.apache.zookeeper-zookeeper-3.9.2.jar:3.9.2]
     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
     at java.lang.Thread.run(Thread.java:840) ~[?:?]
```

We can hardly find the ZK node when encountering the above error.

**Improvement 2**

There is a mechanism that persists cursor ack info into ZK if it matches the two conditions:
- can not create a cursor ledger successfully.
- all messages were acknowledged.

But it will print `Error while using MetaStore, try to persist the position...` even if it will not write ZK because there are still messages to consume.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
@@ -202,12 +202,14 @@ public class ZKMetadataStore extends AbstractBatchedMetadataStore
                                         Collectors.groupingBy(MetadataOp::getType, Collectors.summingInt(op -> 1)))
                                 .entrySet().stream().map(e -> e.getValue() + " " + e.getKey().name() + " entries")
                                 .collect(Collectors.joining(", "));
-                        List<Pair> opsForLog = ops.stream().map(op -> Pair.of(op.getPath(), op.size()))
+                        List<Pair> opsForLog = ops.stream()
+                                .filter(item -> item.size() > 256 * 1024)
+                                .map(op -> Pair.of(op.getPath(), op.size()))
                                 .collect(Collectors.toList());
                         Long totalSize = ops.stream().collect(Collectors.summingLong(MetadataOp::size));
                         log.warn("Connection loss while executing batch operation of {} "
                                 + "of total data size of {}. "
-                                + "Retrying individual operations one-by-one. ops: {}",
+                                + "Retrying individual operations one-by-one. ops whose size > 256KB: {}",
                                 countsByType, totalSize, opsForLog);
 
                         // Retry with the individual operations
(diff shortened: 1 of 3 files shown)


lhotari (dismissed):

lhotari:
@poorbarcode One small detail about the PR titles. Please don't add a space between the 2 prefixes. Instead of `[improve] [log]`, it should be `[improve][log]`.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
```diff
@@ -201,10 +202,13 @@ protected void batchOperation(List<MetadataOp> ops) {
                                         Collectors.groupingBy(MetadataOp::getType, Collectors.summingInt(op -> 1)))
                                 .entrySet().stream().map(e -> e.getValue() + " " + e.getKey().name() + " entries")
                                 .collect(Collectors.joining(", "));
+                        List<Pair> opsForLog = ops.stream().map(op -> Pair.of(op.getPath(), op.size()))
+                                .collect(Collectors.toList());
```
lhotari, line 206: Since this list is only for logging, I think it would be more useful to build a String directly in the desired format. It might be good to limit how many entries are logged and split to multiple log messages in case there could be a large amount of ops. Is the size of the String going to be a problem?

congbobo184 (approved):

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
lhotari: `java.util.StringJoiner` (or `Collectors.joining`) could be used instead.

zjxxzjwang (approved):

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
poorbarcode: Added a filter `op -> op.size > 256K`

dao-jun (approved):

poorbarcode:
@lhotari Do you have other suggestions?

poorbarcode:
> @lhotari Do you have other suggestions?

Since no answer for 5 days, I dismiss the request change

PR #23637, opened 2024-11-25
[fix][broker]: support missing broker level fine-granted permissions

mattisonchao:
### Motivation

Support missing broker-level fine-granted permissions.

### Modifications

- Introduce Broker level operations
- Add `allowBrokerOperationAsync` method for `AuthorizationProvider`.
- Keep compatibility for existing endpoints. 
- Fix admin `backlogQuotaCheck` HTTP request path.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Documentation

- [x] `doc-not-needed`

PR #23663, opened 2024-11-29
[feat][broker] Implement allowBrokerOperationAsync in PulsarAuthorizationProvider to avoid exception thrown

mattisonchao:
### Motivation

#23637 introduces a method called `allowTopicPolicyOperationAsync`. Although we kept the compatibility, the exception thrown will cost more CPU and give GC pressure.

### Modifications

- implemented superuser validation for `PulsarAuthorizationProvider`

### Verifying this change

- The existing test `BrokerEndpointsAuthorizationTest` is already covered. 

### Documentation

- [x] `doc-not-needed`


lhotari (approved):
LGTM

PR #23672, opened 2024-12-03
[fix][ml] The readPosition is greater than the lastPosition

hanmz:
### Motivation
Fix issue: https://github.com/apache/pulsar/issues/23673
We found that sometimes `readPosition` is larger than `lastPosition`. If this happens, messages will not be consumed, messages will be backlogged, and recovery will not be possible.
The reason for this problem is probably when executing `updateReadPosition` method.

### Modifications
When executing the `updateReadPosition` method, the updated position is not allowed to exceed lastPosition+1

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


Coselding (approved):

thetumbled:
> We found that sometimes readPosition is larger than lastPosition. If this happens, messages will not be consumed, messages will be backlogged, and recovery will not be possible.
 
It is possible that the messages has been dispatched to consumer, but not acked. Have you verified this?

PR #23683, opened 2024-12-05
[fix][broker] Catch exception for entry payload interceptor processor

gaoran10:
### Motivation

Currently, the entry payload interceptor processor doesn't catch exceptions, the exception will be swallowed by the executor task.

### Modifications

Catch exception for entry payload interceptor processor.

### Verifying this change

Add a test for catching exceptions from the payload input interceptor processor.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
@@ -139,8 +139,14 @@ public class OpAddEntry implements AddCallback, CloseCallback, Runnable, Managed
             lastInitTime = System.nanoTime();
             if (ml.getManagedLedgerInterceptor() != null) {
                 long originalDataLen = data.readableBytes();
-                payloadProcessorHandle = ml.getManagedLedgerInterceptor()
-                        .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
+                try {
+                    payloadProcessorHandle = ml.getManagedLedgerInterceptor()
+                            .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
+                } catch (Exception e) {
+                    log.error("[{}] Error processing payload before ledger write", ml.getName(), e);
+                    this.failed(new ManagedLedgerException.ManagedLedgerInterceptException(e));
+                    return;
+                }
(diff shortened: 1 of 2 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -139,8 +139,14 @@ public void initiate() {
             lastInitTime = System.nanoTime();
             if (ml.getManagedLedgerInterceptor() != null) {
                 long originalDataLen = data.readableBytes();
-                payloadProcessorHandle = ml.getManagedLedgerInterceptor()
-                        .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
+                try {
+                    payloadProcessorHandle = ml.getManagedLedgerInterceptor()
+                            .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
+                } catch (Exception e) {
+                    log.error("[{}] Error processing payload before ledger write", ml.getName(), e);
```
lhotari, line 146: would it make sense to call `duplicateBuffer.release()` here?

dao-jun: +1.
Referring to: https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L811-L879

it will call OpAddEntry.failed(..) before OpAddEntry.initiate(). Since OpAddEntry.initiate() retained the buffer, so call `duplicateBuffer.release()` here is reasonable.

gaoran10: Good idea.

It makes sense. Normally, the release operation will execute in the Processor's release method, but if exceptions are thrown, the release operation can't be executed, if we don't release data here, the processor needs to release data when throwing exceptions.

dao-jun (approved):

lhotari (approved):

PR #23686, opened 2024-12-06
[fix][broker] Fix the retry mechanism in `MetadataCache#readModifyUpdateOrCreate`

RobertIndie:
## Motivation

The method `MetadataCache#readModifyUpdateOrCreate` should handle the BadVersionException by retrying the modification process, as already noted in the Java documentation: "The modify function can potentially be called multiple times due to concurrent updates."

Currently, `MetadataCache#readModifyUpdateOrCreate` does not catch the BadVersionException on the second attempt, allowing the exception to be passed to the caller. This issue can be easily reproduced by increasing concurrent futures in the test `MetadataCacheTest#readModifyUpdateBadVersionRetry`.

The current retry implementation is incorrect and lacks a backoff mechanism, which could lead to too many requests to the metadata store.

## Modification

- Correct the retry process in `MetadataCache#readModifyUpdateOrCreate` to ensure BadVersionException is caught during each retry.
- Implement a retry backoff mechanism in `MetadataCache#readModifyUpdateOrCreate` to manage the frequency of retries effectively.
- Add new config `retryBackoff` to the MetadataCacheConfig to control the MetadataCache retry backoff.
- Respective the `metadataStoreOperationTimeoutSeconds` for the MetadataCache retry

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
@@ -24,14 +24,12 @@ import com.github.benmanes.caffeine.cache.AsyncCacheLoader;
 import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.annotations.VisibleForTesting;
-import io.netty.util.concurrent.DefaultThreadFactory;
 import java.io.IOException;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -63,21 +61,23 @@ public class MetadataCacheImpl<T> implements MetadataCache<T>, Consumer<Notifica
     private final MetadataStore store;
     private final MetadataStoreExtended storeExtended;
     private final MetadataSerde<T> serde;
-    private final ScheduledExecutorService backoffExecutor =
-            Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("metadata-cache-backoff"));
+    private final ScheduledExecutorService executor;
     private final MetadataCacheConfig<T> cacheConfig;
 
     private final AsyncLoadingCache<String, Optional<CacheGetResult<T>>> objCache;
 
-    public MetadataCacheImpl(MetadataStore store, TypeReference<T> typeRef, MetadataCacheConfig<T> cacheConfig) {
-        this(store, new JSONMetadataSerdeTypeRef<>(typeRef), cacheConfig);
+    public MetadataCacheImpl(MetadataStore store, TypeReference<T> typeRef, MetadataCacheConfig<T> cacheConfig,
+                             ScheduledExecutorService executor) {
+        this(store, new JSONMetadataSerdeTypeRef<>(typeRef), cacheConfig, executor);
     }
 
-    public MetadataCacheImpl(MetadataStore store, JavaType type, MetadataCacheConfig<T> cacheConfig) {
-        this(store, new JSONMetadataSerdeSimpleType<>(type), cacheConfig);
+    public MetadataCacheImpl(MetadataStore store, JavaType type, MetadataCacheConfig<T> cacheConfig,
+                             ScheduledExecutorService executor) {
+        this(store, new JSONMetadataSerdeSimpleType<>(type), cacheConfig, executor);
     }
 
-    public MetadataCacheImpl(MetadataStore store, MetadataSerde<T> serde, MetadataCacheConfig<T> cacheConfig) {
+    public MetadataCacheImpl(MetadataStore store, MetadataSerde<T> serde, MetadataCacheConfig<T> cacheConfig,
+                             ScheduledExecutorService executor) {
         this.store = store;
         if (store instanceof MetadataStoreExtended) {
             this.storeExtended = (MetadataStoreExtended) store;
@@ -86,6 +86,7 @@ public class MetadataCacheImpl<T> implements MetadataCache<T>, Consumer<Notifica
         }
         this.serde = serde;
         this.cacheConfig = cacheConfig;
+        this.executor = executor;
 
         Caffeine<Object, Object> cacheBuilder = Caffeine.newBuilder();
         if (cacheConfig.getRefreshAfterWriteMillis() > 0) {
@@ -336,15 +337,16 @@ public class MetadataCacheImpl<T> implements MetadataCache<T>, Consumer<Notifica
                 // if resource is updated by other than metadata-cache then metadata-cache will get bad-version
                 // exception. so, try to invalidate the cache and try one more time.
                 objCache.synchronous().invalidate(key);
+                long elapsed = System.currentTimeMillis() - backoff.getFirstBackoffTimeInMillis();
                 if (backoff.isMandatoryStopMade()) {
-                    result.completeExceptionally(new TimeoutException(String.format("Timeout to update key %s", key)));
+                    result.completeExceptionally(new TimeoutException(
+                            String.format("Timeout to update key %s. Elapsed time: %d ms", key, elapsed)));
                     return null;
                 }
                 final var next = backoff.next();
-                log.info("Update key {} conflicts. Retrying in {} ms. Mandatory stop: {} ms. Elapsed time: {} ms", key,
-                        next, backoff.isMandatoryStopMade(),
-                        System.currentTimeMillis() - backoff.getFirstBackoffTimeInMillis());
-                backoffExecutor.schedule(() -> execute(op, key, result, backoff), next,
+                log.info("Update key {} conflicts. Retrying in {} ms. Mandatory stop: {}. Elapsed time: {} ms", key,
+                        next, backoff.isMandatoryStopMade(), elapsed);
+                executor.schedule(() -> execute(op, key, result, backoff), next,
                         TimeUnit.MILLISECONDS);
                 return null;
             }
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
@@ -30,6 +30,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -37,6 +39,7 @@ import java.util.function.Supplier;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.common.concurrent.FutureUtils;
+import org.apache.pulsar.common.util.Backoff;
 import org.apache.pulsar.metadata.api.CacheGetResult;
 import org.apache.pulsar.metadata.api.GetResult;
 import org.apache.pulsar.metadata.api.MetadataCache;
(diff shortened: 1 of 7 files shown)


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
```diff
@@ -58,6 +61,7 @@ public class MetadataCacheImpl<T> implements MetadataCache<T>, Consumer<Notifica
     private final MetadataStore store;
     private final MetadataStoreExtended storeExtended;
     private final MetadataSerde<T> serde;
+    private final ScheduledExecutorService backoffExecutor = Executors.newSingleThreadScheduledExecutor();
```
dao-jun, line 64: It' better not use `Executors....` to new an Executor.
BTW, can we reuse the existing scheduler?  Such as `AbstractMetadataStore.executor`, maybe we can increase it's thread num to more.

RobertIndie: But how can we access the `AbstractMetadataStore.executor` here?

dao-jun: can we just pass `AbstractMetadataStore.executor` to  MetaCacheImp via the constructors?

BewareMyPower: > It' better not use Executors.... to new an Executor.

Why?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
```diff
@@ -321,22 +325,25 @@ public void accept(Notification t) {
...
+                        TimeUnit.MILLISECONDS);
                 return null;
             }
             result.completeExceptionally(ex.getCause());
             return null;
         });
+    }
+
+    private CompletableFuture<T> executeWithRetry(Supplier<CompletableFuture<T>> op, String key) {
+        final var backoff = new Backoff(100, TimeUnit.MILLISECONDS, 1, TimeUnit.MINUTES, 0, TimeUnit.MILLISECONDS);
```
BewareMyPower, line 344: 1 minute might be too long. Generally retrying after a few seconds could never fail, but if there is a bug that makes it retry forever, it could reach the maximum backoff value.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
dao-jun: Because it's thread name is meaningless, it maybe could lead to some difficulties when we analysis the thread dump.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
RobertIndie: In most cases, it should work after two or three tries. If there is an unrecoverable bug, it will cause many conccurent requests in a short time. In this situation, retrying many times quickly is meaningless.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
RobertIndie: I get your point. Then we can use `Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("metadata-cache-backoff"))` to give it a meaningful name. https://github.com/apache/pulsar/pull/23686/commits/dbca96ebb98fc682131c779077e5d392e94a0352

WDYT?

dao-jun (approved):
LGTM

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
codelipenghui: After reviewing the details, it seems the retry mechanism is primarily designed to avoid invalidating the cache on the broker. However, I believe this is also incorrectly implemented.

Retries should not occur when another user has updated the data to a new version. Instead, the system should throw a ConcurrentModificationException. This allows users to review the updated data before deciding whether to proceed with their changes. For example, if two users attempt to set the data retention policy for a namespace, one user should receive a ConcurrentModificationException if there is a conflict. Typically, this requires the user to check the latest policy before retrying.

Regarding the incorrect implementation of the retry mechanism, it should verify the data version between the cache and the metadata store. If the versions match, retries should stop, as this indicates the data has been updated elsewhere.

BewareMyPower: In the metadata store API, it seems that we use `BadVersionException` rather than `ConcurrentModificationException` in this case. It means when it tries to update a node in the metadata store, the version is outdated (due to the existing version has already been bumped by another update operation)

RobertIndie: Now the problem becomes whether we should throw the ConcurrentModificationException/BadVersionException. From a design perspective and according to [the readModifyUpdate documentation](https://github.com/apache/pulsar/blob/216b83008deb469e0fc55ed8117f0c393ebcb0ac/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataCache.java#L108), it seems that MetadataCacheImpl should be responsible for retrying and should not throw this exception.

Since readModifyUpdate does not require users to provide a version, it does not seem reasonable to throw a BadVersionException/ConcurrentModificationException to the caller. If the caller needs to ensure that data is not accidentally overwritten, throwing a BadVersionException/ConcurrentModificationException will not solve this problem. Therefore, unless we can explicitly pass in a version, we cannot achieve this semantic.

Besides, I found that introducing unlimited retries in MetadataCache is also not reasonable. We can use `metadataStoreOperationTimeoutSeconds` as the mandatory stop time for retries and throw the TimeoutException if all retires are failed. Also we can reduce the maximum timeout time following @BewareMyPower 's [suggestion](https://github.com/streamnative/ursa-storage/pull/548#discussion_r1873286488). WDYT? @codelipenghui @BewareMyPower

BewareMyPower: Oh it's right. If we follow the API description, we should add the retry mechanism inside the implementation rather than let users retry manually.

BewareMyPower: > , I found that introducing unlimited retries in MetadataCache is also not reasonable

I think we can provide two options to `MetadataCacheConfig` to specify the backoff and the timeout.

RobertIndie: > I think we can provide two options to MetadataCacheConfig to specify the backoff and the timeout.

I think we need 3 options: backoffInitialTimeInMs, backoffMaxTimeInMs and backoffMandatoryStopTimeInMs.
And we should use `metadataStoreOperationTimeoutSeconds * 1000` as the default value of `backoffMandatoryStopTimeInMs`.
The default value of `backoffInitialTimeInMs` could be 5ms and the default `backoffMaxTimeInMs` could be 3000ms.

BewareMyPower: Why not just add `Backoff` as its field?

```java
public class MetadataCacheConfig<T> {
    // ...
    private final Backoff backoff = new Backoff(5, TimeUnit.MILLISECONDS, 3, TimeUnit.SECONDS, 0, TimeUnit.MILLISECONDS);
```

Regarding the timeout, I mean if a modification operation is not done after the timeout, just fail the future with `TimeoutException`. It's independent from the `Backoff`

RobertIndie: > Why not just add Backoff as its field?

OK. That's also fine.

> Regarding the timeout, I mean if a modification operation is not done after the timeout, just fail the future with TimeoutException. It's independent from the Backoff

Should the caller or the metadata cache fail the future? My idea is to let the MetadataCache fail the future with the TimeoutException, so we could use the backoff mandatory stop time as the timeout. But if it's the caller, then it doesn't make sense.  Failing the future this way won't stop the retry task. We could check the future before each retry, but the bigger issue is that the current caller won't fail the future. Without introducing a mandatory stop time, the retry task could run indefinitely.

BewareMyPower: Ah yes, we don't need the timeout. We have `CompletableFuture#orTimeout` to support this purpose. So adding a `Backoff` field is enough

RobertIndie: I added a new config `retryBackoff` to the MetadataCacheConfig. I also used `isMandatoryStopMade` to check for timeouts. If it times out, the future will be failed.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
```diff
@@ -321,22 +330,33 @@ public void accept(Notification t) {
...
-                op.get().thenAccept(result::complete).exceptionally((ex1) -> {
-                    result.completeExceptionally(ex1.getCause());
+                if (backoff.isMandatoryStopMade()) {
+                    result.completeExceptionally(new TimeoutException(String.format("Timeout to update key %s", key)));
                     return null;
-                });
+                }
+                final var next = backoff.next();
+                log.info("Update key {} conflicts. Retrying in {} ms. Mandatory stop: {} ms. Elapsed time: {} ms", key,
+                        next, backoff.isMandatoryStopMade(),
```
BewareMyPower, line 345: `isMandatoryStopMade()` is a boolean but you logs `Mandatory stop: {} ms`.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
```diff
@@ -58,6 +63,9 @@ public class MetadataCacheImpl<T> implements MetadataCache<T>, Consumer<Notifica
     private final MetadataStore store;
     private final MetadataStoreExtended storeExtended;
     private final MetadataSerde<T> serde;
+    private final ScheduledExecutorService backoffExecutor =
+            Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("metadata-cache-backoff"));
```
BewareMyPower, line 67: Actually I'm wondering if we need one thread per `MetadataCache`. It means each time a `MetadataCache` instance is created, a new thread will be created, which might be unnecessary.

I think a better design is to add a common thread pool like `ForkJoinPool#commonPool` and a config in `MetadataCacheConfig` to specify a customized executor if users want.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
```diff
@@ -321,22 +330,33 @@ public void accept(Notification t) {
...
+    private void execute(Supplier<CompletableFuture<T>> op, String key, CompletableFuture<T> result, Backoff backoff) {
         op.get().thenAccept(result::complete).exceptionally((ex) -> {
             if (ex.getCause() instanceof BadVersionException) {
                 // if resource is updated by other than metadata-cache then metadata-cache will get bad-version
                 // exception. so, try to invalidate the cache and try one more time.
                 objCache.synchronous().invalidate(key);
-                op.get().thenAccept(result::complete).exceptionally((ex1) -> {
-                    result.completeExceptionally(ex1.getCause());
+                if (backoff.isMandatoryStopMade()) {
+                    result.completeExceptionally(new TimeoutException(String.format("Timeout to update key %s", key)));
```
BewareMyPower, line 340: It's better to print the elapsed time.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
RobertIndie: We can use the `AbstractMetadataStore.executor` as suggessted here: https://github.com/apache/pulsar/pull/23686#discussion_r1872973847

BewareMyPower (approved):

PR #23688, opened 2024-12-06
[fix][cli] Fix set topic retention policy failed

hangc0276:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Motivation
When using this command `bin/pulsar-admin topics set-retention -s 20g -t 200d persistent://public/default/test_v1` to set a retention policy for a topic, it failed with following exception
```
PicocliException: Could not set value for field private java.lang.Integer org.apache.pulsar.admin.cli.CmdTopics$SetRetention.retentionTimeInSec to 17280000 while processing argument at or before arg[5] '200d' in [topics, set-retention, -s, 20g, -t, 200d, persistent://public/default/test_v1]: picocli.CommandLine$PicocliException: Could not set value for field private java.lang.Integer org.apache.pulsar.admin.cli.CmdTopics$SetRetention.retentionTimeInSec to 17280000
Usage: pulsar-admin topics set-retention [-hv] -s=<sizeLimit>
       -t=<retentionTimeInSec> <topicName>
Set the retention policy for a topic
      <topicName>          persistent://tenant/namespace/topic
  -h, --help               Show this help message and exit.
  -s, --size=<sizeLimit>   Retention size limit with optional size unit suffix.
                             For example, 4096, 10M, 16G, 3T.  The size unit
                             suffix character can be k/K, m/M, g/G, or t/T.  If
                             the size unit suffix is not specified, the default
                             unit is bytes. 0 or less than 1MB means no
                             retention and -1 means infinite size retention
  -t, --time=<retentionTimeInSec>
                           Retention time with optional time unit suffix. For
                             example, 100m, 3h, 2d, 5w. If the time unit is not
                             specified, the default unit is seconds. For
                             example, -t 120 will set retention to 2 minutes. 0
                             means no retention and -1 means infinite time
                             retention.
  -v, --version            Print version information and exit.
```

It is caused by cast Long to Integer failed
https://github.com/apache/pulsar/blob/28e47fa99dcb080385cfa567b844e74ab0cd85b1/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java#L1854-L1855

https://github.com/apache/pulsar/blob/28e47fa99dcb080385cfa567b844e74ab0cd85b1/pulsar-cli-utils/src/main/java/org/apache/pulsar/cli/converters/picocli/TimeUnitToSecondsConverter.java#L28-L35

### Modifications
Change the type of `retentionTimeInSec` from `Integer` to `Long`

### Verifying this change

[...]

PR #22150, opened 2024-02-28
[fix][cli] Fix the bug when set-retention specified size with -T

Technoboy-:
Fixes #22138

### Motivation

When specified with -s with `-xT`, it will overflow when convert long to int.
image
image

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23689, opened 2024-12-06
[fix][cli] Fix set-retention with >2GB size value for topic policy

lhotari:
### Motivation

Fixes a similar issue as #22150, for `pulsar-admin topics set-retention` and `pulsar-admin topicPolicies set-retention` commands. Related to #23688. This adds a test to #23688 changes.

The impact of this bug is that when the value overflows when it's over 2GB. The effective value becomes negative or 0, which disables retention. There won't be a failure. For example, with the parameter `-s 3G`, the value becomes `-1024`. With any value more than 4GB, the value becomes `0`.

### Modifications

- use long value instead of integer
- add tests which also cover #23688 change

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopicPolicies.java
+++ b/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopicPolicies.java
@@ -29,7 +29,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


liangyepianzhou (approved):
LGTM

dao-jun (approved):
lgtm

pdolif (approved):

poorbarcode:
@lhotari 

If do not cherry-pick into `branch-3.0`, there is a downgrade compatibility issue:
- upgrade to `4.0.x`
- Set a larger value for the policy.
- downgrade to `3.0.x`, the value will be transferred to a negative value,

PR #23693, opened 2024-12-09
[fix][client] Fix reader message filtering issue during blue-green cluster switch

hrzzzz:
### Motivation
When the delayed acknowledgment feature is enabled, the method `org.apache.pulsar.client.impl.PersistentAcknowledgmentsGroupingTracker#isDuplicate` filters out messages that have already been acknowledged on the client side. 

When using the blue-green cluster feature, after switching from the blue cluster to the green cluster, the message IDs generated by the green cluster may be smaller than those from the blue cluster. In this case, if a reader is used for consumption, the method `org.apache.pulsar.client.impl.PersistentAcknowledgmentsGroupingTracker#isDuplicate` will filter out all messages with IDs less than or equal to `lastCumulativeAck`, which can result in the loss of messages with IDs less than or equal to `lastCumulativeAck` after switching to the green cluster.

### Modifications
Before switching from the blue cluster to the green cluster, first clear the contents of `org.apache.pulsar.client.impl.AcknowledgmentsGroupingTracker`

### Verifying this change

Added test in: `org.apache.pulsar.broker.service.ClusterMigrationTest#testMigrationWithReader`

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/hrzzzz/pulsar/pull/11

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
@@ -518,6 +518,13 @@ public class PersistentAcknowledgmentsGroupingTracker implements Acknowledgments
     }
 
     @Override
+    public void clean() {
+        lastCumulativeAck.reset();
+        pendingIndividualAcks.clear();
+        pendingIndividualBatchIndexAcks.clear();
+    }
+
+    @Override
     public void close() {
         flush();
         if (scheduledTask != null && !scheduledTask.isCancelled()) {
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ClusterMigrationTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ClusterMigrationTest.java
@@ -45,6 +45,7 @@ import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageRoutingMode;
 import org.apache.pulsar.client.api.Producer;
 import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.Reader;
 import org.apache.pulsar.client.api.SubscriptionType;
 import org.apache.pulsar.common.policies.data.ClusterData;
 import org.apache.pulsar.common.policies.data.ClusterPolicies.ClusterUrl;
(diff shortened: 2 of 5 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
```diff
@@ -517,6 +517,13 @@ public void flushAndClean() {
         pendingIndividualAcks.clear();
```
lhotari, line 517: Just wondering why the existing `flushAndClean()` method couldn't be used? It seems that `pendingIndividualBatchIndexAcks.clear()` is missing from this method. I wonder if that's an issue in this method?

thetumbled:
I am curious about how pulsar ensure the two message ids from two clusters corresponding to the same message are equal, or they are just not equal?  Not equal may introduce lots of problem.

lhotari:
> I am curious about how pulsar ensure the two message ids from two clusters corresponding to the same message are equal, or they are just not equal? Not equal may introduce lots of problem.

Good question. In the blue-green case the message id comparison most likely doesn't apply.
In general there isn't a way to compare message ids between clusters.

The Shadow Topic feature added a field to CommandSend to send the source cluster message id. It would be useful for translations, https://github.com/apache/pulsar/blob/9a97c843a46e23a0811e2172991cd00a3af642c0/pulsar-common/src/main/proto/PulsarApi.proto#L532-L533 . Taking this further and changing geo-replication to include the original message id and original source cluster would be useful for adding individual ack replication support to replicated subscriptions. Replicated subscriptions don't currently support individual acks and being able to compare message ids would be necessary for adding that feature.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
hrzzzz: 1. We can use the `flushAndClean` method, but it performs some additional unnecessary operations. Therefore, I added the `clean` method to simply reset the `acknowledgmentsGroupingTracker`. If you think this is not appropriate, I can reuse the flushAndClean method.
2. The `flushAndClean` method internally removes elements from `pendingIndividualBatchIndexAcks`, so there is no issue with this method.

hrzzzz: @lhotari I switched to using the `flushAndClean` method. Could you take another look, please? Thank you very much!

lhotari (approved):
LGTM

PR #23697, opened 2024-12-09
[fix][broker] Geo Replication lost messages or frequently fails due to Deduplication is not appropriate for Geo-Replication

poorbarcode:
#### Background

**How does deduplication work?**
- Producer sends messages and records sequence-ids that are pending responses in memory, the variable is `{pendingMessages}`
- Brokers will respond to an invalid position `-1:-1` if the sequence ID published is lower than the previous messages.
- Producer checks whether the next sequence id in `{pendingMessages}` is larger than the one that was rejected.
  - `{next} > {rejected}`: ignore the error, and continue work
  - `{next} < {rejected}`: close channels and reconnect.

Conditions that issue happened

- Replicators send all messages on a topic and do not matter what message was sent by which original producer.
  - **(Highlight)** it mixed messages into `{pendingMessages}` with `message.sequenceId` but ignores `message.original-producer-name`, which may cause the sequence-ids in `{pendingMessages}` is not increasing
- The producer checking the sequence ID after receiving a `-1:-1` send response will fail.

---

**Issue-1: loss messages**

- Publishing in the source cluster
  - Producer-1 send 2 messages: M1(`seq: 0`), M2(`seq: 1`)
  - Producer-2 send 2 messages: M3(`seq: 0`), M4(`seq: 1`)
- Replicate messages to the remote cluster
  - Copies M1 and M2
    - `{pendingMessages}`: `[0,1]`
  - Repeatedly copies M1 and M2. and copies M3 and M4.
    - `{pendingMessages}`: `[0,1,0,1]`
    - After repeatedly copying M1 and M2, the network broke.
    - Receive send receipt:
      - `seq 0, position 0:0`
      - `seq 1, position 0:1`
      - `seq 0, position -1: -1`
      - `seq 1, position -1:-1`
    - `{pendingMessages}`: `[empty]`
- After a topic unloading.
  - The replicator will start after the topic is loaded up(backlog is `0` now).
  - The client will create a new connection.
- Result:
  - Disabled deduplication: copied `[M1, M2, M1, M2]`
  - Enabled deduplication: copied `[M1, M2]`

You can reproduce the issue by the test `testDeduplicationNotLostMessage`

---

**Issue-2: frequently fails**
- Original-producer-2: sent msg `3:0` with sequence-id `10`
- Original-producer-1: sent msg `3:1` with sequence-id `1`
- Original-producer-1: sent msg `3:2` with sequence-id `2`
-0 Replicator copies messages 
  - `{pendingMessages}`: `[10,1, 2]`
- Sent `3:0` successfully
  - `{pendingMessages}`: `[1,2]`
- Resend `3:0`(a duplicated publishing)
  - Receive `-1:-1`(new position relates to the latest publishing) for the latest send-response.
  - `failed-sequenced:10 > pendingMessages[0].sequenceId: 1`
  - Replicator-producer reconnects and resets the cursor, which leads to more duplicated publishing.
  - Loop to the step 1.

No test for reproducing this issue.

### Modifications

**Solution: replicators use a specified sequence ID(`ledegrId:entryId` of the original topic) instead of using the original producers’**
- Original-producer-2: sent msg `3:0` with sequence-id `10`
- Original-producer-1: sent msg `3:1` with sequence-id `1`
- Original-producer-1: sent msg `3:2` with sequence-id `2`
- Replicator copies messages with specified sequence ID(`3:0`)
  - `{pendingMessages}`: `3:0, 3:1, 3:2]`
- Sent `3:0` successfully
  - `{pendingMessages}`: `[3:1, 3:2]`
- Resend `3:0`(a duplicated publishing)
- Receive `-1:-1`(new position relates to the latest publishing) for the latest send-response.	
- Ignored the error since `failed-sequenced(3:0) < pendingMessages[0].sequenceId(3:2)`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
@@ -567,9 +567,9 @@ public class Producer {
             String replSequenceLIdStr = String.valueOf(getProperty(MSG_PROP_REPL_SEQUENCE_LID));
             String replSequenceEIdStr = String.valueOf(getProperty(MSG_PROP_REPL_SEQUENCE_EID));
             if (!StringUtils.isNumeric(replSequenceLIdStr) || !StringUtils.isNumeric(replSequenceEIdStr)) {
-                log.error("[{}] Message can not determine whether the message is duplicated due to the acquired messages"
-                                + " props were are invalid. producer={}. supportsDedupReplV2: {}, sequence-id {},"
-                                + " prop-{}: {}, prop-{}: {}",
+                log.error("[{}] Message can not determine whether the message is duplicated due to the acquired"
+                                + " messages props were are invalid. producer={}. supportsDedupReplV2: {},"
+                                + " sequence-id {}, prop-{}: {}, prop-{}: {}",
                         producer.topic.getName(), producer.producerName,
                         supportsDedupReplV2(), getSequenceId(),
                         MSG_PROP_REPL_SEQUENCE_LID, replSequenceLIdStr,
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
@@ -278,16 +278,16 @@ public class Producer {
         return true;
     }
 
-    private boolean isSupportsDedupReplV2() {
+    private boolean isSupportsReplDedupByLidAndEid() {
         // Non-Persistent topic does not have ledger id or entry id, so it does not support.
-        return cnx.isClientSupportsDedupReplV2() && topic.isPersistent();
+        return cnx.isClientSupportsReplDedupByLidAndEid() && topic.isPersistent();
     }
 
     private void publishMessageToTopic(ByteBuf headersAndPayload, long sequenceId, int batchSize, boolean isChunked,
                                        boolean isMarker, Position position) {
         MessagePublishContext messagePublishContext =
                 MessagePublishContext.get(this, sequenceId, headersAndPayload.readableBytes(),
-                        batchSize, isChunked, System.nanoTime(), isMarker, position, isSupportsDedupReplV2());
+                        batchSize, isChunked, System.nanoTime(), isMarker, position, isSupportsReplDedupByLidAndEid());
         if (brokerInterceptor != null) {
             brokerInterceptor
                     .onMessagePublish(this, headersAndPayload, messagePublishContext);
@@ -299,7 +299,7 @@ public class Producer {
                                        int batchSize, boolean isChunked, boolean isMarker, Position position) {
         MessagePublishContext messagePublishContext = MessagePublishContext.get(this, lowestSequenceId,
                 highestSequenceId, headersAndPayload.readableBytes(), batchSize,
-                isChunked, System.nanoTime(), isMarker, position, isSupportsDedupReplV2());
+                isChunked, System.nanoTime(), isMarker, position, isSupportsReplDedupByLidAndEid());
         if (brokerInterceptor != null) {
             brokerInterceptor
                     .onMessagePublish(this, headersAndPayload, messagePublishContext);
@@ -394,7 +394,7 @@ public class Producer {
         private int batchSize;
         private boolean chunked;
         private boolean isMarker;
-        private boolean supportsDedupReplV2;
+        private boolean supportsReplDedupByLidAndEid;
 
         private long startTimeNs;
 
@@ -481,8 +481,8 @@ public class Producer {
         }
 
         @Override
-        public boolean supportsDedupReplV2() {
-            return supportsDedupReplV2;
+        public boolean supportsReplDedupByLidAndEid() {
+            return supportsReplDedupByLidAndEid;
         }
 
         @Override
@@ -559,7 +559,7 @@ public class Producer {
             // stats
             producer.stats.recordMsgIn(batchSize, msgSize);
             producer.topic.recordAddLatency(System.nanoTime() - startTimeNs, TimeUnit.NANOSECONDS);
-            if (producer.isRemoteOrShadow && producer.isSupportsDedupReplV2()) {
+            if (producer.isRemoteOrShadow && producer.isSupportsReplDedupByLidAndEid()) {
                 sendSendReceiptResponseRepl();
             } else {
                 // Repl V1 is the same as normal for this handling.
@@ -590,10 +590,10 @@ public class Producer {
             Object positionPairObj = getProperty(MSG_PROP_REPL_SOURCE_POSITION);
             if (positionPairObj == null || !(positionPairObj instanceof long[])) {
                 log.error("[{}] Message can not determine whether the message is duplicated due to the acquired"
-                                + " messages props were are invalid. producer={}. supportsDedupReplV2: {},"
+                                + " messages props were are invalid. producer={}. supportsReplDedupByLidAndEid: {},"
                                 + " sequence-id {}, prop-{}: not in expected format",
                         producer.topic.getName(), producer.producerName,
-                        supportsDedupReplV2(), getSequenceId(),
+                        supportsReplDedupByLidAndEid(), getSequenceId(),
                         MSG_PROP_REPL_SOURCE_POSITION);
                 producer.cnx.getCommandSender().sendSendError(producer.producerId,
                         Math.max(highestSequenceId, sequenceId),
@@ -615,7 +615,7 @@ public class Producer {
 
         static MessagePublishContext get(Producer producer, long sequenceId, int msgSize, int batchSize,
                                          boolean chunked, long startTimeNs, boolean isMarker, Position position,
-                                         boolean supportsDedupReplV2) {
+                                         boolean supportsReplDedupByLidAndEid) {
             MessagePublishContext callback = RECYCLER.get();
             callback.producer = producer;
             callback.sequenceId = sequenceId;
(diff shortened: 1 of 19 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/GeoPersistentReplicator.java
```diff
@@ -191,9 +193,22 @@ protected boolean replicateEntries(List<Entry> entries) {
                         readMoreEntries();
                     });
                 } else {
+                    // TODO test cases
+                    //   - Reopen Geo-Replication.
+                    //   - Multi version schemas.
+                    //   - Deduplicated replication.
```
poorbarcode, line 199: Just left them here to mention adding test tomorrow

pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
```diff
@@ -0,0 +1,109 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.pulsar.client.api.Schema;
+import org.apache.pulsar.client.impl.conf.ProducerConfigurationData;
+import org.apache.pulsar.common.api.proto.KeyValue;
+import org.apache.pulsar.common.api.proto.MessageMetadata;
+import org.apache.pulsar.common.protocol.Commands;
+
+@Slf4j
+public class GeoReplicationProducerImpl extends ProducerImpl{
```
poorbarcode, line 33: TODO: edit the implementation of `ProducerBuilder`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+                op.setMessageId(ledgerId, entryId, partitionIndex);
+                try {
+                    // Need to protect ourselves from any exception being thrown in the future handler from the
+                    // application
+                    op.sendComplete(null);
+                } catch (Throwable t) {
+                    log.warn("[{}] [{}] Got exception while completing the callback for -- source-message: {}:{} --"
+                                    + " target-msg: {}:{}",
+                            topic, producerName, lIdSent, eIdSent, ledgerId, entryId, t);
+                }
```
poorbarcode, line 119: add a `finally`

codelipenghui:
@poorbarcode It's a good idea to just use the ledger ID and entry ID for the message deduplication. In this case, we can also remove the deduplication state after the ledger get fully replicated. For example:

- Replicating data from ledger 1
- The last entry ID of ledger 1 is 10000
- After the remote cluster get replicated data with ledger ID greater than 1
- Then we can delete the deduplication state for ledger 1

pulsar-common/src/main/proto/PulsarApi.proto
```diff
@@ -302,6 +302,7 @@ message FeatureFlags {
   optional bool supports_partial_producer = 3 [default = false];
   optional bool supports_topic_watchers = 4 [default = false];
   optional bool supports_get_partitioned_metadata_without_auto_creation = 5 [default = false];
+  optional bool supports_dedup_repl_v2 = 6 [default = false];
```
BewareMyPower, line 305: There is no document for how "v2" makes the difference from the existing behavior and why it needs two versions of replicator deduplication. I think it definitely requires a PIP to explain it in details

BewareMyPower (changes requested):

pulsar-common/src/main/proto/PulsarApi.proto
poorbarcode: Renamed it to `supports_repl_dedup_by_lid_and_eid`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java
```diff
@@ -474,18 +480,18 @@ protected void handleSendReceipt(CommandSendReceipt sendReceipt) {
...
-            log.warn("{} Message with sequence-id {} published by producer {} has been dropped", ctx.channel(),
-                    sequenceId, producerId);
-        }
-
-        if (log.isDebugEnabled()) {
-            log.debug("{} Got receipt for producer: {} -- msg: {} -- id: {}:{}", ctx.channel(), producerId, sequenceId,
-                    ledgerId, entryId);
+            log.warn("{} Message with sequence-id {}-{} published by producer [id:{}, name:{}] has been dropped",
+                    ctx.channel(), sequenceId, highestSequenceId, producerId, producer.getProducerName());
+        } else {
```
Technoboy-, line 487: the `producer` maybe null, so NPE here

poorbarcode: FIxed, thanks

pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
```diff
@@ -0,0 +1,248 @@
...
+                && (pendingLId < lastPersistedSourceLedgerId || (pendingLId.longValue() == lastPersistedSourceLedgerId
+                  && pendingEId.longValue() <= lastPersistedSourceEntryId))) {
+            if (log.isDebugEnabled()) {
+                log.debug("[{}] [{}] Received an msg send receipt[pending send is repeated due to repl cursor rewind]:"
+                                + " source entry {}:{}, pending send: {}:{}, latest persisted: {}:{}",
+                        topic, producerName, sourceLId, sourceEId, pendingLId, pendingEId,
+                        lastPersistedSourceLedgerId, lastPersistedSourceEntryId);
+            }
+            removeAndApplyCallback(op, sourceLId, sourceEId, targetLId, targetEid, false);
+            ackReceived(cnx, sourceLId, sourceEId, targetLId, targetEid);
```
gaoran10, line 117: Why does it need to call the method ackReceived?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
```diff
@@ -551,8 +577,45 @@ public void run() {
...
+                        MSG_PROP_REPL_SOURCE_POSITION);
+                producer.cnx.getCommandSender().sendSendError(producer.producerId,
+                        Math.max(highestSequenceId, sequenceId),
+                        ServerError.PersistenceError, "Message can not determine whether the message is"
+                                + " duplicated due to the acquired messages props were are invalid");
+                return;
+            }
+            long[] positionPair = (long[]) positionPairObj;
+            long replSequenceLId = positionPair[0];
+            long replSequenceEId = positionPair[1];
```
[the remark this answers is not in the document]
poorbarcode: Added a check for the error "out-of-index exception", I think the check by `if` has better performance than `try-catch`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -321,27 +327,160 @@ public boolean isEnabled() {
...
+        return isDuplicateNormal(publishContext, headersAndPayload, true);
+    }
+
+    private void setContextPropsIfRepl(PublishContext publishContext, ByteBuf headersAndPayload) {
+        // Case-1: is a replication marker.
+        if (publishContext.isMarkerMessage()) {
+            // Message is coming from replication, we need to use the replication's producer name, ledger id and entry
+            // id for the purpose of deduplication.
+            int readerIndex = headersAndPayload.readerIndex();
+            MessageMetadata md = Commands.parseMessageMetadata(headersAndPayload);
```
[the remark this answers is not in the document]
poorbarcode: Changed to `peekMessageMetadata`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
poorbarcode: This logic branch is used for the following case:
- Replicator sends msg (origin-position: `3:1`)
  - Pending msg queue of the internal producer of the replicator is `[3:1]` now. 
- Replicator sends msg (origin-position: `3:1`), it repeats send the same msg
  -  Pending msg queue of the internal producer of replicator is `[3:1], [3:1]` now.
- This logic branch will remove the first record in Pending msg queue of the internal producer

gaoran10: Got it, thanks.

gaoran10 (approved):

Technoboy- (approved):

BewareMyPower:
@poorbarcode @gaoran10 @Technoboy- @codelipenghui Why could this PR be cherry-picked to branch-3.0 and branch-4.0?

It changes the `PulsarApi.proto` by adding a new **feature** flag. How could 3.0 and 4.0 be declared as LTS releases?

poorbarcode:
> @poorbarcode @gaoran10 @Technoboy- @codelipenghui Why could this PR be cherry-picked to branch-3.0 and branch-4.0?
> It changes the PulsarApi.proto by adding a new feature flag. How could 3.0 and 4.0 be declared as LTS releases?

- This PR fixed the bug and `3.0.x & 4.0.x` also need this fix.
- the change of `feature flag` is used for the compatibility

codelipenghui:
> This PR fixed the bug and 3.0.x & 4.0.x also need this fix.
the change of feature flag is used for the compatibility


That's reasonable @poorbarcode, but we'd better start a discussion in the community to get aligned since the fix has changed the core logic and protocol for geo-replication.

PR #23708, opened 2024-12-10
[improve][fn] Improve implementation for maxPendingAsyncRequests async concurrency limit when return type is CompletableFuture<Void>

### LINKED ISSUE #23706: [Bug] Pulsar Functions async function concurrency limit handling is not optimal when return type is CompletableFuture<Void>
### Search before asking

### Read release policy

### Version

all released versions

### Minimal reproduce step / Problem description

When analyzing the Pulsar Functions async function handling code, there are performance concerns around this code

https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java#L135-L158

The problem this introduces is that in the case of an asynchronous function that has a `CompletableFuture<Void>` return type and doesn't return results, there will be unnecessary processing delays when the concurrency limit is reached. Whenever there's a slow response in the head of the queue, the request processing will pause. This will introduce unnecessary latency for functions which use `Context.newOutputMessage(...).sendAsync()` for sending messages to multiple topics.

The Pulsar Functions async function handling code should be optimal from performance perspective when using async functions with `CompletableFuture<Void>` return type.

### What did you see instead?

Based on the code, it could lead into unnecessary CPU compared since when the result is `CompletableFuture<Void>`, no messages will be sent to an output topic and there's no need for preserving the ordering of processing the async results.

### Anything else?

Instead, a simple java.util.concurrent.Semaphore based implementation should be used as a solution for limiting concurrency for implementing `maxPendingAsyncRequests` when the return type is `CompletableFuture<Void>`.

lhotari:
## Motivation

When using asynchronous functions in Pulsar Functions, concurrency needs to be limited to prevent unbounded resource consumption. This is currently handled by the `maxPendingAsyncRequests` configuration parameter (default value: 1000). Without such limits, asynchronous processing could lead to unbounded message consumption from input topics, potentially causing resource exhaustion (out-of-memory errors) and function crashes.

However, the current implementation has a performance issue specifically for asynchronous functions that return `CompletableFuture<Void>`. In these cases, the current queue-based implementation introduces unnecessary processing delays when the concurrency limit is reached. In particular:

1. When there's a slow response at the head of the queue, request processing gets delayed unnecessarily when the queue is full. This makes the actual concurrency limit lower than the configured limit and it becomes nondeterministic. Under heavy load, the concurrency limit will be introducing backpressure and this is why any slow request will cause an unnecessary additional latency to the messages in the backlog. This could lead to additional cumulative latency for backlogged messages since all previous delays impact the processing latency of later messages. This type of latency is avoidable with the solution explained in this PR.
2. This introduces avoidable latency for functions using `Context.newOutputMessage(...).sendAsync()` to send messages to multiple topics
3. The current queue-based ordering preservation is unnecessary when no results are being returned (when return type is `CompletableFuture<Void>`)

## Modifications
This PR improves the implementation by:

1. Using a `Semaphore` instead of `LinkedBlockingQueue` for concurrency limiting when the function's return type is `CompletableFuture<Void>`
2. Preserving the existing queue-based implementation for other return types where output message ordering needs to match input ordering
3. Making exception handling more robust by properly unwrapping completion exceptions
4. Simplifying the `JavaExecutionResult` class by removing unused `systemException` field
5. Adding a unit test to verify the new implementation

## Implementation Details

- Added logic to detect `CompletableFuture<Void>` return type through function configuration
- Introduced a `Semaphore`-based concurrency limiter that gets used when order preservation isn't needed
- Maintained backwards compatibility by keeping queue-based implementation for non-void return types
- Updated error handling to better handle asynchronous exceptions and to avoid the extra wrapping of `Throwable` with `new Exception(t)` type of code.
- Added test coverage for the new concurrency limiting approach

## Testing

Added new unit test `testAsyncFunctionMaxPendingVoidResult()` that verifies:
- Proper concurrency limiting with Semaphore
- Correct handling of async void results

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaExecutionResult.java
+++ b/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaExecutionResult.java
@@ -27,13 +27,11 @@ import lombok.Data;
  */
 @Data
 public class JavaExecutionResult {
-    private Exception userException;
-    private Exception systemException;
+    private Throwable userException;
     private Object result;
 
     public void reset() {
         setUserException(null);
-        setSystemException(null);
         setResult(null);
     }
 }
--- a/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
+++ b/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
@@ -25,11 +25,13 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.Semaphore;
 import java.util.function.Consumer;
 import lombok.AccessLevel;
 import lombok.Data;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.functions.api.Function;
 import org.apache.pulsar.functions.api.Record;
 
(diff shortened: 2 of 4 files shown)


lhotari:
The previous solution would retain order of the results. I'll have to check when that is necessary when whether that could be made configurable.

nodece (approved):

PR #23719, opened 2024-12-12
[fix] [broker] Fix config replicationStartAt does not work when set it to earliest

poorbarcode:
### Motivation

```java
@FieldContext(
        category = CATEGORY_SERVER,
        dynamic = true,
        doc = "The position that replication task start at, it can be set to earliest or latest (default).")
private String replicationStartAt = "latest";
```

The value of the config named `replicationStartAt` are `earliest` or `latest`, but both implementation and test used `MessageId.earliest.toString()` that value is `-1:-1:-1`( so the test works as expected). 

Therefore, setting `replicationStartAt` to `earliest` does not work.

### Modifications

- Fix the incorrect implementation and keep compatibility with the previous version.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -1449,9 +1449,7 @@ public class ServiceConfiguration implements PulsarConfiguration {
     @FieldContext(
             category = CATEGORY_SERVER,
             dynamic = true,
-            doc = "The position that replication task start at, it can be set to earliest or latest (default). Note: do"
-                    + " not use \"MessageId.earliest\" because its value is \"-1:-1:-1\", which does not equals"
-                    + " \"earliest\"")
+            doc = "The position that replication task start at, it can be set to earliest or latest (default).")
     private String replicationStartAt = "latest";
 
     @FieldContext(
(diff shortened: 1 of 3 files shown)


gaoran10 (approved):
Could we support both formats?

poorbarcode:
@gaoran10 

> Could we support both formats?

Yes, the current implementation supports both formats

shibd (approved):

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -1449,7 +1449,9 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
     @FieldContext(
             category = CATEGORY_SERVER,
             dynamic = true,
-            doc = "The position that replication task start at, it can be set to earliest or latest (default).")
+            doc = "The position that replication task start at, it can be set to earliest or latest (default). Note: do"
+                    + " not use \"MessageId.earliest\" because its value is \"-1:-1:-1\", which does not equals"
+                    + " \"earliest\"")
```
lhotari, line 1454: Why would someone set a String value in the config to "MessageId.earliest"?
The notes about MessageId.earliest is simply confusing. It's better not to add the note at all. 
```suggestion
            doc = "The position that replication task start at, it can be set to earliest or latest (default).")
```

codelipenghui (approved):

PR #23038, opened 2024-07-16
[improve][client] Add exception handle for client send error

congbobo184:
### Motivation

- producer send messages return error will close the current cnx
- if close the current cnx, all producers and consumers witch use this cnx will close and reconnect
(https://github.com/apache/pulsar/blob/5c6602cbb3660a696bf960f2847aac1a2ae037d2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L323-L345)
- this will influence a lot of producers and consumers
- we only close this producer and reconnect enough, don't need to close this cnx

### Modifications

receive send_error, close current producer connection, then it will reconnect automatically

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23721, opened 2024-12-12
[improve] [test] Add more test for the case that client receives a SendError, which relates to the PR #23038

poorbarcode:
### Motivation

https://github.com/apache/pulsar/pull/23038 changed the behavior of the PulsarClient in handling SendErrors.

### Modifications

Add more tests for the change
- Create producers with a customized name, to confirm users will not get the error `Producer with the same id is already created`
- Create a pulsar client with  a value that does not equal `1`, to confirm users will not get the error `Producer with the same id is already created`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
@@ -196,14 +196,15 @@ public class ClientCnxTest extends MockedPulsarServiceBaseTest {
     }
 
     @Test
-    public void testCnxReceiveSendErrorWithConfigMultiPerConnection() throws Exception {
+    public void testCnxReceiveSendErrorWithMultiConnectionsPerBroker() throws Exception {
         PulsarClient client = PulsarClient.builder().serviceUrl(lookupUrl.toString())
                 .connectionsPerBroker(1000).build();
 
         // Create a producer with customized name.
         final String tp = BrokerTestUtil.newUniqueName(NAMESPACE + "/tp");
         admin.topics().createNonPartitionedTopic(tp);
-        Producer<String> p = client.newProducer(Schema.STRING).producerName("p1").topic(tp).create();
+        ProducerImpl<String> p =
+                (ProducerImpl<String>) client.newProducer(Schema.STRING).producerName("p1").topic(tp).create();
 
         // Inject a persistence error.
         org.apache.pulsar.broker.service.Producer serverProducer = pulsar.getBrokerService().getTopic(tp, false)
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
```diff
@@ -173,11 +188,47 @@ public void testCnxReceiveSendError() throws Exception {
...
         producerTwo.send("test");
+        producerThree.send("test");
         producerTwo.close();
         producerOne.close();
+        producerThree.close();
         client.close();
     }
 
+    @Test
+    public void testCnxReceiveSendErrorWithConfigMultiPerConnection() throws Exception {
```
codelipenghui, line 199: testCnxReceiveSendErrorWithMultiConnectionsPerBroker?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
```diff
@@ -173,11 +188,47 @@ public void testCnxReceiveSendError() throws Exception {
...
+        Producer<String> p = client.newProducer(Schema.STRING).producerName("p1").topic(tp).create();
+
+        // Inject a persistence error.
+        org.apache.pulsar.broker.service.Producer serverProducer = pulsar.getBrokerService().getTopic(tp, false)
+                .join().get().getProducers().values().iterator().next();
+        ServerCnx serverCnx = (ServerCnx) serverProducer.getCnx();
+        serverCnx.getCommandSender().sendSendError(serverProducer.getProducerId(), 1/* sequenceId */,
+                ServerError.PersistenceError, "mocked error");
+        // Wait for the client receives the error.
+        Thread.sleep(1000);
```
codelipenghui, line 215: Is it possible to use Awaitability to check the client has received the error?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
poorbarcode: Renamed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
poorbarcode: I added Ping-Pong twice to confirm the client has handled the Error we sent

```
// Wait for the client to receive the error.
// If the client confirmed two Pings, it means the client has handled the PersistenceError we sent.
serverCnx.checkConnectionLiveness().join();
serverCnx.checkConnectionLiveness().join();
```

dao-jun (approved):

Technoboy- (approved):

PR #23738, opened 2024-12-16
[fix] [client] Fix memory leak when publishing encountered a corner case error

poorbarcode:
### Motivation

**Issue 1: memory leak if get errors when publishing**
Conditions:
- Send queue is full
- Or reaches the limitation named `max message size`
- Or publishes after closing the producer
- Or encounters an error when calling `ProducerInterceptor. eligible`
- see `testSendQueueIsFull`, `testSendMessageSizeExceeded`, `testSendAfterClosedProducer`, and `testInterceptorError`

### Modifications

- Fix issues

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -483,19 +486,42 @@ private ByteBuf applyCompression(ByteBuf payload) {
...
         final int uncompressedSize = payload.readableBytes();
 
+        // To guarantee compatibility with customized "SendCallback", see detail the doc of this method.
+        if ((callback instanceof ProducerImpl.DefaultSendMessageCallback
+            || callback.getClass().getName().endsWith("NonPersistentReplicator$ProducerSendCallback")
+            || callback.getClass().getName().endsWith("ProducerSendCallback$ProducerSendCallback"))
+            && msg.getDataBuffer().capacity() > 0) {
+            checkArgument(msg.getDataBuffer().refCnt() >= 2,
+                    "Message's data's refCnt is less than 2, see #23738");
+        }
```
BewareMyPower, line 516: Such type assertions really should be avoided. I mean, the `checkArgument(message instanceof MessageImpl)` assertion before is also redundant.

The reference count validation should not be performed in `sendAsync`, instead, it should be validated before the caller calls `sendAsync` (e.g. in `GeoPersistentReplicator#replicateEntries`).

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
```diff
@@ -0,0 +1,356 @@
...
+    public void testSendAfterClosedProducer() throws Exception {
+        final String topicName = BrokerTestUtil.newUniqueName("persistent://public/default/tp_");
+        admin.topics().createNonPartitionedTopic(topicName);
+        ProducerImpl<String> producer =
+                (ProducerImpl<String>) pulsarClient.newProducer(Schema.STRING).topic(topicName).create();
+        // Publish after the producer was closed.
+        MsgPayloadTouchableMessageBuilder<String> msgBuilder = newMessage(producer);
+        producer.close();
+        msgBuilder.value("msg-1").sendAsync().exceptionally(ex -> {
+            log.warn("expected error", ex);
```
BewareMyPower, line 234: If it's an expected error, don't print warn logs. You'd better use `assertThrows` or a trivial `try-catch`:

```java
try {
    msgBuilder.value("msg-1").send();
    fail();
} catch (PulsarClientException ignored) {
    // or you can add the type assertion here
}
```

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
```diff
@@ -0,0 +1,356 @@
...
+            } catch (Exception ex) {
+                log.warn("", ex);
+                break;
+            }
+        }
+        try{
+            sendFutureList.get(sendFutureList.size() - 1).join();
+        } catch (Exception ex) {
+            log.warn("", ex);
+        }
```
BewareMyPower, line 96: Which case is expected? Should it succeed or fail?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
```diff
@@ -0,0 +1,356 @@
...
+        int indexCalledSend = 0;
+        List<CompletableFuture> sendFutureList = new ArrayList<>();
+        for (MsgPayloadTouchableMessageBuilder<String> msgBuilder: msgBuilderList) {
+            try {
+                sendFutureList.add(msgBuilder.value("msg-1").sendAsync());
+                if (indexCalledSend != 99) {
+                    indexCalledSend++;
+                }
+            } catch (Exception ex) {
+                log.warn("", ex);
```
BewareMyPower, line 88: I see many places calling `log.warn("", ex)`. If the exception is expected, please do type assertions (like `assertTrue(e instanceof xxx)` or `assertTrue(e.getMessage().contains("xxx")`) rather than printing the whole exception stack

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
```diff
@@ -0,0 +1,356 @@
...
+                        }
+
+                    }
+                }).create();
+
+        MsgPayloadTouchableMessageBuilder<String> msgBuilder = newMessage(producer);
+        try {
+            msgBuilder.value("msg-1").sendAsync().get(3, TimeUnit.SECONDS);
+        } catch (Exception ex) {
+            log.warn("Intercept error", ex);
```
BewareMyPower, line 305: Yet another unnecessary exception stack logs, which is noisy. And you should call `fail()` in the try block.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
poorbarcode: Removed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Solved

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Solved

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Solved

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Solved

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java
```diff
@@ -42,8 +43,9 @@ public class ConnectionHandler {
     private volatile ClientCnx clientCnx = null;
     @Getter
     @Setter
+    @VisibleForTesting
     // Since the `clientCnx` variable will be set to null at some times, it is necessary to save this value here.
-    private volatile int maxMessageSize = Commands.DEFAULT_MAX_MESSAGE_SIZE;
+    volatile int maxMessageSize = Commands.DEFAULT_MAX_MESSAGE_SIZE;
```
BewareMyPower, line 48: There is already a getter and setter, we don't need to remove the `private` modifier here.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -483,19 +486,33 @@ private ByteBuf applyCompression(ByteBuf payload) {
...
+     * Otherwise, please confirm the value "refCnt()" of {@param message} {@link MessageImpl#getDataBuffer}' is "2"
+     * when you call this method, because the ByteBuf will be released twice:
+     * 1. When enabled "isBatchMessagingEnabled":
+     *      1-1. Releasing when it was pushed into the batched message queue, see: {@link #doBatchSendAndAdd}
+     *      1-2. Releasing in the method {@link SendCallback#sendComplete(Throwable, OpSendMsgStats)}.
+     * 2. Single message, in other words, disabled "isBatchMessagingEnabled":
+     *      2-1. Releasing when it was writen out by
+     *           {@link ChannelOutboundHandler#write(ChannelHandlerContext, Object, ChannelPromise)}.
+     *      2-2. Releasing in the method {@link SendCallback#sendComplete(Throwable, OpSendMsgStats)}.
+     */
```
BewareMyPower, line 501:
```suggestion
    /**
     * Note on ByteBuf Release Behavior.
     *
     * <p>If you have a customized callback, please ignore the note below.</p>
     *
     * <p>When using the default callback, please confirm that the {@code refCnt()} value of the {@code message}
     * (as returned by {@link MessageImpl#getDataBuffer}) is {@code 2} when you call this method. This is because
     * the {@code ByteBuf} will be released twice under the following conditions:</p>
     *
     * <ul>
     *   <li><b>Batch Messaging Enabled:</b>
     *     <ol>
     *       <li>Release 1: When the message is pushed into the batched message queue (see {@link #doBatchSendAndAdd}).
     *       </li>
     *       <li>Release 2: In the method {@link SendCallback#sendComplete(Throwable, OpSendMsgStats)}.</li>
     *     </ol>
     *   </li>
     *   <li><b>Single Message (Batch Messaging Disabled):</b>
     *     <ol>
     *       <li>Release 1: When the message is written out by
     *       {@link ChannelOutboundHandler#write(ChannelHandlerContext, Object, ChannelPromise)}.</li>
     *       <li>Release 2: In the method {@link SendCallback#sendComplete(Throwable, OpSendMsgStats)}.</li>
     *     </ol>
     *   </li>
     * </ul>
     */
```

Improve the Javadocs via Copilot.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
```diff
@@ -0,0 +1,370 @@
...
+import org.apache.pulsar.common.util.FutureUtil;
+import org.awaitility.Awaitility;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+@Slf4j
+@Test(groups = "broker-api")
+public class ProducerCornerCaseTest extends ProducerConsumerBase {
```
BewareMyPower, line 50: It's better to rename it with `ProducerSendMemoryLeakTest` or just `ProducerMemoryLeakTest`. "corner case" is too general.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
BewareMyPower: it's not solved. When I debugged this test, it never reached here. Should we remove the `try-catch` block?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
BewareMyPower: it's not solved. After I debugged this test, it should fail with `ProducerQueueIsFullError`, so I suggested the following change

```java
        try {
            sendFutureList.get(sendFutureList.size() - 1).join();
            fail();
        } catch (CompletionException e) {
            assertTrue(e.getCause() instanceof PulsarClientException.ProducerQueueIsFullError);
        }
```

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Removed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Renamed to `ProducerMemoryLeakTest`

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerCornerCaseTest.java
poorbarcode: Accepted the suggestion and removed `fail()` to avoid flaky

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java
poorbarcode: Solved

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+        public void setMsgMocker(Function<MessageImpl<T>, MessageImpl<T>> msgMocker) {
+            this.msgMocker = msgMocker;
+        }
+
+        @Override
+        public Message<T> getMessage() {
+            beforeSend();
+            MessageImpl<T> msg = MessageImpl.create(msgMetadata, content, schema,
+                    producer != null ? producer.getTopic() : null);
+            payload = getInternalState(msg, "payload");
```
BewareMyPower, line 353: Whenever possible, don't use reflection unless it's hard to change the visibility of the private field. In this case, `MessageImpl` already provides a `getPayload` method visible for testing. Even it does not provide, it's better to add one rather than using reflection.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java
```diff
@@ -45,10 +46,10 @@ public class TypedMessageBuilderImpl<T> implements TypedMessageBuilder<T> {
...
     private static final ByteBuffer EMPTY_CONTENT = ByteBuffer.allocate(0);
 
-    private final transient ProducerBase<?> producer;
-    private final transient MessageMetadata msgMetadata = new MessageMetadata();
-    private final transient Schema<T> schema;
-    private transient ByteBuffer content;
+    protected final transient ProducerBase<?> producer;
+    protected final transient MessageMetadata msgMetadata = new MessageMetadata();
+    protected final transient Schema<T> schema;
+    protected transient ByteBuffer content;
```
BewareMyPower, line 52: You don't need to change these access modifiers, as well as the `beforeSend` method.

You can reuse the `getMessage` method when you want to mock it in `MsgPayloadTouchableMessageBuilder`.

```java
        @Override
        public Message<T> getMessage() {
            final var msg = (MessageImpl<T>) super.getMessage();
            payload = msg.getPayload().retain();
            if (msgMocker == null) {
                return msg;
            } else {
                return msgMocker.apply(msg);
            }
        }
```

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+    private <T> MsgPayloadTouchableMessageBuilder<T> newMessage(ProducerImpl<T> producer){
+        Schema<T> schema = getInternalState(producer, "schema");
+        return new MsgPayloadTouchableMessageBuilder<T>(producer, schema);
+    }
+
+    private static class MsgPayloadTouchableMessageBuilder<T> extends TypedMessageBuilderImpl {
+
+        public volatile ByteBuf payload;
+
+        private volatile Function<MessageImpl<T>, MessageImpl<T>> msgMocker;
```
BewareMyPower, line 339: This field is never used?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+    protected void cleanup() throws Exception {
+        super.internalCleanup();
+    }
+
+    @Override
+    protected void doInitConf() throws Exception {
+        super.doInitConf();
+        conf.setTopicLevelPoliciesEnabled(false);
+    }
+
```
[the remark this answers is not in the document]
poorbarcode: Removed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+        final String topicName = BrokerTestUtil.newUniqueName("persistent://public/default/tp_");
+        admin.topics().createNonPartitionedTopic(topicName);
+        ProducerImpl<String> producer = (ProducerImpl<String>) pulsarClient.newProducer(Schema.STRING)
+                .blockIfQueueFull(false).maxPendingMessages(1)
+                .enableBatching(true).topic(topicName).create();
+        List<MsgPayloadTouchableMessageBuilder<String>> msgBuilderList = new ArrayList<>();
+        for (int i = 0; i < 100; i++) {
+            msgBuilderList.add(newMessage(producer));
+        }
+        int indexCalledSend = 0;
```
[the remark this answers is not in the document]
poorbarcode: Changed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+            assertEquals(msgBuilder.payload.refCnt(), 1);
+        }
+
+        // cleanup.
+        for (int i = indexCalledSend; i > -1; i--) {
+            MsgPayloadTouchableMessageBuilder<String> msgBuilder = msgBuilderList.get(i);
+            msgBuilder.release();
+            assertEquals(msgBuilder.payload.refCnt(), 0);
+        }
+        for (int i = indexCalledSend + 1; i < 100; i++) {
```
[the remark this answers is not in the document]
poorbarcode: But it is necessary, to avoid flaky.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+            MsgPayloadTouchableMessageBuilder<String> msgBuilder = msgBuilderList.get(i);
+            assertEquals(msgBuilder.payload.refCnt(), 1);
+        }
+
+        // cleanup.
+        for (int i = indexCalledSend; i > -1; i--) {
+            MsgPayloadTouchableMessageBuilder<String> msgBuilder = msgBuilderList.get(i);
+            msgBuilder.release();
+            assertEquals(msgBuilder.payload.refCnt(), 0);
+        }
```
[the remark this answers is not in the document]
poorbarcode: Changed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    @Test(dataProvider = "maxMessageSizeAndCompressions")
+    public void testSendMessageSizeExceeded(int maxMessageSize, CompressionType compressionType) throws Exception {
+        final String topicName = BrokerTestUtil.newUniqueName("persistent://public/default/tp_");
+        admin.topics().createNonPartitionedTopic(topicName);
+        ProducerImpl<String> producer = (ProducerImpl<String>) pulsarClient.newProducer(Schema.STRING).topic(topicName)
+                .compressionType(compressionType)
+                .enableBatching(false)
+                .compressionType(CompressionType.NONE)
```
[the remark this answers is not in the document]
poorbarcode: Removed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java
poorbarcode: Changed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
poorbarcode: Removed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+        admin.topics().delete(topicName);
+        //admin.topics().deleteAsync(topicName).get(10, TimeUnit.SECONDS); TODO fix bug.
+    }
+
+    @DataProvider
+    public Object[][] failedInterceptAt() {
+        return new Object[][]{
+          {"eligible"},
+          {"beforeSend"},
+          {"onSendAcknowledgement"},
```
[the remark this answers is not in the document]
poorbarcode: Added and formatted

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+        assertEquals(msgBuilder.payload.refCnt(), 1);
+
+        // cleanup.
+        msgBuilder.release();
+        assertEquals(msgBuilder.payload.refCnt(), 0);
+        admin.topics().delete(topicName);
+    }
+
+    private <T> MsgPayloadTouchableMessageBuilder<T> newMessage(ProducerImpl<T> producer){
+        Schema<T> schema = getInternalState(producer, "schema");
```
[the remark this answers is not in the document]
poorbarcode: Changed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,367 @@
...
+        Awaitility.await().untilAsserted(() -> {
+            assertEquals(producer.getPendingQueueSize(), 0);
+        });
+        assertEquals(msgBuilder.payload.refCnt(), 1);
+
+        // cleanup.
+        msgBuilder.release();
+        assertEquals(msgBuilder.payload.refCnt(), 0);
+        admin.topics().delete(topicName);
+        //admin.topics().deleteAsync(topicName).get(10, TimeUnit.SECONDS); TODO fix bug.
```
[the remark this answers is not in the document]
poorbarcode: Removed

dao-jun (approved):
I've took a quick review, the fix in ProducerImpl LGTM

codelipenghui (commented):
Wow, we missed so many cases to release the buffer.

The fix looks good me, but when I try to run the newly added test on the master branch (without this fix). A lot of test cases can still get passed. Could you please help if this is expected?

image

poorbarcode:
- `testSendMessageSizeExceeded`: it should fail when the `@param is larger than 1`, I improved the test, and it is fine now.
- `testBatchedSendMessageSizeExceeded`: it is expected that the tests can pass because the batch message will only check the max message size that excludes message metadata
- `testInterceptorError`: only `@param eligible` has an issue, so it is expected.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
poorbarcode: Solved

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,376 @@
...
+        public volatile ByteBuf payload;
+
+        public <T> MsgPayloadTouchableMessageBuilder(ProducerBase producer, Schema<T> schema) {
+            super(producer, schema);
+        }
+
+        @Override
+        public Message<T> getMessage() {
+            beforeSend();
+            MessageImpl<T> msg = (MessageImpl<T>) super.getMessage();
```
BewareMyPower, line 365: `super.getMessage()` already calls `beforeSend`, so I think you don't need to make any change to `beforeSend`

poorbarcode: Removed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java
```diff
@@ -20,6 +20,7 @@
```
BewareMyPower, line 20: The changes of this file are not needed anymore

poorbarcode: Removed the change

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java
```diff
@@ -20,6 +20,7 @@
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.pulsar.client.util.TypeCheckUtil.checkType;
+import com.google.common.annotations.VisibleForTesting;
```
BewareMyPower, line 23:
```suggestion
```

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,375 @@
...
+    @Override
+    protected void cleanup() throws Exception {
+        super.internalCleanup();
+    }
+
+    @Override
+    protected void doInitConf() throws Exception {
+        super.doInitConf();
+    }
+
```
BewareMyPower, line 72:
```suggestion
```

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,370 @@
...
+    @Test(dataProvider = "maxMessageSizeAndCompressions")
+    public void testSendMessageSizeExceeded(int maxMessageSize, CompressionType compressionType) throws Exception {
+        final String topicName = BrokerTestUtil.newUniqueName("persistent://public/default/tp_");
+        admin.topics().createNonPartitionedTopic(topicName);
+        ProducerImpl<String> producer = (ProducerImpl<String>) pulsarClient.newProducer(Schema.STRING).topic(topicName)
+                .compressionType(compressionType)
+                .enableBatching(false)
+                .create();
+        final ClientCnx cnx = producer.getClientCnx();
```
BewareMyPower, line 131: I don't get why we need to get `cnx` here. It seems it only calls `cnx.close()` in the following cide.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerMemoryLeakTest.java
```diff
@@ -0,0 +1,370 @@
...
+            MsgPayloadTouchableMessageBuilder<String> msgBuilder = msgBuilderList.get(i);
+            assertEquals(msgBuilder.payload.refCnt(), 1);
+            msgBuilder.release();
+            assertEquals(msgBuilder.payload.refCnt(), 0);
+        }
+        for (int i = sendFutureList.size(); i < 100; i++) {
+            MsgPayloadTouchableMessageBuilder<String> msgBuilder = msgBuilderList.get(i);
+            msgBuilder.release();
+            assertEquals(msgBuilder.payload.refCnt(), 0);
+        }
```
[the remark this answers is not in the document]
poorbarcode: > Is there any case that sendFutureList.size() is not 100?

I think no case, removed.

BewareMyPower (approved):

lhotari:
@poorbarcode @BewareMyPower Please review #23761, that addresses a long time issue where the completable futures of pending messages weren't completed when close was called. It's also a case which could result into a resource leak situation.

PR #23757, opened 2024-12-19
[improve][client] Make replicateSubscriptionState nullable

nodece:
### Motivation

The current implementation of subscription replication relies on consumer control with the following configuration:

```java
boolean replicateSubscriptionState = false;
// boolean replicateSubscriptionState = true;
pulsarClient.newConsumer(Schema.STRING)
                .replicateSubscriptionState(replicateSubscriptionState)
```

By default, the `replicateSubscriptionState` is set to `false` in the `ConsumerConfigurationData.java`, which creates a mismatch with the `CommandSubscribe` definition, where `replicate_subscription_state` is optional. To address this, I propose changing the default value of `replicateSubscriptionState` to `null`, aligning it with the optional nature of the field in `CommandSubscribe`.

This adjustment would enhance flexibility in managing subscription replication policies, such as applying the policy that comes from the namespace or topic level, which streamlines the configuration of subscription replication behavior. This feature is in progress.

These changes are compatible with both the old and new versions, there won't be any breaking changes.

### Modifications

- The `replicateSubscriptionState` type is changed from `boolean` to `Boolean` and the default value is `null` in the`ConsumerConfigurationData.java`. When it is null, it won't set the value to the `replicate_subscription_state` of `CommandSubscribe`.
- In the broker, use the `Boolean` instead of `boolean` to match the `replicate_subscription_state` of `CommandSubscribe`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ConsumerConfigurationData.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ConsumerConfigurationData.java
@@ -381,7 +381,7 @@ public class ConsumerConfigurationData<T> implements Serializable, Cloneable {
             value = "If `replicateSubscriptionState` is enabled, a subscription state is replicated to geo-replicated"
                     + " clusters."
     )
-    private boolean replicateSubscriptionState = false;
+    private Boolean replicateSubscriptionState;
 
     private boolean resetIncludeHead = false;
 
@@ -436,5 +436,13 @@ public class ConsumerConfigurationData<T> implements Serializable, Cloneable {
         } catch (CloneNotSupportedException e) {
             throw new RuntimeException("Failed to clone ConsumerConfigurationData");
         }
+    }
+
+    /**
+     * @deprecated Using {@link #getReplicateSubscriptionState()} instead.
+     */
+    @Deprecated
+    public boolean isReplicateSubscriptionState() {
+        return replicateSubscriptionState != null && replicateSubscriptionState;
     }
 }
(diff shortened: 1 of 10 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ConsumerConfigurationData.java
```diff
@@ -437,4 +437,12 @@ public ConsumerConfigurationData<T> clone() {
             throw new RuntimeException("Failed to clone ConsumerConfigurationData");
         }
     }
+
+    /**
+     * @deprecated Using {@link #getReplicateSubscriptionState()} instead.
+     */
+    @Deprecated
```
lhotari, line 444: Is `@JsonIgnore` needed on this method?

lhotari (approved):
> This adjustment would enhance flexibility in managing subscription replication policies, such as applying the policy that comes from the namespace or topic level, which streamlines the configuration of subscription replication behavior. This feature is in progress.

This will be a useful feature.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ConsumerConfigurationData.java
nodece: @lhotari Yes, `@JsonIgnore` has been added.

nodece: `@JsonProperty(access = JsonProperty.Access.READ_WRITE)` is required, this will ensure that the ·replicateSubscriptionState· field is properly handled during JSON serialization and deserialization

lhotari (approved):

lhotari:
> > This adjustment would enhance flexibility in managing subscription replication policies, such as applying the policy that comes from the namespace or topic level, which streamlines the configuration of subscription replication behavior. This feature is in progress.
> 
> This will be a useful feature.

@nodece Please check #23769, it seems to be a related new feature.

nodece:
@lhotari  Sure, go to #23769 and #23770.

PR #9292, opened 2021-01-24
[fix][broker] Support large number of unack message store for cursor recovery

rdhabalia:
### Motivation
Right now, managed-cursor serializes individually deleted-message ranges and persist them in bookie ledger-entry for the recovery. managed-cursor can persist max 150K ranges due to limited bookie entry size (5MB). 150K ranges might not be enough for some of the use cases while recovering cursor. so, we need a mechanism to persist a couple of million ranges to support such usecases.

### Modification
with #3818 and #3819 managed-cursor manages individually deleted messages in bitset with OpenRangeSet. Serializing OpenRangeSet can allow managed-cursor to store 10M ranges with 5MB data size. 

### Result
Usecases require a large number of individually deleted-messages that can be supported with this change.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23759, opened 2024-12-19
[fix] [ml] Fix cursor metadata compatability issue when switching the config unackedRangesOpenCacheSetEnabled

poorbarcode:
### Motivation

- https://github.com/apache/pulsar/pull/9292 introduced a new serialization about cursor metadata, but it will be lost. information when switching the config  `unackedRangesOpenCacheSetEnabled`. You can reproduce the issue with the new test `testUnackmessagesAndRecoveryCompatibility`
- [[DISCUSS] Add an optional config to disable the feature that compresses cursor metadata, which was contributed by #9292](https://lists.apache.org/thread/kfm7n6xf9pf7h76k4gx83zv3c2kj6yy1)

### Modifications
- Fix the issue
- Add an optional config `managedLedgerPersistIndividualAckAsLongArray` to enable or disable the feature that https://github.com/apache/pulsar/pull/9292 introduced, and set the default value to `false`. 

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -639,7 +639,22 @@ public class ManagedCursorImpl implements ManagedCursor {
             try {
                 Map<Long, long[]> rangeMap = rangeList.stream().collect(Collectors.toMap(LongListMap::getKey,
                         list -> list.getValuesList().stream().mapToLong(i -> i).toArray()));
-                individualDeletedMessages.build(rangeMap);
+                // Guarantee compatability for the config "unackedRangesOpenCacheSetEnabled".
+                if (getConfig().isUnackedRangesOpenCacheSetEnabled()) {
+                    individualDeletedMessages.build(rangeMap);
+                } else {
+                    RangeSetWrapper<Position> rangeSetWrapperV2 = new RangeSetWrapper<>(positionRangeConverter,
+                            positionRangeReverseConverter, true,
+                            getConfig().isPersistentUnackedRangesWithMultipleEntriesEnabled());
+                    rangeSetWrapperV2.build(rangeMap);
+                    rangeSetWrapperV2.forEach(range -> {
+                        individualDeletedMessages.addOpenClosed(range.lowerEndpoint().getLedgerId(),
+                                range.lowerEndpoint().getEntryId(), range.upperEndpoint().getLedgerId(),
+                                range.upperEndpoint().getEntryId());
+                        return true;
+                    });
+                    rangeSetWrapperV2.clear();
+                }
             } catch (Exception e) {
                 log.warn("[{}]-{} Failed to recover individualDeletedMessages from serialized data", ledger.getName(),
                         name, e);
@@ -2367,7 +2382,14 @@ public class ManagedCursorImpl implements ManagedCursor {
                     }
                     // Add a range (prev, pos] to the set. Adding the previous entry as an open limit to the range will
                     // make the RangeSet recognize the "continuity" between adjacent Positions.
-                    Position previousPosition = ledger.getPreviousPosition(position);
+                    // Before https://github.com/apache/pulsar/pull/21105 is merged, the range does not support crossing
+                    // multi ledgers, so the first position's entryId maybe "-1".
+                    Position previousPosition;
+                    if (position.getEntryId() == 0) {
+                        previousPosition = PositionFactory.create(position.getLedgerId(), -1);
+                    } else {
+                        previousPosition = ledger.getPreviousPosition(position);
+                    }
                     individualDeletedMessages.addOpenClosed(previousPosition.getLedgerId(),
                         previousPosition.getEntryId(), position.getLedgerId(), position.getEntryId());
                     MSG_CONSUMED_COUNTER_UPDATER.incrementAndGet(this);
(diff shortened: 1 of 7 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2367,7 +2383,14 @@ public void asyncDelete(Iterable<Position> positions, AsyncCallbacks.DeleteCallb
...
                     // make the RangeSet recognize the "continuity" between adjacent Positions.
-                    Position previousPosition = ledger.getPreviousPosition(position);
+                    // Before https://github.com/apache/pulsar/pull/21105 is merged, the range does not support crossing
+                    // multi ledgers, so the first position's entryId maybe "-1".
+                    Position previousPosition;
+                    if (position.getEntryId() == 0) {
+                        previousPosition = PositionFactory.create(position.getLedgerId(), -1);
+                    } else {
+                        previousPosition = ledger.getPreviousPosition(position);
+                    }
```
codelipenghui, line 2393: It's better to move to the internal of the getPreviousPosition() method.

codelipenghui (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: > It's better to move to the internal of the getPreviousPosition() method.

`getPreviousPosition` returns a meaningful position, in other words, it returns a position that exactly exists. And other methods rely on this feature, such as https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3621-L3623

```java
if (specifiedLac.getEntryId() < 0) {
     // Calculate the meaningful LAC.
     Position actLac = getPreviousPosition(specifiedLac);
    ...
}
```

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -2243,6 +2243,10 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
             + " will only be tracked in memory and messages will be redelivered in case of"
             + " crashes.")
     private int managedLedgerMaxUnackedRangesToPersist = 10000;
+    @FieldContext(
+            category = CATEGORY_STORAGE_ML,
+            doc = "Whether persist cursor ack stats as long arrays, which will compress the data and reduce GC rate")
+    private boolean managedLedgerPersistIndividualAckAsLongArray = false;
```
lhotari, line 2249: Does change the existing Pulsar 4.0 behavior when this defaults to `false`?

poorbarcode: Yes, will rollback the default behavior to `3.0.x`, in other words, removed the default behavior changes that introduced by #9292

shibd (approved):

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
eolivelli: As written on the mailing list I think that we cannot break compatibility with Pulsar 4.0.0 and 4.0.1, it is "worse" than breaking compatibility with "3.x"

lhotari (changes requested):
Thanks for the contribution, @poorbarcode . This is an essential improvement.

I'd like to request a change for the default value of `managedLedgerPersistIndividualAckAsLongArray`. It should be set to `true` in this PR. There's a long mailing list thread about this:
https://lists.apache.org/thread/kfm7n6xf9pf7h76k4gx83zv3c2kj6yy1

When we say there's support for downgrading (rollback), it doesn't
mean users don't need to take any action.

We ensure downgrade compatibility by providing an upgrade guide that
explains how to configure the system to allow rollbacks without losing
state information. If we didn't do this, we'd always be stuck in the
same situation whenever a new LTS version comes out.

Let's start making progress:
- keep PR 9292 as default
- implement PR 23759 without changing the PR 9292 default
- write a proper upgrade guide for Pulsar 4.0 where rollback
considerations are explained (separate PR)

poorbarcode:
@lhotari 

Answered your question in the discussion channel: https://lists.apache.org/thread/kfm7n6xf9pf7h76k4gx83zv3c2kj6yy1

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
thetumbled: I support that keeping compatibility should comes first, which impact most of the users.
and i wonder whether we can fix the issue without introducing the switch to disable the improvement introduced by #9292.

poorbarcode:
Hi Lari

Since there is no more arguments in the email list, I dismissed your change request

lhotari:
> Hi Lari
> 
> Since there is no more arguments in the email list, I dismissed your change request

@poorbarcode In Apache projects, we continue discussions until there's consensus on important decisions. We haven't reached consensus yet. Let's postpone cherry-picking to branch-4.0 until there's consensus.

lhotari:
@poorbarcode I have replied in https://lists.apache.org/thread/rhb4hbbrqw3238l0563n9nm6s57jrb00 .

Issue #11780, opened 2021-08-25
producer.sendAsync is interrupted by client.close()


geomagilles:
**Describe the bug**
If the client closes while sendAsync() is processed, the message is lost. 

**To Reproduce**
see https://github.com/infiniticio/pulsar-bug-clientAsync

**Expected behavior**
The message should be sent before actually closing the client (https://apache-pulsar.slack.com/archives/C5Z4T36F7/p1629752614463900?thread_ts=1629723649.441700&cid=C5Z4T36F7)

**Additional context**
none

avirlrma:
Hi folks,

Can I take this up? @sijie

codelipenghui:
The issue had no activity for 30 days, mark with Stale label.

geomagilles:
@codelipenghui: @avirlrma proposed to help

PR #23761, opened 2024-12-19
[Fix][Client] Fix pending message not complete when closeAsync

pengxiangrui127:
### Motivation
In the current process of closeAsync() in ProducerImpl.java, the client will firstly set the producer to the Closing state, then stop timerTask of sendTimeout, and then send a http request to broker to close the producer. When the client receives a successful response from broker, it sets the producer to the Closed state, removes the producer from the producers in client, and complete the PendingMessage with AlreadyClosedException.
However, if the communication between client and broker is abnormal due to network or other reasons, the PendingMessage will not be cleared anymore. What's worse is that since the producer is already in the Closing state at this time, there is no chance to trigger the cleanup of PendingMessage by calling the closeAsync() again, which is obviously unreasonable.

### Modifications

There is no need to wait for the broker's response to clean up PendingMessage. Once you decide to shut down the producer, there is no need to retain the PendingMessage in the producer. At this time, failing quickly is more valuable than waiting.

### Verifying this change

Add a test case to ensure closeAndClearPendingMessages() has been invoke when request broker error after closeAsync().

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
@@ -1142,11 +1142,11 @@ public class ProducerImpl<T> extends ProducerBase<T> implements TimerTask, Conne
         CompletableFuture<Void> closeFuture = new CompletableFuture<>();
         cnx.sendRequestWithId(cmd, requestId).handle((v, exception) -> {
             cnx.removeProducer(producerId);
+            closeAndClearPendingMessages();
             if (exception == null || !cnx.ctx().channel().isActive()) {
                 // Either we've received the success response for the close producer command from the broker, or the
                 // connection did break in the meantime. In any case, the producer is gone.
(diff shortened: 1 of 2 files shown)


lhotari:
There's a related issue #11780

lhotari:
There's also #12195

lhotari (approved):
LGTM. This PR makes sense. 

I initially thought that the intention was to also address issues #11780 and #12195 as well. There could be a separately improvement addressing the problem that closeAsync doesn't flush any pending messages before closing. Usually there's an expectation coming from java.io.OutputStream that when you call close, it will flush before closing. This doesn't happen with Producer's close.

nodece (approved):

PR #23766, opened 2024-12-20
[fix][admin] Fix exception loss in getMessageId method

### LINKED ISSUE #23765: [Bug] TimeoutException encountered while accessing the admin's getMessageById API in version 3.0.7
### Search before asking

### Read release policy

### Version

OS：centos7
Jdk：17
Pulsar version：3.0.7

### Minimal reproduce step

1. Add a test case to the testGetMessageById method in PersistentTopicsTest.java.

2. Specifically, you can add the following code:
`Assert.expectThrows(PulsarAdminException.ServerSideErrorException.class, () -> {
    admin.topics().getMessageById(topicName1, id1.getLedgerId(), id1.getEntryId() + 10);
});
`

3. Run this test case to reproduce the issue. You will encounter the following error:
`Caused by: org.apache.pulsar.client.admin.PulsarAdminException$TimeoutException: java.util.concurrent.TimeoutException
    at org.apache.pulsar.client.admin.internal.BaseResource.sync(BaseResource.java:347)
    at org.apache.pulsar.client.admin.internal.TopicsImpl.getMessageById(TopicsImpl.java:1010)
    at org.apache.pulsar.broker.admin.PersistentTopicsTest.lambda$testGetMessageById$11(PersistentTopicsTest.java:1385)
    at org.testng.Assert.expectThrows(Assert.java:2440)
    ... 29 more
`

The issue occurs when trying to query a non-existent message, which usually happens when a topic is newly created but hasn't received any traffic yet. In such cases, querying some information about the topic might invoke this API, leading to a timeout.

For this scenario, I would expect a fast failure, rather than being blocked until the timeout occurs.

### What did you see instead?

What I observed instead is that the getMessageById request gets blocked until the timeout occurs.

The hidden risk is that, since the timeout duration is uncertain, if the user has not configured a timeout (e.g., PulsarAdmin.builder().readTimeout(5, TimeUnit.SECONDS);) or if the timeout configuration is unreasonable, it can cause the TCP connection to enter a CLOSE_WAIT state. In extreme cases, this could potentially lead to a tcp.listenOverflow, which can affect other functionalities.

The following image shows a large number of connections in the CLOSE_WAIT state on the broker's 8080 port:
image

### Anything else?

_No response_

danpi:
Fixes #23765

### Motivation
#23765 demonstrated how to reproduce the TimeoutException bug and explained the potential issue. Therefore, this PR aims to fix this problem.

### Modifications
1. The exception is returned via completeExceptionally instead of being thrown directly.
2. A new test case is added to testGetMessageById to verify the effect of the fix.

### Verifying this change

- Added test in
org/apache/pulsar/broker/admin/PersistentTopicsTest.java#testGetMessageById

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -2774,9 +2774,10 @@ public class PersistentTopicsBase extends AdminResource {
                         public void readEntryFailed(ManagedLedgerException exception,
                                                     Object ctx) {
                             if (exception instanceof ManagedLedgerException.LedgerNotExistException) {
-                                throw new RestException(Status.NOT_FOUND, "Message id not found");
+                                results.completeExceptionally(
+                                        new RestException(Status.NOT_FOUND, "Message id not found"));
                             }
-                            throw new RestException(exception);
+                            results.completeExceptionally(new RestException(exception));
                         }
 
                         @Override
@@ -2784,7 +2785,7 @@ public class PersistentTopicsBase extends AdminResource {
                             try {
                                 results.complete(generateResponseWithEntry(entry, (PersistentTopic) topic));
                             } catch (IOException exception) {
-                                throw new RestException(exception);
+                                results.completeExceptionally(new RestException(exception));
                             } finally {
                                 if (entry != null) {
                                     entry.release();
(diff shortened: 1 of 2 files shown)


danpi:
The reason behind this is that there is a possibility of calling the asyncReadEntry0 method in RangeEntryCacheImpl, where at line 272, a thread switch occurs, causing the exception to be lost. However, in the case of an asynchronous call, if we cannot predict when the thread switch will happen, returning the exception via completeExceptionally instead of throwing it directly might be a better approach.

BewareMyPower (approved):

PR #23770, opened 2024-12-23
[improve][pip] PIP-398: Subscription replication on the namespace and topic levels

nodece:
### Motivation

Enhance the subscription replication.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-398.md
+++ b/pip/pip-398.md
@@ -1,0 +1,216 @@
+# PIP-398: Subscription replication on the namespace and topic levels
+
+# Background knowledge
+
+https://github.com/apache/pulsar/pull/4299 introduces the subscription replication feature on the consumer level:
+
+```java
+Consumer<byte[]> consumer = pulsarClient.newConsumer().topic("topic").replicateSubscriptionState(false/*true*/)
+        .subscriptionName("sub").subscribe();
+```
+
+While this provides flexibility, it introduces overhead in managing replication for a large number of consumers. Users
+need to manually enable the `replicateSubscriptionState` flag for each consumer, which can become cumbersome in
+large-scale deployments.
+
+# Motivation
+
+The key motivation behind this PIP is to simplify subscription replication configuration, especially in failover
+scenarios. When a main cluster goes down and a backup cluster is activated, ensuring that subscription states are
+consistently replicated across clusters is critical for failover scenarios. By extending the replication configuration
+to the namespace and topic levels, the system reduces the need for explicit consumer-level configuration.
+
+# Goals
+
+## In Scope
+
+The PIP aims to provide management of subscription replication at the namespace and topic levels using the Pulsar Admin
+CLI and API.
+
+# Out scope
+
+This feature depends on https://github.com/apache/pulsar/pull/23757, which makes the subscription replication nullable
+on the consumer level. Otherwise, the namespace and topic levels will be ignored.
+
+# High Level Design
+
+Introduces a configuration used for enabling subscription replication on the namespace and topic levels, when enabled,
+all consumers under the topic will automatically replicate their subscription states to remote clusters.
+
+The priority for the subscription replication configuration is as follows:
+
+- consumer level > topic level > namespace level.
+- If `replicateSubscriptionState` is set at the consumer level, configurations at the topic and namespace levels are
+  ignored.
+- If set at the topic level, the namespace-level configuration is ignored.
+
+# Detailed Design
+
+## Design & Implementation Details
+
+### Namespace level
+
+1. Add the field `Boolean replicate_subscriptions_state` to the `org.apache.pulsar.common.policies.data.Policies` class
+   to control subscription replication at the namespace level:
+    ```java
+    public class Policies {
+        @SuppressWarnings("checkstyle:MemberName")
+        public Boolean replicate_subscriptions_enabled;
+    }
+    ```
+2. Add the management methods to the `org.apache.pulsar.client.admin.Namespaces` interface:
+     ```java
+     public interface Namespaces {
+         void setReplicateSubscriptionsState(String namespace, Boolean enabled) throws PulsarAdminException;
+         CompletableFuture<Void> setReplicateSubscriptionsStateAsync(String namespace, Boolean enabled);
+         Boolean getReplicateSubscriptionsState(String namespace) throws PulsarAdminException;
+         CompletableFuture<Boolean> getReplicateSubscriptionsStateAsync(String namespace);
+     }
+     ```
+3. Implement the management methods in the `org.apache.pulsar.client.admin.internal.NamespacesImpl` class.
+
+### Topic level
+
+1. Add the field `Boolean replicateSubscriptionsState` to the `org.apache.pulsar.common.policies.data.TopicPolicies`
+   class to enable subscription replication at the topic level:
+   ```java
+    public class TopicPolicies {
+         public Boolean replicateSubscriptionsState;
+    }
+    ```
+2. Add the management methods to the `org.apache.pulsar.client.admin.TopicPolicies` interface:
+   ```java
+       public interface Topics { 
+           void setReplicateSubscriptionsState(String topic, Boolean enabled) throws PulsarAdminException;
+           CompletableFuture<Void> setReplicateSubscriptionsStateAsync(String topic, Boolean enabled);
+           Boolean getReplicateSubscriptionsState(String topic, boolean applied) throws PulsarAdminException;
+           CompletableFuture<Boolean> getReplicateSubscriptionsStateAsync(String topic, boolean applied);
+       }
+   ```
+3. Implement the management methods in the `org.apache.pulsar.client.admin.internal.TopicPoliciesImpl` class.
+
+### Consumer level
+
+The consumer-level configuration takes precedence over both the topic and namespace levels, ensuring that
+consumer-specific configurations can still override the more general settings.
+
+### Subscription replication applied
+
+When the subscription replication is changed on the namespace or topic level, the subscription replication will be
+applied to all consumers under the topic.
+
+Please notice that the `org.apache.pulsar.client.admin.Topics.setReplicatedSubscriptionStatus()` can override the
+subscription replication configuration for the special subscription, and equals the consumer level.
+
+## Public-facing Changes
+
+### Public API
+
+#### REST API
+
+##### Namespace level
+
+- `/{tenant}/{namespace}/replicateSubscriptionsState`: enable/disable/remove the subscription replication on the
+  namespace level.
+    - Method: `POST`
+    - Content-Type: `application/json`
+    - Body:
+        - true
+        - false
+        - null
+- `GET /{tenant}/{namespace}/replicateSubscriptionsState` to get subscription replication configuration on the namespace
+  level.
+    - Method: `GET`
+    - Response:
+        - true
+        - false
+        - null
+
+##### Topic level
+
+- `/{tenant}/{namespace}/{topic}/replicateSubscriptionsState`: enable/disable/remove the subscription replication on the
+  topic level.
+    - Method: `POST`
+    - Content-Type: `application/json`
+    - Body:
+        - true
+        - false
+        - null
+- `/{tenant}/{namespace}/{topic}/replicateSubscriptionsState` to get subscription replication configuration on the topic
+  level.
+    - Method: `GET`
+    - Parameters:
+        - `applied=true`: get the applied subscription replication configuration, if topic is not set, return the
+          namespace level configuration.
+        - `applied=false`: get the applied subscription replication configuration, if topic is set, return the topic
+          level configuration.
+    - Response:
+        - true
+        - false
+        - null
+
+### CLI
+
+#### Namespace level
+
+- `pulsar-admin namespaces set-replicate-subscriptions-state <tenant>/<namespace> --enabled true/false` to
+  enable/disable the subscription replication on the namespace level.
+- `pulsar-admin namespaces get-replicate-subscriptions-state <tenant>/<namespace>` to get the subscription
+  replication configuration on the namespace level.
+- `pulsar-admin namespaces remove-replicate-subscriptions-state <tenant>/<namespace>` to remove the subscription
+  replication configuration on the namespace level.
+
+#### Topic level
+
+- `pulsar-admin topicPolicies set-replicate-subscriptions-state <tenant>/<namespace>/<topic> --enabled true/false`
+  to enable/disable the subscription replication on the topic level.
+- `pulsar-admin topicPolicies get-replicate-subscriptions-state <tenant>/<namespace>/<topic>` to get the
+  subscription replication configuration on the topic level.
+- `pulsar-admin topicPolicies remove-replicate-subscriptions-state <tenant>/<namespace>/<topic>` to remove the
+  subscription replication configuration on the topic level.
+
+# Security Considerations
+
+Both write and read operations require the necessary permissions, which already exist in Pulsar.
+
+## Namespace level
+
+- Write the subscription replication configuration:
+    - Required: `PolicyName.REPLICATED_SUBSCRIPTION` with `WRITE` permission.
+- Read the subscription replication configuration:
+    - Required: `PolicyName.REPLICATED_SUBSCRIPTION` with `READ` permission.
+
+## Topic level
+
+- Write the subscription replication configuration
+    - Required: `TopicOperation.SET_REPLICATED_SUBSCRIPTION_STATUS` permission.
+- Read the subscription replication configuration:
+    - Required: `TopicOperation.GET_REPLICATED_SUBSCRIPTION_STATUS` permission.
+
+# Backward & Forward Compatibility
+
+## Upgrade
+
+None.
+
+## Downgrade / Rollback
+
+If the broker is downgrade, users will need to ensure that the consumer-level configuration is re-enabled to maintain
+the correct replication behavior.
+
+## Pulsar Geo-Replication Upgrade & Downgrade/Rollback Considerations
+
+None.
+
+# Alternatives
+
+None.
+
+# General Notes
+
+None.
+
+# Links
+
+* Mailing List discussion thread:
+* Mailing List voting thread:


lhotari:
@nodece would it be possible to cover #23769 feature too? That enables subscription replication for all subscriptions at the broker level if I understand it correctly. Could you work together with @yyj8 to cover that?

nodece:
> @nodece would it be possible to cover #23769 feature too? That enables subscription replication for all subscriptions at the broker level if I understand it correctly. Could you work together with @yyj8 to cover that?

#23769 aims to overwrite the subscription replication state from the consumer, which is different from this PIP. Actually, I also need an overwrite feature,  and then the users use the admin API to manage the subscription replication.

The downstream users sometimes enable/disable the subscription replication on the consumer, this lead to the namespace or topic level is ignored.

lhotari:
> > @nodece would it be possible to cover #23769 feature too? That enables subscription replication for all subscriptions at the broker level if I understand it correctly. Could you work together with @yyj8 to cover that?
> 
> #23769 aims to overwrite the subscription replication state from the consumer, which is different from this PIP. Actually, I also need an overwrite feature, and then the users use the admin API to manage the subscription replication.
> 
> The downstream users sometimes enable/disable the subscription replication on the consumer, this lead to the namespace or topic level is ignored.
> 
> We can also merge #23769 to this PIP. What's your opinion? @lhotari @yyj8

@nodece @yyj8 Please work together to combine both cases since they are both valuable and useful. I think it would make
sense to combine the work in a single PIP to ensure that we achieve consistency across the changes.

yyj8:
> > > @nodece would it be possible to cover #23769 feature too? That enables subscription replication for all subscriptions at the broker level if I understand it correctly. Could you work together with @yyj8 to cover that?
> > 
> > 
> > #23769 aims to overwrite the subscription replication state from the consumer, which is different from this PIP. Actually, I also need an overwrite feature, and then the users use the admin API to manage the subscription replication.
> > The downstream users sometimes enable/disable the subscription replication on the consumer, this lead to the namespace or topic level is ignored.
> > We can also merge #23769 to this PIP. What's your opinion? @lhotari @yyj8
> 
> @nodece @yyj8 Please work together to combine both cases since they are both valuable and useful. I think it would make sense to combine the work in a single PIP to ensure that we achieve consistency across the changes.



> > > @nodece would it be possible to cover #23769 feature too? That enables subscription replication for all subscriptions at the broker level if I understand it correctly. Could you work together with @yyj8 to cover that?
> > 
> > 
> > #23769 aims to overwrite the subscription replication state from the consumer, which is different from this PIP. Actually, I also need an overwrite feature, and then the users use the admin API to manage the subscription replication.
> > The downstream users sometimes enable/disable the subscription replication on the consumer, this lead to the namespace or topic level is ignored.
> > We can also merge #23769 to this PIP. What's your opinion? @lhotari @yyj8
> 
> @nodece @yyj8 Please work together to combine both cases since they are both valuable and useful. I think it would make sense to combine the work in a single PIP to ensure that we achieve consistency across the changes.

This is a great idea and suggestion. Can we consider a strategy that divides into three dimensions: cluster dimension, namespace dimension, and topic dimension.
(1)PR https://github.com/apache/pulsar/pull/23769 As a cluster dimension strategy, merge it into this PIP https://github.com/apache/pulsar/pull/23770;
(2)PR https://github.com/apache/pulsar/pull/23769 add the function of dynamically configuring parameter `replicateAllSubscriptionState` values through the command `bin/pulse admin brokers update-dynamic-config 
 --config replicateAllSubscriptionState --value true/false`

@nodece @lhotari Can you see if this method is feasible?

nodece:
@yyj8 

> cluster dimension, namespace dimension, and topic dimension.

Once the consumer level is configured, it is the highest priority, the cluster, namespace, and topic levels will be ignored, please see https://github.com/apache/pulsar/pull/23769#issuecomment-2559221226 for details.

> (2)PR https://github.com/apache/pulsar/pull/23769 add the function of dynamically configuring parameter `replicateAllSubscriptionState` values through the command `bin/pulse admin brokers update-dynamic-config  --config replicateAllSubscriptionState --value true/false`

This is feasible.

yyj8:
> @yyj8
> 
> > cluster dimension, namespace dimension, and topic dimension.
> 
> Once the consumer level is configured, it is the highest priority, the cluster, namespace, and topic levels will be ignored, please see [#23769 (comment)](https://github.com/apache/pulsar/pull/23769#issuecomment-2559221226) for details.
> 
> > (2)PR #23769 add the function of dynamically configuring parameter `replicateAllSubscriptionState` values through the command `bin/pulse admin brokers update-dynamic-config  --config replicateAllSubscriptionState --value true/false`
> 
> This is feasible.

The default configuration for the client is `replicateSubscriptionState=false`. In the scenario of cluster migration, we hope that there is no need for business code modification, and the server will control the replication of the state of all subscriptions in the cluster uniformly. If the configuration priority of the client is the highest, and the client does not display the configuration `replicateSubscriptionState=true`, this goes against our original intention of not requiring business code modification.

My suggestion is to configure the cluster dimension. If `repliceAllSubscriptState=true` is enabled, it will be a mandatory configuration to overwrite the client. If `repliceAllSubscriptState=false`, then use the client's configuration.

nodece:
@yyj8 
> The default configuration for the client is `replicateSubscriptionState=false`.

The latest client defaults to `null`, please see https://github.com/apache/pulsar/pull/23757.

> In the scenario of cluster migration, we hope that there is no need for business code modification, and the server will control the replication of the state of all subscriptions in the cluster uniformly. If the configuration priority of the client is the highest, and the client does not display the configuration `replicateSubscriptionState=true`, this goes against our original intention of not requiring business code modification.
My suggestion is to configure the cluster dimension. If `repliceAllSubscriptState=true` is enabled, it will be a mandatory configuration to overwrite the client. If `repliceAllSubscriptState=false`, then use the client's configuration.

I understand your idea, this can make all subscriptions replicated when `repliceAllSubscriptState=true`, when true, this equals the consumer with `replicateSubscriptionState=true`, and they have the highest.

However, this idea conflicts with PIP-398, which assumes that the consumer-level configuration for `replicateSubscriptionState` is not set, and instead, replication behavior is driven by namespace and topic policies. If the `replicateSubscriptionState` is set,  it means that the specified subscription will be replicated, namespace and topic levels cannot change this behavior. If changed, it will result in a breaking change. This is worth considering. 

In our case, I don't want to configure the `replicateSubscriptionState` on the consumer, but the user has been configured, I cannot change the user code. We have the same case, so I want to ignore the `replicateSubscriptionState`. 

## My idea
To combine your case and my case, I suggest introducing a broker configuration for overwriting the consume configuration, and assuming the namespace and topic level are set, the final result will be so like:

- `overwirteConsumerReplicateSubscriptionState=true`:  The subscription will be replicated. 
- `overwirteConsumerReplicateSubscriptionState=false`: The subscription will be replicated. 
- `overwirteConsumerReplicateSubscriptionState=null`: This mean the consumer level is null, and then the broker depends on the namespace and topic level to check the subscription replication.

yyj8:
> @yyj8
> 
> > The default configuration for the client is `replicateSubscriptionState=false`.
> 
> The latest client defaults to `null`, please see #23757.
> 
> > In the scenario of cluster migration, we hope that there is no need for business code modification, and the server will control the replication of the state of all subscriptions in the cluster uniformly. If the configuration priority of the client is the highest, and the client does not display the configuration `replicateSubscriptionState=true`, this goes against our original intention of not requiring business code modification.
> > My suggestion is to configure the cluster dimension. If `repliceAllSubscriptState=true` is enabled, it will be a mandatory configuration to overwrite the client. If `repliceAllSubscriptState=false`, then use the client's configuration.
> 
> I understand your idea, this can make all subscriptions replicated when `repliceAllSubscriptState=true`, when true, this equals the consumer with `replicateSubscriptionState=true`, and they have the highest.
> 
> However, this idea conflicts with PIP-398, which assumes that the consumer-level configuration for `replicateSubscriptionState` is not set, and instead, replication behavior is driven by namespace and topic policies. If the `replicateSubscriptionState` is set, it means that the specified subscription will be replicated, namespace and topic levels cannot change this behavior. If changed, it will result in a breaking change. This is worth considering.
> 
> In our case, I don't want to configure the `replicateSubscriptionState` on the consumer, but the user has been configured, I cannot change the user code. We have the same case, so I also want to ignore the `replicateSubscriptionState`.
> 
> ## My idea
> To combine your case and my case, I suggest introducing a broker configuration for overwriting the consume configuration, and assuming the namespace and topic level are set, the final result will be so like this:
> 
> * `overwirteConsumerReplicateSubscriptionState=true`:  The subscription will be replicated.
> * `overwirteConsumerReplicateSubscriptionState=false`: The subscription will be replicated.
> * `overwirteConsumerReplicateSubscriptionState=null`: This means the consumer level is null, and then the broker depends on the namespace and topic level to check the subscription replication.
> 
> We can also introduce `replicateSubscriptionsState=true/false/empty` configuration on the broker level, when the namespace/topic/consumer levels are not set, we use the broker level configuration to check if replicates the subscription.

@nodece I agree with your suggestion.
Before using cluster level configuration, we first check if there are topic and namespace dimensions. If there are none, we apply cluster level configuration; The priority of the three dimensions from high to low is: topic -> namespace -> cluster.

Then we can make unified code adjustments on your pip, or you can directly merge my code into your code, or after you submit the code, I can synchronize your code and modify the cluster level configuration before submitting.

nodece:
> Before using cluster level configuration, we first check if there are topic and namespace dimensions. If there are none, we apply cluster level configuration; The priority of the three dimensions from high to low is: topic -> namespace -> cluster.

@yyj8 Correct. The `replicateSubscriptionState` is used to auto-create a subscription. When set to `false`, the broker doesn't persist that to the cursor property in the old version. Therefore, we only need to add `replicateSubscriptionState=true/false/null` at the cluster level.

**Breaking Change**: For an existing subscription where `replicateSubscriptionState=false` is set at the consumer level, the new broker will use the subscription replication policy from the topic/namespace/cluster level.

The new subscription will follow the above priority.

/cc @lhotari

nodece:
@yyj8 This PIP has been updated, could you have a chance to review this PIP?

yyj8:
> @yyj8 This PIP has been updated, could you have a chance to review this PIP?

Are you referring to updating the code and modifying the functional design specifications in this pip [pip/pip-398.md](https://github.com/apache/pulsar/pull/23770/files#diff-67fc7a48cc071911c2239d1c628335d4147f54345051f36efd2f18dbcc8339c6) design documentation?

nodece:
> > @yyj8 This PIP has been updated, could you have a chance to review this PIP?
> 
> Are you referring to updating the code and modifying the functional design specifications in this pip [pip/pip-398.md](https://github.com/apache/pulsar/pull/23770/files#diff-67fc7a48cc071911c2239d1c628335d4147f54345051f36efd2f18dbcc8339c6) design documentation?

Just for this PIP.

pip/pip-398.md
```diff
@@ -0,0 +1,245 @@
...
+        - true
+        - false
+        - null
+- `GET /{tenant}/{namespace}/replicateSubscriptionState` to get subscription replication configuration on the namespace
+  level.
+    - Method: `GET`
+    - Response:
+        - true
+        - false
+        - null
```
lhotari, line 142: Just wondering if this would be more of a namespace policy matter at the namespace level

pip/pip-398.md
```diff
@@ -0,0 +1,245 @@
...
+    - Method: `GET`
+    - Parameters:
+        - `applied=true`: get the applied subscription replication configuration, if topic is not set, return the
+          namespace level configuration.
+        - `applied=false`: get the applied subscription replication configuration, if topic is set, return the topic
+          level configuration.
+    - Response:
+        - true
+        - false
+        - null
```
lhotari, line 165: It feels that this should be part of the topic policy.

pip/pip-398.md
```diff
@@ -0,0 +1,245 @@
...
+enabled, all consumers under the broker/namespace/topic will automatically replicate their subscription states to remote
+clusters.
+
+The priority for the subscription replication configuration is as follows:
+
+- consumer level > topic level > namespace level > broker level.
+- If `replicateSubscriptionState` is set at the consumer level, configurations at the topic, namespace, and broker levels are
+  ignored.
+- If set at the topic level, the namespace-level configuration is ignored.
+- If set at the namespace level, the broker-level configuration is ignored.
```
lhotari, line 42: I think that there could be a need to override any value set on the consumer side. I also added some comments that controlling the behavior at namespace and consumer level should perhaps be a policy concern instead of creating a separate concept for handling the setting.

If the setting at namespace and topic level would be more than just a true/false. It would be possible to add a separate setting that would allow overriding the consumer level setting.

lhotari (commented):
Added some thoughts.

lhotari:
The term `replicateSubscriptionState` is ambiguous (it's not introduced in this PIP). It's unclear whether it refers to "replicating the subscription state" or represents "the state of replicating subscriptions". This demonstrates why putting more thought into naming concepts could be helpful.

pip/pip-398.md
nodece: Yes.

nodece:
> The term `replicateSubscriptionState` is ambiguous (it's not introduced in this PIP). It's unclear whether it refers to "replicating the subscription state" or represents "the state of replicating subscriptions". This demonstrates why putting more thought into naming concepts could be helpful.

"replicating the subscription state" is correct. This term is from the design of existing consumer.

nodece:
@lhotari Do you have more suggestions? If not, I will send a vote to the mailing list.

lhotari:
> @lhotari Do you have more suggestions? If not, I will send a vote to the mailing list.

@nodece I think that at the topic and namespace level, it could be better to combine `overwriteConsumerReplicateSubscriptionState` (you mentioned it in https://github.com/apache/pulsar/pull/23770#issuecomment-2560597473) and `replicateSubscriptionState` into a single enum instead of 2 separate boolean values to be used in policies. Let's say `ReplicateSubscriptionStatePolicy` enum with values:
* `OVERRIDE_ENABLE` - overrides any value set by the consumer (or lower level policy) and enables subscription replication
* `OVERRIDE_DISABLE` - overrides any value set by the consumer (or lower level policy) and enables subscription replication
* `ENABLE` - enables subscription replication if not explicitly disabled by the consumer (or lower level policy)
* `DISABLE` - disables subscription replication if not explicitly enabled by the consumer (or lower level policy)
I don't know if this is accurate, but putting more thought in how to address the topic and namespace level would be useful. 

I haven't seen the APIs being updated to have this information in the policies. I don't think that introducing a completely separate concept for topic and namespace level makes sense.

nodece:
@lhotari 

This is an important section: https://github.com/apache/pulsar/pull/23770/files#diff-67fc7a48cc071911c2239d1c628335d4147f54345051f36efd2f18dbcc8339c6R110

Since the consumer enables the subscription replication when the `replicateSubscriptionState` is true in the consumer level. Prioritizing `true` over `false` ensures consistency, this approach should be applied to topic, namespace, and broker as well. By doing so, we can avoid unnecessary back-and-forth state changes and maintain smoother transitions.

1. If the consumer level is present and its value is `true`, enable subscription replication.
2. If the consumer level is `false`, and the topic level is present with its value set to `true`, enable subscription replication.
3. If the consumer level is `false`, the topic level is not set, but the namespace level is present and its value is `true`, enable subscription replication.
4. If the consumer level is `false`, and both the topic and namespace levels are not set, but the broker level is present with its value set to `true`, enable subscription replication.

> I haven't seen the APIs being updated to have this information in the policies. I don't think that introducing a completely separate concept for topic and namespace level makes sense.

The `replicateSubscriptionState` follows hierarchy topic policies, with the difference being that we only apply topic/namespace/broker policy to subscription when the consumer level is false.

pip/pip-398.md
```diff
@@ -0,0 +1,245 @@
...
+
+When the subscription replication is changed on the broker, namespace or topic level, the subscription replication will
+be applied to all consumers under the broker/namespace/topic.
+
+We exclusively consider the `true` case for replication. If the consumer level is set to `false`, the
+replication configuration defined at the topic, namespace, or broker level will be applied to the subscription.
+
+There is a special case here, if the user intentionally sets `false` at the consumer level, and then set `true` at the
+topic/ns/broker level, this may disrupt the user's behavior. This way, we can minimize changes to the Pulsar public API
+as much as possible.
```
lhotari, line 120: It would be good to explain what "disrupt the user's behavior" means. Perhaps rewording with the help of Generative AI could be useful since "disruption" is typically used in a different context.

lhotari:
> The `replicateSubscriptionState` follows hierarchy topic policies, with the difference being that we only apply topic/namespace/broker policy to subscription when the consumer level is false.

@nodece Yes, I understand that. Isn't there a need to override any client application consumer level setting with topic or namespace level policies? What if Pulsar administrator would explicitly want to disallow or force using replicated subscriptions for a particular topic or namespace while geo-replication is enabled?

pip/pip-398.md
nodece: The user expected to disable subscription replication, however, found that the feature had been enabled.

nodece:
> Isn't there a need to override any client application consumer level setting with topic or namespace level policies? 

When the topic/namespace level is true, the consumer level will be overried.

> What if Pulsar administrator would explicitly want to disallow or force using replicated subscriptions for a particular topic or namespace while geo-replication is enabled?

The administrator is permitted to enable the subscription replication and subsequently disable it.

lhotari:
@nodece Do you have plans to take this PIP forward?

nodece:
@lhotari Vote has been started.

pip/pip-398.md
```diff
@@ -0,0 +1,253 @@
...
+    }
+    ```
+2. Add the management methods to the `org.apache.pulsar.client.admin.TopicPolicies` interface:
+   ```java
+       public interface TopicPolicies { 
+           void setReplicateSubscriptionState(String topic, Boolean enabled) throws PulsarAdminException;
+           CompletableFuture<Void> setReplicateSubscriptionStateAsync(String topic, Boolean enabled);
+           Boolean getReplicateSubscriptionState(String topic, boolean applied) throws PulsarAdminException;
+           CompletableFuture<Boolean> getReplicateSubscriptionStateAsync(String topic, boolean applied);
+       }
```
lhotari: I would strongly suggest replacing the `Boolean` with an enum. Using a boolean value for `replicateSubscriptionState` is very confusing.

Here's an example what `ReplicateSubscriptionStatePolicy` enum could be:
* `OVERRIDE_ENABLE` - overrides any value set by the consumer (or lower level policy) and enables subscription replication
* `OVERRIDE_DISABLE` - overrides any value set by the consumer (or lower level policy) and disables subscription replication
* `ENABLE` - enables subscription replication if not explicitly disabled by the consumer (or lower level policy)
* `DISABLE` - disables subscription replication if not explicitly enabled by the consumer (or lower level policy)

lhotari:
> @lhotari Vote has been started.

@nodece I added a request for changes for the `replicateSubscriptionState` which is very confusing when a `Boolean` is used. It's very hard to understand what meaning `null`, `TRUE` and `FALSE` would have. Using an enum would resolve this confusion since it's possible to document the enum values and make the meaning explicit.

pip/pip-398.md
nodece: We have a `pulsar-admin topics set-replicated-subscription-status` command, the current implementation uses Boolean for similar that, so I followed the same design for consistency and to avoid introducing a new pattern in the CLI/API.

The meaning is:
- null → remove configuration(I can add a remove api instead of this, and then use boolean instead of Boolean in set API)
- true → enable
- false → disable (**but true has priority if both are set in the different levels**)

Using an enum would make the meaning explicit, but it introduces additional complexity.

nodece: @lhotari

nodece: @lhotari Any updates?

lhotari: > We have a `pulsar-admin topics set-replicated-subscription-status` command, the current implementation uses Boolean for similar that, so I followed the same design for consistency and to avoid introducing a new pattern in the CLI/API.

This is a valid argument if we are able to achieve all required use cases with a `Boolean`. 

After thinking about it again, it seems that these choices need to be covered:

1. override any setting by the consumer (or lower level policy) and enable subscription replication for all subscriptions
2. override any setting by the consumer (or lower level policy) and disable subscription replication for all subscriptions
3. enable subscription replication if not explicitly disabled by the consumer (or lower level policy) for all subscriptions
4. no setting at all (higher level setting would take priority). If all policies (broker, ns, topic) are "empty", the consumer setting would get used directly.

@nodece With a single `Boolean` value, how would you distinguish between the 2 different cases (1. and 3. above) where subscription replication is enabled?

(with "lower level policy", it means the hierarchy of broker -> namespace -> topic)

nodece: > @nodece With a single `Boolean` value, how would you distinguish between the 2 different cases (1. and 3. above) where subscription replication is enabled?

> override any setting by the consumer (or lower level policy) and enable subscription replication for all subscriptions

Subscription replication is enabled if:
- The consumer sets true
- The first non-null policy found in the following order is true: topic → namespace → broker. Once a policy is found (even if false), lower levels are ignored.

> no setting at all (higher level setting would take priority). If all policies (broker, ns, topic) are "empty", the consumer setting would get used directly.

Correct.

I understand the concern about using `null` for removing the setting. To make the semantics clearer, I'll add an explicit `removeReplicateSubscriptionState()` API instead of relying on `null`.
So no need to worry about `Boolean` vs `boolean`, the new API will handle the unset case explicitly.

lhotari: > Subscription replication is enabled if:
> 
> * The consumer sets true
> * The first non-null policy found in the following order is true: topic → namespace → broker. Once a policy is found (even if false), lower levels are ignored.

Ok. I was thinking about a scenario where there would be a need to ignore and override any setting on the consumer. Even if the consumer would enable replication, it would be possible to turn replication off with policies. With a `Boolean` this wouldn't be possible. Perhaps we don't need to support overriding the client side setting in this case.

pip/pip-398.md
```diff
@@ -0,0 +1,253 @@
...
+and topic levels, when enabled, all consumers under the broker/namespace/topic will automatically replicate their
+subscription states to remote clusters.
+
+The priority for the subscription replication configuration is as follows:
+
+- consumer level > topic level > namespace level > broker level.
+- If `replicateSubscriptionState` is set at the consumer level, configurations at the topic, namespace, and broker levels are
+  ignored.
+- If set at the topic level, the namespace-level configuration is ignored.
+- If set at the namespace level, the broker-level configuration is ignored.
```
lhotari: Based on the discussion thread at https://github.com/apache/pulsar/pull/23770/changes#r2905108272 , it seems that the PIP considers the case where the setting is "set and enabled (Boolean.TRUE)".
I misunderstood the design since a value could be considered "set" when it's `Boolean.FALSE`.

nodece: @lhotari You are right.

The replicated flag at the subscription level is persisted in BookKeeper cursor metadata, and only `true` indicates a replicated subscription: https://github.com/apache/pulsar/blob/v4.0.9/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java#L1149-L1152

To keep the semantics consistent across subscription, topic, and namespace
levels, the policy only treats `Boolean.TRUE` as enabling the feature. `false` or `null` both mean "not enabling it".

Otherwise, the topic could end up in inconsistent states where a subscription is replicated while the topic policy explicitly disables this.

lhotari: @nodece Please update this to the PIP document too. It's good to explain the "what", "how" and "why" as you already did in your answer above. Having this in the PIP document will be it clear to everyone.

nodece:
@lhotari This PIP has been updated.

PR #23772, opened 2024-12-23
[fix][ml] Topic load timeout due to ml data ledger future never finishes

poorbarcode:
### Motivation

**Background**
There is a mechanism that repeatedly prevents the callback of ML data ledger creation:
- Start a scheduled task to check whether the creation will be timeout.
- Received a callback
  - Check whether the future(`@param ctx` of `BK.createAsync`) has been done or not.
  - If done: it means the creation has timeout before the creation is completed
  - Otherwise: it is a real callback from BK.

**Issue:**
But the timeout event will call the same callback as above, then the steps are as follows, which you ca reproduce by the test `testCreateDataLedgerTimeout`:
- Start creating a data ledger
  - Call `BK.createAsync` 
- Timeout
  - Mark the future(`@param ctx` of `BK.createAsync`) as completed exceptionally. 
  - Trigger the callback related to ledger creation. 
    - Check whether the future(`@param ctx` of `BK.createAsync`) has been done or not.
    - If done: do nothing.
- Creation is compelled.
  - Trigger the callback related to ledger creation. 
    - Check whether the future(`@param ctx` of `BK.createAsync`) has been done or not.
    - If done: do nothing.
- Issue: The callback for ledger creation will never be called.

Screenshot 2024-12-24 at 00 14 38
Screenshot 2024-12-24 at 00 14 08

### Modifications

Fix the issue

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockBookKeeper.java
+++ b/testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockBookKeeper.java
@@ -40,7 +40,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.BiPredicate;
 import org.apache.bookkeeper.client.AsyncCallback.CreateCallback;
 import org.apache.bookkeeper.client.AsyncCallback.DeleteCallback;
 import org.apache.bookkeeper.client.AsyncCallback.OpenCallback;
@@ -61,7 +60,6 @@ import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.stats.StatsLogger;
 import org.apache.bookkeeper.versioning.LongVersion;
 import org.apache.bookkeeper.versioning.Versioned;
-import org.apache.zookeeper.MockZooKeeper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockBookKeeper.java
+++ b/testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockBookKeeper.java
@@ -20,6 +20,7 @@ package org.apache.bookkeeper.client;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import com.google.common.collect.Lists;
+import io.netty.util.concurrent.DefaultThreadFactory;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockBookKeeper.java
```diff
@@ -97,6 +100,7 @@ public static Collection<BookieId> getMockEnsemble() {
     public PulsarMockBookKeeper(OrderedExecutor orderedExecutor) throws Exception {
         this.orderedExecutor = orderedExecutor;
         this.executor = orderedExecutor.chooseThread();
+        scheduler = Executors.newScheduledThreadPool(2, new DefaultThreadFactory("mock-bk-scheduler"));
```
nodece, line 103:
```suggestion
        scheduler = Executors.newScheduledThreadPool(1, new DefaultThreadFactory("mock-bk-scheduler"));
```

nodece (changes requested):
It looks you should fix the `checkAndCompleteLedgerOpTask()`:
```java
    protected boolean checkAndCompleteLedgerOpTask(int rc, LedgerHandle lh, Object ctx) {
        if (ctx instanceof CompletableFuture) {
            // ledger-creation is already timed out and callback is already completed so, delete this ledger and return.
            if (((CompletableFuture) ctx).complete(lh)) {
                return false;
            } else {
                if (rc == BKException.Code.OK) {
                    log.warn("[{}]-{} ledger creation timed-out, deleting ledger", this.name, lh.getId());
                    asyncDeleteLedger(lh.getId(), DEFAULT_LEDGER_DELETE_RETRIES);
                    return true; // changed
                }
                return false; // changed
            }
        }
        return false;
    }
```

gaoran10 (approved):

shibd:
> It looks you should fix the checkAndCompleteLedgerOpTask():

+1

poorbarcode:
@shibd @nodece 

> It looks you should fix the checkAndCompleteLedgerOpTask():

Good suggestion. Modified.

nodece (approved):

dao-jun:
> It looks you should fix the `checkAndCompleteLedgerOpTask()`:
> 
> ```java
>     protected boolean checkAndCompleteLedgerOpTask(int rc, LedgerHandle lh, Object ctx) {
>         if (ctx instanceof CompletableFuture) {
>             // ledger-creation is already timed out and callback is already completed so, delete this ledger and return.
>             if (((CompletableFuture) ctx).complete(lh)) {
>                 return false;
>             } else {
>                 if (rc == BKException.Code.OK) {
>                     log.warn("[{}]-{} ledger creation timed-out, deleting ledger", this.name, lh.getId());
>                     asyncDeleteLedger(lh.getId(), DEFAULT_LEDGER_DELETE_RETRIES);
>                     return true; // changed
>                 }
>                 return false; // changed
>             }
>         }
>         return false;
>     }
> ```

It seems no need to change here, the following steps will handle the case of `rc != OK`

dao-jun (approved):

shibd (approved):

congbobo184:
> > It looks you should fix the `checkAndCompleteLedgerOpTask()`:
> > ```java
> >     protected boolean checkAndCompleteLedgerOpTask(int rc, LedgerHandle lh, Object ctx) {
> >         if (ctx instanceof CompletableFuture) {
> >             // ledger-creation is already timed out and callback is already completed so, delete this ledger and return.
> >             if (((CompletableFuture) ctx).complete(lh)) {
> >                 return false;
> >             } else {
> >                 if (rc == BKException.Code.OK) {
> >                     log.warn("[{}]-{} ledger creation timed-out, deleting ledger", this.name, lh.getId());
> >                     asyncDeleteLedger(lh.getId(), DEFAULT_LEDGER_DELETE_RETRIES);
> >                     return true; // changed
> >                 }
> >                 return false; // changed
> >             }
> >         }
> >         return false;
> >     }
> > ```
> 
> It seems no need to change here, the following steps will handle the case of `rc != OK`

+1

Issue #19806, opened 2023-03-14
PIP-XYZ: Assign topic partitions to bundles by round robin


315157973:
### Motivation

With all existing Pulsar load balancing algorithms, it is difficult to balance the load of Pulsar cluster nodes. It often happens that some nodes are highly loaded while others are idle, and the CPU of each Broker node is very different.
![image](https://user-images.githubusercontent.com/9758905/224992293-8074379f-eaca-4052-a563-4f0a95eda121.png)
There are three reasons why the existing way will make the cluster unbalanced:
1. The load managed by each Bundle is not even. Even if the number of partitions managed by each bundle is the same, there is no guarantee that the sum of the loads of these partitions will be the same.
2. Doesn't shed loads very well. The existing default policy ThresholdShedder has a relatively high usage threshold, and various traffic thresholds need to be set. Many clusters with high TPS and small message bodies may have high CPU but low traffic; And for many small-scale clusters, the threshold needs to be modified according to the actual business.
3. The removed Bundle cannot be well distributed to other Brokers. The load information of each Broker will be reported at regular intervals, so the judgment of the Leader Broker when allocating Bundles cannot be guaranteed to be completely correct. Secondly, if there are a large number of Bundles to be redistributed, the Leader may make the low-load Broker a new high-load node when the load information is not up-to-date.

### Goal

We reimplemented the algorithm in 3 parts and listed our test report at the end:
1. Assigning partitions to Bundle.
2. Allocating Bundle to Broker.
3. The algorithm for the leader to decide which Bundles on the high-load Broker to offload.

In this PIP, we only discuss the first one: `Assigning partitions to Bundle`.
Let each bundle carry the same load as possible.

### API Changes

1.  Add a configuration item `partitionAssignerClassName`, so that different partition assignment algorithms can be dynamically configured.
The existing algorithm will be used as the default `partitionAssignerClassName=ConsistentHashingPartitionAssigner`
3. Implement a new partition assignment class `RoundRobinPartitionAssigner`

### Implementation

The client sends a message to a multi-partition Topic, which uses polling by default. 
Therefore, we believe that the load of partitions of the same topic is balanced.
We  assign partitions of the same topic to bundle by round-robin. 
In this way, the difference in the number of partitions carried by the bundle will not exceed 1. 
Since we consider the load of each partition of the same topic to be balanced, the load carried by each bundle is also balanced.
[...]

PR #23774, opened 2024-12-24
[improve][broker] Improve the extensibility of the TopicBundleAssignmentStrategy interface class (#23773)

### LINKED ISSUE #23773: [improve][broker] Improve the extensibility of the TopicBundleAssignmentStrategy interface class
### Search before asking

### Motivation

During the lookup process, the pulsar broker needs to determine the bundle to which the topic belongs, depending on the hash value calculated based on the topic name. After the [PIP-255](https://github.com/apache/pulsar/issues/19806) makes the partition assignment strategy pluggable, the process of obtaining the bundle triggered by the lookup is as follows:

`Lookup -> NamespaceService#getBrokerServiceUrlAsync -> NamespaceService#getBundleAsync ->
NamespaceBundles#findBundle -> TopicBundleAssignmentStrategy#findBundle -> NamespaceBundles#getBundle(long hash)`

When determining whether the current broker has the topic, the same hash algorithm needs to be used to calculate the hash value based on the topic name.

`NamespaceService#getOwnedTopicListForNamespaceBundle -> NamespaceBundle#includes -> NamespaceBundleFactory#getLongHashCode -> Range#contains(long hash)`

##
### Code bad smell: scatter-point modification

Unfortunately, in the current implementation of pulsar broker, the preceding two processes that need to calculate the hash value are implemented in different methods. Once the hash calculation method needs to be adjusted, the hash calculation method needs to be modified in different classes, which has a typical bad code smell problem.

The detailed description is as follows:

Take the default implementation class `ConsistentHashingTopicBundleAssigner` of the `TopicBundleAssignmentStrategy` interface class as an example. During the lookup process, the hash value is calculated in `ConsistentHashingTopicBundleAssigner#findBundle`.

https://github.com/apache/pulsar/blob/1967a9309586286580ac0f3b75a34e1f70e63f75/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/ConsistentHashingTopicBundleAssigner.java#L25-L40

However, the process of determining whether the current broker has owned the topic is to calculate the hash value in the `NamespaceBundleFactory#getLongHashCode` method.

https://github.com/apache/pulsar/blob/1967a9309586286580ac0f3b75a34e1f70e63f75/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java#L204

https://github.com/apache/pulsar/blob/1967a9309586286580ac0f3b75a34e1f70e63f75/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/NamespaceBundleFactory.java#L78-L79

https://github.com/apache/pulsar/blob/1967a9309586286580ac0f3b75a34e1f70e63f75/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/NamespaceBundleFactory.java#L294-L296

## Insufficient interface scalability

In addition, the `NamespaceBundleFactory#getLongHashCode` method uses a fixed algorithm to calculate the hash value. Therefore, it is difficult to extend other interface class implementations using different hash algorithms without modifying the `NamespaceBundleFactory#getLongHashCode` method, which violates the open and closed principles.

image

### Solution

It is recommended that the implementation of the `NamespaceBundleFactory#getLongHashCode` method be moved to the implementation class of the interface `TopicBundleAssignmentStrategy`. Therefore, we may add a new method `long getHashCode(String name)`  to the `TopicBundleAssignmentStrategy` interface class. When the `NamespaceBundleFactory#getLongHashCode` method invokes `getHashCode`, `NamespaceBundles` is unaware of the specific hash algorithm implementation.

### Alternatives

_No response_

### Anything else?

_No response_

rayluoluo:
[improve][broker] Improve the extensibility of the TopicBundleAssignmentStrategy interface class (#23773)

Fixes #23773

PIP: #19806 

### Motivation

For details, see the description of issue #23773 

### Modifications

The implementation of the `NamespaceBundleFactory#getLongHashCode` method is moved to the implementation class of the interface `TopicBundleAssignmentStrategy`. Therefore, a new method `long getHashCode(String name)`  is added to the `TopicBundleAssignmentStrategy` interface class. The implementation of the hash algorithm is no longer fixed in the `NamespaceBundleFactory#getLongHashCode` method. Instead, the `getHashCode` method implemented by different algorithms is invoked.

### Verifying this change

This change is already covered by existing tests, such as *TopicBundleAssignmentStrategyTest#testStrategyFactory*, and added tests and can be verified as follows:

*TopicBundleAssignmentStrategyTest*
  - *TopicBundleAssignmentStrategyTest#testRoundRobinBundleAssigner：A new partition assignment strategy algorithm `RoundRobinBundleAssigner` is customized and the `TopicBundleAssignmentStrategy` interface is implemented. Only the interfaces in `TopicBundleAssignmentStrategy` need to be implemented. It's not needed to modify the interface implementation of other classes. The open and closed principles are met.*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [X] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-399.md
+++ b/pip/pip-399.md
@@ -1,0 +1,60 @@
+
+# PIP-399: Fix Metric Name for Delayed Queue
+
+# Background knowledge
+
+Pulsar delayed delivery is a feature that allows messages to be delivered to consumers after a certain delay.
+It will expose a metric to monitor the memory usage of delayed queue.
+
+# Motivation
+
+There is already one metric called `pulsar_delayed_message_index_size_bytes` for the total memory occupation used by delayed queue of **one topic**.
+```
+writeMetric(stream, "pulsar_delayed_message_index_size_bytes", stats.delayedTrackerMemoryUsage,
+                cluster, namespace, topic, splitTopicAndPartitionIndexLabel);
+```
+
+Whereas, the metric for **one sub** also called `pulsar_delayed_message_index_size_bytes`, which do not comform the metric name norm and is confusing.
+```
+writeSubscriptionMetric(stream, "pulsar_delayed_message_index_size_bytes",
+                    subsStats.delayedTrackerMemoryUsage, cluster, namespace, topic, sub, splitTopicAndPartitionIndexLabel);
+```
+
+Currently, it can export metric like:
+```
+# TYPE pulsar_delayed_message_index_size_bytes gauge
+pulsar_delayed_message_index_size_bytes{cluster="MyPulsar",namespace="public/default",topic="persistent://public/default/testNack-partition-0"} 0
+pulsar_delayed_message_index_size_bytes{cluster="MyPulsar",namespace="public/default",topic="persistent://public/default/testNack-partition-0",subscription="sub2"} 0
+```
+
+The metric of topic and subscription mix together. If we want to filter out the metric of sub to pick out the metric of topic, we need to use promsql like:
+`pulsar_delayed_message_index_size_bytes{subscription=""}`
+It is quite weird and not friendly to use.
+
+
+# Goals
+
+Rename the metric for **one sub** to `pulsar_subscription_delayed_message_index_size_bytes`.
+
+
+# Backward & Forward Compatibility
+
+## Upgrade
+
+Rename the metric for **one sub** to `pulsar_subscription_delayed_message_index_size_bytes`.
+
+## Downgrade / Rollback
+
+Rename the metric for **one sub** to `pulsar_delayed_message_index_size_bytes`.
+
+
+
+# General Notes
+
+# Links
+
+<!--
+Updated afterwards
+-->
+* Mailing List discussion thread: https://lists.apache.org/thread/b8rqld3cww1t34zntgmld50yz34lxx1d
+* Mailing List voting thread: https://lists.apache.org/thread/cyyx29ggjdpbr3kq5vvd6tk83f9vc112


rayluoluo:
@315157973 @poorbarcode @liangyepianzhou @lhotari PTAL

pulsar-broker/src/main/java/org/apache/pulsar/common/naming/TopicBundleAssignmentStrategy.java
```diff
@@ -23,5 +23,7 @@
 public interface TopicBundleAssignmentStrategy {
     NamespaceBundle findBundle(TopicName topicName,  NamespaceBundles namespaceBundles);
 
+    long getHashCode(String name);
```
lhotari, line 26: It's better to make the name of this method more specific. It's better to use the type `TopicName` as well and provide a default implementation. Please also add javadoc with proper description of the method.

```suggestion
    default long calculateBundleHashCode(TopicName topicName) {
        return Hashing.crc32().hashString(topicName.toString(), StandardCharsets.UTF_8).padToLong();    
    }
```

pulsar-broker/src/main/java/org/apache/pulsar/common/naming/ConsistentHashingTopicBundleAssigner.java
```diff
@@ -23,18 +23,26 @@
...
+        NamespaceBundle bundle = namespaceBundles.getBundle(getHashCode(topicName.toString()));
         if (topicName.getDomain().equals(TopicDomain.non_persistent)) {
             bundle.setHasNonPersistentTopic(true);
         }
         return bundle;
     }
 
+    @Override
+    public long getHashCode(String name) {
+        return pulsar.getNamespaceService().getNamespaceBundleFactory().getHashFunc()
```
lhotari, line 38: this doesn't look great since it's a ["train wreck" anti-pattern](https://wiki.c2.com/?TrainWreck) to have long call chains to reach out to a dependency. could the hash function be passed in the constructor instead?

pulsar-broker/src/main/java/org/apache/pulsar/common/naming/ConsistentHashingTopicBundleAssigner.java
```diff
@@ -23,18 +23,26 @@
...
+                .hashString(name, StandardCharsets.UTF_8)
+                .padToLong();
+    }
+
     @Override
     public void init(PulsarService pulsarService) {
+        this.pulsar = pulsarService;
```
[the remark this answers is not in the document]
rayluoluo: HashFunc is initialized in the NamespaceBundleFactory. When the ConsistentHashingTopicBundleAssigner#init() method needs to be called to initialize the bundle allocation algorithm instance, pulsar.getNamespaceService() is not ready, pulsar.getNamespaceService().getNamespaceBundleFactory() will cause NPE. Therefore, HashFunc is more suitable for lazy acquisition.

pulsar-broker/src/main/java/org/apache/pulsar/common/naming/TopicBundleAssignmentStrategy.java
rayluoluo: Change as suggested, thank you. @lhotari

pulsar-broker/src/main/java/org/apache/pulsar/common/naming/ConsistentHashingTopicBundleAssigner.java
```diff
@@ -18,23 +18,42 @@
...
+        NamespaceBundle bundle = namespaceBundles.getBundle(calculateBundleHashCode(topicName));
         if (topicName.getDomain().equals(TopicDomain.non_persistent)) {
             bundle.setHasNonPersistentTopic(true);
         }
         return bundle;
     }
 
+    @Override
+    public long calculateBundleHashCode(TopicName topicName) {
+        if (hashFunction == null) {
```
315157973, line 44: It seems not thread safety. Why not use double check lock
```
if (hashFunction == null) {
    synchronized(ConsistentHashingTopicBundleAssigner.class) {
        if (hashFunction == null) {
            hashFunction = getBundleHashFunc();
        }
    }
}
```

rayluoluo: Change as suggested, thank you. @315157973

Issue #23707, opened 2024-12-10
[Enhancement] Cache Schema instances for classes in a weak reference cache since creating an instance could be CPU intensive


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

For example, `Schema.AVRO(SomeClass.class)` is fairly CPU intensive. It would be useful it there would be a weak reference cache for caching the schema instance.  

### Solution

* add a weak reference cache for caching a scheme instance for Schema.AVRO, Schema.JSON, Schema.PROTOBUF.
* Don't return the schema instance directly since schema instances are mutable. Instead return a clone of the cached instance.

### Alternatives

-

### Anything else?

-

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

walkinggo:
Are there any limitations in the specific technical framework for implementing weak references? How should we limit the size of the cache?

PR #23777, opened 2024-12-24
Enhancement schema cache

walkinggo:
Main Issue: #23707 

### Motivation
Cache Schema instances for classes in a weak reference cache since creating an instance could be CPU intensive

### Modifications

 - *Added weak reference cache for schema*
 
- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [✅] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/internal/PulsarClientImplementationBinding.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/internal/PulsarClientImplementationBinding.java
@@ -55,6 +55,12 @@ import org.apache.pulsar.common.schema.SchemaType;
  * The actual implementation of this class is loaded from {@link DefaultImplementation}.
  */
 public interface PulsarClientImplementationBinding {
+    public Schema getAvroSchemaCache(Class clazz);
+    public Schema getProtobufSchemaCache(Class clazz);
+    public Schema getJsonSchemaCache(Class clazz);
+    public void setAvroSchemaCache(Class clazz,Schema schema);
+    public void setProtobufSchemaCache(Class clazz,Schema schema);
+    public void setJsonSchemaCache(Class clazz,Schema schema);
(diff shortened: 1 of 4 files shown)


pulsar-client-api/src/main/java/org/apache/pulsar/client/internal/PulsarClientImplementationBinding.java
```diff
@@ -55,6 +55,12 @@
  * The actual implementation of this class is loaded from {@link DefaultImplementation}.
  */
 public interface PulsarClientImplementationBinding {
+    public Schema getAvroSchemaCache(Class clazz);
+    public Schema getProtobufSchemaCache(Class clazz);
+    public Schema getJsonSchemaCache(Class clazz);
+    public void setAvroSchemaCache(Class clazz,Schema schema);
+    public void setProtobufSchemaCache(Class clazz,Schema schema);
+    public void setJsonSchemaCache(Class clazz,Schema schema);
```
lhotari, line 63: The cache should be encapsulated in the PulsarClientImplementationBindingImpl and not exposed outside of the interface.

lhotari (changes requested):
The implementation approach isn't correct. The cache should be an implementation details of the PulsarClientImplementationBindingImpl and not exposed externally. 
The key in the cache could be the org.apache.pulsar.client.impl.schema.SchemaDefinitionImpl, however that would require adding hashCode&equals implementations to that class. The cache should have weak keys to prevent memory leaks.

lhotari:
There's another PR #23808 where I've provided suggestions about the minimal implementation: https://github.com/apache/pulsar/pull/23808#issuecomment-2569689969

Issue #23705, opened 2024-12-10
[Bug] Pulsar Function processing time doesn't get properly recorded for asynchronous functions


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

any released version

### Minimal reproduce step

In the code, it can be seen that asynchronous functions don't get handled properly. `stats.processTimeEnd()` gets called immediately when the result is returned:
https://github.com/apache/pulsar/blob/6fe8100b1fd5d37a6e1bf33803a8904fa3879321/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstanceRunnable.java#L337-L355

### What did you expect to see?

asynchronous functions would also be handled

### What did you see instead?

asynchronous functions have invalid processing time stats

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #23787, opened 2024-12-26
[Bug][pulsar-function]Pulsar Function processing time doesn't get properly recorded for asynchronous functions

walkinggo:
Main Issue: #23705 

### Motivation
In the code, the asynchronous function is not handled properly. `stats.processTimeEnd()` is called immediately when the result is returned, but we should wait for the asynchronous task to complete before processing.

### Modifications
Passed `stats` as a parameter into the function, and called `stats.processTimeEnd()` after the function call completes. Also, updated the corresponding test cases.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
+++ b/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
@@ -34,6 +34,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.functions.api.Function;
 import org.apache.pulsar.functions.api.Record;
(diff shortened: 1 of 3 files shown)


lhotari (changes requested):
This needs a different approach since there could be multiple messages in flight. The current duration calculation expects one-by-one messages handling.

walkinggo:
> This needs a different approach since there could be multiple messages in flight. The current duration calculation expects one-by-one messages handling.

Sorry, I'm not quite sure I understand your point. Are you referring to the potential issue when multiple messages are being processed, where they might share the same stats variable, leading to the modification of the same processTimeStart variable after calling processTimeStart in ComponentStatsManager?

walkinggo:
I try to finish it in another pr #23811 becasue i use different branch.i will close this pr,please check it in next pr.

PR #23793, opened 2024-12-30
[improve][pip] PIP-401: Support set batching configurations for Pulsar Functions&Sources

jiangpengcheng:
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-401.md
+++ b/pip/pip-401.md
@@ -1,0 +1,98 @@
+# PIP-401: Support set batching configurations for Pulsar Functions&Sources
+
+# Background knowledge
+
+Pulsar Functions and Sources enable the batching feature hard-coded, and also set the `batchingMaxPublishDelay` to 10ms, it only
+supports set the `batch-builder` for now, this is not suitable for all the use cases, and also not feasible for users.
+
+# Motivation
+
+Support setting batching configurations for Pulsar Functions&Sources, to make it more flexible and suitable for users.
+
+# Goals
+
+## In Scope
+
+- Support setting batching configurations for Pulsar Functions&Sources.
+
+# High Level Design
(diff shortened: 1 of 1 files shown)


pip/pip-401.md
```diff
@@ -0,0 +1,93 @@
...
+```shell
+./bin/pulsar-admin functions create \
+    --tenant public \
+    --namespace default \
+    --name test-java \
+    --className org.apache.pulsar.functions.api.examples.ExclamationFunction \
+    --inputs persistent://public/default/test-java-input \
+    --producer-config '{"batchingConfig": {"enabled": true, "batchingMaxPublishDelayMs": 100, "roundRobinRouterBatchingPartitionSwitchFrequency": 10, "batchingMaxMessages": 1000}}' \
+    --jar /pulsar/examples/api-examples.jar
+```
```
lhotari, line 58: Regarding the examples, Please share an example also for creating a `source`. Although it seems repetitive, it's useful to have concrete examples (https://lists.apache.org/thread/5cg98tpstjkp66t5zf7mtw1gdr0h1gj6). It would also be helpful to have Integration tests in the implementation to validate that the configuration also works end-to-end.

Besides sharing a command line example, it would be useful to also share a yaml config file example. In cases where a function or source has more complex configuration, it's usually better to use a yaml config file that is passed with with `--function-config-file` or `--source-config-file` command line parameter.

jiangpengcheng: added more examples

and I will add integration tests which implement this PIP

lhotari (approved):
LGTM

PR #23798, opened 2024-12-31
[fix][client] Fix compatibility between kerberos and tls

nodece:
### Motivation

When the pulsar-admin CLI uses the kerberos authentication with tls transport, we encounter the following issue:

```
$ bin/pulsar-admin tenants list
2024-12-31T15:35:56,308+0800 [main] INFO  org.apache.pulsar.client.impl.auth.AuthenticationSasl - JAAS loginContext is: PulsarClient.
2024-12-31T15:35:56,455+0800 [main] INFO  org.apache.pulsar.common.sasl.JAASCredentialsContainer - successfully logged in.
2024-12-31T15:35:56,459+0800 [pulsar-tgt-refresh-thread] INFO  org.apache.pulsar.common.sasl.TGTRefreshThread - TGT refresh thread started.
2024-12-31T15:35:56,463+0800 [pulsar-tgt-refresh-thread] INFO  org.apache.pulsar.common.sasl.TGTRefreshThread - Client principal is "client/localhost@PULSAR.COM".
2024-12-31T15:35:56,463+0800 [pulsar-tgt-refresh-thread] INFO  org.apache.pulsar.common.sasl.TGTRefreshThread - Server principal is "krbtgt/PULSAR.COM@PULSAR.COM".
2024-12-31T15:35:56,483+0800 [pulsar-tgt-refresh-thread] INFO  org.apache.pulsar.common.sasl.TGTRefreshThread - TGT valid starting at:        Tue Dec 31 15:35:56 CST 2024
2024-12-31T15:35:56,484+0800 [pulsar-tgt-refresh-thread] INFO  org.apache.pulsar.common.sasl.TGTRefreshThread - TGT expires:                  Wed Jan 01 15:35:56 CST 2025
2024-12-31T15:35:56,484+0800 [pulsar-tgt-refresh-thread] INFO  org.apache.pulsar.common.sasl.TGTRefreshThread - TGT refresh sleeping until: Wed Jan 01 11:38:40 CST 2025
class org.apache.pulsar.client.api.PulsarClientException$UnsupportedAuthenticationException: Method not implemented!
```

`Authentication` providers two methods to get the authentication data:

- `org.apache.pulsar.client.api.Authentication#getAuthData(java.lang.String)`
   - For compatibility, which will call `org.apache.pulsar.client.api.Authentication#getAuthData()`
- `org.apache.pulsar.client.api.Authentication#getAuthData()`

However, the client/admin/broker/proxy still calls the `org.apache.pulsar.client.api.Authentication#getAuthData()`, this breaks the kerberos authentication.

### Modifications

- Use `org.apache.pulsar.client.impl.auth.AuthenticationSasl#getAuthData(java.lang.String)` instead of `org.apache.pulsar.client.api.Authentication#getAuthData()`.
- Make `org.apache.pulsar.client.api.Authentication#getAuthData()` as deprecated.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarChannelInitializer.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarChannelInitializer.java
@@ -71,7 +71,7 @@ public class PulsarChannelInitializer extends ChannelInitializer<SocketChannel> 
         this.socks5ProxyAddress = conf.getSocks5ProxyAddress();
         this.socks5ProxyUsername = conf.getSocks5ProxyUsername();
         this.socks5ProxyPassword = conf.getSocks5ProxyPassword();
-        this.conf = conf;
+        this.conf = conf.clone();
         if (tlsEnabled) {
             this.pulsarSslFactoryMap = new ConcurrentHashMap<>();
             if (scheduledExecutorService != null && conf.getAutoCertRefreshSeconds() > 0) {
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarChannelInitializer.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarChannelInitializer.java
@@ -27,8 +27,10 @@ import io.netty.handler.flush.FlushConsolidationHandler;
 import io.netty.handler.proxy.Socks5ProxyHandler;
 import io.netty.handler.ssl.SslHandler;
 import java.net.InetSocketAddress;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
@@ -55,8 +57,8 @@ public class PulsarChannelInitializer extends ChannelInitializer<SocketChannel> 
     private final InetSocketAddress socks5ProxyAddress;
     private final String socks5ProxyUsername;
     private final String socks5ProxyPassword;
-
-    private final PulsarSslFactory pulsarSslFactory;
+    private final ClientConfigurationData conf;
+    private final Map<String, PulsarSslFactory> pulsarSslFactoryMap;
 
     private static final long TLS_CERTIFICATE_CACHE_MILLIS = TimeUnit.MINUTES.toMillis(1);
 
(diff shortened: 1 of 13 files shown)


liudezhi2098 (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarChannelInitializer.java
```diff
@@ -69,26 +71,17 @@ public PulsarChannelInitializer(ClientConfigurationData conf, Supplier<ClientCnx
         this.socks5ProxyAddress = conf.getSocks5ProxyAddress();
         this.socks5ProxyUsername = conf.getSocks5ProxyUsername();
         this.socks5ProxyPassword = conf.getSocks5ProxyPassword();
-
+        this.conf = conf;
```
liudezhi2098, line 74: fix spotbugs

nodece: @liudezhi2098 Done.

dao-jun (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarChannelInitializer.java
```diff
@@ -123,6 +116,23 @@ CompletableFuture<Channel> initTls(Channel ch, InetSocketAddress sniHost) {
...
+                        return factory;
+                    } catch (Exception e) {
+                        log.error("Unable to initialize and create the ssl context", e);
+                        initTlsFuture.completeExceptionally(e);
+                        return null;
+                    }
+                });
+                if (pulsarSslFactory == null) {
+                    return;
+                }
```
lhotari, line 135: @nodece What is the reason for this change? The PulsarSslFactory shouldn't be holding any state related to the hostname. Is this really needed?

lhotari: This change wasn't listed in the PR changes.

lhotari: I realize it now. This is a shortcoming of the authentication interfaces. I think this should be fixed in the interfaces so that PulsarSslFactory could be shared while also supporting this use case. I have some ideas in mind to address this. I'm planning to add a new interface which extends org.apache.pulsar.client.api.Authentication to address this. This per-host solution could be a fallback if the org.apache.pulsar.client.api.Authentication implementation doesn't support the new interface to support this use case. org.apache.pulsar.client.impl.auth.AuthenticationTls and org.apache.pulsar.client.impl.auth.AuthenticationSasl would implement this interface.
This change could also impact the PIP-337 interfaces (org.apache.pulsar.common.util.PulsarSslFactory related). Currently it doesn't make much sense that a new PulsarSslFactory instance is created for each host. As long as the client's key material doesn't change, it should be possible to use the same instance. This is something that the extended interface could handle so that it returns a PulsarSslFactory directly and caches the instance in the AuthenticationTls instance.

PR #22705, opened 2024-05-13
[improve] [broker] [break change] Do not create partitioned DLQ/Retry topic automatically

poorbarcode:
Discussion: https://lists.apache.org/thread/t756oy8881vqf9fz2zzgk3y04zsyykbh

### Motivation

After you set `defaultNumPartitions` to `16`, you will get `16` partitions per topic, and `16*16` DLQ partitions per subscription, you will get a huge number of DLQs if you have more than one subscription under one topic.  For example: 
- create topic `t1` with `16` partitions
- you will get `t1-partition-0`, `t1-partition-1`...`tp-partition-15`.
- after you enable DLQ, you will get the following `16*16` DLQs per subscription
  - `t1-partition-0-{subscription}-DLQ-partition-0`
  - `t1-partition-0-{subscription}-DLQ-partition-1`
  - ...
  - `t1-partition-15-{subscription}-DLQ-partition-15`
  - `t1-partition-1-{subscription}-DLQ-partition-0`
  - `t1-partition-1-{subscription}-DLQ-partition-1`
  - ...
  - ...
  - ...
  - `t1-partition-15-{subscription}-partition-15`

### Issue both partitioned DLQ and non-partitioned DLQ were created automatically.

- Background
  - Broker settings: `{allowAutoTopicCreation=true, defaultNumPartitions=1, allowAutoTopicCreationType=partitioned}`
  - client version `< 3.1`
- Flows of creating DLQ.
  - Trigger the DLQ creation.
  - Get partitioned metadata of DLQ: `partitions: 1`.
    - Broker created the partitioned metadata at this step. 
  - Before `3.1`, the client mistakenly assumed `{tenant}/{namespace}/{topic}-partition-0-{subscription}-DLQ-partition-0` is `{tenant}/{namespace}/{topic}-partition-0-{subscription}-DLQ` due to the bug that was fixed by https://github.com/apache/pulsar/pull/19841.
  - Start a dead-letter producer under `{tenant}/{namespace}/{topic}-partition-0-{subscription}-DLQ`
    - Triggered a non-partitioned topic creation.

Eventually, you will get both partitioned and non-partitioned DLQ.
```
> bin/pulsar-admin topics list-partitioned-topics {tenant}/{namespace}
persistent://{tenant}/{namespace}/{topic}-partition-0-{subscription}-DLQ

> bin/pulsar-admin topics list {tenant}/{namespace}                  
persistent://{tenant}/{namespace}/{topic}-partition-0-{subscription}-DLQ
``` 

[...]

PR #23806, opened 2025-01-03
[improve][broker] Optimization only automatically creates non-partitioned retry-topic and DLQ-topic.

zjxxzjwang:
### Motivation

when the code logic automatically creates a try-topic or DLQ-topic, it only creates non-partitioned retry-topic and non-partitioned DLQ-topic, regardless of the value of the “allowAutoTopicCreationType” configuration.

### Modifications

Depending on the value of the “allowAutoTopicCreationType” configuration item,  control the type of auto-creation of a try-topic or DLQ-topic.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -198,9 +198,6 @@ allowAutoTopicCreation=true
 # The type of topic that is allowed to be automatically created.(partitioned/non-partitioned)
 allowAutoTopicCreationType=non-partitioned
 
-# The type of retry-topic or DLQ-topic that is allowed to be automatically created.(partitioned/non-partitioned)
-allowAutoRetryOrDLQTopicCreationType=non-partitioned
-
 # If 'allowAutoTopicCreation' is true and the name of the topic contains 'cluster',
 # the topic cannot be automatically created.
 allowAutoTopicCreationWithLegacyNamingScheme=true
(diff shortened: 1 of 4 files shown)


lhotari:
@zjxxzjwang Please rename this PR. The current title is confusing. 

Hint: using ChatGPT or some other LLM could help provide better names for PRs. LLMs are also useful in fixing grammar mistakes.

One way to pass the PR details to LLM is to append `.patch` to the url, for example https://github.com/apache/pulsar/pull/23806.patch and then store this in a file that is passed to an LLM. Also store the output of https://api.github.com/repos/apache/pulsar/pulls/23806 and pass that in a file. 
In the prompt "please provide a better title and description for this PR with fixed grammar and improved clarity. return the result in downloadable markdown format.".

Here's an example from Claude:
image

lhotari (changes requested):
I believe that it's been an intentional change to create non-partitioned retry-topics and DLQ-topics. This change was made in PR #22705.
I guess we would need a separate configuration or namespace level policy for controlling the behavior. This type of decisions are made on the [dev@pulsar.apache.org mailing list](https://pulsar.apache.org/contact/#mailing-lists) and they are usually documented as PIPs.

PR #23811, opened 2025-01-04
[fix][fn] Record Pulsar Function processing time properly for asynchronous functions

### LINKED ISSUE #23705: [Bug] Pulsar Function processing time doesn't get properly recorded for asynchronous functions

walkinggo:
Fixes #23705 

### Motivation

See #23705 
In the realm of asynchronous processing, precise timing and performance metrics are essential for effective monitoring and optimization. Apache Pulsar Functions, as a distributed compute platform, relies heavily on asynchronous operations to process and transform data streams. However, the current implementation lacks a robust mechanism for accurately capturing and reporting the execution times of these asynchronous functions.

### Modifications
This Pull Request introduces several modifications to the Apache Pulsar Functions project aimed at improving the accuracy and reliability of asynchronous function execution time statistics. The primary goal is to enhance monitoring and analysis capabilities for function performance. Here’s a breakdown of the key changes:

1. Removal of processTimeStart Method: The processTimeStart method in ComponentStatsManager has been removed. This method was previously used to record the start time of asynchronous function execution. The start time is now recorded in the JavaExecutionResult object, providing a more centralized and consistent approach.

2. Modification of JavaInstanceRunnable: The run method in JavaInstanceRunnable has been updated to reflect the removal of processTimeStart. The stats.processTimeStart() call has been deleted, and the stats.processTimeEnd() method now accepts the start time as a parameter to calculate the total execution time.

3. Update to FunctionStatsManager: The processTimeEnd method in FunctionStatsManager has been modified to remove the processTimeStart member variable and accept the start time as a parameter. This allows for accurate calculation of the function’s execution time.

4. Changes to SinkStatsManager and SourceStatsManager: The processTimeEnd methods in SinkStatsManager and SourceStatsManager have been updated to accept the start time as a parameter. However, since these classes do not record processing time, the method bodies remain empty.

5. Addition of startTime in JavaExecutionResult: The JavaExecutionResult class now includes a startTime member variable to store the start time of asynchronous function execution. This allows for accurate calculation of execution time within the handleResult method of JavaInstanceRunnable.

6. Modification of AsyncFuncRequest: The AsyncFuncRequest class in JavaInstance now includes a result member variable of type JavaExecutionResult. This change ensures that the processAsyncResultsInInputOrder method uses the existing JavaExecutionResult object instead of creating a new one, maintaining consistency and avoiding duplication.

7. Use of Same ExecutionResult in Non-asyncPreserveInputOrder Mode: In scenarios where asyncPreserveInputOrderForOutputMessages is disabled, the same executionResult object is now used to avoid unnecessary object creation and potential issues with result assignment.

8. Fix for Result and Exception Handling: Two patches address potential bugs related to result and exception handling in JavaInstance. The processAsyncResultsInInputOrder method now ensures that the result and userException fields of JavaExecutionResult are properly set, improving the reliability of the execution result.

9. Addition of Test Case: A new test case, testAsyncFunctionTime, has been added to verify the accuracy of asynchronous function execution time recording and calculation. This test ensures that the start time recorded in JavaExecutionResult is within an acceptable range of the actual start time.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/walkinggo/pulsar/pull/7


pulsar-client/src/test/java/org/apache/pulsar/client/examples/TransactionExample.java
lhotari, line 1: This change doesn't seem to be related to the PR.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstanceRunnable.java
```diff
@@ -329,26 +334,29 @@ public void run() {
...
+                            asyncResultConsumer,
+                            asyncErrorHandler);
+                    Thread.currentThread().setContextClassLoader(instanceClassLoader);
+                    instanceObserver.setJavaExecutionResult(result);
+                    return instanceObserver;
+                }).whenComplete((res, ex) -> {
+                    stats.processTimeEnd(res.getStartTime());
+                });
 
+                JavaExecutionResult result = future.join().getJavaExecutionResult();
```
lhotari, line 359: This doesn't make sense and is completely unnecessary. 
Instead, you should simply pass the starting time nanos as an argument that gets added to the `JavaExecutionResult` instance field (add a new field `startTimeNanos` to JavaExecutionResult).
The result can be recorded in the `handleResult` method. The `processTimeStart` method should be completely removed from the `ComponentStatsManager` class and the `processTimeEnd` method should accept the starting time as a parameter. The value gets taken from the `startTimeNanos` field of the `JavaExecutionResult` instance.

lhotari (changes requested):

walkinggo:
It looks like due to code formatting, many unintended changes were introduced. I added `startTimeNanos` in `JavaExecutionResult` and removed it from `ComponentStatsManager`. Now, the `processTimeEnd` method in `ComponentStatsManager` takes the start time as a parameter and calculates the final time. I also removed the `processTimeStart` method.

walkinggo:
Do I need to create a new PR to remove the unnecessary line changes caused by formatting? @ @lhotari

pulsar-client/src/test/java/org/apache/pulsar/client/examples/TransactionExample.java
walkinggo: Sorry, uploading these file changes was a mistake. I've force-pushed in a later commit to remove those changes.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstanceRunnable.java
walkinggo: Thank you, I have completed these tasks.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/stats/ComponentStatsManager.java
```diff
@@ -100,9 +102,7 @@ public ComponentStatsManager(FunctionCollectorRegistry collectorRegistry,
 
     public abstract void setLastInvocation(long ts);
 
-    public abstract void processTimeStart();
-
-    public abstract void processTimeEnd();
+    public abstract void processTimeEnd(double processTimeStart);
```
lhotari, line 105: the parameter should be a `long` instead of `double`.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaExecutionResult.java
```diff
@@ -29,6 +29,7 @@
 public class JavaExecutionResult {
     private Throwable userException;
     private Object result;
+    private double startTimeNanos;
```
lhotari, line 32:
```suggestion
    private final long startTimeNanos = System.nanoTime();
```
It should be a `long` field and it should be made final.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
```diff
@@ -117,6 +117,7 @@ public JavaExecutionResult handleMessage(Record<?> record, Object input,
         }
 
         JavaExecutionResult executionResult = new JavaExecutionResult();
+        executionResult.setStartTimeNanos(System.nanoTime());
```
lhotari, line 120: This is obsolete since startTimeNanos is set when the instance gets constructed.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstanceRunnable.java
```diff
@@ -346,10 +348,10 @@ public void run() {
                         asyncErrorHandler);
                 Thread.currentThread().setContextClassLoader(instanceClassLoader);
 
-                // register end time
-                stats.processTimeEnd();
 
                 if (result != null) {
+                    // register end time
+                    stats.processTimeEnd(result.getStartTimeNanos());
```
lhotari, line 354: Remove this. This should be handled in the handleResult method instead.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
lhotari: You will also need to pass the `JavaExecutionResult` instance into a field `javaExecutionResult` of the `AsyncFuncRequest` so that the same instance is being used for the `asyncPreserveInputOrderForOutputMessages` case on line 139. Replace the instantiation of `JavaExecutionResult` in `processAsyncResultsInInputOrder` method on line 191 with this instance that is carried in the `javaExecutionResult` of the `AsyncFuncRequest` instance.
Also replace the instance on line 152 with this same `executionResult` on line 152 so that the starting time will be handled for the non `asyncPreserveInputOrderForOutputMessages` case of async processing.
With these changes, the solution would have the basics in place. There would also need to be a way to test all of this. 

For testing, it might be useful to have a `java.util.function.LongSupplier` function for getting the nanoTime. In production code, this would default to `() -> System.nanoTime()`, but in tests, this could be used to mock the "clock" for getting nanoTime. Adding testability would cause some additional refactorings.

lhotari (changes requested):
Good start, going in the right direction!

lhotari:
> Do I need to create a new PR to remove the unnecessary line changes caused by formatting? @ @lhotari

It makes it slightly harder to read the diff when there are formatting changes. However there's the "hide whitespace" option to address that in GitHub UI.
Are you using IntelliJ/IDEA for development? Have you setup the code style according to the instructions at https://pulsar.apache.org/contribute/setup-ide/#configure-code-style ? I noticed that the import formatting doesn't match the Pulsar formatting since there are no empty lines in the Pulsar code style.
In the past, I have noticed that some code in the Pulsar Functions packages is using some non-default formatting so if you reformat the code, it will result in many changes. As long as you are using the Pulsar code style, the formatting changes are fine to be included in this PR.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstance.java
walkinggo: I have completed these tasks, please check them.

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
```diff
@@ -309,4 +314,22 @@ public void testAsyncFunctionMaxPendingVoidResult() throws Exception {
...
+                    Thread.sleep(500);
+                    return input;
+                },
+                new InstanceConfig());
+        String testString = "ABC123";
+        JavaExecutionResult result = instance.handleMessage(mock(Record.class), testString);
+        LongSupplier timeSupplier = () -> System.nanoTime() - 500_000_000L;
+        assertNotNull(result.getResult());
+        long beforeTime = timeSupplier.getAsLong();
+        assertTrue(Math.abs(beforeTime - result.getStartTime()) <= 20_000_000);
```
lhotari, line 332: the usage of `LongSupplier timeSupplier` doesn't make much sense in this way. just get rid of it for now. What I was trying to explain earlier was that a common pattern for testing time is to have a mock clock that gets injected also in the production code. Instead of calling `System.nanoTime()` in production code, the "clock" interface (`LongSupplier` for example) would be used. Let's forget that for now.

lhotari (changes requested):

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
```diff
@@ -24,12 +24,16 @@
...
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.function.LongSupplier;
```
lhotari, line 36: These seem to be unused imports

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
```diff
@@ -309,4 +317,25 @@ public void testAsyncFunctionMaxPendingVoidResult() throws Exception {
...
+                        e.printStackTrace();
+                    }
+                    return input;
+                },
+                new InstanceConfig());
+        String testString = "ABC123";
+        JavaExecutionResult result = instance.handleMessage(mock(Record.class), testString);
+        assertNotNull(result.getResult());
+        long beforeTime = System.nanoTime() - 500_000_000L;
+        assertTrue(Math.abs(beforeTime - result.getStartTime()) <= 20_000_000);
```
lhotari, line 338: What is the logic behind this assertion?

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
```diff
@@ -309,4 +317,25 @@ public void testAsyncFunctionMaxPendingVoidResult() throws Exception {
         log.info("start:{} end:{} during:{}", startTime, endTime, endTime - startTime);
         instance.close();
     }
+
+    @Test
+    public void testAsyncFunctionTime() {
```
lhotari, line 322: This doesn't seem to be testing an async function since it's a `Function<String, String>` that is the function.

lhotari (changes requested):

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
walkinggo: Sorry, the import statements here were left over from my coding, and I should have removed them.

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
walkinggo: The logic here is to give a task that takes 500ms, and then compare the difference between the task's completion time and the recorded time. If the difference is less than 20ms, we can consider the recorded time as accurate.

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
walkinggo: Sorry, I should have used the `CompletableFuture` class for testing. Is that acceptable?

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
lhotari: It seems pointless to test that at all in this unit test class. It would essentially test that `result.getStartTime()` was set before the execution started and that execution took more than 500ms. Tests should be about testing expected behavior. 

In this PR, we are more interested that the processing time gets properly recorded. That would have to be tested at a different level since JavaInstanceTest is at the level of the JavaInstance class. The processing time handling is in JavaInstanceRunnable. I think that tests for this PR should reside in JavaInstanceRunnable and possibly also in one of the end-to-end tests (for example `pulsar-broker/src/test/java/org/apache/pulsar/io/PulsarFunctionE2ETest.java`).

btw. One tip about assertions. For newly added tests, I'd recommend using [AssertJ fluent assertions](https://assertj.github.io/doc/) for more complex assertions. We have AssertJ already available as a test dependency in Pulsar. It's not that I'm saying that AssertJ must be used, it's just that if there are more complex assertions, it's a better option than TestNG's assertions. AssertJ assertions are easier to read and when there are failures, the error messages are really great.

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
lhotari: This test can be removed completely. More context in https://github.com/apache/pulsar/pull/23811#discussion_r1906727569 . Tests for this PR should go in different classes.

walkinggo: Alright, I will remove this test case. Should I add additional test cases in the other test files you mentioned?

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
walkinggo: How should I run `PulsarFunctionE2ETest.java` located in `pulsar-broker/src/test/java/org/apache/pulsar/io`? I tried running it locally in IDEA, but encountered an error. Is there any related documentation?

lhotari: > How should I run `PulsarFunctionE2ETest.java` located in `pulsar-broker/src/test/java/org/apache/pulsar/io`? I tried running it locally in IDEA, but encountered an error. Is there any related documentation?

Most likely you will first have to compile Pulsar on the command line with maven.
either

build core-modules:


```shell
 mvn -Pcore-modules,-main -T 1C clean install -DskipTests -Dspotbugs.skip=true
```

or build everything, skipping some checks to speed it up:

```shell
mvn -T 1C clean install -DskipTests -Dspotbugs.skip=true -Dcheckstyle.skip=true -Dlicense.skip=true
```

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
lhotari: yes

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceTest.java
walkinggo: Is this test program only runnable on Linux or macOS? I ran it on Windows using the command you provided, but encountered an error indicating that I don't have permission to delete certain files and that the path characters are incorrect.

lhotari: > Is this test program only runnable on Linux or macOS? I ran it on Windows using the command you provided, but encountered an error indicating that I don't have permission to delete certain files and that the path characters are incorrect.

@walkinggo I'd recommend using [Windows Subsystem for Linux (WSL) on Windows](https://learn.microsoft.com/en-us/windows/wsl/install) with Ubuntu for developing Pulsar. We don't support Windows for Pulsar broker at all. It might work, but the support is only for Linux and macOS. For the Pulsar client, we also support Windows.

lhotari: IntelliJ has also special support for WSL on Windows: https://www.jetbrains.com/help/idea/how-to-use-wsl-development-environment-in-product.html

walkinggo: Thank you for your patience. I have successfully run the program on Ubuntu. @lhotari

walkinggo:
Should I modify the related code logic based on the following error that appeared in my local CI check? @lhotari 

[ERROR] Medium: org.apache.pulsar.functions.instance.JavaInstance$AsyncFuncRequest.getResult() may expose internal representation by returning JavaInstance$AsyncFuncRequest.result [org.apache.pulsar.functions.instance.JavaInstance$AsyncFuncRequest] At JavaInstance.java:[line 50] EI_EXPOSE_REP

walkinggo:
I pass the start time of the result as a parameter rather than making it an inherent attribute of the result. This is because I don't want the result to become an inherent variable of `AsyncFuncRequest`. If it were an inherent variable, the result would be exposed, triggering the EI_EXPOSE_REP warning.

lhotari:
> I pass the start time of the result as a parameter rather than making it an inherent attribute of the result. This is because I don't want the result to become an inherent variable of `AsyncFuncRequest`. If it were an inherent variable, the result would be exposed, triggering the EI_EXPOSE_REP warning.

That makes the code worse in my opinion. I'd recommend adding a definition to the `src/main/resources/findbugsExclude.xml` file to ignore the `EI_EXPOSE_REP` SpotBugs/FindBugs check for the particular method. We do  that already for many other locations. This is the file where you can add a new entry: https://github.com/apache/pulsar/blob/master/pulsar-functions/instance/src/main/resources/findbugsExclude.xml .

```xml
  <Match>
    <Class name="org.apache.pulsar.functions.instance.JavaInstance$AsyncFuncRequest"/>
    <Method name="getResult"/>
    <Bug pattern="EI_EXPOSE_REP"/>
  </Match>
```

walkinggo:
I have added the content to the corresponding file. Thank you for your help.

pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/instance/JavaInstanceRunnableTest.java
```diff
@@ -173,6 +178,20 @@ public Void process(String input, Context context) throws Exception {
...
+        JavaInstanceRunnable runnable =
+                createRunnable(FunctionDetails.newBuilder().setClassName(TestFunction.class.getName()).build());
+        Instant before = Instant.now();
+        Instant join = CompletableFuture.supplyAsync(() -> {
+            new Thread(runnable).start();
+            return Instant.now();
+        }).join();
+        assertThat(join).isNotNull();
+        Duration duration = Duration.between(before, join);
+        assertThat(duration.toMillis()).isLessThan(20);
```
lhotari, line 192: this assertion seems pointless. It would be more meaningful to validate calls to `processTimeEnd`.

pulsar-broker/src/test/java/org/apache/pulsar/io/PulsarFunctionE2ETest.java
```diff
@@ -296,6 +305,92 @@ public void testReadCompactedFunction() throws Exception {
...
 
+    @Test(timeOut = 20000)
+    public void testPulsarFunctionAsyncStatTime() throws Exception {
+        final String namespacePortion = "io";
+        final String replNamespace = tenant + "/" + namespacePortion;
+        final String sourceTopic = "persistent://" + replNamespace + "/my-topic1";
+        final String sinkTopic = "persistent://" + replNamespace + "/output";
+        final String propertyKey = "key";
+        final String propertyValue = "value";
+        final String functionName = SpendTimeFunction.class.getName();
```
nodece, line 316:
```suggestion
        final String functionName = "org.apache.pulsar.functions.api.examples.JavaNativeAsyncExclamationFunction";
```

pulsar-broker/src/test/java/org/apache/pulsar/io/PulsarFunctionE2ETest.java
```diff
@@ -296,6 +305,92 @@ public void testReadCompactedFunction() throws Exception {
...
     }
 
+    @Test(timeOut = 20000)
+    public void testPulsarFunctionAsyncStatTime() throws Exception {
+        final String namespacePortion = "io";
+        final String replNamespace = tenant + "/" + namespacePortion;
+        final String sourceTopic = "persistent://" + replNamespace + "/my-topic1";
+        final String sinkTopic = "persistent://" + replNamespace + "/output";
+        final String propertyKey = "key";
+        final String propertyValue = "value";
```
nodece, line 315:
```suggestion
```

pulsar-broker/src/test/java/org/apache/pulsar/io/PulsarFunctionE2ETest.java
```diff
@@ -1204,4 +1307,12 @@ public ByteBuffer process(ByteBuffer input, Context context) throws Exception {
             return input;
         }
     }
+
+    public static class SpendTimeFunction implements org.apache.pulsar.functions.api.Function<String, String> {
```
nodece, line 1311: `org.apache.pulsar.functions.api.examples.JavaNativeAsyncExclamationFunction` can be instead of this function, please remove this class.

BTW, this is a sync function, not async.

pulsar-broker/src/test/java/org/apache/pulsar/io/PulsarFunctionE2ETest.java
```diff
@@ -93,6 +95,8 @@ protected static FunctionConfig createFunctionConfig(String tenant, String names
         functionConfig.setAutoAck(true);
         if (!isBuiltin) {
             functionConfig.setClassName("org.apache.pulsar.functions.api.examples.ExclamationFunction");
+        } else {
+            functionConfig.setClassName(functionName);
```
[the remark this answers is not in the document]
walkinggo: I added this line to modify the internal function name so that I could use the `SpendTimeFunction` method in the test cases. However, I plan to remove this method and use `org.apache.pulsar.functions.api.examples.JavaNativeAsyncExclamationFunction` instead. Therefore, I will revert this modification.

nodece (approved):
LGTM

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaExecutionResult.java
```diff
@@ -29,6 +29,7 @@
 public class JavaExecutionResult {
     private Throwable userException;
     private Object result;
+    private final long startTime = System.nanoTime();
 
     public void reset() {
```
lhotari, line 34: I think it would be useful to remove this unused `reset` method from this class. The new final `startTime` field makes this method inconsistent as well and the resolution is to simply remove this method.

lhotari (changes requested):
Good work @walkinggo. This looks very close to completion. I added a comment about removing the unused org.apache.pulsar.functions.instance.JavaExecutionResult#reset method. I think it makes sense removing it in this PR since we added a final field to the class in this PR. Since the method is unused, it's better to simply remove it.

pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaExecutionResult.java
walkinggo: Thank you for your help and patience. I really appreciate the time and effort you've spent on my PR. I have removed the method. Could you please review it?

lhotari (approved):

PR #23814, opened 2025-01-05
[Fix][client]Fix status html endpoint

### LINKED ISSUE #23809: [Bug] /status.html endpoint fails when statusFilePath is not set
### Search before asking

### Read release policy

### Version

any released version

### Minimal reproduce step

1. set statusFilePath to empty in `standalone.conf`
2. start pulsar standalone with `bin/pulsar standalone`
3. go to http://localhost:8080/status.html (or use `curl http://localhost:8080/status.html`)

that /status.html endpoint returns "OK" when `statusFilePath` is not set and the broker is ready.

### What did you see instead?

The /status.html will always return 404 failure

### Anything else?

There's also invalid default configuration in `standalone.conf`:
https://github.com/apache/pulsar/blob/1f7a79f4c240c8075c217ac417dffa5bc72a4d5f/conf/standalone.conf#L215
This value doesn't make any sense. The default value should be empty (which results in `null`).

yunmaoQu:
close #23809 
### Modifications
To address these issues, the following changes have been made:
1. **Dedicated Servlet Handler**: A new servlet handler has been added specifically for the `/status.html` endpoint. This handler directly checks the broker's readiness state.
2. **Default `statusFilePath` Update**: The default value for `statusFilePath` in `standalone.conf` has been changed to an empty value. This prevents the use of an invalid default path.
3. **Independence from `statusFilePath`**: The status endpoint has been modified to operate independently of the `statusFilePath` configuration. This means its functionality is no longer dependent on this setting.

### Documentation
- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository


lhotari (changes requested):
This PR doesn't make sense at the moment. I'd prefer that @yyj8 handles the issue as I have suggested in comment https://github.com/apache/pulsar/issues/23809#issuecomment-2569629329 instead of explaining the context to another contributor.

PR #23829, opened 2025-01-09
[improve][fn] Add API endpoint for function-worker for liveness check with configurable flag

mukesh154:
### Motivation

This pull request introduces a health check functionality for Kubernetes deployments, specifically adding a liveness probe for the function worker. The liveness probe is crucial for Kubernetes-based applications, enabling automated pod restarts in case of failure. This change ensures that the function worker recovers when a `ProducerFencedException` occurs, which causes the worker to get stuck and not recover.

For instance, when a client makes a request like:

```
curl --location --request PUT 'https://localhost:6651/admin/v3/functions/test/test/test' --header 'Authorization: Bearer <token>' --header '...' --form 'data=@functions-test-1.0.0.jar'
```

For `POST`, `PUT`, and `DELETE` operations, the following error is returned under heavy load:
```json
{"reason":"Internal Error updating function at the leader"}
```

And, when the following error occurs in the function worker currently:

```
ERROR org.apache.pulsar.functions.worker.FunctionMetaDataManager - Could not write into Function Metadata topic │
│ org.apache.pulsar.client.api.PulsarClientException$ProducerFencedException: Producer was fenced
```

The function worker does not recover, leading to an ongoing failure. With this update, the worker will automatically restart with the help of health check with liveliness probe upon encountering this error, ensuring proper recovery and continuity of operations.

### Modifications

This update introduces an API endpoint to perform a liveliness check on the function worker pod. The API returns an HTTP status of `200 (OK)` when the `isLive` flag within `FunctionImpl` is true. If the flag is false, typically after a `ProducerFencedException` occurs, the API will return a status of `503 (Service Unavailable)`.

The Kubernetes deployment configuration has been updated to use this new API endpoint in the liveness probe along with existing `metrics` endpoint, allowing the system to monitor the health and availability of the function worker.

### Verifying this change

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API: (no)
  - The schema: (no)
  - The default values of configurations: (no)
  - The wire protocol: (no)
  - The rest endpoints: (no)
  - The admin cli options: (no)
  - Anything that affects deployment: (no)

### Documentation

Check the box below or label this PR directly (if you have committer privilege).

Need to update docs? 

- [ ] `doc-required` 
- [X] `no-need-doc` 
- [ ] `doc`

### DIFF
--- a/pulsar-functions/worker/src/main/java/org/apache/pulsar/functions/worker/FunctionMetaDataManager.java
+++ b/pulsar-functions/worker/src/main/java/org/apache/pulsar/functions/worker/FunctionMetaDataManager.java
@@ -81,6 +81,8 @@ public class FunctionMetaDataManager implements AutoCloseable {
     @Getter
     private CompletableFuture<Void> isInitialized = new CompletableFuture<>();
 
+    private boolean isFunctionWorkerAlive = true;
+
     public FunctionMetaDataManager(WorkerConfig workerConfig,
                                    SchedulerManager schedulerManager,
                                    PulsarClient pulsarClient,
@@ -243,6 +245,10 @@ public class FunctionMetaDataManager implements AutoCloseable {
                 needsScheduling = processUpdate(functionMetaData);
             }
         } catch (Exception e) {
+            if (e.getCause() instanceof PulsarClientException.ProducerFencedException) {
+                log.error("Function worker status has been set to false due to ProducerFencedException.");
+                this.isFunctionWorkerAlive = false;
+            }
             log.error("Could not write into Function Metadata topic", e);
             throw new IllegalStateException("Internal Error updating function at the leader", e);
         }
@@ -499,5 +505,9 @@ public class FunctionMetaDataManager implements AutoCloseable {
                 pulsarClient.newReader(), this.workerConfig, lastMessageSeen, this.errorNotifier);
         this.functionMetaDataTopicTailer.start();
         log.info("MetaData Manager Tailer started");
+    }
+
+    public boolean checkLiveliness() {
+        return this.isFunctionWorkerAlive;
     }
 }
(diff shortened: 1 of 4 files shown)


lhotari:
> The function worker does not recover, leading to an ongoing failure. With this update, the worker will automatically restart with the help of health check with liveliness probe upon encountering this error, ensuring proper recovery and continuity of operations.

Although adding a liveness check could be useful for many reasons, it would be better to primarily address this issue. Are you able to isolate and reproduce the issue? What Pulsar version are you running?

It looks like the leader election for the function worker uses a consumer with `consumerEventListener` to find who is the leader:
https://github.com/apache/pulsar/blob/3396065a3a62af3e3586700f0bbfcbff93716b48/pulsar-functions/worker/src/main/java/org/apache/pulsar/functions/worker/LeaderService.java#L76-L87

There might be a bug in the Pulsar implementation of notifying the active consumer change:
https://github.com/apache/pulsar/blob/49aa3080d422994baa036ed0b743a2fa18a6d530/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java#L213-L223

At least by reading the code, it's hard to see how that could work.

lhotari:
> At least by reading the code, it's hard to see how that could work.

Now I can see it. There are multiple references to `notifyActiveConsumerChanged` in other code locations. It seems that the solution works, but the code is just hard to understand. Found #1818 with some explanations.

lhotari:
How many function worker instances do you have when you encounter this problem?

mukesh154:
> Although adding a liveness check could be useful for many reasons, it would be better to primarily address this issue. Are you able to isolate and reproduce the issue? What Pulsar version are you running?


Thanks for the feedback! I understand your point about addressing the primary issue first. Regarding the issue reproduction, I haven’t been able to isolate it myself yet. I'm using Pulsar 3.0, so it could potentially be related to that version.

mukesh154:
> How many function worker instances do you have when you encounter this problem?

I have 2 function worker instances when I encounter the problem.

PR #23824, opened 2025-01-08
[fix][client] Prevent retry topic and dead letter topic producer leaks when sending of message fails

lhotari:
Fixes #20635

### Motivation

The client currently leaks producers when sending of the message fails and a new producer is created after an exception occurs. The previous producer wasn't closed in these cases.
In case of such failures, there should be a retry backoff so that such cases don't cause excessive load on the broker side.

### Modifications

- properly close retry topic and dead letter topic producers when they are discarded
- add backoff for the creation of new producers when there are subsequent failures. Reset the backoff when a message is sent successfully.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23830, opened 2025-01-09
[fix][client] Client interceptors close method not called after producer/consumer/reader close

Shawyeok:
### Motivation

ClientInterceptors close methods (`ProducerInterceptor#close`, `ConsumerInterceptor#close`, `ReaderInterceptor#close`) give interceptor developer a chance to release resources allocated in the interceptor when producer/consumer/reader is closing.

However, the interceptor's close method not get called in the implementation of [PIP-23][pip-23].

### Modifications

Call interceptors close method in the final stage of closing process, add some unit tests to cover user case.

### Verifying this change

  - *Add integration test for ProducerInterceptor with non-partition and partition topic*
  - *Add integration test for ConsumerInterceptor with non-partition and partition topic*
  - *Add integration test for ReaderInterceptor with non-partition and partition topic*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -72,6 +72,7 @@ import lombok.AccessLevel;
 import lombok.Getter;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.tuple.Triple;
 import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.ConsumerCryptoFailureAction;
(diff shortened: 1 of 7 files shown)


Shawyeok:
@codelipenghui Please feel free take a look, thanks.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1142,7 +1143,14 @@ public synchronized CompletableFuture<Void> closeAsync() {
         }
 
         ArrayList<CompletableFuture<Void>> closeFutures = new ArrayList<>(4);
-        closeFutures.add(closeFuture);
+        closeFutures.add(closeFuture.whenCompleteAsync((nil, ex) -> {
```
Denovo1998, line 1146: @Shawyeok @lhotari 
In the case where the State is Closing or Closed, should interceptors.close() also be executed?

It looks like the code for `closeAsync()` has been updated after https://github.com/apache/pulsar/pull/23824, so the `whenCompleteAsync` here should be set on the `compositeCloseFuture`?

PR #23831, opened 2025-01-09
[fix][misc]: ignore deleted ledger when tear down cluster

mattisonchao:
### Motivation

The current tear-down logic is listing ledgers and then deleting them. So, there's a race condition if some internal logic has deleted the schema ledger, which will cause the down process to fail.

### Modifications

- Ignore deleted ledgers while tear-down.

### Verifying this change

### Documentation
- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
@@ -230,7 +230,8 @@ public class PulsarClusterMetadataTeardown {
                     case BKException.Code.NoSuchLedgerExistsException:
                     case BKException.Code.NoSuchLedgerExistsOnMetadataServerException:
                     case BKException.Code.NoSuchEntryException:
-                        log.warn("Failed to delete ledger.{}", ledgerId, Throwables.getRootCause(ex));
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataTeardown.java
```diff
@@ -222,9 +222,19 @@ private static void deleteLedger(BookKeeper bookKeeper, long ledgerId) {
...
             }
-        } catch (InterruptedException | BKException e) {
-            log.error("Failed to delete ledger {}: {}", ledgerId, e);
-            throw new RuntimeException(e);
+        } catch (InterruptedException | BKException ex) {
+            if (ex instanceof BKException bkException) {
+                switch (bkException.getCode()) {
+                    case BKException.Code.NoSuchLedgerExistsException:
+                    case BKException.Code.NoSuchLedgerExistsOnMetadataServerException:
+                    case BKException.Code.NoSuchEntryException:
```
dao-jun, line 230: LGTM, but why to catch `NoSuchEntryException ` ?

dao-jun (approved):

shibd (approved):

PR #23833, opened 2025-01-09
[fix][broker] Fix possible mark delete NPE when batch index ack is enabled

BewareMyPower:
### Motivation

`MLPendingAckStoreTest#testMainProcess` will fail by NPE if batch index ACK is enabled. The NPE happens when the position argument does not has a non-null `ackSet`.

### Modifications

- Add an `ackBatchPosition` method to handle `Optional` objects correctly with elegant and clear code.
- Replace `BitSetRecyclable` with `BitSet` as the value type of `batchDeletedIndexes` to simplify the unnecessarily complicated code of calling `recycle()`. For example, if `recycle()` is called in a callback of `ConcurrentSkipListMap#compute`, it could be called multiple times. Additionally, the optimization for GC is not provided because `BitSetRecyclable` cannot avoid the memory allocation of its underlying long arrays.
- Mark `batchDeletedIndexes` as `@Nullable` and check if it's not null instead of `getConfig().isDeletionAtBatchIndexLevelEnabled()` to eliminate all NPE warnings.

### Verifying this change

Enable batch index ACK in `MLPendingAckStoreTest` and guarantee this test passes.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/BitSetRecyclable.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/BitSetRecyclable.java
@@ -1211,9 +1211,6 @@ public class BitSetRecyclable implements Cloneable, java.io.Serializable {
         if (recyclerHandle != null) {
             this.clear();
             recyclerHandle.recycle(this);
-            // Reset with null so that recycle() can be called many times but only the 1st time takes effect.
-            // It's used when recycle() is called in ConcurrentSkipListMap's callback
-            recyclerHandle = null;
         }
     }
 }
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/BitSetRecyclable.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/BitSetRecyclable.java
@@ -1211,6 +1211,9 @@ public class BitSetRecyclable implements Cloneable, java.io.Serializable {
         if (recyclerHandle != null) {
             this.clear();
             recyclerHandle.recycle(this);
+            // Reset with null so that recycle() can be called many times but only the 1st time takes effect.
+            // It's used when recycle() is called in ConcurrentSkipListMap's callback
+            recyclerHandle = null;
         }
     }
 }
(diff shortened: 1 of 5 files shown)


pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/BitSetRecyclable.java
```diff
@@ -1211,6 +1211,9 @@ public void recycle() {
         if (recyclerHandle != null) {
             this.clear();
             recyclerHandle.recycle(this);
+            // Reset with null so that recycle() can be called many times but only the 1st time takes effect.
```
dao-jun, line 1214: If we set the  `recycleHandler = null` here, looks like it will downgrade to  shortlive objects

BewareMyPower: Yes. What's the issue here?

dao-jun: I am just wandering that since it can be downgrade to shortlive objects, why not just `new BitSetRecyclable()` here directly, so we don't need to call `recycle` and don't need to set `recycleHandle=null`

dao-jun: By this way, we can keep the semantic align, WDYT?

dao-jun: +1, use bitset here directly will not lead to performance regression, and makes it more clearly. The key point is `BitSet#words` field, since we just wrap the given `ackSet` by BitSet, it is OK to use `BitSet` instead.

BewareMyPower: Yes, it's hardly to see two `BitSetRecyclable` objects share the same underlying long array (`ackSet`). In this case, even if two allocations of `BitSetRecyclable` share the same slot in the recycler's local pool (`Recycler#threadLocal`), the long array is still allocated twice from JVM heap memory, which should take the majority of the memory.

----

The change is done, PTAL again @dao-jun

dao-jun (approved):
lgtm

Demogorgon314 (approved):

PR #23839, opened 2025-01-10
[improve][broker] Reduce unnecessary REPLICATED_SUBSCRIPTION_SNAPSHOT_REQUEST

shibd:
### Motivation
Let's say two clusters to two-way replication: cluster-1 and cluster-2

- cluster-1: has enabled subscription replication and correctly configured the cluster info for `cluster-2` 
- cluster-2:has either incorrectly configured the information for `cluster-1` or disabled subscription replication.

As a result, `cluster-1` will send **SNAPSHOT_REQUEST** to `cluster-2` at a frequency of `replicatedSubscriptionsSnapshotFrequencyMillis` (1 second by default). 

In extreme cases, this can lead to a large number of consecutive `marker` between `cluster-1` and `cluster-2` topic.

This may affect the performance of `reader` and `consumer`, or cause `reader.readNext()` to time out.

### Modifications
- When there is no complete snapshot, retry according to replicatedSubscriptionsSnapshotTimeoutSeconds to minimize the number of markers.

### Verifying this change
- Add `testReplicatedSubscriptionOneWay` to cover this case.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
@@ -228,9 +228,9 @@ public class ReplicatedSubscriptionsController implements AutoCloseable, Topic.P
             // 2. This approach prevents sending additional SNAPSHOT_REQUEST to both local_topic and remote_topic.
             // 3. Since it's uncertain when the remote cluster will enable subscription replication,
             //    the timeout mechanism of pendingSnapshots is used to ensure retries.
-            
-            // In other words, when hit this case, The frequency of sending SNAPSHOT_REQUEST has changed from 
-            // `replicatedSubscriptionsSnapshotFrequencyMillis` to `replicatedSubscriptionsSnapshotTimeoutSeconds`.
+            //
+            // In other words, when hit this case, The frequency of sending SNAPSHOT_REQUEST
+            // will use `replicatedSubscriptionsSnapshotTimeoutSeconds`.
             if (log.isDebugEnabled()) {
                 log.debug("[{}] PendingSnapshot exists but has never succeeded. "
                         + "Skipping snapshot creation until pending snapshot timeout.", topic.getName());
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatedSubscriptionTest.java
```diff
@@ -1002,6 +1006,85 @@ public void testReplicatedSubscriptionWithCompaction() throws Exception {
         Assert.assertEquals(result, List.of("V2"));
     }
 
+    @Test
+    public void testReplicatedSubscriptionOneWay() throws Exception {
+        final String namespace = BrokerTestUtil.newUniqueName("pulsar-r4/replicatedsubscription");
+        final String topicName = "persistent://" + namespace + "/one-way";
+        config1.setReplicatedSubscriptionsSnapshotTimeoutSeconds(10);
```
lhotari, line 1013: Is it possible to use a shorter timeout to make the test complete faster?

lhotari: Could the snapshot frequency be increased for the test as well?

shibd: Yes, thanks for the suggestion. Addressed.

lhotari (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
```diff
@@ -220,6 +221,23 @@ private void receiveSubscriptionUpdated(ReplicatedSubscriptionsUpdate update) {
     private void startNewSnapshot() {
         cleanupTimedOutSnapshots();
 
+        if (lastCompletedSnapshotStartTime == 0 && !pendingSnapshots.isEmpty()) {
```
poorbarcode, line 224: Since you skipped the snapshot here, how can `lastCompletedSnapshotStartTime` be set to a non-null value>

poorbarcode (changes requested):

poorbarcode (approved):

Issue #23843, opened 2025-01-13
[Bug] can't create topic in k8s ipv6. broker 


zey1996:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

k8s with ipv6
imag version:3.3.3
no client, just cli.

### Minimal reproduce step

this is my broker deploy yaml file:
```yaml
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: "my-release-pulsar-broker"
  namespace: milvus
  labels:
    app: pulsar
    chart: pulsar-2.7.8
    release: my-release
    heritage: Helm
    cluster: my-release-pulsar
    component: broker
spec:
  serviceName: "my-release-pulsar-broker"
  replicas: 3
  selector:
    matchLabels:
      app: pulsar
      release: my-release
      component: broker
  updateStrategy:
    type: RollingUpdate
  podManagementPolicy: Parallel
  template:
    metadata:
      labels:
        app: pulsar
        release: my-release
        cluster: my-release-pulsar
        component: broker
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "8080"
    spec:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
            - podAffinityTerm:
                labelSelector:
                  matchLabels:
                    app: broker
                    component: bookie
                topologyKey: kubernetes.io/hostname
              weight: 1
      serviceAccountName: "my-release-pulsar-broker-acct"
      terminationGracePeriodSeconds: 30
      initContainers:
      - name: wait-zookeeper-ready
        image: "registry.paas/eos-indexer/pulsar:3.3.3"
        imagePullPolicy: IfNotPresent
        command: ["sh", "-c"]
        args:
          - >-
            until bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server my-release-pulsar-zookeeper:2181 get /admin/clusters/my-release-pulsar; do
              echo "pulsar cluster my-release-pulsar isn't initialized yet ... check in 3 seconds ..." && sleep 3;
            done;
        volumeMounts:
      - name: wait-bookkeeper-ready
        image: "registry.paas/eos-indexer/pulsar:3.3.3"
        imagePullPolicy: IfNotPresent
[...]

PR #23846, opened 2025-01-14
[improve] Support overriding java.net.preferIPv4Stack with OPTS

lhotari:
### Motivation

When using Pulsar with IPv6, it's necessary to set `-Djava.net.preferIPv4Stack=false` since Pulsar startup scripts set `-Djava.net.preferIPv4Stack=true`. Currently it's not possible to override the Pulsar default by passing `-Djava.net.preferIPv4Stack=false` in `OPTS` environment variable. This PR intends to make it possible.
This PR is related to the issue #23843.

### Modifications

Prepend the default value `-Djava.net.preferIPv4Stack=true` to `OPTS` instead of appending to it so that the setting in `OPTS` takes preference. The last setting in the argument line wins when setting system properties for Java processes.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/bin/bookkeeper
+++ b/bin/bookkeeper
@@ -201,7 +201,7 @@ OPTS="$OPTS --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/jdk
 OPTS="-cp $BOOKIE_CLASSPATH $OPTS"
 
 # Disable ipv6 as it can cause issues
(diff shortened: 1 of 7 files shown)


nodece (approved):
LGTM

nicoloboschi (approved):
+1

PR #23854, opened 2025-01-15
[fix][broker] Fix deadlock in Key_Shared PIP-379 implementation

### LINKED ISSUE #23848: [Bug] Deadlock in DrainingHashesTracker due to synchronized methods
### Search before asking

### Read release policy

### Version

Pulsar 4.0.1

### Minimal reproduce step

Exact steps to reproduce aren't yet confirmed.

This problem was faced in a test. The problem is that there isn't a `jstack -l <PID>` thread dump of the deadlocked state. I have a heap dump, but since heap dumps don't contain exactly the same information as a thread dump about threads, it's hard to find out the exact reason why the dead lock happens. There are multiple blocked threads which indicates a dead lock:

Image

Image

Image

Image

Image

Image

Image

Image

Image

No dead locks.

### What did you see instead?

There are multiple threads in blocked state which indicates a dead lock.

### Anything else?

_No response_

lhotari:
Fixes #23848

### Motivation

The Pulsar 4.0 Key_Shared PIP-379 implementation could deadlock and this problem was first captured in test environment heapdump where the thread stack information included in the heapdump showed multiple threads in blocked state, hinting a deadlock.
With the help of @pdolif, the problem was reproduced in isolation by simply adding this type of test code in a KeySharedSubscriptionTest test case.

```java
// test code that triggered the deadlock issue
new Thread(() -> {
    while(true) { pulsar.getBrokerService().updateRates(); }
}).start();
```
In the test case, this triggered a similar pattern of blocked threads and therefore could be considered a reproduction of the previously faced dead lock issue in a test environment.

threaddump with deadlock:
* in jstack.review https://jstack.review/?https://gist.github.com/lhotari/75eebb84687333bd7b133f2cbf3b6af6#tda_1_dump
* deadlock details: https://gist.github.com/lhotari/75eebb84687333bd7b133f2cbf3b6af6#file-deadlock-txt-L2244

### Modifications

- add test case to reproduce the dead lock case which will prevent future regressions in the same area
- fix the dead lock issue by replacing synchronization with read-write locks
  - locks are for minimum "scope"
  - callbacks are called outside of a lock scope to reduce the chances of potential other deadlocks
- use volatile fields for counters in DrainingHashEntry so that entry methods can be used without separate locks

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
@@ -26,6 +26,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.PrimitiveIterator;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.common.policies.data.DrainingHash;
@@ -34,7 +36,7 @@ import org.apache.pulsar.common.policies.data.stats.DrainingHashImpl;
 import org.roaringbitmap.RoaringBitmap;
 
 /**
- * A thread-safe map to store draining hashes in the consumer.
+ * A thread-safe map to store draining hashes in the consumer using read-write locks for improved concurrency.
  */
 @Slf4j
 public class DrainingHashesTracker {
@@ -42,6 +44,7 @@ public class DrainingHashesTracker {
     private final UnblockingHandler unblockingHandler;
     // optimize the memory consumption of the map by using primitive int keys
     private final Int2ObjectOpenHashMap<DrainingHashEntry> drainingHashes = new Int2ObjectOpenHashMap<>();
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
     int batchLevel;
     boolean unblockedWhileBatching;
     private final Map<ConsumerIdentityWrapper, ConsumerDrainingHashesStats> consumerDrainingHashesStatsMap =
@@ -52,9 +55,14 @@ public class DrainingHashesTracker {
      */
     @ToString
     public static class DrainingHashEntry {
+        private static final AtomicIntegerFieldUpdater<DrainingHashEntry> REF_COUNT_UPDATER =
+                AtomicIntegerFieldUpdater.newUpdater(DrainingHashEntry.class, "refCount");
+        private static final AtomicIntegerFieldUpdater<DrainingHashEntry> BLOCKED_COUNT_UPDATER =
+                AtomicIntegerFieldUpdater.newUpdater(DrainingHashEntry.class, "blockedCount");
+
         private final Consumer consumer;
-        private int refCount;
-        private int blockedCount;
+        private volatile int refCount;
+        private volatile int blockedCount;
 
         /**
          * Constructs a new DrainingHashEntry with the specified Consumer.
@@ -81,7 +89,7 @@ public class DrainingHashesTracker {
          * Increments the reference count.
          */
         void incrementRefCount() {
-            refCount++;
+            REF_COUNT_UPDATER.incrementAndGet(this);
         }
 
         /**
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
```diff
@@ -179,31 +221,43 @@ public DrainingHashesTracker(String dispatcherName, UnblockingHandler unblocking
...
+                    log.debug("[{}] Draining hash {} incrementing {} consumer id:{} name:{}", dispatcherName,
+                            stickyHash, entry.getRefCount() + 1, consumer.consumerId(), consumer.consumerName());
+                }
             }
+        } finally {
+            lock.writeLock().unlock();
+        }
+        if (addedStats != null) {
+            // add hash to added stats
+            addedStats.addHash(stickyHash);
```
heesung-sn, line 260: Is this required? It seems like this is no-op as addedStats is a local var in this func.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
```diff
@@ -231,13 +300,14 @@ public synchronized void endBatch() {
...
+     * @param closing    whether the consumer is closing
      */
-    public synchronized void reduceRefCount(Consumer consumer, int stickyHash, boolean closing) {
+    public void reduceRefCount(Consumer consumer, int stickyHash, boolean closing) {
         if (stickyHash == 0) {
             return;
         }
-        DrainingHashEntry entry = drainingHashes.get(stickyHash);
+
+        DrainingHashEntry entry = getEntry(stickyHash);
```
heesung-sn, line 310: is this thread-safe? should we put this after the lock is acquired?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
```diff
@@ -252,24 +322,40 @@ public synchronized void reduceRefCount(Consumer consumer, int stickyHash, boole
                 log.debug("[{}] Draining hash {} removing consumer id:{} name:{}", dispatcherName, stickyHash,
                         consumer.consumerId(), consumer.consumerName());
             }
-            DrainingHashEntry removed = drainingHashes.remove(stickyHash);
+            DrainingHashEntry removed;
+            lock.writeLock().lock();
+            try {
+                removed = drainingHashes.remove(stickyHash);
+            } finally {
+                lock.writeLock().unlock();
```
heesung-sn, line 330: is this necessary to lock and release this quick?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
```diff
@@ -34,14 +36,15 @@
 import org.roaringbitmap.RoaringBitmap;
 
 /**
- * A thread-safe map to store draining hashes in the consumer.
+ * A thread-safe map to store draining hashes in the consumer using read-write locks for improved concurrency.
```
heesung-sn, line 39: is this necessary to access this instance by multiple threads? Can we run this ops by a single thread?

lhotari: Yes, it's necessary. The draining hashes state is exposed in topic stats and multi-thread access for stats cannot be easily eliminated without additional complexity.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
lhotari: Yes, this is necessary. The stats instance has a separate lock. By making the call outside of the lock of the draining hashes instance, it reduces the chances for deadlocks. The deadlock that this PR is fixing was caused by such an access pattern in the reduceRefCount method.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
lhotari: Yes. getEntry method uses a read lock. It's intentional that the lock lifetime is as short as possible and that no other locks are acquired while holding the lock to reduce the chances for deadlocks. That's the strategy of this PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/DrainingHashesTracker.java
lhotari: Fair point. I'll also handle the "unblock" accesses during the same acquired lock.

heesung-sn (approved):

Issue #23845, opened 2025-01-13
[Bug] Key_Shared subscription doesn't deliver messages in the replay queue when no new messages are produced


lhotari:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

Pulsar 4.0.1

### Minimal reproduce step

Exact steps to reproduce aren't yet confirmed.

This problem was faced in a test where there was a large number of consumers that were scaled in a way where consumers were added and removed. The problem was noticed at the end of the test case, where all messages didn't get delivered to consumers and remained in the backlog.

In the topic stats for the subscription, `msgInReplay` showed a positive value and in internal stats for the subscription `subscriptionHavePendingRead` was true. By looking at the code, it seems to be a case that isn't handled for PersistentDispatcherMultipleConsumers/PersistentStickyKeyDispatcherMultipleConsumers. 

### What did you expect to see?

The cursor shouldn't go into completely into "waiting" state when there are messages in the replay queue.

### What did you see instead?

Messages in the replay queue don't get dispatched to consumers.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

lhotari:
There's already a solution to cancel a pending read when a hash gets unblocked:
https://github.com/apache/pulsar/blob/acac72ea03f7c38cab99ec011b309d5e6bb4fe9d/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L109-L122

There seems to be a bug in this behavior so that it didn't catch the case that was encountered. One possible reason for this is that a consumer didn't have permits when the unblocking happened. There would need to be some logic to handle that case.

walkinggo:
It looks like this piece of code triggers a re-read here. I tried the related test cases, but no errors occurred. Could you provide more details about what you meant by "permit"? @lhotari

lhotari:
> It looks like this piece of code triggers a re-read here. I tried the related test cases, but no errors occurred. Could you provide more details about what you meant by "permit"? [@lhotari](https://github.com/lhotari)

[...]

PR #23855, opened 2025-01-16
[fix][broker] Revert "[fix][broker] Cancel possible pending replay read in cancelPendingRead (#23384)"

lhotari:
### Motivation

There was a misunderstanding in making the change #23384. This PR reverts the change and adds javadoc to the `cancelPendingReadRequest` method to explain the purpose of the method.

### Modifications

- revert #23384 changes
- revert also this in PersistentDispatcherMultipleConsumersClassic
- add javadoc

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherMultipleConsumers.java
@@ -68,6 +68,10 @@ public abstract class AbstractDispatcherMultipleConsumers extends AbstractBaseDi
 
     public abstract boolean isConsumerAvailable(Consumer consumer);
 
+    /**
+     * Cancel a possible pending read that is a Managed Cursor waiting to be notified for more entries.
+     * This won't cancel any other pending reads that are currently in progress.
+     */
     protected void cancelPendingRead() {}
 
     /**
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -687,9 +687,8 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
 
     @Override
     protected void cancelPendingRead() {
-        if ((havePendingRead || havePendingReplayRead) && cursor.cancelPendingReadRequest()) {
+        if (havePendingRead && cursor.cancelPendingReadRequest()) {
             havePendingRead = false;
-            havePendingReplayRead = false;
         }
     }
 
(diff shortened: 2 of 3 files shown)


heesung-sn (approved):

shibd (approved):

poorbarcode:
Seems I have mentioned this issue at https://github.com/apache/pulsar/pull/23803, it does not matter, could you add a test for the current PR to avoid others doing the same change as https://github.com/apache/pulsar/pull/23384

poorbarcode (changes requested):
Seems I have mentioned this issue at https://github.com/apache/pulsar/pull/23803, it does not matter, could you add a test for the current PR to avoid others doing the same change as https://github.com/apache/pulsar/pull/23384

lhotari:
> Seems I have mentioned this issue at #23803, it does not matter, could you add a test for the current PR to avoid others doing the same change as #23384?

It's a common practice to revert a change that turned out to be invalid. In this case, although a similar change is made in #23384, it won't cause merge conflicts even if it were to be merged later. This PR contains a javadoc comment to improve the clarity of `cancelPendingRead`. I don't think that it's justified to block this PR with "request for changes". It's not a common practice to include tests for revert commits. We could add tests in other PRs.

poorbarcode:
@lhotari 

>  I don't think that it's justified to block this PR with "request for changes". It's not a common practice to include tests for revert commits. We could add tests in other PRs.

Sure, please create a new PR to add the test

poorbarcode (approved):

lhotari:
> @lhotari
> 
> > I don't think that it's justified to block this PR with "request for changes". It's not a common practice to include tests for revert commits. We could add tests in other PRs.
> 
> Sure, please create a new PR to add the test

> Seems I have mentioned this issue at https://github.com/apache/pulsar/pull/23803, it does not matter, could you add a test for the current PR to avoid others doing the same change as https://github.com/apache/pulsar/pull/23384?

@poorbarcode What type of test do you have in mind? In #23803 there aren't any comments about what you have in mind and there are no tests. Adding a test to avoid others doing the same change doesn't really make sense to me. A better way to address the problem would be to rename the method if we'd like to improve the clarity.
btw. I'm currently working on #23845 and it might result in adding some other tests in this area.

poorbarcode:
@lhotari 

> What type of test do you have in mind?

A test to reproduce the issue that #23384 introduced, which can prevent others from making the same change as https://github.com/apache/pulsar/pull/23384

lhotari:
> @lhotari
> 
> > What type of test do you have in mind?
> 
> A test to reproduce the issue that #23384 introduced, which can prevent others from making the same change as #23384

@poorbarcode There was no known issue that #23384 introduced. It's just a wrong change. For read requests, it's not possible to "cancel" a replay read and therefore having the logic in `cancelPendingReadRequest` method for setting `havePendingReplayRead=false` in the method doesn't simply make sense. I have clarified the meaning in the javadoc in this PR: "Cancel a possible pending read that is a Managed Cursor waiting to be notified for more entries. This won't cancel any other pending reads that are currently in progress.".

As I mentioned in the previous comment, renaming this method to a more specific name would be a better way to ensure that the abstraction doesn't get misunderstood in the future.

PR #23859, opened 2025-01-17
[fix][client] fix retry topic with exclusive mode.

thetumbled:
### Motivation

- Retry topic relies on the delayed queue feature. Once user call `reconsumeLater` for a message, this message will be produce as a delayed message to the corresponding retry topic. 
- Delayed queue feature can only work with shared/key-shared subscription mode, or the delayed message will be dispatched immediately with exclusive/failover mode.

Based on the analysis above, we can come to the conclusion that the consumer of the retry topic must be shared/key-shared.

### Modifications

Restrict the subscription type of the retry topic to be shared.

### Verifying this change

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


thetumbled:
And i have a question: why delayed message do not support exclusive type? is it a task that can't be implemented? or a pending task?

lhotari:
> And i have a question: why delayed message do not support exclusive type? is it a task that can't be implemented? or a pending task?

@thetumbled This question has been asked multiple times before and I didn't search for previous answers. I think that one of the reasons is that the semantics of delayed messages hasn't been defined and designed with Failover or Exclusive type. 
There's also technical reasons from implementation perspective. Naively adding support for delayed messages would result in a lot of code duplication in the dispatcher implementation classes. Avoiding code duplication would require refactoring of the dispatcher code and there's currently a lot of resistance in doing that. I noticed that when working on PIP-379 that there's a preference to minimize refactoring due to the risk of regressions. The state management in dispatchers isn't that great and there are bugs that pop up when a subtle change is made. I'm currently investigating #23845 and that's some of my observations about the dispatcher code. It's fairly hard to grasp it how it works due to the multiple levels of states that impact the execution. Multiple levels of state result in a vast state space, which isn't great. A better approach would be to minimize the state space. That could be achieved with a simpler execution model that addresses the different states that exist due to various hacks that have been added to address issues in the original implementation.

thetumbled:
> > And i have a question: why delayed message do not support exclusive type? is it a task that can't be implemented? or a pending task?
> 
> @thetumbled This question has been asked multiple times before and I didn't search for previous answers. I think that one of the reasons is that the semantics of delayed messages hasn't been defined and designed with Failover or Exclusive type. There's also technical reasons from implementation perspective. Naively adding support for delayed messages would result in a lot of code duplication in the dispatcher implementation classes. Avoiding code duplication would require refactoring of the dispatcher code and there's currently a lot of resistance in doing that. I noticed that when working on PIP-379 that there's a preference to minimize refactoring due to the risk of regressions. The state management in dispatchers isn't that great and there are bugs that pop up when a subtle change is made. I'm currently investigating #23845 and that's some of my observations about the dispatcher code. It's fairly hard to grasp it how it works due to the multiple levels of states that impact the execution. Multiple levels of state result in a vast state space, which isn't great. A better approach would be to minimize the state space. That could be achieved with a simpler execution model that addresses the different states that exist due to various hacks that have been added to address issues in the original implementation.

I agree with the complexity of the dispatcher. It is too complicate to master this module and it is hard to guarantee that there is nothing wrong when a seemingly good pr is merged.

thetumbled:
PATL, thanks. @lhotari @BewareMyPower @poorbarcode @dao-jun @codelipenghui @nodece

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBuilderImpl.java
```diff
@@ -151,7 +151,7 @@ public CompletableFuture<Consumer<T>> subscribeAsync() {
                     Math.max(conf.getBatchReceivePolicy().getMaxNumMessages(), conf.getReceiverQueueSize()));
         }
         CompletableFuture<Void> applyDLQConfig;
-        if (conf.isRetryEnable() && conf.getTopicNames().size() > 0) {
+        if (conf.isRetryEnable() && !conf.getTopicNames().isEmpty()) {
```
BewareMyPower, line 154: This change is not related to the PR. It's better to revert it or use a separated PR for such small trivial changes.

poorbarcode (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBuilderImpl.java
thetumbled: Updated.

dao-jun (approved):

BewareMyPower (approved):

shibd:
May be there is a break change:

1. There are 2 consumers.
2. Use key_share to subscribe topic, and key_share will also subscribe retry-topic
3. Rolling upgrade to 3.3.5
4. The first consumer will fail to start because it attempts to subscribe to the retry topic using the "shared" subscription type.

```
"errorMsg":"Subscription is of different type. Active subscription type of 'Key_Shared' is different than the connecting consumer's type 'Shared'.",
```

lhotari:
> > Maybe there is a breaking change:
> > 
> > 1. There are two consumers for a subscription
> > 2. Use key_share to subscribe topic, and key_share will also subscribe retry-topic
> > 3. Rolling upgrade to 3.3.5
> > 4. The first consumer will fail to start because it attempts to subscribe to the retry topic using the "shared" subscription type.
> > 
> > ```
> > "errorMsg":"Subscription is of different type. Active subscription type of 'Key_Shared' is different than the connecting consumer's type 'Shared'.",
> > ```
> 
> Maybe a complete stop and start process can avoid the problem?

This isn't a broker-side problem. The problem is caused by clients connecting to the broker. This is a challenging problem to address since mixed client versions cause the issue. I don't really see a way around it other than to keep the fix. The decision to upgrade the client version can be controlled and coordinated by users. There should be better release notes in Pulsar where such breaking changes are highlighted and where there would be instructions on how to deal with them.

PR #17241, opened 2022-08-23
[managed-ledger] Do not send duplicate reads to BK/offloaders

eolivelli:
### Motivation
When you have many subscriptions on the same topic and they are catching up (for instance after some downtime of the consumer application) Pulsar will start to perform many reads from BK for the same entries.

### Modifications
Prevent concurrent reads of the same entries.

### Verifying this change

This change is already covered by existing tests.

I also tested  manually that with this fix we are reducing a lot the pressure on the bookies. This is something we cannot code in a integration test as it will use too many resources on CI.

With some tests with only 64 subscription we saw a reduction from 100K reads/s from BK to 250 reads/s.

PR #23869, opened 2025-01-20
[fix][broker] Fix repeatedly acquired pending reads quota

poorbarcode:
### Motivation

**Background**
- Broker will limit the pending reading to Bookies if enabled `managedLedgerMaxReadsInFlightSize`
- https://github.com/apache/pulsar/pull/17241 introduced a mechanism that merges BK reading requests if we can, for example:
  - Request-1: Read entry `50~59` 
  - Request-2: Read entry `50~69`
  - The `Request-2` will wait for `Request-1`, and only send a real request that reads `60~69` after `Request-1` is finished.

**Issue**
- The `Request-2` above requests repeatedly
  - It requests `50~70` when creating.
  - It requests another range `61~70` after `Request-1` is finished.
- Expected: `Request-2` acquire `20` quota.
- Exact:  `Request-2` acquire `40` quota.
- You can call `testPreciseLimitation` to reproduce the issue

### Modifications

- Fix the issue.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/InflightReadsLimiterIntegrationTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/InflightReadsLimiterIntegrationTest.java
@@ -1,0 +1,231 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.mledger.impl;
+
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.doAnswer;
+import io.netty.util.concurrent.DefaultThreadFactory;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.client.LedgerHandle;
+import org.apache.bookkeeper.client.api.LedgerEntries;
+import org.apache.bookkeeper.mledger.AsyncCallbacks;
+import org.apache.bookkeeper.mledger.Entry;
+import org.apache.bookkeeper.mledger.ManagedLedgerConfig;
+import org.apache.bookkeeper.mledger.ManagedLedgerException;
+import org.apache.bookkeeper.mledger.ManagedLedgerFactoryConfig;
+import org.apache.bookkeeper.mledger.impl.cache.InflightReadsLimiter;
+import org.apache.bookkeeper.mledger.impl.cache.RangeEntryCacheImpl;
+import org.apache.bookkeeper.mledger.impl.cache.RangeEntryCacheManagerImpl;
+import org.apache.bookkeeper.test.MockedBookKeeperTestCase;
+import org.awaitility.Awaitility;
+import org.awaitility.reflect.WhiteboxImpl;
+import org.mockito.Mockito;
+import org.mockito.stubbing.Answer;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+@Slf4j
+public class InflightReadsLimiterIntegrationTest extends MockedBookKeeperTestCase {
+
+    @DataProvider
+    public Object[][] readMissingCases() {
+        return new Object[][]{
+                {"missRight"},
+                {"missLeft"},
+                {"bothMiss"}
+        };
+    }
+
+    @Test(dataProvider = "readMissingCases")
+    public void testPreciseLimitation(String missingCase) throws Exception {
+        final long start1 = 50;
+        final long start2 = "missLeft".endsWith(missingCase) || "bothMiss".equals(missingCase) ? 30 : 50;
+        final long end1 = 99;
+        final long end2 = "missRight".endsWith(missingCase) || "bothMiss".equals(missingCase) ? 109 : 99;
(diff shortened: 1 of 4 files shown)


managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/InflightReadsLimiterIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+                        res.completeExceptionally(ex);
+                    }
+                }).start();
+                return res;
+            } else {
+                return invocation.callRealMethod();
+            }
+        };
+        doAnswer(answer).when(spyCurrentLedger).readAsync(anyLong(), anyLong());
+        doAnswer(answer).when(spyCurrentLedger).readUnconfirmedAsync(anyLong(), anyLong());
```
lhotari, line 138: I wonder if the PulsarMockBookKeeper read handle interceptor (added in #23875) could be used here to avoid the use of Mockito?

lhotari:
I agree that this issue exists, however there are broader issues in the `managedLedgerMaxReadsInFlightSizeInMB` solution.

Sharing some context about the issues I have found and what I have currently in progress:

I have reported issues apache/pulsar#23482, apache/pulsar#23504, apache/pulsar#23505 and apache/pulsar#23506 .
These issues are related to the broker cache and `managedLedgerMaxReadsInFlightSizeInMB` and `dispatcherMaxReadSizeBytes` limits.

I have already changes to address these issues [in an experimental branch](https://github.com/apache/pulsar/compare/master...lhotari:lh-fix-individual-read-deduplication-and-limiting), pending the submission of individual PRs. Addressing the lack of caching for replay queue messages requires broader changes to the the broker cache and those will be covered with a new PIP.

Some earlier details shared in this comment: https://github.com/apache/pulsar/pull/23524#issuecomment-2461686043 .

Due to issues #23482 and #23506, I don't think that this PR would resolve problems alone. Regarding broker OOMEs, issue #23504 would also need to be resolved. It's possible that `dispatcherDispatchMessagesInSubscriptionThread=false` could be a mitigation to some issues.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/InflightReadsLimiterIntegrationTest.java
poorbarcode: I agree, but the current mock is compatible with both the mocked Bookie client and the real Bookie client.

poorbarcode:
@lhotari 

> I have already changes to address these issues [in an experimental branch](https://github.com/apache/pulsar/compare/master...lhotari:lh-fix-individual-read-deduplication-and-limiting), pending the submission of individual PRs. Addressing the lack of caching for replay queue messages requires broader changes to the the broker cache and those will be covered with a new PIP.

> Some earlier details shared in this comment: https://github.com/apache/pulsar/pull/23524#issuecomment-2461686043 .

> Due to issues https://github.com/apache/pulsar/issues/23482 and https://github.com/apache/pulsar/issues/23506, I don't think that this PR would resolve problems alone. Regarding broker OOMEs, issue https://github.com/apache/pulsar/issues/23504 would also need to be resolved. It's possible that dispatcherDispatchMessagesInSubscriptionThread=false and managedLedgerReadEntryTimeoutSeconds=15 (some reasonable value) could be a mitigation to some issues.

Agree with you, let us fix them one by one, which is easier to review, and we should add tests for each case.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/InflightReadsLimiterIntegrationTest.java
lhotari: That's true, but this test uses `PulsarMockBookKeeper`. The problem with Mockito is that it requires injecting changes deeply into the implementation. Incorrect use of Mockito could result in thread safety issues and other issues, causing flaky tests.
btw. When there are a large number of operations, Mockito Spies that record invocations should only be used when the invocations are validated. We have a helper for spies that don't record invocations (that could cause OOME issues in tests): https://github.com/apache/pulsar/blob/52e8730613c36008ea57a0ca5c10231512232d7e/pulsar-broker/src/test/java/org/apache/pulsar/broker/BrokerTestUtil.java#L99-L112 . It's not a problem in this test since there's not a large invocation count. The point is that with the read handle interceptor support in PulsarMockBookKeeper, this type of issues could be avoided in the first place.

heesung-sn:
I think we should use AsyncTokenBucket or Guava.RateLimiter.tryAcquire for this rate limiter. 

IMHO, the current logic, requiring acquiredPermit to release is error-prone, when the caller forgets to release it. Instead, I think we better use token bucket-based one, which can automatically fill the bucket.

lhotari:
> I think we should use AsyncTokenBucket or Guava.RateLimiter.tryAcquire for this rate limiter.
> 
> IMHO, the current logic, requiring acquiredPermit to release is error-prone, when the caller forgets to release it. Instead, I think we better use token bucket-based one, which can automatically fill the bucket.

@heesung-sn I agree, the current solution is problematic. However, a token bucket isn't most optimal for this use case. I have a redesigned solution the `managedLedgerMaxReadsInFlightSize` solution [in my experimental branch](https://github.com/apache/pulsar/compare/master...lhotari:lh-fix-individual-read-deduplication-and-limiting#diff-f63b07497cff099f2245505beafb4134c8ad983d2d0a984ee53fba90488713f8). There are a lot of different changes in branch so it might be hard to see what it's about.  I'll extract that into a clear PR when time comes.
There's first #23892 and #23894 which need reviews so that I could make further progress. Do you have a chance to review them?

lhotari (approved):
LGTM

lhotari:
> > I think we should use AsyncTokenBucket or Guava.RateLimiter.tryAcquire for this rate limiter.
> > IMHO, the current logic, requiring acquiredPermit to release is error-prone, when the caller forgets to release it. Instead, I think we better use token bucket-based one, which can automatically fill the bucket.
> 
> @heesung-sn I agree, the current solution is problematic. However, a token bucket isn't most optimal for this use case. I have a redesigned solution the `managedLedgerMaxReadsInFlightSize` solution [in my experimental branch](https://github.com/apache/pulsar/compare/master...lhotari:lh-fix-individual-read-deduplication-and-limiting#diff-f63b07497cff099f2245505beafb4134c8ad983d2d0a984ee53fba90488713f8) ([example](https://github.com/lhotari/pulsar/blob/lh-fix-individual-read-deduplication-and-limiting/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java)). There are a lot of different changes in branch so it might be hard to see what it's about. I'll extract that into a clear PR when time comes. There's first #23892 and #23894 which need reviews so that I could make further progress. Do you have a chance to review them?

@heesung-sn @poorbarcode To continue the improvements to address issues with `managedLedgerMaxReadsInFlightSize`, I have created #23901 which addresses the problems I have referred to earlier. Please review

PR #23874, opened 2025-01-21
[fix][broker] PIP-379 Key_Shared implementation race condition causing out-of-order message delivery

### LINKED ISSUE #23870: [Bug] PIP-379 Key_Shared implementation can deliver messages out-of-order due to racecondition when a hash gets unblocked
### Search before asking

### Read release policy

### Version

4.0.0, 4.0.1, 4.0.2

### Minimal reproduce step

It was possible to reproduce the issue in a test case by disabling broker cache and by setting dispatcher retry backoff and keySharedUnblockingIntervalMs to 0. 
```
        conf.setManagedLedgerCacheSizeMB(0);
        conf.setManagedLedgerMaxReadsInFlightSizeInMB(0);
        conf.setDispatcherRetryBackoffInitialTimeInMs(0);
        conf.setDispatcherRetryBackoffMaxTimeInMs(0);
        conf.setKeySharedUnblockingIntervalMs(0);
```
The test case is
https://github.com/lhotari/pulsar/blob/caac334465ca3de9a87e93a13b8fee377a2a467b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/KeySharedSubscriptionDisabledBrokerCacheTest.java

The problem doesn't reproduce in testing with the default configuration when the retry backoff and keySharedUnblockingIntervalMs are set to default. However, the test reveals a race condition that needs to be addressed.  

Messages should be delivered in order, also when retry backoff and keySharedUnblockingIntervalMs is set to 0.

### What did you see instead?

There are message ordering issues in 25%-40% of the test runs.

### Anything else?

The problem seems to be a race condition of the hash getting unblocked while sending of entries is in progress.

lhotari:
Fixes #23870

### Motivation

See #23870 for details. It was possible to reproduce a racecondition that caused out-of-order message delivery.
This happened when a draining hash got unblocked by an ack while message replay filtering was in progress.

### Modifications

Fix provided by @poorbarcode in comment https://github.com/apache/pulsar/pull/23874#pullrequestreview-2567038955
- track the hashes that have been blocked during the filtering of replay entries
  - it is necessary to block all later messages after a hash gets blocked so that ordering is preserved

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
There's some remaining issue in the test where it fails in a different way. Will investigate later.

lhotari:
> There's some remaining issue in the test where it fails in a different way. Will investigate later.

Resolved the issue. The test contained a race condition.

testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockReadHandleInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.bookkeeper.client;
+
+import java.util.concurrent.CompletableFuture;
+import org.apache.bookkeeper.client.api.LedgerEntries;
+
+/**
+ * Interceptor interface for intercepting read handle readAsync operations.
+ * This is useful for testing purposes, for example for introducing delays.
+ */
+public interface PulsarMockReadHandleInterceptor {
```
poorbarcode, line 28: Seems it never been used, if yes, can you remove these changes that related this interface

poorbarcode (changes requested):

testmocks/src/main/java/org/apache/bookkeeper/client/PulsarMockReadHandleInterceptor.java
lhotari: The test framework improvements and fixes have been moved to #23875 and #23876

poorbarcode (changes requested):
I think we'd better solve the issue this way, which has less lock range, better performance, and easier to read

**PersistentStickyKeyDispatcherMultipleConsumers.ReplayPositionFilter.java**
```java
    private class ReplayPositionFilter implements Predicate<Position> {
        // tracks the available permits for each consumer for the duration of the filter usage
        // the filter is stateful and shouldn't be shared or reused later
        private final Map<Consumer, MutableInt> availablePermitsMap = new HashMap<>();

+       private HashSet<Long> stuckKeys = new HashSet<>();

        @Override
        public boolean test(Position position) {
            // if out of order delivery is allowed, then any position will be replayed
            if (isAllowOutOfOrderDelivery()) {
                return true;
            }
            // lookup the sticky key hash for the entry at the replay position
            Long stickyKeyHash = redeliveryMessages.getHash(position.getLedgerId(), position.getEntryId());
            if (stickyKeyHash == null) {
                // the sticky key hash is missing for delayed messages, the filtering will happen at the time of
                // dispatch after reading the entry from the ledger
                if (log.isDebugEnabled()) {
                    log.debug("[{}] replay of entry at position {} doesn't contain sticky key hash.", name, position);
                }
                return true;
            }
+           if (stuckKeys.contains(stickyKeyHash)) {
+               return false;
+           }

            // find the consumer for the sticky key hash
            Consumer consumer = selector.select(stickyKeyHash.intValue());
            // skip replaying the message position if there's no assigned consumer
            if (consumer == null) {
+               stuckKeys.add(stickyKeyHash);
                return false;
            }
            // lookup the available permits for the consumer
            MutableInt availablePermits =
                    availablePermitsMap.computeIfAbsent(consumer,
                            k -> new MutableInt(getAvailablePermits(consumer)));
            // skip replaying the message position if the consumer has no available permits
            if (availablePermits.intValue() <= 0) {
 +             stuckKeys.add(stickyKeyHash);
                return false;
            }

            if (drainingHashesRequired
                    && drainingHashesTracker.shouldBlockStickyKeyHash(consumer, stickyKeyHash.intValue())) {
                // the hash is draining and the consumer is not the draining consumer
 +              stuckKeys.add(stickyKeyHash);
                return false;
            }

            availablePermits.decrement();
            return true;
        }
    }
}
```

lhotari:
> I think we'd better solve the issue this way, which has less lock range, better performance, and easier to read

@poorbarcode This looks like a useful optimization, but I don't think that it addresses the reported issue #23870. In that case, the problem occurs at a different stage. 

In the issue, it's an ordinary read that is happening and when the hash gets unblocked, this is the line where the hash is no longer blocked:
https://github.com/apache/pulsar/blob/e5bd77419e91d1602731cd0c1d02a738e1b7ebc7/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L510

The last acknowledgement will remove the hash from unblocking before it has been added to the replay queue. The alternative solution for this PR would be to postpone the removal after the message has been added to the replay queue.

lhotari:
> > I think we'd better solve the issue this way, which has less lock range, better performance, and easier to read
> 
> @poorbarcode thanks, I'll check the details to see if it addresses the issue.

@poorbarcode It fixes the test case in this PR, but it doesn't address another race condition that is possible. I'd have to come up with another test case that reproduces another possible race condition which would occur when there would be a race in an ordinary read and unblocking of a "draining" hash.

@poorbarcode It would be helpful if you could review the test framework improvement PRs #23875 and #23876 since I might need the read handle interceptor solution for introducing a race condition in a realistic scenario. The read handle interceptor solution is much cleaner than injecting behaviors with Mockito.

lhotari:
> > > I think we'd better solve the issue this way, which has less lock range, better performance, and easier to read
> > 
> > 
> > @poorbarcode thanks, I'll check the details to see if it addresses the issue.
> 
> @poorbarcode It fixes the test case in this PR, but it doesn't address another race condition that is possible. I'd have to come up with another test case that reproduces another possible race condition which would occur when there would be a race in an ordinary read and unblocking of a "draining" hash.
> 
> @poorbarcode It would be helpful if you could review the test framework improvement PRs #23875 and #23876 since I might need the read handle interceptor solution for introducing a race condition in a realistic scenario. The read handle interceptor solution is much cleaner than injecting behaviors with Mockito.

lhotari:
> > I think we'd better solve the issue this way, which has less lock range, better performance, and easier to read
> 
> @poorbarcode thanks, I'll check the details to see if it addresses the issue.

@poorbarcode Yes, it looks good. Great solution.

Just thinking if a race could happen at that stage (in sending/dispatching) when a draining hash is unblocked: https://github.com/apache/pulsar/blob/e5bd77419e91d1602731cd0c1d02a738e1b7ebc7/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L510

lhotari:
> Just thinking if a race could happen at that stage (in sending/dispatching) when a draining hash is unblocked:
> 
> https://github.com/apache/pulsar/blob/e5bd77419e91d1602731cd0c1d02a738e1b7ebc7/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L510

This is the complete method:
https://github.com/apache/pulsar/blob/e5bd77419e91d1602731cd0c1d02a738e1b7ebc7/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentStickyKeyDispatcherMultipleConsumers.java#L495-L519

I guess that unblocking at this stage wouldn't cause a race where messages could get delivered out-of-order. At least I'm not seeing it, although I first thought that there could be such a case.

lhotari:
> I guess that unblocking at this stage wouldn't cause a race where messages could get delivered out-of-order. At least I'm not seeing it, although I first thought that there could be such a case.

Just to be sure, I added a similar solution in a5d2029. There's also a slight performance benefit that there won't be a need to run all checks when the hash is already blocked in one round. I used an efficient `IntSet` implementation from fastutil to avoid int -> Integer boxing.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -335,6 +335,13 @@ public synchronized void readMoreEntries() {
             }
             return;
         }
+        if (havePendingReplayRead) {
```
poorbarcode, line 338: I am considering the changes in this class is unnecessary, could you remove them?

poorbarcode (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
lhotari: Ok, I'll move the havePendingReplayRead state handling changes to a separate PR. It's strictly not necessary for addressing the issue that the test reproduced.

lhotari: In any case, I'll move the havePendingReplayRead changes to another PR. Reproducing the race condition referred to in the previous comment is hard to reproduce and most likely reproducing would require introducing delays. The PulsarMockBookKeeper read handle interceptor in #23875 would be useful for that purpose. @poorbarcode Could you please review it in the mean time? It should be a quick review to cover #23875 and #23876.

lhotari: @poorbarcode havePendingReplayRead state handling changes have been removed. PTAL

poorbarcode (approved):
LGTM

PR #22860, opened 2024-06-06
[fix][broker] The topic might reference a closed ledger

shibd:
### Motivation
We observe that a `normal topic` might reference a `closed ledger` and it never auto recover. will cause the producer and customer stuck.

The root cause is related to `topic create timeout`. Assuming we have two concurrent requests to create a topic: `firstCreate(Thread-1)`, `secondCreate(Thread-2)`

| Thread-1                                            | Thread-2                                                                      |
|----------------|-------------------------------------------------|
| firstCreate topic timeout                              |                                                                               |
|                                                     | [remove this old topic and recreate a new topic](https://github.com/shibd/pulsar/blob/f07b3a030179c38f9786b3e26c82aa13e00b34a6/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1009-L1012)                                |
|                                                     | Open ledger from cache, and get an old ledger  that create by firstCreate topic                                        |
| call [topic.close  ](https://github.com/apache/pulsar/blob/31bfded7d717afd3e76de4d839bdb3daebf704a0/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1749-L1755)                              |                                                                               |
| old ledger close and remove it from cache |                                                                               |
|                                                     | but this `old ledger` being referenced to `new topic` and that stats is close |

- When the `firstCreate` topic timeout, will call  `topic.close`. it will close the ledger, and remove it from the ledger cache.
https://github.com/apache/pulsar/blob/f07b3a030179c38f9786b3e26c82aa13e00b34a6/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1745-L1752

- If the `secondCreate` request successfully creates a topic before the `old ledger closes`, the reference will be made to the `old ledger`.

### Modifications
Refactor `BrokerService#getTopic()` method:
  - If any exceptions occur during the creation topics process, they should be proactively and asynchronously removed from the topics: `pulsar.getExecutor().execute(() -> topics.remove(topicName.toString()));`
  - The new get topic operation should not remove the `topicFuture` that created by previous get topic operation. If the previous topicFuture is not removed from the map yet, the broker should always use the existing topicFuture (waiting for completion or return an error to the client side directly)
  - Add documentation to clearly define the method boundaries
[...]

PR #17526, opened 2022-09-07
[fix][broker]Consumer can't consume messages because there has two sames topics in one broker

poorbarcode:
### Motivation

With the transaction feature, we send and receive messages, and at the same time, execute `admin API: unload namespace` 1000 times. Then the problem occur: Consumer could not receive any message, and there has no error log. After that we tried `admin API: get topic stats`, and the response showed only producers are registered on topic, and no consumers are registered on topic, but consumer stat is `Ready` in the client. This means that the state of the consumer is inconsistent between the broker and the client.

#### Location problem

Then we found the problem: Two PersistentTopic which have the same name registered at a broker node, consumer registered on one (aka `topic-c`), and producer registered on another one (aka `topic-p`). At this time, when we send messages, the data flow like this : 

```text
client: producer sends a message

broker: handle cmd-send

broker: find the topic by name, it is "topic-p"

broker: find all subscriptions registered on "topic-p"

broker: found one subscription, but it has no consumers registered

broker: no need to send the message to the client
``` 

But the consumer exactly registered on another topic: `topic-c`, so consumer could not receive any message.

#### Repreduce

> *How to reproduce two topics registered at the same broker node ?*

Make  `transaction buffer recover`, `admin unload namespace`, `client create consumer`, ` client create producer` executed at the same time, the process flow like this (at the step-11, the problem occurs ):

| Time | `transaction buffer recoverr` | `admin unload namespace` | `client create consumer` | `client create producer` |
| ----------- | ----------- | ----------- | ----------- | ----------- |
| 1 | TB recover |  |  |  |
| 2 | TB recover failure | topic.unload |  |  |
| 3 | topic.close(false) | topic.close(true) |  |  |
| 4 | brokerService.topics.remove(topicName) |  |  |  |
| 5 | remove topic finish |  | lookup |  |
| 6 |  |  | create `topic-c` |  |
| 7 |  |  | consumer registered on `topic-c` |  |
| 8 |  | brokerService.topics.remove(topic) |  |  |
| 9 |  | remove `topic-c` finish |  | lookup |
| 10 |  |  |  | create `topic-p`  |
| 11 |  |  |  | producer registered on `topic-p`  |

- Each column means the individual process. e.g. `client create consumer`,  `client create producer`.
- Multiple processes are going on at the same time, and all effet the `brokerService.topics`.
- Column `Time` is used only to indicate the order of each step, not the actual time.
- The important steps are explained below:

> step 3

[...]

PR #23884, opened 2025-01-23
[fix][broker] Closed topics won't be removed from the cache

shibd:
### Motivation
When transaction component init fails, it will try to close the topic, but maybe not remove topic from cache.

| time | thread `createTransactionBufferSystemTopicClient` | thread `create topic` |
| --- | --- | --- |
| 1 |  | topic ledger open complete and [new PersistentTopic object](https://github.com/apache/pulsar/blob/b02d52ca8c8afd62681c0e243d16d8958abb5380/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1800-L1802)|
| 2 |  | new TransactionBuffer |
| 3 |  | Try get [AbortedTxnProcessor](https://github.com/apache/pulsar/blob/9ebd97941d8ebe074423349d03307683e7fa3de2/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java#L56-L60) |
| 4 | [`Failed to create snapshot writer` and close topic](https://github.com/apache/pulsar/blob/9ebd97941d8ebe074423349d03307683e7fa3de2/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java#L56-L60) |  |
| 5 | set topic fenced and closing |  |
| 6 | try remove topic from broker cache, [since the topic feature is not done, will ignore it](https://github.com/apache/pulsar/blob/b02d52ca8c8afd62681c0e243d16d8958abb5380/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L2427-L2430) |  |
| 7 |  | Topic future success complete |

### Modifications
- Add topic create future to topic.
- When remove a topic, call `removeTopicFutureFromCache` and passed topic future.
- Remove method `removeTopicFromCache(Topic)` that intrudce from #17526
  - At that time, there could have been two topicFutures for the same topic simultaneously (one creating, one closing). However, this case no longer occurs because we fixed it in [#22860](https://github.com/apache/pulsar/pull/22860/files#diff-0210356c8a88e4efa89eb769a027fa6c166db479dbad8bbbbc704c6ed6e317f5L1007-L1032), ensuring that a topic cannot be removed until it is fully created or closed. 

### Verifying this change
- You can run `testNoOrphanClosedTopicIfTxnInternalFailed` to reproduce this.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

## Test PR
https://github.com/shibd/pulsar/pull/53

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -96,6 +96,8 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
     protected static final long POLICY_UPDATE_FAILURE_RETRY_TIME_SECONDS = 60;
 
     protected final String topic;
+
+    // createFuture associated with the topic to ensure safe removal from the cache of brokerService
     @Getter
     @Setter
     protected volatile CompletableFuture<Optional<Topic>> createFuture;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -97,7 +97,9 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
 
     protected final String topic;
 
-    // createFuture associated with the topic to ensure safe removal from the cache of brokerService
+    // Reference to the CompletableFuture returned when creating this topic in BrokerService.
+    // Used to safely remove the topic from BrokerService's cache by ensuring we remove the exact
+    // topic instance that was created.
     @Getter
     @Setter
     protected volatile CompletableFuture<Optional<Topic>> createFuture;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -48,6 +48,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.ToLongFunction;
 import javax.annotation.Nonnull;
 import lombok.Getter;
+import lombok.Setter;
 import org.apache.bookkeeper.mledger.util.StatsBuckets;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
@@ -95,6 +96,9 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
     protected static final long POLICY_UPDATE_FAILURE_RETRY_TIME_SECONDS = 60;
 
     protected final String topic;
+    @Getter
+    @Setter
+    protected volatile CompletableFuture<Optional<Topic>> createFuture;
 
     // Producers currently connected to this topic
     protected final ConcurrentHashMap<String, Producer> producers;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -1326,6 +1326,7 @@ public class BrokerService implements Closeable {
         NonPersistentTopic nonPersistentTopic;
         try {
             nonPersistentTopic = newTopic(topic, null, this, NonPersistentTopic.class);
+            nonPersistentTopic.setCreateFuture(topicFuture);
         } catch (Throwable e) {
             log.warn("Failed to create topic {}", topic, e);
             topicFuture.completeExceptionally(e);
@@ -1800,6 +1801,7 @@ public class BrokerService implements Closeable {
                                 PersistentTopic persistentTopic = isSystemTopic(topic)
                                         ? new SystemTopic(topic, ledger, BrokerService.this)
                                         : newTopic(topic, ledger, BrokerService.this, PersistentTopic.class);
+                                persistentTopic.setCreateFuture(topicFuture);
                                 persistentTopic
                                         .initialize()
                                         .thenCompose(__ -> persistentTopic.preCreateSubscriptionForCompactionIfNeeded())
@@ -2409,42 +2411,7 @@ public class BrokerService implements Closeable {
         return authorizationService;
     }
 
-    public CompletableFuture<Void> removeTopicFromCache(Topic topic) {
-        Optional<CompletableFuture<Optional<Topic>>> createTopicFuture = findTopicFutureInCache(topic);
-        if (createTopicFuture.isEmpty()){
-            return CompletableFuture.completedFuture(null);
-        }
-        return removeTopicFutureFromCache(topic.getName(), createTopicFuture.get());
-    }
-
-    private Optional<CompletableFuture<Optional<Topic>>> findTopicFutureInCache(Topic topic){
-        if (topic == null){
-            return Optional.empty();
-        }
-        final CompletableFuture<Optional<Topic>> createTopicFuture = topics.get(topic.getName());
-        // If not exists in cache, do nothing.
-        if (createTopicFuture == null){
-            return Optional.empty();
-        }
-        // If the future in cache is not yet complete, the topic instance in the cache is not the same with the topic.
-        if (!createTopicFuture.isDone()){
-            return Optional.empty();
-        }
-        // If the future in cache has exception complete,
-        // the topic instance in the cache is not the same with the topic.
-        if (createTopicFuture.isCompletedExceptionally()){
-            return Optional.empty();
-        }
-        Optional<Topic> optionalTopic = createTopicFuture.join();
-        Topic topicInCache = optionalTopic.orElse(null);
-        if (topicInCache == null || topicInCache != topic){
-            return Optional.empty();
-        } else {
-            return Optional.of(createTopicFuture);
-        }
-    }
-
-    private CompletableFuture<Void> removeTopicFutureFromCache(String topic,
+    public CompletableFuture<Void> removeTopicFutureFromCache(String topic,
                                                         CompletableFuture<Optional<Topic>> createTopicFuture) {
         TopicName topicName = TopicName.get(topic);
         return pulsar.getNamespaceService().getBundleAsync(topicName)
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -2411,12 +2411,18 @@ public class BrokerService implements Closeable {
         return authorizationService;
     }
 
-    public CompletableFuture<Void> removeTopicFutureFromCache(String topic,
-                                                        CompletableFuture<Optional<Topic>> createTopicFuture) {
-        TopicName topicName = TopicName.get(topic);
+    /**
+     * Removes the topic from the cache only if the topicName and associated createFuture match exactly.
+     * The TopicEvent.UNLOAD event will be triggered before and after removal.
+     *
+     * @param topic The topic to be removed.
+     * @return A CompletableFuture that completes when the operation is done.
+     */
+    public CompletableFuture<Void> removeTopicFromCache(AbstractTopic topic) {
+        TopicName topicName = TopicName.get(topic.getName());
         return pulsar.getNamespaceService().getBundleAsync(topicName)
                 .thenAccept(namespaceBundle -> {
-                    removeTopicFromCache(topic, namespaceBundle, createTopicFuture);
+                    removeTopicFromCache(topic.getName(), namespaceBundle, topic.getCreateFuture());
                 });
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TopicTransactionBuffer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TopicTransactionBuffer.java
@@ -109,7 +109,25 @@ public class TopicTransactionBuffer extends TopicTransactionBufferState implemen
     private final AbortedTxnProcessor.SnapshotType snapshotType;
     private final MaxReadPositionCallBack maxReadPositionCallBack;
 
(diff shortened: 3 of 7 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
```diff
@@ -54,10 +56,20 @@ public SingleSnapshotAbortedTxnProcessorImpl(PersistentTopic topic) {
...
+            // be completed at that time, which could leave closed topics in the cache(at BrokerService).
+            CompletableFuture<Optional<Topic>> topicFuture = topic.getBrokerService().getTopics().get(topic.getName());
+            if (topicFuture != null) {
+                topicFuture.thenAccept(t -> t.ifPresent(v -> {
+                    v.close(true).exceptionally(ec -> {
+                        log.error("Close topic {} exception", v.getName(), ec);
+                        return null;
+                    });
+                }));
+            }
```
lhotari, line 70: This code is now duplicated in multiple locations. Instead of adding code duplication, 
I'd suggest to create a new public method directly in BrokerService which called  `closeTopicForcefullyIfExists`. The javadoc should describe the purpose.

lhotari (changes requested):
Please reduce code duplication

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
BewareMyPower: +1

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionPersistentTopicTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void testNoOrphanClosedTopicIfTxnInternalFailed() throws Exception {
+        String tpName = BrokerTestUtil.newUniqueName("persistent://public/default/tp2");
+
+        // 1. Intercept when the `topicFuture` is about to complete and wait until the topic close operation finishes.
+        BrokerService brokerService = pulsar.getBrokerService();
+        Field pulsarStatsField = BrokerService.class.getDeclaredField("pulsarStats");
+        pulsarStatsField.setAccessible(true);
+        PulsarStats pulsarStats = brokerService.getPulsarStats();
```
BewareMyPower, line 76: Please avoid using reflection as much as possible in tests. It makes refactoring very hard. My experience refactoring some code of Pulsar is really terrible due to the reflection everywhere.

Actually, there is already a getter to access `pulsarStats` because `BrokerService` is annotated with `@Getter`.

BewareMyPower (changes requested):

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionPersistentTopicTest.java
dao-jun: +1

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
```diff
@@ -54,10 +56,20 @@ public SingleSnapshotAbortedTxnProcessorImpl(PersistentTopic topic) {
                 .getTransactionBufferSnapshotServiceFactory()
                 .getTxnBufferSnapshotService().getReferenceWriter(TopicName.get(topic.getName()).getNamespaceObject());
         this.takeSnapshotWriter.getFuture().exceptionally((ex) -> {
-                    log.error("{} Failed to create snapshot writer", topic.getName());
-                    topic.close();
-                    return null;
-                });
+            log.error("{} Failed to create snapshot writer", topic.getName());
```
BewareMyPower, line 59: It would be better to add `ex` to the log message here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
```diff
@@ -54,10 +56,20 @@ public SingleSnapshotAbortedTxnProcessorImpl(PersistentTopic topic) {
...
+            // Don't directly use the topic object to close, because the topicFuture might not
+            // be completed at that time, which could leave closed topics in the cache(at BrokerService).
+            CompletableFuture<Optional<Topic>> topicFuture = topic.getBrokerService().getTopics().get(topic.getName());
+            if (topicFuture != null) {
+                topicFuture.thenAccept(t -> t.ifPresent(v -> {
```
BewareMyPower, line 64: I'm not sure why the value of `topics` is an `Optional`. I just have a question that when could the value be `Optional.empty()`? If so, an empty `Optional` will be cached and never removed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
```diff
@@ -54,10 +56,20 @@ public SingleSnapshotAbortedTxnProcessorImpl(PersistentTopic topic) {
...
+                    v.close(true).exceptionally(ec -> {
+                        log.error("Close topic {} exception", v.getName(), ec);
```
BewareMyPower, line 66: If `close` failed, the topic will still be cached. Should you call `BrokerService#removeTopicFromCache` in this case?

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionPersistentTopicTest.java
BewareMyPower: I see you want to delay the `openLedgerComplete`  method by sleeping 1 second in `recordTopicLoadTimeValue` so that `topicFuture.complete` will happen after the transaction failure.

But I think you can customize the `TopicFactory` to achieve the same goal. You can override the `create` method to create a `PersistentTopic`'s subclass whose `initialize()` method will be delayed for some time

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TopicTransactionBuffer.java
```diff
@@ -129,6 +131,23 @@ public TopicTransactionBuffer(PersistentTopic topic) {
...
+        this.takeSnapshotIntervalNumber = topic.getBrokerService().getPulsar()
+                .getConfiguration().getTransactionBufferSnapshotMaxTransactionCount();
+        this.takeSnapshotIntervalTime = topic.getBrokerService().getPulsar()
+                .getConfiguration().getTransactionBufferSnapshotMinTimeInMillis();
+        this.maxReadPosition = topic.getManagedLedger().getLastConfirmedEntry();
+        this.snapshotAbortedTxnProcessor = snapshotAbortedTxnProcessor;
+        this.snapshotType = snapshotType;
+        this.maxReadPositionCallBack = topic.getMaxReadPositionCallBack();
+        this.recover();
+    }
```
BewareMyPower, line 149: This new constructor overload has much duplicated code with the original constructor. Please reuse the code. It  only just reduces duplicated code, but also simulate the real case more likely.

Here is a bad example. Assuming there is a `Demo` class whose constructor is:

```java
public Demo(Argument arg) {
    this.field1 = arg.f();
    this.field2 = arg.g();
}
```

If you wanted to pass a mocked `field2` and added a new constructor;

```java
Demo(Argument arg, Field2 field2) {
    this.field1 = arg.h();
    this.field2 = field2;
}
```

Then the new constructor will be meaningless. Because now `field` is not `arg.f()` anymore.

BewareMyPower (changes requested):

shibd:
hi, @lhotari @BewareMyPower @dao-jun Thanks for details review, 

After discussing with @poorbarcode , we've decided to pass `createFuture` into the `topic` and then wrap the `close` method. I'll push a new version later.

poorbarcode (approved):

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionPersistentTopicTest.java
```diff
@@ -0,0 +1,148 @@
...
+        super.internalCleanup();
+    }
+
+    @Test
+    public void testNoOrphanClosedTopicIfTxnInternalFailed() {
+        String tpName = BrokerTestUtil.newUniqueName("persistent://public/default/tp2");
+
+        BrokerService brokerService = pulsar.getBrokerService();
+
+        // 2. Mock close topic when create transactionBuffer
```
RobertIndie, line 77:
```suggestion
        // 1. Mock close topic when create transactionBuffer
```

Seems the `1. `is gone.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TopicTransactionBuffer.java
```diff
@@ -118,13 +136,8 @@ public TopicTransactionBuffer(PersistentTopic topic) {
...
         this.maxReadPosition = topic.getManagedLedger().getLastConfirmedEntry();
-        if (topic.getBrokerService().getPulsar().getConfiguration().isTransactionBufferSegmentedSnapshotEnabled()) {
-            snapshotAbortedTxnProcessor = new SnapshotSegmentAbortedTxnProcessorImpl(topic);
-            snapshotType = AbortedTxnProcessor.SnapshotType.Segment;
-        } else {
-            snapshotAbortedTxnProcessor = new SingleSnapshotAbortedTxnProcessorImpl(topic);
-            snapshotType = AbortedTxnProcessor.SnapshotType.Single;
-        }
+        this.snapshotAbortedTxnProcessor = snapshotAbortedTxnProcessor;
+        this.snapshotType = snapshotType;
```
RobertIndie, line 140: Why do we make this change?

RobertIndie (commented):
Overall looks good to me. Thanks.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TopicTransactionBuffer.java
shibd: To passed mocked snapshotAbortedTxnProcessor object in unit test

BewareMyPower (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
```diff
@@ -95,6 +96,9 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
     protected static final long POLICY_UPDATE_FAILURE_RETRY_TIME_SECONDS = 60;
 
     protected final String topic;
+    @Getter
+    @Setter
+    protected volatile CompletableFuture<Optional<Topic>> createFuture;
```
lhotari, line 101: It would be useful to document the purpose of this `createFuture`. Without documentation, it could be hard to understand the reason why this field exists.
Please add javadocs. It could be useful to use ordinary getter and setter methods so that javadocs could be added.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2409,42 +2411,7 @@ public AuthorizationService getAuthorizationService() {
...
-        if (topicInCache == null || topicInCache != topic){
-            return Optional.empty();
-        } else {
-            return Optional.of(createTopicFuture);
-        }
-    }
-
-    private CompletableFuture<Void> removeTopicFutureFromCache(String topic,
+    public CompletableFuture<Void> removeTopicFutureFromCache(String topic,
                                                         CompletableFuture<Optional<Topic>> createTopicFuture) {
```
lhotari, line 2415: Please add javadoc to this method.
Instead of passing `String topic, CompletableFuture<Optional<Topic>> createTopicFuture`, arguments to this method, I think that it could be a single `AbstractTopic topic` parameter since AbstractTopic contains the getter for createTopicFuture. This would properly encapsulate the implementation details.
The name of the method could be reconsidered. For example `removeTopicFromCache` and documenting in the javadoc that it removes a specific instance from the cache by cache key which is the topic future.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
lhotari: I wonder if it would make more sense to call this field `cacheKey`?

lhotari (changes requested):
Please check the review comments.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
shibd: Good point, changed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
shibd: Thanks for reviewing and adding documentation. I believe keeping the name as createFuture makes it clearer.

lhotari (changes requested):
Please check comment about the comment.

lhotari (approved):
LGTM

PR #23886, opened 2025-01-23
[fix][client] Fix LoadManagerReport not found

nodece:
### Motivation

When using the `pulsar-client-all` and `pulsar-client-admin` libraries, the `admin.brokerStats().getLoadReport()` doesn't work:

```
DEBUG o.a.p.c.util.ObjectMapperFactory - Add LoadManagerReport deserializer failed because LoadManagerReport.class has been shaded
java.lang.ClassNotFoundException: org.apache.pulsar.shade.org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport
2025-01-23 16:15:16.816 [taskScheduler-9] DEBUG o.a.p.c.util.ObjectMapperFactory - Add LoadManagerReport deserializer failed because LoadManagerReport.class has been shaded
java.lang.ClassNotFoundException: org.apache.pulsar.shade.org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport
2025-01-23 16:15:28.232 [http-nio-10000-exec-1] DEBUG o.s.w.s.m.m.a.ExceptionHandlerExceptionResolver - Resolved [org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.AbstractMethodError: Receiver class org.apache.pulsar.client.admin.internal.BrokerStatsImpl does not define or inherit an implementation of the resolved method 'abstract org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport getLoadReport()' of interface org.apache.pulsar.client.admin.BrokerStats.]
```

We got this error because the shade plugin relocates the `org.apache.pulsar.policies.data.loadbalancer.*` package, and `org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport` didn't added to the shaded-jar. 

### Modifications

- Relocating the `org.apache.pulsar.policies` is meaningless, so remove that from the shade configuration.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-admin-shaded/pom.xml
+++ b/pulsar-client-admin-shaded/pom.xml
@@ -367,6 +367,13 @@
                 <relocation>
                   <pattern>org.apache.pulsar.policies</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.policies</shadedPattern>
+                  <!-- exclude references to unshaded classes and interfaces in https://github.com/apache/pulsar/tree/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer -->
+                  <excludes>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.NamespaceBundleStats</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.ResourceUsage</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.ServiceLookupData</exclude>
+                  </excludes>
                 </relocation>
                 <relocation>
                   <pattern>org.asynchttpclient</pattern>
--- a/pulsar-client-all/pom.xml
+++ b/pulsar-client-all/pom.xml
@@ -427,6 +427,13 @@
                 <relocation>
                   <pattern>org.apache.pulsar.policies</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.policies</shadedPattern>
+                  <!-- exclude references to unshaded classes and interfaces in https://github.com/apache/pulsar/tree/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer -->
+                  <excludes>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.NamespaceBundleStats</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.ResourceUsage</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.ServiceLookupData</exclude>
+                  </excludes>
                 </relocation>
                 <relocation>
                   <pattern>org.asynchttpclient</pattern>
--- a/pulsar-client-shaded/pom.xml
+++ b/pulsar-client-shaded/pom.xml
@@ -339,6 +339,13 @@
                 <relocation>
                   <pattern>org.apache.pulsar.policies</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.policies</shadedPattern>
+                  <!-- exclude references to unshaded classes and interfaces in https://github.com/apache/pulsar/tree/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer -->
+                  <excludes>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.NamespaceBundleStats</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.ResourceUsage</exclude>
+                    <exclude>org.apache.pulsar.policies.data.loadbalancer.ServiceLookupData</exclude>
+                  </excludes>
                 </relocation>
                 <relocation>
                   <pattern>org.asynchttpclient</pattern>
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/ObjectMapperFactory.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/ObjectMapperFactory.java
@@ -32,7 +32,6 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.fasterxml.jackson.module.paramnames.ParameterNamesModule;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.ClassUtils;
 import org.apache.pulsar.client.admin.internal.data.AuthPoliciesImpl;
 import org.apache.pulsar.common.functions.FunctionConfig;
 import org.apache.pulsar.common.functions.FunctionState;
@@ -263,15 +262,7 @@ public class ObjectMapperFactory {
         mapper.addMixIn(FunctionState.class, JsonIgnorePropertiesMixIn.class);
         mapper.addMixIn(Metrics.class, MetricsMixIn.class);
 
-        try {
-            // We look for LoadManagerReport first, then add deserializer to the module
-            // With shaded client, org.apache.pulsar.policies is relocated to
-            // org.apache.pulsar.shade.org.apache.pulsar.policies
-            ClassUtils.getClass("org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport");
-            module.addDeserializer(LoadManagerReport.class, new LoadReportDeserializer());
-        } catch (ClassNotFoundException e) {
-            log.debug("Add LoadManagerReport deserializer failed because LoadManagerReport.class has been shaded", e);
-        }
+        module.addDeserializer(LoadManagerReport.class, new LoadReportDeserializer());
 
         module.setAbstractTypes(resolver);
 
--- a/tests/pulsar-client-admin-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
+++ b/tests/pulsar-client-admin-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
@@ -87,6 +87,7 @@ public class SmokeTest extends TestRetrySupport {
         expectedNamespacesList.add("public/default");
         expectedNamespacesList.add("public/functions");
         Assert.assertEquals(admin.namespaces().getNamespaces("public"), expectedNamespacesList);
+        admin.brokerStats().getLoadReport();
     }
 
     @Override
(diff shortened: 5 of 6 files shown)


pulsar-common/src/main/java/org/apache/pulsar/common/util/ObjectMapperFactory.java
```diff
@@ -263,15 +262,7 @@ private static void setAnnotationsModule(ObjectMapper mapper) {
         mapper.addMixIn(FunctionState.class, JsonIgnorePropertiesMixIn.class);
         mapper.addMixIn(Metrics.class, MetricsMixIn.class);
 
-        try {
-            // We look for LoadManagerReport first, then add deserializer to the module
-            // With shaded client, org.apache.pulsar.policies is relocated to
-            // org.apache.pulsar.shade.org.apache.pulsar.policies
-            ClassUtils.getClass("org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport");
```
lhotari, line 270: just wondering if a solution would work where this line `ClassUtils.getClass("org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport")` is removed. I wonder why this is present?

It seems strange that there would need to be separate handling for the `org.apache.pulsar.policies.data.loadbalancer` package. It wouldn't be consistent with the remaining parts of the shading solution.

tests/pulsar-client-admin-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
```diff
@@ -87,6 +87,7 @@ public void checkAdmin() throws PulsarClientException, PulsarAdminException {
         expectedNamespacesList.add("public/default");
         expectedNamespacesList.add("public/functions");
         Assert.assertEquals(admin.namespaces().getNamespaces("public"), expectedNamespacesList);
+        admin.brokerStats().getLoadReport();
```
lhotari, line 90: would it be possible to also validate that there's valid content?

pulsar-common/src/main/java/org/apache/pulsar/common/util/ObjectMapperFactory.java
nodece: > just wondering if a solution would work where this line `ClassUtils.getClass("org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport")` is removed. 

Yes. It works fine.

> I wonder why this is present?

Usually, we don't shade the API package.

However, the `org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport` exists in the `pulsar-client-admin-api` module, the shade plugin didn't ignore that.

> It seems strange that there would need to be separate handling for the `org.apache.pulsar.policies.data.loadbalancer` package. It wouldn't be consistent with the remaining parts of the shading solution.

It looks confusing, I guess that is because of the package name, and this is a special case, we must take care of that.

tests/pulsar-client-admin-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
nodece: I have added not null validation.

tests/pulsar-client-all-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
```diff
@@ -87,6 +87,7 @@ public void checkAdmin() throws PulsarClientException, PulsarAdminException {
         expectedNamespacesList.add("public/default");
         expectedNamespacesList.add("public/functions");
         Assert.assertEquals(admin.namespaces().getNamespaces("public"), expectedNamespacesList);
+        admin.brokerStats().getLoadReport();
```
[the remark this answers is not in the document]
nodece: I have added not null validation.

dao-jun (approved):

tests/pulsar-client-admin-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
nodece: Because the standalone uses `org.apache.pulsar.broker. load-balance.NoopLoadManager`, which returns null, we just check that no exception is thrown.

tests/pulsar-client-all-shade-test/src/test/java/org/apache/pulsar/tests/integration/SmokeTest.java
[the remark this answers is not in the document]
nodece: Because the standalone uses `org.apache.pulsar.broker. load-balance.NoopLoadManager`, which returns null, we just check that no exception is thrown.

pulsar-client-shaded/pom.xml
```diff
@@ -336,10 +336,6 @@
                   <pattern>org.apache.commons</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.apache.commons</shadedPattern>
                 </relocation>
-                <relocation>
-                  <pattern>org.apache.pulsar.policies</pattern>
-                  <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.policies</shadedPattern>
-                </relocation>
```
lhotari, line 342: this change could be removed with the problem is correctly addressed in LoadReportDeserializer

pulsar-client-admin-shaded/pom.xml
```diff
@@ -364,10 +364,6 @@
                   <pattern>org.apache.pulsar.checksum</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.checksum</shadedPattern>
                 </relocation>
-                <relocation>
-                  <pattern>org.apache.pulsar.policies</pattern>
-                  <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.policies</shadedPattern>
-                </relocation>
```
lhotari, line 370: this change could be removed with the problem is correctly addressed in LoadReportDeserializer

pulsar-client-all/pom.xml
```diff
@@ -424,10 +424,6 @@
                   <pattern>org.apache.commons</pattern>
                   <shadedPattern>org.apache.pulsar.shade.org.apache.commons</shadedPattern>
                 </relocation>
-                <relocation>
-                  <pattern>org.apache.pulsar.policies</pattern>
-                  <shadedPattern>org.apache.pulsar.shade.org.apache.pulsar.policies</shadedPattern>
-                </relocation>
```
lhotari, line 430: this change could be removed with the problem is correctly addressed in LoadReportDeserializer

nodece:
> It looks like that shading changes should be reverted and there should be logic in LoadReportDeserializer to handle the unshaded classname that comes from the broker:

Both the interface (`org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport`) and its implementation are being shaded, but the interface class isn’t being properly added to the shaded package. This mismatch is what’s causing the class resolution issue when using the shaded jar.

To resolve this, we should ensure that the interface is also included in the shaded package, keeping it consistent with the implementation. Alternatively, as a simpler solution, we could consider removing that shading the `org.apache.pulsar.policies`, which would eliminate these classpath problems.

@lhotari What's your opinion?

nodece:
BTW, it works fine in our private pulsar. We are using the `pulsar-client-all` to feat the broker resource(broker-stats/topic/namespace/subscript/so on) info.

lhotari:
> @lhotari What's your opinion?

@nodece I've suggested a fix in https://github.com/apache/pulsar/pull/23886#pullrequestreview-2572592840 . Have you checked that?

lhotari:
> Both the interface (`org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport`) and its implementation are being shaded, but the interface class isn’t being properly added to the shaded package. This mismatch is what’s causing the class resolution issue when using the shaded jar.
> 
> To resolve this, we should ensure that the interface is also included in the shaded package, keeping it consistent with the implementation. Alternatively, as a simpler solution, we could consider removing that shading the `org.apache.pulsar.policies`, which would eliminate these classpath problems.

Public interfaces shouldn't be shaded. Implementation classes should be shaded. That's the original intent.
The mistake has been to use the classname in org.apache.pulsar.policies.data.loadbalancer.LoadReport#loadReportType field. Shading will change that to be the shaded class name and as a result of this, the deserialization will fail since there's a difference between client side and broker side for the value of the `loadReportType` field. This problem is addressed by the changes that I suggested in https://github.com/apache/pulsar/pull/23886#pullrequestreview-2572592840.

lhotari:
The interface classes present here should be excluded from shading: https://github.com/apache/pulsar/tree/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer

pulsar-client-admin-shaded/pom.xml
lhotari: Instead of removing the relocation rule, add a exclude rule so that the classes and interfaces in https://github.com/apache/pulsar/tree/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer aren't shaded:
```xml
  <excludes>
    <exclude>org\.apache\.pulsar\.policies\.data\.loadbalancer\.(LoadManagerReport|NamespaceBundleStats|ResourceUsage|ServiceLookupData)($.*)?</exclude>
  </excludes>
```

pulsar-client-all/pom.xml
lhotari: Add the excludes rule described earlier.

pulsar-client-shaded/pom.xml
lhotari: Add the excludes rule described earlier.

lhotari (changes requested):
I added some more details about how to resolve the problem.

nodece:
> > @lhotari What's your opinion?
> 
> @nodece I've suggested a fix in [#23886 (review)](https://github.com/apache/pulsar/pull/23886#pullrequestreview-2572592840) . Have you checked that?

This is an error fix, I still got the same error.

`org.apache.pulsar.shade.org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport` does not exist.

> Public interfaces shouldn't be shaded. Implementation classes should be shaded.

I understand that this is the correct way. Your changes are correct.

> Shading will change that to be the shaded class name and as a result of this, the deserialization will fail since there's a difference between client side and broker side for the value of the `loadReportType` field. 

```
    public static final String loadReportType = LoadReport.class.getSimpleName();
```

The `loadReportType` value is a simple name, so we don't need to consider the full name, and this means `LoadReportDeserializer` should not modified.


```
  <excludes>
    <!-- exclude references to unshaded classes and interfaces in https://github.com/apache/pulsar/tree/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer -->   
    <exclude>org\.apache\.pulsar\.policies\.data\.loadbalancer\.(LoadManagerReport|NamespaceBundleStats|ResourceUsage|ServiceLookupData)($.*)?</exclude>
  </excludes>
```

It doesn't work, so I use the following code instead of that:
```
<excludes>
  <exclude>org.apache.pulsar.policies.data.loadbalancer.LoadManagerReport</exclude>
  <exclude>org.apache.pulsar.policies.data.loadbalancer.NamespaceBundleStats</exclude>
  <exclude>org.apache.pulsar.policies.data.loadbalancer.ResourceUsage</exclude>
  <exclude>org.apache.pulsar.policies.data.loadbalancer.ServiceLookupData</exclude>
</excludes>
```

@lhotari Thank you!

pulsar-client-admin-shaded/pom.xml
nodece: Done.

pulsar-client-shaded/pom.xml
nodece: Done.

lhotari (approved):
LGTM

PR #23892, opened 2025-01-24
[improve][test] Support decorating topic, subscription, dispatcher, ManagedLedger and ManagedCursors instances in tests

lhotari:
### Motivation

In some Pulsar tests, there's a need to inject special behaviors such as delays, race conditions or failures.
The current solutions require using reflection and replacing instances, which results in hard to maintain test code.

### Modifications

- Add a solution for Pulsar broker tests that allows to decorate persistent topics, subscriptions, dispatchers
managed ledger factory, managed ledger and managed cursor instances.
- Refactor production code so that there are hooks for intercepting the instances in subclasses
- Add a helper methods for creating a Mockito Spy decorator
  - for a dispatcher instance
  - for a ManagedCursorImpl instance

example usage for spying on a dispatcher and cursor instance:

```java
    @Override
    protected void doInitConf() throws Exception {
        super.doInitConf();
        BrokerTestInterceptor.INSTANCE.configure(conf);
    }

    @AfterMethod(alwaysRun = true)
    protected void resetInterceptors() throws Exception {
        BrokerTestInterceptor.INSTANCE.reset();
    }

    @Test
    public void testSomething() throws Exception {
        BrokerTestInterceptor.INSTANCE.applyDispatcherSpyDecorator(PersistentDispatcherMultipleConsumers.class,
                spy -> {
                    doAnswer(invocation -> {
                        List<Entry> entries = invocation.getArgument(0);
                        log.info("intercepted readEntriesComplete with {} entries", entries.size());
                        return invocation.callRealMethod();
                    }).when(spy).readEntriesComplete(any(), any());
                }
        );

        BrokerTestInterceptor.INSTANCE.applyCursorSpyDecorator(spy -> {
            doAnswer(invocation -> {
                log.info("intercepted asyncReadEntriesWithSkipOrWait");
                return invocation.callRealMethod();
            }).when(spy).asyncReadEntriesWithSkipOrWait(anyInt(), anyLong(), any(), any(), any(), any());
            doAnswer(invocation -> {
                log.info("intercepted asyncReplayEntries");
                return invocation.callRealMethod();
            }).when(spy).asyncReplayEntries(any(), any(), any(), anyBoolean());
        });
   ...
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23894, opened 2025-01-24
[fix][broker] Apply dispatcherMaxReadSizeBytes also for replay reads for Shared and Key_Shared subscriptions

### LINKED ISSUE #23505: [Bug] Redelivering messages doesn't take dispatcherMaxReadSizeBytes into account in Shared and Key_Shared subscriptions

lhotari:
Fixes #23505

### Motivation

See #23505. It is expected that dispatcherMaxReadSizeBytes also limits replay queue reads. Replay reads are currently unbounded in byte size, which is unexpected behavior.

### Modifications

- use `org.apache.bookkeeper.mledger.ManagedCursor#applyMaxSizeCap` also in replay queue reads
- add test
  - test uses #23892. That PR should be merged first to clean up the diff of this PR.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -384,7 +384,7 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
             }
 
             Set<Position> messagesToReplayNow =
-                    canReplayMessages() ? getMessagesToReplayNow(messagesToRead) : Collections.emptySet();
+                    canReplayMessages() ? getMessagesToReplayNow(messagesToRead, bytesToRead) : Collections.emptySet();
             if (!messagesToReplayNow.isEmpty()) {
                 if (log.isDebugEnabled()) {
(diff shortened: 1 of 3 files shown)


lhotari:
@heesung-sn This PR has been rebased now. Please review

heesung-sn (approved):

Issue #23889, opened 2025-01-24
Metadata information is not cleaned when broker exits abnormally


Joforde:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

When a broker starts, it registers its metadata with the metadata service (such as Zookeeper or ETCD) under the `/loadbalance/brokers` directory.  When the broker exits gracefully, it actively calls the unregister method to remove its own metadata.
https://github.com/apache/pulsar/blob/66d1bb0d734f12d758b0f0e9e3c0b42543508f8d/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java#L134-L143

https://github.com/apache/pulsar/blob/66d1bb0d734f12d758b0f0e9e3c0b42543508f8d/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java#L172-L178

However, if the broker is forced to exit due to issues like hardware failure, network problems, or being terminated with `kill -9`, it does not call the unregister method to delete its metadata.  This results in the metadata for brokers that are no longer accessible remaining in the system.

Currently, Pulsar retrieves all active brokers by fetching all child nodes under the metadata service's `/loadbalance/brokers` path.  This can lead to offline brokers being considered active, and bundles may be assigned to brokers that are not accessible, causing new namespace clients to experience read and write failures.
https://github.com/apache/pulsar/blob/66d1bb0d734f12d758b0f0e9e3c0b42543508f8d/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/BrokerRegistryImpl.java#L197-L201

### Solution

I am not yet certain how to resolve this issue. I am currently reviewing the relevant code and drafting a solution document. If you have any suggestions, please feel free to leave a comment.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

lhotari:
> However, if the broker is forced to exit due to issues like hardware failure, network problems, or being terminated with `kill -9`, it does not call the unregister method to delete its metadata. This results in the metadata for brokers that are no longer accessible remaining in the system.

The metadata should get deleted automatically when the session expires. When a broker crashes, a new broker instance would not be able to start until the session expires and the empheral metadata entries expire.   

There's `EnumSet.of(CreateOption.Ephemeral)`:
[...]

PR #13066, opened 2021-12-01
[Broker] Fix LeaderElectionService.getCurrentLeader and add support for empheralOwner in MockZooKeeper

lhotari:
### Motivation

`LeaderElectionService.getCurrentLeader()` always returned `Optional.empty()` on all other than the leader broker.
This has several severe consequences in the Pulsar Broker code base.

When a lookup is made to a topic that isn't currently loaded, the decision will be made in a distributed fashion on the follower brokers since the information about the leader broker is missing (because `LeaderElectionService.getCurrentLeader()` always returned `Optional.empty()`). This leads to races when assigning the topic ownership to a broker since the decision isn't made centrally on the leader broker.
I'll provide a separate PR for improving and fixing the topic ownership assignment. That PR is #13069 and it contains a failing test case for the topic ownership assignment / lookup problem.

### Modifications

- Add support for empheralOwner in MockZooKeeper
- Add MultiBrokerBaseTest base test class for writing tests that involve multiple brokers
- Add test case that fails without the fix
- Make the fix to LeaderElectionImpl

PR #23902, opened 2025-01-27
[fix][metadata] fixed ephemeral zk put

heesung-sn:
Fixes https://github.com/apache/pulsar/issues/23889

### Motivation

Fixes https://github.com/apache/pulsar/issues/23889
zk.put creates persistent znode although it passes ephemeral node creation option.

### Modifications

- pass the creation option when set fails(if the node exists already)
- fix zk stat ephemeral node check
> ephemeralOwner The session id of the owner of this znode if the znode is an ephemeral node. If it is not an ephemeral node, it will be zero.

ref : https://zookeeper.apache.org/doc/r3.5.5/zookeeperProgrammers.html

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
@@ -439,8 +439,8 @@ public class ZKMetadataStore extends AbstractBatchedMetadataStore
                                 future.completeExceptionally(getException(Code.BADVERSION, opPut.getPath()));
                             } else {
                                 // The z-node does not exist, let's create it first
-                                put(opPut.getPath(), opPut.getData(), Optional.of(-1L)).thenAccept(
-                                                s -> future.complete(s))
+                                put(opPut.getPath(), opPut.getData(), Optional.of(-1L), opPut.getOptions())
+                                        .thenAccept(s -> future.complete(s))
                                         .exceptionally(ex -> {
                                             if (ex.getCause() instanceof BadVersionException) {
                                                 // The z-node exist now, let's overwrite it
@@ -478,7 +478,7 @@ public class ZKMetadataStore extends AbstractBatchedMetadataStore
 
     private Stat getStat(String path, org.apache.zookeeper.data.Stat zkStat) {
         return new Stat(path, zkStat.getVersion(), zkStat.getCtime(), zkStat.getMtime(),
-                zkStat.getEphemeralOwner() != -1,
+                zkStat.getEphemeralOwner() != 0,
                 zkStat.getEphemeralOwner() == zkc.getSessionId());
     }
 
--- a/pulsar-metadata/src/test/java/org/apache/pulsar/metadata/MetadataStoreExtendedTest.java
+++ b/pulsar-metadata/src/test/java/org/apache/pulsar/metadata/MetadataStoreExtendedTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.pulsar.metadata;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotEquals;
 import static org.testng.Assert.assertNotNull;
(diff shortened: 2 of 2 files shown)


lhotari (approved):
LGTM, great work @heesung-sn

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/ResourceLockImpl.java
```diff
@@ -188,6 +188,17 @@ private CompletableFuture<Void> acquireWithNoRevalidation(T newValue) {
         CompletableFuture<Void> result = new CompletableFuture<>();
         store.put(path, payload, Optional.of(version), EnumSet.of(CreateOption.Ephemeral))
                 .thenAccept(stat -> {
+                    if (!stat.isEphemeral()) {
```
heesung-sn, line 191: I hit this error from `testLookupConnectionNotCloseIfFailedToAcquireOwnershipOfBundle`, where the lock suddenly became persistent, after invalidating the cache and updating it with null.

This is a bit surprising to me that the ephemeral lock can suddenly become persistent(afaik, this is from the zk server behavior, not from zk client). I couldn't find enough zk doc/references to explain this behavior.

```java
        cache.invalidateLocalOwnerCache();
        final var lock = pulsar.getCoordinationService().getLockManager(NamespaceEphemeralData.class)
                .acquireLock(ServiceUnitUtils.path(bundle), new NamespaceEphemeralData()).join();
        lock.updateValue(null);
```

lhotari: Rings a bell, found #13066. There's probably more changes required to implement it properly in MockZooKeeper.

lhotari: There's https://github.com/apache/pulsar/blob/master/pulsar-broker/src/test/java/org/apache/pulsar/broker/MultiBrokerTestZKBaseTest.java as a way how it's currently possible to use a real Zookeeper with tests. However, adding direct support to PulsarTestContext and making it easy to override some protected method in MockedPulsarServiceBaseTest to choose it would be more flexible.

lhotari: @heesung-sn I pushed some changes to this PR for MockZooKeeperSession/MockZooKeeper to support ephemeral owner. The value for persistent nodes is now 0. Support was missing completely for multi-ops. In the mock zookeeper solution, the session id gets passed in a thread local since there's not another way.

lhotari: 👆@heesung-sn do you have a chance to check what @Joforde suggested?

heesung-sn: I tried 2. and still see some test failures.

heesung-sn: Planning to get back to this work on Thursday this week.

lhotari: @heesung-sn I added `.withTestZookeeper` support to `PulsarTestContext` and an easy way to use this in tests, that's in commit 17ec37a. I modified BrokerServiceLookupTest to use it so that it runs for both. That's useful for comparing the results and catching problems when using real ZooKeeper. It seems that there's a bigger mess to fix.

lhotari:
@heesung-sn I'm closing this PR since this is superseded by #23984. I'll push a new PR with the remaining fixes, including fixes to MockZooKeeper to properly support ephemeral nodes and properly return ZK's stat information. That appears to be broken in MockZooKeeper with many other details.

lhotari:
PR to fix remaining issues: #23988

PR #22789, opened 2024-05-28
[fix][ml] Fix race conditions in RangeCache

lhotari:
### Motivation

The RangeCache class contains several race conditions which cause instability.

When one thread removes the entry and another one uses it, that will become a problem. 

The `cacheEvictionIntervalMs` setting is 10 ms by default. This results in the 
`RangeCache.evictLEntriesBeforeTimestamp` method getting called about 100 times per second.
The default expiration is `managedLedgerCacheEvictionTimeThresholdMillis` which is 1000 ms by default.

It's also possible that 2 threads remove the entry at the same time. `ManagedLedgerImpl.invalidateEntriesUpToSlowestReaderPosition` will result in calls to `RangeCache.removeRange` method. These calls happen independently of the `RangeCache.evictLEntriesBeforeTimestamp` calls so there's a chance for race conditions.

### Modifications

- Add checks to validate that the value hasn't already been released before it is removed.
  - add extra `.retain()` and `.release()` calls
  - add a way to validate that the value matches the key
    - this is to prevent a possible race condition where the entry has already been recycled and contains another valid entry which is for a different key
- Make all removals using the `remove(key, value)` method so that removals remove the correct value exactly once
  - Add a wrapper object for the value to ensure this so that exact identity match is used for the removal.
  - The wrapper instance is recycled to reduce the overhead.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #22892, opened 2024-06-11
[improve][misc] Upgrade to Netty 4.1.111.Final and switch to use grpc-netty-shaded

lhotari:
Fixes #22601
Fixes #21892
Fixes #19460

### Motivation

[Netty 4.1.111.Final](https://netty.io/news/2024/06/11/4-1-111-Final.html) contains important fixes:
- https://github.com/netty/netty/pull/14086
- https://github.com/netty/netty/pull/14093

These address Broker stability when TLS with Bookkeeper V2 protocol is used between Broker and Bookies.
This will also allow removing the extra buffer copies that have been in place in Pulsar broker and Pulsar client when TLS is enabled. 

https://github.com/apache/pulsar/blob/266243cae246a6fa52b4b6c626932885ad44cbf4/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/ByteBufPair.java#L138-L157

This will be removed in a separate PR.

### Modifications

- Upgrade Netty to 4.1.111.Final version

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23903, opened 2025-01-28
[fix][broker Fix bug in RangeCache where different instance of the key wouldn't ever match

### LINKED ISSUE #23900: [Bug] Redelivery messages (replay queue reads) don't get read from cache even when they exist in cache, causing extra bookkeeper reads
### Search before asking

### Read release policy

### Version

Since Pulsar 3.0.6, 3.2.4, 3.3.1, 4.0.0

### Minimal reproduce step

Replay queue reads cache cache misses in all cases, even when the entry is already available in the cache.
One detail of the current broker cache implementation is that it is necessary to configure `cacheEvictionByMarkDeletedPosition=true` so that cached entries don't get evicted by the read position in the managed ledger. (There's a separate issue #16421 about by-passing the cache when `cacheEvictionByMarkDeletedPosition=true` isn't used.)

Replay queue reads should get read from the cache.

### What did you see instead?

No caching

### Anything else?

This was previously reported as part of #23504. In addition, there's another issue that when reads are performed, the returned entries aren't added to the cache.

lhotari:
Fixes #23900

### Motivation

During work on PR #23901, a severe issue was discovered in the RangeCache implementation. Cache lookups consistently fail when the key used to look up an entry differs from the instance used to put the entry into the cache.

One detail of the current broker cache implementation is that it is necessary to configure `cacheEvictionByMarkDeletedPosition=true` so that cached entries don't get evicted by the read position in the managed ledger.

This problematic solution was introduced in PR #22789, which aimed to address race conditions in the RangeCache and prevent the use of already recycled object instances. The solution underwent subsequent refactoring in PRs #22814 and #22818. While the value wrapper solution ensures consistency between cached entries and their corresponding keys, the original design primarily focused on the `getRange` and `removeRange` methods.

In these method implementations, the ConcurrentSkipListMap's subMap method ensures the original key remains available, preventing the lookup problem from manifesting. This explains why the issue remained undetected for so long, as caching functions correctly in most scenarios.

The impact primarily affects single reads, such as initial readings of replay queue entries. Single reads have never added entries to the cache, a bug originally reported as part of #23504 and subsequently moved to #23900.

Issue #23900 represents a regression.
Prior to the changes introduced in PRs #22789, #22814, and #22818, replay queue entries might have been cached due to previous reads adding entries to the cache. However, the functionality to add replay queue reads to the cache has never been implemented.

### Modifications

- add unit test to verify that an entry can be looked up with a different key instance
- fix the issue in RangeCache implementation

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
@@ -297,6 +297,9 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
      * @return an pair of ints, containing the number of removed entries and the total size
      */
     public Pair<Integer, Long> removeRange(Key first, Key last, boolean lastInclusive) {
+        if (log.isDebugEnabled()) {
+            log.debug("Removing entries in range [{}, {}], lastInclusive: {}", first, last, lastInclusive);
+        }
         RemovalCounters counters = RemovalCounters.create();
         Map<Key, EntryWrapper<Key, Value>> subMap = entries.subMap(first, true, last, lastInclusive);
         for (Map.Entry<Key, EntryWrapper<Key, Value>> entry : subMap.entrySet()) {
@@ -404,6 +407,9 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
      * @return a pair containing the number of entries evicted and their total size
      */
     public Pair<Integer, Long> evictLeastAccessedEntries(long minSize) {
+        if (log.isDebugEnabled()) {
+            log.debug("Evicting entries to reach a minimum size of {}", minSize);
+        }
         checkArgument(minSize > 0);
         RemovalCounters counters = RemovalCounters.create();
         while (counters.removedSize < minSize && !Thread.currentThread().isInterrupted()) {
@@ -422,6 +428,9 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
     * @return the tota
     */
    public Pair<Integer, Long> evictLEntriesBeforeTimestamp(long maxTimestamp) {
+       if (log.isDebugEnabled()) {
+              log.debug("Evicting entries with timestamp <= {}", maxTimestamp);
+       }
        RemovalCounters counters = RemovalCounters.create();
        while (!Thread.currentThread().isInterrupted()) {
            Map.Entry<Key, EntryWrapper<Key, Value>> entry = entries.firstEntry();
@@ -453,6 +462,9 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
      * @return size of removed entries
      */
     public Pair<Integer, Long> clear() {
+        if (log.isDebugEnabled()) {
+            log.debug("Clearing the cache with {} entries and size {}", entries.size(), size.get());
+        }
         RemovalCounters counters = RemovalCounters.create();
         while (!Thread.currentThread().isInterrupted()) {
             Map.Entry<Key, EntryWrapper<Key, Value>> entry = entries.firstEntry();
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
@@ -106,6 +106,17 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
             return localKey;
         }
 
+        /**
+         * Get the value associated with the key. Returns null if the key does not match the key associated with the
+         * value.
+         *
+         * @param key                    the key to match
+         * @param requireSameKeyInstance when true, the matching will be restricted to exactly the same instance of the
+         *                               key as the one stored in the wrapper. This is used to avoid any races
+         *                               when retrieving or removing the entries from the cache when the key and value
+         *                               instances are available.
+         * @return the value associated with the key, or null if the key does not match
+         */
         V getValue(K key, boolean requireSameKeyInstance) {
             long stamp = lock.tryOptimisticRead();
             K localKey = this.key;
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
@@ -106,7 +106,7 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
             return localKey;
         }
 
-        V getValue(K key) {
+        V getValue(K key, boolean requireSameKeyInstance) {
             long stamp = lock.tryOptimisticRead();
             K localKey = this.key;
             V localValue = this.value;
@@ -116,7 +116,7 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
                 localValue = this.value;
                 lock.unlockRead(stamp);
             }
-            if (localKey != key) {
+            if (localKey != key && (requireSameKeyInstance || localKey == null || !localKey.equals(key))) {
                 return null;
             }
             return localValue;
@@ -236,14 +236,14 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
      * The caller is responsible for releasing the reference.
      */
     public Value get(Key key) {
-        return getValue(key, entries.get(key));
+        return getValue(key, entries.get(key), false);
     }
 
-    private  Value getValue(Key key, EntryWrapper<Key, Value> valueWrapper) {
+    private  Value getValue(Key key, EntryWrapper<Key, Value> valueWrapper, boolean requireSameKeyInstance) {
         if (valueWrapper == null) {
             return null;
         } else {
-            Value value = valueWrapper.getValue(key);
+            Value value = valueWrapper.getValue(key, requireSameKeyInstance);
             if (value == null) {
                 // the wrapper has been recycled and contains another key
                 return null;
@@ -280,7 +280,7 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
 
         // Return the values of the entries found in cache
         for (Map.Entry<Key, EntryWrapper<Key, Value>> entry : entries.subMap(first, true, last, true).entrySet()) {
-            Value value = getValue(entry.getKey(), entry.getValue());
+            Value value = getValue(entry.getKey(), entry.getValue(), true);
             if (value != null) {
                 values.add(value);
             }
@@ -320,7 +320,7 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
                                           boolean skipInvalid, Predicate<Value> removeCondition) {
         Key key = entry.getKey();
         EntryWrapper<Key, Value> entryWrapper = entry.getValue();
-        Value value = entryWrapper.getValue(key);
+        Value value = entryWrapper.getValue(key, true);
         if (value == null) {
             // the wrapper has already been recycled and contains another key
             if (!skipInvalid) {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
@@ -107,6 +107,27 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
         }
 
         /**
+         * Get the value associated with the key. Returns null if the key does not match the key.
+         *
+         * @param key the key to match
+         * @return the value associated with the key, or null if the value has already been recycled or the key does not
+         * match
+         */
+        V getValue(K key) {
+            return getValueInternal(key, false);
+        }
+
+        /**
+         * Get the value associated with the Map.Entry's key and value. Exact instance of the key is required to match.
+         * @param entry the entry which contains the key and {@link EntryWrapper} value to get the value from
+         * @return the value associated with the key, or null if the value has already been recycled or the key does not
+         * exactly match the same instance
+         */
+        static <K, V> V getValueMatchingMapEntry(Map.Entry<K, EntryWrapper<K, V>> entry) {
+            return entry.getValue().getValueInternal(entry.getKey(), true);
+        }
+
+        /**
          * Get the value associated with the key. Returns null if the key does not match the key associated with the
          * value.
          *
@@ -117,7 +138,7 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
          *                               instances are available.
          * @return the value associated with the key, or null if the key does not match
          */
-        V getValue(K key, boolean requireSameKeyInstance) {
+        private V getValueInternal(K key, boolean requireSameKeyInstance) {
             long stamp = lock.tryOptimisticRead();
             K localKey = this.key;
             V localValue = this.value;
@@ -127,6 +148,10 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
                 localValue = this.value;
                 lock.unlockRead(stamp);
             }
+            // check that the given key matches the key associated with the value in the entry
+            // this is used to detect if the entry has already been recycled and contains another key
+            // when requireSameKeyInstance is true, the key must be exactly the same instance as the one stored in the
+            // entry to match
             if (localKey != key && (requireSameKeyInstance || localKey == null || !localKey.equals(key))) {
                 return null;
             }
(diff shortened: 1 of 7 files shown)


lhotari:
For preventing future regressions like #23900, it would be necessary to add specific tests for that purpose, to ensure that replay queue reads don't cause cache misses in a tailing read scenario. The regression is specific to tailing reads.

In the current broker cache implementation, catchup reads will result in cache misses for replay queue reads. This behavior is not a regression. This has a performance impact at least on Key_Shared subscriptions where it's common that entries get added to the replay queue. It's a future improvement that I'm planning to address that there would be caching also for that type of scenario.

Since we don't have currently sufficient test coverage for various scenarios, regressions like this can happen.
The goal of recent test framework improvements such as #23875 and #23892 are to make it easier to write tests also for this type of scenarios.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -106,7 +106,7 @@ K getKey() {
             return localKey;
         }
 
-        V getValue(K key) {
+        V getValue(K key, boolean requireSameKeyInstance) {
```
eolivelli, line 109: what about adding a second method that is more lenient instead of increasing the complexity of the existing ?
this is called on the hot paths, I think that it is worth to add some code duplication

lhotari: Actually, the code becomes more complex when adding more methods. I've added javadoc to explain parameter value. I'll push the changes soon.

lhotari: @eolivelli I made a second attempt to refactor the code based on your review comment. PTAL. The RangeCache design is intentionally racy without any exclusive locks and since the values are reference counted recycled objects, it has resulted in this solution where consistency is checked with the solution that is fairly complex. 
After applying this solution, I haven't seen reports about bugs where the root cause was that there were races in RangeCache eviction and cache usage. Some of the complexity would be reduced if we wouldn't recycle the EntryWrapper instances in RangeCache. However that comes with a tradeoff regarding GC. Extensively testing performance could shade more light in actual differences in practice.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -106,7 +106,39 @@ K getKey() {
...
+         * value.
+         *
+         * @param key                    the key to match
+         * @param requireSameKeyInstance when true, the matching will be restricted to exactly the same instance of the
+         *                               key as the one stored in the wrapper. This is used to avoid any races
+         *                               when retrieving or removing the entries from the cache when the key and value
+         *                               instances are available.
+         * @return the value associated with the key, or null if the key does not match
+         */
+        private V getValueInternal(K key, boolean requireSameKeyInstance) {
```
heesung-sn, line 141: General Question:

Object pooling can introduce complexity, such as synchronization overhead, which can negate performance gains.

How do we know if this object pooling is required here?

lhotari: > Object pooling can introduce complexity, such as synchronization overhead, which can negate performance gains.

@heesung-sn There's no synchronization that is used in this class. Netty's object pooling solution is very well optimized. I have commented on some of the tradeoffs in this comment: https://github.com/apache/pulsar/pull/23903#discussion_r1932599350

> How do we know if this object pooling is required here?

"required" can be understood in many ways. In Pulsar, we do use object pooling extensively. It's a chosen style in the Pulsar code base. In some cases, it's not useful. In some cases, it's useful. The only way to be sure would be to put the effort to benchmark and compare differences in performance. Removing object pooling isn't the scope of this PR.

With concurrency issues, there's always the doubt whether something is necessary or not. The changes made in PRs #22789, #22814 and #22818 seem to have eliminated "use after release" type of bugs that we were facing in Pulsar. Some of the issues were addressed with bug fixes to Netty in 4.1.111.Final (mentioned in #22892) and my assumption is that other "use after release" issues were addressed by the RangeCache changes.

I hope we can continue reviewing this PR and address the bug that this PR fixes. There are tests that reproduce the issue at different levels.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -106,7 +106,39 @@ K getKey() {
...
+        }
+
+        /**
+         * Get the value associated with the Map.Entry's key and value. Exact instance of the key is required to match.
+         * @param entry the entry which contains the key and {@link EntryWrapper} value to get the value from
+         * @return the value associated with the key, or null if the value has already been recycled or the key does not
+         * exactly match the same instance
+         */
+        static <K, V> V getValueMatchingMapEntry(Map.Entry<K, EntryWrapper<K, V>> entry) {
+            return entry.getValue().getValueInternal(entry.getKey(), true);
```
heesung-sn, line 127: any NPE from entry.getValue()?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -116,7 +148,11 @@ V getValue(K key) {
                 localValue = this.value;
                 lock.unlockRead(stamp);
             }
-            if (localKey != key) {
+            // check that the given key matches the key associated with the value in the entry
```
heesung-sn, line 151: What happens when this.key and this.value are recycled and have new key/value after this line? Are we ok to return the old value here?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -106,7 +106,39 @@ K getKey() {
...
 
+        /**
+         * Get the value associated with the key. Returns null if the key does not match the key.
+         *
+         * @param key the key to match
+         * @return the value associated with the key, or null if the value has already been recycled or the key does not
+         * match
+         */
         V getValue(K key) {
+            return getValueInternal(key, false);
```
heesung-sn, line 117: don't we always need to get value with requireSameKeyInstance=true?
In what case, are we ok with requireSameKeyInstance=false here?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
heesung-sn: Don' we need to wrap `void recycle() ` with the write lock?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -116,7 +148,11 @@ V getValue(K key) {
                 localValue = this.value;
                 lock.unlockRead(stamp);
```
heesung-sn, line 149: don't we need to release this in try-finally block?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
heesung-sn: Don' we need to wrap `void recycle() ` with the write lock?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
lhotari: No. Essentially this single change is addressing the bug fix. 
Getting by key should also match by `.equals` check.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
lhotari: No, since the value shouldn't be null. If it would be null, it would be detected since there would be an exception. Suppressing such issues would be a wrong solution.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
lhotari: > What happens when this wrapper is recycled and has a new value for the same key, after this line? Are we ok to return the old value here?

The value would be returned from this method. There's additional logic in the now extracted `getRetainedValueMatchingKey` method. The method contains verbose comments explaining the way that it's ensured that a recycled value is never returned from the public methods of RangeCache.

lhotari: > Don' we need to wrap void recycle() with the write lock?

That's not necessary since the solution already prevents race reads. In Java on 64 bit JVMs, there's no data corruption problems with racy writes and reads to fields. The problem to solve is about consistency across multiple fields and about visibility of the changes. When recycling, it would just be unnecessary overhead to add a write lock. The `getRetainedValueMatchingKey` will detect if the key and value don't match by calling the `matchesKey` method, which in `EntryImpl` is implemented in this way:
```java
    public boolean matchesKey(Position key) {
        return key.compareTo(ledgerId, entryId) == 0;
    }
```
One of the goals of RangeCache is to avoid exclusive locks as much as possible. The write lock on StampedLock is cheap, but we can avoid even that cost.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
lhotari: In this case, that's not necessary. In this class, when locks are held, only field assignments are performed.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
```diff
@@ -236,34 +272,45 @@ public boolean exists(Key key) {
...
+        try {
+            value.retain();
+        } catch (IllegalReferenceCountException e) {
+            // Value was already deallocated
+            return null;
+        }
+        // check that the value matches the key and that there's at least 2 references to it since
+        // the cache should be holding one reference and a new reference was just added in this method
+        if (value.refCnt() > 1 && value.matchesKey(key)) {
+            return value;
```
heesung-sn, line 308: What if this value is recycled after `if (value.refCnt() > 1 && value.matchesKey(key)) {` has a different value?

heesung-sn: It seems like this value is a mutable obj. How do we guarantee this value's state remains the same in this method?

lhotari: > What if this value is recycled after `if (value.refCnt() > 1 && value.matchesKey(key)) {` has a different value?

There's a call to `value.retain();` a few lines earlier. That is expected to ensure that there's a reference to the value which prevents recycling.  The prevents "use after release" issues since it validates that the value matches the key after the reference count has been incremented. Obviously it won't prevent issues if there's a bug in application reference counting where the application does a double release. That would be an unrealistic expectation.

lhotari: > It seems like this value is a mutable obj. How do we guarantee this value's state remains the same in this method?

There's no such guarantees here and that's not necessary. My previous answer explains that "use after release" issues are prevents and there's a guarantee that at this time the check passes. As explained in the previous answer, it would be unrealistic to expect that this would prevent application double release bugs.

merlimat (approved):

heesung-sn (commented):
Lgtm

heesung-sn:
```java
package org.apache.bookkeeper.mledger.util;

import io.netty.util.Recycler;
import java.util.Collections;
import java.util.concurrent.locks.StampedLock;
import org.testng.annotations.Test;
import java.lang.management.GarbageCollectorMXBean;
import java.lang.management.ManagementFactory;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.*;


public class EntryWrapperPerformanceTestNG {

    private static class EntryWrapper<K, V> {
        private final Recycler.Handle<EntryWrapper> recyclerHandle;
        private static final Recycler<EntryWrapper> RECYCLER = new Recycler<EntryWrapper>() {
            @Override
            protected EntryWrapper newObject(Handle<EntryWrapper> recyclerHandle) {
                return new EntryWrapper(recyclerHandle);
            }
        };
        private final StampedLock lock = new StampedLock();
        private K key;
        private V value;
        long size;

        private EntryWrapper(Recycler.Handle<EntryWrapper> recyclerHandle) {
            this.recyclerHandle = recyclerHandle;
        }

        static <K, V> EntryWrapper<K, V> create(K key, V value, long size) {
            EntryWrapper<K, V> entryWrapper = RECYCLER.get();
            long stamp = entryWrapper.lock.writeLock();
            entryWrapper.key = key;
            entryWrapper.value = value;
            entryWrapper.size = size;
            entryWrapper.lock.unlockWrite(stamp);
            return entryWrapper;
        }

        K getKey() {
            long stamp = lock.tryOptimisticRead();
            K localKey = key;
            if (!lock.validate(stamp)) {
                stamp = lock.readLock();
                localKey = key;
                lock.unlockRead(stamp);
            }
            return localKey;
        }

        V getValue(K key) {
            long stamp = lock.tryOptimisticRead();
            K localKey = this.key;
            V localValue = this.value;
            if (!lock.validate(stamp)) {
                stamp = lock.readLock();
                localKey = this.key;
                localValue = this.value;
                lock.unlockRead(stamp);
            }
            if (localKey != key) {
                return null;
            }
            return localValue;
        }

        long getSize() {
            long stamp = lock.tryOptimisticRead();
            long localSize = size;
            if (!lock.validate(stamp)) {
                stamp = lock.readLock();
                localSize = size;
                lock.unlockRead(stamp);
            }
            return localSize;
        }

        void recycle() {
            key = null;
            value = null;
            size = 0;
            recyclerHandle.recycle(this);
        }
    }


    private static final int ITERATIONS = 1_000_000;
    private static final int THREAD_COUNT = 10; // Simulating 10 concurrent threads
    private static final int TEST_RUNS = 100; // Run each test 10 times

    private long getGCCount() {
        long count = 0;
        for (GarbageCollectorMXBean gcBean : ManagementFactory.getGarbageCollectorMXBeans()) {
            long gcCount = gcBean.getCollectionCount();
            if (gcCount != -1) {
                count += gcCount;
            }
        }
        return count;
    }

    private long getUsedMemory() {
        System.gc(); // Suggest GC before measuring memory usage
        return Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
    }

    private void runTest(String testName, Runnable testLogic) {
        List<Long> times = new ArrayList<>();
        List<Long> gcCounts = new ArrayList<>();
        List<Long> memoryUsages = new ArrayList<>();

        for (int i = 0; i < TEST_RUNS; i++) {
            long startGC = getGCCount();
            long startMemory = getUsedMemory();
            long startTime = System.nanoTime();

            testLogic.run();

            long elapsedTime = System.nanoTime() - startTime;
            long endGC = getGCCount();
            long endMemory = getUsedMemory();

            times.add(elapsedTime);
            gcCounts.add(endGC - startGC);
            memoryUsages.add(endMemory - startMemory);

            System.out.println(testName + " - Run " + (i + 1) + " - Time: " + TimeUnit.NANOSECONDS.toMillis(elapsedTime) + " ms, GC: " + (endGC - startGC) + ", Memory: " + (endMemory - startMemory) + " bytes");
        }

        printStatistics(testName, times, gcCounts, memoryUsages);
    }

    private void printStatistics(String testName, List<Long> times, List<Long> gcCounts, List<Long> memoryUsages) {
        System.out.println("========= " + testName + " Summary =========");
        System.out.println("Average Time: " + TimeUnit.NANOSECONDS.toMillis(average(times)) + " ms");
        System.out.println("Median Time: " + TimeUnit.NANOSECONDS.toMillis(median(times)) + " ms");
        System.out.println("P99 Time: " + TimeUnit.NANOSECONDS.toMillis(percentile(times, 99)) + " ms");
        System.out.println("Max Time: " + TimeUnit.NANOSECONDS.toMillis(Collections.max(times)) + " ms");

        System.out.println("Average GC: " + average(gcCounts));
        System.out.println("Median GC: " + median(gcCounts));
        System.out.println("P99 GC: " + percentile(gcCounts, 99));
        System.out.println("Max GC: " + Collections.max(gcCounts));

        System.out.println("Average Memory: " + average(memoryUsages) + " bytes");
        System.out.println("Median Memory: " + median(memoryUsages) + " bytes");
        System.out.println("P99 Memory: " + percentile(memoryUsages, 99) + " bytes");
        System.out.println("Max Memory: " + Collections.max(memoryUsages) + " bytes");
        System.out.println("====================================");
    }

    private long average(List<Long> values) {
        return values.stream().mapToLong(Long::longValue).sum() / values.size();
    }

    private long median(List<Long> values) {
        List<Long> sorted = new ArrayList<>(values);
        Collections.sort(sorted);
        int middle = sorted.size() / 2;
        return (sorted.size() % 2 == 0) ? (sorted.get(middle - 1) + sorted.get(middle)) / 2 : sorted.get(middle);
    }

    private long percentile(List<Long> values, int percentile) {
        List<Long> sorted = new ArrayList<>(values);
        Collections.sort(sorted);
        int index = (int) Math.ceil(percentile / 100.0 * sorted.size()) - 1;
        return sorted.get(Math.max(0, Math.min(index, sorted.size() - 1)));
    }

    @Test
    public void testObjectPoolingPerformance() {
        runTest("Object Pooling", () -> {
            for (int i = 0; i < ITERATIONS; i++) {
                EntryWrapper<Integer, String> entry = EntryWrapper.create(i, "Value" + i, i);
                entry.recycle();
            }
        });
    }

    @Test
    public void testNewObjectPerformance() {
        runTest("New Object Creation", () -> {
            for (int i = 0; i < ITERATIONS; i++) {
                var str = new String("Value" + i); // Creates a new String object each time
            }
        });
    }

    @Test
    public void testConcurrentObjectPooling() {
        runTest("Concurrent Pooling", () -> {
            ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
            List<Future<?>> futures = new ArrayList<>();

            for (int t = 0; t < THREAD_COUNT; t++) {
                futures.add(executor.submit(() -> {
                    for (int i = 0; i < ITERATIONS / THREAD_COUNT; i++) {
                        EntryWrapper<Integer, String> entry = EntryWrapper.create(i, "Value" + i, i);
                        entry.recycle();
                    }
                }));
            }

            futures.forEach(f -> {
                try {
                    f.get();
                } catch (Exception ignored) {}
            });

            executor.shutdown();
        });
    }

    @Test
    public void testConcurrentNewObjectCreation() {
        runTest("Concurrent New Object", () -> {
            ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
            List<Future<?>> futures = new ArrayList<>();

            for (int t = 0; t < THREAD_COUNT; t++) {
                futures.add(executor.submit(() -> {
                    //List<String> strings = new ArrayList<>();
                    for (int i = 0; i < ITERATIONS / THREAD_COUNT; i++) {
                        var str = new String("Value" + i); // Creates a new String object each time
                    }
                    //strings.clear();
                }));
            }

            futures.forEach(f -> {
                try {
                    f.get();
                } catch (Exception ignored) {}
            });

            executor.shutdown();
        });
    }
}

```

```
========= Concurrent New Object Summary =========
Average Time: 23 ms
Median Time: 22 ms
P99 Time: 32 ms
Max Time: 255 ms
Average GC: 4
Median GC: 4
P99 GC: 4
Max GC: 4
Average Memory: -20971 bytes
Median Memory: 0 bytes
P99 Memory: 0 bytes
Max Memory: 0 bytes
====================================

========= Concurrent Pooling Summary =========
Average Time: 40 ms
Median Time: 36 ms
P99 Time: 77 ms
Max Time: 439 ms
Average GC: 4
Median GC: 4
P99 GC: 4
Max GC: 4
Average Memory: -713031 bytes
Median Memory: 0 bytes
P99 Memory: 2097152 bytes
Max Memory: 2097152 bytes
====================================


========= New Object Creation Summary =========
Average Time: 10 ms
Median Time: 10 ms
P99 Time: 22 ms
Max Time: 40 ms
Average GC: 4
Median GC: 4
P99 GC: 4
Max GC: 4
Average Memory: -524288 bytes
Median Memory: 0 bytes
P99 Memory: 2097152 bytes
Max Memory: 2097152 bytes
====================================


========= Object Pooling Summary =========
Average Time: 54 ms
Median Time: 53 ms
P99 Time: 66 ms
Max Time: 92 ms
Average GC: 4
Median GC: 4
P99 GC: 4
Max GC: 4
Average Memory: -796917 bytes
Median Memory: 0 bytes
P99 Memory: 0 bytes
Max Memory: 2097152 bytes
====================================
```


Sharing some test result: object pooling vs jvm young gc performance test

eolivelli (commented):
+1

lhotari:
> Sharing some test result: object pooling vs jvm young gc performance test

Thank you @heesung-sn for conducting these performance tests comparing object pooling and JVM young generation garbage collection. While the initial results are interesting, I have some concerns about the reliability of manual benchmarking on the JVM. Microbenchmarks are particularly susceptible to JVM optimizations and warmup effects that can skew results. I'd strongly recommend using JMH (Java Microbenchmark Harness) for this type of analysis, as it:
- Handles JVM warmup properly
- Provides detailed GC statistics
- Accounts for common benchmarking pitfalls
- Offers consistent methodology

The good news is that we already have JMH configured in our project at https://github.com/apache/pulsar/tree/master/microbench, so adding new benchmarks should be straightforward.

Some specific suggestions for the next round of testing:
- Create comparative benchmarks for RangeCache with and without object recycling
- Include tests with both G1GC and ZGC ZGenerational on Java 21 to understand performance across different garbage collectors

Once we have these JMH results, we can make a data-driven decision about whether to remove the recycled objects usage from RangeCache entirely. Let me know if you'd like help setting up these JMH benchmarks.

Issue #23149, opened 2024-08-09
[Bug][client] Delay messages can be ack before reaching the delivery time


Denovo1998:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

master

### Minimal reproduce step

I found this when trying to implement the abort function of delayed messages. Here is the test code.

```java
    @Test(timeOut = 20000)
    public void testAbortDelayedMessage() throws Exception {
        final String topic = "persistent://my-property/my-ns/testAbortDelayedMessage";

        Consumer<byte[]> consumer = pulsarClient.newConsumer().topic(topic)
                .subscriptionName("my-subscription")
                .subscriptionType(SubscriptionType.Shared)
                .subscribe();

        Producer<byte[]> producer = pulsarClient.newProducer().topic(topic).enableBatching(false).create();

        admin.topicPolicies().setDelayedDeliveryPolicy(topic,
                DelayedDeliveryPolicies.builder()
                        .active(true)
                        .tickTime(100L)
                        .maxDeliveryDelayInMillis(5000)
                        .build());

        MessageId msgId = producer.newMessage()
                .value("Test delayed message".getBytes())
                .deliverAfter(5, TimeUnit.SECONDS)
                .send();
        log.warn("Message sent with ID: " + msgId);
        producer.flush();

        consumer.acknowledge(msgId);

        while (true) {
            Message<byte[]> msg = consumer.receive(100, TimeUnit.MILLISECONDS);
            if (msg != null) {
                log.warn("Received message: " + new String(msg.getData()));
                consumer.acknowledge(msg.getMessageId());
            }
            log.warn("Waiting for messages...");
        }
    }
```

### What did you expect to see?

Delayed messages are consumed normally. Because consumer have not received this message.

### What did you see instead?

After the delay message corresponding to the active ack is consumer, the delay message is no longer consumed.

### Anything else?

Will there be any problem with ack like this?

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

dao-jun:
firstly, I dont think it's a bug. This is an incorrect usage of the messaging system.  A message should be acked after consumed, not before.

[...]

PR #23907, opened 2025-01-29
[improve][broker] Implementing delayed message cancellation in pulsar

Denovo1998:
Fixes #xyz

Main Issue: [#23149](https://github.com/apache/pulsar/issues/23149)

PIP: #xyz 

### Motivation

In dynamic event-driven architectures, the relevance and timing of messages are crucial. Apache Pulsar currently supports delayed message delivery, which is beneficial for scheduling future processing. However, once scheduled, there is no native support to modify or cancel these messages based on changing circumstances. This limitation poses challenges in situations where messages become obsolete before their scheduled delivery or when system conditions change, necessitating the early processing or cancellation of delayed messages.

如果目前想要实现延迟消息的取消，我们只能在consumer端的内存中存储需要取消的延迟消息，如果延迟时间过长会有很多问题。

### Modifications

The following changes were introduced in Pulsar’s message dispatching components to enable delayed message cancellation:
- Extended AbstractBaseDispatcher with the ability to recognize and handle cancellation properties in message metadata.
- Added constants for identifying cancellation-specific properties in message metadata.

The key implementation detail involves checking each message for a cancellation marker as it becomes eligible for dispatch. If a message is marked for cancellation, it is dropped from the dispatch queue and its associated resources are cleaned up immediately.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java
@@ -20,6 +20,8 @@ package org.apache.pulsar.broker.service;
 
 import static org.apache.bookkeeper.mledger.util.PositionAckSetUtil.andAckSet;
 import static org.apache.bookkeeper.mledger.util.PositionAckSetUtil.isAckSetEmpty;
+import static org.apache.pulsar.common.naming.Constants.DELAY_CANCELED_MESSAGE_POSITION;
+import static org.apache.pulsar.common.naming.Constants.IS_MARK_DELETE_DELAY_MESSAGE;
 import io.netty.buffer.ByteBuf;
 import io.prometheus.client.Gauge;
 import java.util.ArrayList;
@@ -28,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAdder;
+import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.Entry;
 import org.apache.bookkeeper.mledger.ManagedCursor;
@@ -46,10 +49,12 @@ import org.apache.pulsar.broker.service.plugin.EntryFilter;
 import org.apache.pulsar.broker.transaction.pendingack.impl.PendingAckHandleImpl;
 import org.apache.pulsar.client.api.transaction.TxnID;
 import org.apache.pulsar.common.api.proto.CommandAck.AckType;
+import org.apache.pulsar.common.api.proto.KeyValue;
 import org.apache.pulsar.common.api.proto.MessageMetadata;
 import org.apache.pulsar.common.api.proto.ReplicatedSubscriptionsSnapshot;
 import org.apache.pulsar.common.protocol.Commands;
 import org.apache.pulsar.common.protocol.Markers;
+import org.apache.pulsar.common.util.collections.ConcurrentLongLongPairHashMap;
 import org.apache.pulsar.compaction.Compactor;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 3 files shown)


lhotari (commented):
Thank you, @Denovo1998. Since there's a major impact to the core parts of Pulsar, more effort would have to be spent in design decisions. 

All major project decisions (excluding new security issues) are made by the whole community [in the Apache way](https://community.apache.org/committers/decisionMaking.html) on the [dev@pulsar.apache.org mailing list](https://lists.apache.org/list.html?dev@pulsar.apache.org).

The most important thing about engaging with any Apache project is that everyone is equal. All project participants with an opinion can express that opinion and, where appropriate, have the community consider it.
Unless you already have, please join the dev mailing list using [these instructions ](https://pulsar.apache.org/contact/#mailing-lists), and start a discussion about this proposal.

Pulsar project decisions are usually documented in the form of [Pulsar Improvement Proposals (PIPs)](https://github.com/apache/pulsar/tree/master/pip#pulsar-improvement-proposal-pip) and a new PIP is started with a discussion on the mailing list.

In this case, I believe that the use case is very clear, but the solution might be different from what the current pull request includes. 

It's a useful starting point to have this PR in place before starting a PIP, however it won't be merged before we have concluded on a solution that meets our functional and non-functional requirements. For example, we are looking for maintainability (fits the current architecture), performance (doesn't have a negative performance impact on other workloads) and compatibility (doesn't break existing brokers and clients).

At a quick glance it seems that from the architecture side, that the current DelayedDeliveryTracker should be leveraged to support this use case. For encoding the marker messages, Pulsar contains marker messages. It should be considered how these cancellation events would be stored in the topic. Compatibility concerns are around possible limitations with geo-replication. It might not even make sense to store this information in the topic in the first place. When the cancellation messages are stored in the topic, cancellation would only work when the DelayedDeliveryTracker state is such that it has "indexed" the delayed messages and the cancellation messages. For example, the InMemoryDelayedDeliveryTracker keeps state only in memory. The impact of the cancellation commands in the topic would be that before delivering any scheduled message, the state would have to be caught up before delivering a scheduled message. This is just a first thought about the possible impact of supporting cancellation. Due to such performance impacting details, it's more likely that this type of cancellation support would have to be enabled for a namespace or topic explicitly.

I hope this comment helps in making further progress.

lhotari:
> When the cancellation messages are stored in the topic, cancellation would only work when the DelayedDeliveryTracker state is such that it has "indexed" the delayed messages and the cancellation messages. For example, the InMemoryDelayedDeliveryTracker keeps state only in memory. The impact of the cancellation commands in the topic would be that before delivering any scheduled message, the state would have to be caught up before delivering a scheduled message. This is just a first thought about the possible impact of supporting cancellation. Due to such performance impacting details, it's more likely that this type of cancellation support would have to be enabled for a namespace or topic explicitly.

Not directly related, but contains some details about the current delayed delivery tracker solution: #23912 . 
For the `BucketDelayedDeliveryTracker`, it could be feasible to add the cancellation support since it keeps persistent state. However, the challenge that the cancellation commands might not be processed while making delayed message delivery decisions remains. It seems that some sort of RPC would be needed and storing the cancellation commands directly in the BucketDelayedDeliveryTracker persistence solution could be feasible. For the InMemoryDelayedDeliveryTracker, I don't think that it would be feasible to add support for cancellation at all since based on my initial analysis, I don't think that it makes sense to store the cancellation state in the topic itself due to the problems with unprocessed cancellation commands while making delivery decisions.

thetumbled:
First of all, what do you mean by cancelling a delayed message? do you want to delete such message in broker and don't dispatch it, or cancel the delayed time and dispatch it immediately?

As a message queue, pulsar do not support message modification, but if you do want to withdraw the message produced, you can try the transaction feature or topic compaction feature. Frankly speaking, we may meet some issues when combining delayed message feature with  transaction feature or topic compaction feature.

It is also reasonable that pulsar do not support the modification of delayed time or cancellation of delayed message, just like we can't withdraw the message produced. There are too many features in pulsar, we can't integrate all features into pulsar, most of these so-called requirement should be implemented in user's side. And the complexity of Pulsar is pretty high, we need to **balance between the benefit of this minority demand and the risk of breaking existing logic, the complexity of the project.** 
Combination of different features of pulsar can result into lots of problem. IMO, what pulsar mostly need is not adding edge features that only minority would use, but the reliability, stability, simplicity. We need to decrease the complexity of the project.

thetumbled:
This is a great new feature, impacting the core logic, a pip may be needed to get more opinions from the community.

Denovo1998:
bump;

PR #23918, opened 2025-02-02
[improve]Optimize InMemoryDelayedDeliveryTracker by maintaining state

### LINKED ISSUE #23912: [Enhancement] Optimize delayed message delivery by reducing unnecessary reads to storage in InMemoryDelayedDeliveryTracker
### Search before asking

### Motivation

In the current delayed message delivery, there's an opportunity to reduce unnecessary reads to storage.

In Pulsar, there are 2 implementations for the delayed delivery tracker, InMemoryDelayedDeliveryTracker and BucketDelayedDeliveryTracker.
This is configured by the `delayedDeliveryTrackerFactoryClassName` configuration key. The default setting chooses the in memory implementation:
https://github.com/apache/pulsar/blob/b02d52ca8c8afd62681c0e243d16d8958abb5380/conf/broker.conf#L614-L617

The BucketDelayedDeliveryTracker contains an optimization to skip messages in reading which have been "indexed":
https://github.com/apache/pulsar/blob/11a615e7cecb57c8c240cc565c5ac62d938c490d/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L446-L457

This already reduces reads when BucketDelayedDeliveryTracker is in use.

The state of the InMemoryDelayedDeliveryTracker gets cleared after all consumers have disconnected:
https://github.com/apache/pulsar/blob/12b057999f5674ae6da6b04ff8bd212d1904cc88/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L187-L202

### Solution

It would be useful to keep state also in the InMemoryDelayedDeliveryTracker and skip reading delayed messages when the information is already available for the delivery time of a specific entry.

### Alternatives

_No response_

### Anything else?

_No response_

yunmaoQu:
Fixes #23912

Main Issue: #23912

PIP: #xyz 

### Motivation

In the current delayed message delivery, there's an opportunity to reduce unnecessary reads to storage.
### Modifications
It would be useful to keep state also in the InMemoryDelayedDeliveryTracker and skip reading delayed messages when the information is already available for the delivery time of a specific entry.

### Verifying this change

This change is already covered by existing tests.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: 
https://github.com/yunmaoQu/pulsar/pull/2

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -218,9 +218,26 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
         return positions;
     }
 
+    public boolean shouldSkipMessage(long ledgerId, long entryId) {
+        for (Long2ObjectMap<Roaring64Bitmap> ledgerMap : delayedMessageMap.values()) {
+            Roaring64Bitmap entryIds = ledgerMap.get(ledgerId);
+            if (entryIds != null && entryIds.contains(entryId)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     @Override
     public CompletableFuture<Void> clear() {
-        this.delayedMessageMap.clear();
+        long cutoffTime = getCutoffTime();
+        delayedMessageMap.headMap(cutoffTime).clear();
+
+        if (log.isDebugEnabled()) {
+            log.debug("[{}] Cleared expired delayed messages before {}, remaining messages: {}",
+                    dispatcher.getName(), cutoffTime, getNumberOfDelayedMessages());
+        }
+
         return CompletableFuture.completedFuture(null);
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -193,12 +193,6 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
                 shouldRewindBeforeReadingOrReplaying = false;
             }
             redeliveryMessages.clear();
-            delayedDeliveryTracker.ifPresent(tracker -> {
-                // Don't clean up BucketDelayedDeliveryTracker, otherwise we will lose the bucket snapshot
-                if (tracker instanceof InMemoryDelayedDeliveryTracker) {
-                    tracker.clear();
-                }
-            });
         }
 
         if (isConsumersExceededOnSubscription()) {
@@ -448,7 +442,10 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
         // Filter out and skip read delayed messages exist in DelayedDeliveryTracker
         if (delayedDeliveryTracker.isPresent()) {
             final DelayedDeliveryTracker deliveryTracker = delayedDeliveryTracker.get();
-            if (deliveryTracker instanceof BucketDelayedDeliveryTracker) {
+            if (deliveryTracker instanceof InMemoryDelayedDeliveryTracker) {
+                skipCondition = position -> ((InMemoryDelayedDeliveryTracker) deliveryTracker)
+                        .shouldSkipMessage(position.getLedgerId(), position.getEntryId());
+            } else if (deliveryTracker instanceof BucketDelayedDeliveryTracker) {
                 skipCondition = position -> ((BucketDelayedDeliveryTracker) deliveryTracker)
                         .containsMessage(position.getLedgerId(), position.getEntryId());
             }


pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -218,9 +218,26 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
     @Override
     public CompletableFuture<Void> clear() {
-        this.delayedMessageMap.clear();
+        long cutoffTime = getCutoffTime();
```
dao-jun, line 233: why only clear expired index?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -218,9 +218,26 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
         return positions;
     }
 
+    public boolean shouldSkipMessage(long ledgerId, long entryId) {
```
dao-jun, line 221: it better to keep the method name with `BucketDelayedDeliveryTracker`, change `shouldSkipMessage `  to `contains`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
yunmaoQu: The decision to clear only expired indices in the clear() method of the InMemoryDelayedDeliveryTracker is aimed at optimizing performance and maintaining the logical state of the message delivery system.

By focusing on expired messages, we reduce the overhead associated with clearing and re-adding valid messages, which enhances performance, especially in scenarios with a high volume of delayed messages. This approach also allows us to retain the state of valid messages, enabling more efficient message delivery without needing to re-read them from storage.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
yunmaoQu: ok

yunmaoQu:
Hi @dao-jun , I'm in need of your review on this PR to move forward with the feature. Could you please check it out as soon as you can? Thanks a lot!

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: I doubt that it will introduce memory leak problem. For example, we need to clear all delayed messages at method:
org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers#clearDelayedMessages

IMO, the payoff of re-adding valid messages for once is acceptable, while the risk of such change is great.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -448,7 +442,10 @@ protected Predicate<Position> createReadEntriesSkipConditionForNormalRead() {
         // Filter out and skip read delayed messages exist in DelayedDeliveryTracker
         if (delayedDeliveryTracker.isPresent()) {
             final DelayedDeliveryTracker deliveryTracker = delayedDeliveryTracker.get();
-            if (deliveryTracker instanceof BucketDelayedDeliveryTracker) {
+            if (deliveryTracker instanceof InMemoryDelayedDeliveryTracker) {
+                skipCondition = position -> ((InMemoryDelayedDeliveryTracker) deliveryTracker)
+                        .shouldSkipMessage(position.getLedgerId(), position.getEntryId());
+            } else if (deliveryTracker instanceof BucketDelayedDeliveryTracker) {
                 skipCondition = position -> ((BucketDelayedDeliveryTracker) deliveryTracker)
                         .containsMessage(position.getLedgerId(), position.getEntryId());
```
thetumbled, line 450: `change shouldSkipMessage`  to `contains` to avoid duplicate code.

thetumbled (changes requested):
Could you help to add unit test to cover the logic changed?
We need to ensure that such kind of skipping reading delayed messages won't impact the normal read when we want to dispatch them.
And i wonder that how much improvement this pr can offer? Do unnecessary reads occur frequently or not? If the answer is positive, i am glad to accept this improvement.

Issue #23910, opened 2025-01-29
[Bug] Seek on a timestamp causes the subscription to be resetted to the earliest position.


Gilthoniel:
### Search before asking

- [ ] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

Pulsar v4.0.2
Pulsar Go client v0.14

### Minimal reproduce step

With v4.0.2, after restarting brokers, we observe a wrong position in subscription after a seek by timestamp. The position seems to be the earliest position instead of the next message after the timestamp.

Note that after downgrading to v4.0.1, the seeking position gets back to the proper one so the message after the timestamp.

### What did you expect to see?

On a seek by timestamp, we expect the subscription to reset to the next message after the timestamp available in the topic.

### What did you see instead?

We observe subscriptions to reset on the earliest position when we use seek by timestamp.

May that be related to #22129 ?

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
Thanks for reporting this issue.

The change added in #22792 can be disabled by configuring `managedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis=-1`. This configuration value doesn't exist in `broker.conf`. For k8s deployments, setting `PULSAR_PREFIX_managedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis: "-1"` in the config map is a way to configure it.

Do you have a chance to test if the problem reproduces with 4.0.2 with this configuration?

Gilthoniel:
@lhotari thanks, it indeed works properly with v4.0.2 when using `-1` so that should confirm the change has introduced a regression.

lhotari:
> [@lhotari](https://github.com/lhotari) thanks, it indeed works properly with v4.0.2 when using `-1` so that should confirm the change has introduced a regression.

Thanks for confirming. I have updated the release notes with information about this regression and how to apply the workaround: https://github.com/apache/pulsar/releases/tag/v4.0.2 . It will soon be published in https://pulsar.apache.org/release-notes/versioned/pulsar-4.0.2/ .

> ### Known Major Issues in this Release
> 
> This release contains a known issue impacting the `seek` method by timestamp. This issue was reported after the release was published.
> 
[...]

PR #22792, opened 2024-05-28
[improve][broker] Optimize subscription seek (cursor reset) by timestamp

dao-jun:
Fixes https://github.com/apache/pulsar/issues/22129

### Motivation

Pulsar uses binary search to find the message by timestamp, it will reduce the number of iterations to find the message, and make it more efficient and faster.

Even though the current implementation is correct, and using binary search to speed-up, but it's still not efficient *enough*.
The current implementation is to scan all the ledgers to find the message by timestamp.
This is a performance bottleneck, especially for large topics with many messages.
Say, if there is a topic which has 1m entries, through the binary search, it will take 20 iterations to find the message.

In some extreme cases, it may lead to a timeout, and the client will not be able to seeking by timestamp.

The motivation of this PR is to optimize the finding message by timestamp, to make it more efficient and faster.

### Modifications

Before search entires, calculate the `start`, `end` position by `LedgerInfo#timestamp` and *only* search entries in the range to avoid search the entire ledgers.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23919, opened 2025-02-02
[fix][broker] Fix seeking by timestamp can be reset the cursor position to earliest

dao-jun:
Fixes https://github.com/apache/pulsar/issues/23910

### Motivation

As the we don't enable `AppendBrokerTimestampMetadataInterceptor` by default, so the entries timestamp is not Strictly Increasing.

Because the message timestamp is generated by the clients, the messages from different producers maybe not in global ordering(because of network delay, backpressure, thread scheduling, etc)

In a single ledger, they may be arranged in the following way:
[2, 1, 3, 5, 4, 6, 7, 9, 8.....]
Overall, they have a self increasing trend, but locally, it may be possible not.

1. If the `Position` is null when call `PersistentMessageFinder.findMessages`, it will reset the cursor position to `earliest`, see: https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L804-L824
2. If the first entry we read from bk cannot meet the condition, it will return null. see: https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java#L87-L91

In https://github.com/apache/pulsar/pull/22792/commits/823a55dd659da6c1c54dca55d08f92053c91013e#diff-1d1f02c0ae1aed67e77512aebe4d7233705490b14960ee428611462e446861e5R133-R142, we optimized the case of the target entry maybe in the last opening ledger. It's very intuitive but the actual situation is more complicated:

If we want to find the message's position whose timestamp is 101, the second-to-last ledger's close timestamp is 100, and the entries's timestamp in the last opening ledger arranged as [102,103,101,104...]
The first entry's timestamp is greater than 101, so it will not meet the condition of https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java#L74-L75, and the return value will be null, the cursor position will be finally set to earliest.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

The tests will be run in the forked repository until all PR review comments have
been handled, the tests pass and the PR is approved by a reviewer.
-->

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -1336,7 +1336,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         }
 
         if (max <= 0) {
-            callback.findEntryFailed(new ManagedLedgerException("No entries available"), Optional.empty(), ctx);
+            callback.findEntryComplete(null, ctx);
             return;
         }
 
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -1336,7 +1336,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         }
 
         if (max <= 0) {
-            callback.findEntryComplete(null, ctx);
+            callback.findEntryFailed(new ManagedLedgerException("No entries available"), Optional.empty(), ctx);
             return;
         }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java
@@ -21,6 +21,8 @@ package org.apache.pulsar.broker.service.persistent;
 import static com.google.common.base.Preconditions.checkArgument;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.bookkeeper.mledger.AsyncCallbacks;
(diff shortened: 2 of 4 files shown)


lhotari (approved):
Thanks for investigating the issue and providing a fix, @dao-jun . I wonder if there would be a way to add a failing test case which would then prevent future regressions?

lhotari:
Checkstyle errors:
```
[INFO] There are 2 errors reported by Checkstyle 10.14.2 with /home/runner/work/pulsar/pulsar/buildtools/src/main/resources/pulsar/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java:[25,1] (imports) ImportOrder: Extra separation in import group before 'com.google.common.annotations.VisibleForTesting'
Error:  src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java:[25,1] (imports) ImportOrder: Import com.google.common.annotations.VisibleForTesting appears after other imports that it should precede
```

lhotari:
@dao-jun Do you have thoughts about adding a failing test case that would prevent future regressions? It seems that we don't currently have proper tests that would test the seeking behavior at the Pulsar client level.

liudezhi2098 (approved):

dao-jun:
@lhotari I've added a client level seek test, PTAL

dao-jun:
It looks there is another case which may cause the issue: the timestamp's target entry belongs to a trimmed ledger.

lhotari (approved):

dao-jun:
> Good work @dao-jun , thanks for handling this.

Thanks for review, you are welcome

poorbarcode:
https://lists.apache.org/thread/jp1rpjqz9cczsro3b36t15z239zqfqmy Cherry-picked

PR #23924, opened 2025-02-05
[fix][client] Replace String with WebTarget in authenticationStage

nodece:
### Motivation

This PR is based on #23905 and is used to fix the TLS context loss.

### Modifications

- Add `org.apache.pulsar.client.api.Authentication#authenticationStage(javax.ws.rs.client.WebTarget, org.apache.pulsar.client.api.AuthenticationDataProvider, java.util.Map<java.lang.String,java.lang.String>, java.util.concurrent.CompletableFuture<java.util.Map<java.lang.String,java.lang.String>>)` instead of `org.apache.pulsar.client.api.Authentication#authenticationStage(java.lang.String, org.apache.pulsar.client.api.AuthenticationDataProvider, java.util.Map<java.lang.String,java.lang.String>, java.util.concurrent.CompletableFuture<java.util.Map<java.lang.String,java.lang.String>>)` in the `org.apache.pulsar.client.api.Authentication`, and fix call.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari (changes requested):
This is a breaking change in the org.apache.pulsar.client.api.Authentication API that is exposed for plugins to implement. 
There should be a PIP in place and the implementation should be made in a backwards compatible way.
It should be possible to add conditional logic when the Authentication instance implements some other interface which is for handling this particular case.

Issue #23925, opened 2025-02-05
[Bug] The RawMessageImpl.getProperties throw `Duplicate key` IllegalStateException.


horizonzy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

master

### Minimal reproduce step

You can run this code to reproduce it.
```
import io.netty.buffer.ByteBuf;
import io.netty.buffer.Unpooled;
import java.util.Map;
import org.apache.pulsar.common.api.proto.KeyValue;
import org.apache.pulsar.common.api.proto.MessageMetadata;
import org.apache.pulsar.common.api.raw.RawMessage;
import org.apache.pulsar.common.api.raw.RawMessageImpl;
import org.apache.pulsar.common.api.raw.ReferenceCountedMessageMetadata;
import org.apache.pulsar.shade.io.netty.buffer.ByteBufAllocator;

public class Test {
    public static void main(String[] args) {
        byte[] messageContent = "Hello, Pulsar!".getBytes();
        ByteBuf payload = Unpooled.wrappedBuffer(messageContent);

        org.apache.pulsar.shade.io.netty.buffer.ByteBuf buffer = ByteBufAllocator.DEFAULT.buffer();
        ReferenceCountedMessageMetadata referenceCountedMessageMetadata = ReferenceCountedMessageMetadata.get(buffer);
        MessageMetadata metadata1 = referenceCountedMessageMetadata.getMetadata();
        KeyValue keyValue = metadata1.addProperty();
        keyValue.setKey("aa");
        keyValue.setValue("bb");

        KeyValue keyValue1 = metadata1.addProperty();
        keyValue1.setKey("aa");
        keyValue1.setValue("cc");
        RawMessage rawMessage = RawMessageImpl.get(referenceCountedMessageMetadata, null, null, 0, 0, 0);
        Map<String, String> properties = rawMessage.getProperties();
    }
}

```

### What did you expect to see?

Do not throw the exception.

### What did you see instead?

It will throw the following exception.
```
Exception in thread "main" java.lang.IllegalStateException: Duplicate key aa (attempted merging values bb and cc)
	at java.base/java.util.stream.Collectors.duplicateKeyException(Collectors.java:135)
[...]

PR #6390, opened 2020-02-21
fix duplicate key to send propertys

liudezhi2098:
**Motivation**
Fix when sending a message, set duplicate key to properties, can't pull the message while concumer #6388 
```javascript
//org.apache.pulsar.client.impl.MessageImpl
if (msgMetadata.getPropertiesCount() > 0) {
            this.properties = Collections.unmodifiableMap(msgMetadataBuilder.getPropertiesList().stream()
                    .collect(Collectors.toMap(KeyValue::getKey, KeyValue::getValue)));
        } else {
            properties = Collections.emptyMap();
        }
        this.schema = schema;
```
Collectors.toMap can not allowed duplicate key

**Changes**
Replace old value with new value
```javascript
if (msgMetadata.getPropertiesCount() > 0) {
            this.properties = Collections.unmodifiableMap(msgMetadataBuilder.getPropertiesList().stream()
                    .collect(Collectors.toMap(KeyValue::getKey, KeyValue::getValue,
                            (oldValue,newValue) -> newValue)));
        } else {
            properties = Collections.emptyMap();
        }
        this.schema = schema;
```

PR #23927, opened 2025-02-05
[fix] [common] Fix RawMessageImpl.getProperties() failed when the message metadata contains the same key but with different values

horizonzy:
Main Issue: #23925

My fork PR: https://github.com/horizonzy/pulsar/pull/20

### Motivation
If the message metadata contains the same key but with different values, we should make the later value override the previous value.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/api/raw/RawMessageImpl.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/api/raw/RawMessageImpl.java
@@ -99,7 +99,8 @@ public class RawMessageImpl implements RawMessage {
                               (oldValue, newValue) -> newValue));
         } else if (msgMetadata.getMetadata().getPropertiesCount() > 0) {
             return msgMetadata.getMetadata().getPropertiesList().stream()
-                    .collect(Collectors.toMap(KeyValue::getKey, KeyValue::getValue));
+                    .collect(Collectors.toMap(KeyValue::getKey, KeyValue::getValue,
+                            (oldValue, newValue) -> newValue));
         } else {
             return Collections.emptyMap();
         }
--- a/pulsar-common/src/test/java/org/apache/pulsar/common/api/raw/RawMessageImplTest.java
+++ b/pulsar-common/src/test/java/org/apache/pulsar/common/api/raw/RawMessageImplTest.java
@@ -41,7 +41,7 @@ public class RawMessageImplTest {
     private static final String HARD_CODE_KEY_ID_VALUE  = "__pfn_input_msg_id_value__";
 
     @Test
-    public void testGetProperties() {
+    public void testGetMessageSingleMetadataProperties() {
         ReferenceCountedMessageMetadata refCntMsgMetadata =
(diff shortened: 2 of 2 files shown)


nodece (approved):

mattisonchao (dismissed):

mattisonchao:
Some questions:

1. why do the message properties allow duplicated keys?
2. If the message allows a duplicate key, then we remove one of them during reading. Is it some properties lost for the user?

BewareMyPower (changes requested):
+1 to @mattisonchao's comment. We should fix the case that duplicated keys are included in the same message.

horizonzy:
> If a message has duplicated keys in its property, I think we should allow the downstream to determine how to process such duplicated keys rather than retain a random value for such keys. For example, we can support throwing an exception that contains a list of `<duplicated-key, values...>`.

I think not. We define the MessageMetadata properties as followings.
```
message MessageMetadata {
    required string producer_name   = 1;
    required uint64 sequence_id     = 2;
    required uint64 publish_time    = 3;
    repeated KeyValue properties    = 4;
}
```
We use the proto repeated to define the properties like map, and the field properties type is repeated, so the order is determined. 

For the map case, we should always use the new k,v to override the old k,v. They didn't care whether the message has duplicated keys, they just need use the newest k,v.

horizonzy:
> > If a message has duplicated keys in its property, I think we should allow the downstream to determine how to process such duplicated keys rather than retain a random value for such keys. For example, we can support throwing an exception that contains a list of `<duplicated-key, values...>`.
> 
> I think not. We define the MessageMetadata properties as followings.
> 
> ```
> message MessageMetadata {
>     required string producer_name   = 1;
>     required uint64 sequence_id     = 2;
>     required uint64 publish_time    = 3;
>     repeated KeyValue properties    = 4;
> }
> ```
> 
> We use the proto repeated to define the properties like map, and the field properties type is repeated, so the order is determined.
> 
> For the map case, we should always use the new k,v to override the old k,v. They didn't care whether the message has duplicated keys, they just need use the newest k,v.

@mattisonchao here

BewareMyPower (approved):

BewareMyPower:
That makes sense to me. I approved it because it solves the issue when duplicated keys are already written as the properties. Regarding the producer side to prevent duplicated keys in `MessageMetadata`, I think it can be done in a separated PR.

mattisonchao:
well, I can approve it since the `MessageImpl` side has a kind of same logic. https://github.com/apache/pulsar/pull/6390

umm. it seems like that was caused by proto2 not supporting map type, and we didn't limit the duplicate key on the protocol side.

mattisonchao (approved):

PR #23930, opened 2025-02-05
[fix][broker] Fix rate limiter token bucket and clock consistency issues causing excessive throttling and connection timeouts

### LINKED ISSUE #23920: [Bug] Producers getting timed out intermittently
### Search before asking

### Read release policy

### Version

Setup summary:
 * Pulsar 4.0.1, 3 brokers, 5 bookies, 3 ZKs, JVM - 17.0.13
 * Produce and Consume through Pulsar Java client (4.0.1) - Clients are using Java 11

Also tried with 4.0.2 version

Pulsar Deployment - VM based, we are running on EC2 machines.

As far as configurations are concerned, we had following settings enabled
 * Extensible load balnacer and TransferShedder 
 * Bundle split algorithm - `flow_or_qps_equally_divide`
 * Tiered storage to S3

We updated the load balancer configurations to use the Pulsar defaults, i.e. Modular Load balancer, ThresholdShedder and `range_equally_divide` for bundle split strategy. Reason for this change was because we observed errors in broker logs related to topic shedding.

Then we also disabled tiered storage.

In all cases, the reported issue is observed.

### Minimal reproduce step

This issue is continuation of Apache Pulsar's slack thread - https://apache-pulsar.slack.com/archives/C5Z4T36F7/p1738158630009249
To and fro discussion with @lhotari can be found there.

We are observing an intermittent issue where Pulsar producer client is unable to produce message to a partitioned topic, it is failing with org.apache.pulsar.client.api.PulsarClientException$TimeoutException  after 30 seconds.

```java
// Pulsar client
this.pulsarClient =
          PulsarClient.builder()
              .serviceUrl(connectionUrl)
              .memoryLimit(0, SizeUnit.MEGA_BYTES)
              .build();

// Producer configurations
ProducerBuilder<T> builder =
        pulsarClient
            .newProducer(schema)
            .producerName(producerName)
            .topic(topicName)
            .compressionType(CompressionType.LZ4)
            .enableBatching(true)
            .batchingMaxPublishDelay(30, TimeUnit.MILLISECONDS)
            .blockIfQueueFull(true);
```

Pulsar brokers, bookies and client instances are running in the same AWS VPC, so chances of network outage are quite low. Brokers are configured to use advertisedAddress , advertisedListeners is not configured.

In the client application, the Pulsar client and Producer object is initialized once and then messages are produced in sporadic intervals.
There is no other producer or consumer running on the cluster and there is more than sufficient resources available to cater to produce request.

Pulsar topics are partitioned.
We also have applied publish rate and dispatch rate policies of 1 MBps on the said topic (per partition)
But the throughput of messages being produced is far less than applied quota, for example, 1-5 messages in 2-5 minutes.

Image

We did try to remove the publish and dispatch rates on the topic and issue was resolved.
Although we could not conclusively say that issue got resolved after removing those policies.
We also applied those policies back and did not observe issue for a day and when issue occurred again, we unload the namespace bundles and assigned to different brokers.

We did this several times so unloading namespace bundles seems to quick fix for the issue.

We also enabled debug logs on client and the server side.

Observed a "Setting auto read to false" false in broker logs (see below)

```
2025-02-02T16:49:45,932+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-8] Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf Got request to create producer 
2025-02-02T16:49:45,932+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-8] [Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf] Added producer -- count: 1
2025-02-02T16:49:45,932+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Put schema finished
2025-02-02T16:49:45,933+0000 [broker-topic-workers-OrderedExecutor-5-0] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:39103] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-8}, client=[id: 0x8eacecff, L:/10.10.17.40:6650 - R:/10.22.2.15:39103] [SR:10.22.2.15, state:Connected], producerName=Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf, producerId=8}, role: null
2025-02-02T16:49:45,933+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Schema is already exists
2025-02-02T16:49:45,933+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-9] Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf Got request to create producer 
2025-02-02T16:49:45,933+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-9] [Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf] Added producer -- count: 1
2025-02-02T16:49:45,933+0000 [broker-topic-workers-OrderedExecutor-5-0] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:39103] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-9}, client=[id: 0x8eacecff, L:/10.10.17.40:6650 - R:/10.22.2.15:39103] [SR:10.22.2.15, state:Connected], producerName=Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf, producerId=9}, role: null
2025-02-02T16:49:45,933+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Put schema finished
2025-02-02T16:49:46,175+0000 [pulsar-io-3-12] DEBUG org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:39103] Received send message request. producer: 9:0 Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf:0 size: 133, partition key is: 1, ordering key is null, uncompressedSize is 45
2025-02-02T16:49:46,175+0000 [pulsar-io-3-12] DEBUG org.apache.pulsar.broker.service.ServerCnxThrottleTracker - [[id: 0x8eacecff, L:/10.10.17.40:6650 - R:/10.22.2.15:39103] [SR:10.22.2.15, state:Connected]] Setting auto read to false
2025-02-02T16:49:46,180+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.Producer - [PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-9}] [Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf] [9] triggered send callback. cnx /10.22.2.15:39103, sequenceId 0
2025-02-02T16:49:46,180+0000 [pulsar-io-3-12] DEBUG org.apache.pulsar.broker.service.Producer - [PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-9}] [Producer352f00f6-a898-4ed4-9469-8eb62c2ed0bf] [9] Persisted message. cnx [id: 0x8eacecff, L:/10.10.17.40:6650 - R:/10.22.2.15:39103] [SR:10.22.2.15, state:Connected], sequenceId 0
2025-02-02T16:49:47,082+0000 [pulsar-io-3-13] DEBUG org.apache.pulsar.broker.service.ServerCnx - New connection from /127.0.0.1:57030
2025-02-02T16:49:47,083+0000 [pulsar-io-3-13] INFO  org.apache.pulsar.broker.service.ServerCnx - Closed connection from /127.0.0.1:57030
2025-02-02T16:49:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-6] No backlog. Update old position info is null
2025-02-02T16:49:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-6] Time backlog quota = [-1]. Checking if exceeded.
2025-02-02T16:49:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.BrokerService - quota not exceeded for [persistent://public/default/default_shivam_stream-partition-6]
2025-02-02T16:49:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-7] No backlog. Update old position info is null
2025-02-02T16:49:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-7] Time backlog quota = [-1]. Checking if exceeded.
2025-02-02T16:49:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.BrokerService - quota not exceeded for [persistent://public/default/default_shivam_stream-partition-7]
--
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-7] Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf Got request to create producer 
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-7] [Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf] Added producer -- count: 1
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Put schema finished
2025-02-02T16:52:39,369+0000 [broker-topic-workers-OrderedExecutor-4-0] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:54474] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-7}, client=[id: 0x83526df1, L:/10.10.17.40:6650 - R:/10.22.2.15:54474] [SR:10.22.2.15, state:Connected], producerName=Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf, producerId=7}, role: null
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Schema is already exists
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-4] Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf Got request to create producer 
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-4] [Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf] Added producer -- count: 1
2025-02-02T16:52:39,369+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Put schema finished
2025-02-02T16:52:39,369+0000 [broker-topic-workers-OrderedExecutor-3-0] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:54474] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-4}, client=[id: 0x83526df1, L:/10.10.17.40:6650 - R:/10.22.2.15:54474] [SR:10.22.2.15, state:Connected], producerName=Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf, producerId=4}, role: null
2025-02-02T16:52:39,596+0000 [pulsar-io-3-7] DEBUG org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:54474] Received send message request. producer: 9:0 Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf:0 size: 133, partition key is: 1, ordering key is null, uncompressedSize is 45
2025-02-02T16:52:39,596+0000 [pulsar-io-3-7] DEBUG org.apache.pulsar.broker.service.ServerCnxThrottleTracker - [[id: 0x83526df1, L:/10.10.17.40:6650 - R:/10.22.2.15:54474] [SR:10.22.2.15, state:Connected]] Setting auto read to false
2025-02-02T16:52:39,601+0000 [BookKeeperClientWorker-OrderedExecutor-4-0] DEBUG org.apache.pulsar.broker.service.Producer - [PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-9}] [Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf] [9] triggered send callback. cnx /10.22.2.15:54474, sequenceId 0
2025-02-02T16:52:39,601+0000 [pulsar-io-3-7] DEBUG org.apache.pulsar.broker.service.Producer - [PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-9}] [Producer3e526aac-10f4-4a5e-8c51-2dbda8545bbf] [9] Persisted message. cnx [id: 0x83526df1, L:/10.10.17.40:6650 - R:/10.22.2.15:54474] [SR:10.22.2.15, state:Connected], sequenceId 0
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-6] No backlog. Update old position info is null
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-6] Time backlog quota = [-1]. Checking if exceeded.
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.BrokerService - quota not exceeded for [persistent://public/default/default_shivam_stream-partition-6]
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-7] No backlog. Update old position info is null
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-7] Time backlog quota = [-1]. Checking if exceeded.
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.BrokerService - quota not exceeded for [persistent://public/default/default_shivam_stream-partition-7]
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-8] No backlog. Update old position info is null
2025-02-02T16:52:51,001+0000 [pulsar-backlog-quota-checker-OrderedScheduler-0-0] DEBUG org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/default_shivam_stream-partition-8] Time backlog quota = [-1]. Checking if exceeded.
--
2025-02-02T16:54:29,252+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-4] Producerbc584de6-13d3-46c3-9d7f-9f5c4f7619b2 Got request to create producer 
2025-02-02T16:54:29,252+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.AbstractTopic - [persistent://public/default/default_shivam_stream-partition-4] [Producerbc584de6-13d3-46c3-9d7f-9f5c4f7619b2] Added producer -- count: 1
2025-02-02T16:54:29,252+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Put schema finished
2025-02-02T16:54:29,252+0000 [broker-topic-workers-OrderedExecutor-3-0] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.22.2.15:55876] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/default_shivam_stream-partition-4}, client=[id: 0x4b78fe29, L:/10.10.17.40:6650 - R:/10.22.2.15:55876] [SR:10.22.2.15, state:Connected], producerName=Producerbc584de6-13d3-46c3-9d7f-9f5c4f7619b2, producerId=4}, role: null
2025-02-02T16:54:29,253+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] DEBUG org.apache.pulsar.broker.service.schema.SchemaRegistryServiceImpl - [public/default/default_shivam_stream] Schema is already exists
20...
```

We also observed that `pulsar.pulsar_publish_rate_limit_times` for the topics with this issue and found that they had non-zero values for this metric.

Image

Another observation when we encounter this issue is that no messages are delivered to subscriptions. We tried consuming data from topic using Pulsar's CLI client from earliest position but no records were delivered.
We enabled `unblockStuckSubscriptionEnabled=true` in broker.conf but still observed this issue.

Currently we are removing the topicPolicies - publish rate and dispatch rate and will observe if the issue recurs.

As per our tests, these policies are working fine when there is sufficient records being produced and consumed from pulsar topics.
But when producer load is sparse, we are encountering the issue where producer is unable to produce message.

Let me know if any other details are required around this.

Records getting produced successfully

### What did you see instead?

Records not being produced or consumed at sporadic intervals.
The issue happens intermittently and gets some solved on it's own, we haven't been able to nail the scenarios which deterministically leads to this issue.

The client patterns would be sparse producer and non zero values of `pulsar.pulsar_publish_rate_limit_times` during the duration of issue.

### Anything else?

_No response_

lhotari:
Fixes #23920

### Motivation

Prior to this PR, connections in Pulsar could experience excessive throttling and timeouts due to issues in rate limiting implementation. This was caused by both calculation errors in token bucket logic and problems with clock source handling in virtualized environments. The PR addresses these issues to prevent unintended long throttling periods and connection timeouts.

Two key issues are addressed in this PR:

1. Calculation Issue in AsyncTokenBucket:
   - The current token calculation logic is flawed when handling consumed tokens in conjunction with newly added tokens
   - When adding new tokens, the eventually consistent token calculation should happen before limiting the tokens to the bucket capacity limit. Since the new tokens were first added and capped to the capacity limit, this led to a situation where the tokens would become negative when the DefaultMonotonicSnapshotClock's thread was not executed with the expected interval for example due to high CPU usage.

2. Clock Source Consistency:
   - The current rate limiting implementation in Pulsar broker can be affected by clock source issues where `System.nanoTime()` isn't strictly monotonic and consistent across multiple CPUs
   - This is a known issue in virtualized environments and certain hardware configurations, particularly:
     - Multi-socket systems where each CPU may have its own TSC (Time Stamp Counter)
     - When CPUs are affected by power management features and running at different frequencies
     - The previous solution was impacted since `System.nanoTime()` was read on different threads.

For details about timekeeping challenges in virtualized environments, see [Linux kernel documentation on timekeeping](https://docs.kernel.org/virt/kvm/x86/timekeeping.html).

Without this fix, when time leaps backwards (for example due to `System.nanoTime` being read by a thread running on a different CPU with unsynchronized TSC), the rate limiter would incorrectly throttle connections for the duration of the time leap. In virtualized environments, the time difference between CPU cores can be significant. Even though such large variances are uncommon, they can cause extended connection timeouts when they occur. This PR significantly improves reliability by preventing such time-variance-induced throttling entirely.

While both forward and backward time leaps can occur in these environments, only backward leaps are explicitly handled. Forward leaps cannot be reliably detected within the eventually consistent logic used in AsyncTokenBucket, and their impact is bounded - at most one rate period (default 1 second) of potential overuse. This design choice favors implementation simplicity and performance while still protecting against the more problematic backward leap case that could cause extended throttling.

Although this PR also covers clock source consistency issues, the most probable root cause for the reported issue #23920 is, however, the calculation logic that was assuming that the DefaultMonotonicSnapshotClock's update thread would always be executing consistently.

### Modifications

1. **AsyncTokenBucket improvements**:
   - Fixed token calculation logic to properly handle consumed tokens:
     - Subtract pendingConsumedTokens from new tokens before capping at bucket capacity
     - This ensures correct token balance even when DefaultMonotonicSnapshotClock's update thread is delayed
   - Added comprehensive test coverage for negative balance and eventual consistency scenarios
   - Added test cases to verify proper token bucket behavior under various conditions

2. **DefaultMonotonicSnapshotClock enhancements**:
   - Implemented a single dedicated thread approach to minimize clock inconsistencies
     - Since System.nanoTime() values can vary slightly between CPUs in multi-core/multi-socket systems, using a single thread ensures time values are read consistently from one CPU
     - This effectively eliminates time variance issues by avoiding cross-CPU clock reads
   - Added handling of backward time leaps in clock source
   - Forward time leaps are intentionally not handled since they cannot be reliably detected in the eventually consistent token bucket logic
   - Forward leaps have minimal impact - at most 1 second of potential overuse with default rate period settings
   - Implemented a low-overhead synchronization mechanism for obtaining consistent clock values when required
   - The implementation is optimized for the common path while providing guarantees when needed

3. **Rate limiter implementations**:
   - Updated PublishRateLimiterImpl to use proper executor for unthrottling
   - Added error handling for producer unthrottling
   - Modified DispatchRateLimiter and SubscribeRateLimiter to use the monotonic clock instance

### Performance

[AsyncTokenBucketBenchmark](https://github.com/lhotari/pulsar/blob/lh-fix-rate-limiting-connection-timeouts/microbench/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucketBenchmark.java) and [DefaultMonotonicSnapshotClockBenchmark](https://github.com/lhotari/pulsar/blob/lh-fix-rate-limiting-connection-timeouts/microbench/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClockBenchmark.java) results with JMH demonstrate high throughput (test results with Apple M3 Max):

```
Benchmark                                                                      Mode  Cnt            Score             Error  Units
AsyncTokenBucketBenchmark.consumeTokensBenchmark001Threads                    thrpt    3    235889973.656 ±    20309920.612  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark010Threads                    thrpt    3   2092331094.972 ±     4154666.893  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark100Threads                    thrpt    3   2423947092.794 ±  1185164011.463  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos001Threads                 thrpt    3   2031791421.024 ±   281179462.872  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos010Threads                 thrpt    3  19192224651.486 ±  3492465614.920  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos100Threads                 thrpt    3  22500950540.764 ± 15333188461.467  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot001Threads  thrpt    3     71259586.362 ±     7340997.362  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot010Threads  thrpt    3     22880953.901 ±     1521620.667  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot100Threads  thrpt    3     16236151.209 ±     5642180.371  ops/s
```

Results with Dell XPS 15 7590 i9-9980HK on Linux:
```
Benchmark                                                                      Mode  Cnt            Score            Error  Units
AsyncTokenBucketBenchmark.consumeTokensBenchmark001Threads                    thrpt    3     54622720.701 ±    4953944.847  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark010Threads                    thrpt    3    361746065.391 ±  155306621.346  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark100Threads                    thrpt    3    410425438.712 ±  524401828.340  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos001Threads                 thrpt    3   1776174139.570 ±  604272160.441  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos010Threads                 thrpt    3  10480409935.377 ± 3681168439.649  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos100Threads                 thrpt    3  10572683864.234 ±  783977752.777  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot001Threads  thrpt    3     30378314.738 ±   11642686.027  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot010Threads  thrpt    3    222770085.250 ±   42935412.356  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot100Threads  thrpt    3    287689452.858 ±  153899991.642  ops/s
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
@@ -363,7 +363,7 @@ public abstract class AsyncTokenBucket {
      * @param forceUpdateTokens if true, the token balance is updated before the comparison
      * @return true if the bucket contains tokens, false otherwise
      */
-    public boolean containsTokens(boolean forceUpdateTokens) {
+    protected boolean containsTokens(boolean forceUpdateTokens) {
         return tokens(forceUpdateTokens) > 0;
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
@@ -119,14 +119,14 @@ public abstract class AsyncTokenBucket {
      */
     private final LongAdder pendingConsumedTokens = new LongAdder();
 
-    private final boolean getTokensUpdatesTokens;
+    private final boolean consistentTokensView;
 
     protected AsyncTokenBucket(MonotonicSnapshotClock clockSource, long resolutionNanos,
-                               boolean getTokensUpdatesTokens) {
+                               boolean consistentTokensView) {
         this.clockSource = clockSource;
         this.resolutionNanos = resolutionNanos;
         this.lastNanos = Long.MIN_VALUE;
-        this.getTokensUpdatesTokens = getTokensUpdatesTokens;
+        this.consistentTokensView = consistentTokensView;
     }
 
     public static FinalRateAsyncTokenBucketBuilder builder() {
@@ -334,10 +334,10 @@ public abstract class AsyncTokenBucket {
     /**
      * Returns the current number of tokens in the bucket.
      * The token balance is updated if the configured resolutionNanos has passed since the last update unless
-     * getTokensUpdatesTokens is true.
+     * consistentTokensView is true.
      */
     public final long getTokens() {
-        return tokens(getTokensUpdatesTokens);
+        return tokens(consistentTokensView);
     }
 
     public abstract long getRate();
@@ -346,12 +346,12 @@ public abstract class AsyncTokenBucket {
      * Checks if the bucket contains tokens.
      * The token balance is updated before the comparison if the configured resolutionNanos has passed since the last
      * update. It's possible that the returned result is not definite since the token balance is eventually consistent
-     * if getTokensUpdatesTokens is false.
+     * if consistentTokensView is false.
      *
      * @return true if the bucket contains tokens, false otherwise
      */
     public boolean containsTokens() {
-        return containsTokens(getTokensUpdatesTokens);
+        return containsTokens(consistentTokensView);
     }
 
     /**
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
@@ -219,17 +219,19 @@ public abstract class AsyncTokenBucket {
      * @return the number of new tokens to add since the last update
      */
     private long calculateNewTokensSinceLastUpdate(long currentNanos) {
+        long previousLastNanos = lastNanos;
+        long newLastNanos;
+        // update lastNanos only if at least resolutionNanos/2 nanoseconds has passed since the last update
+        if (currentNanos >= previousLastNanos + resolutionNanos / 2) {
+            newLastNanos = currentNanos;
+        } else {
+            newLastNanos = previousLastNanos;
+        }
         long newTokens;
-        long previousLastNanos = LAST_NANOS_UPDATER.getAndUpdate(this,
-                currentLastNanos -> {
-                    // update lastNanos only if at least resolutionNanos/2 nanoseconds has passed since the last update
-                    if (currentNanos >= currentLastNanos + resolutionNanos / 2) {
-                        return currentNanos;
-                    } else {
-                        return currentLastNanos;
-                    }
-                });
-        if (previousLastNanos == Long.MIN_VALUE || currentNanos != lastNanos) {
+        if (newLastNanos == previousLastNanos
+                // prevent races with a CAS update of lastNanos
+                || !LAST_NANOS_UPDATER.compareAndSet(this, previousLastNanos, newLastNanos)
+                || previousLastNanos == Long.MIN_VALUE) {
             newTokens = 0;
         } else {
             long durationNanos = currentNanos - previousLastNanos + REMAINDER_NANOS_UPDATER.getAndSet(this, 0);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
@@ -42,6 +42,10 @@ import java.util.concurrent.atomic.LongAdder;
  * connection or client from the throttling queue to unthrottle. Before unthrottling, the application should check
  * for available tokens. If tokens are still not available, the application should continue with throttling and
  * repeat the throttling loop.
+ * <p>By default, the AsyncTokenBucket is eventually consistent. This means that the token balance is updated
+ * with added tokens and consumed tokens at most once during each "increment", when time advances more than the
+ * configured resolution. There are settings for configuring consistency, please see {@link AsyncTokenBucketBuilder}
+ * for details.
  * <p>This class does not produce side effects outside its own scope. It functions similarly to a stateful function,
  * akin to a counter function. In essence, it is a sophisticated counter. It can serve as a foundational component for
  * constructing higher-level asynchronous rate limiter implementations, which require side effects for throttling.
@@ -119,14 +123,28 @@ public abstract class AsyncTokenBucket {
      */
     private final LongAdder pendingConsumedTokens = new LongAdder();
 
-    private final boolean consistentTokensView;
+    /**
+     * By default, AsyncTokenBucket is eventually consistent. This means that the consumed tokens are subtracted from
+     * the total amount of tokens at most once during each "increment", when time advances more than the configured
+     * resolution. This setting determines if the consumed tokens are subtracted from tokens balance consistently.
+     * For high performance, it is recommended to keep this setting as false.
+     */
+    private final boolean consistentConsumedTokens;
+    /**
+     * By default, AsyncTokenBucket is eventually consistent. This means that the added tokens are calculated based
+     * on elapsed time at most once during each "increment", when time advances more than the configured
+     * resolution. This setting determines if the added tokens are calculated and added to tokens balance consistently.
+     * For high performance, it is recommended to keep this setting as false.
+     */
+    private final boolean consistentAddedTokens;
 
     protected AsyncTokenBucket(MonotonicSnapshotClock clockSource, long resolutionNanos,
-                               boolean consistentTokensView) {
+                               boolean consistentConsumedTokens, boolean consistentAddedTokens) {
         this.clockSource = clockSource;
         this.resolutionNanos = resolutionNanos;
         this.lastNanos = Long.MIN_VALUE;
-        this.consistentTokensView = consistentTokensView;
+        this.consistentConsumedTokens = consistentConsumedTokens;
+        this.consistentAddedTokens = consistentAddedTokens;
     }
 
     public static FinalRateAsyncTokenBucketBuilder builder() {
@@ -144,40 +162,46 @@ public abstract class AsyncTokenBucket {
     /**
      * Consumes tokens and possibly updates the tokens balance. New tokens are calculated and added to the current
      * tokens balance each time the update takes place. The update takes place once in every interval of the configured
-     * resolutionNanos or when the forceUpdateTokens parameter is true.
+     * resolutionNanos or when the forceConsistentTokens parameter is true.
      * When the tokens balance isn't updated, the consumed tokens are added to the pendingConsumedTokens LongAdder
      * counter which gets flushed the next time the tokens are updated. This makes the tokens balance
      * eventually consistent. The reason for this design choice is to optimize performance by preventing CAS loop
      * contention which could cause excessive CPU consumption.
      *
      * @param consumeTokens     number of tokens to consume, can be 0 to update the tokens balance
-     * @param forceUpdateTokens if true, the tokens are updated even if the configured resolution hasn't passed
+     * @param forceConsistentTokens if true, the token balance is updated consistently
      * @return the current number of tokens in the bucket or Long.MIN_VALUE when the number of tokens is unknown due
      * to eventual consistency
      */
-    private long consumeTokensAndMaybeUpdateTokensBalance(long consumeTokens, boolean forceUpdateTokens) {
+    private long consumeTokensAndMaybeUpdateTokensBalance(long consumeTokens, boolean forceConsistentTokens) {
         if (consumeTokens < 0) {
             throw new IllegalArgumentException("consumeTokens must be >= 0");
         }
-        long currentNanos = clockSource.getTickNanos(forceUpdateTokens);
+        boolean requestConsistentTickNanosSnapshot =
+                consistentAddedTokens || consistentConsumedTokens || forceConsistentTokens || resolutionNanos == 0;
+        long currentNanos = clockSource.getTickNanos(requestConsistentTickNanosSnapshot);
         long newTokens = 0;
         // check if the tokens should be updated immediately
-        if (shouldUpdateTokensImmediately(currentNanos, forceUpdateTokens)) {
+        if (shouldAddTokensImmediately(currentNanos, forceConsistentTokens)) {
             // calculate the number of new tokens since the last update
-            newTokens = calculateNewTokensSinceLastUpdate(currentNanos);
+            newTokens = calculateNewTokensSinceLastUpdate(currentNanos, forceConsistentTokens);
         }
         // update tokens if there are new tokens or if resolutionNanos is set to 0 which is currently used for testing
-        if (newTokens > 0 || resolutionNanos == 0) {
+        if (newTokens > 0 || resolutionNanos == 0 || consistentConsumedTokens || forceConsistentTokens) {
             // flush the pendingConsumedTokens by calling "sumThenReset"
             long currentPendingConsumedTokens = pendingConsumedTokens.sumThenReset();
             // calculate the token delta by subtracting the consumed tokens from the new tokens
             long tokenDelta = newTokens - currentPendingConsumedTokens;
-            // update the tokens and return the current token value
-            return TOKENS_UPDATER.updateAndGet(this,
-                    // limit the tokens to the capacity of the bucket
-                    currentTokens -> Math.min(currentTokens + tokenDelta, getCapacity())
-                            // subtract the consumed tokens from the capped tokens
-                            - consumeTokens);
+            if (tokenDelta != 0 || consumeTokens != 0) {
+                // update the tokens and return the current token value
+                return TOKENS_UPDATER.updateAndGet(this,
+                        // limit the tokens to the capacity of the bucket
+                        currentTokens -> Math.min(currentTokens + tokenDelta, getCapacity())
+                                // subtract the consumed tokens from the capped tokens
+                                - consumeTokens);
+            } else {
+                return tokens;
+            }
         } else {
             // eventual consistent fast path, tokens are not updated immediately
 
@@ -186,13 +210,8 @@ public abstract class AsyncTokenBucket {
                 pendingConsumedTokens.add(consumeTokens);
             }
 
-            if (forceUpdateTokens) {
-                // return the current tokens balance without updating the tokens and resetting the pendingConsumedTokens
-                return tokens - pendingConsumedTokens.sum();
-            } else {
-                // return Long.MIN_VALUE if the current value of tokens is unknown due to the eventual consistency
-                return Long.MIN_VALUE;
-            }
+            // return Long.MIN_VALUE if the current value of tokens is unknown due to the eventual consistency
+            return Long.MIN_VALUE;
         }
     }
 
@@ -201,19 +220,19 @@ public abstract class AsyncTokenBucket {
      *
      * The tokens will be updated once every resolutionNanos nanoseconds.
      * This method checks if the configured resolutionNanos has passed since the last update.
-     * If the forceUpdateTokens is true, the tokens will be updated immediately.
+     * If the forceConsistentTokens is true, the tokens will be updated immediately.
      *
-     * @param currentNanos the current monotonic clock time in nanoseconds
-     * @param forceUpdateTokens if true, the tokens will be updated immediately
+     * @param currentNanos      the current monotonic clock time in nanoseconds
+     * @param forceConsistentTokens if true, the tokens are added even if the configured resolution hasn't fully passed
      * @return true if the tokens should be updated immediately, false otherwise
      */
-    private boolean shouldUpdateTokensImmediately(long currentNanos, boolean forceUpdateTokens) {
+    private boolean shouldAddTokensImmediately(long currentNanos, boolean forceConsistentTokens) {
         long currentIncrement = resolutionNanos != 0 ? currentNanos / resolutionNanos : 0;
         long currentLastIncrement = lastIncrement;
         return currentIncrement == 0
                 || (currentIncrement > currentLastIncrement
                 && LAST_INCREMENT_UPDATER.compareAndSet(this, currentLastIncrement, currentIncrement))
-                || forceUpdateTokens;
+                || consistentAddedTokens || forceConsistentTokens;
     }
 
     /**
@@ -223,11 +242,13 @@ public abstract class AsyncTokenBucket {
      * @param currentNanos the current monotonic clock time in nanoseconds
      * @return the number of new tokens to add since the last update
      */
-    private long calculateNewTokensSinceLastUpdate(long currentNanos) {
+    private long calculateNewTokensSinceLastUpdate(long currentNanos, boolean forceConsistentTokens) {
         long previousLastNanos = lastNanos;
         long newLastNanos;
         // update lastNanos only if at least resolutionNanos/2 nanoseconds has passed since the last update
-        if (currentNanos >= previousLastNanos + resolutionNanos / 2) {
+        // unless consistency is needed
+        long minimumIncrementNanos = forceConsistentTokens || consistentAddedTokens ? 0L : resolutionNanos / 2;
+        if (currentNanos > previousLastNanos + minimumIncrementNanos) {
             newLastNanos = currentNanos;
         } else {
             newLastNanos = previousLastNanos;
@@ -291,15 +312,14 @@ public abstract class AsyncTokenBucket {
     }
 
     /**
-     * Returns the current token balance. When forceUpdateTokens is true, the tokens balance is updated before
-     * returning. If forceUpdateTokens is false, the tokens balance could be updated if the last updated happened
+     * Returns the current token balance. When forceConsistentTokens is true, the tokens balance is updated before
+     * returning. If forceConsistentTokens is false, the tokens balance could be updated if the last updated happened
      * more than resolutionNanos nanoseconds ago.
      *
-     * @param forceUpdateTokens if true, the tokens balance is updated before returning
      * @return the current token balance
      */
-    protected long tokens(boolean forceUpdateTokens) {
-        long currentTokens = consumeTokensAndMaybeUpdateTokensBalance(0, forceUpdateTokens);
+    private long tokens() {
+        long currentTokens = consumeTokensAndMaybeUpdateTokensBalance(0, false);
         if (currentTokens != Long.MIN_VALUE) {
             // when currentTokens isn't Long.MIN_VALUE, the current tokens balance is known
             return currentTokens;
@@ -319,7 +339,7 @@ public abstract class AsyncTokenBucket {
         long currentTokens = consumeTokensAndMaybeUpdateTokensBalance(0, true);
         if (currentTokens == Long.MIN_VALUE) {
             throw new IllegalArgumentException(
-                    "Unexpected result from updateAndConsumeTokens with forceUpdateTokens set to true");
+                    "Unexpected result from updateAndConsumeTokens with forceConsistentTokens set to true");
         }
         if (currentTokens > 0) {
             return 0L;
@@ -334,10 +354,10 @@ public abstract class AsyncTokenBucket {
     /**
      * Returns the current number of tokens in the bucket.
      * The token balance is updated if the configured resolutionNanos has passed since the last update unless
-     * consistentTokensView is true.
+     * consistentConsumedTokens is true.
      */
     public final long getTokens() {
-        return tokens(consistentTokensView);
+        return tokens();
     }
 
     public abstract long getRate();
@@ -346,25 +366,12 @@ public abstract class AsyncTokenBucket {
      * Checks if the bucket contains tokens.
      * The token balance is updated before the comparison if the configured resolutionNanos has passed since the last
      * update. It's possible that the returned result is not definite since the token balance is eventually consistent
-     * if consistentTokensView is false.
+     * if consistentConsumedTokens is false.
      *
      * @return true if the bucket contains tokens, false otherwise
      */
     public boolean containsTokens() {
-        return containsTokens(consistentTokensView);
-    }
-
-    /**
-     * Checks if the bucket contains tokens.
-     * The token balance is updated before the comparison if the configured resolutionNanos has passed since the last
-     * update. The token balance is also updated when forceUpdateTokens is true.
-     * It's possible that the returned result is not definite since the token balance is eventually consistent.
-     *
-     * @param forceUpdateTokens if true, the token balance is updated before the comparison
-     * @return true if the bucket contains tokens, false otherwise
-     */
-    protected boolean containsTokens(boolean forceUpdateTokens) {
-        return tokens(forceUpdateTokens) > 0;
+        return tokens() > 0;
     }
 
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
@@ -119,10 +119,14 @@ public abstract class AsyncTokenBucket {
      */
     private final LongAdder pendingConsumedTokens = new LongAdder();
 
-    protected AsyncTokenBucket(MonotonicSnapshotClock clockSource, long resolutionNanos) {
+    private final boolean getTokensUpdatesTokens;
+
+    protected AsyncTokenBucket(MonotonicSnapshotClock clockSource, long resolutionNanos,
+                               boolean getTokensUpdatesTokens) {
         this.clockSource = clockSource;
         this.resolutionNanos = resolutionNanos;
         this.lastNanos = Long.MIN_VALUE;
+        this.getTokensUpdatesTokens = getTokensUpdatesTokens;
     }
 
     public static FinalRateAsyncTokenBucketBuilder builder() {
@@ -328,10 +332,11 @@ public abstract class AsyncTokenBucket {
 
     /**
      * Returns the current number of tokens in the bucket.
-     * The token balance is updated if the configured resolutionNanos has passed since the last update.
+     * The token balance is updated if the configured resolutionNanos has passed since the last update unless
+     * getTokensUpdatesTokens is true.
      */
     public final long getTokens() {
-        return tokens(false);
+        return tokens(getTokensUpdatesTokens);
     }
 
     public abstract long getRate();
@@ -339,12 +344,13 @@ public abstract class AsyncTokenBucket {
     /**
      * Checks if the bucket contains tokens.
      * The token balance is updated before the comparison if the configured resolutionNanos has passed since the last
-     * update. It's possible that the returned result is not definite since the token balance is eventually consistent.
+     * update. It's possible that the returned result is not definite since the token balance is eventually consistent
+     * if getTokensUpdatesTokens is false.
      *
      * @return true if the bucket contains tokens, false otherwise
      */
     public boolean containsTokens() {
-        return containsTokens(false);
+        return containsTokens(getTokensUpdatesTokens);
     }
 
     /**
@@ -356,7 +362,7 @@ public abstract class AsyncTokenBucket {
      * @param forceUpdateTokens if true, the token balance is updated before the comparison
      * @return true if the bucket contains tokens, false otherwise
      */
-    public boolean containsTokens(boolean forceUpdateTokens) {
+    protected boolean containsTokens(boolean forceUpdateTokens) {
         return tokens(forceUpdateTokens) > 0;
     }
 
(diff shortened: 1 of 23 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
```diff
@@ -45,45 +49,170 @@ public DefaultMonotonicSnapshotClock(long snapshotIntervalNanos, LongSupplier cl
...
+            long newSnapshotTickNanos = baseSnapshotTickNanos + durationSinceReference;
+
+            // reset the reference clock source value if the clock source value leaps backward or forward
+            if (newSnapshotTickNanos < previousSnapshotTickNanos - maxDelta
+                    || newSnapshotTickNanos > previousSnapshotTickNanos + maxDelta) {
+                referenceClockSourceValue = clockValue;
+                baseSnapshotTickNanos = previousSnapshotTickNanos;
+                if (!snapshotRequested) {
+                    // if the snapshot value is not requested, increment by the snapshot interval
+                    baseSnapshotTickNanos += snapshotIntervalNanos;
```
rdhabalia, line 158: here, we assume that `updateSnapshotTickNanos` gets invoked at every `snapshotIntervalNanos` time so, we are adding it to `baseSnapshotTickNanos`. 
instead of if we pass `snapshotIntervalNanos` as a variable to this method then method could be less error prone and can be also used in unit-test if needed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
```diff
@@ -45,45 +49,170 @@ public DefaultMonotonicSnapshotClock(long snapshotIntervalNanos, LongSupplier cl
...
+            }
+
+            // calculate the duration since the reference clock source value
+            // so that the snapshot value is always increasing and tolerates it when the clock source is not strictly
+            // monotonic across all CPUs and leaps backward or forward
+            long durationSinceReference = clockValue - referenceClockSourceValue;
+            long newSnapshotTickNanos = baseSnapshotTickNanos + durationSinceReference;
+
+            // reset the reference clock source value if the clock source value leaps backward or forward
+            if (newSnapshotTickNanos < previousSnapshotTickNanos - maxDelta
```
rdhabalia, line 152: now, as we are anyway, using single thread update, then can't we just avoid this complex check, and just directly use `System.nanoTime()` retrieving from the single CPU and use that incremented count?

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
```diff
@@ -45,45 +49,170 @@ public DefaultMonotonicSnapshotClock(long snapshotIntervalNanos, LongSupplier cl
...
+                    || newSnapshotTickNanos > previousSnapshotTickNanos + maxDelta) {
```
rdhabalia, line 153: btw, can we document what exactly we are doing with `baseSnapshotTickNanos` to explain the logic here as it's little hard to understand what is happening here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
```diff
@@ -134,7 +135,11 @@ private void unthrottleQueuedProducers(ScheduledExecutorService executor) {
             // unthrottle as many producers as possible while there are token available
             while ((throttlingDuration = calculateThrottlingDurationNanos()) == 0L
                     && (producer = unthrottlingQueue.poll()) != null) {
-                producer.decrementThrottleCount();
+                try {
+                    producer.decrementThrottleCount();
+                } catch (Exception e) {
+                    log.error("Failed to unthrottle producer {}", producer, e);
```
rdhabalia, line 141: worth to add topic name here

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
```diff
@@ -45,45 +49,170 @@ public DefaultMonotonicSnapshotClock(long snapshotIntervalNanos, LongSupplier cl
...
+                // this is very unlikely to happen, but it's better to log it in any case
+                LOG.error("Unexpected fatal error that stopped the clock.", t);
+            } finally {
+                LOG.info("DefaultMonotonicSnapshotClock's TickUpdaterThread stopped. {},tid={}", this, getId());
+                running = false;
+                notifyAllTickUpdated();
+            }
+        }
+
+        private void updateSnapshotTickNanos(boolean snapshotRequested) {
```
rdhabalia, line 133: it would be great if we can add unit-test for this method and try to get max code-coverage from unit-test as we can see high cyclomatic complexity in this method code.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
```diff
@@ -45,45 +49,170 @@ public DefaultMonotonicSnapshotClock(long snapshotIntervalNanos, LongSupplier cl
...
-            // report unexpected error since this would be a fatal error when the clock doesn't progress anymore
-            // this is very unlikely to happen, but it's better to log it in any case
-            LOG.error("Unexpected fatal error that stopped the clock.", t);
         }
     }
 
     @Override
     public void close() {
-        thread.interrupt();
+        tickUpdaterThread.interrupt();
```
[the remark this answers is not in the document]
lhotari: I don't think that there's a problem in using `.interrupt()` in this case. Internally, interrupt is a flag in the Thread class.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
lhotari: A single thread could get migrated from one CPU to another. In certain virtualized environments, the counter can leap forward or backward. It can also happen on real multi-core/multi-socket hardware with TSC clock source when the hardware doesn't support CPU instructions which are required to address challenges in synchronizing the TSC values across CPUs. Some of the low level details is explained in ["TSC hardware"](https://docs.kernel.org/virt/kvm/x86/timekeeping.html#tsc-hardware).

The purpose of this class is to address any such problems where the time leap forward or backward and removing this logic would make this solution pointless.

Although this logic might seem complex, the problem is also a very low level problem. The logic is performant and has been validated with JMH benchmarks.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
lhotari: I added a few tests and the line coverage is now 97% (73/75) and branch coverage is 90% (18/20) with the current DefaultMonotonicSnapshotClock. There's only 2 lines and 2 branches that aren't covered which are lines 196 and 209 where there's handling for InterruptedException. Adding test coverage for those lines isn't reasonable, so this means that max code coverage is reached.

Based on metrics, the complexity of the of method code isn't high since methods are fairly small.

image

for example, PersistentTopic contains methods which have high complexity according to metrics:
image

Metrics calculated with IntelliJ plugin "[MetricsReloaded](https://plugins.jetbrains.com/plugin/93-metricsreloaded)"

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
lhotari: The topic name is already included in producer's toString() method implementation.
https://github.com/apache/pulsar/blob/3d0625ba64294fb0fe7dafc27c7a34883b4be51b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java#L659-L662

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
lhotari: Sure, I'll add some more comments to describe the logic.

lhotari:
Although clock sources aren't necessarily monotonic in all platforms, it seems that the root cause of this issue is could be different. 
The DefaultMonotonicSnapshotClock was originally added to address a performance bottleneck in calling `System.nanoTime()` on MacOS. This wasn't a significant bottleneck on Linux platform. The solution was to update the value in a single thread and add eventual consistency to the rate limiter. The problem seen in issue #23920 could occur when this thread gets starved and the updates are delayed. This PR already contains 2 mitigations to this problem by modifying the thread priority to `MAX_PRIORITY` and by modifying the logic how new tokens are added to the bucket when the token update happens.
However, I'll need to find a solution where the problem would be properly avoided without adding contention since contention would cause the rate limiting solution to have a relatively high overhead.

lhotari:
> Although clock sources aren't necessarily monotonic in all platforms, it seems that the root cause of this issue is could be different. The DefaultMonotonicSnapshotClock was originally added to address a performance bottleneck in calling `System.nanoTime()` on MacOS. This wasn't a significant bottleneck on Linux platform. The solution was to update the value in a single thread and add eventual consistency to the rate limiter. The problem seen in issue #23920 could occur when this thread gets starved and the updates are delayed. This PR already contains 2 mitigations to this problem by modifying the thread priority to `MAX_PRIORITY` and by modifying the logic how new tokens are added to the bucket when the token update happens. However, I'll need to find a solution where the problem would be properly avoided without adding contention since contention would cause the rate limiting solution to have a relatively high overhead.

I found a solution to the problem. It's possible to detect leaps backwards, but forward leaps should be unhandled since the tick updater thread could starve and time should be able to leap forward in that case when it finally gets updated.
Time leaping forward would cause issues with the changes in this PR. Before the changes made in this PR, there was an issue which caused negative tokens. I'll continue adding more tests to describe the scenario.

My assumption currently is that the minimal fix for #23920 is this diff:

```patch
diff --git a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
index ac9a1f03e59..36edb1c5d8a 100644
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
@@ -159,16 +159,15 @@ public abstract class AsyncTokenBucket {
         if (shouldUpdateTokensImmediately(currentNanos, forceUpdateTokens)) {
             // calculate the number of new tokens since the last update
             long newTokens = calculateNewTokensSinceLastUpdate(currentNanos);
-            // calculate the total amount of tokens to consume in this update
             // flush the pendingConsumedTokens by calling "sumThenReset"
-            long totalConsumedTokens = consumeTokens + pendingConsumedTokens.sumThenReset();
+            long currentPendingConsumedTokens = pendingConsumedTokens.sumThenReset();
             // update the tokens and return the current token value
             return TOKENS_UPDATER.updateAndGet(this,
-                    currentTokens ->
-                            // after adding new tokens, limit the tokens to the capacity
-                            Math.min(currentTokens + newTokens, getCapacity())
-                                    // subtract the consumed tokens
-                                    - totalConsumedTokens);
+                    // after adding new tokens subtract the pending consumed tokens and
+                    // limit the tokens to the capacity of the bucket
+                    currentTokens -> Math.min(currentTokens + newTokens - currentPendingConsumedTokens, getCapacity())
+                            // subtract the consumed tokens
+                            - consumeTokens);
         } else {
             // eventual consistent fast path, tokens are not updated immediately
```

The problem with the previous solution was that it would first add new tokens and cap it to the capacity. With the eventually consistency logic in AsyncTokenBucket, that's a wrong solution. I'll provide a separate explanation for this in later updates.

nodece:
> `System.nanoTime()` isn't strictly monotonic and consistent across multiple CPUs.

Perhaps this is not the root reason, many rate limiters use `System.nanoTime`, such as guava. We should avoid introducing a complex patch, even though the logic is clear. 

The minimal fix seems correct.

lhotari:
> Perhaps this is not the root reason, many rate limiters use `System.nanoTime`, such as guava. We should avoid introducing a complex patch, even though the logic is clear.
> 
> The minimal fix seems correct.

@nodece Just a reminder that this patch isn't complex based on code metrics. The problem domain is unusual and the feeling about complexity is mainly about unfamiliarity. This PR is still WIP and I'll try to get this into a reasonable state for final review.

lhotari:
I have also added a JMH benchmark for DefaultMonotonicSnapshotClock. Results on Apple Mac M3 Max:
```
Benchmark                                                                      Mode  Cnt            Score             Error  Units
DefaultMonotonicSnapshotClockBenchmark.getTickNanos001Threads                 thrpt    3   2033254223.572 ±    41581820.145  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos010Threads                 thrpt    3  18719597588.557 ±  2124798619.451  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanos100Threads                 thrpt    3  21049656355.267 ± 22701352214.776  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot001Threads  thrpt    3     70079972.669 ±     1667215.446  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot010Threads  thrpt    3     22865514.539 ±     1216715.428  ops/s
DefaultMonotonicSnapshotClockBenchmark.getTickNanosRequestSnapshot100Threads  thrpt    3     16422340.515 ±     2346690.916  ops/s
```
With the consistent mode (`requestSnapshot=true`), the throughput is about 16.4 M ops/s with 100 threads so the implementation scales very well for it's purpose. The AsyncTokenBucket implementation has a solution to minimize the calls for a consistent view to up to 1 call for a snapshot interval so this will scale very well.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
lhotari: I have updated the PR description with the rationale. As mentioned before, this code isn't complex based on code metrics. The problem domain is unusual and the feeling about complexity is mainly about unfamiliarity.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
lhotari: The code has been refactored and there are comments to explain the logic.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicSnapshotClock.java
lhotari: Code coverage is maxed currently for DefaultMonotonicSnapshotClock. 100% branch coverage and 97% line coverage (everything else except `currentThread().interrupt()` lines in catching `InterruptedException`).

lhotari:
This PR is now ready for final review. Please PTAL @rdhabalia @nodece. I'd appreciate accurate feedback about possible concerns about complexity since the based on code metrics this PR doesn't contain complex code. The problem domain is unusual and the feeling about complexity is mainly about unfamiliarity. I'll be happy to improve the description if this is the case. I'm confident that this PR is essential and important for the [PIP-322 Rate Limiting implementation](https://github.com/apache/pulsar/blob/master/pip/pip-322.md) that has been included in Pulsar since Pulsar 3.2.0.

dao-jun (approved):

lhotari:
I fixed the flaky throttling tests and moved them to broker-api group. There were multiple issues in the flaky tests.

nodece (approved):
Overall, LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/AsyncTokenBucket.java
```diff
@@ -177,8 +186,13 @@ private long consumeTokensAndMaybeUpdateTokensBalance(long consumeTokens, boolea
                 pendingConsumedTokens.add(consumeTokens);
             }
 
-            // return Long.MIN_VALUE if the current value of tokens is unknown due to the eventual consistency
-            return Long.MIN_VALUE;
+            if (forceUpdateTokens) {
+                // return the current tokens balance without updating the tokens and resetting the pendingConsumedTokens
+                return tokens - pendingConsumedTokens.sum();
```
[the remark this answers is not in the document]
lhotari: @nodece This code is now eliminated. I refactored AsyncTokenBucket to make the eventual consistency of it more understandable. There's javadoc in AsyncTokenBucketBuilder for the options to tune consistency since by default AsyncTokenBucket is by default eventually consistent so that the token balance is updated at most once during one "increment" which is configured with the resolutionNanos setting. By default it's 16 milliseconds.

PR #23931, opened 2025-02-05
[improve] [broker] Make the estimated entry size more accurate

poorbarcode:
### Motivation

**Background**
- `B1`: Broker assumes the entry size is `1` when no publishing after the topic was loaded up at the latest time. Otherwise, it uses the average entries size published during the lastest time-range of the topic loaded up.
  - Issue: the average entry size is not prectice for the catch-up read after a topic is reloaded. 
- `B2`: The feature `managedLedgerMaxReadsInFlightSize` assumes the entry size is `10kb` when it read nothing. Otherwise, it uses the average entries size of the entries read out before.

**Issue**
Regarding scenario `B2`, it may cause an error `Time-out elapsed while acquiring enough permits on the memory limiter to read from ledger 33, {tenant}/{ns}/persistent/{topic}-partition-0, estimated read size 1937 bytes for 1 entries (check managedLedgerMaxReadsInFlightSizeInMB)` when there is hundreds topic being read and the exact entry size is very little.

### More accurate solutions
- `P0`: Use the exact size of the entries that we attempt to read.
- `P1`: Use the exact size of the ledger that the entries were stored at to estimate the entries' size.
- `P2`: Use the size of the topic that the entries were stored at to estimate the entries' size.

### Modifications

- Rather than use the `P2` solution, change the solution to `P1`, and it will never return an inaccurate value such as `1` or `10k`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/ConsumerStatsTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/ConsumerStatsTest.java
@@ -445,8 +445,13 @@ public class ConsumerStatsTest extends ProducerConsumerBase {
                 .batchingMaxPublishDelay(5, TimeUnit.SECONDS)
                 .batchingMaxBytes(Integer.MAX_VALUE)
                 .create();
-
-        producer.send("first-message");
+        // The first messages deliver: 20 msgs.
+        // Average of "messages per batch" is "1".
+        for (int i = 0; i < 20; i++) {
+            producer.send("first-message");
+        }
+        // The second messages deliver: 20 msgs.
+        // Average of "messages per batch" is "Math.round(1 * 0.9 + 20 * 0.1) = 2.9 ～ 3".
         List<CompletableFuture<MessageId>> futures = new ArrayList<>();
         for (int i = 0; i < 20; i++) {
             futures.add(producer.sendAsync("message"));
@@ -480,10 +485,7 @@ public class ConsumerStatsTest extends ProducerConsumerBase {
         metadataConsumer.put("matchValueReschedule", "producer2");
         @Cleanup
         Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING).topic(topic).properties(metadataConsumer)
-                // Since https://github.com/apache/pulsar/pull/23931 improved the performance of delivery, the consumer
-                // will get more messages than before(it only receives 1 messages at the first delivery), we set queue
-                // size to `1` to keep the test passing.
-                .receiverQueueSize(1)
+                .receiverQueueSize(20)
                 .subscriptionName(subName).subscriptionInitialPosition(SubscriptionInitialPosition.Earliest).subscribe();
 
         int counter = 0;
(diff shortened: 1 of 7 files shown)


lhotari:
> * The feature `managedLedgerMaxReadsInFlightSize` assumes the entry size is `10kb` when it read nothing. Otherwise, it uses the average entries size of the entries read out before.

After #23901 changes this case would be very rare since it would apply only for the first read in case no entries were produced to the topic.

dao-jun (approved):

pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/ConsumerStatsTest.java
```diff
@@ -480,6 +480,10 @@ public void testAvgMessagesPerEntry() throws Exception {
         metadataConsumer.put("matchValueReschedule", "producer2");
         @Cleanup
         Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING).topic(topic).properties(metadataConsumer)
+                // Since https://github.com/apache/pulsar/pull/23931 improved the performance of delivery, the consumer
+                // will get more messages than before(it only receives 1 messages at the first delivery), we set queue
+                // size to `1` to keep the test passing.
+                .receiverQueueSize(1)
```
codelipenghui, line 486: I'm not sure why set queue to 1 will get the issue passed. But from the purpose of this test method, it looks like it should not related to the queue size. @poorbarcode Could you please provide more context about this change?

codelipenghui (commented):
The change looks good, left a minor comments about this test change.

pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/ConsumerStatsTest.java
poorbarcode: > Could you please provide more context about this change?

- In original design, the first read of a subscription only has one message, because broker calculate the avg message size by `msg read out / count`, so it read `1` message when `{msg read out}` is `0`
- This test always assumed the quantity of messages read out at the first time is `1`, see: https://github.com/apache/pulsar/pull/23931/files#diff-6e0b3de39cd005ecdedb9375327d65773640082cf70d44407ec79b2d553080b8R508-L505

codelipenghui: Ok, so how about change the below assert instead of adding the queue size config?

codelipenghui: I think that make sense to not let the test only test the consumer with 1 queue size.

poorbarcode: Sure

Technoboy- (approved):

pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/ConsumerStatsTest.java
poorbarcode: @codelipenghui 

- The main purpose of the test is the mechanism of `AvgMessagesPerEntry`: `Math.round( {pervious avg} * 0.9 + {new avg} * 0.1 )`
- Without a param `queueSize`, we need at least send `1000` messages 😂 
- I changed the `queueSize` to `20` and changed the test

codelipenghui (approved):

lhotari:
@poorbarcode Great work, Yubiao! I can backport and cherry-pick this to branch-3.3 and branch-3.0.

PR #23940, opened 2025-02-06
[improve][ml] Do not switch thread to execute asyncAddEntry's core logic

BewareMyPower:
### Motivation

In protocol handler's implementation, sometimes it needs to record the current base offset before the actual `asyncAddEntry` call (or the wrapped `PersistentTopic#publishMessages` call), for example:

```java
    private final PersistentTopic persistentTopic;
    private final Set<Long> pendingBaseOffsets = ConcurrentHashMap.newKeySet();

    private synchronized void add(ByteBuf buffer, int batchSize) {
        final var ml = (ManagedLedgerImpl) persistentTopic.getManagedLedger();
        final var interceptor = (ManagedLedgerInterceptorImpl) ml.getManagedLedgerInterceptor();
        final var baseOffset = interceptor.getIndex(); // the base offset of the next batch to write
        pendingBaseOffsets.add(baseOffset);
        ml.asyncAddEntry(buffer, batchSize, new AsyncCallbacks.AddEntryCallback() {

            @Override
            public void addComplete(Position position, ByteBuf entryData, Object ctx) {
                if (!pendingBaseOffsets.remove(baseOffset)) {
                    log.error("Failed to remove {}", baseOffset);
                }
```

However, the existing implementation won't work as expected that even if the downstream protocol handler guarantees that `asyncAddEntry` is called in a single thread. See https://github.com/apache/pulsar/blob/a19eaa2b972ab8fe1a2a45e65df1566cf28fb336/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L804-L809

First, the code above does not make sense because `internalAsyncAddEntry` itself is already synchronized.

Second, to guarantee thread safety for concurrent add operations, switching to another thread is not more efficient than synchronizing the method call. Mostly, it's less efficient because `execute` involves with at least two method calls (`offer` and `poll`) on a `BlockingQueue`, as well as some other operations (like CAS). Take `ArrayBlockingQueue` as example, both its `offer` and `poll` implementations need to acquire the internal lock.

Third, switching to another thread to execute the core logic is anti-intuitive, especially for modifying some fields that represent the current states. For example, to achieve the same goal at the beginning, I have to write the code like:

```java
        ml.getExecutor().execute(() -> {
            pendingBaseOffsets.add(baseOffset);
            ml.asyncAddEntry(buffer, batchSize, callback, ctx);
        });
```

### Modifications

Split `internalAsyncAddEntry` to two methods `beforeAddEntryToQueue` and `afterAddEntryToQueue` that are called before and after adding the operation to `pendingAddEntries`. Then simplify the code by throwing an exception and pass the exception to the callback. It's also more efficient because `OpAddEntry#failed` won't be called in the synchronized block.

Add `testBeforeAddEntry` to protect the change.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -804,29 +804,39 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
         // Jump to specific thread to avoid contention from writers writing from different threads
         final var addOperation = OpAddEntry.createNoRetainBuffer(this, buffer, numberOfMessages, callback, ctx,
                 currentLedgerTimeoutTriggered);
-        internalAsyncAddEntry(addOperation);
+        var added = false;
+        try {
+            // Use synchronized to ensure if `addOperation` is added to queue and fails later, it will be the first
+            // element in `pendingAddEntries`.
+            synchronized (this) {
+                if (managedLedgerInterceptor != null) {
+                    managedLedgerInterceptor.beforeAddEntry(addOperation, addOperation.getNumberOfMessages());
+                }
+                final var state = STATE_UPDATER.get(this);
+                beforeAddEntryToQueue(state);
+                pendingAddEntries.add(addOperation);
+                added = true;
+                afterAddEntryToQueue(state, addOperation);
+            }
+        } catch (Throwable throwable) {
+            if (!added) {
+                addOperation.failed(ManagedLedgerException.getManagedLedgerException(throwable));
+            } // else: all elements of `pendingAddEntries` will fail in another thread
+        }
     }
 
-    protected synchronized void internalAsyncAddEntry(OpAddEntry addOperation) {
-        if (!beforeAddEntry(addOperation)) {
-            return;
-        }
-        final State state = STATE_UPDATER.get(this);
+    protected void beforeAddEntryToQueue(State state) throws ManagedLedgerException {
         if (state.isFenced()) {
-            addOperation.failed(new ManagedLedgerFencedException());
-            return;
-        } else if (state == State.Terminated) {
-            addOperation.failed(new ManagedLedgerTerminatedException("Managed ledger was already terminated"));
-            return;
-        } else if (state == State.Closed) {
-            addOperation.failed(new ManagedLedgerAlreadyClosedException("Managed ledger was already closed"));
-            return;
-        } else if (state == State.WriteFailed) {
-            addOperation.failed(new ManagedLedgerAlreadyClosedException("Waiting to recover from failure"));
-            return;
+            throw new ManagedLedgerFencedException();
         }
-        pendingAddEntries.add(addOperation);
+        switch (state) {
+            case Terminated -> throw new ManagedLedgerTerminatedException("Managed ledger was already terminated");
+            case Closed -> throw new ManagedLedgerAlreadyClosedException("Managed ledger was already closed");
+            case WriteFailed -> throw new ManagedLedgerAlreadyClosedException("Waiting to recover from failure");
+        }
+    }
 
+    protected void afterAddEntryToQueue(State state, OpAddEntry addOperation) throws ManagedLedgerException {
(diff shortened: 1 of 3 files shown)


dao-jun:
> In protocol handler's implementation, sometimes it needs to record the current base offset before the actual asyncAddEntry call (or the wrapped PersistentTopic#publishMessages call)

I don't understand this part, could you pls explain more clear?

BewareMyPower:
@dao-jun I found this issue when I implemented the Kafka transaction. In Kafka, `analyzeAndValidateProducerState` is called before writing messages, this method will update a map that maps the next message's offset (key) to an object and it will remove the key after the record batch is written.

For example, assuming there are 3 ongoing writes, the flow looks like the following pseudo code:

```java
for (int i = 0; i < 3; i++) {
    long nextOffset = interceptor.getIndex() + 1; // LEO
    ongoingTxns.put(nextOffset, txn); // record the ongoing txn of this record batch
    asyncAddEntry().thenAccept(__ -> ongoingTxns.remove(nextOffset));
}
```

Assuming each record batch has only 1 message, ideally, after these 3 writes, `ongoingTxns` will have 3 keys (0, 1, 2). However, since `asyncAddEntry` switches thread to the ML's executor to call interceptor's `beforeAddEntry` method, there is a chance that `nextOffset` is always 0 in these 3 loops and `ongoingTxns.remove(nextOffset)` will take effect only once. Then it could hit [here](https://github.com/streamnative/kop/blob/14f9832d8a437c5b717e3c74fd7e4cda9810aff7/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManager.java#L312) in the actual code.

The details above are beyond the scope of this PR but could help you understand the motivation.

dao-jun:
> @dao-jun I found this issue when I implemented the Kafka transaction. In Kafka, `analyzeAndValidateProducerState` is called before writing messages, this method will update a map that maps the next message's offset (key) to an object and it will remove the key after the record batch is written.
> 
> For example, assuming there are 3 ongoing writes, the flow looks like the following pseudo code:
> 
> ```java
> for (int i = 0; i < 3; i++) {
>     long nextOffset = interceptor.getIndex() + 1; // LEO
>     ongoingTxns.put(nextOffset, txn); // record the ongoing txn of this record batch
>     asyncAddEntry().thenAccept(__ -> ongoingTxns.remove(nextOffset));
> }
> ```
> 
> Assuming each record batch has only 1 message, ideally, after these 3 writes, `ongoingTxns` will have 3 keys (0, 1, 2). However, since `asyncAddEntry` switches thread to the ML's executor to call interceptor's `beforeAddEntry` method, there is a chance that `nextOffset` is always 0 in these 3 loops and `ongoingTxns.remove(nextOffset)` will take effect only once. Then it could hit [here](https://github.com/streamnative/kop/blob/14f9832d8a437c5b717e3c74fd7e4cda9810aff7/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManager.java#L312) in the actual code, which is much different from the OSS KoP.
> 
> The details above are beyond the scope of this PR but could help you understand the motivation.

Thanks for your explain!

dao-jun (approved):

BewareMyPower:
Some shadow topic related tests failed, I will fix them

lhotari:
@BewareMyPower Please also remove the comment "Jump to specific thread to avoid contention from writers writing from different threads" that is now obsolete: https://github.com/apache/pulsar/blob/cc7b3816ce14a2ddff18dbf04216769fdb8751db/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L804

lhotari:
@BewareMyPower This logic doesn't either make sense any more after the changes:
https://github.com/apache/pulsar/blob/cc7b3816ce14a2ddff18dbf04216769fdb8751db/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L801-L806

Issue #15966, opened 2022-06-07
PIP-175: Extend time based release process


merlimat:
## Motivation

In PIP-47 (https://github.com/apache/pulsar/wiki/PIP-47:-Time-Based-Release-Plan), we have adopted a time-based release plan. This was the first attempt at establishing a new principle on how releases should b

The main two benefits of this approach have been: 

 1. Clarity for users and developers on when to expect a release
 2. Breaking a hard relationship between feature and release: a particular feature will be included in the release if it is completed in time. Otherwise, it will be bubbled up to the next release.

The motivation for the current proposal is to extend the existing process to address the issues that we have seen and that were left out of the scope of PIP-47.

### Short maintenance cycles for releases

Since we're doing a 3 months release cycle, we are ending with 4 releases done per year, even though it's more close to 3 releases. 

There is a high cost to maintain a lot of old releases, backport bug fixes, and security patches. In general, we actively support the last 3 minor releases while continuing to develop the next release. E.g., 2.8, 2.9, and 2.10, while 2.11 is under development.

The result is that a user adopting a particular release is forced to upgrade in a < 1-year timeframe to keep up to date and use a supported release. This timeframe is too short for many users as it imposes a lot of forced upgrades, for which they are not prepared in terms of available time and required effort.

### Live Upgrade/Downgrade compatibility path

In Pulsar, we guarantee that users have a way to do live upgrades and downgrades with zero downtime.

This is very powerful because it gives them the freedom to upgrade to a new release with the assurance of being able to roll back to the previous release in case any functional or performance regressions are encountered.

Today, this compatibility is guaranteed across minor versions. Eg: I can do  `2.7 -> 2.8 -> 2.7` as a live upgrade. 

What is not guaranteed is to "skip" releases. E.g.: `2.7 -> 2.9` might work or not, but it's not guaranteed. In that case an intermediated upgrade would be required: `2.7 -> 2.8 -> 2.9`.

The reasons for which the "skip" upgrade might not work are multiple:
  1. Incompatible upgrade of some dependency (e.g., ZooKeeper) that might not be compatible with an older version.
  2. Adoption of a new metadata format or data format on disk.
     Every time we introduce a new incompatible format change (outside of a regular Protobuf field addition), we do it in a 2 steps way: 
      - In a new release, we introduce the new feature/format, disabled by default. The new release can read both old and new formats, though it keeps writing the old format by default.
      - In a subsequent release, we change the default to the new format

[...]

PR #23942, opened 2025-02-06
[improve][client] Add newMessage with schema and transactions

omarkj:
Pulsar Client allows callers to create messages with a schema or a transaction, but not both.

This commit adds a new method in the producer that allows callers to create a message with both a schema and transaction.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
```diff
@@ -132,6 +132,21 @@ public interface Producer<T> extends Closeable {
...
+     *
+     * <p>After the transaction commit, it will be made visible to consumer.
+     *
+     * <p>After the transaction abort, it will never be visible to consumer.
+     *
+     * @return a typed message builder that can be used to construct the message to be sent through this producer
+     * @see #newMessage()
+     */
+    <V> TypedMessageBuilder<V> newMessage(Schema<V> schema,
+                                          Transaction txn);
```
BewareMyPower, line 148: This change adds a new public API, which requires a PIP. https://github.com/apache/pulsar/tree/master/pip#when-is-a-pip-required

BTW, for this implementation PR, you should add a test to ensure it works.

BewareMyPower (changes requested):

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
omarkj: Is there a test suite for the producer? I was unable to find one to add to.

dao-jun:
For PIP process, pls referring to https://github.com/apache/pulsar/tree/master/pip#pulsar-improvement-proposal-pip

omarkj:
PIP PR filed here: https://github.com/apache/pulsar/pull/23950, will follow those guidelines by emailing out to the thread soon, etc.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
BewareMyPower: No. Instead, I think you should add a test to `TransactionTest` to show which case this new API should be used. e.g.

```java
// producer is a Producer<String> while value is a String rather than byte[]
producer.newMessage(schema, txn).value(value).send();
```

lhotari (approved):
LGTM

lhotari:
@BewareMyPower @dao-jun The PIP has been approved and merged, https://github.com/apache/pulsar/blob/master/pip/pip-407.md . PTAL and review

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
lhotari, line 148: I have a concern about backward compatibility with this interface change. Adding a new method without a default implementation will break source compatibility for any custom implementations of the `Producer` interface.
While it might be uncommon for users to implement this interface directly, it could happen in testing scenarios or custom proxy implementations. Java source compatibility rules require all implementors to provide all interface methods.

We could address the source compatibility issue by adding a default implementation:

```suggestion
    default <V> TypedMessageBuilder<V> newMessage(Schema<V> schema, Transaction txn) {
        // to retain source compatibilty of Producer interface for implementors of the Producer interface
        throw new UnsupportedOperationException(
                String.format("Method not implemented in: %s", getClass().getName()));
    }
```

The benefit of making it source compatible is that we could also backport this change to the 4.0.x client without introducing breaking changes.

BewareMyPower (approved):

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
BewareMyPower: Do we really need to backport this change to 4.0.x client? LTS releases should not introduce such new APIs even if it does not break compatibility.

BewareMyPower: Once it's ported to 4.0.x, the forward compatibility will be broken. According to https://github.com/apache/pulsar/issues/15966, we should also guarantee the downgrade compatibility

dao-jun (approved):

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
BewareMyPower: Resolve this conversation. We can open another thread for the backport topic. For a new feature, adding a new non-default method should be accepted

lhotari: > Do we really need to backport this change to 4.0.x client? LTS releases should not introduce such new APIs even if it does not break compatibility.

This is a feature which would benefit existing 4.0.x users. By making the solution compatible without introducing breaking changes to the interface, we could deliver this feature to 4.0.x users immediately. Obviously there's the question whether we need to do this or not.

BewareMyPower: I reviewed the PIP-175 and LTS releases recently again. PIP-175 mainly talked about the server side compatibility. Regarding the client side,

> Note that this consideration is separate from the compatibility between clients and brokers, where we never break compatibility. The oldest available Pulsar client can still talk with the newest Pulsar broker, and vice versa, a new client, will be perfectly fine with an older broker (except the new features won't be working).

From the existing definitions, it's acceptable to cherry-pick this **client-side** feature for 4.0.x. It does not break any compatibility: with this new API, the latest 4.0.x client can still access an older broker (that supports transaction).

The only question is the source compatibility you mentioned by making it `default`. Could you show a concrete example for how could a non-default new method affect users?

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/TransactionTest.java
```diff
@@ -484,6 +485,42 @@ public void testAsyncSendOrAckForSingleFuture() throws Exception {
...
+                .build()
+                .get();
+
+        MessageId messageId = producer.newMessage(Schema.STRING, txn)
+                .value("testSendAndAckWithSchema")
+                .send();
+
+        txn.commit().get();
+
+        Message<byte[]> message = consumer.receive();
```
congbobo184, line 519: Could we add an Avro or another schema test? The bytes type is not sufficiently representative.

congbobo184 (changes requested):

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/TransactionTest.java
liangyepianzhou: +1

lhotari: @congbobo184 There's already `Schema.STRING`, why wouldn't that be representative? I'd like to get this PR merged since it's also required by upcoming PIP-439, #24704.

PR #23943, opened 2025-02-07
[fix] [client] call redeliver 1 msg but did 2 msgs

poorbarcode:
### Motivation

Issue: call redeliver `1` message, but it will redeliver `2` messages, you can reproduce the issue by the new test `testBatchMessageNAck`

### Modifications

fix the bug

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BatchMessageWithBatchIndexLevelTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BatchMessageWithBatchIndexLevelTest.java
@@ -59,6 +59,7 @@ import org.apache.pulsar.common.util.collections.BitSetRecyclable;
 import org.awaitility.Awaitility;
 import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 @Slf4j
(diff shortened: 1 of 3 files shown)


shibd (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2831,27 +2831,18 @@ private Optional<EncryptionContext> createEncryptionContext(MessageMetadata msgM
...
-                if (!messageIds.contains(id)) {
-                    messageIds.add(id);
-                    break;
-                }
-                message.release();
-                message = incomingMessages.poll();
+        Message<T> message;
+        while (true) {
+            message = incomingMessages.pollIf(msg -> {
+                MessageId idPoled = NegativeAcksTracker.discardBatchAndPartitionIndex(msg.getMessageId());
```
dao-jun, line 2837: idPolled?

dao-jun (approved):

Technoboy- (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
poorbarcode: Fixed

gaoran10 (approved):

nodece:
@poorbarcode Could you fix the build error on the branch-3.0?

```
import static com.google.common.base.Preconditions.checkState;
```
This import is not used.

poorbarcode:
@nodece 

> Could you fix the build error on the branch-3.0?
> ```
> import static com.google.common.base.Preconditions.checkState;
> ```
> This import is not used.

@lhotari has fixed, Thanks

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BatchMessageWithBatchIndexLevelTest.java
```diff
@@ -137,6 +138,67 @@ public void testBatchMessageAck() {
         });
     }
 
+    @DataProvider
+    public Object[][] enabledBatchSend() {
+        return new Object[][] {
+                {false},
```
BewareMyPower, line 144: When it's false, even without this change, `testBatchMessageNAck` could still pass

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BatchMessageWithBatchIndexLevelTest.java
```diff
@@ -137,6 +138,67 @@ public void testBatchMessageAck() {
...
+            if (enabledBatchSend) {
+                assertEquals(consumer.numMessagesInQueue(), 40);
+            } else {
+                assertEquals(consumer.numMessagesInQueue(), 21);
+            }
+        });
+
+        // Negative ack and verify result/
+        Message<byte[]> receive1 = consumer.receive();
+        consumer.pause();
```
BewareMyPower, line 189: It's meaningless here. `pause()` should only affect the message listener

lhotari: In the Java client, `pause()` will pause sending new permits, so it's not specific to message listener.
https://github.com/apache/pulsar/blob/1220951ac74fb4742abbbd331d6e751234c47015/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1867-L1880

dao-jun: `pause` is for stop receiving msgs from broker

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BatchMessageWithBatchIndexLevelTest.java
poorbarcode: Yes, it only affects bacthed messages

PR #23945, opened 2025-02-07
[improve][pip] PIP-406: Introduce metrics related to dispatch throttled events

shibd:
### Motivation
- Discuss email: https://lists.apache.org/thread/w9pq1jp3f22prsr97mxzj1yhdoo1cnc4
- Vote thread: https://lists.apache.org/thread/pmgkf67ph98k5l3lr60rtg96octb397q
- Implatement PR: https://github.com/apache/pulsar/pull/23946

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-406.md
+++ b/pip/pip-406.md
@@ -1,4 +1,4 @@
-# PIP-406: Introduce pulsar_subscription_dispatch_throttled_msgs and bytes metrics 
+# PIP-406: Introduce metrics related to dispatch_throttled_count
 
 # Background knowledge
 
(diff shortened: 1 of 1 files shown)


pip/pip-406.md
```diff
@@ -0,0 +1,75 @@
...
+this metric may not accurately reflect whether the backlog is due to insufficient consumer capacity, as it could be caused by dispatch throttling.
+
+## Goals
+
+Introduce metrics to indicate the number of `messages/bytes throttled` for a subscription. This allows users to write PromQL queries to identify subscriptions with high backlogs but low or no throttling, pinpointing backlogs caused by insufficient consumer capacity.
+
+## In Scope
+- Introduce the metric `pulsar_subscription_dispatch_throttled_msgs` to represent the total number of messages throttled for a subscription.
+- Introduce the metric `pulsar_subscription_dispatch_throttled_bytes` to represent the total number of bytes throttled for a subscription.
+- Add `dispatchThrottledMsgs` and `dispatchThrottledBytes` fields to topic subscription stats.
```
poorbarcode, line 18: Maybe `remaining_msg_permits` is more meaningful, because the value of  `pulsar_subscription_dispatch_throttled_msgs` will keep increasing once the throttling worked

pip/pip-406.md
```diff
@@ -0,0 +1,75 @@
...
+However, if [dispatch throttling](https://pulsar.apache.org/docs/next/concepts-throttling/) is configured at the broker/topic/subscription level,
+this metric may not accurately reflect whether the backlog is due to insufficient consumer capacity, as it could be caused by dispatch throttling.
+
+## Goals
+
+Introduce metrics to indicate the number of `messages/bytes throttled` for a subscription. This allows users to write PromQL queries to identify subscriptions with high backlogs but low or no throttling, pinpointing backlogs caused by insufficient consumer capacity.
+
+## In Scope
+- Introduce the metric `pulsar_subscription_dispatch_throttled_msgs` to represent the total number of messages throttled for a subscription.
+- Introduce the metric `pulsar_subscription_dispatch_throttled_bytes` to represent the total number of bytes throttled for a subscription.
```
poorbarcode, line 17: BTW, most of users will not enable subscription level metrics

pip/pip-406.md
shibd: We can't get real `remaining_msg_permits`, because if a subscription dispatch is throttled there may be limitations across three level (broker, topic, subscription), which can be **difficult** for users to assess. 

For example, if a broker has 1000 permits remaining, it might impose a restriction on this subscription, or it might not.

pip/pip-406.md
shibd: Yes, this metrics just work on enable subscitpion users.

In other words, it can accurately represent whether each subscription is being throttled, rather than just providing configurations for users to view.

pip/pip-406.md
shibd: > because the value of pulsar_subscription_dispatch_throttled_msgs will keep increasing once the throttling worked

Users can use something like Prometheus to fetch and store metrics every 30 seconds, making it easy to calculate increments.

pip/pip-406.md
```diff
@@ -0,0 +1,75 @@
...
++                }
++                return current + increment;
++            });
++        }
++        return result;
+     }
+```
+
+## Public-facing Changes
+- None
```
RobertIndie, line 67: Two new metrics have been added in this PIP. We should document them here.

shibd: Done

pip/pip-406.md
```diff
@@ -0,0 +1,152 @@
...
++            if (!localDispatchRateLimiterOnMessage.consumeTokensAndCheckIfContainsTokens(numberOfMessages)) {
++                dispatchThrottleMsgCount.increment();
++            }
+         }
+         AsyncTokenBucket localDispatchRateLimiterOnByte = dispatchRateLimiterOnByte;
+         if (byteSize > 0 && localDispatchRateLimiterOnByte != null) {
+-            localDispatchRateLimiterOnByte.consumeTokens(byteSize);
++            if (!localDispatchRateLimiterOnByte.consumeTokensAndCheckIfContainsTokens(byteSize)) {
++                dispatchThrottleBytesCount.increment();
++            }
```
lhotari, line 64: this isn't a correct place to check which got throttled. The tokens get consumed after the messages have been sent. Throttling happens here: https://github.com/apache/pulsar/blob/28f7845c93edc0dcbc068a71bec45ab4b3402d4e/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L565-L615

shibd: If the record is here will get huge throttled count, Because once rate limiting occurs, there will be multiple repeated requests in a short period.

lhotari: > If the record is here will get huge throttled count, Because once rate limiting occurs, there will be multiple repeated requests(asyncRead) in a short period.

@shibd What I meant to say earlier is that if you handle the metric this way, it's completely bogus metric. The reason for this is that when the tokens are consumed, the messages have already been sent. The metric would be tracking sent messages and bytes and not throttled messages and bytes.

I understand that when throttling occurs, the same messages might get throttled again. You would need to address that problem if you would like that this metric counts the same entry only once.

Duplication in the throttling code was recently eliminated and there's now a single location in the code base where the dispatch rate limiter throttling happens, it's here: https://github.com/apache/pulsar/blob/efcf7c27b3f0fa5dc0252ce46f54b28a897e08bd/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java#L417-L428

lhotari: @shibd I voted -1 on the PIP due to this issue with more details in the email message, https://lists.apache.org/thread/6447tlcvbrkszqnb24q5q4xsc4yc08zb .

shibd: Yes, from the name of the metric, it does seem ambiguous.

Initially, I added it during reading, but then I thought about it and realized that for metrics like `throttled_count`, it can only represent the `degree` of throttling.

So, I added it when `consuming tokens`, rather than before reading messages. If a limit throttles, then there will definitely be a shortage when consuming tokens, and we can then increment the metric.

Whether it solves the frequent reading issue wasn't my main focus at the time.

We can discuss:

Regarding the metric for message `throttling counts`, we can't obtain an exact value, right? Even with the new code, you've changed the logic: `delaying the re-read when messages can't be read at all`

However, if we count the times here, it can only represent a `degree` of throttling. This is because the metric's count is also `related to the number of reads`.

I don't have any strong suggestions(I think adding it in either place would be fine.). If you can explain more insights, I can learn from them.

For now, I'll follow your suggestion and add the metric during reading.

lhotari: > Initially, I added it during reading, but then I thought about it and realized that for metrics like `throttled_count`, it can only represent the `degree` of throttling.
> 
> So, I added it when `consuming tokens`, rather than before reading messages. If a limit throttles, then there will definitely be a shortage when consuming tokens, and we can then increment the metric.

Yes, that's completely understandable and logical.

> Regarding the metric for message throttling counts, we can't obtain an exact value, right? 

Do you mean the exact value of how many messages and bytes gets throttled? Yes, it doesn't seem to be possible since the dispatch rate limiter will set limits on how many messages could be read. At the time that the limit is set, we don't know whether there would be messages to be sent out.

> Even with the new code, you've changed the logic: delaying the re-read when messages can't be read at all

The new code in #24012 didn't really change the core logic in this area. It removes code duplication that was present in multiple dispatcher classes and consolidates the logic to a single method in AbstractBaseDispatcher.

> However, if we count the times here, it can only represent a `degree` of throttling. This is because the metric's count is also `related to the number of reads`.

Exactly.

> I don't have any strong suggestions(I think adding it in either place would be fine.). 

Yes, it's definitely valuable to add the counter to be able to observe when throttling is occuring.

> If you can explain more insights, I can learn from them.

Not directly related, but I just created #24036 about making the fixed 1-second backoff period configurable and what would need to be taken into account if we were to change that.

shibd: Agree with these. thanks. 

when you have time, please take another look at the PIP. I think we can vote again.

lhotari: @shibd You found a great solution! I like it. The only remaining problem is about naming in the PIP document. 

For example where it's "total count of bytes throttled", it is currently misleading.
> Introduce the metric `pulsar_broker_dispatch_throttled_bytes_count` to represent the total count of bytes throttled for a broker.

Instead of "total count of bytes throttled", it's actually "total number of times bytes were throttled". 
I'll check what Claude AI would suggest to resolve this. I strongly recommend to use AI assistant for editing PIP docs.

shibd: hi, @lhotari. Thanks for the review.

The incorrect names and statements have been revised. Please review.

For subscription metrics implement, your understanding is correct: **the subscription throttling metrics only represent the impact of SubscriptionRateLimit**. I intentionally designed it this way. The reasons are:

1. If it were designed to increase with subscription, topic, and broker impacts, users would find it difficult to determine which dimension's rate limit configuration is causing the issue.
2. In the current design, it can be understood this way: `when users see that the broker or topic level has already been throttled, the subscription throttling metrics become irrelevant, as it indicates that the broker or topic configuration is relatively small, and the first step should be addressing the broker or topic level configuration`

shibd: @lhotari Could you help look at the new comment?

RobertIndie (commented):
I think we don't need separate metrics for both `messages throttled` and `bytes throttled`. This could make it complex for users. Could we just name it `pulsar_dispatch_throttled_events` and only count the number of throttled events where the message count **or** bytes exceed the rate limits?

So the logic would be like:

```diff
     private boolean applyDispatchRateLimitsToReadLimits(DispatchRateLimiter rateLimiter,
                                                         MutablePair<Integer, Long> readLimits,
                                                         DispatchRateLimiter.Type limiterType) {
+        int originalMessagesToRead = readLimits.getLeft();
+        long originalBytesToRead = readLimits.getRight();
         // update messagesToRead according to available dispatch rate limit.
         int availablePermitsOnMsg = (int) rateLimiter.getAvailableDispatchRateLimitOnMsg();
         if (availablePermitsOnMsg >= 0) {
@@ -414,6 +416,12 @@ private boolean applyDispatchRateLimitsToReadLimits(DispatchRateLimiter rateLimi
         if (availablePermitsOnByte >= 0) {
             readLimits.setRight(Math.min(readLimits.getRight(), availablePermitsOnByte));
         }
+        if (readLimits.getLeft() < originalMessagesToRead || readLimits.getRight() < originalBytesToRead) {
+            rateLimiter.increaseDispatchThrottleMsgCount();
+        }
```

pip/pip-406.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Goals
+
+Introduce metrics to indicate the count of `messages/bytes throttled` for **broker/topic/subscription** level rate limit. 
+This allows users to write PromQL queries to identify subscriptions with high backlogs but low or no throttling, pinpointing backlogs caused by insufficient consumer capacity.
+
+## In Scope
+
+Broker Level:
+- Introduce the metric `pulsar_broker_dispatch_throttled_msg_count` to represent the total number of times message dispatching was throttled on a broker due to broker rate limits.
```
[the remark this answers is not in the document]
shibd: Yes, but I think we still need to keep the keyword 'dispatch'. 

How about the following names?

- pulsar_broker_dispatch_throttle_msg_events
- pulsar_broker_dispatch_throttle_byte_events

pip/pip-406.md
lhotari: @shibd I think that the problem hasn't been addressed. 

> For subscription metrics implement, your understanding is correct: **the subscription throttling metrics only represent the impact of SubscriptionRateLimit**. I intentionally designed it this way. The reasons are:
> 1. If it were designed to increase with subscription, topic, and broker impacts, users would find it difficult to determine which dimension's rate limit configuration is causing the issue.

This is a valid concern and there are ways to address your concern. 

> 2. In the current design, it can be understood this way: `when users see that the broker or topic level has already been throttled, the subscription throttling metrics become irrelevant, as it indicates that the broker or topic configuration is relatively small, and the first step should be addressing the broker or topic level configuration`

I don't think that this holds since "when users see that the broker or topic level has already been throttled" doesn't make sense in reality. How would a user "see" that? Metrics should be capturing the information so that there isn't a need to guess.

The metrics should be present to be able to notice whether a subscription gets throttled or not. The reason (subscription level, topic level, broker level) could be tracked separately. One simple way is to track all counters at the subscription level and aggregate them upwards to topic and broker level. This way there's no need to duplicate counters at runtime for stats. Many stats already have been implemented this way.

In dashboards, it's possible to calculate sums effectively directly from the subscription level metrics. There's no need to have Prometheus or OTel metrics for dispatch throttling at topic or broker level when the subscription level metric has labels which allow associating the values at topic or broker level.

shibd: > This is a valid concern and there are ways to address your concern.

How can we determine this? Initially, the design was as you mentioned (the subscription represented by all three dimensions). After discussing with @poorbarcode , we realized that this design makes it difficult for users to identify which dimension's configuration is affecting them.

> I don't think that this holds since "when users see that the broker or topic level has already been throttled" doesn't make sense in reality. How would a user "see" that? Metrics should be capturing the information so that there isn't a need to guess.

In the current implementation, users can monitor metrics for the three dimensions. Once it hits the expected threshold, they can modify the corresponding configuration. If the broker's configuration is low, adjust the broker; if the topic's configuration is low, adjust the topic.

> The metrics should be present to be able to notice whether a subscription gets throttled or not. The reason (subscription level, topic level, broker level) could be tracked separately. One simple way is to track all counters at the subscription level and aggregate them upwards to topic and broker level. This way there's no need to duplicate counters at runtime for stats. Many stats already have been implemented this way.

> In dashboards, it's possible to calculate sums effectively directly from the subscription level metrics. There's no need to have Prometheus or OTel metrics for dispatch throttling at topic or broker level when the subscription level metric has labels which allow associating the values at topic or broker level.

If we obtain the rate trottled metrics for the broker and topic through aggregation, it can't accurately represent whether the impact is due to broker or topic level configurations.


@poorbarcode What do you think?

pip/pip-406.md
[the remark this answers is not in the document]
RobertIndie: Yes. LGTM. I just missed the word "dispatch".

pip/pip-406.md
lhotari: > hi, @lhotar I agree with you point, but, How can we determine this? Initially, the design was as you mentioned (the subscription metrics represented by all three(broker/topic/subscription rate limit) dimensions). After discussing with @poorbarcode , we realized that this design makes it difficult for users to identify which dimension's configuration is affecting them.

There seems to be some misunderstanding. That's exactly my point too that it would be possible to determine what is causing the issue and when an individual subscription has been throttled. You seem to be suggesting that users don't need that information. Assuming that all subscriptions get equally throttled is something that might not hold. Keeping the metrics also on subscription level is a way to address that so that we know exactly and don't have to guess why and when a subscription was throttled.

What I'm suggesting is that we would simply keep the throttling even information. It's possible to slice-and-dice it for dashboards and show it it understandable format. That's the metrics side. Similarly, the topic stats information at subscription would contain multiple counters instead of just one so that it's possible to know exact the subscription was throttled.

In metrics, it's possible to have a single metric and use labels to capture the topic and subscription and reason.
These labels are there for any subscription metrics in any case. To get topic level metrics, you'd just filter the ones where the throttling reason is "topic". I'd assume that this is the recommended approach for exposing metrics in this case.

> If we obtain the rate trottled metrics for the broker and topic through aggregation, it can't accurately represent whether the impact is due to broker or topic level configurations.

There's a misunderstanding. All throttling events happen at the subscription level. The reason is either broker, topic or subscription, but they all happen at the subscription level. That's why it's sufficient to keep 3 counters at subscription level and aggregate the topic and broker level counters up to the topic and further to the broker level when generating topic / broker stats.

Makes sense?

shibd: Hi, Lari.
Thanks for reviewing. It seems like adding labels is a good idea. We can add a label to distinguish whether throttling is caused by a subscription, broker, or topic.

By doing this, we actually only need two metrics:
- `pulsar_subscription_dispatch_throttled_msg_events`
  - Labels:
    - topic
    - subscription
    - `reason`: broker/topic/subscription
- `pulsar_subscription_dispatch_throttled_bytes_events`
  - Labels:
    - topic
    - subscription
    - `reason`: broker/topic/subscription

When a user wants to view all throttling occurrences for a subscription, they just need to use a query like the one below (without the reason label).
```java
        double subscriptionAllDispatchThrottledMsgCount = subscriptionDispatchThrottledMsgCountMetrics.stream()
                .filter(m -> m.tags.get("subscription").equals(subName) && m.tags.get("topic").equals(topic))
                .mapToDouble(m-> m.value).sum();
```

if the user want get a subscription throttle events by topic rate limit:
```java
        double subscriptionDispatchThrottledBytesCount = subscriptionDispatchThrottledBytesCountMetrics.stream()
                .filter(m -> m.tags.get("subscription").equals(subName)
                        && m.tags.get("topic").equals(topic) && m.tags.get("reason").equals("topic"))
                .mapToDouble(m-> m.value).sum();
```

This way, we don’t need to add extra throttle metrics for `topics` and `brokers`; using this one metric can handle all scenarios.

For example, If the user wants to obtain the metric for the total number of times the broker rate limit was subscription throttled.
```java
        double brokerAllDispatchThrottledMsgCount = subscriptionDispatchThrottledMsgCountMetrics.stream()
                .filter(m -> m.tags.get("reason").equals("broker"))
                .mapToDouble(m-> m.value).sum();
```


What do you think? I’ve drafted a PR, and you can take a look at the implementation(You can review user scenarios through unit tests)

https://github.com/shibd/pulsar/pull/56

If there are no issues, I'll start modifying the PIP.

lhotari: > By doing this, we actually only need two metrics:
> 
> * `pulsar_subscription_dispatch_throttled_msg_events`
>   
>   * Labels:
>     
>     * topic
>     * subscription
>     * `reason`: broker/topic/subscription
> * `pulsar_subscription_dispatch_throttled_bytes_events`
>   
>   * Labels:
>     
>     * topic
>     * subscription
>     * `reason`: broker/topic/subscription

@shibd Yes, this would be the way in this approach from the metrics perspective when they get exposed for scraping at the /metrics endpoint.

> When a user wants to view all throttling occurrences for a subscription, they just need to use a query like the one below (without the reason label).

In this case, there is no "user" in Pulsar who would do these queries with code. For metrics exposed to a metrics database like Prometheus or VictoriaMetrics, filtering by labels is how the "user" would use the metrics.

Within Pulsar code base, there are these use cases to solve internally:
- exposing metrics to /metrics endpoint
- exposing stats to topic stats at subscription and topic level
- exposing stats to namespace stats
- exposing stats to broker stats

There's already examples of other metrics where the values get aggregated from subscription level to the topic level and namespace level. A similar solution would be preferred also for this new metric. I found that the `msgBacklog` is such metric currently, there are also other examples.

The difference here would be that at the subscription level, there would have to be 6 counters to hold the metrics.
They would be `dispatchThrottledMsgEventsSubscriptionLevel`, `dispatchThrottledMsgEventsTopicLevel`, `dispatchThrottledMsgEventsBrokerLevel`, `dispatchThrottledBytesEventsSubscriptionLevel`, `dispatchThrottledBytesEventsTopicLevel` and `dispatchThrottledBytesEventsBrokerLevel`. When aggregating (rolling up) the metric values, these would be needed. These would also be needed for the /metrics endpoint.

lhotari: @shibd I now noticed that https://github.com/shibd/pulsar/pull/56 already has the 6 different counters approach so that looks good to me. The only consideration could be to have a "container" class for exposing the stats externally instead of having individual fields, but since we don't do grouping currently, it's fine to handle it like you are already doing it in your PR.

lhotari (approved):
LGTM, thank you @shibd

RobertIndie (approved):

shibd:
https://lists.apache.org/thread/pmgkf67ph98k5l3lr60rtg96octb397q

The vote passed with 3 +1 binding.
- Lari
- Zike
- Yubiao

PR #23951, opened 2025-02-08
[fix] [ml] incorrect non-durable cursor's backlog due to concurrently trimming ledger and non-durable cursor creation

poorbarcode:
### Motivation
**Background**: the logic of calculating `backlog`
- There is a variable named `manegedledger.entriesAddedCounter`, which means how many entries were added after the topic was loaded up at the latest time.
- there is a variable named `cursor.messagesConsumedCounter`, which means how many entries were acked after the cursor was loaded up at the latest time, it can be a negative value.
- Backlog calculating mechanism: `manegedledger.entriesAddedCounter - cursor.messagesConsumedCounter`

**Steps to reproduce the issue**
- `ledgers: [ 1, 2, 3 ]`, `100` entries per ledger.
- `durable-cursor: {mark-deleted-position: 3:0}`
- `manegedledger.entriesAddedCounter`: `0`, no entries were added after the topic was loaded up.
- concurrently trimming ledger and non-durable cursor creation

| time | `trim ledgers` |  non-durable cursor |
| --- | --- | --- |
| 1 | | Init the variable `messagesConsumedCounter` by `managedLegder.ledgers`: `-300` |
| 2 | Calculate ledgers to delete |
| 3 | Move non-durable cursors' mark deleted position |
| 4 | Skip the non-durable cursor because it has not been added to the list `cursors` |
| 5 | Delete ledgers | 
| 6 | | added to the managed ledger via `addCursors` |

After that, if the non-durable cursor receives messages `3:0~3:99` and acknowledges them, the `messagesConsumedCounter` will be -200 (-300 + 100). Then backlog will be 200, but there is no message to consume|

**The issue we encountered**

Screenshot 2025-02-08 at 17 35 40

Screenshot 2025-02-08 at 17 35 18

### Modifications

- fix the race condition: since trimming ledger is executed in a lock block `ml.synchronized`, add a lock when creating a non-durable cursor
- Since it is hard to add a test to reproduce the issue, I did not add a test.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -1144,16 +1144,14 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
             return cachedCursor;
         }
 
-        NonDurableCursorImpl cursor = new NonDurableCursorImpl(bookKeeper, this, cursorName,
-                startCursorPosition, initialPosition, isReadCompacted);
-        cursor.setActive();
-
-        log.info("[{}] Opened new cursor: {}", name, cursor);
         synchronized (this) {
+            NonDurableCursorImpl cursor = new NonDurableCursorImpl(bookKeeper, this, cursorName,
+                    startCursorPosition, initialPosition, isReadCompacted);
+            cursor.setActive();
+            log.info("[{}] Opened new cursor: {}", name, cursor);
             addCursor(cursor);
+            return cursor;
         }
-
-        return cursor;
     }
 
     @Override
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -1144,8 +1144,9 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
             return cachedCursor;
         }
 
-        // To avoid cursor initialization and "ML.trimLedgers" execute concurrency, we add this lock block.
-        // See details https://github.com/apache/pulsar/pull/23951.
+        // The backlog of a non-durable cursor could be incorrect if the cursor is created before `internalTrimLedgers`
+        // and added to the managed ledger after `internalTrimLedgers`.
+        // For more details, see https://github.com/apache/pulsar/pull/23951.
         synchronized (this) {
             NonDurableCursorImpl cursor = new NonDurableCursorImpl(bookKeeper, this, cursorName,
                     startCursorPosition, initialPosition, isReadCompacted);
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -1144,6 +1144,8 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
             return cachedCursor;
         }
 
+        // To avoid cursor initialization and "ML.trimLedgers" execute concurrency, we add this lock block.
+        // See details https://github.com/apache/pulsar/pull/23951.
         synchronized (this) {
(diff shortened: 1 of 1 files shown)


shibd (approved):

gaoran10 (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1144,16 +1144,14 @@ public ManagedCursor newNonDurableCursor(Position startCursorPosition, String cu
...
-        cursor.setActive();
-
-        log.info("[{}] Opened new cursor: {}", name, cursor);
         synchronized (this) {
+            NonDurableCursorImpl cursor = new NonDurableCursorImpl(bookKeeper, this, cursorName,
+                    startCursorPosition, initialPosition, isReadCompacted);
+            cursor.setActive();
+            log.info("[{}] Opened new cursor: {}", name, cursor);
             addCursor(cursor);
+            return cursor;
```
BewareMyPower, line 1153: Since there is no test to protect this change (it might be hard to add the test), to avoid possible regression in futurem you'd better add some comments here to explain why the creation of the `NonDurableCursorImpl` should be put inside the synchronized block.

BewareMyPower: From the PR description, it seems that we should only synchronize the `initializeCursorPosition` method with `internalTrimLedgers`, would it be better to just minimum the synchronize block like:

```java
    void initializeCursorPosition(Pair<Position, Long> lastPositionCounter) {
        synchronized (ledger) {
            /* ... */
        }
    }
```

BewareMyPower:
> Receive messages 3:0~3:99 and ack them, cursor.messagesConsumedCounter += 100 (it will be -200)

Could you explain which method does it happen? From the table in the PR description, I can only see:
- 2, 3, 4, 5 happen in `ManagedLedgerImpl#internalTrimLedgers`
- 1 happens in `ManagedLedgerImpl#initializeCursorPosition`

but I'm not sure where does 6 happen.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Added

poorbarcode:
@BewareMyPower 

> but I'm not sure where does 6 happen.

It is in the method `ManagedCursorImpl.setAcknowledgedPosition`,  https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L2001

dao-jun (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1144,16 +1144,16 @@ public ManagedCursor newNonDurableCursor(Position startCursorPosition, String cu
             return cachedCursor;
         }
 
-        NonDurableCursorImpl cursor = new NonDurableCursorImpl(bookKeeper, this, cursorName,
-                startCursorPosition, initialPosition, isReadCompacted);
-        cursor.setActive();
-
-        log.info("[{}] Opened new cursor: {}", name, cursor);
+        // To avoid cursor initialization and "ML.trimLedgers" execute concurrency, we add this lock block.
+        // See details https://github.com/apache/pulsar/pull/23951.
```
BewareMyPower, line 1148:
```suggestion
        // Add this lock block to avoid cursor initialization and `internalTrimLedgers` execution concurrency.
        // See details in https://github.com/apache/pulsar/pull/23951.
```

`ML.trimLedgers` is hard to understand, use the actual method name `internalTrimLedgers` instead.

Besides, correct the grammar errors (via the Copilot plugin in VSCode)

BewareMyPower: I've got the fix now. Here are the improved comments, please help double confirm it @poorbarcode 

The key point is that we want to synchronize both the creation of a non-durable cursor and `addCursor()` call with the trim operation, not only the creation.

PR #23955, opened 2025-02-10
[fix][ml] Fix memory leak due to duplicated RangeCache value retain operations 

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/23903 introduces a memory leak issue in `RangeCache#removeEntry`.

```diff
-        Value value = entryWrapper.getValue(key);
+        Value value = getValueMatchingEntry(entry);
```

Unlike `entryWrapper.getValue`, `getValueMatchingEntry` will increase the reference count of `entry`'s value.

### Modifications

- Remove the duplicated retain operation in `RangeCache#removeEntry`. Add some API notes for the private methods that might increase the reference count of the value
- Apply the reference count validation for eviction on a `RangeCache` (`RangeCacheTest.customTimeExtraction`).

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/RangeCache.java
@@ -284,6 +284,9 @@ public class RangeCache<Key extends Comparable<Key>, Value extends ValueWithKeyV
         }
     }
 
+    /**
+     * @apiNote the returned value must be released if it's not null
+     */
     private Value getValueMatchingEntry(Map.Entry<Key, EntryWrapper<Key, Value>> entry) {
         Value valueMatchingEntry = EntryWrapper.getValueMatchingMapEntry(entry);
         return getRetainedValueMatchingKey(entry.getKey(), valueMatchingEntry);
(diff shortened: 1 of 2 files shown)


BewareMyPower:
Though I think catching the `IllegalReferenceCountException` from `value.retain()` is not a good practice and error prone, let's just keep the fix simple.

lhotari:
> Though I think catching the `IllegalReferenceCountException` from `value.retain()` is not a good practice and error prone, let's just keep the fix simple.

`value.retain()` is essential since RangeCache is not synchronized. Removing the logic will cause problems in concurrent removals since that could currently happen.

lhotari (approved):
Oh sorry, reviewed again and noticed that it's fine.

lhotari:
I have created https://github.com/apache/pulsar/pull/23956 (WIP, draft) to address future memory leak regressions. It seems that there are quite a few tests where leaks happen currently and some of them are hard to fix. I have added a solution where it's possible to ignore specific test classes.

Issue #23723, opened 2024-12-12
[feat] Pulsar Functions should provide a way to set Pulsar client memory limit


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

I haven't found a way to tune the memory limit for the Pulsar client used in Pulsar Functions. 

The memory limit is disabled in Pulsar Functions: https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/InstanceUtils.java#L161

There's code for configuring it:
https://github.com/apache/pulsar/blob/05e57dd3a443c5b99c21054c56a1b497455fa867/pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/runtime/thread/ThreadRuntimeFactory.java#L106-L168

but it doesn't get called in the JavaInstanceStarter case. The JavaInstanceStarter is used when running Java functions in process / k8s / function mesh runtimes (other than thread runtime):
https://github.com/apache/pulsar/blob/567174f43528c0f7ae917bfb5166213973c62c29/pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/runtime/JavaInstanceStarter.java#L236-L245
`Optional.empty()` is passed to memory limit.
https://github.com/apache/pulsar/blob/05e57dd3a443c5b99c21054c56a1b497455fa867/pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/runtime/thread/ThreadRuntimeFactory.java#L92-L104

It looks like the memory limit solution has been only implemented for the ThreadRuntimeFactory in #9320. When using ThreadRuntimeFactory, it's possible to set the limit in `conf/functions_worker.yml`. This isn't helpful for other runtime types (process, k8s).

### Solution

Add the Pulsar client memory limit configuration to `org.apache.pulsar.common.functions.FunctionConfig`, `org.apache.pulsar.functions.instance.InstanceConfig` and `org.apache.pulsar.functions.worker.WorkerConfig` and make it configurable in the same way as other instance config parameters such as `maxPendingAsyncRequests`. 
The default would be configured in `WorkerConfig` for all functions and the function configuration would be in `InstanceConfig` and configurable with `FunctionConfig` in yaml and command line. 
[...]

PR #23959, opened 2025-02-11
[improve][fn] Provide a way to set Pulsar client memory limit in Pulsar Function

walkinggo:
Main Issue: #23723

### Motivation

see #23723 . The main work of this PR is to add a method for setting the memory limit of the Pulsar client in a Pulsar Function.

### Modifications

1.Memory Limit Support:
Added MemoryLimit class to define memory constraints for Pulsar clients, which includes both an absolute value and a percentage of the maximum direct memory.
This memory configuration is applied across several components including FunctionConfig, WorkerConfig, InstanceConfig, and ThreadRuntimeFactory.
2.Class Updates:
Introduced new fields in configuration classes such as FunctionConfig, WorkerConfig, and InstanceConfig to hold memory limit settings.
Updated ThreadRuntimeFactory to handle the memory limit during initialization, ensuring that the client memory limit is respected.
3.Proto File Changes:
Updated Function.proto to include a new MemoryLimit field, allowing the configuration of memory limits directly via function details.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/walkinggo/pulsar/pull/8


pulsar-client-admin-api/pom.xml
```diff
@@ -43,6 +43,21 @@
...
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-resolver-dns</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
```
BewareMyPower, line 60: Please avoid introducing Netty dependencies to the API module

BewareMyPower (changes requested):

pulsar-client-admin-api/pom.xml
walkinggo: The main reason I introduced it here is that I moved `MemoryLimit` and its required `DirectMemoryUtils` to this location. However, `DirectMemoryUtils` requires importing `io.netty.util.internal.PlatformDependent`.

walkinggo: I’m not sure where I should place this utility class that is meant to be used by multiple components.

> Please avoid introducing Netty dependencies to the API module

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.common.functions;
+
+import static org.apache.pulsar.common.policies.data.impl.BookieInfoImpl.BookieInfoImplBuilder.checkArgument;
```
lhotari, line 21: this import doesn't make sense. Use `com.google.common.base.Preconditions.checkArgument` instead.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
```diff
@@ -0,0 +1,66 @@
...
+            if (absolute != null) {
+                return Optional.of(absolute);
+            }
+
+            if (percentOfDirectMem != null) {
+                return Optional.of(getBytesPercentDirectMem(percentOfDirectMem));
+            }
+        }
+        return Optional.empty();
+    }
```
lhotari, line 61: What's the point of this static method? Why not have an instance method directly in the class to do the calculation?

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private static long getBytesPercentDirectMem(double percent) {
+        return (long) (DirectMemoryUtils.jvmMaxDirectMemory() * (percent / 100));
+    }
```
lhotari, line 65: Couldn't this method be moved to `DirectMemoryUtils` with a reasonable name , for example `jvmMaxDirectMemoryPercentage`?

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/DirectMemoryUtils.java
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.pulsar.common.util;
+package org.apache.pulsar.common.functions;
```
lhotari, line 19: Why are you moving this class. Couldn't we keep it in the existing location?

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
lhotari: If this was an instance method, you could pass the direct memory size as a parameter to the instance method if you'd like to capsulate the logic here. However it seems unnecessary in the first place.

pulsar-functions/proto/src/main/proto/Function.proto
```diff
@@ -98,8 +98,16 @@ message FunctionDetails {
     bool retainOrdering = 21;
     bool retainKeyOrdering = 22;
     SubscriptionPosition subscriptionPosition = 23;
+    MemoryLimit memoryLimit = 24;
 }
 
+message MemoryLimit {
+    // absolute value in bytes, represents the memory limit in bytes
+    int64 absoluteValue = 1;
```
lhotari, line 106: why is this called `absoluteValue` if it's about the memory limit in bytes? a better name is `limitInBytes` or picking a name that is also used in existing code.

pulsar-client-admin-api/pom.xml
BewareMyPower: >  its required DirectMemoryUtils

I don't think `MemoryLimit` is coupled with `DirectMemoryUtils`.

lhotari: It seems that the MemoryLimit class shouldn't even depend on DirectMemoryUtils so moving it should be reverted. Related comment https://github.com/apache/pulsar/pull/23959#discussion_r1951215801

lhotari: Although I suggested making it an instance method as the first thought, it seems that there's no need to have the calculation together with the MemoryLimit class. The calculation could be handled when the value is needed. The MemoryLimit class could be kept as a plain data transfer object (DTO).

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
lhotari: update: https://github.com/apache/pulsar/pull/23959#discussion_r1952299178

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
lhotari: update: https://github.com/apache/pulsar/pull/23959#discussion_r1952299178

pulsar-client-admin-api/pom.xml
walkinggo: Thank you for your help. They indeed aren't tightly coupled, so I have moved `calculateClientMemoryLimit` back to its original location and also restored `DirectMemoryUtils` to its original position. This way, there's no need to add the Netty dependency to the `pom.xml` file.

pulsar-functions/proto/src/main/proto/Function.proto
walkinggo: > why is this called `absoluteValue` if it's about the memory limit in bytes? a better name is `limitInBytes` or picking a name that is also used in existing code.

I think you're right. I have renamed the variable accordingly.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/FunctionConfig.java
```diff
@@ -133,7 +133,8 @@ public enum Runtime {
     private Boolean exposePulsarAdminClientEnabled;
     // Whether the consumer should skip to latest position in case of failure recovery
     private Boolean skipToLatest;
-
+    // The momoryLimit of pulsar client
```
lhotari, line 136: typo

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/MemoryLimit.java
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.pulsar.common.functions;
+
+import lombok.Data;
+import lombok.experimental.Accessors;
+
+@Data
+@Accessors(chain = true)
+/**
+ * Memory limit set for the pulsar client used by all instances
+ * If `absoluteValue` and `percentOfMaxDirectMemory` are both set, then the min of the two will be used.
```
lhotari, line 28: referencing old field. Javadoc doesn't yet support markdown, so using backticks in Javadoc isn't useful.

pulsar-functions/proto/src/main/proto/Function.proto
lhotari, line 106: rename field

pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/worker/WorkerConfig.java
```diff
@@ -377,6 +378,11 @@ public class WorkerConfig implements Serializable, PulsarConfiguration {
         doc = "The directory to download functions by runtime manager"
     )
     private String downloadDirectory;
+    @FieldContext(
+            category = CATEGORY_FUNC_RUNTIME_MNG,
+            doc = "Memory limit set for the pulsar client used by all instances"
```
lhotari, line 383:
```suggestion
            doc = "Default memory limit set for the pulsar client used instances"
```

walkinggo:
> @walkinggo Please first make sure that the CI build passes in your own fork. I see that [walkinggo#8](https://github.com/walkinggo/pulsar/pull/8) referenced in the PR description isn't passing.

The error in my PR is related to json-smart, but my code doesn't involve modifications to this component. I suspect there might be an issue with Maven Central, and I also noticed that this [PR](https://github.com/apache/pulsar/pull/23966#pullrequestreview-2611658179) is updating the version of json-smart.

walkinggo:
> > > @walkinggo Please first make sure that the CI build passes in your own fork. I see that [walkinggo#8](https://github.com/walkinggo/pulsar/pull/8) referenced in the PR description isn't passing.
> > 
> > 
> > The error in my PR is related to json-smart, but my code doesn't involve modifications to this component. I suspect there might be an issue with Maven Central, and I also noticed that this [PR](https://github.com/apache/pulsar/pull/23966#pullrequestreview-2611658179) is updating the version of json-smart.
> 
> @walkinggo PR #23966 has been merged. Please merge origin/master to your PR to get the changes included.

The PR has been successfully merged and passed the CI checks.

Issue #14529, opened 2022-03-02
PIP-146: ManagedCursorInfo compression


nodece:
Discussion thread: https://lists.apache.org/thread/j92bzsby9n2ozc9gcw5psgcy2026l1wm

## Motivation

The cursor data is managed by ZooKeeper/etcd metadata store. When cursor data becomes more and more, the data size will increase and will take a lot of time to pull the data. Therefore, it is necessary to add compression for the cursor, which can reduce the size of data and reduce the time of pulling data.

## Goal

Support use the LZ4/ZLIB/ZSTD/SNAPPY to compress the ManagedCursorInfo.

### CursorInfo compression format

[MAGIC_NUMBER] + [METADATA_SIZE] + [METADATA_PAYLOAD] + [MANAGED_CURSOR_INFO_PAYLOAD] 
 
- MAGIC_NUMBER
Use 0x4778, it is the same as the magic number of ledger info.

- METADATA
Add a named `ManagedCursorInfoMetadata` message to `MLDataFormats.proto`

```
message ManagedCursorInfoMetadata {
    required CompressionType compressionType = 1;
    required int32 uncompressedSize = 2;
}
```

### CursorInfo compression and decompression design

Currently, these compressions types have been defined and implemented by Pulsar, we only need to deal with compression and decompression of the `ManagedCursorInfo` data:

- Get CursorInfo from the metadata store
We will check the cursor data header, if it is compressed, we will parse the bytes data by compressed format, otherwise we will parse the cursor data directly by the original way.

 - Add/Update CursorInfo to the metadata store
The default is to use compression if the compression type is specified, otherwise we will put this data to the metadata store directly.

### CursorInfo compression type configuration

Add `managedCursorInfoCompressionType` in `org.apache.pulsar.broker.ServiceConfiguration` and `org.apache.bookkeeper.mledger.ManagedLedgerFactoryConfig`.

## Compatibility

1. The compression is disabled by default
2. Data can be upgraded safely or downgraded. When enabled, we can migrate the old data to new data with compression metadata in updating action. When disabled, we can revert this data to the previous version in update action and can parse the compressed data by the compression metadata in getting action

nodece:
This proposal has 3 (+1) bindings and 0 (-1) and has stayed open for at least 48 hours:
https://lists.apache.org/thread/kvpostlbqty87cdjgt7vrfozn3qz2j1t

PR #11490, opened 2021-07-28
[ManagedLedger] Compress managed ledger info

gaoran10:
### Motivation

Currently, the `ManagedLedgerInfo` contains offload context info, if there are too many ledgers in one ManagedLedger, a Zookeeper ZNode data will increase, it's hard for Zookeeper to manage these data. We could compress the `ManagedLedgerInfo` data, this will decrease the data size.

For example, if one `ManagedLedgerInfo` contains 30000 ledgers and each `LedgerInfo` contains offload context, the uncompressed data size is about 6MB, after compress with ZSTD, the compression data size could be decreased to about 1.3MB.

### Modifications

Add a `ManagedLedgerInfoMetadata` before `ManagedLedgerInfo`, the data structure as below.

```
[MAGIC_NUMBER] (2) + [METADATA_SIZE] (4) + [METADATA_PAYLOAD] + [MANAGED_LEDGER_INFO_PAYLOAD]
```

Add a new configuration `managedLedgerInfoCompressionType` to control ManagedLedgerInfo compression type, if not set this configuration, then don't compress ManagedLedgerInfo data.

### Migration

When reading data from the Zookeeper, check the magic number of the data, if the head data match the magic number then try to parse metadata and uncompress data, if encounter errors or not match fall back to parse `ManagedLedgerInfo` directly.

### Verifying this change

Verify compress and uncompress `ManagedLedgerInfo` data.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API: (no)
  - The schema: (no)
  - The default values of configurations: (no)
  - The wire protocol: (no)
  - The rest endpoints: (no)
  - The admin cli options: (no)
  - Anything that affects deployment: (no)

#### For committer

For this PR, do we need to update docs?

- If yes,
  
  - if you update docs in this PR, label this PR with the `doc` label.
  
  - if you plan to update docs later, label this PR with the `doc-required` label.

  - if you need help on updating docs, create a follow-up issue with the `doc-required` label.
  
- If no, label this PR with the `no-need-doc` label and explain why.

PR #23960, opened 2025-02-11
[fix][ml] Fix memory leaks in ManagedCursorInfo and ManagedLedgerInfo decompression and compression

lhotari:
### Motivation

Running `ManagedCursorInfoMetadataTest` and `ManagedLedgerInfoMetadataTest` with Netty leak detector's `paranoid` level reveal Netty `ByteBuf` leaks in `ManagedCursorInfo` and `ManagedLedgerInfo` decompression and compression.

ManagedCursorInfo compression was introduced in [PIP-146](https://github.com/apache/pulsar/issues/14529). ManagedLedgerInfo compression was introduced earlier, in PR #11490.

### Modifications

Fix the leak by properly releasing buffers. 
* In the `parseManagedCursorInfo` and `parseManagedLedgerInfo` methods there was a clear leak where the ByteBuf instance wasn't released
* In the `compressManagedInfo` method, the `CompositeByteBuf` instance was never released. `CompositeByteBuf` might copy input buffers into a new buffer and releasing the components separately is not a correct approach. Handling exceptional cases such as decompression failing cannot be handled in the compressManagedInfo method. The compression codec implementation should handle releasing any buffers it has allocated in that case. That's why the original handling didn't even make sense.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/MetaStoreImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/MetaStoreImpl.java
@@ -478,8 +478,13 @@ public class MetaStoreImpl implements MetaStore, Consumer<Notification> {
             try {
                 MLDataFormats.ManagedCursorInfoMetadata metadata =
                         MLDataFormats.ManagedCursorInfoMetadata.parseFrom(metadataBytes);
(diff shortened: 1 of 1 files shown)


dao-jun (approved):
Great catch! LGTM

PR #23974, opened 2025-02-12
[improve][broker] Support showing client ip address in client stats while using reverse proxy

rdhabalia:
### Motivation

In Apache Pulsar, the broker enables producers and consumers to connect to a topic and provides an API to retrieve topic statistics. These stats include a list of connected producers and consumers, along with their IP addresses and connection times. This information is particularly valuable when dealing with a large number of producers and consumers from various client hosts, as it helps troubleshoot issues such as:

Identifying which client host has an active consumer
Detecting if a client host has stopped consuming messages
Diagnosing message backlogs
Thus, mapping the client host IP to the corresponding producer or consumer is crucial.

**The Issue with Reverse Proxies**
However, this mapping breaks when a reverse proxy is used between the client and broker. In such cases, the broker records only the proxy's IP address for all connected producers and consumers, making it difficult to identify the actual client host. Apache Pulsar supports multiple proxy solutions, such as Pulsar-Proxy and SNI Proxy, which further complicates troubleshooting by obscuring client IPs.

To resolve this, this PR ensures that when a client library connects to a broker via a proxy, it sends the actual client IP address. The broker then correctly identifies and records this IP in the stats API, mapping it to the appropriate producer or consumer. This approach abstracts the proxy layer from users, allowing them to see accurate client IPs without any additional effort.

This PR doesn't change client-broker protocol, API definition or configuration.

### Modifications

Client lib sends an ip-address property when client lib detects a proxy, and the broker shows it in the client stats.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
@@ -48,6 +48,7 @@ import org.apache.bookkeeper.mledger.Entry;
 import org.apache.bookkeeper.mledger.Position;
 import org.apache.bookkeeper.mledger.PositionFactory;
 import org.apache.bookkeeper.mledger.impl.AckSetStateUtil;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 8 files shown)


rdhabalia:
ping,. if anyone can review it.

lhotari:
This PR introduces a security issue since any client could impersonate another client IP by setting the `Metadata.CLIENT_IP` metadata. We need to address this before releasing this feature. For new features like this one, there should be a PIP.

lhotari:
Pulsar already supports the standard way to pass the remote client's IP address in a secure way. That is by using that HA Proxy Protocol. All major proxy implementations support this. In the case of SNI proxy, there should be a way to configure this too.

PR #21865, opened 2024-01-08
[fix][broker] Fix compaction/replication data loss when expire messages

coderzc:
### Motivation

When expiring messages during compaction, we may lose the middle message due to `__compaction` cursor mark-delete position being advanced. So I think we should skip TTL check for `__compaction`.

replicators also have a similar problem, so also skip the expiration check of replicator

### Modifications

* Skip TTL check for `__compaction` and skip expire `__compaction` subscriptions.
* Passing lost `cursor.properties` when the mark deletes for the cursor.
* Skip TTL check for `replicator`, and remove the expired message replicator.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #23975, opened 2025-02-13
[fix][broker] Add expire check for replicator

nodece:
### Motivation

https://github.com/apache/pulsar/pull/21865 removes the expire check for the replicator and compaction subscription. However, this introduces a breaking change in that the replication message never expires.

The TTL policy should be applied to the subscription and replicator.

### Modifications

- Add expire check for the replicator.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari (approved):
LGTM, just wondering whether we need more comments in the `testReplicatorWithTTL` test case so that it's clear what the expected behavior is. It's currently hard to validate the test case itself.

lhotari (approved):
LGTM, thanks for improving the test case too

PR #23976, opened 2025-02-13
[fix] Potential fix for code scanning alert no. 22: HTTP response splitting

merlimat:
Potential fix for [https://github.com/apache/pulsar/security/code-scanning/22](https://github.com/apache/pulsar/security/code-scanning/22)

To fix the problem, we need to ensure that any user-provided input used in HTTP headers is properly sanitized to prevent HTTP response splitting. This can be done by removing or escaping special characters such as CRLF from the input before using it in the response header.

The best way to fix this issue is to create a method that sanitizes the input by removing or escaping special characters and then use this method to sanitize the `SASL_STATE_SERVER` header value before setting it in the response.

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari (approved):
LGTM

PR #21552, opened 2023-11-10
[fix][broker] Fix topic compaction is failed after compactedLedger's all quorum is being recover

TakaHiR07:
Fixes #21551

### Motivation

Fix topic compaction is failed after compactedLedger's all quorum is being recover, which is described in issue. 

### Modifications

When CompactedTopicImpl try to open compactedLedger, use asyncOpenLedgerNoRecovery instead of asyncOpenLedger. Then compactedLedger can watch the zk node change and update the ledger metadata in broker.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/TakaHiR07/pulsar/pull/19

PR #23977, opened 2025-02-13
[fix][broker] Invalid regex in PulsarLedgerManager causes zk data notification to be ignored

TakaHiR07:
### Motivation

ledger zk path is like "/ledgers/00/0601/L7170". But currently it exist pattern regex error cause zk data notification can not execute. 

ledgerPathRegex.matcher(n.getPath()).matches() is always false.

企业微信截图_c58d5c24-43ee-4bb1-b1b5-9d9fbb3747e2

### Modifications

use correct pattern. 

Alternative modification:  remove the judgement in handleDataNotification(), since getLedgerId(n.getPath()) would throw error if the path is not ledger path.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] The public API

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


thetumbled:
Some test is requried to ensure the notification logic is triggered and works correct.

lhotari:
> ledger zk path is like "/ledgers/00/0601/L7170". But currently it exist pattern regex error cause zk data notification can not execute.

Great catch @TakaHiR07. What is the current impact of this in Pulsar & Bookkeeper (which is using PulsarLedgerManager in the Pulsar distribution of Bookkeeper)?

TakaHiR07:
> Great catch @TakaHiR07. What is the current impact of this in Pulsar & Bookkeeper (which is using PulsarLedgerManager in the Pulsar distribution of Bookkeeper)?

One impact is all the asyncOpenLedgerNoRecovery in pulsar can not register successful MetadataListener. The code is here:  https://github.com/apache/bookkeeper/blob/606db747eae9856fed0aeb3f16ef01e7c9254e26/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java#L95-L105

I am not sure whether other place use PulsarLedgerManager and register zk listener.

lhotari:
> Some test is requried to ensure the notification logic is triggered and works correct.

@thetumbled That's right that there should be tests, but this just shows that the original code didn't have proper test coverage if it's currently broken. 

One possible resolution would be to add an issue report about the missing test coverage and add the tests later. That moment usually never comes, but it's also bad to have this issue around.

TakaHiR07:
> > > Great catch @TakaHiR07. What is the current impact of this in Pulsar & Bookkeeper (which is using PulsarLedgerManager in the Pulsar distribution of Bookkeeper)?
> > 
> > 
> > One impact is all the asyncOpenLedgerNoRecovery in pulsar can not register successful MetadataListener. The code is here: https://github.com/apache/bookkeeper/blob/606db747eae9856fed0aeb3f16ef01e7c9254e26/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java#L95-L105
> > I am not sure whether other place use PulsarLedgerManager and register zk listener.
> 
> I wonder what parts of the metadata could change. My guess is LAC (lastAddConfirmed) and length based on this: https://github.com/apache/bookkeeper/blob/54bdc0d60b32830b513089167cee67f52f4735eb/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L367-L370 . I would assume that this would be relevant when the ledger is in recovery state. States: https://github.com/apache/bookkeeper/blob/2192caaf9738cf4efb799647cc5a5f68bf1823b2/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java#L154-L167

@lhotari I think if ledger is in recover state, LAC would be changed. But we should not use asyncOpenLedgerNoRecovery to register zk metadata listener, instead should use asyncOpenLedger to update ledgerHandler's metadata. This is no problem. 

But if ledger is already closed, and then trigger bookkeeper auto-recovery because of disk error, ledger's quorum would be changed. 

Actually, the issue is found when I fix another issue, you can see here. https://github.com/apache/pulsar/pull/21552

lhotari (approved):
LGTM. The pattern looks correct although there aren't tests. The problem isn't caused by this PR since there doesn't seem to be existing tests for PulsarLedgerManager.

PR #23979, opened 2025-02-13
[fix][io] Fix pulsar-io:pom not found

nodece:
### Motivation

#23932 should not set the skip in the `pulsar-io/pom.xml`, which breaks the `pulsar-io-core` dependency:
```
Error:  Failed to read artifact descriptor for com.***.pulsar:pulsar-io-core:jar:3.0.10.0-32dfc7d-SNAPSHOT
Error:  	Caused by: The following artifacts could not be resolved: com.***.pulsar:pulsar-io:pom:3.0.10.0-32dfc7d-SNAPSHOT (absent): com.***.pulsar:pulsar-io:pom:3.0.10.0-32dfc7d-SNAPSHOT was not found in
```

### Modifications

- Add maven-deploy-plugin with skip=true for each connector.
- Remove maven-deploy-plugin with skip=true from the `pulsar-io/pom.xml`.
- Remove maven-deploy-plugin with skip=true from the `pulsar-io/core/pom.xml` and `pulsar-io/common/pom.xml`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-io/aerospike/pom.xml
+++ b/pulsar-io/aerospike/pom.xml
@@ -69,6 +69,13 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>${skipDeployConnector}</skip>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.apache.nifi</groupId>
         <artifactId>nifi-nar-maven-plugin</artifactId>
       </plugin>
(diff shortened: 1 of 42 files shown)


lhotari (approved):
LGTM

merlimat (commented):
Can we also set to skip by default?

lhotari:
> Can we also set to skip by default?

@merlimat I think that would be useful starting from Pulsar 4.1 . Do we need a PIP for making that change?

PR #23980, opened 2025-02-13
[fix][broker]Consumer stuck when delete subscription __compaction failed

poorbarcode:
### Motivation

**Background**
- The steps of unsubscribing `__compaction`
  - Delete compacted ledger
  - Delete cursor
  - Remove subscription

**Issue 1**: consumer will stuck if deleting cursor(the step 2 above) failed 
- Delete compacted ledger
- Delete cursor failed
- Reload the topic
- Compactor subscription relates to a deleted ledger
- Consumers are stuck because the compacted ledger has been deleted, and the messages were lost. The broker will keep printing the error log below

```
2025-02-13T17:23:51,448 - ERROR - [broker-topic-workers-OrderedExecutor-4-0:PersistentDispatcherSingleActiveConsumer] - [persistent://public/default/tp-74ea581c-d7ca-4558-999b-804b6e33faee / reader-801f712e61-Consumer{subscription=PersistentSubscription{topic=persistent://public/default/tp-74ea581c-d7ca-4558-999b-804b6e33faee, name=reader-801f712e61}, consumerId=1, consumerName=Ti6va, address=[id: 0xce1b338f, L:/127.0.0.1:52193 - R:/127.0.0.1:52199] [SR:127.0.0.1, state:Connected]}] Error reading entries at 11:0 : org.apache.bookkeeper.client.BKException$BKNoSuchLedgerExistsException: No such ledger exists on Bookies - Retrying to read in 15.0 seconds
```

You can reproduce the issue by the test `testReadMsgsAfterDisableCompaction(true)`

---

**issue 2**
The compaction task can concurrently execute by deleting the cursor `__compaction`

### Modifications

- Fix issue 1
- Fix issue 2
- Print an error log if a compactor subscription is initialized twice, which may cause a ledger lost/

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -245,6 +245,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
     private static final Long COMPACTION_NEVER_RUN = -0xfebecffeL;
     private volatile CompletableFuture<Long> currentCompaction = CompletableFuture.completedFuture(
             COMPACTION_NEVER_RUN);
+    private volatile AtomicBoolean disablingCompaction = new AtomicBoolean(false);
     private TopicCompactionService topicCompactionService;
 
     // TODO: Create compaction strategy from topic policy when exposing strategic compaction to users.
@@ -1340,18 +1341,42 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             return;
         }
 
-        currentCompaction.handle((__, e) -> {
-            if (e != null) {
+        // Avoid concurrently execute compaction and unsubscribing.
+        synchronized (this) {
+            if (!disablingCompaction.compareAndSet(false, true)) {
+                unsubscribeFuture.completeExceptionally(
+                        new SubscriptionBusyException("the subscription is deleting by another task"));
+                return;
+            }
+        }
+        // Unsubscribe compaction cursor and delete compacted ledger.
+        currentCompaction.thenCompose(__ -> {
+            asyncDeleteCursor(subscriptionName, unsubscribeFuture);
+            return unsubscribeFuture;
+        }).thenAccept(__ -> {
+            try {
+                ((PulsarCompactorSubscription) subscription).cleanCompactedLedger();
+            } catch (Exception ex) {
+                Long compactedLedger = null;
+                Optional<CompactedTopicContext> compactedTopicContext = getCompactedTopicContext();
+                if (compactedTopicContext.isPresent() && compactedTopicContext.get().getLedger() != null) {
+                    compactedLedger = compactedTopicContext.get().getLedger().getId();
+                }
+                log.error("[{}][{}][{}] Error cleaning compacted ledger", topic, subscriptionName, compactedLedger, ex);
+            } finally {
+                // Reset the variable: disablingCompaction,
+                disablingCompaction.compareAndSet(true, false);
+            }
+        }).exceptionally(ex -> {
+            if (currentCompaction.isCompletedExceptionally()) {
                 log.warn("[{}][{}] Last compaction task failed", topic, subscriptionName);
-            }
-            return ((PulsarCompactorSubscription) subscription).cleanCompactedLedger();
-        }).whenComplete((__, ex) -> {
-            if (ex != null) {
-                log.error("[{}][{}] Error cleaning compacted ledger", topic, subscriptionName, ex);
-                unsubscribeFuture.completeExceptionally(ex);
             } else {
-                asyncDeleteCursor(subscriptionName, unsubscribeFuture);
+                log.warn("[{}][{}] Failed to delete cursor task failed", topic, subscriptionName);
             }
+            // Reset the variable: disablingCompaction,
+            disablingCompaction.compareAndSet(true, false);
+            unsubscribeFuture.completeExceptionally(ex);
+            return null;
         });
     }
 
@@ -3932,6 +3957,10 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             try {
                 if (isClosingOrDeleting) {
                     log.info("[{}] Topic is closing or deleting, skip triggering compaction", topic);
+                    return;
+                }
+                if (disablingCompaction.get()) {
+                    log.info("[{}] Compaction is disabling, skip triggering compaction", topic);
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -245,6 +245,7 @@ public static boolean isDedupCursorName(String name) {
     private static final Long COMPACTION_NEVER_RUN = -0xfebecffeL;
     private volatile CompletableFuture<Long> currentCompaction = CompletableFuture.completedFuture(
             COMPACTION_NEVER_RUN);
+    private volatile AtomicBoolean disablingCompaction = new AtomicBoolean(false);
```
lhotari, line 248:
```suggestion
    private final AtomicBoolean disablingCompaction = new AtomicBoolean(false);
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1340,18 +1341,42 @@ private void asyncDeleteCursorWithCleanCompactionLedger(PersistentSubscription s
             return;
         }
 
-        currentCompaction.handle((__, e) -> {
-            if (e != null) {
-                log.warn("[{}][{}] Last compaction task failed", topic, subscriptionName);
+        // Avoid concurrently execute compaction and unsubscribing.
+        synchronized (this) {
```
lhotari, line 1345: Is there really a need for a synchronized block here?

pulsar-broker/src/test/java/org/apache/pulsar/compaction/GetLastMessageIdCompactedTest.java
```diff
@@ -308,6 +319,145 @@ public void testGetLastMessageIdAfterCompactionWithCompression(boolean enabledBa
...
+        admin.topicPolicies().setCompactionThreshold(topicName, 1);
+        admin.topics().createSubscription(topicName, "s1", MessageId.earliest);
+        var producer = pulsarClient.newProducer(Schema.STRING).topic(topicName).enableBatching(false).create();
+        producer.newMessage().key("k0").value("v0").send();
+        triggerCompactionAndWait(topicName);
+        admin.topics().deleteSubscription(topicName, "s1");
+        PersistentTopic persistentTopic =
+                (PersistentTopic) pulsar.getBrokerService().getTopic(topicName, false).get().get();
+        AtomicBoolean disablingCompaction =
+                WhiteboxImpl.getInternalState(persistentTopic, "disablingCompaction");
```
lhotari, line 402: Please avoid using reflection/`WhiteboxImpl`. A better approach is to have a default access method which is annotation with `@VisibleForTesting`.

pulsar-broker/src/test/java/org/apache/pulsar/compaction/GetLastMessageIdCompactedTest.java
```diff
@@ -308,6 +319,145 @@ public void testGetLastMessageIdAfterCompactionWithCompression(boolean enabledBa
...
+        Awaitility.await().untilAsserted(() -> {
+            assertTrue(disablingCompaction.get());
+        });
+        producer.newMessage().key("k1").value("v1").send();
+        producer.newMessage().key("k2").value("v2").send();
+        CompletableFuture<Long> currentCompaction1 =
+                WhiteboxImpl.getInternalState(persistentTopic, "currentCompaction");
+        persistentTopic.triggerCompaction();
+        CompletableFuture<Long> currentCompaction2 =
+                WhiteboxImpl.getInternalState(persistentTopic, "currentCompaction");
```
lhotari, line 435: Again, avoid reflection/`WhiteboxImpl`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1340,18 +1341,42 @@ private void asyncDeleteCursorWithCleanCompactionLedger(PersistentSubscription s
...
+            if (!disablingCompaction.compareAndSet(false, true)) {
+                unsubscribeFuture.completeExceptionally(
```
lhotari, line 1347: this should be protected with `if (!unsubscribeFuture.isDone()) {` so that creating the exception could be avoided.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: Yes, it is. It prevents the modifying of the variable `currentCompaction`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: Could you mention the code line? I could not find it.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: Good point, I have changed the modifier

pulsar-broker/src/test/java/org/apache/pulsar/compaction/GetLastMessageIdCompactedTest.java
poorbarcode: Changed

pulsar-broker/src/test/java/org/apache/pulsar/compaction/GetLastMessageIdCompactedTest.java
poorbarcode: Changed

coderzc (approved):

dao-jun (approved):

lhotari:
@poorbarcode I fixed a checkstyle issue and a problem in using MockZooKeeper's `delay` method. After #23988, MockZooKeeper is correctly single threaded and a delay will block the execution of this code in ZKSessionWatcher:

This will trigger a ConnectionLost session event. That's why a delay should be kept under 2000 ms. 
Currently MockZooKeeper's session timeout is hard coded to 30000ms and the check interval is 1/15 of this, therefore 2000ms.

lhotari:
I created a separate PR to fix the invalid test that fails now: #24166

lhotari (approved):
LGTM. Good work @poorbarcode

PR #12606, opened 2021-11-03
[ML] Avoid passing OpAddEntry across a thread boundary in asyncAddEntry

lhotari:
### Motivation

There's a concern that passing OpAddEntry across thread boundaries has thread safety issues combined to the use of recycled OpAddEntry object instances.

### Modifications

To avoid any thread safety issues caused by passing OpAddEntry, the OpAddEntry instance can be created in the same thread where it is handled.

PR #23956, opened 2025-02-10
[WIP][improve][ci] Enable Netty leak detection in CI and make leaks fail the build

lhotari:
### Motivation

To prevent future memory leak regressions, it's useful to Netty leak detection enabled in CI and make leaks fail the build.

### Modifications

- add custom io.netty.util.ResourceLeakDetector implementation that logs leaks to files
- configure Netty leak detector for unit tests and integration tests
- show detected leaks directly in GitHub Action UI and make log records downloadable

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/lhotari/pulsar/pull/201

Issue #23865, opened 2025-01-18
[Bug] pulsar-io EventLoopGroup is used in dispatchers to execute synchronized code and CPU intensive work


lhotari:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

All released versions

### Minimal reproduce step

In dispatchers, `topic.getBrokerService().executor()`, the "pulsar-io" EventLoopGroup is used to execute synchronized code and CPU intensive work. The "pulsar-io" EventLoopGroup is intended to be the worker for Pulsar's ServerCnx Netty handler so that network connections remain responsive. Since dispatcher implementations contain multiple synchronized methods, blocking operations could impact all connections that are bound to the same IO thread. 

Another potential problem that the solution causes unnecessary context switching since the current solution doesn't use a single thread for the scheduled jobs and the work that it performs.

### What did you expect to see?

A pinned executor to be used for executing all operations for a dispatcher or replicator.

### What did you see instead?

EventLoopGroup is used as executor and the thread could change for each execution.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

lhotari:
this is a real performance issue with Pulsar. In #24623, I have added a solution for profiling test cases with Async Profiler with JFR integrated to it. It's possible to analyze the results with [Eclipse Mission Control](https://adoptium.net/jmc).

The lock analysis shows that dispatchers contribute to most blocking on pulsar-io threads:
<img width="1089" height="1408" alt="Image" src="https://github.com/user-attachments/assets/f3af93eb-469b-4fe4-a6a2-3b3990cfce99" />

This is a problem since Netty IO threads shouldn't ever be blocked since one thread is used across multiple connections to handle the asynchronous IO events. Blocking will cause a "noisy neighbor" problem across connections.

PR #9039, opened 2020-12-23
[Feature] Introduce continuous offset for pulsar

aloyszhang:
Fixes #9038 
### Motivation

As described in [PIP-70](https://github.com/apache/pulsar/wiki/PIP-70%3A-Introduce-lightweight-broker-entry-metadata).
One of the use case for Broker entry metadata is  providing continuous message sequence-Id for messages in one topic-partition which is useful for Protocol Hanlder like KOP.

This PR enable Pulsar to support continuous offset for message based on Broker entry metadata.

### Modifications

Introduce a new field for broker entry metadta named `offset`;
Introduce a new interceptor type `ManagedLedgerInterceptor` which intercept entry in `ManagedLedger`；
Each partition will be assigned a `ManagedLedgerInterceptor` when `ManagedLedger` created;
Each Entry will be intercept for adding a  monotone increasing offset in Broker entry metadata and the offet is added by batchSize of entry;
Support find position by a given offset.

PR #23983, opened 2025-02-13
[improve][broker] Don't call ManagedLedger#asyncAddEntry in Netty I/O thread

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/23940 brings a behavior change that the core logic of `ManagedLedger#asyncAddEntry` now won't switch threads, which means it will be executed directly in Netty I/O thread via `PersistentTopic#asyncAddEntry`.

The `beforeAddEntry` method calls the`intercept` and `interceptWithNumberOfMessages` methods for **all broker entry interceptors** and prepends a new broker entry metadata buffer on the original buffer (though it's just a composite buffer).

There is a risk that when many producers send messages to the same managed ledger concurrently, the process of `asyncAddEntry` might block the Netty I/O thread for some time and cause the performance regression.

### Modifications

In `PersistentTopic#publishMessage`, expose the `getExecutor()` method for `ManagedLedger` and execute `ManagedLedger#asyncAddEntry` in that executor. The change of https://github.com/apache/pulsar/pull/12606 is moved to `PersistentTopic` as well that the buffer is retained before switching to another thread. After that, only synchronize `afterAddEntryToQueue` with other synchronized methods of `ManagedLedgerImpl`.

`ManagedLedgerImpl#asyncAddEntry` still doesn't switch the thread, so it would still be possible for the downstream application to synchronize `asyncAddEntry`, either by adding a lock (e.g. `synchronized`) or executing this method is a single thread.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -671,8 +671,20 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
     }
 
     private void asyncAddEntry(ByteBuf headersAndPayload, PublishContext publishContext) {
-        ledger.asyncAddEntry(headersAndPayload,
-            (int) publishContext.getNumberOfMessages(), this, publishContext);
+        // Retain the buffer in advance to avoid the buffer might have been released when it's passed to `asyncAddEntry`
+        final var buffer = headersAndPayload.retain();
+        try {
+            ledger.getExecutor().execute(() -> {
+                try {
+                    ledger.asyncAddEntry(buffer, (int) publishContext.getNumberOfMessages(), this, publishContext);
+                } finally {
+                    buffer.release();
+                }
+            });
+        } catch (Exception e) {
+            buffer.release();
+            publishContext.completed(e, -1L, -1L);
+        }
     }
 
     public void asyncReadEntry(Position position, AsyncCallbacks.ReadEntryCallback callback, Object ctx) {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.NavigableMap;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
 import java.util.function.Predicate;
 import org.apache.bookkeeper.common.annotation.InterfaceAudience;
 import org.apache.bookkeeper.common.annotation.InterfaceStability;
@@ -199,6 +200,7 @@ public interface ManagedLedger {
      *            callback object
      * @param ctx
      *            opaque context
+     * @apiNote This method is not guaranteed to be thread safe. The caller should be responsible to call it in order.
      */
     void asyncAddEntry(ByteBuf buffer, int numberOfMessages, AddEntryCallback callback, Object ctx);
 
@@ -733,4 +735,13 @@ public interface ManagedLedger {
     }
 
     Position getFirstPosition();
+
+    /**
+     * In the internal implementation of a managed ledger, it is reasonable to use a single-thread executor to execute
+     * tasks that need to be performed in order. By exposing this internal executor, the caller can synchronize its
+     * custom tasks, as well as the internal tasks.
+     */
+    default Executor getExecutor() {
+        return Runnable::run;
+    }
 }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -798,26 +798,21 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
             log.debug("[{}] asyncAddEntry size={} state={}", name, buffer.readableBytes(), state);
         }
 
-        // retain buffer in this thread
+        // The buffer will be queued in `pendingAddEntries` and might be polled later in a different thread. However,
+        // the caller could release it after this method returns. To ensure the buffer is not released when it's polled,
+        // increase the reference count, which should be decreased by `OpAddEntry`'s methods later.
         buffer.retain();
-
-        // Jump to specific thread to avoid contention from writers writing from different threads
         final var addOperation = OpAddEntry.createNoRetainBuffer(this, buffer, numberOfMessages, callback, ctx,
                 currentLedgerTimeoutTriggered);
         var added = false;
         try {
-            // Use synchronized to ensure if `addOperation` is added to queue and fails later, it will be the first
-            // element in `pendingAddEntries`.
-            synchronized (this) {
-                if (managedLedgerInterceptor != null) {
-                    managedLedgerInterceptor.beforeAddEntry(addOperation, addOperation.getNumberOfMessages());
-                }
-                final var state = STATE_UPDATER.get(this);
-                beforeAddEntryToQueue(state);
-                pendingAddEntries.add(addOperation);
-                added = true;
-                afterAddEntryToQueue(state, addOperation);
+            if (managedLedgerInterceptor != null) {
+                managedLedgerInterceptor.beforeAddEntry(addOperation, addOperation.getNumberOfMessages());
             }
+            beforeAddEntryToQueue();
+            pendingAddEntries.add(addOperation);
+            added = true;
+            afterAddEntryToQueue(addOperation);
         } catch (Throwable throwable) {
             if (!added) {
                 addOperation.failed(ManagedLedgerException.getManagedLedgerException(throwable));
@@ -825,7 +820,8 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
         }
     }
 
-    protected void beforeAddEntryToQueue(State state) throws ManagedLedgerException {
+    protected void beforeAddEntryToQueue() throws ManagedLedgerException {
+        final var state = STATE_UPDATER.get(this);
         if (state.isFenced()) {
             throw new ManagedLedgerFencedException();
         }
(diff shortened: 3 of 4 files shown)


merlimat (commented):
An extra context switch for each entry is costly, especially when you have many small entries and little or no batching. That's why we put it on the same thread. 

If the interceptor needs to do expensive work, maybe only the interceptor part should be done in a different thread, though it shouldn't affect it when we don't use interceptor.

lhotari:
> An extra context switch for each entry is costly, especially when you have many small entries and little or no batching. That's why we put it on the same thread.

@merlimat The thread switching was added in PR https://github.com/apache/pulsar/pull/9039, already in December 2020. The reason to make this change is related to a performance concern of #23940 changes which removed the thread switching.

https://github.com/apache/pulsar/blob/ee5b13af5cf229c2e4846c6d34ebda59eb82330a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L796-L826

In Pulsar use cases, synchronization on CPU intensive operations (or blocking IO operations) in Netty IO threads could cause performance regressions. In this case, it would impact use cases where there's a large number of producers producing to a single topic.
Blocking IO threads will have a broader impact since it will impact Netty IO of all connections sharing the same IO thread.

Before #23940, the code looks like this:
https://github.com/apache/pulsar/blob/7a79c78f8e6f4b52f13be1c6441f4b007d9a00fe/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L796-L810

btw. In the Pulsar code base, we have a problem in how IO threads are used. IO threads are used to process work that shouldn't be handled with IO threads at all. I have created an issue #23865. There should be a separate thread pool for running blocking operations and CPU intensive synchronized operations.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -671,8 +671,15 @@ public void updateSubscribeRateLimiter() {
     }
 
     private void asyncAddEntry(ByteBuf headersAndPayload, PublishContext publishContext) {
-        ledger.asyncAddEntry(headersAndPayload,
-            (int) publishContext.getNumberOfMessages(), this, publishContext);
+        // Retain the buffer in advance to avoid the buffer might have been released when it's passed to `asyncAddEntry`
+        final var buffer = headersAndPayload.retain();
```
lhotari, line 675: Just wondering if this is going to add an extra reference compared to the previous code. 
I have a solution in #23956 to reliably detect Netty ByteBuf leaks so those changes are one way to validate whether there's a new leak.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -671,8 +671,15 @@ public void updateSubscribeRateLimiter() {
...
+        try {
+            ledger.getExecutor().execute(() -> ledger.asyncAddEntry(buffer, (int) publishContext.getNumberOfMessages(),
+                    this, publishContext));
```
lhotari, line 678: this should be an internal concern of the ManagedLedger implementation's `asyncAddEntry` method. One reason for this is that there's multiple `asyncAddEntry` signatures and we'd like to add them all in order.

lhotari (commented):
Great work @BewareMyPower. Some comments added in this first pass.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
BewareMyPower: Oh yes, it's a memory leak. Let me think how to handle it better.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
BewareMyPower: I think we should restore the `synchronized` keyword to the `asyncAddEntry` method to make it thread safe.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
BewareMyPower: Oh, I realized that I made a regression in https://github.com/apache/pulsar/pull/23983/commits/61dcac775d35e4abfada6af749ae7cdf66405248

BewareMyPower:
> @merlimat The thread switching was added in PR https://github.com/apache/pulsar/pull/9039, already in December 2020.

@merlimat @lhotari to correct it, this is the very early behavior introduced in https://github.com/apache/pulsar/pull/1521.

This PR intends to decouple `ManagedLedger#asyncAddEntry` and `PersistentTopic#asyncAddEntry` so that the managed ledger interface can be more flexible for the downstream protocol handlers to use.

After that, all write operations from Pulsar client will still keep the original behavior that switches to managed ledger's executor to call `ManagedLedger#asyncAddEntry`.

However, regarding the downstream, for example, in my Kafka protocol handler implementation, `PersistentTopic#publishMessage` is not called in an I/O thread. Instead, it's called in an independent worker thread. Then I can choose to call `persistentTopic.getManagedLedger().asyncAddEntry(/* ... */)` in order, which can be achieved by adding the `synchronized` keyword or using the same worker thread for the same topic.

The comment [here](https://github.com/apache/pulsar/pull/23983#pullrequestreview-2615729014) makes sense to a certain extent, but it might be a new topic (e.g. thread switching vs. synchronized) to discuss, which is beyond the scope of this PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
lhotari: There are 2 different aspects to consider: thread safety and ordering.

Regarding "I still think the synchronization should be performed from the caller":
In Java, synchronization is not only about performing operations one by one under a mutually exclusive lock. "Visibility" is an important aspect of Java thread safety. That's why it doesn't make sense for callers to synchronize calls to `asyncAddEntry` since all callers would need to use the same lock for both ordering and thread safety.

Snippet from "Java Concurrency in Practice", Chapter 2 "Thread safety":
> It is a common mistake to assume that synchronization needs to be used only when writing to shared variables; this is simply not true. 
> For each mutable state variable that may be accessed by more than one thread, all accesses to that variable must be performed **with the same lock held**. In this case, we say that the variable is guarded by that lock.

How a ManagedLedger implementation achieves ordering guarantees and thread safety is an internal implementation detail. In the case of ManagedLedger, it doesn't make sense to delegate the responsibility of thread safety to the caller.

Another downside of the `ledger.getExecutor().execute` solution is that it exposes internal implementation details that callers of the API must be aware of. This is not great API design when such implementation details are exposed.

I think we need to consider a different solution. I can see some code examples in #23940 of what needs to be solved. To me, it seems this could be solved with the `Object ctx` parameter by passing a `ctx` that is also understood by the interceptor. @BewareMyPower, would you be able to research that type of solution instead?

lhotari (changes requested):
Please check the review comments

BewareMyPower:
Let me close this PR first.

> An extra context switch for each entry is costly

It's true, but we still need a benchmark. I ran a simple test by adding the metrics to `ManagedLedgerMBeanImpl`

```java
    @Getter
    private final StatsBuckets threadSwitchingStats = new StatsBuckets(ENTRY_LATENCY_BUCKETS_USEC);
    @Getter
    private final StatsBuckets executeStats = new StatsBuckets(ENTRY_LATENCY_BUCKETS_USEC);
    @Getter
    private final StatsBuckets asyncAddEntryStats = new StatsBuckets(ENTRY_LATENCY_BUCKETS_USEC);
```

and then record the latency in `asyncAddEntry`:

```java
        final var beforeExecute = System.nanoTime();
        executor.execute(() -> {
            final var afterExecute = System.nanoTime();
            OpAddEntry addOperation = OpAddEntry.createNoRetainBuffer(this, buffer, numberOfMessages, callback, ctx,
                    currentLedgerTimeoutTriggered);
            internalAsyncAddEntry(addOperation);
            final var completeAsyncAddEntry = System.nanoTime();
            mbean.getThreadSwitchingStats().addValue(afterExecute - beforeExecute);
            mbean.getAsyncAddEntryStats().addValue(completeAsyncAddEntry - afterExecute);
        });
        final var afterExecute = System.nanoTime();
        mbean.getExecuteStats().addValue(afterExecute - beforeExecute);
```

Then I added a test:

```java
    @Test
    public void testThreadSwitchingLatency() throws Exception {
        ManagedLedgerImpl ml = (ManagedLedgerImpl) factory.open("ml");
        final var data = new byte[1000];
        for (int i = 0; i < 1000; i++) {
            ml.addEntry(data);
        }
        ml.mbean.getThreadSwitchingStats().refresh();
        ml.mbean.getAsyncAddEntryStats().refresh();
        ml.mbean.getExecuteStats().refresh();
        log.info("1: {}", ml.mbean.getThreadSwitchingStats().getAvg());
        log.info("2: {}", ml.mbean.getAsyncAddEntryStats().getAvg());
        log.info("3: {}", ml.mbean.getExecuteStats().getAvg());
    }
```

Outputs:

```
1: 26924.005
2: 35933.837
3: 8866.484
```

1. Task execution delay: 27 us
2. Asynchronous operation itself: 36 us
3. The `execute` method call itself: 9 us

All of these should not be a significant impact because the whole add latency is milliseconds level.

The main concern is that if `internalAsyncAddEntry` could cost much time, blocking the Netty I/O thread might have an impact for other requests. Hence, switching to a dedicated thread pool (managed ledger's executor) might make sense.

We should not use `synchronized` to replace `executor.execute` here, because currently `ManagedLedgerImpl` uses `synchronized` as a coarse-grained lock in many other places as well:
- `asyncOpenCursor`
- cursor initialization callback in `asyncOpenCursor`
- `asyncDeleteCursor`
- `newNonDurableCursor`
- `getEstimatedBacklogSize`
- `asyncTerminate`
- ...

It's hard to analyze how much time is blocked by acquiring the lock.

Using a dedicated lock for `asyncAddEntry` should make sense, which is much more efficient than thread switching. However, we cannot simply replace `synchronized` with a different lock on `internalAsyncAddEntry` because it could access fields that are protected by `synchronized` in other threads.

In summary, the current design should not be changed, including:
- Adding `synchronized` on `internalAsyncAddEntry`
- Switching to managed ledger's executor to execute `internalAsyncAddEntry`

We should not care much about the thread switching because its latency is microseconds level.

Issue #23885, opened 2025-01-23
Flaky-test: NegativeAcksTest.testNegativeAcksWithBackoff


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failures

- [2025-01-21T06:22:11.8350781Z](https://github.com/apache/pulsar/actions/runs/12881023585/job/35911260374#step:10:690) 

### Exception stacktrace

```
java.lang.AssertionError: expected [null] but found [org.apache.pulsar.client.impl.MessageImpl@6c98609e]
	at org.testng.Assert.fail(Assert.java:110)
	at org.testng.Assert.failNotSame(Assert.java:1573)
	at org.testng.Assert.assertNull(Assert.java:1506)
	at org.testng.Assert.assertNull(Assert.java:1494)
	at org.apache.pulsar.client.impl.NegativeAcksTest.testNegativeAcksWithBackoff(NegativeAcksTest.java:285)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:11)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
[...]

PR #23986, opened 2025-02-14
[fix][test] fix flaky testNegativeAcksWithBackoff when batch enabled.

thetumbled:
Main Issue: https://github.com/apache/pulsar/issues/23885

### Motivation

When batch message is negative ack, some undetermined behavior may occur. For example, 
- There are 10 messages batched into one message, all of them are unbatch into consumer's receiver queue.
- consumer negative ack the first 5 messages.
- before the consumer negative ack another 5 messages, the negatve ack tracker trigger the redelivery request.
- broker dispatch this batched message to consumer.
- consumer clear the remaining 5 messages in the queue, and refill 10 messages into the queue.
- Message duplication issue occurs.
IMO, suck kind of problem is reasonable as we only provide At-Least-Once semantices.

### Modifications

This unit test don't handle the batch case at some time. 
Negative ack one message is enough for this test in batch case.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
@@ -255,10 +255,16 @@ public class NegativeAcksTest extends ProducerConsumerBase {
         final int redeliverCount = 5;
         long firstReceivedAt = System.currentTimeMillis();
         long expectedTotalRedeliveryDelay = 0;
+        Message<String> msg = null;
         for (int i = 0; i < redeliverCount; i++) {
             for (int j = 0; j < N; j++) {
-                Message<String> msg = consumer.receive();
+                msg = consumer.receive();
                 log.info("Received message {}", msg.getValue());
+                if (!batching) {
+                    consumer.negativeAcknowledge(msg);
+                }
+            }
+            if (batching) {
                 consumer.negativeAcknowledge(msg);
             }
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
```diff
@@ -255,10 +255,17 @@ public void testNegativeAcksWithBackoff(boolean batching, boolean usePartitions,
         final int redeliverCount = 5;
         long firstReceivedAt = System.currentTimeMillis();
         long expectedTotalRedeliveryDelay = 0;
+        Message<String> msg = null;
```
lhotari, line 258: I think that this change is not relevant. In a good coding style it's better to reduce the scope of variables to minimal instead of reusing variables for different purposes.

lhotari (approved):

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
thetumbled: Updated.

lhotari (approved):

PR #23984, opened 2025-02-13
[fix][meta] Fix ephemeral Zookeeper put which creates a persistent znode

heesung-sn:
Fixes https://github.com/apache/pulsar/issues/23889

### Motivation

Fixes https://github.com/apache/pulsar/issues/23889
zk.put creates persistent znode although it passes ephemeral node creation option.

### Modifications

- pass the creation option when set fails(if the node exists already

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/heesung-sn/pulsar/pull/87

PR #23988, opened 2025-02-14
[fix][meta] Fix ephemeral handling of ZK nodes and fix MockZooKeeper ephemeral and ZK stat handling

lhotari:
### Motivation

This PR is a follow-up to #23984 to fix ephemeral status handling in ZKMetadataStore and to improve the ZooKeeper testing infrastructure.

A key addition is enabling pulsar-metadata tests to run against MockZooKeeper through a proper MetadataStoreProvider implementation, ensuring MockZooKeeper correctly implements all features that Pulsar depends on.

### Modifications

1. Fixed Production Code Issues:
  - Fixed ephemeral status handling in ZKMetadataStore to properly detect ephemeral nodes by using the correct comparison value (0L) for non-ephemeral (== persistent) nodes

2. MockZooKeeper improvements and fixes:
  - Fixed ephemeral node handling and cleanup on session termination
  - Improved ZK stat handling with proper timestamps and version numbers
  - Added proper session ID management and consistent thread safety
  - Fixed persistent watcher cleanup on session termination
  - Fixed getChildren implementation and removed code duplication
  - Fixed ordering issues by migrating to use a single threaded executor model
    - This allowed to remove locks and synchronization in most cases

3. Test Framework Improvements::
  - Implemented MockZooKeeperMetadataStoreProvider to allow pulsar-metadata tests to use MockZooKeeper
    - This ensures that MockZooKeeper correctly implements all MetadataStore features that Pulsar depends on
  - Enhanced PulsarTestContext to support toggling between MockZooKeeper and real ZK (TestZKServer) implementations
  - Added ability to run MockedPulsarServiceBaseTest tests with both MockZooKeeper and real ZK (TestZKServer)

There are some unrelated NPE fixes and improvements for better exception messages. I faced those issues while debugging BrokerServiceTest and noticed those exceptions in the execution logs. Some of the exceptions were due to invalid tests, however the improvements will be useful also later.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/LoadManagerShared.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/LoadManagerShared.java
@@ -285,7 +285,7 @@ public class LoadManagerShared {
     public static String getBundleRangeFromBundleName(String bundleName) {
         // the bundle format is property/cluster/namespace/0x00000000_0xFFFFFFFF
         int pos = bundleName.lastIndexOf("/");
-        checkArgument(pos != -1);
+        checkArgument(pos != -1, "Invalid bundle name format: %s", bundleName);
         return bundleName.substring(pos + 1);
     }
 
@@ -293,7 +293,7 @@ public class LoadManagerShared {
     public static String getNamespaceNameFromBundleName(String bundleName) {
         // the bundle format is property/cluster/namespace/0x00000000_0xFFFFFFFF
         int pos = bundleName.lastIndexOf('/');
-        checkArgument(pos != -1);
+        checkArgument(pos != -1, "Invalid bundle name format: %s", bundleName);
         return bundleName.substring(0, pos);
     }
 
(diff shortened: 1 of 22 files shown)


lhotari:
There might be production code issues too. 
Failures with TestZKServer in org.apache.pulsar.client.api.BrokerServiceLookupTest#testMultipleBrokerLookup and org.apache.pulsar.client.api.BrokerServiceLookupTest#testWebserviceServiceTls
Logs in https://gist.github.com/lhotari/f00645438de72b992c1a48def0fce59e

heesung-sn (approved):

lhotari:
cherry-picking depends on #23686

lhotari:
> Some Pulsar tests are using MockZooKeeper's `delay` method. After #23988, MockZooKeeper is correctly single threaded and a delay will block the execution of this code in ZKSessionWatcher:
> 
> https://github.com/apache/pulsar/blob/6e8c3496647df275071190aec9d29ecbcee55e80/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L108-L115
> 
> This will trigger a ConnectionLost session event. That's why a delay should be kept under 2000 ms. Currently MockZooKeeper's session timeout is hard coded to 30000ms and the check interval is 1/15 of this, therefore 2000ms. It would be a good improvement to use the configured `metadataStoreSessionTimeoutMillis` instead.

This is handled in #24171

PR #23989, opened 2025-02-15
[fix][client] add maxRetryRequestTimes in ClientConfigurationData and remove Duplicate retry logic

walkinggo:
### Motivation
The maxRetries property of AsyncHttpConnector is derived from the config of DefaultAsyncHttpClient. 
```
@SneakyThrows
    public AsyncHttpConnector(int connectTimeoutMs, int readTimeoutMs,
                              int requestTimeoutMs,
                              int autoCertRefreshTimeSeconds, ClientConfigurationData conf,
                              boolean acceptGzipCompression) {
        Validate.notEmpty(conf.getServiceUrl(), "Service URL is not provided");
        serviceNameResolver = new PulsarServiceNameResolver();
        String serviceUrl = conf.getServiceUrl();
        serviceNameResolver.updateServiceUrl(serviceUrl);
        this.acceptGzipCompression = acceptGzipCompression;
        AsyncHttpClientConfig asyncHttpClientConfig =
                createAsyncHttpClientConfig(conf, connectTimeoutMs, readTimeoutMs, requestTimeoutMs,
                        autoCertRefreshTimeSeconds);
        httpClient = createAsyncHttpClient(asyncHttpClientConfig);
        this.requestTimeout = requestTimeoutMs > 0 ? Duration.ofMillis(requestTimeoutMs) : null;
        this.maxRetries = httpClient.getConfig().getMaxRequestRetry();
    }
```
This causes both DefaultAsyncHttpClient and AsyncHttpConnector to share the same retry count value. However, since both components implement their own retry logic, this configuration actually results in accumulated retries exceeding the configured limit. Given that DefaultAsyncHttpClient has a default retry count of 5, I propose disabling retries in DefaultAsyncHttpClient (setting its retry count to 0) while maintaining the default configuration, to centralize retry handling exclusively within AsyncHttpConnector.

### Modifications

1. **Retry Configuration Support**  
Added `maxRetryRequestTimes` in `ClientConfigurationData` (default: 5) to control admin client request retry behavior. Configuration propagates to `AsyncHttpConnector` implementations.

2. **Client API Enhancements**  
- Extended `ClientBuilder` interface with new `maxRetryTimes()` method
- Implemented configuration injection in `ClientBuilderImpl`

3. **HTTP Connector Refactoring**  
- Modified `AsyncHttpConnector` to use centralized configuration instead of AsyncHttpClient's internal retry setting
- Explicitly disabled underlying library retry via `confBuilder.setMaxRequestRetry(0)`

4. **Validation**  
Added parameter verification test in `ClientBuilderImplTest` to validate retry time configuration constraint

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/walkinggo/pulsar/pull/9

### DIFF
--- a/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
+++ b/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
@@ -130,7 +130,7 @@ public class AsyncHttpConnector implements Connector, AsyncHttpRequestExecutor {
                         autoCertRefreshTimeSeconds);
         httpClient = createAsyncHttpClient(asyncHttpClientConfig);
         this.requestTimeout = requestTimeoutMs > 0 ? Duration.ofMillis(requestTimeoutMs) : null;
-        this.maxRetries = httpClient.getConfig().getMaxRequestRetry();
+        this.maxRetries = conf.getMaxRetryRequestTimes();
     }
 
     private AsyncHttpClientConfig createAsyncHttpClientConfig(ClientConfigurationData conf, int connectTimeoutMs,
@@ -178,6 +178,7 @@ public class AsyncHttpConnector implements Connector, AsyncHttpRequestExecutor {
             }
         });
         confBuilder.setDisableHttpsEndpointIdentificationAlgorithm(!conf.isTlsHostnameVerificationEnable());
+        confBuilder.setMaxRequestRetry(0);
     }
 
     protected AsyncHttpClient createAsyncHttpClient(AsyncHttpClientConfig asyncHttpClientConfig) {
@@ -293,7 +294,6 @@ public class AsyncHttpConnector implements Connector, AsyncHttpRequestExecutor {
         return resultFuture;
     }
 
-    // TODO: There are problems with this solution since AsyncHttpClient already contains logic to retry requests.
     // This solution doesn't contain backoff handling.
     private <T> void retryOperation(
             final CompletableFuture<T> resultFuture,
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
@@ -679,4 +679,12 @@ public interface ClientBuilder extends Serializable, Cloneable {
      * - The `loadManagerClassName` config in broker is a class that implements the `ExtensibleLoadManager` interface
      */
     ClientBuilder lookupProperties(Map<String, String> properties);
+
+    /**
+     * Set the max retry times for a request.
+     *
+     * @param maxRetryTimes
+     * @return the client builder instance
+     */
+    ClientBuilder maxRetryTimes(int maxRetryTimes);
 }
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientBuilderImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientBuilderImpl.java
@@ -482,4 +482,10 @@ public class ClientBuilderImpl implements ClientBuilder {
         conf.setLookupProperties(properties);
         return this;
     }
+
+    @Override
+    public ClientBuilder maxRetryTimes(int maxRetryTimes) {
+        conf.setMaxRetryRequestTimes(maxRetryTimes);
+        return this;
+    }
 }
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java
@@ -374,6 +374,12 @@ public class ClientConfigurationData implements Serializable, Cloneable {
     private String dnsLookupBindAddress = null;
 
     @ApiModelProperty(
+            name = "maxRetryRequestTimes",
+            value = "The Pulsar admin client max retry request times, default value is 5"
+    )
+    private int maxRetryRequestTimes = 5;
+
+    @ApiModelProperty(
             name = "dnsLookupBindPort",
             value = "The Pulsar client dns lookup bind port, takes effect when dnsLookupBindAddress is configured,"
                     + " default value is 0."
--- a/pulsar-client/src/test/java/org/apache/pulsar/client/impl/ClientBuilderImplTest.java
+++ b/pulsar-client/src/test/java/org/apache/pulsar/client/impl/ClientBuilderImplTest.java
@@ -170,6 +170,11 @@ public class ClientBuilderImplTest {
         assertThatAuthIsNotSet(auth);
     }
 
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testClientBuilderWithMaxRetryTime() throws PulsarClientException {
+        PulsarClient.builder().maxRetryTimes(5).build();
+    }
+
     @Test
     public void testLoadConfAuthNotSetWhenEmptyAuthParamsSpecified() {
         Map<String, Object> confProps = new HashMap<>();


walkinggo:
> Changes related to configuration modification need a pip.

Apologies, this PR was created in response to an existing TODO comment. I'll promptly draft a PIP document and submit it to the community mailing list.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
```diff
@@ -178,6 +178,7 @@ public boolean keepAlive(InetSocketAddress remoteAddress, Request ahcRequest,
             }
         });
         confBuilder.setDisableHttpsEndpointIdentificationAlgorithm(!conf.isTlsHostnameVerificationEnable());
+        confBuilder.setMaxRequestRetry(0);
```
thetumbled, line 181: Adding some comment on it will be better.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java
```diff
@@ -373,6 +373,12 @@ public class ClientConfigurationData implements Serializable, Cloneable {
     )
     private String dnsLookupBindAddress = null;
 
+    @ApiModelProperty(
+            name = "maxRetryRequestTimes",
+            value = "The Pulsar admin client max retry request times, default value is 5"
```
thetumbled, line 378: Need to illustrate more about this conf. 
This configuration works together with following conf:
```
    @ApiModelProperty(
            name = "connectionTimeoutMs",
            value = "Duration of waiting for a connection to a broker to be established."
                    + "If the duration passes without a response from a broker, the connection attempt is dropped."
    )
    private int connectionTimeoutMs = 10000;
    @ApiModelProperty(
            name = "requestTimeoutMs",
            value = "Maximum duration for completing a request."
    )
    private int requestTimeoutMs = 60000;

    @ApiModelProperty(
            name = "readTimeoutMs",
            value = "Maximum read time of a request."
    )
    private int readTimeoutMs = 60000;
```
The max retry count may  be unreachable.

thetumbled (commented):
Good catch! I leave some comments.
Since this pr aim to fix a bug, could you add a unit test to reproduce the bug that max retry times double the maxRequestRetry in the httpclient's configuration?

walkinggo:
> Good catch! I leave some comments. Since this pr aim to fix a bug, could you add a unit test to reproduce the bug that max retry times double the maxRequestRetry in the httpclient's configuration?

I'm afraid adding corresponding test cases here could be quite challenging. I'm not sure how to validate the service's retry count – perhaps we could verify whether the configuration file is successfully set through test cases.

The retry logic is specifically implemented in the onFailure(Channel channel, Throwable cause) method of class org.asynchttpclient.netty.channel.NettyConnectListener from async-http-client.

On a related note, I don't think the retry logic would result in double the attempts. Since the admin-client invokes async-http-client for each request, the actual retry attempts should be the square of the original configured count (n² if n is the configured retry value).

thetumbled:
> > Good catch! I leave some comments. Since this pr aim to fix a bug, could you add a unit test to reproduce the bug that max retry times double the maxRequestRetry in the httpclient's configuration?
> 
> I'm afraid adding corresponding test cases here could be quite challenging. I'm not sure how to validate the service's retry count – perhaps we could verify whether the configuration file is successfully set through test cases.
> 
> The retry logic is specifically implemented in the onFailure(Channel channel, Throwable cause) method of class org.asynchttpclient.netty.channel.NettyConnectListener from async-http-client.
> 
> On a related note, I don't think the retry logic would result in double the attempts. Since the admin-client invokes async-http-client for each request, the actual retry attempts should be the square of the original configured count (n² if n is the configured retry value).

Or have you meet this problem in real enviroment? We should be cautious about the so-called bug fix right? what if the real behaviour is correct? 
IMO,  Maybe you can push forward this pip without the reproducer, but the reproducer will be favourable.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
```diff
@@ -718,6 +718,13 @@ ClientBuilder authentication(String authPluginClassName, Map<String, String> aut
      */
     ClientBuilder lookupProperties(Map<String, String> properties);
 
+    /**
+     * Set the max retry times for a request.
+     *
+     * @param maxRetryTimes
+     * @return the client builder instance
+     */
+    ClientBuilder maxRetryTimes(int maxRetryTimes);
```
lhotari: `maxRetryTimes` is not very descriptive. Since this is about http client retries, it should be also explained in the javadoc and the method name and parameter name should also make that obvious.

`maxHttpRequestRetries` would be one choice.

lhotari: ClientBuilder isn't used for PulsarAdmin client. It's the org.apache.pulsar.client.admin.PulsarAdminBuilder that is used. I'd expect this setting to be in PulsarAdminBuilder instead of ClientBuilder, however there's also a Http client in the Pulsar client itself which is used when the provided serviceUrl is a http/https url.

lhotari: The Pulsar Admin Builder implementation also uses the ClientConfigurationData to hold configuration data (which isn't the greatest solution since not all configuration applies to both ClientBuilder and PulsarAdminBuilder).

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
```diff
@@ -721,10 +721,10 @@ ClientBuilder authentication(String authPluginClassName, Map<String, String> aut
     /**
      * Set the max retry times for a request.
      *
-     * @param maxRetryTimes
+     * @param maxHttpRequestRetries
      * @return the client builder instance
      */
-    ClientBuilder maxRetryTimes(int maxRetryTimes);
+    ClientBuilder MaxHttpRequestRetries(int maxHttpRequestRetries);
```
lhotari:
```suggestion
    ClientBuilder maxHttpRequestRetries(int maxHttpRequestRetries);
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientBuilderImpl.java
```diff
@@ -509,8 +509,8 @@ public ClientBuilder lookupProperties(Map<String, String> properties) {
     }
 
     @Override
-    public ClientBuilder maxRetryTimes(int maxRetryTimes) {
-        conf.setMaxRetryTimes(maxRetryTimes);
+    public ClientBuilder MaxHttpRequestRetries(int maxHttpRequestRetries) {
```
lhotari:
```suggestion
    public ClientBuilder maxHttpRequestRetries(int maxHttpRequestRetries) {
```

pulsar-client/src/test/java/org/apache/pulsar/client/impl/ClientBuilderImplTest.java
```diff
@@ -174,7 +174,7 @@ public void testLoadConfAuthNotSetWhenNoPropsAvailable() {
 
     @Test(expectedExceptions = IllegalArgumentException.class)
     public void testClientBuilderWithMaxRetryTime() throws PulsarClientException {
-        PulsarClient.builder().maxRetryTimes(5).build();
+        PulsarClient.builder().MaxHttpRequestRetries(5).build();
```
lhotari:
```suggestion
        PulsarClient.builder().maxHttpRequestRetries(5).build();
```

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
```diff
@@ -721,10 +721,10 @@ ClientBuilder authentication(String authPluginClassName, Map<String, String> aut
     /**
      * Set the max retry times for a request.
```
lhotari:
```suggestion
     * Set the maximum number of retries for http requests.
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java
lhotari: update the name

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
walkinggo: > The Pulsar Admin Builder implementation also uses the ClientConfigurationData to hold configuration data (which isn't the greatest solution since not all configuration applies to both ClientBuilder and PulsarAdminBuilder).

I think this is also a part that needs to be revised, and I will try to clarify their relationships.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/ClientBuilderImplTest.java
```diff
@@ -172,6 +172,11 @@ public void testLoadConfAuthNotSetWhenNoPropsAvailable() {
...
     }
 
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testClientBuilderWithMaxRetryTime() throws PulsarClientException {
+        PulsarClient.builder().maxHttpRequestRetries(5).build();
+    }
+
     @Test
     public void testLoadConfAuthNotSetWhenEmptyAuthParamsSpecified() {
         Map<String, Object> confProps = new HashMap<>();
```
nodece, line 182: Please add tests to cover the default value and different configured values for `maxHttpRequestRetries`. Verify both the setter and getter behavior.

nodece: `testClientBuilderWithMaxRetryTime` looks meaningless.

PR #23995, opened 2025-02-17
[improve][meta] Stop election operations when LeaderElectionImpl gets closed

lhotari:
### Motivation

In some tests, I noticed that LeaderElectionImpl could continue to make operations and override the closed state.

### Modifications

Add checks if the internalState is Closed

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/LeaderElectionImpl.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/LeaderElectionImpl.java
@@ -104,6 +104,9 @@ class LeaderElectionImpl<T> implements LeaderElection<T> {
     }
 
     private synchronized CompletableFuture<LeaderElectionState> elect() {
+        if (internalState == InternalState.Closed) {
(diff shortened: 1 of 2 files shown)


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/LeaderElectionImpl.java
```diff
@@ -113,6 +116,11 @@ private synchronized CompletableFuture<LeaderElectionState> elect() {
                 return tryToBecomeLeader();
             }
         }).thenCompose(leaderElectionState -> {
+            synchronized (this) {
+                if (internalState == InternalState.Closed) {
```
dao-jun, line 120: Need to add `volatile` on `internalState`?

lhotari: Already covered with `synchronized (this)`. `this` points to `LeaderElectionImpl` instance here.

PR #24012, opened 2025-02-21
[fix][broker] PIP-322 Fix issue with rate limiters where rates can exceed limits initially and consumption pauses until token balance is positive

### LINKED ISSUE #23920: [Bug] Producers getting timed out intermittently

### LINKED ISSUE #24001: [Bug] Rate-Limiter fails with a huge spike in a traffic, and publish/consume stuck for a longer time.

lhotari:
Fixes #23920
Fixes #24001

### Motivation

["PIP-322: Pulsar Rate Limiting Refactoring"](https://github.com/apache/pulsar/blob/master/pip/pip-322.md) changes caused a regression in dispatcher rate limiter and publish rate limiter behavior. When consumption starts, the rate can initially exceed limits, and consumption pauses until the token balance has caught up.

Due to the design of AsyncTokenBucket, the token balance is updated periodically as part of the usage calls. There is no separate scheduler required to update the token balance.

In the initial PIP-322 implementation, consistent token balance calculation was avoided since there was an assumption that the solution to calculate a consistent token balance every 16 milliseconds (configurable) was a sufficient solution. This assumption caused problems since in 16 milliseconds, the rate limiter could let a lot of traffic through which would be accounted by throttling when the token balance got updated. The tests `DispatchRateLimiterOverconsumingTest` and `PublishRateLimiterOverconsumingTest` added in this PR reproduced this issue and will help prevent future regressions in this area.

### Modifications

- When externally returning the token balance for the AsyncTokenBucket, always return a consistent token balance. Remove the now unnecessary options for configurable consistency.
  - leave a configurable option `addTokensResolutionNanos` which configures the interval when new tokens get added to the token balance.
  - For dispatch rate limiters, the PIP-322 implementation will behave in the same way as the original dispatch rate limiter implementation when the tokens get added once per second (rate period) as the previous implementation does.
- Make the dispatch rate limiter implementation configurable with a feature flag
  - Return old implementation for dispatch rate limiter and make it possible to use it
- For AsyncTokenBucket, use a clock implementation that delegates directly to `System.nanoTime()`
 
### Verifying this change

This PR adds a new test case `DispatchRateLimiterOverconsumingTest` which attempts to reproduce the issue described in #24001. The test runs on both PIP-322 implementation and the old dispatch rate limiter implementation.

There's currently similar behavior for both implementations. The rate of the first second of the message receiving results in a rate that is 2x the configured rate. This issue doesn't reproduce exactly in the same way on branch-3.0 (experimental commit: https://github.com/lhotari/pulsar/commit/a602428327cce088566fa90babf1a86aacd441b9), however also on branch-3.0, the behavior of the dispatch rate limiter is not always stable for the first 2 seconds. The test has been modified to allow this behavior and this problem can be investigated later since it's not a regression in the dispatch rate limiters.

This PR also reproduces the issue #23920 in a new test case `PublishRateLimiterOverconsumingTest` and demonstrates that this PR addresses the issue.

### JMH benchmark results

XPS 15 7590 (2019), i9-9980HK CPU @ 2.40GHz
```
Benchmark                                                    Mode  Cnt          Score           Error  Units
AsyncTokenBucketBenchmark.consumeTokensBenchmark001Threads  thrpt    3   49256534.199 ±   2696712.077  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark010Threads  thrpt    3  355694311.827 ± 193606239.269  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark100Threads  thrpt    3  439857131.356 ± 148199215.903  ops/s
DefaultMonotonicClockBenchmark.getTickNanos001Threads       thrpt    3   69711113.428 ±   9052629.012  ops/s
DefaultMonotonicClockBenchmark.getTickNanos010Threads       thrpt    3  475899645.775 ± 183135317.282  ops/s
DefaultMonotonicClockBenchmark.getTickNanos100Threads       thrpt    3  616883563.660 ± 269262678.166  ops/s
```
About 440 million ops for the AsyncTokenBucketBenchmark with 100 threads.

On Mac M3 Max, the results are now poor, possibly due to `System.nanoTime()` bottleneck.
```
Benchmark                                                    Mode  Cnt         Score          Error  Units
AsyncTokenBucketBenchmark.consumeTokensBenchmark001Threads  thrpt    3  84775402.806 ± 21346715.569  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark010Threads  thrpt    3  22746705.158 ±  2899591.678  ops/s
AsyncTokenBucketBenchmark.consumeTokensBenchmark100Threads  thrpt    3  17764575.534 ±  1805527.872  ops/s
DefaultMonotonicClockBenchmark.getTickNanos001Threads       thrpt    3  99204803.147 ±  5349820.538  ops/s
DefaultMonotonicClockBenchmark.getTickNanos010Threads       thrpt    3  22802728.081 ±  3517752.042  ops/s
DefaultMonotonicClockBenchmark.getTickNanos100Threads       thrpt    3  17440254.540 ±   711659.845  ops/s
```

The DefaultMonotonicClockBenchmark now measures `System.nanoTime()`. On MacOS, performance degrades when more threads are added. That's why the AsyncTokenBucketBenchmark numbers are also bad on MacOS. 
Previously, the implementation avoided calling System.nanoTime, which added complexity. Since Pulsar deployments target mainly Linux OS, there isn't a need to address the issue now.

The source of slowness of MacOS System.nanoTime can be found in the native code: https://github.com/openjdk/jdk21u/blob/b7d92cd0ae781de6c51ea965c16bd6e0c396e9f7/src/hotspot/os/bsd/os_bsd.cpp#L760-L782 
Explained in https://serce.me/posts/2019-05-16-the-matter-of-time#the-nanos-of-the-current-time
Linux uses the Posix implementation in Java 21 and doesn't have such a scalability issue: https://github.com/openjdk/jdk21u/blob/b7d92cd0ae781de6c51ea965c16bd6e0c396e9f7/src/hotspot/os/posix/os_posix.cpp#L1414-L1438

This also PR adds instructions how to run the JMH benchmarks with async-profiler. Initial profiling confirms the assumption that MacOS `System.nanoTime` implementation is the bottleneck.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


rdhabalia:
should we block this PR with the same reason you provided in another PR: https://github.com/apache/pulsar/pull/24002 where you said it's fixed as a part of https://github.com/apache/pulsar/pull/23930. and not prepared for taking for anyone's feedback and concerns. ? 
FIrst of all, we need to disable this rate-limiting for dispatcher until we have stable one in place as it's blocking and impacting many systems.

lhotari:
> The test case in this PR demonstrates this from an end-to-end perspective. While the excessive negative tokens problem is solved when using AsyncTokenBucket, there's another challenging issue that would require major changes to how dispatchers acquire tokens.

It seems that there's some other regression in both branch-3.3 and master that causes the first second rate to be 2x of the configured rate. Now that the pre-PIP-322 rate limiter is used in this PR, it shows that this isn't related to the use of AsyncTokenBucket or branch-4.0 changes since the problem is present in branch-3.3. (when I cherry-pick the changes)

lhotari:
I pushed changes to add `dispatchRateLimiterFactoryClassName` setting for toggling between the "classic" (before PIP-322) implementation and PIP-322 implementation using AsyncTokenBucket. We can default to the "classic" implementation which reverts the PIP-322 changes for dispatch rate limiter until the default configuration value is changed.

rdhabalia:
> I apologize for the communication issues on my end. I'm definitely taking all feedback and concerns seriously, though it does take me some time to address everything properly.

if that's true, then please rethink about this rate-limiter, token-bucket algorithm cab be exploited by malicious or greedy clients and your implementation to prevent it by introducing -ve token has serious implications of delaying or making topics unavailable without any reason. I implemented dispatch-rate limiter 3-4 years ago and we have been using it without any issue with highest accuracy and efficiency including system stability.  Yes, I felt that it was very hard for you to take feedbacks even while discussing about this PIP and other other PIP proposed by some user. That's why currently I am not proposing the solution but let's revisit this implementation because **in my opinion this implementation won't work** because **if -ve token stays then it will impact topic's availability** and **even if we put a cap then also refreshing token happens slowly that user will not be able to get consistent rate** instead publiish/dispatch happens in chunk which is not expected behavior. so, please step back, rethink, and let's revisit.

lhotari:
> if that's true, then please rethink about this rate-limiter, token-bucket algorithm cab be exploited by malicious or greedy clients and your implementation to prevent it by introducing -ve token has serious implications of delaying or making topics unavailable without any reason. I implemented dispatch-rate limiter 3-4 years ago and we have been using it without any issue with highest accuracy and efficiency including system stability. Yes, I felt that it was very hard for you to take feedbacks even while discussing about this PIP and other other PIP proposed by some user. That's why currently I am not proposing the solution but let's revisit this implementation because **in my opinion this implementation won't work** because **if -ve token stays then it will impact topic's availability** and **even if we put a cap then also refreshing token happens slowly that user will not be able to get consistent rate** instead publiish/dispatch happens in chunk which is not expected behavior. so, please step back, rethink, and let's revisit.

@rdhabalia Just to clarify: There's no issue in the AsyncTokenBucket core algorithm itself. The remaining problems after #23930 were caused by eventual consistency of getting the token balance. That was a mistake to assume that eventual consistent token balance would be useful. It's not useful for most usecases. I'll be addressing that usability issue with AsyncTokenBucket since token balance should always be consistent externally.

As mentioned in the PR, there is some change since branch-3.0 which causes the current test to fail. The rate for the first second is about 2x of the configured rate. I'll dig more into that.

lhotari:
@rdhabalia One of the key behavior differences has been that by default AsyncTokenBucket adds tokens to the bucket very frequently (every 16 milliseconds) based on the consumed time and the "classic" RateLimiterImpl adds tokens once in a rate period (1 second). In this commit 9c0e4cf9, I made changes to have similar behavior for the AsyncTokenBucket implementation. It makes the behavior very similar to RateLimiterImpl. The rate during the first second results in 2x rate. You can see the behavior if you wish by `gh pr checkout 24012` and running the `org.apache.pulsar.broker.service.DispatchRateLimiterOverconsumingTest` test which now runs both implementations.

rdhabalia:
> I pushed changes to add dispatchRateLimiterFactoryClassName setting for toggling between the "classic" (before PIP-322) implementation and PIP-322 implementation 

Seems like a usecase of Pluggable rate-limiter interface. please find below discussion where someone again refused to take the feedback.
https://lists.apache.org/thread/lz57m8smqolc34h6r5nymd8ppxgv26p7

we have made interface for taking time `System.nanoTime()`  by having interface of `LongSupplier clockSource()` but refused to add an interface for rate-limiter implementation and impacting business critical usecases. 
anyways, let's go ahead and revert back the behavior of rate-limiting asap so it can unblock many mission critical systems that depend on Pulsar.

lhotari:
> > I pushed changes to add dispatchRateLimiterFactoryClassName setting for toggling between the "classic" (before PIP-322) implementation and PIP-322 implementation
> 
> Seems like a usecase of Pluggable rate-limiter interface. please find below discussion where someone again refused to take the feedback. https://lists.apache.org/thread/lz57m8smqolc34h6r5nymd8ppxgv26p7

That's exactly why I added this feature in the PR - to give users the flexibility to continue using the "classic" implementation if needed. I appreciate you pointing this out. "Strong opinions loosely held," for the betterment of Apache Pulsar.

Repeating once again: I've found that there's no fundamental issue with the AsyncTokenBucket implementation itself. The regression occurred due to specific reasons that have been addressed in this PR. Once merged, those problems will be resolved.

> we have made interface for taking time `System.nanoTime()` by having interface of `LongSupplier clockSource()` 

The reason why MonotonicClock is now pluggable is that for some platforms, there could be a need for the "LeapTolerantMonotonicClock" implementation which handles issues where System.nanoTime implementation is slow or not strictly monotonic. These platforms do exist. Regarding the slowness, a good example is MacOS, although not a very practical deployment platform for Pulsar. This `System.nanoTime()` slowness shows up in JMH benchmarks on MacOS.
For platforms where System.nanoTime works well, there's no need for the complexity that "LeapTolerantMonotonicClock" contains. Previously I made incorrect assumptions of the root causes of negative tokens and now that the root cause is clear and fixed in this PR, there's no need to make "LeapTolerantMonotonicClock" the default in Pulsar.

> but refused to add an interface for rate-limiter implementation and impacting business critical usecases. anyways, let's go ahead and revert back the behavior of rate-limiting asap so it can unblock many mission critical systems that depend on Pulsar.

@rdhabalia Please review this PR so that we can revert back the behavior of dispatch rate limiting and get 3.3.5 and 4.0.3 released. These releases are urgent since [many users are waiting for a Pulsar release](https://lists.apache.org/thread/pdw6br1bnfp5xydnpwo8n5wm4dm1jvrw) without the critical vulnerability, CVE-2025-24970. I have already aborted the release votes for 3.3.5-candidate-1 and 4.0.3-candidate-1.

I'd like to remind everyone that all the valuable reasons for [PIP-322](https://github.com/apache/pulsar/blob/master/pip/pip-322.md) continue to exist. PIP-322 is a foundation for future improvements. These future improvements could be similar to what Confluent's Kora has with "backpressure and auto-tuning" as well as "dynamic quota management," according to the [Kora paper](https://www.vldb.org/pvldb/vol16/p3822-povzner.pdf).

Just to be clear: there's no need to revert PIP-322 completely. Instead, please report any issues you encounter so we can fix them together. I sincerely appreciate the effort you've put in while facing regressions caused by PIP-322 changes.

As mentioned earlier, one of the gaps in our LTS release policy is about how a new LTS release becomes the new stable LTS release "officially." For now, 3.0.x LTS is the stable LTS release until we have a proven track record of 4.0.x LTS being sufficiently stable. This will only happen with more community involvement in two critical areas:

1. Contributors willing to report issues with sufficient information
2. Contributors who put sufficient regression tests in place to prevent future regressions

Besides regression tests, we also need more people testing new release candidates. Currently, too few community members are doing this for Apache Pulsar.

lhotari:
The test case in 088ca04 reproduced the issue #29320 and demonstrates that it is now fixed with the changes in this PR.

lhotari:
Since the tests prove that problems in AsyncTokenBucket are fixed, after all, I'll make the AsyncTokenBucket implementation the default. Users that would like to switch to the previous implementation can configure it with `dispatchRateLimiterFactoryClassName=org.apache.pulsar.broker.service.persistent.DispatchRateLimiterFactoryClassic`.

rdhabalia:
Please don't mix two things: bug-fix and new configuration. can you please create a separate PR for asyncToken algo fix and have a new PR to add previous implementation back. also, you might have to create for 2nd PR as it will add new configuration.

lhotari:
> Please don't mix two things: bug-fix and new configuration. can you please create a separate PR for asyncToken algo fix and have a new PR to add previous implementation back. also, you might have to create for 2nd PR as it will add new configuration.

@rdhabalia I understand your concern about separating bug fixes from configuration changes, which is generally good practice. However, given our time constraints with two urgent releases blocked by this PR, I'd really appreciate if we could review it as is.

The AsyncTokenBucket class is relatively small (303 lines including the license header), and the actual fix for the consistency issue is quite targeted - just changing the return value to `tokens - pendingConsumedTokens.sum()` in the else branch of the `consumeTokensAndMaybeUpdateTokensBalance` method.

During #23930, I made some changes based on incorrect assumptions that I've now removed in this PR. While I understand that splitting this into multiple PRs might seem cleaner, it could actually create more confusion about review order and cause further delays.

If you feel strongly about separating parts like the configurability of the dispatch rate implementation or bringing back the pre-PIP-322 implementation, I'm open to discussing it - I just want to make sure we can move forward efficiently given our release timeline.

rdhabalia:
@lhotari they are two different chanegs. 1. fixing a bug and 2. add new implementation.
We must have two separate PRs without any argument.
**you still want to discuss about this simple thing?** 

**oh..I must admit, even getting simple things out with you is quite challenging.* 

> it could actually create more confusion about review order and cause further delays.

and don't worry about delay because we did rush to merge this rate-limiting and made Pulsar unstable. 

there is no argument: we can't mix two things. 

 @codelipenghui @eolivelli   is it unreasonable to ask two separate PRs for two different things?

conf/broker.conf
```diff
@@ -423,6 +423,11 @@ subscribeThrottlingRatePerConsumer=0
 # Rate period for {subscribeThrottlingRatePerConsumer}. Default is 30s.
 subscribeRatePeriodPerConsumerInSecond=30
 
+# The class name of the factory that creates DispatchRateLimiter implementations. Current options are
```
merlimat, line 426: It might be good to summarize here, in a few words, what are the advantages of `DispatchRateLimiterFactoryAsyncTokenBucket`

merlimat (approved):
👍 Nice work and thanks for adding new unit test to cover the existing behaviors

merlimat:
>  to ask two separate PRs for two different things?

@rdhabalia 
Why are these 2 different things? You asked to have a config switch to retain the old buggy behavior. It was added. I don't understand what it is the problem here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/qos/DefaultMonotonicClock.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * <p>
+ * This delegates directly to System.nanoTime() and relies on the underlying system clock for monotonic timekeeping.
+ * </p>
+ */
+public class DefaultMonotonicClock implements MonotonicClock {
+    public static class Factory implements MonotonicClockFactory {
+        @Override
+        public MonotonicClock createMonotonicClock() {
+            return new DefaultMonotonicClock();
```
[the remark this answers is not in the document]
lhotari: Makes sense. I ended up removing the option to use a configurable clock instance to reduce the scope of the PR so this part was removed.

conf/broker.conf
lhotari: Sure. I guess that the main advantages for dispatch rate limiters are internal implementation details and efficiency. Adding tokens doesn't need a separate scheduled job, that is a clear difference in the implementation. The implementation is lockless where contention is minimized by avoiding CAS loops which could be a source of contention when using CAS. During usage, there is no object allocation, only possible stack allocations for primitives. In a busy system, the scheduler execution doesn't add another source of object allocations and GC pressure. All this mainly matters for dispatch rate limiters when a single broker would be holding a very high amount of topics and subscriptions and let's say every topic and subscription would be rate limited. When there are 10000s of instances of a rate limiter, things start to add up. I cannot currently think of a single sentence that I'd put in a configuration file. These were the benefits related to dispatch rate limiters. For publish rate limiters, the benefits are broaders and that is hopefully explained in the [PIP-322 doc](https://github.com/apache/pulsar/blob/master/pip/pip-322.md).

rdhabalia:
@merlimat 
It was again unexpected from you.

First, this PR had a configuration change that required to creation of a PIP. second, It had two separate things a. fix bug in the async-token implementation and second revert old implementation. 

This PR has 52 files changes, it was super difficult to find out which file has a bug fix. I have been spending time to to help in reporting and fixing the issue. You didn't even care to listen and see the feedback because of which @lhotari started again working on it.  You guys just went ahead and merged the PR. I was raising many concerns but you guys are just keep putting patches and doing what you want. It's completely unacceptable.

You guys block PRs from others without any reason, and here I tried to avoid that destructive practice and didn't block it but you just merged it.

BewareMyPower:
> this PR had a configuration change that required to creation of a PIP.

```
# org.apache.pulsar.broker.service.persistent.DispatchRateLimiterFactoryAsyncTokenBucket (default, PIP-322 implementation)
# and org.apache.pulsar.broker.service.persistent.DispatchRateLimiterFactoryClassic (legacy implementation)
```

I see the documents here. But I cannot see what's the difference between them. The difference should be documented in a PIP from a high-level perspective. Or at least, if this is a missed part of PIP-322, we should update https://github.com/apache/pulsar/blob/master/pip/pip-322.md for why do we decide to add such config.

I didn't have a chance to go through all conversations above, but it seems that the reason is that the implementation of PIP-322 was reported to have many bugs. Though it seems to be fixed by this PR, how can we know if this fix brings a new bug? The key point should be that this feature might not be used in production for some time, so this feature should still be considered as "unstable". Then we should provide an option for users to switch back to the legacy rate limiter.

thetumbled:
> @merlimat It was again unexpected from you.
> 
> First, this PR had a configuration change that required to creation of a PIP. second, It had two separate things a. fix bug in the async-token implementation and second revert old implementation.
> 
> This PR has 52 files changes, it was super difficult to find out which file has a bug fix. I have been spending time to to help in reporting and fixing the issue. You didn't even care to listen and see the feedback because of which @lhotari started again working on it. I was raising many concerns but everything was completely ignored and you just went ahead and merged the PR. It's completely unacceptable.
> 
> You block PRs from others without any reason, and here I tried to avoid that destructive practice and didn't block it but you just merged it.

I can't go through all conversations above too, but i agree that we should be cautious about critical and huge change that modify tens of files and thousands of code. This pr is created just few days ago, more review are needed especially when there is still controvesy unsolved. 
We don't need to rush, right? We can't rush to fix one bug but introduce other bugs.

PR #24020, opened 2025-02-25
[improve][client] PIP-409: support producer configuration for retry/dead letter topic producer

### LINKED ISSUE #19463: Support configuring batch message behavior in Java client's DeadLetterPolicy / DLQ feature
### Search before asking

### Motivation

Pulsar's DLQ feature is mainly implemented on the client side. 
It is not currently possible to disable batch messages that might get produced to the dead letter topic when 2 messages happen to get delivered to the dead letter topic about the same time.

For some applications and environments, it could be desired to be able to configure this behavior. The broker configuration doesn't set `acknowledgmentAtBatchIndexLevelEnabled=true` by default which would be necessary for using the [PIP-54 Support acknowledgment at the batch index level](https://pulsar.apache.org/blog/2020/06/18/Apache-Pulsar-2-6-0/#pip-54-support-acknowledgment-at-the-batch-index-level) feature needed for individual message acknowledgement within batchmessages. Re-trying messages within the DLQ topic might fail and without PIP-54, all messages in the batch message would get redelived when even one of the messages in the batch message isn't successfully acknowledged. 

To support retaining messages as non-batch messages, there should be a setting in DeadLetterPolicy to control this.

### Solution

Add a setting to DeadLetterPolicy for controlling batch message behavior. 

Setting `DeadLetterPolicy.BatchBehavior` (name could be adjusted later)
```java
enum BatchBehavior {
    ENABLED,
    DISABLED,
    ENABLE_FOR_BATCHED_MESSAGES
}
```

In addition to disabling/enabling a batch, there could be a third option `ENABLE_FOR_BATCHED_MESSAGES` which would enable batching only when the incoming message is already a batch message.

### Alternatives

_No response_

### Anything else?

_No response_

### LINKED ISSUE #21954: [Bug] Java Client : Pulsar Java client should be able to configure compression for message rejection in the DLQ
### Search before asking

### Version

* Pulsar 3.0.1
* Java Client 3.0.1

### Minimal reproduce step

. Configure a producer with LZ4 compression enabled
. Configure a consumer with DLQ.
. Reject a very big message (negative ack)

Message sent to the DLQ

### What did you see instead?

Execption : o.a.pulsar.client.impl.ConsumerImpl : [persistent://zzz/mailbox/message-event] Failed to send DLQ message to zzz/mailbox/message-event-recorder-DLQ for message id 64563:0:-1
java.util.concurrent.CompletionException: org.apache.pulsar.client.api.PulsarClientException$InvalidMessageException: Message size is bigger than 20971520 bytes

NB: Max Message size and netty are set 20Mb.

### Anything else?

Maybe related to : 
* https://github.com/apache/pulsar/issues/19463
* https://github.com/apache/pulsar/pull/20959

thetumbled:
PIP: #24022 
Fixes #19463
Fixes #21954

### Motivation

Currently, we don't support configure the producer of retry/dead letter topic. But enable the chunk message feature
and disable the batch configuration in hard code, which can't handle many situations. For example, when the throughput 
of message of retry topic become considerable, the resource consumed by the un-batched messages is pretty large. 
There is no reason that we disable the batch message feature.

For better control for the retry/dead letter topic feature, we can support configuration for the producer of 
retry/dead letter topic.

### Modifications

Support configuration for the producer of retry/dead letter topic.

### Verifying this change

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/71


thetumbled:
The pip document has been merged, we can continue to review this pr. Thanks for your review. @BewareMyPower @nodece @dao-jun @codelipenghui @shibd @Technoboy-

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/DeadLetterPolicy.java
```diff
@@ -62,4 +63,16 @@ public class DeadLetterPolicy implements Serializable {
...
+     * Function to build the producer for the retry letter topic.
+     * The input parameter is the topic name.
+     */
+    private Function<String, ProducerBuilder<byte[]>> retryLetterProducerBuilder;
+
+    /**
+     * Function to build the producer for the dead letter topic.
+     * The input parameter is the topic name.
+     */
+    private Function<String, ProducerBuilder<byte[]>> deadLetterProducerBuilder;
```
lhotari, line 77: Instead of using java.util.function.Function, I'd suggest using a richer interface so that more context parameters can be added besides the topic name. 

```suggestion
    /**
     * Customizer for configuring the producer builder for the retry letter topic.
     *
     * <p>This field holds a function that allows the caller to customize the producer builder
     * settings for the retry letter topic before the producer is created. The customization logic
     * can use the provided context (which includes input topic and subscription details) to adjust
     * configurations such as timeouts, batching, or message routing.
     */
    private ProducerBuilderCustomizer retryLetterProducerBuilder;

    /**
     * Customizer for configuring the producer builder for the dead letter topic.
     *
     * <p>This field holds a function that allows the caller to customize the producer builder
     * settings for the dead letter topic before the producer is created. Using the provided context,
     * implementations can perform specific adjustments that ensure the dead letter queue operates
     * with the appropriate configurations tailored for handling undeliverable messages.
     */
    private ProducerBuilderCustomizer deadLetterProducerBuilder;


    /**
     * Functional interface for customizing a producer builder for a specific topic.
     *
     * <p>This interface allows for customizing the producer builder configuration for either the retry letter topic
     * or the dead letter topic. The customization might include setting producer properties such as batching, timeouts,
     * or any other producer-specific configuration.
     *
     * @see ProducerBuilderContext
     */
    public interface ProducerBuilderCustomizer {
        /**
         * Customize the given producer builder with settings specific to the topic context provided.
         *
         * @param context         the context containing information about the input topic and the subscription
         * @param producerBuilder the producer builder instance to be customized
         */
        void customize(ProducerBuilderContext context, ProducerBuilder<byte[]> producerBuilder);
    }

    /**
     * Provides context information required for customizing a producer builder.
     *
     * <p>This interface supplies relevant details such as the name of the input topic and associated subscription name.
     * This contextual information helps in correctly configuring the producer for the appropriate topic.
     */
    public interface ProducerBuilderContext {
        /**
         * Returns the default name of topic for the dead letter or retry letter producer. This topic name is used
         * unless the ProducerBuilderCustomizer overrides it.
         *
         * @return a {@code String} representing the input topic name
         */
        String getDefaultTopicName();
        
        /**
         * Returns the name of the input topic for which the dead letter or retry letter producer is being configured.
         *
         * @return a {@code String} representing the input topic name
         */
        String getInputTopicName();

        /**
         * Returns the name of the subscription for which the dead letter or retry letter producer is being configured.
         *
         * @return a {@code String} representing the subscription name
         */
        String getInputTopicSubscriptionName();
    }
```

lhotari (changes requested):
Instead of using java.util.function.Function, I'd suggest using a richer interface so that more context parameters can be added besides the topic name. This pattern would also allow adding more context parameters later without breaking the API in backwards incompatible ways.

lhotari:
Great work, @thetumbled. Thanks for making this happen. The PIP-409 solution will help resolve long time issues such as #19463 and #21954.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/DeadLetterPolicy.java
thetumbled: Adding `ProducerBuilderContext` is good.
But `ProducerBuilderCustomizer` is unnecessary, and the grammar is incorrect.
`void customize(ProducerBuilderContext context, ProducerBuilder<byte[]> producerBuilder);`
should be changed to 
`ProducerBuilder<byte[]> producerBuilder customize(ProducerBuilderContext context);`

I prefer to use `Function<ProducerBuilderContext, ProducerBuilder<byte[]>>`, which reduce the burden of understanding new interface `ProducerBuilderCustomizer`. WDYT, @lhotari @BewareMyPower @nodece @dao-jun

BewareMyPower: >  so that more context parameters can be added besides the topic name.

It's a good point.

I'm also +1 to `Function<ProducerBuilderContext, ProducerBuilder<byte[]>>`.

nodece: > I'm also +1 to `Function<ProducerBuilderContext, ProducerBuilder<byte[]>>`.

It makes sense to me.

lhotari: > Adding `ProducerBuilderContext` is good.
> But `ProducerBuilderCustomizer` is unnecessary, and the grammar is incorrect.
> `void customize(ProducerBuilderContext context, ProducerBuilder<byte[]> producerBuilder);`
> should be changed to 
> `ProducerBuilder<byte[]> customize(ProducerBuilderContext context);`
> 
> I prefer to use `Function<ProducerBuilderContext, ProducerBuilder<byte[]>>`, which reduce the burden of understanding new interface `ProducerBuilderCustomizer`. WDYT, @lhotari @BewareMyPower @nodece @dao-jun

The reason why both the context and builder are passed are due to the fact that the runtime will preconfigure the builder and the customizer will only override settings.

One possibility would be to also return a builder so that the builder instance could be completely switched.

This would be a way to support different use cases. If you'd replace the builder, you could even produce to a different cluster with a producer builder created from a different pulsar client.

BewareMyPower: > One possibility would be to also return a builder so that the builder instance could be completely switched.

That makes sense to me. Passing a `ProducerBuilder` can keep the immutable fields (`client`) still immutable.

BTW, @thetumbled once the design is updated, please update the PIP document to make it consistent with the latest design

thetumbled: Agree, i just updated the implementation and pip document, PTAL again, thanks for your reviews!

pip/pip-409.md
```diff
@@ -33,25 +33,88 @@ retry/dead letter topic.
...
+}
+
+public interface ProducerBuilderCustomizer {
+    /**
+     * Customize the given producer builder with settings specific to the topic context provided.
+     *
+     * @param context         the context containing information about the input topic and the subscription
+     * @param producerBuilder the producer builder instance to be customized
+     */
+    void customize(ProducerBuilderContext context, ProducerBuilder<byte[]> producerBuilder);
```
nodece, line 69:
```suggestion
    ProducerBuilder<byte[]> customize(ProducerBuilderContext context, ProducerBuilder<byte[]> producerBuilder);
```

Is it better? The user can provide the different producer builder.

BewareMyPower: No. I think passing `ProducerBuilder` as the argument is to prevent from returning a different producer builder.

thetumbled: > No. I think passing ProducerBuilder as the argument is to prevent from returning a different producer builder.

I agree. The purpose of introducing interface `ProducerBuilderCustomizer` is to retrict the creation of `ProducerBuilder`. 
If we allow returning a different producer builder created by user, there is no need for ``ProducerBuilderCustomizer`.

lhotari: > I agree. The purpose of introducing interface `ProducerBuilderCustomizer` is to retrict the creation of `ProducerBuilder`. 

If there's a need to restrict the creation, the return type should be `void` in that case. However, wouldn't it be a valid use case to use a different Pulsar client instance for the producer? In that case allowing to return a different instance should be made possible.


> If we allow returning a different producer builder created by user, there is no need for `ProducerBuilderCustomizer`.
> 

That's not case. A user might want to rely on the defaults for all other setting than what they are overriding in the customizer.

thetumbled: I don't know which kind of senerio need to use a different Pulsar client? Take the example you mentioned above:

>  This would be a way to support different use cases. If you'd replace the builder, you could even produce to a different cluster with a producer builder created from a different pulsar client.

Producing the retry messages to another cluster will break the logical design of retry topic. Let's explain it in more detail. User use `client1` to create consumer for `inputTopic` and `retryTopic`, both of them belonging to `cluster1`. But user use `client2`  belonging to `cluster2` to create the producer for `retryTopic`, resulting the message produced to `cluster2`. However, the consumer for `retryTopic` in `cluster1` can't receive any messages, which break the logic design completely.

BewareMyPower: This interface can be implemented from the client side. Unnecessary flexibility might cause unexpected behaviors that are hard to diagnose. The current interceptor is such a bad example: https://github.com/apache/pulsar/blob/998bb51f5ea256c59110e5b8bef0f9daef1ac48d/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/interceptor/ProducerInterceptor.java#L83

With incorrect implementations, `beforeSend` could return `null` or a customized `Message` implementation. What's worse, if it returns a valid `MessageImpl` with broken `getValue()` implementation, it won't fail fast. Instead, a broken data could be written.

Therefore, I still suggest limiting such tricky use cases.

BewareMyPower: The original motivation of PIP-409 is to support customizing a more efficient config of retry/dead letter topic producer. We should not make it too flexible or complicated for those cases out of that scope.

thetumbled: WDYT @lhotari

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/DeadLetterPolicy.java
```diff
@@ -62,4 +62,29 @@ public class DeadLetterPolicy implements Serializable {
...
+     * settings for the dead letter topic before the producer is created. Using the provided context,
+     * implementations can perform specific adjustments that ensure the dead letter queue operates
+     * with the appropriate configurations tailored for handling undeliverable messages.
+     */
+    private ProducerBuilderCustomizer deadLetterProducerBuilderCustomizer = DEFAULT_PRODUCER_BUILDER_CUSTOMIZER;
+
+    // keep compatibility with old configuration
+    private static final ProducerBuilderCustomizer DEFAULT_PRODUCER_BUILDER_CUSTOMIZER =
+            (context, builder) ->
+                    builder.enableBatching(false).enableChunking(true).blockIfQueueFull(false);
```
lhotari, line 89: Instead of having a default customizer, I'd suggest a solution where in the the runtime, the old configuration is always applied and then a user provider customizer is called if it has been set. If the user would like to change the defaults, they could do that in the customizer.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2294,17 +2295,16 @@ private CompletableFuture<Producer<byte[]>> initDeadLetterProducerIfNeeded() {
...
-                        CompletableFuture<Producer<byte[]>> newProducer =
-                                ((ProducerBuilderImpl<byte[]>) client.newProducer(Schema.AUTO_PRODUCE_BYTES(schema)))
-                                        .initialSubscriptionName(this.deadLetterPolicy.getInitialSubscriptionName())
-                                        .topic(this.deadLetterPolicy.getDeadLetterTopic())
-                                        .producerName(
-                                                String.format("%s-%s-%s-%s-DLQ", this.topicName, this.subscription,
-                                                        this.consumerName, RandomStringUtils.randomAlphanumeric(5)))
-                                        .blockIfQueueFull(false)
-                                        .enableBatching(false)
-                                        .enableChunking(true)
```
lhotari, line 2306: keep these as explained in the previous comment

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2294,17 +2295,16 @@ private CompletableFuture<Producer<byte[]>> initDeadLetterProducerIfNeeded() {
...
-                                        .createAsync();
+                        ProducerBuilderImpl<byte[]> builder = (ProducerBuilderImpl<byte[]>) client
+                                .newProducer(Schema.AUTO_PRODUCE_BYTES(schema))
+                                .topic(deadLetterPolicy.getDeadLetterTopic());
+                        builder.initialSubscriptionName(deadLetterPolicy.getInitialSubscriptionName());
+                        if (deadLetterPolicy.getDeadLetterProducerBuilderCustomizer() != null) {
+                            ProducerBuilderContext context = new ProducerBuilderContextImpl(
+                                    deadLetterPolicy.getDeadLetterTopic(), topic, subscription);
+                            deadLetterPolicy.getDeadLetterProducerBuilderCustomizer().customize(context, builder);
```
lhotari, line 2305: only create the context instance and call the customizer if the customizer is non-null

pip/pip-409.md
lhotari: I'm fine with the customizer approach without allowing other use cases such as replacing the producer builder. That could be added later if needed, without breaking the API.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/DeadLetterPolicy.java
thetumbled: The implementation is updated, PTAL again, thanks.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
lhotari: There was a slight misunderstanding regarding the `DEFAULT_PRODUCER_BUILDER_CUSTOMIZER`. Remove it completely, it's not needed when you revert the change to remove these lines above.

If the user would like to change the above settings, they could do that in the customizer. It also is important for the "customizer" abstraction that unless you do any changes in the customizer, the defaults will be use. The customizer customizes the producer builder after the runtime has configured the defaults. Makes sense?

thetumbled: Maybe you can suggest any code, or it is difficult to understand the subtle difference between current implementation and your expectation.

lhotari: @thetumbled I pushed directly to this PR, f353e02. There was another issue that the ProducerBuilderContextImpl was in the wrong package. This is a class that should be kept private in the same package as ConsumerImpl. It could even be a private inner class in ConsumerImpl, there's no value in reusing it.

lhotari (approved):
LGTM, good work @thetumbled . Great to see this finally solving many use cases.

thetumbled:
> LGTM, good work @thetumbled . Great to see this finally solving many use cases.

Thanks for your review and contributions!

thetumbled:
> One remaining issue about naming. Commented added.

fixed, thanks.

pip/pip-409.md
```diff
@@ -33,25 +33,88 @@ retry/dead letter topic.
 
 ## Design & Implementation Details
 
-- Add two new configurations in `DeadLetterPolicy`:
+- Add `ProducerBuilderContext` and `ProducerBuilderCustomizer` interface:
```
lhotari, line 36: Update the naming also in the PIP document since it's being maintained in this PR.

thetumbled: fixed.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2303,8 +2305,13 @@ private CompletableFuture<Producer<byte[]>> initDeadLetterProducerIfNeeded() {
...
                                         .blockIfQueueFull(false)
                                         .enableBatching(false)
-                                        .enableChunking(true)
-                                        .createAsync();
+                                        .enableChunking(true);
+                        if (deadLetterPolicy.getDeadLetterProducerBuilderCustomizer() != null) {
+                            DeadLetterProducerBuilderContext context = new DeadLetterProducerBuilderContextImpl(
+                                    deadLetterPolicy.getDeadLetterTopic(), topic, subscription, consumerName);
+                            deadLetterPolicy.getDeadLetterProducerBuilderCustomizer().customize(context, builder);
+                        }
```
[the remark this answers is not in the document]
thetumbled: updated.

lhotari (approved):

PR #23888, opened 2025-01-24
[improve][ml]Set default value of managedLedgerPersistIndividualAckAsLongArray to true

poorbarcode:
### Motivation

Regard [[DISCUSS] Add an optional config to disable the feature that compresses cursor metadata, which was contributed by #9292](https://lists.apache.org/thread/kfm7n6xf9pf7h76k4gx83zv3c2kj6yy1), we get a conclusion:
- `branch-3.0, branch-3.3, branch-4.0`: set default value of the feature `managedLedgerPersistIndividualAckAsLongArray` to `false`
- `branch-4.1`: set the default value of the feature `managedLedgerPersistIndividualAckAsLongArray` to `true`

### Modifications
- Set the default value of managedLedgerPersistIndividualAckAsLongArray to `true`. This PR should not be cherry-picked into previous versions.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24027, opened 2025-02-26
[improve][broker] Clean up config changes without a PIP since the 4.0 LTS

BewareMyPower:
- Remove `managedLedgerOffloadReadThreads` from https://github.com/apache/pulsar/pull/24025
- Remove `managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis` and `managedLedgerMaxReadsInFlightPermitsAcquireQueueSize` from https://github.com/apache/pulsar/pull/23901
- Remove `managedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis` from https://github.com/apache/pulsar/pull/22792 The configs above only increase the complexity and are hard to configure.

Add more comments to `managedLedgerPersistIndividualAckAsLongArray` from https://github.com/apache/pulsar/pull/23759. This config was added to keep the compatibility from 3.x or earlier so it has value to retain. 3.x users must configure it with false when upgrading to 4.0.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1769,7 +1769,7 @@ managedLedgerUnackedRangesOpenCacheSetEnabled=true
 # Whether persist cursor ack stats as long arrays, which will compress the data and reduce GC rate
 # When it's false, the behavior will be the same with 3.x or earlier
 # Modifying this config could lose existing individual acknowledgments, so you should configure it with false when
-# upgrading from 3.x to 4.0 or later if you don't want to lose these acknowledgments.
+# upgrading from version <=4.0 to 4.1 or later if you don't want to lose these acknowledgments.
 managedLedgerPersistIndividualAckAsLongArray=true
 
 # For Amazon S3 ledger offload, AWS region
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1755,9 +1755,6 @@ managedLedgerOffloadDriver=
 # Maximum number of thread pool threads for ledger offloading
 managedLedgerOffloadMaxThreads=2
 
-# Maximum number of read thread pool threads for ledger offloading
-managedLedgerOffloadReadThreads=2
-
 # The extraction directory of the nar package.
 # Available for Protocol Handler, Additional Servlets, Entry Filter, Offloaders, Broker Interceptor.
 # Default is System.getProperty("java.io.tmpdir").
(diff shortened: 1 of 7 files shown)


lhotari (changes requested):
For some of the added configuration you are right that PIPs are needed.

Most configurables are settings for internal implementation details that don't need to be changed. In many cases, it will be useful if the value can be modified if a user faces an issue that could be solved by tuning the limit.

For example, #23901 introduced `managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis` and `managedLedgerMaxReadsInFlightPermitsAcquireQueueSize`. The original feature contained a gap where the solution would never work properly when the system was under heavy load. That's the whole point of the `managedLedgerMaxReadsInFlightSizeInMB` solution that it adds a limit. If the limit doesn't work, that's a bug. It's not a new feature when that problem is fixed.

What is the problem that these new configs have caused to our users?
We can continue this discussion on the mailing list.

BewareMyPower:
> That's the whole point of the managedLedgerMaxReadsInFlightSizeInMB solution that it adds a limit. 

Just a simple question. When should users modify this config?

conf/broker.conf
```diff
@@ -1769,6 +1766,12 @@ managedLedgerOffloadPrefetchRounds=1
 # Use Open Range-Set to cache unacked messages
 managedLedgerUnackedRangesOpenCacheSetEnabled=true
 
+# Whether persist cursor ack stats as long arrays, which will compress the data and reduce GC rate
+# When it's false, the behavior will be the same with 3.x or earlier
+# Modifying this config could lose existing individual acknowledgments, so you should configure it with false when
+# upgrading from 3.x to 4.0 or later if you don't want to lose these acknowledgments.
+managedLedgerPersistIndividualAckAsLongArray=true
```
poorbarcode, line 1773: There is a conclusion that we will never set the default value of `managedLedgerPersistIndividualAckAsLongArray ` to `true` for `branch-4.0.x`
- see details here : https://github.com/apache/pulsar/pull/23888#issue-2808801657

poorbarcode (changes requested):

conf/broker.conf
BewareMyPower: Updated

PR #24035, opened 2025-02-28
[fix][broker] fix delay queue sequence issue.

thetumbled:
### Motivation

When a group of delay messages reach to theire timestamp, we expect that the dispatch sequence in accordance with the message id.
Yet there is corner case that will break this rule, we can reproduce the problem with the unit test `testDelaySequence` in this pr.
The root reason is that the inner map of `delayedMessageMap` is `Long2ObjectMap<Roaring64Bitmap>` instead of `Long2ObjectSortedMap<Roaring64Bitmap>`.

### Modifications

Change the type of `delayedMessageMap` to `Long2ObjectSortedMap<Long2ObjectSortedMap<Roaring64Bitmap>>`.

### Verifying this change

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/72

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -22,7 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import io.netty.util.Timer;
 import it.unimi.dsi.fastutil.longs.Long2ObjectAVLTreeMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
-import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectRBTreeMap;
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM

lhotari:
@thetumbled Is it #23611 where the regression was introduced? Does the previous implementation have a similar issue?

dao-jun (approved):

thetumbled:
> @thetumbled Is it #23611 where the regression was introduced? Does the previous implementation have a similar issue?

The previous implementation has sequence issue too, but triggered by other way. Previouse implementation sort the triple tuple (timestamp, ledgerid, entryid) with heap sort algorithm, which is not a stable sort method. 
We may meet issue like this: 
```
tracker.addMessage(0, 0, 1)
tracker.addMessage(1, 1,  1)
tracker.addMessage(2, 2, 1)
```
These three messages are scheduled to be delivered at the same time, but the dispatch sequence may not be 0, 1, 2 due to the sort algorithm.

PR #24055, opened 2025-03-05
[fix] Avoid negative estimated entry count

gaoran10:
### Motivation

The estimated entry count may be a negative value in some cases, then the read operation will get an empty result.

### Modifications

Add a negative check after the long value converts to an int value.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3814,6 +3814,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         return Math.min(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger), maxEntries);
     }
 
+    // The minimum value is 1
     static int estimateEntryCountBySize(long bytesSize, Position readPosition, ManagedLedgerImpl ml) {
         Position posToRead = readPosition;
         if (!ml.isValidPosition(readPosition)) {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3811,11 +3811,14 @@ public class ManagedCursorImpl implements ManagedCursor {
         if (maxSizeBytes == NO_MAX_SIZE_LIMIT) {
             return maxEntries;
         }
-        return Math.min(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger), maxEntries);
+        long estimatedEntryCount = estimateEntryCountBySize(maxSizeBytes, readPosition, ledger);
+        if (estimatedEntryCount > Integer.MAX_VALUE) {
+            return maxEntries;
+        }
+        return Math.min((int) estimatedEntryCount, maxEntries);
     }
 
-    // The minimum value is 1
-    static int estimateEntryCountBySize(long bytesSize, Position readPosition, ManagedLedgerImpl ml) {
+    static long estimateEntryCountBySize(long bytesSize, Position readPosition, ManagedLedgerImpl ml) {
         Position posToRead = readPosition;
         if (!ml.isValidPosition(readPosition)) {
             posToRead = ml.getNextValidPosition(readPosition);
@@ -3854,8 +3857,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                 posToRead = ml.getNextValidPosition(PositionFactory.create(posToRead.getLedgerId(), Long.MAX_VALUE));
             }
         }
-        int safeInt = result > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) result;
-        return Math.max(safeInt, 1);
+        return Math.max(result, 1);
     }
 
     @Override
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3854,8 +3854,8 @@ public class ManagedCursorImpl implements ManagedCursor {
                 posToRead = ml.getNextValidPosition(PositionFactory.create(posToRead.getLedgerId(), Long.MAX_VALUE));
             }
         }
-        int resultInt = Long.valueOf(result).intValue();
-        return Math.max(resultInt < 0 ? Integer.MAX_VALUE : resultInt, 1);
+        int safeInt = result > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) result;
+        return Math.max(safeInt, 1);
     }
 
     @Override
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3811,12 +3811,10 @@ public class ManagedCursorImpl implements ManagedCursor {
         if (maxSizeBytes == NO_MAX_SIZE_LIMIT) {
             return maxEntries;
         }
-        int maxEntriesBasedOnSize =
-                Long.valueOf(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger)).intValue();
-        return Math.min(maxEntriesBasedOnSize, maxEntries);
+        return Math.min(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger), maxEntries);
     }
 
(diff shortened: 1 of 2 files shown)


poorbarcode (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3855,7 +3854,8 @@ static long estimateEntryCountBySize(long bytesSize, Position readPosition, Mana
                 posToRead = ml.getNextValidPosition(PositionFactory.create(posToRead.getLedgerId(), Long.MAX_VALUE));
             }
         }
-        return Math.max(result, 1);
+        int resultInt = Long.valueOf(result).intValue();
+        return Math.max(resultInt < 0 ? Integer.MAX_VALUE : resultInt, 1);
```
lhotari, line 3858: A better solution would be to check the long value before converting it, since in theory, it could "wrap around" multiple times.
```suggestion
        int resultInt = result > Integer.MAX_VALUE ? Integer.MAX_VALUE : result;
        return Math.max(resultInt, 1);
```

lhotari:
> The estimated entry count may be a negative value in some cases, then the read operation will get an empty result.

@gaoran10 Just wondering how it could result could exceed `Integer.MAX_VALUE` so that the integer value becomes negative in the current logic. Is it the line `result += remainEntriesOfLedger;` in `estimateEntryCountBySize` method that causes it? Could we also improve the existing logic and add some test cases to reproduce and prevent such regressions?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
gaoran10: > A better solution would be to check the long value before converting it, since in theory, it could "wrap around" multiple times.

I changed the method return type from long to int, because we still need to convert it from long to int when using it.

gaoran10: fixed

gaoran10:
> Just wondering how it could the result could exceed Integer.MAX_VALUE so that the integer value becomes negative in the current logic. Is it the line result += remainEntriesOfLedger; in estimateEntryCountBySize method that causes it? Could we also improve the existing logic and add some test cases to reproduce and prevent such regressions?

You can try to use this newly added case in the test.
```
var ml = (ManagedLedgerImpl) factory.open(mlName);
ml.addEntry(new byte[1000]);
int entryCount11 = ManagedCursorImpl.estimateEntryCountBySize(
        Long.MAX_VALUE, PositionFactory.create(ml.getCurrentLedger().getId(), 0), ml);
```

BewareMyPower (dismissed):
The root cause should be here: https://github.com/apache/pulsar/blob/8303b96c4d44681f81279c87495b95ee50e32013/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L3815

Calling `intValue` on a `Long` is a dangerous operation that could have overflow issue easily

```java
        int maxEntriesBasedOnSize =
                Long.valueOf(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger)).intValue();
```

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -5206,44 +5206,52 @@ public void testEstimateEntryCountBySize() throws Exception {
...
         assertEquals(entryCount10, 304);
 
         // cleanup.
         ml.delete();
+
+        // test estimated long value convert to an int value
+        ml = (ManagedLedgerImpl) factory.open(mlName);
+        ml.addEntry(new byte[1000]);
+        int entryCount11 = ManagedCursorImpl.estimateEntryCountBySize(
+                Long.MAX_VALUE, PositionFactory.create(ml.getCurrentLedger().getId(), 0), ml);
```
BewareMyPower, line 5252: From my perspective, it seems to be an issue with https://github.com/apache/pulsar/pull/23931. In your test case, the managed ledger has only 1 ledger and the internal states are:

```yaml
- currentLedgerSize: 1000
  currentLedgerEntries: 1
```

However, the `estimateEntryCountBySize` method call returns 8668582741404864, which is very unexpected.

It happens when `Long.MAX_VALUE` is passed to `estimateEntryCountBySize`, which means `applyMaxSizeCap` accepts `Long.MAX_VALUE`. In which case could it happen?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3811,12 +3811,11 @@ public int applyMaxSizeCap(int maxEntries, long maxSizeBytes) {
         if (maxSizeBytes == NO_MAX_SIZE_LIMIT) {
             return maxEntries;
         }
-        int maxEntriesBasedOnSize =
-                Long.valueOf(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger)).intValue();
-        return Math.min(maxEntriesBasedOnSize, maxEntries);
+        return Math.min(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger), maxEntries);
```
BewareMyPower, line 3814: I just checked the downstream side, which could call `asyncReadEntries(maxEntries, Long.MAX_VALUE, ...)` to only limit by `maxEntries`.

A more proper fix should be:

```java
        final var estimateEntryCount = estimateEntryCountBySize(maxSizeBytes, readPosition, ledger);
        if (estimateEntryCount >= Integer.MAX_VALUE) {
            return maxEntries;
        }
        return Math.min((int) estimateEntryCount, maxEntries);
```

And we should add tests to `applyMaxSizeCap` rather than the internal `estimateEntryCountBySize` method.

This PR changes the semantic of `estimateEntryCountBySize` that when the average size is greater than `Integer.MAX_VALUE`, it will return `Integer.MAX_VALUE`, which is anti-intuitive.

BewareMyPower: Generally, we should take care of the type cast. It's intuitive to return a long for the result of `longVariable / intVariable`. But when comparing a long and an integer, we should not simply call `Long#intValue()` method.

BewareMyPower: It's not a good practice to test an internal method but not a public method. Public methods can be called directly.

It's okay for us to remove the `estimateEntryCountBySize` and the related test in future, which could bring the regression again because we don't have tests for public methods like `applyMaxSizeCap` and `asyncReadEntries`.

BewareMyPower (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3811,9 +3811,11 @@ public int applyMaxSizeCap(int maxEntries, long maxSizeBytes) {
         if (maxSizeBytes == NO_MAX_SIZE_LIMIT) {
             return maxEntries;
         }
-        int maxEntriesBasedOnSize =
-                Long.valueOf(estimateEntryCountBySize(maxSizeBytes, readPosition, ledger)).intValue();
-        return Math.min(maxEntriesBasedOnSize, maxEntries);
+        long estimatedEntryCount = estimateEntryCountBySize(maxSizeBytes, readPosition, ledger);
+        if (estimatedEntryCount > Integer.MAX_VALUE) {
```
BewareMyPower, line 3815: I have a question that could `estimateEntryCountBySize` return a negative value? If so, should we check `estimatedEntryCount < 0` as well?

BewareMyPower: This is a value estimated from `maxSizeBytes`, if it's invalid (either negative or a value bigger than a 32-bits integer), we should treat it as unlimited.

gaoran10: The estimateEntryCountBySize method returns this `return Math.max(result, 1);`, the value shouldn't be less than 1.

BewareMyPower:
Add the same `release/xxx` labels since it's fixing the regression of https://github.com/apache/pulsar/pull/23931

lhotari:
There was an additional regression caused by #23931 changes. I've created a PR #24089 to address it. Please review

Issue #24038, opened 2025-02-28
[Bug] PR Checks are failing on branch-3.0


mukesh154:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

branch-3.0

### Minimal reproduce step

Check this link to get the failures: [PR Check Failues on branch-3.0](https://github.com/apache/pulsar/actions/runs/13520588520)

### What did you expect to see?

The Pulsar CI PR checks should pass.

### What did you see instead?

The Pulsar CI PR checks are failing for below test group:

- [CI - Unit - Brokers - Broker Group 4](https://github.com/apache/pulsar/actions/runs/13520588520/job/37779201256#logs)
```
  Error:  Tests run: 20, Failures: 4, Errors: 0, Skipped: 4, Time elapsed: 727.575 s <<< FAILURE! - in org.apache.pulsar.broker.service.ClusterMigrationTest
  Error:  org.apache.pulsar.broker.service.ClusterMigrationTest.testClusterMigration[true, Shared](4)  Time elapsed: 32.444 s  <<< FAILURE!
  java.lang.NullPointerException: Cannot invoke "org.apache.pulsar.broker.service.Subscription.getConsumers()" because the return value of "org.apache.pulsar.broker.service.AbstractTopic.getSubscription(String)" is null
  	at org.apache.pulsar.broker.service.ClusterMigrationTest.testClusterMigration(ClusterMigrationTest.java:358)
  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
  	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
  	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
  	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
  	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
[...]

PR #24058, opened 2025-03-05
[fix][ml] Avoid NPE when getCurrentLedger() returns null

lhotari:
### Motivation

In branch-3.0, tests are failing with logs about NPEs such as 
```
  java.lang.NullPointerException: Cannot invoke "org.apache.bookkeeper.client.LedgerHandle.getId()" because the return value of "org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.getCurrentLedger()" is null
  	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.estimateEntryCountBySize(ManagedCursorImpl.java:3705)
  	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.applyMaxSizeCap(ManagedCursorImpl.java:3691)
  	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.asyncReadEntriesWithSkip(ManagedCursorImpl.java:853)
  	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.asyncReadEntries(ManagedCursorImpl.java:840)
```

This is due to #23931 changes.

### Modifications

Handle the case when getCurrentLedger() returns null

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3826,7 +3826,8 @@ public class ManagedCursorImpl implements ManagedCursor {
 
         while (remainingBytesSize > 0) {
             // Last ledger.
-            if (posToRead.getLedgerId() == ml.getCurrentLedger().getId()) {
+            LedgerHandle currentLedger = ml.getCurrentLedger();
+            if (currentLedger != null && posToRead.getLedgerId() == currentLedger.getId()) {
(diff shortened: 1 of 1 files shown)


poorbarcode (commented):
I am considering why `managedLedger.currentLedger` will be `null`, could you explain the test's name?

lhotari:
@poorbarcode In branch-3.0, there are at least 2 tests where this can be seen in the output. 
org.apache.pulsar.broker.service.ClusterMigrationTest#testClusterMigration is one. The issue is #24038

lhotari:
It seems that `currentLedger` could be null immediately after the ledger has been created, but not after that.

PR #24060, opened 2025-03-05
[improve][monitor] Add version=0.0.4 to /metrics content type for Prometheus 3.x compatibility

lhotari:
### Motivation

Prometheus 3.x requires explicit version information in the content type header when scraping metrics. Without this change, Prometheus 3.x might fail scraping metrics from Pulsar endpoints in certain cases where scraping succeeded with Prometheus 2.x. This would cause monitoring issues for users who upgrade to Prometheus 3.x.
See https://prometheus.io/docs/prometheus/latest/migration/#scrape-protocols for more details.

A similar change has been made in BookKeeper with https://github.com/apache/bookkeeper/pull/4208

### Modifications

- Added a constant `PROMETHEUS_CONTENT_TYPE_004` in `PrometheusMetricsServlet` class
- Updated the content type in all metrics endpoints to include `version=0.0.4`:
  - `PrometheusMetricsServlet`
  - `PulsarPrometheusMetricsServlet`
  - `FunctionsMetricsResource`
- Added a test in `BrokerServiceTest` to verify the correct content type is returned

### Additional context

- Until this change is available, it's possible to configure `fallbackScrapeProtocol` in Prometheus at multiple levels. Here's an example for Apache Pulsar Helm chart https://github.com/apache/pulsar-helm-chart/pull/577 .

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #13785, opened 2022-01-17
Bump prometheus client version from 0.5.0 to 0.15.0

Shoothzj:
### Motivation
- prometheus `simpleclient` has broken api, it makes `pulsar-client-origin` can not well used with other libraries
- specify `OpenMetrics`

This PR contains metric name broken changes.

### Modifications
Bump prometheus client version from 0.5.0 to 0.15.0

### changes
- bump the client version
- rename the field prometheus client `UNTYPED` to `UNKNOWN`
- `OpenMetrics`'s counter name needs a `_total` suffix

### metrics name broken changes
- rename `pulsar_expired_token_count` to `pulsar_expired_token_total`
- rename `pulsar_authentication_success_count` to `pulsar_authentication_success_total`
- rename `pulsar_authentication_failures_count` to `pulsar_authentication_failures_total`
- rename `pulsar_sink_sink_exceptions_total_1min` to `pulsar_sink_sink_exceptions_1min_total`
- rename `pulsar_source_received_total_1min` to `pulsar_source_received_1min_total`
- rename `pulsar_source_written_total_1min` to `pulsar_source_written_1min_total`
- rename `pulsar_source_source_total_1min` to `pulsar_source_source_exceptions_1min_total `
- rename `pulsar_source_system_exceptions_total_1min` to `pulsar_source_system_exceptions_1min_total `
- rename `pulsar_function_received_total_1min` to `pulsar_function_received_1min_total `
- rename `pulsar_function_user_exceptions_total_1min` to `pulsar_function_user_exceptions_1min_total`
- rename `pulsar_function_system_exceptions_total_1min` to `pulsar_function_system_exceptions_1min_total`
- rename `pulsar_function_processed_successfully_total_1min` to `pulsar_function_processed_successfully_1min_total`
- rename `pulsar_function_received_total_1min` to `pulsar_function_received_1min_total`
- rename `pulsar_function_user_exceptions_total_1min` to `pulsar_function_user_exceptions_1min_total`
- rename `pulsar_function_system_exceptions_total_1min` to `pulsar_function_system_exceptions_1min_total`
- rename `pulsar_function_processed_successfully_total_1min` to `pulsar_function_processed_successfully_1min_total`
- rename `pulsar_sink_received_total_1min` to `pulsar_sink_received_1min_total`
- rename `pulsar_sink_written_total_1min` to `pulsar_sink_written_1min_total`
[...]

PR #24065, opened 2025-03-07
[fix][monitor] Set correct content-type for metrics endpoints in Pulsar for Prometheus 3.x compatibility

lhotari:
### Motivation

Follow up on #24060 which set an incorrect content type.

Metrics endpoint uses version 1.0.0 of the Prometheus/OpenMetrics text format.

This content-type value ensures compatibility with Prometheus 3.x and later versions.
For details, refer to the Prometheus 3.x migration guide:
https://prometheus.io/docs/prometheus/latest/migration/#scrape-protocols

The difference between version 1.0.0 and 0.0.4 formats is mainly about counters.
The prometheus client_java library >=0.10.0 creates OpenMetrics compatible counters which are not compatible with the 0.0.4 format which is currently in use in BookKeeper.

For implementation details of Prometheus client_java, see:
https://github.com/prometheus/client_java/blob/parent0.16.0/simpleclient/src/main/java/io/prometheus/client/Counter.java#L76-L80
The library will always append "_total" to the counter name unless the counter name already contains "_total" suffix and will have a separate "_created" counter to ensure OpenMetrics compatibility. This change was introduced in https://github.com/prometheus/client_java/pull/615 in version 0.10.0.
Release notes: https://github.com/prometheus/client_java/releases/tag/parent-0.10.0

In Pulsar, the Prometheus client_java library was updated to 0.15.0 with https://github.com/apache/pulsar/pull/13785 in Pulsar 2.11.0.

### Modifications

- set version to 1.0.0 in the content type so that scraping will result in correct metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/stats/prometheus/PrometheusMetricsServlet.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/stats/prometheus/PrometheusMetricsServlet.java
@@ -40,11 +40,28 @@ import org.slf4j.LoggerFactory;
 
 public class PrometheusMetricsServlet extends HttpServlet {
     public static final String DEFAULT_METRICS_PATH = "/metrics";
-    // Prometheus uses version 0.0.4 of the text format for metrics.
-    // This content-type value ensures compatibility with Prometheus 3.x and later versions.
-    // For details, refer to the Prometheus 3.x migration guide:
-    // https://prometheus.io/docs/prometheus/latest/migration/#scrape-protocols
-    public static final String PROMETHEUS_CONTENT_TYPE_004 = "text/plain; version=0.0.4; charset=utf-8";
+    /**
+     * Metrics endpoint uses version 1.0.0 of the Prometheus/OpenMetrics text format.
+     * <p>
+     * This content-type value ensures compatibility with Prometheus 3.x and later versions.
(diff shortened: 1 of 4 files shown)


lhotari:
The content type might have to be configurable. I'm still unsure whether 1.0.0 is correct or do we need to switch to use `application/openmetrics-text; version=1.0.0; charset=utf-8` as specified in https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure

eolivelli (approved):

eolivelli:
+1 to make it configurable, but to allowing tweaks. Let's not change the way we output metrics depending on the overridden value

lhotari:
This turned to be the wrong solution. #24060 is correct. It was a bug in Prometheus which caused the issues. After spending quite a mount of effort I wasn't able to track down the problem with Prometheus 3.2.1. I switched to VictoriaMetrics in the Pulsar Helm chart to address the problem. 

[OpenMetrics format](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure) requires "EOF" as the last line:
> Expositions MUST end with EOF and SHOULD end with 'EOF\n'

Pulsar's format isn't OpenMetrics, but closer to the 0.0.4 format which isn't so strictly defined.

PR #24072, opened 2025-03-12
[fix][broker] Pattern subscription doesn't work when the pattern excludes the topic domain.

shibd:
### Motivation

After #16062, we use a notification machine to discover topic changes for pattern subscriptions.

If the user sets a pattern string like `my-property/my-ns/test-pattern.*`(without topic domain) when creating a subscription.

It doesn't receive updates for the latest topics because the topic names include the domain when matching.

For example:
- topicName: `persistent://my-property/my-ns/test-pattern-1`
- topicsPattern: `my-property/my-ns/test-pattern.*`

https://github.com/apache/pulsar/blob/fad925f7c522a30ba9fa005e5788162dd17b21ab/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java#L73

You can run test to reproduce it: `testSubscribePatterWithOutTopicDomain`

BTW: This issue has always existed, but some clients had a fallback by periodically querying topics for updates.
For the Java client, this has been removed after #20779, which is why the issue occurs.

### Modifications
- Remove the domain from both the `topic name` and the `topic pattern` for matching.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
@@ -57,7 +57,7 @@ public class TopicListService {
                                 Pattern topicsPattern, List<String> topics) {
             this.topicListService = topicListService;
             this.id = id;
-            this.topicsPattern = Pattern.compile(TopicList.removeTopicDomainScheme(topicsPattern.toString()));
+            this.topicsPattern = topicsPattern;
             this.matchingTopics = TopicList.filterTopics(topics, topicsPattern);
         }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
@@ -57,7 +57,7 @@ public class TopicListService {
                                 Pattern topicsPattern, List<String> topics) {
             this.topicListService = topicListService;
             this.id = id;
-            this.topicsPattern = topicsPattern;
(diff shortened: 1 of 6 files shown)


poorbarcode (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -57,7 +57,7 @@ public TopicListWatcher(TopicListService topicListService, long id,
                                 Pattern topicsPattern, List<String> topics) {
```
lhotari, line 57: It seems that the pattern gets compiled twice. Shouldn't the `removeTopicDomainScheme` happen before `TopicListWatcher` is constructed?

lhotari: looks like org.apache.pulsar.broker.service.ServerCnx#handleCommandWatchTopicList would be the location.

lhotari: https://github.com/apache/pulsar/blob/4ac4f3c91bfdb7307c642fcbd35040015b5a8dd6/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L3156-L3173

lhotari (approved):
LGTM

lhotari:
Good work @shibd

PR #24073, opened 2025-03-12
[improve][broker]Optimize message expiration rate repeated update issues

zjxxzjwang:
### Motivation

The message expiration rate is updated in both tasks, as follows:
Clipboard_Screenshot_1741772793
Clipboard_Screenshot_1741772909

the time period for calculating the rate is the interval between two tasks. If the time interval between two tasks is very small, the calculation speed will be very large.

### Modifications

Delete the update method in the message expiration task, and retain the unified update method in the topic

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
@@ -190,7 +190,6 @@ public class PersistentMessageExpiryMonitor implements FindEntryCallback, Messag
             long numMessagesExpired = (long) ctx - cursor.getNumberOfEntriesInBacklog(false);
             msgExpired.recordMultipleEvents(numMessagesExpired, 0 /* no value stats */);
             totalMsgExpired.add(numMessagesExpired);
-            updateRates();
             // If the subscription is a Key_Shared subscription, we should to trigger message dispatch.
             if (subscription != null && subscription.getType() == SubType.Key_Shared) {
                 subscription.getDispatcher().markDeletePositionMoveForward();


lhotari (changes requested):
> the time period for calculating the rate is the interval between two tasks. If the time interval between two tasks is very small, the calculation speed will be very large.

Good observations, @zjxxzjwang. It seems that this PR is not only an optimization, but targets fixing invalid rate stats.

However, the current change would make "topic stats" inaccurate since the rate wouldn't get updated in that case. For consistency reasons, updating the rate should happen in the same way for all code paths. 
One possible solution would be to make updateRates method private in PersistentMessageExpiryMonitor and call updateRates in the implementation of getMessageExpiryRate method, before the updated rate is returned.

zjxxzjwang:
> > 计算速率的时间段就是两个任务之间的间隔，如果两个任务之间的时间间隔很小，计算速度就会很大。
> 
> 很好的观察，@zjxxzjwang看来这个 PR 不仅仅是一个优化，还针对修复无效率统计。
> 
> 但是，当前的更改会使“主题统计信息”不准确，因为在这种情况下速率不会更新。出于一致性原因，更新速率应以相同的方式在所有代码路径中进行。 一种可能的解决方案是将 PersistentMessageExpiryMonitor 中的 updateRates 方法设为私有，并在返回更新后的速率之前在 getMessageExpiryRate 方法的实现中调用 updateRates。



> > 计算速率的时间段就是两个任务之间的间隔，如果两个任务之间的时间间隔很小，计算速度就会很大。
> 
> 很好的观察，@zjxxzjwang看来这个 PR 不仅仅是一个优化，还针对修复无效率统计。
> 
> 但是，当前的更改会使“主题统计信息”不准确，因为在这种情况下速率不会更新。出于一致性原因，更新速率应以相同的方式在所有代码路径中进行。 一种可能的解决方案是将 PersistentMessageExpiryMonitor 中的 updateRates 方法设为私有，并在返回更新后的速率之前在 getMessageExpiryRate 方法的实现中调用 updateRates。

Hello, the rate will continue to be updated ，because the updateRates method of the PersistentTopic class will also be executed at regular intervals, and the time period used to calculate the rate will be fixed (the timing period of the updateRates method). The calculated rate can be explained more strongly.

lhotari:
> However, the current change would make "topic stats" inaccurate since the rate wouldn't get updated in that case. For consistency reasons, updating the rate should happen in the same way for all code paths.
> One possible solution would be to make updateRates method private in PersistentMessageExpiryMonitor and call updateRates in the implementation of getMessageExpiryRate method, before the updated rate is returned.

@zjxxzjwang Please address this review comment. The solution would be somewhat similar as there is in #24067.

zjxxzjwang:
> > 但是，当前的更改会使“主题统计信息”不准确，因为在这种情况下速率不会更新。出于一致性原因，更新速率应以相同的方式在所有代码路径中进行。
> > 一种可能的解决方案是将 PersistentMessageExpiryMonitor 中的 updateRates 方法设为私有，并在返回更新后的速率之前在 getMessageExpiryRate 方法的实现中调用 updateRates。
> 
> @zjxxzjwang[请回复此评论。解决方案与#24067](https://github.com/apache/pulsar/pull/24067)中的有些相似。

Good suggestion, I have made the modifications according to your suggestion. Please help me review it。 @lhotari

zjxxzjwang:
@lhotari Hello, please help me review it

lhotari (approved):

PR #23837, opened 2025-01-10
[improve][pip] PIP-404: Introduce per ledger properties

dao-jun:
### Motivation

PIP doc

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #24076, opened 2025-03-13
[improve] PIP-404: Introduce per ledger properties

dao-jun:
PIP: https://github.com/apache/pulsar/pull/23837

### Motivation

Introduce per ledger properties

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


liudezhi2098 (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,109 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
...
+            f.completeExceptionally(new ManagedLedgerFencedException());
+            return f;
+        }
+        LedgerInfo li = ledgers.get(ledgerId);
+        if (li == null) {
+            f.completeExceptionally(new ManagedLedgerException("Ledger not found"));
+            return f;
+        }
+
+        synchronized (this) {
```
nodece, line 1373: I think `synchronized (this)` is unnecessary. What are your concerns?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,109 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
...
+            if (!metadataMutex.tryLock()) {
```
nodece, line 1374: You don't need to call `tryLock` here because the `org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl#asyncUpdateProperties` will try lock.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,109 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
         return result;
     }
 
+    @Override
+    public CompletableFuture<Void> asyncAddLedgerProperty(long ledgerId, String key, String value) {
```
nodece, line 1360: The same feedback applies to `asyncRemoveLedgerProperty`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,109 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
...
+        synchronized (this) {
+            if (!metadataMutex.tryLock()) {
+                // retry in 100 milliseconds
+                scheduledExecutor.schedule(
+                        () -> asyncRemoveLedgerProperty(ledgerId, key), 100,
+                        TimeUnit.MILLISECONDS);
+            } else {
+                f.whenComplete((v, t) -> metadataMutex.unlock());
+                List<MLDataFormats.KeyValue> oldProperties = li.getPropertiesList();
+                Map<String, String> newPropertiesMap = new HashMap<>();
```
nodece, line 1413: We can optimize this by adding a fast path: if the properties don't contain the given key, we can skip rebuilding and return early.

nodece (changes requested):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,109 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
...
 
+    @Override
+    public CompletableFuture<Void> asyncAddLedgerProperty(long ledgerId, String key, String value) {
+        CompletableFuture<Void> f = new CompletableFuture<>();
+        // Execute the operation in ML executor to avoid concurrent modifications to ledger info.
+        if (state.isFenced()) {
+            f.completeExceptionally(new ManagedLedgerFencedException());
+            return f;
+        }
+        LedgerInfo li = ledgers.get(ledgerId);
```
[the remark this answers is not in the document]
dao-jun: No need to consider this, because if we want to add a new property into the ledger, the ledger must exists in ledgers map.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
dao-jun: This is to avoid ledger info overwritten. 
Offloader also can update ledger info.
see:
ManagedLedger.tryTransformLedgerInfo

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
dao-jun: same as above

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
dao-jun: same as above

jiazhai (approved):
+1. over all LGTM. This change adds an optional field, it is safe for the existing code. 
A small suggestion: If we will have more than 1 property in per ledger, It would be better to have batched operations (e.g. set/get/removeLedgerProperties), looks like the PerLedgerProperties related operations 
 are a little heavy (operate the data in ZK, need locks...).

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
nodece: LGTM

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
nodece: "metadataMutex is different from synchronized. `synchronized` is reentrant, but metadataMutex is not.
When running the test locally, the thread can get stuck in a deadlock if it tries to acquire metadataMutex recursively."

`metadataMutex` differs from `synchronized` because it is not reentrant.
While `synchronized` allows the same thread to reacquire the lock, `metadataMutex` does not, leading to a deadlock if the thread attempts to lock it again.

This issue causes the test to get stuck in a loop.

nodece (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,116 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
...
+                f.complete(null);
+            }
+
+            @Override
+            public void operationFailed(MetaStoreException e) {
+                log.error("[{}] Update managedLedger's properties failed", name, e);
+                handleBadVersion(e);
+                f.completeExceptionally(e);
+            }
+        });
```
lhotari, line 1466: It seems that the existing `transformLedgerInfo` method should be used instead.

dao-jun: Good suggestion, addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1356,6 +1356,96 @@ public void terminateFailed(ManagedLedgerException exception, Object ctx) {
...
+                    List<MLDataFormats.KeyValue> newProperties = newPropertiesMap.entrySet().stream()
+                            .map(e -> MLDataFormats.KeyValue.newBuilder()
+                                    .setKey(e.getKey()).setValue(e.getValue()).build()).toList();
+                    return oldInfo.toBuilder().clearProperties().addAllProperties(newProperties).build();
+                })
+                .thenAccept(v -> f.complete(null))
+                .exceptionally(t -> {
+                    Throwable cause = FutureUtil.unwrapCompletionException(t);
+                    if (cause instanceof OffloadConflict) {
+                        f.completeExceptionally(new ManagedLedgerNotFoundException("Offload conflict detected"));
```
lhotari, line 1385: "Offload conflict detected" is a useless message. It's better to use the original message in this case:
```suggestion
                        f.completeExceptionally(new ManagedLedgerNotFoundException(cause.getMessage()));
```

lhotari (approved):

PR #24089, opened 2025-03-17
[fix][ml] Fix issues in estimateEntryCountBySize

lhotari:
## Motivation

The changes introduced in PR #23931 caused a regression in the entry count estimation logic for branch-3.0. While PR #24055 attempted to address this regression, it did not fully resolve the issues, leaving Pulsar's branch-3.0 in a broken state. A `NullPointerException` could occur when using `ReadOnlyManagedLedgerImpl` instances, as the `currentLedger` is null in these cases. `ReadOnlyManagedLedgerImpl` is used in branch-3.0 by PulsarSQL. In master branch, it's currently unused, but since the code exists, it makes sense to fix this bug in the master branch. This PR also contains other improvements.

A problem was also detected in using ManagedLedgerImpl.getNextValidPosition method to find the next valid position. This method is relatively costly and takes the current LAC into account which doesn't make much sense just for estimation purposes.

## Modifications
1. **NPE protection**: Added null checks for `currentLedger` to handle `ReadOnlyManagedLedgerImpl` instances properly
2. **Code reorganization**: Extracted the estimation logic into a separate `EntryCountEstimator` class to improve testability and code clarity
3. **Parameter enhancements**: Modified method signature to accept maxEntries and stop calculations when this limit is reached
4. **Return type change**: Changed return type from `long` to `int` to match the actual usage in the codebase
5. **Documentation**: Added clear comments about implementation details
6. **Tests**: Added comprehensive unit tests in `EntryCountEstimatorTest` that validate various scenarios.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3826,9 +3826,19 @@ public class ManagedCursorImpl implements ManagedCursor {
         long result = 0;
         long remainingBytesSize = bytesSize;
 
+        Position previousPosToRead = null;
         while (remainingBytesSize > 0) {
-            // Last ledger.
-            if (posToRead.getLedgerId() == ml.getCurrentLedger().getId()) {
+            if (previousPosToRead != null && previousPosToRead.compareTo(posToRead) >= 0) {
+                // Exit the loop if posToRead is not advancing to avoid counting the same position
+                // multiple times, which could cause an unnecessarily high number of iterations,
+                // especially when bytesSize is set to Long.MAX_VALUE.
+                break;
+            }
+            // Save the current posToRead in a variable to verify if it advances in the next iteration.
+            previousPosToRead = posToRead;
+            // Last ledger. This is null in ReadOnlyManagedLedgerImpl
+            LedgerHandle currentLedger = ml.getCurrentLedger();
+            if (currentLedger != null && posToRead.getLedgerId() == currentLedger.getId()) {
                 if (ml.getCurrentLedgerSize() == 0 ||  ml.getCurrentLedgerEntries() == 0) {
                     // Only read 1 entry if no entries to read.
                     return 1;
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3811,11 +3811,8 @@ public class ManagedCursorImpl implements ManagedCursor {
         if (maxSizeBytes == NO_MAX_SIZE_LIMIT) {
             return maxEntries;
         }
-        long estimatedEntryCount = estimateEntryCountByBytesSize(maxSizeBytes, readPosition, ledger);
-        if (estimatedEntryCount > Integer.MAX_VALUE) {
-            return maxEntries;
-        }
-        return Math.min((int) estimatedEntryCount, maxEntries);
+        int estimatedEntryCount = estimateEntryCountByBytesSize(maxEntries, maxSizeBytes, readPosition, ledger);
+        return Math.min(estimatedEntryCount, maxEntries);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


dao-jun (approved):

lhotari:
There seem to be other gaps in the estimation logic. I'll revisit this in this PR in further commits.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3826,9 +3826,19 @@ static long estimateEntryCountBySize(long bytesSize, Position readPosition, Mana
         long result = 0;
         long remainingBytesSize = bytesSize;
 
+        Position previousPosToRead = null;
         while (remainingBytesSize > 0) {
-            // Last ledger.
-            if (posToRead.getLedgerId() == ml.getCurrentLedger().getId()) {
```
poorbarcode, line 3831: I think the following predicate is better

```java
(ml.getCurrentLedger() != null && posToRead.getLedgerId() == ml.getCurrentLedger().getId())
|| (ml.getCurrentLedger() == null && posToRead.getLedgerId() == ml.ledgers.lastEntry().getKey())
```

BewareMyPower: @poorbarcode It's just a coding style issue and I don't think one is better than the other.  BTW, the change you suggested has a race condition because `getCurrentLedger` is called three times without any synchronization, which means they could return different results.

For your specific case, I'd like a functional coding style to simplify your code and make it thread-safe:

```java
        final Position posToRead;
        if (ml.isValidPosition(readPosition)) {
            posToRead = readPosition;
        } else {
            posToRead = ml.getNextValidPosition(readPosition);
        }
        long result = 0;
        long remainingBytesSize = bytesSize;

        while (remainingBytesSize > 0) {
            final var flag = Optional.ofNullable(ml.getCurrentLedger())
                    .map(currentLedger -> currentLedger.getId() == posToRead.getLedgerId())
                    .orElseGet(() -> posToRead.getLedgerId() == ml.ledgers.lastEntry().getKey());
```

poorbarcode: @BewareMyPower Please review the reason why `currentLedger` is null before arguing, the variable will always be `null` or always `non-null`.

>  It's just a coding style issue and I don't think one is better than the other.

By the way, I often saw that you argued with coding styles

BewareMyPower: Yeah it's right that the `currentLedger` could only be null when it's a readonly cursor. Hence there should be no race condition. But the pre-condition is **implicit**. Generally, adding null validation usually means the field is nullable without extra explanation.

```
            // Last ledger. This is null in ReadOnlyManagedLedgerImpl
```

And this PR already contains the comment above.

poorbarcode: > // Last ledger. This is null in ReadOnlyManagedLedgerImpl
> And this PR already contains the comment above.

- Should we back to the original design of `ReadOnlyManagedLedgerImpl` to confirm whether it is correct?
- And check all places that calls `managedledger.getCurrentLedger()`?

BewareMyPower: It makes sense. We can wait until this PR is ready to review.

lhotari: I have moved the estimation logic to a separate class so that it's easier to unit test the logic without mocking. It makes it easier to make it more understandable when this logic isn't part of the ManagedCursorImpl class. We have too large classes already and splitting into smaller classes would be a desirable direction.
Please review @poorbarcode and @BewareMyPower

BewareMyPower (approved):

lhotari:
follow up in #24125, please review

PR #24091, opened 2025-03-18
[fix][broker] Fix some problems in calculate totalAvailableBookies in method getExcludedBookiesWithIsolationGroups when some bookies belongs to multiple isolation groups.

hanmz:
### Motivation

Currently, the values ​​of `totalAvailableBookiesInPrimaryGroup` and `totalAvailableBookiesFromPrimaryAndSecondary` may not be accurate because there may be repeated accumulation when some bookies belongs to multiple isolation groups.

For example: There are 3 bookies in total and has 3 group, the distribution is as follows
Group1: bk1
Group2: bk2,bk3
Group3: bk1
When `primaryIsolationGroup` contains Group1 and Group3, `secondaryIsolationGroup` contains Group2, when execute the method `getExcludedBookiesWithIsolationGroups(2, groups[primaryIsolationGroup,secondaryIsolationGroup])`. The correct return value should be empty but current return bk2 and bk3.

The calculation of `totalAvailableBookiesFromPrimaryAndSecondary` has similar problems.

### Modifications
Fix the calculation logic of `totalAvailableBookiesInPrimaryGroup` and `totalAvailableBookiesFromPrimaryAndSecondary`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/bookie/rackawareness/IsolatedBookieEnsemblePlacementPolicy.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/bookie/rackawareness/IsolatedBookieEnsemblePlacementPolicy.java
@@ -206,6 +206,9 @@ public class IsolatedBookieEnsemblePlacementPolicy extends RackawareEnsemblePlac
                     return excludedBookies;
                 }
                 Set<String> allGroups = allGroupsBookieMapping.keySet();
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM, good work @hanmz

PR #7266, opened 2020-06-13
[pulsar-broker] Dispatch batch messages according consumer permits

rdhabalia:
### Motivation
Right now, dispatching of batch messages is not predictable for a shared-subscription. Dispatcher doesn't consider number of batch messages in an entry which can cause higher number of message delivery than consumer receiver-queue.
For example:
If a topic has message-entries with 100 batched message in an entry.
If consumer has receiver queue=40 and has available-permits=20. then broker dispatches 20 entries with total 2000 (20*100) messages in it. because of that consumer's will high number of messages than expected and we can see available-permits in negative.

### Modification
- Find out average number of batch messages in an entry
- Dispatch messages considering total batch-messages in an entry

### Result
It will avoid dispatching higher number of messages than consumer's available-permit and will fix -ve available-permits.

PR #24092, opened 2025-03-18
[fix][broker] Restore the behavior to dispatch batch messages according to consumer permits

### LINKED ISSUE #23825: Flaky-test: BatchMessageTest.testBatchMessageDispatchingAccordingToPermits
### Search before asking

### Example failures

- [2025-01-07T10:46:47.4307801Z](https://github.com/apache/pulsar/actions/runs/12635032254/job/35247450509#step:11:1178) 
- [2025-01-03T19:03:01.9270671Z](https://github.com/apache/pulsar/actions/runs/12602408785/job/35125717078#step:11:1222) 

### Exception stacktrace

```
java.lang.AssertionError: expected [10.0] but found [100.0]
	at org.testng.Assert.assertEquals(Assert.java:836)
	at org.testng.Assert.assertEquals(Assert.java:849)
	at org.apache.pulsar.broker.service.BatchMessageTest.testBatchMessageDispatchingAccordingToPermits(BatchMessageTest.java:1016)
```

BewareMyPower:
Fixes #23825

### Motivation

It seems that the PIP-379 breaks https://github.com/apache/pulsar/pull/7266. For example, given the following configs:
- The consumer whose available permits is 10
- There are 10 entries available, each entry's batch size is 10.

When `subscriptionSharedUseClassicPersistentImplementation` is false (by default), 10 entries will be dispatched to the consumer. When it's true (the behavior before 4.0), only 1 entry will be dispatched to the consumer.

### Modifications

Fix the `maxEntriesInThisBatch` calculation in `PersistentDispatcherMultipleConsumers`. In `testBatchMessageDispatchingAccordingToPermits`, wait until the entries are dispatched to consumers. Then this test would fail without this fix.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -850,7 +850,7 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
                 int maxAdditionalUnackedMessages = Math.max(c.getMaxUnackedMessages() - c.getUnackedMessages(), 0);
                 maxMessagesInThisBatch = Math.min(maxMessagesInThisBatch, maxAdditionalUnackedMessages);
             }
-            int maxEntriesInThisBatch = Math.min(availablePermits,
+            int maxEntriesInThisBatch = Math.min(availablePermits / avgBatchSizePerMsg,
                             // use the average batch size per message to calculate the number of entries to
                             // dispatch. round up to the next integer without using floating point arithmetic.
                             (maxMessagesInThisBatch + avgBatchSizePerMsg - 1) / avgBatchSizePerMsg);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -843,14 +843,14 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
                         c, c.getAvailablePermits());
             }
 
-            int maxMessagesInThisBatch =
-                    Math.max(remainingMessages, serviceConfig.getDispatcherMaxRoundRobinBatchSize());
+            int maxMessagesInThisBatch = Math.min(remainingMessages, availablePermits);
             if (c.getMaxUnackedMessages() > 0) {
                 // Calculate the maximum number of additional unacked messages allowed
                 int maxAdditionalUnackedMessages = Math.max(c.getMaxUnackedMessages() - c.getUnackedMessages(), 0);
                 maxMessagesInThisBatch = Math.min(maxMessagesInThisBatch, maxAdditionalUnackedMessages);
             }
-            int maxEntriesInThisBatch = Math.min(availablePermits / avgBatchSizePerMsg,
+            // TODO: add tests to verify dispatcherMaxRoundRobinBatchSize is respected
+            int maxEntriesInThisBatch = Math.min(serviceConfig.getDispatcherMaxRoundRobinBatchSize(),
                             // use the average batch size per message to calculate the number of entries to
                             // dispatch. round up to the next integer without using floating point arithmetic.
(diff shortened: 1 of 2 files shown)


poorbarcode (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -850,7 +850,7 @@ protected synchronized boolean trySendMessagesToConsumers(ReadType readType, Lis
                 int maxAdditionalUnackedMessages = Math.max(c.getMaxUnackedMessages() - c.getUnackedMessages(), 0);
                 maxMessagesInThisBatch = Math.min(maxMessagesInThisBatch, maxAdditionalUnackedMessages);
             }
-            int maxEntriesInThisBatch = Math.min(availablePermits,
+            int maxEntriesInThisBatch = Math.min(availablePermits / avgBatchSizePerMsg,
                             // use the average batch size per message to calculate the number of entries to
                             // dispatch. round up to the next integer without using floating point arithmetic.
                             (maxMessagesInThisBatch + avgBatchSizePerMsg - 1) / avgBatchSizePerMsg);
```
lhotari, line 856: Thanks for spotting this issue. For clarity of the code, it could be better to move the `availablePermits` consideration on line 847 as `Math.max(Math.min(remainingMessages, availablePermits),` and remove the `Math.min` here so that it becomes:
```suggestion
            int maxEntriesInThisBatch = 
                            // use the average batch size per message to calculate the number of entries to
                            // dispatch. round up to the next integer without using floating point arithmetic.
                            (maxMessagesInThisBatch + avgBatchSizePerMsg - 1) / avgBatchSizePerMsg);
```

in the original code, this was how it was handled: https://github.com/apache/pulsar/blob/e547beaa6dad72359bb5b8d30ce9fa80a6989f71/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumersClassic.java#L738-L740

The reason to use different variable names was to improve the clarity of the code, but it introduced the regression which thanks to you is now getting fixed.

In the original code, the problem with clarity is caused by using abbreviated variable names such as `messagesForC` and reusing the same variable for different purposes. Abbreviated variable names and reusing variables is not a great idea from code maintainability perspective.

lhotari (approved):
Thanks for spotting this regression @BewareMyPower. I added one suggestion.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
BewareMyPower: I updated the local variables with correct meanings:
- `maxMessagesInThisBatch = Math.min(remainingMessages, availablePermits)`
- `maxEntriesInThisBatch = Math.min(serviceConfig.getDispatcherMaxRoundRobinBatchSize(), (maxMessagesInThisBatch + avgBatchSizePerMsg - 1) / avgBatchSizePerMsg))`

The previous code compares `dispatcherMaxRoundRobinBatchSize` with `remainingMessages` to get `maxMessagesInThisBatch`, which is incorrect because `dispatcherMaxRoundRobinBatchSize` represents the number of entries.

```java
    @FieldContext(
        dynamic = true,
        category = CATEGORY_SERVER,
        doc = "Max number of entries to dispatch for a shared subscription. By default it is 20 entries."
    )
    private int dispatcherMaxRoundRobinBatchSize = 20;
```

However, I found there is no test to protect that `dispatcherMaxRoundRobinBatchSize` takes effect. It's beyond the scope of this PR, so I just added a TODO here.

gaoran10 (approved):

nodece (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
lhotari: Good stuff, @BewareMyPower

lhotari (approved):
LGTM

PR #24096, opened 2025-03-19
[fix] [broker] Fix negative subscription/consumer's unack-messages

poorbarcode:
### Motivation

**Issue**: negative unack-messages
- Consumer-1 received messages.
- Unload the topic.
- The message may be sent to consumer-2, but the consumption of consumer-1 is still in progress now.
- Consumer-1 and consumer-2 acknowledge the message concurrently.

`unack-messages` may be reduced twice in above scenario, you can reproduce the issue by the new test `testAcknowledgeConcurrently`

### Modifications

- Only reduce `unack-messages` who actually deleted messages successfully.
- This PR also fixes the issue of the `unack-messages` is not accurate

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/AckSetPositionImpl.java
```diff
@@ -31,6 +33,11 @@ public class AckSetPositionImpl implements Position, AckSetState {
     protected final long ledgerId;
     protected final long entryId;
     protected volatile long[] ackSet;
+    @Getter
+    @Setter
+    private int batchMessagesAckedCount;
+    @Getter
+    private boolean positionRemovedFromCursor;
```
lhotari, line 40: mutable fields should be `volatile` unless there's other ways to ensure thread safety

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -560,74 +560,48 @@ public CompletableFuture<Void> messageAcked(CommandAck ack) {
...
-                //check if the position can remove from the consumer pending acks.
-                // the bit set is empty in pending ack handle.
-                if (AckSetStateUtil.hasAckSet(position)) {
-                    if (((PersistentSubscription) subscription)
-                            .checkIsCanDeleteConsumerPendingAck(position)) {
-                        removePendingAcks(ackOwnerConsumer, position);
-                    }
-                }
-            }));
         }
```
lhotari, line 602: How is transactional cases handled with this change? Is it fine to remove this?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -538,6 +556,127 @@ public void deleteFailed(ManagedLedgerException exception, Object ctx) {
...
+             * - topic was unloaded
+             *   - reset messages to consumer-2
+             * At this moment, race-condition occurs:
+             * - consumer-1 is acknowledging msg-A
+             * - dispatcher is delivering msg-A to consumer-2
+             * Issue: broker received the acknowledging of msg-A, but no consumer has pending acknowledge that relate
+             *  to msg-A so broker can not know how many single messages in the batched message.
+             * Solve: to get a precise messages number, this "synchronized" block is needed.
+             */
+            synchronized (dispatcher0) {
```
lhotari, line 575: does adding this synchronized lock cause a risk for deadlocks? In general, there's a risk of deadlocks at least in a case where the nested call acquires other locks.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -538,6 +556,127 @@ public void deleteFailed(ManagedLedgerException exception, Object ctx) {
...
+                // Regarding this case, only cursor know how many messages in batch were acked, because
+                // "consumer.pendingAcks" does not know how many messages were acked count before, only "cursor"
+                // knows.
+                actualAcked = batchMessagesAckedCount;
+            }
+            attemptAckMsgs += actualAcked;
+            // Reduce un-acknowledged messages.
+            owner.addAndGetUnAckedMsgs(owner, -actualAcked);
+            owner.updateBlockedConsumerOnUnackedMsgs(owner);
+            if (log.isInfoEnabled()) {
```
lhotari, line 665: Is info level too high for this log message?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -538,6 +556,127 @@ public void deleteFailed(ManagedLedgerException exception, Object ctx) {
...
+            final long ledgerId = position.getLedgerId();
+            final long entryId = position.getEntryId();
+            final int batchMessagesAckedCount = AckSetStateUtil.getBatchMessagesAckedCount(position);
+            final boolean positionRemovedFromCursor = AckSetStateUtil.isPositionRemovedFromCursor(position);
+            if (batchMessagesAckedCount == 0) {
+                // All messages were skipped.
+                // Since we can not get how many msgs that were attempted to ack, just plus 1 into the
+                // "attemptAckMsgs".
+                attemptAckMsgs++;
+                log.warn("[{}][{}]{}-{}-{} is acknowledging {}:{}, which has been acked before. consumer_size: {}",
```
lhotari, line 595: Is warn level too high? Isn't the actual problem resolved by this PR and there's nothing to do about it?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -538,6 +556,127 @@ public void deleteFailed(ManagedLedgerException exception, Object ctx) {
...
+                        owner = consumer;
+                        break;
+                    }
+                }
+            }
+            if (owner == null) {
+                // Since we can not get how many msgs that were attempted to ack, just plus 1 into the
+                // "attemptAckMsgs".
+                attemptAckMsgs++;
+                log.warn("[{}][{}]{}-{}-{} skipped to reduce un-ack-msgs for {}:{}, because could not find the"
```
lhotari, line 630: Is warn level too high? Isn't the actual problem resolved by this PR and there's nothing to do about it?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -362,10 +362,6 @@ public synchronized void readMoreEntries() {
         Position markDeletePosition = cursor.getMarkDeletedPosition();
         if (lastMarkDeletePositionBeforeReadMoreEntries != markDeletePosition) {
             redeliveryMessages.removeAllUpTo(markDeletePosition.getLedgerId(), markDeletePosition.getEntryId());
-            for (Consumer consumer : consumerList) {
-                consumer.getPendingAcks()
-                        .removeAllUpTo(markDeletePosition.getLedgerId(), markDeletePosition.getEntryId());
-            }
```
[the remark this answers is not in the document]
poorbarcode: > or the mark delete position moves forward as a result of admin operations

I think that rather than always removing consumers' pending ack when reading entries, we'd better delete consumers' pending acks when the md-position is moving.

> Why is this removed in this PR?

`readMoreEntries` runs in a different thread, which may cause a race condition below
- `readMoreEntries` removes consumers' pending acks
- `after delete messages` reduces un-ack messages by consumers' pending acks, but it get a non value because the task `readMoreEntries` already removed them.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
poorbarcode: - There is a race condition which was described by the code comments.
- There are no other locks in the thread, `cursor.lock.writeLock()` has released when execute `delete callback`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
poorbarcode: After the transaction is complete, it will call `cursor.delete`, and the callback of `cursor.delete` will remove consumers' pending acks

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
poorbarcode: I think it is okay, this warning log indicates there are repeatable consumption

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
poorbarcode: I think it is okay, this warning log indicates there are repeatable consumption

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
poorbarcode: Changed to `debug` level.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/AckSetPositionImpl.java
poorbarcode: Fixed

poorbarcode:
Rebased master branch

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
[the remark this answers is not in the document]
lhotari: > > or the mark delete position moves forward as a result of admin operations
> 
> I think that rather than always removing consumers' pending ack when reading entries, we'd better delete consumers' pending acks when the md-position is moving.

Since this logic isn't covered in this PR, I think that removing the existing logic should be omitted from this PR. As mentioned in the previous comment there would be a temporarily memory leak and potential draining hashes being blocked infinitely in Key_Share if this logic would be removed. Most likely we don't have tests to cover this case at the moment, but it doesn't mean that we could just remove the logic in this PR

The mark delete position is updated frequently compared to `readMoreEntries` calls. That's the reason why it's more efficient to handle the consumer side of mark delete position advancing.

> > Why is this removed in this PR?
> 
> `readMoreEntries` runs in a different thread, which may cause a race condition below
> 
> * `readMoreEntries` removes consumers' pending acks
> * `after delete messages` reduces un-ack messages by consumers' pending acks, but it gets a null value because the task `readMoreEntries` already removed them.

Also for the negative unack message count, there would need to be a solution for handling message expiration, where the mark delete position moves forward without all messages being acknowledged.
It's not OK to remove calls to `getPendingAcks().removeAllUpTo` in this PR.

I just wonder what the race conditions are since in `PendingAcksMap`, there are locks to prevent race conditions. That's exactly the reason why the there's a read and write lock so that the acknowledgements happening in a different thread don't cause problems.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
lhotari: Well, for a Pulsar administrator, they could be confused about the warning message, especially if a lot of these warnings end up in log files of a large system and it triggers an alert. Error and warning logs should be understandable for that audience instead of the ones that know Pulsar implementation level details. 
I'd rather say that this should be `info` level instead of `warn` level when there's nothing to do about it.

Google has some guidelines of good error messages that also apply to warnings. When there's an error or warning, it would be useful to explain what to do about it or whether action needs to be taken. 
Some references from Google's guidelines:
https://developers.google.com/tech-writing/error-messages/identify-the-cause
https://developers.google.com/tech-writing/error-messages/show-fix

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
lhotari: I'd suggest changing from warn to info, similar rationale as in https://github.com/apache/pulsar/pull/24096#discussion_r2007094289.

poorbarcode: Improved

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
poorbarcode: Improved

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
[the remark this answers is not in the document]
poorbarcode: After you run `ConsumerAckAndDisableBrokerBatchAckTest.testAcknowledgeConcurrently` more than `200` times, you will get an issue, which was leads by these lines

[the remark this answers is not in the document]
poorbarcode: > How are pending acks removed if this code is being removed? 

I described here: https://github.com/apache/pulsar/pull/24096#discussion_r2006825594

> That's something that you should address in this PR if you prefer to remove this code in this PR.

I feel that pip-379 should solve this problem because these lines of code were added by PIP-379 in order avoid to solving the problems you described at https://github.com/apache/pulsar/pull/24096#discussion_r2003692893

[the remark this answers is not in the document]
poorbarcode: @lhotari 

Maybe we can solve the potential issue of pending acks that you mentioned [here](https://github.com/apache/pulsar/pull/24096#discussion_r2003692893) one by one, but it will be slow because I am busy with other works. Could you explain more information you know, more detail is better. ❤️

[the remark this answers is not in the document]
lhotari: @poorbarcode Simply don't remove this code that is critical fro PIP-379 in this PR. That would be a way to address this review comment.

poorbarcode:
Since there are too many other bugs that lead to the tests can not pass, and the current PR is too large to review, I will seperate the current PR to some little PRs
- [ ] https://github.com/apache/pulsar/pull/24405
- [ ] https://github.com/apache/pulsar/pull/24406
- [ ] There are others later

poorbarcode:
Pushed a new PR to instead the current one
- https://github.com/apache/pulsar/pull/24496

lhotari:
closing this one since it has been replaced by #24496

PR #24097, opened 2025-03-19
[fix] [broker] topics infinitely failed to delete after remove cluster from replicated clusters modifying when using partitioned system topic

poorbarcode:
#### Background
- Envs: `broker.conf: allowAutoTopicCreationType = TopicType.PARTITIONED`
- Users may start 2 clusters with the Geo-Replication feature with Global ZK
- Users want to hire one cluster, so remove the cluster from `namespace-level replicatedClusters`
  - Broker will remove the topics of the cluster removed automatically.
- Issue occurs:
  - The system topic `__change_events` being deleted first.
  - User topics can not be deleted anymore, because the operation `delete topic-level policies` can not be executed successfully anymore.
  - The task `topic.checkReplication` and other operations will retry again and again

You can reproduce by the new test `OneWayReplicatorUsingGlobalPartitionedTest. testRemoveCluster`

### Modifications
- Skip removing topic-level policies if the system topic `__change_event` has been deleted.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
@@ -137,10 +137,18 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
             return CompletableFuture.completedFuture(null);
         }
         TopicName changeEvents = NamespaceEventsSystemTopicFactory.getEventsTopicName(topicName.getNamespaceObject());
-        return pulsarService.getNamespaceService().checkTopicExists(changeEvents).thenCompose(topicExistsInfo -> {
+        CompletableFuture<Boolean> changeEventTopicExists = pulsarService.getPulsarResources().getTopicResources()
+                .persistentTopicExists(changeEvents).thenCompose(nonPartitionedExists -> {
+            if (!nonPartitionedExists) {
+                return pulsarService.getPulsarResources().getTopicResources()
+                        .persistentTopicExists(changeEvents.getPartition(0));
+            }
+            return CompletableFuture.completedFuture(true);
+        });
+        return changeEventTopicExists.thenCompose(exists -> {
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -137,10 +137,18 @@ public CompletableFuture<Void> deleteTopicPoliciesAsync(TopicName topicName) {
             return CompletableFuture.completedFuture(null);
         }
         TopicName changeEvents = NamespaceEventsSystemTopicFactory.getEventsTopicName(topicName.getNamespaceObject());
-        return pulsarService.getNamespaceService().checkTopicExists(changeEvents).thenCompose(topicExistsInfo -> {
+        CompletableFuture<Boolean> changeEventTopicExists = pulsarService.getPulsarResources().getTopicResources()
+                .persistentTopicExists(changeEvents).thenCompose(nonPartitionedExists -> {
+            if (!nonPartitionedExists) {
+                return pulsarService.getPulsarResources().getTopicResources()
```
shibd, line 143: Why do we need to double-check partition-0 here?

poorbarcode: - When users use global ZK, the partitioned topic `admin/partitioned-topic` ZK node was relied on both clusters.

- After users remove `local cluster` from the namespace level policies, the node `/managed-ledger/<topic>-partition-x` will be deleted by `PersistentTopic.checkReplication`, but the ZK node named `admin/partitioned-topic/<topic>` will not be removed because the remote cluster still uses it.

Uses the expression whether `partition-0` was deleted  is better than `partitoned metedata`

Technoboy- (approved):

PR #24098, opened 2025-03-19
[fix][ml] Skip deleting cursor if it was already deleted before calling unsubscribe

poorbarcode:
### Motivation

We encountered an issue after trying to delete a topic multiple times, and all deletions failed, and there are some bundle rebalnces. After checking the heap dump. We found a subscription that relates to a delete cursor.

**Issue**:
- The subscription has a cursor
- but the cursor has been removed from the managed ledger

Screenshot 2025-03-26 at 16 18 10
Screenshot 2025-03-26 at 16 20 06

### Modifications
- Skip deleting the cursor if it has already was deleted

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/ConsumerUnsubscribeIntegrationTest.java
+++ b/pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/ConsumerUnsubscribeIntegrationTest.java
@@ -28,7 +28,7 @@ import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.Producer;
 import org.apache.pulsar.client.api.ProducerConsumerBase;
 import org.apache.pulsar.client.api.Schema;
-import org.testcontainers.shaded.org.awaitility.Awaitility;
+import org.awaitility.Awaitility;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/ConsumerUnsubscribeIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.concurrent.TimeUnit;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.broker.BrokerTestUtil;
+import org.apache.pulsar.broker.service.persistent.PersistentTopic;
+import org.apache.pulsar.client.api.Consumer;
+import org.apache.pulsar.client.api.MessageId;
+import org.apache.pulsar.client.api.Producer;
+import org.apache.pulsar.client.api.ProducerConsumerBase;
+import org.apache.pulsar.client.api.Schema;
+import org.testcontainers.shaded.org.awaitility.Awaitility;
```
lhotari, line 31: Shouldn't use shaded Awaitility

poorbarcode: Solved

lhotari (approved):
LGTM

PR #24100, opened 2025-03-20
[fix][client] Fix consumer leak when thread is interrupted before subscribe completes

shibd:
### Motivation

When the user has a thread try to subscribe(create a consumer) to a topic.
But the thread get interrupted before getting a response for the consumer creation. 

For example: 
- Thread A - call subscribe() method
- Thread A - blocking on waiting for the response of the consumer creation
- The user framework interrupted `Thread A` for aborting a job.
- Cannot close the consumer since user will get InterruptedException and in this case user can't get consumer object
- So, in the end, this consumer leaked.

### Modifications
- Added a variable named `interruptedBeforeConsumerCreation`. When this variable is true, the consumer needs to be closed.

### Verifying this change
- You can run  `testInterruptedWhenCreateConsumer ` to reproduce this issue.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBuilderImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBuilderImpl.java
@@ -31,6 +31,7 @@ import java.util.stream.Collectors;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.pulsar.client.admin.PulsarAdminException;
 import org.apache.pulsar.client.api.BatchReceivePolicy;
@@ -62,6 +63,7 @@ import org.apache.pulsar.client.util.RetryMessageUtil;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.util.FutureUtil;
 
+@Slf4j
 @Getter(AccessLevel.PUBLIC)
 public class ConsumerBuilderImpl<T> implements ConsumerBuilder<T> {
 
@@ -69,6 +71,7 @@ public class ConsumerBuilderImpl<T> implements ConsumerBuilder<T> {
     private ConsumerConfigurationData<T> conf;
     private final Schema<T> schema;
     private List<ConsumerInterceptor<T>> interceptorList;
+    private volatile boolean interruptedBeforeConsumerCreation;
 
     private static final long MIN_ACK_TIMEOUT_MILLIS = 1000;
     private static final long MIN_TICK_TIME_MILLIS = 100;
@@ -98,8 +101,31 @@ public class ConsumerBuilderImpl<T> implements ConsumerBuilder<T> {
 
     @Override
     public Consumer<T> subscribe() throws PulsarClientException {
+        CompletableFuture<Consumer<T>> future = new CompletableFuture<>();
         try {
-            return subscribeAsync().get();
+            subscribeAsync().whenComplete((c, e) -> {
+                if (e != null) {
+                    // If the subscription fails, there is no need to close the consumer here,
+                    // as it will be handled in the subscribeAsync method.
+                    future.completeExceptionally(e);
+                    return;
+                }
+                if (interruptedBeforeConsumerCreation) {
+                    c.closeAsync().exceptionally(closeEx -> {
+                        log.error("Failed to close consumer after interruption", closeEx.getCause());
+                        return null;
+                    });
+                    future.completeExceptionally(new PulsarClientException(
+                            "Subscription was interrupted before the consumer could be fully created"));
+                } else {
+                    future.complete(c);
+                }
+            });
+            return future.get();
+        } catch (InterruptedException e) {
+            interruptedBeforeConsumerCreation = true;
+            Thread.currentThread().interrupt();
+            throw PulsarClientException.unwrap(e);
         } catch (Exception e) {
             throw PulsarClientException.unwrap(e);
         }
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ConsumerCloseTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ConsumerCloseTest.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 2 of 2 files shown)


lhotari:
> * Cannot close the consumer since user will get InterruptedException and can't get consumer object

@shibd Good work! It would be useful to have a failing test case that reproduces this problem included in this PR. Could you please add it too? It seems that it would be a slight variation of the test case that you have added. The "can't get consumer object" part isn't currently tested so that after fixing the issue the "consumer object can be get" (the test would clarify what this actually means).

shibd:
> The "can't get consumer object" part isn't currently tested so that after fixing the issue the "consumer object can be get" (the test would clarify what this actually means).

In this case, regardless of this fix. the user wont get the consumer object, which is expected.

Because the subscribe throw an exception, that's not the focus. The focus is that we need to close this consumer.

lhotari:
> In this case, regardless of this fix. the user wont get the consumer object, which is expected.
> 
> Because the subscribe throw an exception, that's not the focus. The focus is that we need to close this consumer.

@shibd I point was that once there's a bug, a good practice would be to add a failing test case which reproduces the issue and clarifies it. In this case "the user wont get the consumer object, which is expected" could be understood in many ways. It might be clear to yourself, but not for code reviewers. That's why "failing test cases" reproducing the exact problem are always desired when it's possible to add one. Obviously for many Pulsar bugs that's really hard. In this case, it most likely isn't hard.

shibd:
You can run `testInterruptedWhenCreateConsumer ` to reproduce it.

lhotari:
> You can run `testInterruptedWhenCreateConsumer ` to reproduce it.

@shibd That's not reproducing this flow what you described in the PR description

> For example:
> 
> * Thread A - call subscribe() method
> * Thread A - blocking on waiting for the response of the consumer creation
> * The user framework interrupted `Thread A` for aborting a job.
> * Cannot close the consumer since user will get InterruptedException and can't get consumer object

The step "Cannot close the consumer since user will get InterruptedException and can't get consumer object" is missing. Adding a separate test case that reproduces the user's flow would be useful as I have explained before. Just simply put it in code in a test case which clarifies what "Cannot close the consumer since user will get InterruptedException and can't get consumer object" means. That's the most simplest way without wasting time in writing comments. :)

lhotari:
> I don't want to waste your time. I believe I've already added tests to cover it.

That's true, that there's a test, but the test is not from the viewpoint of the user of our software. The flow described in the description isn't performed in the test case. That would be an optimal "failing test case". In many software teams, this is a well known concept. However, in the Pulsar code base, most of our tests are not following usual best practices and don't appear to be good examples of how to write good tests. 

> You'll see that I've covered in the tests that the consumer will be removed(closed) after this PR fix.

That's not "Cannot close the consumer since user will get InterruptedException and can't get consumer object", what was in the description. I agree that the internal behavior change is covered in the test, but not the original bug that appeared.

> In this scenario, not getting the consumer object is expected. Why would we add a test case for it? A method throws an exception, and you want to cover that it returns null?

Simply test the scenario "Cannot close the consumer since user will get InterruptedException and can't get consumer object" and assume that it results in whatever is the expected result. It's not more harder than that.

> If you can review the code directly and let me know where improvements are needed, I'd be happy to optimize it.

As long as the mentioned scenario is handled in a separate test case, I'm fine with this PR. I know that it's unusual to knit pick on such details in our PRs, but unless attention is paid to details we won't be able to learn from each other and improve as a whole.

shibd:
First, I clearly asserted here that when the user calls `thread.interrupt `, the `subscribe` method will throw an exception. 

Isn't that enough?

https://github.com/apache/pulsar/pull/24100/files#diff-16facc169f1bbd7ab0a86d608b2aaa7ab8a578ff83c1d71d784b3df55d651559R71-R81

shibd:
hi, @lhotari , I respect all your opinions, but I really didn't understand what you meant.

Could you comment directly on the code about what needs to be changed? Or could you show the test you want directly in the code?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBuilderImpl.java
```diff
@@ -98,8 +101,31 @@ public ConsumerBuilder<T> clone() {
...
+                            "Subscription was interrupted before the consumer could be fully created"));
+                } else {
+                    future.complete(c);
+                }
+            });
+            return future.get();
+        } catch (InterruptedException e) {
+            interruptedBeforeConsumerCreation = true;
+            Thread.currentThread().interrupt();
+            throw PulsarClientException.unwrap(e);
```
codelipenghui, line 128:
```suggestion
            subscribeAsync().whenComplete((c, e) -> {
                if (e != null) {
                    // If the subscription fails, there is no need to close the consumer here,
                    // as it will be handled in the subscribeAsync method.
                    future.completeExceptionally(e);
                    return;
                }
                if (interruptedBeforeConsumerCreation) {
                    c.closeAsync().whenComplete((__, closeEx) -> {
                        if (closeEx != null) {
                            log.error("Failed to close consumer after interruption", closeEx.getCause());
                            future.completeExceptionally(new PulsarClientException(interruptedException));
                        } else {
                            future.completeExceptionally(new PulsarClientException(interruptedException));
                        }
                    })                    
                } else {
                    future.complete(c);
                }
            });
            return future.get();
        } catch (InterruptedException e) {
            interruptedBeforeConsumerCreation = true;
            interruptedException = e;
            Thread.currentThread().interrupt();
```

codelipenghui: Is it better to let the consumer get closed first and then return exception to user?

shibd: Yes, this is better from the code side.

However, to clarify, in this case, the user won't actually get this exception(`future.completeExceptionally(new PulsarClientException(interruptedException));`) because future.get() has already thrown an InterruptedException and returned it to the user.

codelipenghui (approved):

nodece (commented):
This issue is somewhat peripheral, but it poses a potential risk. In Java, thread interruptions are often ignored, which can lead to resource leaks. Additionally, we have many similar methods that do not properly handle interruptions.

To ensure correctness, users should be aware that they must handle thread interruptions appropriately.

When using the Pulsar Client/Admin, the users should not call the thread interrupt.

shibd:
hi, @nodece For operation type of api, such as: `ack()`, `nack()`. In my understanding, not handling interruptions is acceptable because, even during an interruption, whether the API operation succeeds or fails, it is acceptable to the user.

However, we cannot consider the consumer leak described in this PR as expected behavior; we should fix it.

lhotari:
> hi, @lhotari , I respect all your opinions, but I really didn't understand what you meant.
> 
> Could you comment directly on the code about what needs to be changed? Or could you show the test you want directly in the code?

Thanks @shibd. 

It is simply about "Cannot close the consumer since user will get InterruptedException and in this case user can't get consumer object". I don't understand the part "and in this case user can't get consumer object" and was expecting that a test case would have clarified it. 

Now, reading your comments again, it seems that "user can't get consumer object" is expected behavior and this sentence is just misleading and that's not something to be tested in the first place as you have described.
Rephrasing the sentence in the description would be a way to resolve this in that case. This isn't a blocker.
I'll check what a LLM would suggest for the PR description using the solution described in https://gist.github.com/lhotari/81f533af4b9ad515e02d96e543c4408b. It usually gives a pretty good draft and an LLM can be used to also fix grammar issues in the PR description.

lhotari:
@shibd this is the unedited DeepSeek suggestion: https://gist.github.com/lhotari/c76c08a59c6c17b27425cea04b7fa215

PR #24101, opened 2025-03-20
[fix][test] Move ReplicatorRateLimiterTest to broker test group

nodece:
### Motivation

The `ReplicatorRateLimiterTest` is crucial for verifying the correct behavior of the rate limiter in replication scenarios. If this test is marked as part of the `quarantine` test group, its failure will not block PR merges. This poses a risk: if the replicator limiter is broken, we will not detect regressions promptly.  

To maintain reliability and catch potential issues early, this test should remain a required test and not be moved to `quarantine`.  

### Modifications

- Move `ReplicatorRateLimiterTest` from the `quarantine` test group to the `broker` test group.
- Refactor assert by `Awaitility`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorRateLimiterTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorRateLimiterTest.java
@@ -109,19 +109,23 @@ public class ReplicatorRateLimiterTest extends ReplicatorTestBase {
             .ratePeriodInSecond(30)
             .build();
         admin1.topics().setReplicatorDispatchRate(topicName, topicRate);
-        Awaitility.await().untilAsserted(() ->
-            assertEquals(admin1.topics().getReplicatorDispatchRate(topicName), topicRate));
-        assertTrue(getRateLimiter(topic).isPresent());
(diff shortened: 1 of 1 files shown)


lhotari (dismissed):
> To maintain reliability and catch potential issues early, this test should remain a required test and not be moved to `quarantine`.

@nodece I agree. However the current tests might not be sufficient to ensure that the feature is working as expected.
It would be good to address the flakiness in the current tests as well. The tests don't seem to be very convincing from flakiness perspective. 

Awaitility should be used instead of using this type of logic:
```java
        replicatorUpdated = false;
        for (int i = 0; i < retry; i++) {
            if (topic.getReplicators().values().get(0).getRateLimiter().get().getDispatchRateOnByte() == messageRate) {
                replicatorUpdated = true;
                break;
            } else {
                if (i != retry - 1) {
                    Thread.sleep(100);
                }
            }
        }
        Assert.assertTrue(replicatorUpdated);
```

Before moving from quarantine, I think that the test code should be revisited.

dao-jun (approved):

PR #21853, opened 2024-01-04
[fix] [client] Messages lost due to TopicListWatcher reconnect

poorbarcode:
#### Issue
- Start a consumer subscribe with regexp `public/default/*`
- Create a new topic `public/default/tp-1`
- Send messages to the new topic
- Consumers will not receive any messages, and the messages will be deleted due to there are no subscription.

#### Root cause

**Case-1:**
| Time | Thread `start consumer` | Thread `create topic` |
| --- | --- | --- |
| 1 | Start subscribe |
| 2 | Start to register `TopicListWatcher` | Create new topic |
| 3 | | Topic created |
| 4 | | Since there are no listeners registered, there is no need to notice clients |
| 5 | `TopicListWatcher` registered, the consumer will not receive any messages from the new topic |

**Case-2:**
| Time | Thread `TopicListWatcher reconnect` | Thread `create topic` |
| --- | --- | --- |
| 1 | Start reconnect |
| 2 |  | Create new topic |
| 3 | | Topic created |
| 4 | | Since there are no listeners registered, there is no need to notice clients |
| 5 | `TopicListWatcher` reconnected |
| 6 | The consumer will not receive any messages from the new topic |

This issue only affects the releases `>=3.1`, because before `3.1` there is a supplementary mechanism `scheduled recheckPatternTimeout`, it makes the issues will not occur.  This supplementary mechanism was removed at `3.1`, see https://github.com/apache/pulsar/pull/20779

### Modifications
- After `TopicListWatcher` reconnected, re-check the topics list.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24104, opened 2025-03-21
[fix][client] Pattern subscription regression when broker-side evaluation is disabled

shibd:
### Motivation

When **EnableBrokerSideSubscriptionPatternEvaluation** is set to false, the pattern subscription can't discover newly created topics. 

This regression was introduced by PR #21853 and #22854, which removed pattern consumer regularity discovery topic logic: 

https://github.com/apache/pulsar/pull/22854/files#diff-779a47f80b0048bab183870ed4c1ecd07ffd538006cf9358e494bee23405d924L154-L159

### Modifications

- Rollback to poll mode for topic discovery when `watcherFuture` returns an exception.
- Remove `recheckPatternTaskBackoff` as it is unused and does not align with the user-side configuration `patternAutoDiscoveryPeriod`.
- Remove test code like `consumer.getRecheckPatternTimeout().task().run(consumer.getRecheckPatternTimeout()); ` from the unit tests, because these tests now operate under the premise of the broker actively pushing topic changes. 

### Verifying this change
- To reproduce and verify this issue, you can run the test testSubscribePatterBrokerDisable.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
@@ -56,6 +56,8 @@ public class PatternMultiTopicsConsumerImpl<T> extends MultiTopicsConsumerImpl<T
     protected NamespaceName namespaceName;
 
     private final AtomicInteger recheckPatternEpoch = new AtomicInteger();
+    // If recheckPatternTimeout is not null, it means the broker's topic watcher is disabled.
+    // The client need falls back to the polling model.
     private volatile Timeout recheckPatternTimeout = null;
     private volatile String topicsHash;
 
@@ -407,6 +409,11 @@ public class PatternMultiTopicsConsumerImpl<T> extends MultiTopicsConsumerImpl<T
     @VisibleForTesting
     int getRecheckPatternEpoch() {
         return recheckPatternEpoch.get();
+    }
+
+    @VisibleForTesting
+    Timeout getRecheckPatternTimeout() {
+        return recheckPatternTimeout;
     }
 
     protected void handleSubscribeOneTopicError(String topicName,
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
@@ -405,11 +405,6 @@ public class PatternMultiTopicsConsumerImpl<T> extends MultiTopicsConsumerImpl<T
     }
 
     @VisibleForTesting
-    Timeout getRecheckPatternTimeout() {
-        return recheckPatternTimeout;
-    }
-
-    @VisibleForTesting
     int getRecheckPatternEpoch() {
         return recheckPatternEpoch.get();
     }
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
@@ -44,8 +44,6 @@ import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.partition.PartitionedTopicMetadata;
 import org.apache.pulsar.common.topics.TopicList;
-import org.apache.pulsar.common.util.Backoff;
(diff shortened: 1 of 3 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
```diff
@@ -173,6 +158,11 @@ CompletableFuture<Void> recheckTopicsChange() {
                     return updateSubscriptions(topicsPattern, this::setTopicsHash, getTopicsResult,
                             topicsChangeListener, oldTopics, subscription);
                 }
+            }).thenAccept(__ -> {
+                if (recheckPatternTimeout != null) {
```
poorbarcode, line 162: After the change here, even if enabled `broker -> enableBrokerSideSubscriptionPatternEvaluation`, this task will also be scheduled triggered, which is unnecessary.

shibd: if enableBrokerSideSubscriptionPatternEvaluation, the recheckPatternTimeout  will == null.

I placed the initialization of recheckPatternTimeout in the exception of watcherFuture.

poorbarcode: Sure, we'd better add a comment for this variable to mark that it has two meanings

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
```diff
@@ -102,23 +86,24 @@ public PatternMultiTopicsConsumerImpl(Pattern topicsPattern,
...
                            + " for new topics", conf.getSubscriptionName(), ex);
+                   this.recheckPatternTimeout = client.timer().newTimeout(
+                           this, Math.max(1, conf.getPatternAutoDiscoveryPeriod()), TimeUnit.SECONDS);
                    return null;
                });
         } else {
             log.debug("Pattern consumer [{}] not creating topic list watcher for subscription mode {}",
                     conf.getSubscriptionName(), subscriptionMode);
             watcherFuture.complete(null);
+            this.recheckPatternTimeout = client.timer().newTimeout(
```
poorbarcode, line 105: How about adding a test that covers the scenarios that use non-persistent topic regexp?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
```diff
@@ -102,23 +86,24 @@ public PatternMultiTopicsConsumerImpl(Pattern topicsPattern,
...
             long watcherId = client.newTopicListWatcherId();
             new TopicListWatcher(updateTaskQueue, client, topicsPattern, watcherId,
                 namespaceName, topicsHash, watcherFuture, () -> recheckTopicsChangeAfterReconnect());
             watcherFuture
-               .thenAccept(__ -> recheckPatternTimeout.cancel())
                .exceptionally(ex -> {
                    log.warn("Pattern consumer [{}] unable to create topic list watcher. Falling back to only polling"
                            + " for new topics", conf.getSubscriptionName(), ex);
+                   this.recheckPatternTimeout = client.timer().newTimeout(
+                           this, Math.max(1, conf.getPatternAutoDiscoveryPeriod()), TimeUnit.SECONDS);
```
poorbarcode, line 98: We'd better set a larger min value because the frequent calling of `getTopicsUnderNamespace` will cause the issue that the CPU is fully used when there are thousands of topics under the same namespace.

poorbarcode (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
shibd: We need to keep the client's behavior as before, if broker is concerned about frequent polling, it should enable `EnableBrokerSideSubscriptionPatternEvaluation ` to avoid it.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java
shibd: Thanks, added.

poorbarcode (approved):

Issue #24094, opened 2025-03-19
Add tests for the dispatcherMaxRoundRobinBatchSize config


BewareMyPower:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

This config limits the max number of entries to dispatch for a `Shared` or `Key_Shared` subscription. However, there is no related test, so a regression could be introduced.

See https://github.com/apache/pulsar/blob/3e6f7deb0afec2675aac74750f7526557f43c9e7/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java#L1299

I added a TODO here: https://github.com/apache/pulsar/blob/76e9f9f927e9e21784761ea222cc155f40169c94/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L852

If this line was changed like:

```java
            int maxEntriesInThisBatch =
                            // use the average batch size per message to calculate the number of entries to
                            // dispatch. round up to the next integer without using floating point arithmetic.
                            (maxMessagesInThisBatch + avgBatchSizePerMsg - 1) / avgBatchSizePerMsg);
```

no tests will fail. This is dangerous because there is no test that guarantees the config works

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

3pacccccc:
i'm interest in this issue, assign to me please.

jojowyh:
I'm new to an open source project in China, and I would like to participate in this issue as well

BewareMyPower:
I see @jojowyh already pushed a PR, so I assigned this issue to him. But I left a comment here: https://github.com/apache/pulsar/pull/24110#discussion_r2009122125

PR #24110, opened 2025-03-22
[test] Add validation tests for dispatcherMaxRoundRobinBatchSize (https://github.com/apache/pulsar/issues/24094)

jojowyh:
[test] Add validation tests for dispatcherMaxRoundRobinBatchSize (https://github.com/apache/pulsar/issues/24094)

Fixes (https://github.com/apache/pulsar/issues/24094)

Main Issue:  (https://github.com/apache/pulsar/issues/24094)

Motivation: The `dispatcherMaxRoundRobinBatchSize` config currently lacks validation tests, which poses a risk of regressions during code changes. Adding these tests ensures that invalid configurations (e.g., negative values, non-integer types) are caught early.

Modifications:  
- Added unit tests in  `pulsar-broker/src/test/java/org/apache/pulsar/common/naming/ServiceConfigurationTest.java` to validate:  
  - Empty configuration uses default value (`dispatcherMaxRoundRobinBatchSize=20`).  
  - Valid numeric value (`30`) is accepted.  
  - Non-numeric input (`"invalidValueType"`) triggers `IllegalArgumentException` with error message matching `"failed to initialize dispatcherMaxRoundRobinBatchSize field while setting value invalidValueType"`. 

### Verifying this change
- [x] This change added tests and can be verified as follows:
  - Run `mvn test` to ensure all tests pass.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: (https://github.com/jojowyh/pulsar/commit/5bbc9ae3b27106f051b912f39205e6d210ffde34)

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/common/naming/ServiceConfigurationTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/common/naming/ServiceConfigurationTest.java
@@ -407,4 +407,32 @@ public class ServiceConfigurationTest {
                 ServiceConfiguration.class);
         assertEquals(conf.lookupProperties(), Map.of("lookup.key2", "value2"));
     }
+
+    @Test
+    public void testDispatcherMaxRoundRobinBatchSizeProperty() throws Exception {
+        // 1. set empty value
+        String emptyConf = "dispatcherMaxRoundRobinBatchSize=\n";
+        ServiceConfiguration defaultConfig = PulsarConfigurationLoader.create(
+                new ByteArrayInputStream(emptyConf.getBytes()),
+                ServiceConfiguration.class
+        );
+        assertEquals(defaultConfig.getDispatcherMaxRoundRobinBatchSize(),20);
+
+        // 2. set valid value: 30
+        String validConf = "dispatcherMaxRoundRobinBatchSize=30\n";
+        try (InputStream stream = new ByteArrayInputStream(validConf.getBytes())) {
+            ServiceConfiguration config = PulsarConfigurationLoader.create(stream, ServiceConfiguration.class);
+            assertEquals(config.getDispatcherMaxRoundRobinBatchSize(),30);
+        }
+
+        // 3. set invalid type: "invalidValueType"
+        String invalidConf = "dispatcherMaxRoundRobinBatchSize=invalidValueType\n";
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/common/naming/ServiceConfigurationTest.java
```diff
@@ -407,4 +407,32 @@ public void testLookupProperties() throws Exception {
...
+
+    @Test
+    public void testDispatcherMaxRoundRobinBatchSizeProperty() throws Exception {
+        // 1. set empty value
+        String emptyConf = "dispatcherMaxRoundRobinBatchSize=\n";
+        ServiceConfiguration defaultConfig = PulsarConfigurationLoader.create(
+                new ByteArrayInputStream(emptyConf.getBytes()),
+                ServiceConfiguration.class
+        );
+        assertEquals(defaultConfig.getDispatcherMaxRoundRobinBatchSize(),20);
```
BewareMyPower, line 419: Verify if the config change works is not what I want to do in https://github.com/apache/pulsar/issues/24094.

Instead, I want to verify if this work takes effect. For example, when `dispatcherMaxRoundRobinBatchSize` is 20, we should guarantee the dispatcher for `Shared` or `Key_Shared` subscriptions should not dispatch more than 20 entries each time.

See my TODO here: https://github.com/apache/pulsar/blob/76e9f9f927e9e21784761ea222cc155f40169c94/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L852

If this line was changed like:

```java
            int maxEntriesInThisBatch =
                            // use the average batch size per message to calculate the number of entries to
                            // dispatch. round up to the next integer without using floating point arithmetic.
                            (maxMessagesInThisBatch + avgBatchSizePerMsg - 1) / avgBatchSizePerMsg);
```

no tests will fail. This is dangerous because there is no test that guarantees the config works

BewareMyPower (changes requested):

PR #19171, opened 2023-01-10
[fix][broker] Allow user lookup topic name with `-partition-` but no metadata

mattisonchao:
### Motivation

This behaviour is broken by PR #18193,

In the previous version, we could use the pulsar client to look up the topic with the `-partition-` keyword. Even this topic doesn't have partition metadata. But currently, we can't do that. You can check the test to know the detail.

### Modifications

- Allow user lookup topic name with `-partition-` but no metadata

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Documentation

- [x] `doc-not-needed`

PR #24118, opened 2025-03-25
[fix][broker] Add topic consistency check

nodece:
### Motivation

In Apache Pulsar, clients can produce or consume messages using a fully qualified partition topic name (`public/default/test-partitioned-topic-partition-N`). However, if the topic metadata does not exist and `allowAutoTopicCreationType=non-partitioned` is enabled, Pulsar automatically creates the topic as non-partitioned. This leads to inconsistencies when listing topics, as the expected partitioned topic structure is not maintained.

This issue is especially prevalent in geo-replicated environments, where topic metadata may not be fully synchronized across all regions. Currently, Pulsar lacks validation to ensure that the topic type (partitioned or non-partitioned) matches the actual partition metadata.

### Modifications

- Added a topic consistency check:
  - **For non-partitioned topic**: If partition metadata exists (which shouldn't be there), the broker will reject message produce and consume operations.
  - **For partitioned topic**: If partition metadata is missing (which should be there), the broker will reject message produce and consume operations.
- Fixed affected tests to align with the new validation logic.
- Added `testGetPartitionedTopicMetadataByPulsarClient` and `testGetPartitionedTopicMedataByAdmin` instead of `testCompatibilityWithPartitionKeyword` because the broker rejects message produce and consume operations when no metadata. Note: we still allow user lookup topic name with `-partition-` but no metadata: https://github.com/apache/pulsar/pull/19171.
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
@@ -1412,17 +1412,20 @@ public class NamespaceService implements AutoCloseable {
      * Check topic exists( partitioned or non-partitioned ).
      */
     public CompletableFuture<TopicExistsInfo> checkTopicExists(TopicName topic) {
-        return pulsar.getBrokerService()
-            .fetchPartitionedTopicMetadataAsync(TopicName.get(topic.toString()))
-            .thenCompose(metadata -> {
-                if (metadata.partitions > 0) {
-                    return CompletableFuture.completedFuture(
-                            TopicExistsInfo.newPartitionedTopicExists(metadata.partitions));
-                }
-                return checkNonPartitionedTopicExists(topic)
-                    .thenApply(b -> b ? TopicExistsInfo.newNonPartitionedTopicExists()
-                            : TopicExistsInfo.newTopicNotExists());
-            });
+        // For non-persistent/persistent partitioned topic, which has metadata.
+        return pulsar.getBrokerService().fetchPartitionedTopicMetadataAsync(
+                        topic.isPartitioned() ? TopicName.get(topic.getPartitionedTopicName()) : topic)
+                .thenCompose(metadata -> {
+                    if (metadata.partitions > 0) {
+                        return CompletableFuture.completedFuture(
+                                topic.isPartitioned() ? TopicExistsInfo.newNonPartitionedTopicExists() :
+                                        TopicExistsInfo.newPartitionedTopicExists(metadata.partitions));
+                    }
+                    // Direct query the single topic.
+                    return checkNonPartitionedTopicExists(topic).thenApply(
+                            b -> b ? TopicExistsInfo.newNonPartitionedTopicExists() :
+                                    TopicExistsInfo.newTopicNotExists());
+                });
     }
 
     /***
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
@@ -1443,12 +1443,12 @@ public class NamespaceService implements AutoCloseable {
      */
     public CompletableFuture<Boolean> checkNonPersistentNonPartitionedTopicExists(String topic) {
         TopicName topicName = TopicName.get(topic);
-        // "non-partitioned & non-persistent" topics only exist on the owner broker.
+        // "non-partitioned & non-persistent" topics only exist on the cache of the owner broker.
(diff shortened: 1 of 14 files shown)


dao-jun (approved):

pulsar-broker/src/test/java/org/apache/pulsar/client/api/ConsumerCreationTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.pulsar.client.api.PulsarClientException.NotAllowedException;
+import org.apache.pulsar.common.naming.TopicDomain;
+import org.apache.pulsar.common.naming.TopicName;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+@Test(groups = "broker-api")
+public class ConsumerCreationTest extends ProducerConsumerBase {
```
lhotari, line 34: Would it make sense to add a test case where there's a partitioned topic (let's say with 4 partitions) and one of the partitions (lets say *-partition-2)  is deleted before the customer subscribes. Validating that the *-partition-2 gets autocreated as expected in cases where it should.

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
```diff
@@ -1412,17 +1412,20 @@ public CompletableFuture<List<String>> getOwnedTopicListForNamespaceBundle(Names
...
-                    .thenApply(b -> b ? TopicExistsInfo.newNonPartitionedTopicExists()
-                            : TopicExistsInfo.newTopicNotExists());
-            });
+        // For non-persistent/persistent partitioned topic, which has metadata.
+        return pulsar.getBrokerService().fetchPartitionedTopicMetadataAsync(
+                        topic.isPartitioned() ? TopicName.get(topic.getPartitionedTopicName()) : topic)
+                .thenCompose(metadata -> {
+                    if (metadata.partitions > 0) {
+                        return CompletableFuture.completedFuture(
+                                topic.isPartitioned() ? TopicExistsInfo.newNonPartitionedTopicExists() :
```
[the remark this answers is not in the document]
nodece: Good catch, `topic.getPartitionIndex() < metadata.partitions` should be added to here.

pulsar-broker/src/test/java/org/apache/pulsar/client/api/ConsumerCreationTest.java
nodece: Your case makes sense. When the auto-creation is enabled, it should work fine, otherwise, throw a not found exception.

nodece:
@lhotari Your coment has been addressed.

lhotari (approved):
LGTM, good work @nodece

lhotari:
@nodece What Pulsar versions are targeted for this change? Although the change LGTM, I wonder if this could introduce any regressions if this is cherry-picked to maintenance branches? Could @poorbarcode and @shibd also review?

nodece:
> I wonder if this could introduce any regressions if this is cherry-picked to maintenance branches

Once the user uses the `topic-partition-N` format, but no metadata, the pulsar rejects the message production and consumption, but looking at it from a different perspective, I believe this is correct as it can help users use pulsar correctly.

So I want to cherry-pick to `branch-3.0`.

lhotari:
> > I wonder if this could introduce any regressions if this is cherry-picked to maintenance branches
> 
> Once the user uses the `topic-partition-N` format, but no metadata, the pulsar rejects the message production and consumption, but looking at it from a different perspective, I believe this is correct as it can help users use pulsar correctly.
> 
> So I want to cherry-pick to `branch-3.0`.

@nodece yes, makes sense.

BewareMyPower:
Unfortunately, this PR DOES BREAK the downstream applications. I can even accept that new features are cherry-picked into LTS releases. But it's hard to accept a PR that changes many existing tests is cherry-picked without broader notifications and discussion.

If an existing test can be reverted so easily, why are we still adding tests? Just delete or modify a test if we want to make a breaking change.

Let's continue the discussion in https://lists.apache.org/thread/llobypz1nc9w6wdnwd3z1h52z7bygdby

lhotari:
Reverted from branch-3.0, branch-3.3 and branch-4.0 as per [this dev mailing list message](https://lists.apache.org/thread/vdtwlr87kyvn1s7rqozxglz381d8mhp0).

PR #24130, opened 2025-03-26
[improve][client] Refactor wrap and unwrap methods in PulsarClientException

nodece:
### Motivation

The current implementation of `wrap` and `unwrap` methods in PulsarClientException use long chains of if-else statements to handle different exception types. This approach has several drawbacks:

- Maintainability issues: When new exception types are added, developers must remember to update both methods, which is error-prone and can lead to inconsistent behavior.

- Code duplication: The same pattern is repeated for each exception type, resulting in nearly 200 lines of duplicated logic.

- Potential for bugs: If a new exception type is added but the methods aren't updated, the exception handling will fall back to generic handling, potentially losing important context.

This PR refactors these methods to use a more maintainable and robust approach that automatically handles all exception types, including future ones.

### Modifications

- Refactored wrap method:
  - Replaced the if-else chain with a reflection-based approach that creates a new exception of the same type.
  - Preserved stack traces and cause information when wrapping exceptions.
  - Added special handling for exceptions that shouldn't be wrapped (CompletionException, InterruptedException, ExecutionException).

- Refactored unwrap method:

  - Simplified the method to recursively unwrap nested exceptions.
  - Eliminated the need to maintain a list of all exception types.
  - Preserved stack traces in the unwrapped exceptions.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
@@ -995,8 +995,6 @@ public class PulsarClientException extends IOException {
     public static PulsarClientException unwrap(Throwable t) {
         if (t instanceof PulsarClientException) {
             return (PulsarClientException) t;
-        } else if (t instanceof RuntimeException) {
-            throw (RuntimeException) t;
         } else if (t instanceof InterruptedException) {
             return createPulsarException(t);
         } else if (!(t instanceof ExecutionException)) {
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
@@ -19,6 +19,7 @@
 package org.apache.pulsar.client.api;
 
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
```diff
@@ -959,177 +960,55 @@ public TransactionHasOperationFailedException(String msg) {
...
 
-        return t;
+        try {
+            String combinedMessage = msg;
+            String originalMessage = t.getMessage();
+            if (originalMessage != null) {
+                combinedMessage += "\n" + originalMessage;
+            }
+            Constructor<? extends Throwable> constructor = t.getClass().getConstructor(String.class);
+            Throwable newException = constructor.newInstance(combinedMessage);
```
lhotari, line 978: Using reflection is a problem. It's better to remain using pure Java for this use case. Reflection will also cause issues with GraalVM native compilation.

lhotari (changes requested):
Replacing the existing code with reflection isn't a great solution. This will cause a breaking change and possibly also break GraalVM native compilation, if someone is already doing that for the Pulsar client.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java
lhotari: It might be possible to achieve a similar solution without reflection. I'd prefer instance methods in this case. The method names "wrap" and "unwrap" are not very intuitive, but it's impossible to rename the methods since that could possible break the API if these are considered to be part of the public API.

nodece: Makes scense.

PR #24166, opened 2025-04-10
[fix][test] Fix invalid test CompactionTest.testDeleteCompactedLedgerWithSlowAck

lhotari:
### Motivation

The test CompactionTest.testDeleteCompactedLedgerWithSlowAck fails with PR #23980. The test appears to be invalid and needs to be revisited.

### Modifications

- revisit CompactionTest.testDeleteCompactedLedgerWithSlowAck
- replace the usage of long deprecated topic naming format where the cluster name is part of the topic name
  - ["Many test classes are using deprecated topic naming format V1 which includes the cluster name" on dev mailing list](https://lists.apache.org/thread/60ysbhrgxfltyjwqybgzct1t5csztn2s)
    - I'm planning to do this modification to other test classes later
  - this format is not compatible with topic policies
- improve the CompactionTest test class by doing some other cleanup such are using static imports for assertion methods.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
@@ -18,15 +18,19 @@
  */
 package org.apache.pulsar.compaction;
 
+import static org.apache.pulsar.broker.BrokerTestUtil.newUniqueName;
+import static org.apache.pulsar.broker.BrokerTestUtil.spyWithoutRecordingInvocations;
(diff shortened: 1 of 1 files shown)


nodece (approved):
LGTM.
﻿
However, this PR changes the topic name format from v1 to v2, which I am concerned will cause a conflict when cherry-picking this PR to the maintenance branches.

Would make sense that create a new PR to improve that?

dao-jun (approved):
lgtm

PR #23093, opened 2024-07-29
[improve][build] Move docker-push profile to submodule

nodece:
### Motivation

The profile refactoring breaks the pulsar release in the #23091.

### Modifications

- Move docker-push profile to docker/pulsar and docker/pulsar-all modules

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #23091, opened 2024-07-29
[improve][build] Upgrade docker-maven-plugin to 0.45.0

nodece:
### Motivation

docker-maven-plugin keeps up to the latest version.

### Modifications

- Improve `docker-push` profile
- Remove tag goal, because build goal includes tag feature
- Upgrade docker-maven-plugin to 0.45.0

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24168, opened 2025-04-10
[fix][build] Fix skipTag and use explicit tag for image name

nodece:
### Motivation

The `docker-maven-plugin` supports skipping Docker image tags using the `docker.skip.tag` property. However, it does not work correctly when building multi-architecture images using `buildx`. As a workaround, the plugin provides a `skipTag` option in the [build configuration](https://dmp.fabric8.io/#build-configuration), which successfully skips tagging in multi-arch builds/pushes.

**Issue (Old behavior):**

Running the following command:
```shell
cd docker/pulsar
mvn install -Pdocker -Ddocker.skip.tag=true -Ddocker.platforms=linux/arm64,linux/amd64
```

Still resulted in tagging:
```
--tag apachepulsar/pulsar:latest
--tag apachepulsar/pulsar:latest
--tag apachepulsar/pulsar:4.1.0-SNAPSHOT-e8be56d
```

**Fixed behavior:**

By using the `skipTag` configuration:
```
--tag apachepulsar/pulsar:latest
```

Only the explicitly defined tag is applied.

### My use case

If you want to publish only a single tag like `myorg/pulsar:3.0.1` to Docker Hub, you can use:
```shell
mvn -B -ntp package -pl docker/pulsar -Pdocker,docker-push \
  -Ddocker.platforms=linux/amd64,linux/arm64 \
  -Ddocker.organization=myorg \
  -Ddocker.image=pulsar \
  -Ddocker.tag=3.0.1 \
  -Ddocker.skip.tag=true
```

This ensures that only `myorg/pulsar:3.0.1` is pushed.

If you want to publish both `myorg/pulsar:3.0.1` and the default tag `myorg/pulsar:${project.version}-${git.commit.id.abbrev}` (commonly used for integration testing), set:
```shell
-Ddocker.skip.tag=false
```

For most production cases, a single stable tag is preferred, and the additional tag can be reserved for internal test pipelines.

### Modifications

- Explicitly set the Docker image tag using the `docker.tag` property.
- Added `skipTag` option in the `<build>` configuration for both `pulsar` and `pulsar-all` images to prevent unnecessary automatic tagging during multi-architecture builds, default to `false`.
- Removed the `<tags>` configuration from the test image, as no tags are required for it.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docker/pulsar-all/pom.xml
+++ b/docker/pulsar-all/pom.xml
@@ -154,11 +154,11 @@
                 <configuration>
                   <images>
                     <image>
-                      <name>${docker.organization}/${docker.image}-all</name>
+                      <name>${docker.organization}/${docker.image}-all:${docker.tag}</name>
                       <build>
                         <contextDir>${project.basedir}</contextDir>
+                        <skipTag>${docker.skip.tag}</skipTag>
                         <tags>
-                          <tag>${docker.tag}</tag>
                           <tag>${project.version}-${git.commit.id.abbrev}</tag>
                         </tags>
                         <args>
--- a/docker/pulsar/pom.xml
+++ b/docker/pulsar/pom.xml
@@ -77,7 +77,7 @@
                 <configuration>
                   <images>
                     <image>
-                      <name>${docker.organization}/${docker.image}</name>
+                      <name>${docker.organization}/${docker.image}:${docker.tag}</name>
                       <build>
                         <args>
                           <PULSAR_TARBALL>target/pulsar-server-distribution-${project.version}-bin.tar.gz</PULSAR_TARBALL>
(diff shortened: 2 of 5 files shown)


lhotari:
@nodece Please explain the use case for using skipTag. Adding that context to the description would be helpful.

lhotari (approved):

nodece:
@lhotari This PR description has been updated.

nodece:
`org.apache.pulsar.broker.intercept.ManagedLedgerInterceptorImplTest.testManagedLedgerPayloadInputProcessorFailure` always fails.

lhotari:
This depends on #23091 and #23093 which aren't in branch-3.3, removing release label for branch-3.3

lhotari:
Adding `-Ddocker.skip.tag=false` to release process instructions, since releasing 4.0.5 initially failed due to this change.

nodece:
Got, because the pulsar-all depends on the image with git hash. I will fix this tomorrow.

PR #24171, opened 2025-04-10
[improve][test] Use configured session timeout for MockZooKeeper and TestZKServer in PulsarTestContext

lhotari:
### Motivation

After #23988, MockZooKeeper is correctly single threaded and an injected delay will block the execution of this code in ZKSessionWatcher:

This will trigger a ConnectionLost session event. That's why a delay should be kept under 2000 ms. 
Currently MockZooKeeper's session timeout is hard coded to 30000ms and the check interval is 1/15 of this, therefore 2000ms.

When there's a test that blocks the ZooKeeper thread for more than 2000ms, it will result in a ConnectionLost event.
This PR makes the session timeout configurable and it's possible to increase the session timeout for tests where it's necessary to use a longer session timeout to avoid ConnectionLost session events.

### Modifications

- refactor the code in PulsarTestContext to postpone creation of the instances after the configuration including the configured session timeout is available
- also configure other `MetadataStoreConfig` settings and pass them from the `ServiceConfiguration` instance

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/testcontext/PulsarTestContext.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/testcontext/PulsarTestContext.java
@@ -239,6 +239,18 @@ public class PulsarTestContext implements AutoCloseable {
                 getPulsarService());
     }
 
+    private enum WithMockZooKeeperOrTestZKServer {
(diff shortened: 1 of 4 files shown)


nodece (approved):
LGTM

PR #24165, opened 2025-04-09
[fix][test] Fix UnfinishedStubbing issue in AuthZTests

lhotari:
### Motivation

[Mockito is not thread safe](https://github.com/mockito/mockito/wiki/FAQ#is-mockito-thread-safe) and modifying mocks in multi-threaded code will result in failures such as this one:

```
   [ERROR] Failures:
    [ERROR]   NamespaceAuthZTest.testMaxConsumersPerTopic:1537->setAuthorizationPolicyOperationChecker:183 » UnfinishedStubbing
    Unfinished stubbing detected here:
    -> at org.apache.pulsar.broker.admin.NamespaceAuthZTest.setAuthorizationPolicyOperationChecker(NamespaceAuthZTest.java:173)

    E.g. thenReturn() may be missing.
    Examples of correct stubbing:
        when(mock.isOk()).thenReturn(true);
        when(mock.isOk()).thenThrow(exception);
        doThrow(exception).when(mock).someVoidMethod();
    Hints:
     1. missing thenReturn()
     2. you are trying to stub a final method, which is not supported
     3. you are stubbing the behaviour of another mock inside before 'thenReturn' instruction is completed
```

### Modifications

- refactor the test code to setup the mock once and switch the implementation dynamically

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24174, opened 2025-04-11
[fix][test] Fix remaining UnfinishedStubbingException issue with AuthZTests

lhotari:
### Motivation

Follow up on #24165. There's a remaining issue since mocks get modified in multi-threaded context in tests using `AuthZTest` as base class which modifies mocks in `@BeforeMethod`. That's not necessary with the solution and it could happen as soon as the broker is started so that the problem doesn't occur.

```
Error:  org.apache.pulsar.broker.admin.TransactionAndSchemaAuthZTest.after  Time elapsed: 0.012 s  <<< FAILURE!
  org.mockito.exceptions.misusing.UnfinishedStubbingException: 
  
  Unfinished stubbing detected here:
  -> at org.apache.pulsar.broker.admin.AuthZTest.before(AuthZTest.java:88)
  
  E.g. thenReturn() may be missing.
  Examples of correct stubbing:
      when(mock.isOk()).thenReturn(true);
      when(mock.isOk()).thenThrow(exception);
      doThrow(exception).when(mock).someVoidMethod();
  Hints:
   1. missing thenReturn()
   2. you are trying to stub a final method, which is not supported
   3. you are stubbing the behaviour of another mock inside before 'thenReturn' instruction is completed
  
  	at org.apache.pulsar.broker.PulsarService.getBrokerService(PulsarService.java:217)
  	at org.apache.pulsar.broker.admin.AuthZTest.after(AuthZTest.java:101)
```

### Modifications

- move mock setup from a `@BeforeMethod` to the `startup` method which runs after the broker has been started.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AuthZTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AuthZTest.java
@@ -23,6 +23,7 @@ import io.jsonwebtoken.Jwts;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
+import lombok.SneakyThrows;
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.pulsar.broker.BrokerTestUtil;
 import org.apache.pulsar.broker.authorization.AuthorizationService;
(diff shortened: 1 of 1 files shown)


nodece (approved):

lhotari:
depends on #22546 which isn't included in branch-3.0, dropping cherry-picking to branch-3.0 for now

PR #24175, opened 2025-04-11
[fix][client] Set local address only if not present

nodece:
### Motivation

When a consumer or producer connects to the broker via Pulsar Proxy, it sets the `X-Pulsar-Client-IP` header to indicate the real client IP.  
To avoid overwriting this user-defined value, the local address should only be added to the metadata if the header is missing.  
This ensures that `admin.topics().stats()` returns the correct address.

`X-Pulsar-Client-IP` is introduced by https://github.com/apache/pulsar/pull/23974

### Modifications

- Use `putIfAbsent` instead of `put` to avoid overwriting X-Pulsar-Client-IP in the metadata.
- Added a test to cover this behavior.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -3163,7 +3163,7 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         boolean isProxy = cnx.isProxy() || client.getConfiguration().getProxyServiceUrl() != null;
         if (isProxy && cnx.getLocalAddress() != null) {
             metadata = metadata.isEmpty() ? new HashMap<>() : metadata;
-            metadata.put(Metadata.CLIENT_IP, cnx.getLocalAddress().toString());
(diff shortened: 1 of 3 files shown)


lhotari (changes requested):
The `X-Pulsar-Client-IP` solution isn't great at all. The information cannot be trusted.
Any client could pass on any information they wish and impersonate another client IP.

The Pulsar Proxy already implements HA Proxy Protocol for passing on the client IP address from proxy to broker.
Pulsar Proxy code location:
https://github.com/apache/pulsar/blob/2a89b7f1bea4c8dfcf48798346e3b02683fa87a2/pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/DirectProxyHandler.java#L295-L297

In a proper configuration, you'd only allow traffic to your Pulsar Proxies via external load balancers that pass on the client IP using HA Proxy protocol.

lhotari:
I have created a PR to revert #23974 in #24177.

nodece:
https://github.com/apache/pulsar/pull/24177 has been merged, so this PR is unless.

PR #24177, opened 2025-04-11
[revert] Revert "[improve][broker] Support showing client ip address in client stats while using reverse proxy (#23974)"

lhotari:
This reverts commit 0cc266d54d69205c78dc1bcf03f0b608c20fb62b / #23974

### Motivation

The solution in #23974 introduces a security issue since any client can impersonate any client IP address.
Adding new features commonly are introduced in Pulsar by first creating a PIP. That would be necessary for such changes since there's a security impact.

Pulsar contains already support for HA Proxy Protocol for passing original client IP information from reverse proxies and load balancers to Pulsar Proxy and/or Pulsar Broker. That would also cover this use case since HA Proxy Protocol support exists also in SNI reverse proxies such as Apache Traffic Server. The documentation in Pulsar should be improved to contain detailed instructions about HA Proxy Protocol support for preserving client IP information across the different layers in a secure way.

### Modifications

Revert #23974 

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
@@ -48,7 +48,6 @@ import org.apache.bookkeeper.mledger.Entry;
 import org.apache.bookkeeper.mledger.Position;
 import org.apache.bookkeeper.mledger.PositionFactory;
 import org.apache.bookkeeper.mledger.impl.AckSetStateUtil;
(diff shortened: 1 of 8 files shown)


nodece (approved):
Using HAProxy is the correct approach, and I've verified that it works as expected. 

When the producer/consumer metadata is configured with a local address such as 127.0.0.1 or localhost, calling `admin.topics().stats(TOPIC)` will get to that local address.

PR #24181, opened 2025-04-15
[fix][proxy] Fix incorrect client error when calling get topic metadata

poorbarcode:
### Motivation

Proxy overwrites multiple error codes to `LookupError`, which may cause issues that many expressions, such as `ex instance of TopicDoesNotExistException` get a `false` result even if the broker responds a `TopicDoesNotExistException`

The code lines that use the expression `ex instance of TopicDoesNotExistException`
- https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L988
- https://github.com/apache/pulsar/blob/master/pulsar-websocket/src/main/java/org/apache/pulsar/websocket/AbstractWebSocketHandler.java#L160
- https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java#L429
- ...

### Modifications

- Correct the error code

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-proxy/src/test/java/org/apache/pulsar/proxy/server/ProxyTest.java
+++ b/pulsar-proxy/src/test/java/org/apache/pulsar/proxy/server/ProxyTest.java
@@ -24,6 +24,7 @@ import static java.util.Objects.requireNonNull;
 import static org.mockito.Mockito.doReturn;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.EventLoopGroup;
 import io.netty.util.concurrent.DefaultThreadFactory;
@@ -52,6 +53,7 @@ import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessageRoutingMode;
 import org.apache.pulsar.client.api.Producer;
 import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.client.api.Schema;
 import org.apache.pulsar.client.api.SubscriptionInitialPosition;
 import org.apache.pulsar.client.api.SubscriptionType;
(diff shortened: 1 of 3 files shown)


Technoboy- (approved):

nodece:
I suggest you check the `org.apache.pulsar.client.api.PulsarClientException#wrap` and `org.apache.pulsar.client.api.PulsarClientException#unwrap` calls to get the correct error.

poorbarcode:
@nodece @liangyepianzhou 

> I suggest you check the org.apache.pulsar.client.api.PulsarClientException#wrap and org.apache.pulsar.client.api.PulsarClientException#unwrap calls to get the correct error.

Good pointer, changed the code

pulsar-proxy/src/test/java/org/apache/pulsar/proxy/server/ProxyTest.java
```diff
@@ -415,6 +426,35 @@ public void testProtocolVersionAdvertisement() throws Exception {
...
+                (PulsarClientImpl) getClientActiveConsumerChangeNotSupported(proxyClientConf);
+        PartitionedTopicMetadata proxyMetadata =
+                proxyClient.getPartitionedTopicMetadata(topic, false, false).get();
+        assertEquals(proxyMetadata.partitions, 1);
+        try {
+            proxyClient.getPartitionedTopicMetadata(topic + "-partition-0", false, false).get();
+        } catch (Exception ex) {
+            assertTrue(FutureUtil.unwrapCompletionException(ex)
+                    instanceof PulsarClientException.TopicDoesNotExistException);
+        }
```
nodece, line 452:
```suggestion
        assertThat(proxyClient.getPartitionedTopicMetadata(topic + "-partition-0", false, false))
                .failsWithin(5, TimeUnit.SECONDS)
                .withThrowableThat()
                .satisfies(n -> {
                    assertThat(FutureUtil.unwrapCompletionException(n)).isInstanceOf(
                            PulsarClientException.TopicDoesNotExistException.class);
                });
```

poorbarcode: I think the current code is simpler and more readable.

nodece: Totally fair — the `try-catch` is very direct. I lean toward the AssertJ style because it gives a built-in timeout and keeps the assertion logic more fluent, especially if we expand the test later. But I'm fine either way — just wanted to share the option.

- (try-catch + assertTrue)
  Cons:
    - No timeout control — may hang if the future never completes.
    - Less elegant when checking nested or multiple failure cases.

nodece (approved):
Overall looks good.

lhotari (approved):

PR #24186, opened 2025-04-17
[fix][broker] Fixes Inconsistent ServiceUnitStateData View (ExtensibleLoadManagerImpl only)

heesung-sn:
### Motivation

Although we do CAS updates on ServiceUnitStateMetadataStoreTableViewImpl.put, ServiceUnitStateData.versionId can still conflict if cache.readModifyUpdateOrCreate retries multiple times with the updated value's GetResult.Stats.version. 

We should fail-fast if the readModifyUpdateOrCreate fails once(e.g. upon first BadVersionException) not to cause inconsistent ServiceUnitStateData views from different brokers.

Also, to fix the worst case, any inconsistent ServiceUnitStateData views, ServiceUnitStateMetadataStoreTableViewImpl should update the latest value(not only compare the versionId but also all fields) from the MetadataStore by its cache refresh cycles.

### Modifications

- Added No backoff config in MetadataCacheConfig and use it from ServiceUnitStateMetadataStoreTableViewImpl.
- Update ServiceUnitStateData.equals to compare all fields.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataCache.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataCache.java
@@ -103,22 +103,6 @@ public interface MetadataCache<T> {
     CompletableFuture<T> readModifyUpdateOrCreate(String path, Function<Optional<T>, T> modifyFunction);
 
     /**
-     * Perform an atomic read-modify-update of the value with the cached version.
-     * <p>
-     * This fail-fasts without retries (e.g. fail-fast upon BadVersionException)
-     * <p>
-     * If the object does not exist yet, the <code>modifyFunction</code> will get passed an {@link Optional#empty()}
-     * object.
-     *
-     * @param path
-     *            the path of the object in the metadata store
-     * @param modifyFunction
-     *            a function that will be passed the current value and returns a modified value to be stored
-     * @return a future to track the completion of the operation
-     */
-    CompletableFuture<T> readModifyUpdateOrCreateOnce(String path, Function<Optional<T>, T> modifyFunction);
-
-    /**
      * Perform an atomic read-modify-update of the value.
      * <p>
      * The modify function can potentially be called multiple times if there are concurrent updates happening.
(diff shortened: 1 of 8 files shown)


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/tableview/impl/MetadataStoreTableViewImpl.java
```diff
@@ -299,14 +299,23 @@ public T get(String key) {
...
-        }).thenCompose(__ -> doHandleNotification(path)); // immediately notify local tableview
+        return cache.readModifyUpdateOrCreateOnce(path, (old) -> {
+                    if (conflictResolver.test(old.orElse(null), value)) {
+                        return value;
+                    } else {
+                        throw new ConflictException(
+                                String.format("Failed to update from old:%s to value:%s", old, value));
+                    }
+                })
+                .thenCompose(__ -> doHandleNotification(path))  // immediately notify local tableview
```
BewareMyPower, line 310: These indent changes seems not necessary.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataCache.java
```diff
@@ -102,6 +102,22 @@ public interface MetadataCache<T> {
...
+     * If the object does not exist yet, the <code>modifyFunction</code> will get passed an {@link Optional#empty()}
+     * object.
+     *
+     * @param path
+     *            the path of the object in the metadata store
+     * @param modifyFunction
+     *            a function that will be passed the current value and returns a modified value to be stored
+     * @return a future to track the completion of the operation
+     */
+    CompletableFuture<T> readModifyUpdateOrCreateOnce(String path, Function<Optional<T>, T> modifyFunction);
```
BewareMyPower, line 119: It would be better to add a config (e.g. `maxRetries`) to `MetadataCacheConfig` rather than adding a new API.

Currently we already have a `retryBackoff` config to control the retry backoff. But it cannot prevent the 1st retry.

BewareMyPower: With a config, it will be more flexible to users that they can choose to implement their own retry mechanism or leverage the built-in retry mechanism.

heesung-sn: updated

BewareMyPower (approved):

PR #24190, opened 2025-04-21
[fix][broker] Resolve the issue of frequent updates in message expiration deletion rate

zjxxzjwang:
### Modifications

At present, the message expiration deletion rate is updated in the getmessageExpiryRate() method, and this rate is updated every time the topic status is obtained. This results in the statistical analysis of message expiration and deletion rates not having a fixed periodicity, making it difficult to represent the true message expiration rate.
### Documentation

Only update the rate in the updateRate() method and only retrieve the value in the getmessageExpiryRate() method

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -2536,6 +2536,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                 // Populate subscription specific stats here
                 topicStatsStream.writePair("msgBacklog",
                         subscription.getNumberOfEntriesInBacklog(true));
+                subscription.getExpiryMonitor().updateRates();
                 topicStatsStream.writePair("msgRateExpired", subscription.getExpiredMessageRate());
                 topicStatsStream.writePair("msgRateOut", subMsgRateOut);
                 topicStatsStream.writePair("messageAckRate", subMsgAckRate);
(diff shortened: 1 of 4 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -2457,6 +2457,7 @@ public void updateRates(NamespaceStats nsStats, NamespaceBundleStats bundleStats
                 // Populate subscription specific stats here
                 topicStatsStream.writePair("msgBacklog",
                         subscription.getNumberOfEntriesInBacklog(true));
+                subscription.getExpiryMonitor().updateRates();
```
nodece, line 2460: Unnecessary.

zjxxzjwang: Hello, if not here, how can the message expiration rate be updated

nodece (approved):
LGTM

dao-jun (approved):

jiazhai:
This is similar to PR https://github.com/apache/pulsar/pull/19759, but not sure how it was reverted

jiazhai:
Get the reverting information: https://github.com/apache/pulsar/pull/24073

jiazhai:
@zjxxzjwang  Would you please help add a unit test for this change following PR https://github.com/apache/pulsar/pull/19759. Seems this part of code is easy to be broken, it is better to protect it.

zjxxzjwang:
> @zjxxzjwang Would you please help add a unit test for this change following PR #19759. Seems this part of code is easy to be broken, it is better to protect it.

Good suggestion, I have also considered it before. But it seems that there are no good test examples to determine if this has not been broken.

zjxxzjwang:
@lhotari Hello, can you review it

nodece:
@zjxxzjwang Please rebase the master branch, and then I will merge this PR after CI passes.

zjxxzjwang:
> @zjxxzjwang Please rebase the master branch, and then I will merge this PR after CI passes.

hello , I have rebased the master branch and added test cases. Can you help me merge it

PR #24207, opened 2025-04-24
[fix][broker] Fix compaction service log's wrong condition

poorbarcode:
#### background
- This change is a follow-up on https://github.com/apache/pulsar/pull/23980, which added the log.
- If `CompactedTopicImpl.newCompactedLedger` was called more than once with the same `compactedLedgerId`, it will trigger a ledger deletion, which causes a data loss event.

#### Issue
- The condition of the log should be `previousLedger == @param-newLedger`, but it is `@param-newLedger == @param-newLedger`
- The impact is that there would be invalid error logs for every compaction

### Modifications

Fix the wrong condition of printing error log

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicImpl.java
@@ -84,13 +84,20 @@ public class CompactedTopicImpl implements CompactedTopic {
 
             // delete the ledger from the old context once the new one is open
             return compactedTopicContext.thenCompose(ctx -> {
-                if (ctx != null && ctx.getLedger() != null && ctx.getLedger().getId() == compactedLedgerId) {
-                    // Print an error log here, which is not expected.
-                    log.error("[__compaction] Using the same compacted ledger to override the old one, which is not"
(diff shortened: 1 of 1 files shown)


shibd (approved):

zymap (approved):

dao-jun (approved):

nodece (approved):

lhotari (commented):
The change itself is ok, but the title and description of the PR are not describing the relevant context of the change:
- The impact isn't described at all. The impact seems to be that there would be invalid error logs for every compaction. Is that correct?
- This change is a follow up on #23980 changes that haven't been released yet. It would be useful to mention this in the description that this is a follow up on #23980 changes.

poorbarcode:
@lhotari 

> The impact isn't described at all. The impact seems to be that there would be invalid error logs for every compaction. Is that correct?
> This change is a follow up on https://github.com/apache/pulsar/pull/23980 changes that haven't been released yet. It would be useful to mention this in the description that this is a follow up on https://github.com/apache/pulsar/pull/23980 changes.

Correct, Updated in the Motivation

lhotari (approved):
LGTM

PR #24200, opened 2025-04-23
[cleanup][test] Cleanup OneWayReplicatorUsingGlobalPartitionedTest and OneWayReplicatorUsingGlobalZKTest

nodece:
### Motivation

Both `OneWayReplicatorUsingGlobalPartitionedTest` and `OneWayReplicatorUsingGlobalZKTest` extend `OneWayReplicatorTest`. As a result, when a new test case is added to `OneWayReplicatorTest`, it is also executed in the `OneWayReplicatorUsingGlobalPartitionedTest` and `OneWayReplicatorUsingGlobalZKTest`. However, not all test cases are applicable to the global zk scenarios. This forces developers to explicitly disable certain tests in `OneWayReplicatorUsingGlobalPartitionedTest` and `OneWayReplicatorUsingGlobalZKTest`, leading to increased complexity and reduced maintainability.

testcases:
```
[root]
pulsar
OneWayReplicatorUsingGlobalPartitionedTest
OneWayReplicatorUsingGlobalPartitionedTest.testCloseTopicAfterStartReplicationFailed
OneWayReplicatorUsingGlobalPartitionedTest.testRemoveCluster
OneWayReplicatorUsingGlobalPartitionedTest.testPartitionedTopicWithTopicPolicyAndNoReplicationClusters
OneWayReplicatorUsingGlobalPartitionedTest.testCreateRemoteAdminFailed

[root]
pulsar
OneWayReplicatorUsingGlobalZKTest
OneWayReplicatorUsingGlobalZKTest.testCloseTopicAfterStartReplicationFailed
OneWayReplicatorUsingGlobalZKTest.testRemoveCluster
OneWayReplicatorUsingGlobalZKTest.testConfigReplicationStartAt
OneWayReplicatorUsingGlobalZKTest.testDeleteNonPartitionedTopic
OneWayReplicatorUsingGlobalZKTest.testDeletePartitionedTopic
OneWayReplicatorUsingGlobalZKTest.testCreateRemoteAdminFailed
OneWayReplicatorUsingGlobalZKTest.testPartitionedTopicWithTopicPolicyAndNoReplicationClusters
```

### Modifications

- Refactor `OneWayReplicatorUsingGlobalPartitionedTest` and `OneWayReplicatorUsingGlobalZKTest` to extend the base class: `OneWayReplicatorTestBase`.
- Mark the `OneWayReplicatorTestBase` class as `final` to prevent further extension.
[...]

PR #24211, opened 2025-04-24
[revert][broker] Revert #24200: Cleanup OneWayReplicatorUsingGlobalPartitionedTest and OneWayReplicatorUsingGlobalZKTest

poorbarcode:
### Motivation

See also https://github.com/apache/pulsar/pull/24200#issuecomment-2827963819

> Sorry, I did not completely review #24200. I assumed that you added new tests in `OneWayReplicatorUsingGlobalZKTest` and `OneWayReplicatorUsingGlobalPartitionedTest`, which makes #24200 meaningful. But the exact changes are that you just copied tests from the super class into the child classes. I will revert this PR, because it did nothing.

- `OneWayReplicatorTest` -- has sub classes -> `OneWayReplicatorUsingGlobalZKTest` and `OneWayReplicatorUsingGlobalPartitionedTest`, they shared some tests 
- Other tests in `OneWayReplicatorUsingGlobalZKTest` and `OneWayReplicatorUsingGlobalPartitionedTest` have not be implemented completely.

https://github.com/apache/pulsar/pull/24200 makes the three test classes independent, and copies the shared tests into `OneWayReplicatorUsingGlobalZKTest` and `OneWayReplicatorUsingGlobalPartitionedTest`, which is meaningless.

### Modifications
- Revert #24200
- I will finish the tests that has modifier `@Test(enabled = false)` one by one in the future, they are needed

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
@@ -105,7 +105,7 @@ import org.testng.annotations.Test;
 
 @Slf4j
 @Test(groups = "broker")
-public final class OneWayReplicatorTest extends OneWayReplicatorTestBase {
+public class OneWayReplicatorTest extends OneWayReplicatorTestBase {
 
     @Override
     @BeforeClass(alwaysRun = true, timeOut = 300000)
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalPartitionedTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalPartitionedTest.java
@@ -18,44 +18,31 @@
  */
 package org.apache.pulsar.broker.service;
 
-import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
-import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
-import java.lang.reflect.Field;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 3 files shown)


nodece:
Let me clarify the reasoning behind #24200 and why I chose to make the test classes independent.

In Pulsar, we support both local ZooKeeper and global ZooKeeper modes. While cross-AZ deployments may require global ZK, we generally try to avoid it because of the increased latency and operational complexity. Therefore, running the exact same test cases in both global and local ZK environments is not always necessary.

My intention was to allow targeted testing of global ZK behavior only when needed, by making `OneWayReplicatorUsingGlobalZKTest` and `OneWayReplicatorUsingGlobalPartitionedTest` fully independent from the base class. This avoids shared inheritance patterns where test cases in the base class are always executed, even when they aren't relevant to global ZK.

Currently, the pattern of adding tests in the parent class and disabling them in the subclasses is not ideal—it:
- Increases confusion for contributors.
- Makes test behavior unpredictable.
- And increases test time unnecessarily.

Instead, I believe each test class should include only the tests it truly needs. In this PR, I copied only the relevant ones into the subclasses(There are some duplicate codes present), so the test scope is explicit. This gives us:
- Better control over what gets tested under global ZK.
- Cleaner test separation.
- And avoids redundant execution.

If needed, I'm happy to further reduce or adjust the test coverage in the global ZK classes. But I believe the direction of separating tests by environment is more scalable and maintainable than relying on shared inheritance and selective disabling.

poorbarcode:
> My intention was to allow targeted testing of global ZK behavior only when needed, by making OneWayReplicatorUsingGlobalZKTest and OneWayReplicatorUsingGlobalPartitionedTest fully independent from the base class. This avoids shared inheritance patterns where test cases in the base class are always executed, even when they aren't relevant to global ZK.
> Currently, the pattern of adding tests in the parent class and disabling them in the subclasses is not ideal—it:
> - Increases confusion for contributors.
> - Makes test behavior unpredictable.
> - And increases test time unnecessarily.

> Instead, I believe each test class should include only the tests it truly needs. In this PR, I copied only the relevant ones into > the subclasses(There are some duplicate codes present), so the test scope is explicit. This gives us:
> - Better control over what gets tested under global ZK.
> - Cleaner test separation.
> - And avoids redundant execution.

- All the test cases are needed for the feature Global ZK, which I have mentioned https://github.com/apache/pulsar/pull/24200#issuecomment-2823692026. I will finish them in the future.
- Maintaining the inheritance relationship will increase maintainability and readability.

lhotari (approved):

PR #24213, opened 2025-04-25
[improve][pip] PIP-414: Enforce topic consistency check

nodece:
### Motivation

The changes from #24118 have been merged and should be documented in a Pulsar Improvement Proposal (PIP) to record these important updates.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-414.md
+++ b/pip/pip-414.md
@@ -10,7 +10,7 @@ scalability. Partitioned topic is backed by multiple internal non-partitioned to
 ## Motivation
 
 Apache Pulsar allows clients to produce to or consume from a fully qualified partitioned topic name (e.g.,
-`public/default/test-partitioned-topic-partition-N`), even if the topic metadata does not yet exist. If topic
+`public/default/test-partitioned-topic-partition-N`), even if the partitioned topic metadata does not yet exist. If topic
 auto-creation is enabled and `allowAutoTopicCreationType=non-partitioned` is set, the broker will automatically create a
 non-partitioned topic for that partition name.
 
--- a/pip/pip-414.md
+++ b/pip/pip-414.md
@@ -63,6 +63,15 @@ These errors help identify misconfigured clients.
 No special upgrade procedure is required. Once upgraded, clients/brokers attempting invalid topic-type interactions will
 receive an error.
 
+If you encounter the error `Partition metadata not found for the partitioned topic`, you can recreate the metadata using
+the following command:
+
+```
+bin/pulsar-admin topics create-partitioned-topic $TOPIC -p $PARITIONS
+```
+
+This behavior is supported by [#24225](https://github.com/apache/pulsar/pull/24225).
+
 ### Downgrade / Rollback
 
 On downgrade, previous behavior resumes: mismatched topic types may be silently created or accessed. Operators should
--- a/pip/pip-414.md
+++ b/pip/pip-414.md
@@ -97,5 +97,5 @@ This change increases Pulsar robustness and predictability by enforcing stricter
 ## Links
 
 * Implement PR: https://github.com/apache/pulsar/pull/24118
-* Mailing List discussion thread:
-* Mailing List voting thread:
+* Mailing List discussion thread: https://lists.apache.org/thread/v7v889qzfjpqznm713wvogmx2bs4pkkb
+* Mailing List voting thread: https://lists.apache.org/thread/92zpyqwbrfbjrsxbh1dsqksn65scsc5m
--- a/pip/pip-414.md
+++ b/pip/pip-414.md
@@ -1,0 +1,92 @@
+# PIP-414: Enforce topic consistency check
+
+## Background Knowledge
+
+In Apache Pulsar, topics can be either non-partitioned or partitioned. A non-partitioned topic handles all messages on a
+single topic instance, while a partitioned topic is logically split across multiple partitions for parallelism and
+scalability. Partitioned topic is backed by multiple internal non-partitioned topics, one per partition (e.g.,
+`topic-partition-0`, `topic-partition-1`, etc.).
+
+## Motivation
+
+Apache Pulsar allows clients to produce to or consume from a fully qualified partitioned topic name (e.g.,
+`public/default/test-partitioned-topic-partition-N`), even if the topic metadata does not yet exist. If topic
+auto-creation is enabled and `allowAutoTopicCreationType=non-partitioned` is set, the broker will automatically create a
+non-partitioned topic for that partition name.
+
+This behavior can lead to inconsistencies: users expecting to interact with a partitioned topic may unknowingly create
+non-partitioned topics for individual partitions. As a result, listing topics may yield unexpected structures, with
+partitioned and non-partitioned topics mixed under similar names.
+
+While this issue can also occur in GEO-replicated environments due to asynchronous metadata propagation, the core
+problem is the lack of topic validation. Disabling this behavior and enforcing
+strict topic-type consistency is essential to ensure clarity, prevent accidental misconfigurations, and maintain the
+integrity of topic structures in Pulsar.
+
+## Goals
+
+### In Scope
+
+- Introduce validation to enforce consistency between topic metadata and topic type.
+- Prevent producing or consuming from a topic when metadata indicates a mismatch.
+
+## High Level Design
+
+The proposed solution adds server-side validation to ensure that the topic type matches the metadata.
+
+- When accessing a non-partitioned topic, the broker checks that no partition metadata exists(which shouldn't be there).
+- When accessing a partitioned topic (via a `-partition-N` name), the broker ensures that valid partition metadata
+  exists and that the topic is indeed part of a partitioned group.
+
+- If these checks fail, the broker rejects the client request with an appropriate error.
+
+## Detailed Design
+
+### Design & Implementation Details
+
+- Validation is added when loading the topic.
+- If a mismatch is detected:
+    - Non-partitioned topic with partition metadata → reject request.
+    - Partitioned topic (e.g., `-partition-0`) with missing metadata → reject request.
+
+### Public-facing Changes
+
+## Monitoring
+
+Operators should monitor for client/broker errors indicating rejected produce/consume requests due to topic-type mismatches.
+These errors help identify misconfigured clients.
+
+## Backward & Forward Compatibility
+
+### Upgrade
+
+No special upgrade procedure is required. Once upgraded, clients/brokers attempting invalid topic-type interactions will
+receive an error.
+
+### Downgrade / Rollback
+
+On downgrade, previous behavior resumes: mismatched topic types may be silently created or accessed. Operators should
+monitor carefully if a downgrade is necessary.
+
+### Pulsar Geo-Replication Upgrade & Downgrade/Rollback Considerations
+
+The old version may allow inconsistent topic-type usage while newer ones enforce validation.
+
+# Alternatives
+
+@BewareMyPower proposed adding a configuration option that `validatePartitionMetadataConsistency` to allow enabling or
+disabling this validation behavior. While this would provide flexibility, we believe it is not the right direction.
+
+Pulsar already has an extensive set of configuration parameters. Introducing more options for narrowly scoped behaviors
+increases maintenance overhead and adds to user confusion. Since topic-type consistency is fundamental to correct Pulsar
+usage, this validation should be enforced by default, not made optional.
+
+## General Notes
+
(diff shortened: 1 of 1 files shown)


BewareMyPower (changes requested):
I agree not to add a new configuration. I proposed the option just to keep compatibility for old branches.

However, we still need to take care of the compatibility issues for upgrade.

For example, given a partitioned topic whose partition metadata is accidentally deleted.

```bash
> admin topics get-partitioned-topic-metadata my-topic
Topic persistent://public/default/my-topic not found

Reason: Topic persistent://public/default/my-topic not found
> admin topics list public/default
persistent://public/default/my-topic-partition-0
```

There is a reader reading messages directly from `my-topic-partition-0` and a producer 
sending messages to this partition.

After the upgrade, the producer and reader will not be able to serve this topic.

> These errors help identify misconfigured clients.

It's not misconfigured. It's on purpose. Especially for the case that some clients don't support multi-topics readers. Users have to create readers on partitions directly.

At least we need to recover the partition metadata for such orphan partitions. Currently it's impossible:

```bash
> admin topics create-partitioned-topic my-topic -p 1
2025-04-25T19:45:14,373+0800 [AsyncHttpClient-7-2] WARN  org.apache.pulsar.client.admin.internal.BaseResource - [http://localhost:8080/admin/v2/persistent/public/default/my-topic/partitions?createLocalTopicOnly=false] Failed to perform http put request: javax.ws.rs.ClientErrorException: HTTP 409 {"reason":"This topic already exists"}
This topic already exists
```

BewareMyPower:
Another approach is to create the missed partitioned metadata automatically. For example, with this PIP implemented, when the broker detects `my-topic-partition-0` is accessed by a client but there is no partition metadata for `my-topic`. The broker should try creating the partition metadata to unblock the client. The number of partitions are determined by the largest partition index `N` in `my-topic-partition-N`.

nodece:
> At least we need to recover the partition metadata for such orphan partitions. Currently it's impossible:
> 
> ```shell
> > admin topics create-partitioned-topic my-topic -p 1
> 2025-04-25T19:45:14,373+0800 [AsyncHttpClient-7-2] WARN  org.apache.pulsar.client.admin.internal.BaseResource - [http://localhost:8080/admin/v2/persistent/public/default/my-topic/partitions?createLocalTopicOnly=false] Failed to perform http put request: javax.ws.rs.ClientErrorException: HTTP 409 {"reason":"This topic already exists"}
> This topic already exists
> ```

Good catch, I can fix this bug. I think this command helps the user to recover the partitioned metadata.

> Another approach is to create the missed partitioned metadata automatically.

This is difficult because we can not obtain the maximum partition index.

BewareMyPower:
> This is difficult because we can not obtain the maximum partition index.

We can list all non-partitioned topics of the `<topic>-partition-<index>` pattern and find the maximum `index`.

nodece:
> > This is difficult because we can not obtain the maximum partition index.
> 
> We can list all non-partitioned topics of the `<topic>-partition-<index>` pattern and find the maximum `index`.

@BewareMyPower Do this during topic creation (e.g., via admin topics create-partitioned-topic) or at runtime?

crossoverJie:
> Another approach is to create the missed partitioned metadata automatically. For example, with this PIP implemented, when the broker detects `my-topic-partition-0` is accessed by a client but there is no partition metadata for `my-topic`. The broker should try creating the partition metadata to unblock the client. The number of partitions are determined by the largest partition index `N` in `my-topic-partition-N`.

I agree with this plan. Our cluster has been running for many years, so there are likely quite a few topics with metadata errors.

Perhaps we should provide an additional admin interface to query topics with metadata errors.

nodece:
> I agree with this plan. Our cluster has been running for many years, so there are likely quite a few topics with metadata errors.

@crossoverJie If the broker automatically creates the partitioned metadata, this may break the user behavior.

> Perhaps we should provide an additional admin interface to query topics with metadata errors.

If the topic type is incorrect, the user can use the create partitioned topic command to fix the metadata error.

BewareMyPower:
> Do this during topic creation (e.g., via admin topics create-partitioned-topic) or at runtime?

Topic creation, or `BrokerService#getTopic` on a non-partitioned topic whose name ends with `-partition-<index>`.

nodece:
> > Do this during topic creation (e.g., via admin topics create-partitioned-topic) or at runtime?
> 
> Topic creation, or `BrokerService#getTopic` on a non-partitioned topic whose name ends with `-partition-<index>`.


@BewareMyPower Topic creation is the best place to handle this. Doing it at runtime (e.g., in BrokerService#getTopic) would be too late and would introduce black-box behavior that's harder to reason about and debug.

dao-jun:
I don't think create missing topic partition metadata in runtime is OK, as @nodece says, it would introduce unpredictable consequences. Actually, I think add a new config is the relatively good way.

nodece:
> I don't think create missing topic partition metadata in runtime is OK, as @nodece says, it would introduce unpredictable consequences. Actually, I think add a new config is the relatively good way.

@dao-jun Instead of introducing a flag to enable or disable the creation of missing topic partition metadata at runtime, this feature will always be enabled. Since disabling this feature can lead to unexpected behavior and complexity, enabling it unconditionally simplifies the design and avoids confusion.

nodece:
> Then support creating missed partitions via admin API for such cases could be the solution. Currently the create-missed-partitions API does not work.

@BewareMyPower `admin topics create-partitioned-topic my-topic -p 1` has been fixed by #24225

Could you review this PIP again?

BewareMyPower (approved):

pip/pip-414.md
```diff
@@ -0,0 +1,101 @@
...
+
+In Apache Pulsar, topics can be either non-partitioned or partitioned. A non-partitioned topic handles all messages on a
+single topic instance, while a partitioned topic is logically split across multiple partitions for parallelism and
+scalability. Partitioned topic is backed by multiple internal non-partitioned topics, one per partition (e.g.,
+`topic-partition-0`, `topic-partition-1`, etc.).
+
+## Motivation
+
+Apache Pulsar allows clients to produce to or consume from a fully qualified partitioned topic name (e.g.,
+`public/default/test-partitioned-topic-partition-N`), even if the topic metadata does not yet exist. If topic
```
codelipenghui:
```suggestion
`public/default/test-partitioned-topic-partition-N`), even if the partitioned topic metadata does not yet exist. If topic
```

codelipenghui: Is it correct?

pip/pip-414.md
```diff
@@ -0,0 +1,101 @@
...
+
+- If these checks fail, the broker rejects the client request with an appropriate error.
+
+## Detailed Design
+
+### Design & Implementation Details
+
+- Validation is added when loading the topic.
+- If a mismatch is detected:
+    - Non-partitioned topic with partition metadata → reject request.
```
codelipenghui, line 49: How do we know the requested topic is non-partitioned? Is there a concrete example demonstrating this case?

From the broker’s perspective, it always fetches the partitioned topic metadata first. So it’s unclear what “non-partitioned topic with partition metadata” actually refers to here. Could you clarify this?

pip/pip-414.md
```diff
@@ -0,0 +1,101 @@
...
+    - Partitioned topic (e.g., `-partition-0`) with missing metadata → reject request.
```
codelipenghui, line 50: I believe this validation is already added.

I tried to build from the master branch and start a standalone.
Then, using the CLI try to publish messages

```
bin/pulsar-client produce -m "{}" test-partition-0
```

The error messages generated:

```
2025-06-16T12:16:25,670-0700 [main] ERROR org.apache.pulsar.client.cli.PulsarClientTool - Error while producing messages
2025-06-16T12:16:25,671-0700 [main] ERROR org.apache.pulsar.client.cli.PulsarClientTool - {"errorMsg":"Partition metadata not found for the partitioned topic: persistent://public/default/test-partition-0","reqId":1347957860605533513, "remote":"localhost/127.0.0.1:6650", "local":"/127.0.0.1:57439"}
org.apache.pulsar.client.api.PulsarClientException$NotAllowedException: {"errorMsg":"Partition metadata not found for the partitioned topic: persistent://public/default/test-partition-0","reqId":1347957860605533513, "remote":"localhost/127.0.0.1:6650", "local":"/127.0.0.1:57439"}
        at org.apache.pulsar.client.api.PulsarClientException.unwrap(PulsarClientException.java:1112) ~[org.apache.pulsar-pulsar-client-api-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at org.apache.pulsar.client.impl.ProducerBuilderImpl.create(ProducerBuilderImpl.java:89) ~[org.apache.pulsar-pulsar-client-original-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at org.apache.pulsar.client.cli.CmdProduce.publish(CmdProduce.java:308) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at org.apache.pulsar.client.cli.CmdProduce.run(CmdProduce.java:287) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at org.apache.pulsar.client.cli.AbstractCmd.call(AbstractCmd.java:27) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at org.apache.pulsar.client.cli.AbstractCmd.call(AbstractCmd.java:23) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at picocli.CommandLine.executeUserObject(CommandLine.java:2041) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at picocli.CommandLine.access$1500(CommandLine.java:148) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2453) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2415) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at picocli.CommandLine$RunLast.execute(CommandLine.java:2417) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at org.apache.pulsar.internal.CommanderFactory$CommandExecutionStrategy.execute(CommanderFactory.java:57) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at picocli.CommandLine.execute(CommandLine.java:2170) [info.picocli-picocli-4.7.5.jar:4.7.5]
        at org.apache.pulsar.client.cli.PulsarClientTool.run(PulsarClientTool.java:216) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at org.apache.pulsar.client.cli.PulsarClientTool.main(PulsarClientTool.java:234) [org.apache.pulsar-pulsar-client-tools-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
2025-06-16T12:16:25,678-0700 [main] INFO  org.apache.pulsar.client.cli.PulsarClientTool - 0 messages successfully produced
```

Could you please share the steps for reproducing this issue?

pip/pip-414.md
nodece: > From the broker’s perspective, it always fetches the partitioned topic metadata first. So it’s unclear what “non-partitioned topic with partition metadata” actually refers to here. Could you clarify this?

When the client provides a topic name that does not include the `-partition-N` suffix, but the topic does have partitioned metadata, it means the client is referring to the base topic of a partitioned topic. While this scenario is uncommon in normal operation, it's still important to handle it properly by checking for the presence of partitioned metadata.

pip/pip-414.md
nodece: This PIP has already been implemented in Pulsar, but it's important to formally document these changes.

codelipenghui: Oh I see.

codelipenghui (approved):

PR #24219, opened 2025-04-26
[improve][broker]Improve the feature "Optimize subscription seek (cursor reset) by timestamp": search less entries

poorbarcode:
### Motivation

**Background**
https://github.com/apache/pulsar/pull/22792 improved the behavior that searching messages by timestamp. For example:
There are ledgers as follows:

```json
[{
	"ledger": "1",
	"entries": "10",
	"closed_time": "1000"
},{
	"ledger": "2",
	"entries": "10",
	"closed_time": "1010"
},{
	"ledger": "3",
	"entries": "10",
	"closed_time": "3000"
}]
```

And the config `broker.conf -> managedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis` is `20ms`, when you search for timestamp `1000`, the broker will find ledgers whose timestamp is larger than `ledger.closed_time-20` and less than `ledger.closed_time+20`. Then the result range is `[1:0 ~ 3:9]`.

### Modifications

Since we have the configuration `broker.conf -> managedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis`, the entries `[1:0 ~ 1:8]` can be skipped to search, because we have already given enough scope expanded by `managedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis`, which improves the search.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageFinder.java
@@ -72,6 +72,7 @@ public class PersistentMessageFinder implements AsyncCallbacks.FindEntryCallback
                             ledgerCloseTimestampMaxClockSkewMillis);
             cursor.asyncFindNewestMatching(ManagedCursor.FindPositionConstraint.SearchAllAvailableEntries, entry -> {
                 try {
+                    // Find the latest entry that is earlier than the target timestamp.
                     long entryTimestamp = Commands.getEntryTimestamp(entry.getDataBuffer());
                     return MessageImpl.isEntryPublishedEarlierThan(entryTimestamp, timestamp);
                 } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


lhotari (commented):
I haven't reviewed this yet, but we might need more real world test cases to get confidence that we don't cause regressions. Previous optimization changes caused regression #23910 which was addressed by #23919. The optimization was different that time, so perhaps things are fine with this optimization.

merlimat (approved):
The change looks good. I'm this case it's safe to skip the first ledger in the search since all the entries would have a smaller timestamp. Including last entry is enough.

lhotari (approved):

Technoboy- (approved):

poorbarcode:
https://lists.apache.org/thread/jp1rpjqz9cczsro3b36t15z239zqfqmy Cherry-picked

PR #24220, opened 2025-04-27
[feat][pip]PIP-415: Support getting message ID by Index

liangyepianzhou:
### Motivation
 we can now obtain the offset of a message by its message id:

1. Get the message by id using `get-message-by-id` cmd
2. Get the index of the message using `Message.getIndex()`

But we cannot obtain the message id by offset. Then we need to add a new API to get the message id by offset.

### Modifications

Add a new http API to retrieve the message ID by offset.
We propose to add a new API to retrieve the message ID by offset, enabling us to cache the mapping between message ID and offset.
This will allow us to use offsets for seek and acknowledgment operations when consuming messages through the standardized API.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-415.md
+++ b/pip/pip-415.md
@@ -86,15 +86,33 @@ Add a new admin API endpoint `getMessageIdByIndex` to `org.apache.pulsar.broker.
 
 **notes: If the specified index points to a system message, return the first valid user message following this index.**
 
+When retrieving a message ID by index, the resolution is limited to the **entry** level (an entry is the minimal storage unit for messages in Pulsar's persistence layer). If message batching is enabled, a single entry may contain multiple messages with distinct indexes.
+
+**Example Scenario** (partition with 2 entries):
+
+| Entry | Ledger ID | Entry ID | Index | Messages |  
+| :--- | ---: | ---: | ---: | ---: |  
+| A | 0 | 0 | 2 | 0,1,2 |  
+| B | 0 | 1 | 4 | 3,4 |  
+
+- Param indexes 0,1,2 or 3,4 will return the **same MessageID** (e.g., `MessageId(0:0:*)` for Entry A).
+- To achieve **precise message-level consumption**, you can first locate the entry using the MessageID, then filter messages by their indexes during client-side processing.
+  - Example: To consume from index 1, retrieve all messages in Entry A and filter out messages with indexes below 1.
+
+**Why Precise Index Matching Isn't Implemented on the Broker Side:**
+1. **Metadata Parsing Constraints**: Precise matching requires parsing `MessageMetadata`, but brokers are designed to parse only `BrokerEntryMetadata`.
+2. **Limited Benefits**: Even if the exact `batchIndex` and `batchSize` are determined, the broker must still send the **entire entry** to the client, offering no network bandwidth savings.
+3. **Client-Side Efficiency**: Once the client receives the entry, users can directly filter messages by index (after parsing `MessageMetadata`, which includes unique indexes for each message).
+
 ##### Error Responses
 
-| Status Code                 | Description                                                                 | 
-|-----------------------------|-----------------------------------------------------------------------------|
-| **307 Temporary Redirect**  | The current broker does not own this topic. Redirect to the correct broker. | 
-| **403 Forbidden**           | Client lacks permissions to access the topic/namespace.                     |       
-| **404 Not Found**           | Topic/namespace does not exist, or the specified `index` is invalid.        |
-| **406 Not Acceptable**      | The topic is not a persistent topic.                                        |
-| **412 Precondition Failed** | The broker is not configured to enable broker entry metadata.               |
+| Status Code                 | Description                                                                                                                                                  | 
+|-----------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **307 Temporary Redirect**  | The current broker does not own this topic. Redirect to the correct broker.                                                                                  | 
+| **403 Forbidden**           | Client lacks permissions to access the topic/namespace.                                                                                                      |       
+| **404 Not Found**           | Topic/namespace does not exist, or the specified `index` is invalid. An index value of -1 or one that exceeds the maximum index range is considered invalid. |
+| **406 Not Acceptable**      | The topic is not a persistent topic.                                                                                                                         |
+| **412 Precondition Failed** | The broker is not configured to enable broker entry metadata.                                                                                                |
 
 #### Pulsar Admin
 Add two api in `pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java`
@@ -103,7 +121,15 @@ Add two api in `pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/a
      * Get the message id by index.
      * @param topicName either the specific partition name of a partitioned topic (e.g. my-topic-partition-0) or the original topic name for non-partitioned topics.
      * @param index the index of a message
-     * @return the message id of the message
+     * @return the message id of the message. 
+     * When retrieving a message ID by index, the resolution is limited to the **entry** level (an entry is the minimal storage unit for messages in Pulsar's persistence layer).
+     * If message batching is enabled, a single entry may contain multiple messages with distinct indexes.
+     * Example Scenario (partition with 2 entries):
+     * | Entry | Ledger ID | Entry ID | Index | Messages |
+     * | :--- | ---: | ---: | ---: | ---: |
+     * | A | 0 | 0 | 2 | 0,1,2 |
+     * | B | 0 | 1 | 4 | 3,4 |
+     * Param with indexes 0,1,2 or 3,4 will return the **same MessageID** (e.g., `MessageId(0:0:*)` for Entry A).
      */
     MessageId getMessageIdByIndex(String topicName, long index) throws PulsarAdminException;
 
@@ -112,6 +138,14 @@ Add two api in `pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/a
      * @param topicName either the specific partition name of a partitioned topic (e.g. my-topic-partition-0) or the original topic name for non-partitioned topics.
      * @param index the index of a message
      * @return the message id of the message
+     * When retrieving a message ID by index, the resolution is limited to the **entry** level (an entry is the minimal storage unit for messages in Pulsar's persistence layer).
+     * If message batching is enabled, a single entry may contain multiple messages with distinct indexes.
+     * Example Scenario (partition with 2 entries):
+     * | Entry | Ledger ID | Entry ID | Index | Messages |
+     * | :--- | ---: | ---: | ---: | ---: |
+     * | A | 0 | 0 | 2 | 0,1,2 |
+     * | B | 0 | 1 | 4 | 3,4 |
+     * Param with indexes 0,1,2 or 3,4 will return the **same MessageID** (e.g., `MessageId(0:0:*)` for Entry A).
      */
     CompletableFuture<MessageId> getMessageIdByIndexAsync(String topicName, long index);
 ```
--- a/pip/pip-415.md
+++ b/pip/pip-415.md
@@ -143,5 +143,5 @@ Add new subcommand to pulsar-admin:
 
 # Links
 
-* Mailing List discussion thread:
-* Mailing List voting thread:
+* Mailing List discussion thread: https://lists.apache.org/thread/7kmo3robyx74p81891h3q8f1cq5lomfv
+* Mailing List voting thread: https://lists.apache.org/thread/mn694h98nmdtddx4co9w0zo94jolzj68
--- a/pip/pip-415.md
+++ b/pip/pip-415.md
@@ -1,0 +1,112 @@
+# PIP-415: Support getting message ID by offset
+
+# Background knowledge
+
+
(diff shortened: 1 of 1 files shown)


StevenLuMT:
Motivation is written relatively simply. I feel that some index planning can be added. Currently, the index function is open ( brokerEntryMetadataInterceptors=org.apache.pulsar.common.intercept.AppendIndexMetadataInterceptor
), but some functions are actually imperfect, such as:
1. Get MessageId through index
2. Get the corresponding index through MessageId
This time, the function of getting MessageId through index will be improved first

pip/pip-415.md
```diff
@@ -0,0 +1,112 @@
...
+
+After these three works, we can now obtain the offset of a message by its message id:
+
+1. Get the message by id using `get-message-by-id` cmd
+2. Get the index of the message using `Message.getIndex()`
+
+But we cannot obtain the message id by offset. Then we need to add a new API to get the message id by offset.
+
+# Motivation
+
```
StevenLuMT, line 19: I think we should not blindly compare pulsar and Rocketmq. This is not convincing at all. I think we can look at it from our own perspective.


Motivation is written relatively simply. I feel that some index planning can be added. Currently, the index function is open ( brokerEntryMetadataInterceptors=org.apache.pulsar.common.intercept.AppendIndexMetadataInterceptor
), but some functions are actually imperfect, such as:

Get MessageId through index
Get the corresponding index through MessageId
This time, the function of getting MessageId through index will be improved first

pip/pip-415.md
```diff
@@ -0,0 +1,112 @@
...
+### Binary protocol
+
+### Configuration
+
+### CLI
+Add new subcommand to pulsar-admin:
+
+```bash
+    pulsar-admin topics get-message-id-by-offset \
```
BewareMyPower, line 87: Since we're using "index" as the concept, it's better to replace "offset" with "index" in all commands and methods.

pip/pip-415.md
```diff
@@ -0,0 +1,112 @@
...
+        internalGetMessageIDByOffsetAsync(offset, authoritative)
+                .thenAccept(asyncResponse::resume)
+                .exceptionally(ex -> {
+                    if (!isRedirectException(ex)) {
+                        log.error("[{}] Failed to get message id by offset for topic {}, partition id {}, offset {}",
+                                clientAppId(), topicName, offset, ex);
+                    }
+                    resumeAsyncResponseExceptionally(asyncResponse, ex);
+                    return null;
+                });
```
BewareMyPower, line 75:
```suggestion
```

No need to mention any implementation here.

In addition, you only mention the HTTP protocol here. It seems that you should add the client side API (`PulsarAdmin`) as well

liangyepianzhou: You are right. I have modified it according to your suggestion.

pip/pip-415.md
```diff
@@ -0,0 +1,152 @@
...
+  ```json
+  {
+    "ledgerId": 12345,
+    "entryId": 67890,
+    "partitionIndex": 0
+  }
+  ```  
+    - `ledgerId`: ID of the ledger where the message is stored.
+    - `entryId`: Unique identifier of the message within the ledger.
+    - `partitionIndex`: Partition index (relevant for partitioned topics; `-1` for non-partitioned topics).
```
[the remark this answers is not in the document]
liangyepianzhou: @BewareMyPower Thank you for sharing your thoughts.

>Should you add the batchIndex and batchSize to the response body? 

I do not believe we should return `batchSize` and `batchIndex`. Currently, the `index` can only pinpoint the granularity of an **entry**. If a user enables batch processing and an entry contains multiple messages, we would still convert the `index` to a `messageId` for response. Users can only consume messages at the **entry level** (if they use the `index` to locate messages). If they require precise consumption based on individual messages, they should implement filtering in their business logic. Returning `batchIndex` and `batchSize` likely adds no value.  

>BTW, the topicName should already include the partition index, so here the partitionIndex might be redundant.

The Pulsar admin API returns a `messageId`, which is minimally defined by the triplet: `ledgerId`, `entryId`, and `partitionId`. While `partitionIndex` is technically redundant (since the topic name already includes the partition index), omitting it might confuse users. Without returning `partitionIndex`, users retrieving it from the `messageId` would only get `0` or `-1`, which could lead to ambiguity. For UX clarity, retaining `partitionIndex` is justified despite its redundancy.

pip/pip-415.md
```diff
@@ -0,0 +1,152 @@
...
+#### Pulsar Admin
+Add two api in `pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java`
+```java
+    /**
+     * Get the message id by index.
+     * @param topicName either the specific partition name of a partitioned topic (e.g. my-topic-partition-0) or the original topic name for non-partitioned topics.
+     * @param index the index of a message
+     * @return the message id of the message
+     */
+    MessageId getMessageIdByIndex(String topicName, long index) throws PulsarAdminException;
```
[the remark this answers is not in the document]
liangyepianzhou: From a logical standpoint, users should already have the **topic name**, **partition index**, and **index** upfront before using these three components to retrieve the `messageId`. This means they *already possess the partition index* and do not need to extract it from the response. The reason I retain `partitionIndex` in the response is for **UX clarity**, specifically to assist in constructing the client-side `messageId`.  As mentioned in my other comment here: https://github.com/apache/pulsar/pull/24220#discussion_r2092242630.

pip/pip-415.md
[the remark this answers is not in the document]
BewareMyPower: It's okay to return the entry's position. But `MessageId` should represent the position of a "message" rather than an "entry". You'd better explain the following behavior in the API docs and this proposal:

Given 2 entries:
| entry | ledger id | entry id | index | messages |
| :- | -: | -: | -: | -: |
| A | 0 | 0 | 3 | 0,1,2 |
| B | 0 | 1 | 5 | 3,4 |

You should tell users that the same `MessageId` will be returned for indexes 0,1,2 or 3,4.

In addition, there are corner cases that are not considered. What will `getMessageIdByIndex` return when the index is -1 or 5?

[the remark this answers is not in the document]
liangyepianzhou: Thank you for your suggestions. I agree that users should be explicitly alerted to these details.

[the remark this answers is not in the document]
liangyepianzhou: **Updated.**  
> In addition, there are corner cases that are not considered. What will `getMessageIdByIndex` return when the index is `-1` or `5`?  

Such cases should be considered **invalid indexes**, and the API should **return a 404 error**.  
The previous version did not explicitly define invalid indexes; this update formalizes the definition:  
`An index value of -1 or one that exceeds the maximum index range is considered invalid.`

pip/pip-415.md
```diff
@@ -0,0 +1,186 @@
...
+     * When retrieving a message ID by index, the resolution is limited to the **entry** level (an entry is the minimal storage unit for messages in Pulsar's persistence layer).
+     * If message batching is enabled, a single entry may contain multiple messages with distinct indexes.
+     * Example Scenario (partition with 2 entries):
+     * | Entry | Ledger ID | Entry ID | Index | Messages |
+     * | :--- | ---: | ---: | ---: | ---: |
+     * | A | 0 | 0 | 2 | 0,1,2 |
+     * | B | 0 | 1 | 4 | 3,4 |
+     * Param with indexes 0,1,2 or 3,4 will return the **same MessageID** (e.g., `MessageId(0:0:*)` for Entry A).
+     */
+    CompletableFuture<MessageId> getMessageIdByIndexAsync(String topicName, long index);
```
BewareMyPower, line 150: We need to add a `@throws` here to indicate that when the index is invalid, the future will complete exceptionally with `PulsarAdminException#NotFoundException`.

pip/pip-415.md
```diff
@@ -0,0 +1,186 @@
...
+3. **Client-Side Efficiency**: Once the client receives the entry, users can directly filter messages by index (after parsing `MessageMetadata`, which includes unique indexes for each message).
+
+##### Error Responses
+
+| Status Code                 | Description                                                                                                                                                  | 
+|-----------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **307 Temporary Redirect**  | The current broker does not own this topic. Redirect to the correct broker.                                                                                  | 
+| **403 Forbidden**           | Client lacks permissions to access the topic/namespace.                                                                                                      |       
+| **404 Not Found**           | Topic/namespace does not exist, or the specified `index` is invalid. An index value of -1 or one that exceeds the maximum index range is considered invalid. |
```
BewareMyPower, line 113: It's not correct. For example, if old messages are trimmed (e.g. message 0,1,2), the left messages might be 4,5, then the index where `index < 4` will be invalid.

In addition, this REST API should only work when `BrokerEntryMetadataInterceptor` is configured and `exposingBrokerEntryMetadataToClientEnabled` is true, right? IMO, in this case, we should return a 500 Internal Server Error

liangyepianzhou: > That's not accurate. For example, if old messages are trimmed (e.g., messages 0, 1, 2), the remaining messages might be 4, 5, making any index < 4 invalid.  

I hold a different view. If old messages are trimmed, the system should return the first valid message. Similarly, if an index points to a system marker, it should return the first valid message after that marker.  

> In addition, this REST API should only function when the `BrokerEntryMetadataInterceptor` is configured and `exposingBrokerEntryMetadataToClientEnabled` is true, correct? IMO, in such cases, we should return a 500 Internal Server Error.  

Correct. The server must enable `BrokerEntryMetadataInterceptor`, but for this specific API, `exposingBrokerEntryMetadataToClientEnabled` might not be strictly mandatory. Although, in practice, these two configurations are often required together.  

For such cases, I opted to return **412 Precondition Failed**. I deliberated extensively between 412 and 500 but ultimately concluded that 412 is more appropriate. This is not a server-side error but rather a failure to meet request prerequisites—the request was sent to a broker cluster lacking the required configuration. Thus, it qualifies as a precondition error.

pip/pip-415.md
liangyepianzhou: According to convention, we typically do not include @throws in Javadoc comments for asynchronous methods. Perhaps using @implNote to clarify could be an alternative approach.

pip/pip-415.md
BewareMyPower: > If old messages are trimmed, the system should return the first valid message.

Then if there are two `getMessageIdByIndex` calls with the same index argument, these two calls could both succeed but return different results, which could be confusing to users.

liangyepianzhou: >Then if there are two getMessageIdByIndex calls with the same index argument, these two calls could both succeed but return different results, which could be confusing to users.

Let me explain why returning the first valid message for expired/indexed-out data is acceptable in practical scenarios:

**Primary Application Scenarios**  
1. **Management Scenarios** - Users query message IDs or messages via frontend interfaces.  
2. **Data Consumption Scenarios** - Clients retrieve message IDs by index to locate and consume messages.  

**Behavior Analysis**  
*For expired data returning the first valid message:*  
1. **Management Scenarios**:  
   - Users should be explicitly notified that the API returns the first valid message corresponding to the requested index.  
   - When the index points to system markers or expired messages, a discrepancy between the query index and the actual message index may occur. Clear documentation/UI warnings would prevent confusion.  

2. **Data Consumption Scenarios**:  
   - Clients can directly consume from a valid message position.  
   - If the returned message index differs from the requested index, clients may log warnings/debug messages without disrupting normal operations.  

*Alternative approach (error reporting):*  
1. **Management Scenarios**:  
   - Still requires explaining the "first valid message" semantics. Additional error handling complexity arises for expired indices.  
2. **Data Consumption Scenarios**:  
   - API errors for expired messages would force clients to handle exceptions, re-query starting positions, or consume from earliest. Unhandled errors could crash applications.  

**Conclusion**  
The semantics of this API inherently resolve to *"return the first valid message"* due to the existence of system markers. Returning 404 for expired messages introduces unnecessary complexity without tangible benefits. A documented behavior (auto-forwarding to first valid message) maintains usability for both management and consumption scenarios while avoiding error-handling overhead.

BewareMyPower: The `MessageId` is mainly used for `seek` as the proposal documents

> We propose to add a new API to retrieve the message ID by index, enabling us to cache the mapping between message ID and index.
> This will allow us to use index for seek and acknowledgment operations when consuming messages through the standardized API.

So the cache might be:
- t0: 0 => (0, 0), 1 => (0, 1), 2 => (1, 0) 
- t1: 0 => (0, 1), 1 => (0, 1), 2 => (1, 0)
- t2: 0 => (1, 0), 1 => (1, 0), 2 => (1, 0)

Returning the first valid message's position cannot differ if the message of the index has been removed.

If you care much about the seek case, returning `MessageId.earliest` could achieve the same goal, where it can also tell users the message of the given index has been removed.

- t0: 0 => (0, 0), 1 => (0, 1), 2 => (1, 0) 
- t1: 0 => earliest, 1 => (0, 1), 2 => (1, 0)
- t2: 0 => earliest, 1 => earliest, 2 => (1, 0)

liangyepianzhou: Good suggestion!

BewareMyPower (approved):

pip/pip-415.md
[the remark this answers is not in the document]
StevenLuMT: This situation of batch can indeed be supported, but a batch returns the same offset, and the engine also conforms to the semantics of batch granularity, which is a good suggestion

PR #23078, opened 2024-07-26
[fix][io] Upgrade Debezium connector version to 2.6.1.Final

mukesh154:
Fixes #22626

### Motivation
To Upgrade debezium version to 2.6.1.Final for all source connectors.

It will address [this comment](https://github.com/apache/pulsar/pull/22813#discussion_r1634194181) as well.

### Modifications
Upgraded debezium version from 1.9.7.Final to 2.6.1.Final all debezium source connectors which includes mysql, postgresql, mongodb, mssql & oracle.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [x] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

Check the box below or label this PR directly (if you have committer privilege).

Need to update docs? 

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR #24221, opened 2025-04-27
[improve][io] support kafka connect transforms and predicates

efcasado:
### Motivation

The current implementation of the Kafka Connect adaptor in Pulsar IO does not support [Kafka Connect transforms](https://github.com/apache/kafka/tree/3.8/connect/transforms/src/main/java/org/apache/kafka/connect/transforms) and [predicates](https://github.com/apache/kafka/tree/3.8/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/predicates). This limits the flexibility and potential of using Kafka Connect-based source connectors, such as the Debezium PostgreSQL source connector, which [rely heavily on transformations](https://debezium.io/documentation/reference/1.9/transformations/index.html) to modify, filter, or enrich records before they are ingested. By adding support for transforms, we enable users to fully leverage these connectors' capabilities, sparing them the need to deploy additional Pulsar Functions to perform simple message-level transformations.

### Modifications

Extended `KafkaConnectSource` with three additional steps: `initPredicates`, `initTransforms` and `applyTransforms`.

- `initPredicates` and `initTransforms` are called during the initialization of the Kafka Connect adaptor and are responsible for setting up any configured predicates and transforms.

- `applyTransforms` is executed during `processRecords` and applies all available transforms to each source record before it is processed.

`kafka-connect-adaptor` has been updated to include the `connect-transforms` package.

### Verifying this change

- Extended the existing `KafkaConnectSourceTest` test to verify that transforms can be initialized and applied

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-io/kafka-connect-adaptor/pom.xml
+++ b/pulsar-io/kafka-connect-adaptor/pom.xml
@@ -110,6 +110,18 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>connect-transforms</artifactId>
+      <version>${kafka-client.version}</version>
+      <exclusions>
+        <exclusion>
+          <artifactId>jose4j</artifactId>
+          <groupId>org.bitbucket.b_c</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
     <!-- pulsar-client is only needed for MessageId conversion (for seeking), commons-lang3 and Netty buffer manipulation -->
     <dependency>
       <groupId>${project.groupId}</groupId>
@@ -156,19 +168,6 @@
     <dependency>
       <groupId>org.apache.kafka</groupId>
       <artifactId>connect-file</artifactId>
-      <version>${kafka-client.version}</version>
-      <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <artifactId>jose4j</artifactId>
-          <groupId>org.bitbucket.b_c</groupId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.kafka</groupId>
-      <artifactId>connect-transforms</artifactId>
       <version>${kafka-client.version}</version>
       <scope>test</scope>
       <exclusions>
--- a/pulsar-io/kafka-connect-adaptor/pom.xml
+++ b/pulsar-io/kafka-connect-adaptor/pom.xml
@@ -167,6 +167,19 @@
     </dependency>
 
     <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>connect-transforms</artifactId>
+      <version>${kafka-client.version}</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <artifactId>jose4j</artifactId>
+          <groupId>org.bitbucket.b_c</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>testmocks</artifactId>
       <version>${project.version}</version>
--- a/pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
+++ b/pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
@@ -30,13 +30,12 @@ import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.json.JsonConverterConfig;
 import org.apache.kafka.connect.source.SourceRecord;
-import org.apache.kafka.connect.transforms.predicates.Predicate;
 import org.apache.kafka.connect.transforms.Transformation;
+import org.apache.kafka.connect.transforms.predicates.Predicate;
 import org.apache.pulsar.client.api.Schema;
 import org.apache.pulsar.common.schema.KeyValue;
 import org.apache.pulsar.common.schema.KeyValueEncodingType;
@@ -58,7 +57,7 @@ public class KafkaConnectSource extends AbstractKafkaConnectSource<KeyValue<byte
     private static final String JSON_WITH_ENVELOPE_CONFIG = "json-with-envelope";
 
     private Map<String, Predicate<SourceRecord>> predicates = new HashMap<>();
-    
+
     private record PredicatedTransform(
             Predicate<SourceRecord> predicate,
             Transformation<SourceRecord> transform,
@@ -177,11 +176,11 @@ public class KafkaConnectSource extends AbstractKafkaConnectSource<KeyValue<byte
             if (current == null) {
                 break;
             }
-            
+
             if (pt.predicate != null && (pt.negated != pt.predicate.test(current))) {
                 continue;
             }
-            
+
             current = pt.transform.apply(current);
         }
         return current;
--- a/pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
+++ b/pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
@@ -178,7 +178,7 @@ public class KafkaConnectSource extends AbstractKafkaConnectSource<KeyValue<byte
                 break;
             }
 
-            if (pt.predicate != null && (pt.negated != pt.predicate.test(current))) {
+            if (pt.predicate != null && !(pt.negated != pt.predicate.test(current))) {
                 continue;
             }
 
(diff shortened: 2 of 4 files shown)


lhotari:
Thanks for the contribution, @efcasado!

lhotari (approved):
LGTM

lhotari:
> Connectors that rely on transforms, such as Debezium, typically package their own set of Kafka Connect transforms, making it unnecessary to include them in the adaptor itself.

@efcasado Are you also planning to contribute [to the Pulsar IO Debezium connectors](https://github.com/apache/pulsar/tree/master/pulsar-io/debezium)? 

Debezium version is outdated in Pulsar:
https://github.com/apache/pulsar/blob/a4a34091cbf58bfa080b42c101bd5094d82c41f3/pom.xml#L204 . 
It would be great to get that updated too. Would you be interested in handling that too? Previous upgrading attempt was https://github.com/apache/pulsar/pull/23078, but that's stalled.

lhotari:
@efcasado In SMT, there's also [predicates](https://docs.confluent.io/kafka-connectors/transforms/current/filter-ak.html). Do you have intention to add support for that too?

pulsar-io/kafka-connect-adaptor/pom.xml
```diff
@@ -166,6 +166,19 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>connect-transforms</artifactId>
+      <version>${kafka-client.version}</version>
+      <scope>test</scope>
```
lhotari, line 173: It seems that we should include this in default scope so that default transformations (and later predicates) would be available.

lhotari: The jar file is just over 100kB, so it won't add much overhead to the current connectors.

efcasado: Makes sense. I was overly defensive, it seems 😄

> The connect-transforms package has only been added as a test dependency. This is to keep the Kafka Connect adaptor bundle as lightweight as possible. Connectors that rely on transforms, such as Debezium, typically package their own set of Kafka Connect transforms, making it unnecessary to include them in the adaptor itself.

I will take care of this shortly. Thanks for taking the time to review this change! 🙏

efcasado:
> @efcasado In SMT, there's also [predicates](https://kafka.apache.org/documentation/#connect_predicates) (Confluent doc for [predicates](https://docs.confluent.io/kafka-connectors/transforms/current/filter-ak.html)). Do you have intention to add support for that too?

I guess it makes sense to include them in this `PR` since they are somewhat related. I can give it a try unless you tell me to only do this in a separate `PR`.

pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
```diff
@@ -95,8 +134,17 @@ private void initTransforms(Map<String, Object> config) {
...
+                    if (predicateName != null) {
+                        predicate = predicates.get(predicateName);
+                        if (predicate == null) {
+                            log.warn("Transform {} references non-existent predicate: {}",
+                                    transformName, predicateName);
+                        }
+                    }
                     transform.configure(transformConfig);
-                    transformations.add(transform);
+                    transformations.add(Pair.of(predicate, transform));
```
lhotari, line 147: In addition, there would be a need to support `negate` as documented [in the SMT predicate docs](https://kafka.apache.org/documentation/#connect_predicates): "The transformation's implicit `negate` config properties allows us to invert the set of records which a predicate matches.".

Instead of a `Pair`, it's better to use a Java record to hold information.

lhotari (commented):
Good work! Added a comment about supporting `negate` for predicates. It would be great to have some unit test coverage too.

pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
efcasado: Thanks for holding hands with me. As an Elixir developer who hasn't touched Java in years, I genuinely appreciate the time you are investing on this :pray: I should have something ready soon, and I will also add a couple of unit tests to cover predicates and negated predicates.

pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
```diff
@@ -135,6 +141,7 @@ private void initTransforms(Map<String, Object> config) {
                         ));
                     log.info("transform config: {}", transformConfig);
                     String predicateName = (String) transformConfig.get("predicate");
+                    boolean negated = (boolean) transformConfig.getOrDefault("negated", false);
```
lhotari, line 144: Just wondering if it would be useful to support cases where the value type of the property would be a `String` at runtime.

Perhaps something like this:
```suggestion
                    boolean negated = Boolean.parseBoolean(String.valueOf(transformConfig.getOrDefault("negated", "false")));
```

Not beautiful code, but would address all cases.

pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
lhotari: Good work so far!

pulsar-io/kafka-connect-adaptor/src/test/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSourceTest.java
```diff
@@ -101,6 +104,132 @@ public void testOpenAndReadTaskDirect() throws Exception {
         testOpenAndReadTask(config);
     }
 
+    @Test
+    void testCastTransformation() throws Exception{
```
efcasado, line 108: There is a fair amount of duplication in these tests. I will address this shortly.

efcasado:
@lhotari I believe I addressed all the above points and the change is ready for another round of reviews 😊

pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
```diff
@@ -60,17 +74,118 @@ public void open(Map<String, Object> config, SourceContext sourceContext) throws
...
+    }
+
+    private static final AvroData avroData = new AvroData(1000);
 
     public synchronized KafkaSourceRecord processSourceRecord(final SourceRecord srcRecord) {
-        KafkaSourceRecord record = new KafkaSourceRecord(srcRecord);
-        offsetWriter.offset(srcRecord.sourcePartition(), srcRecord.sourceOffset());
+        SourceRecord transformedRecord = applyTransforms(srcRecord);
+        if (transformedRecord == null) {
+            return null;
```
lhotari, line 167: just wondering if the `offsetWriter.offset` method should be called before returning

lhotari:
> @lhotari I believe I addressed all the above points and the change is ready for another round of reviews 😊

@efcasado great work! one more question about skipping source records.

pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/KafkaConnectSource.java
```diff
@@ -60,17 +74,118 @@ public void open(Map<String, Object> config, SourceContext sourceContext) throws
...
+        }
+        KafkaSourceRecord record = new KafkaSourceRecord(transformedRecord);
+        offsetWriter.offset(transformedRecord.sourcePartition(), transformedRecord.sourceOffset());
```
lhotari, line 170: would it better to keep this as `offsetWriter.offset(srcRecord.sourcePartition(), srcRecord.sourceOffset())` ?

efcasado: Makes sense. Otherwise, it may result in unnecessary re-deliveries of messages that will anyways not be ingested. I'll address this shortly.

lhotari (approved):
LGTM

lhotari:
Great work on this PR @efcasado! Merging this soon.

> Actually, that's how this `PR` started but I ended up opting for the path of least resistance 😅 I got a version of Debezium 3.1 partially working, but I ran into some issues I could not fix. I will be very happy to give it another try in the coming days.

Looking forward to future contributions!

efcasado:
@lhotari Will this change be cherry-picked for `3.x` too? We are running Pulsar 3.3.5 and we are wondering if we should just use our own fork or wait for you and the team to promote this change to the `3.x` branch.

lhotari:
> @lhotari Will this change be cherry-picked for `3.x` too? We are running Pulsar 3.3.5 and we are wondering if we should just use our own fork or wait for you and the team to promote this change to the `3.x` branch.

I don't see barriers in cherry-picking to 3.3.x since this isn't introducing breaking changes. I'll mark it for 3.0.x and 3.3.x too.

lhotari:
@efcasado FYI, There's an open issue #24646 related to this PR. Being fixed in #24654

PR #24222, opened 2025-04-27
[feat][admin] PIP-415: Support getting message ID by index

liangyepianzhou:
PIP:https://github.com/apache/pulsar/pull/24220

### Motivation
 we can now obtain the offset of a message by its message id:

1. Get the message by id using `get-message-by-id` cmd
2. Get the index of the message using `Message.getIndex()`

But we cannot obtain the message id by offset. Then we need to add a new API to get the message id by offset.

### Modifications

Add a new http API to retrieve the message ID by offset.
We propose to add a new API to retrieve the message ID by offset, enabling us to cache the mapping between message ID and offset.
This will allow us to use offsets for seek and acknowledgment operations when consuming messages through the standardized API.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5533,8 +5533,7 @@ public class PersistentTopicsBase extends AdminResource {
                     ManagedLedger managedLedger = persistentTopic.getManagedLedger();
                     Position lastPosition = managedLedger.getLastConfirmedEntry();
                     Position firstPosition = managedLedger.getFirstPosition();
-                    if (firstPosition == null || lastPosition == null ||
-                            firstPosition.equals(lastPosition)) {
+                    if (firstPosition == null || lastPosition == null || firstPosition.equals(lastPosition)) {
                         return FutureUtil.failedFuture(new RestException(Status.NOT_FOUND,
                                 "No messages found in topic " + topicName));
                     }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5489,4 +5489,53 @@ public class PersistentTopicsBase extends AdminResource {
                             return null;
                         }));
     }
+
+    protected CompletableFuture<MessageId> internalGetMessageIDByOffsetAsync(Long offset, boolean authoritative) {
+        int partitionIndex = topicName.getPartitionIndex();
+        CompletableFuture<Void> future = validateTopicOperationAsync(topicName, TopicOperation.PEEK_MESSAGES);
+        return future.thenCompose(__ -> {
+                    if (topicName.isGlobal()) {
+                        return validateGlobalNamespaceOwnershipAsync(namespaceName);
+                    } else {
+                        return CompletableFuture.completedFuture(null);
+                    }
+                }).thenCompose(__ -> {
+                    if (topicName.isPartitioned()) {
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        return getPartitionedTopicMetadataAsync(topicName, authoritative, false)
+                                .thenAccept(topicMetadata -> {
+                                    if (topicMetadata.partitions > 0) {
+                                        log.warn("[{}] Not supported getMessageIdByOffset operation on "
+                                                        + "partitioned-topic {}", clientAppId(), topicName);
+                                        throw new RestException(Status.METHOD_NOT_ALLOWED,
+                                                "GetMessageIDByOffset is not allowed on partitioned-topic");
+                                    }
+                                });
+                    }
+                }).thenCompose(ignore -> validateTopicOwnershipAsync(topicName, authoritative))
+                .thenCompose(__ -> getTopicReferenceAsync(topicName))
+                .thenCompose(topic -> {
+            if (!(topic instanceof PersistentTopic persistentTopic)) {
+                log.error("[{}] Get message id by offset on a non-persistent topic {} is not allowed",
+                        clientAppId(), topicName);
+                return FutureUtil.failedFuture(new RestException(Status.METHOD_NOT_ALLOWED,
+                        "Get message id by offset on a non-persistent topic is not allowed"));
+            }
+            ManagedLedger managedLedger = persistentTopic.getManagedLedger();
+            return managedLedger.asyncFindPosition(entry -> {
+                try {
+                    BrokerEntryMetadata brokerEntryMetadata =
+                            Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+                    return brokerEntryMetadata.getIndex() < offset;
+                } catch (Exception e) {
+                    log.error("Error deserialize message for message position find", e);
+                } finally {
+                    entry.release();
+                }
+                return false;
+            });
+        }).thenCompose(position -> CompletableFuture
+                .completedFuture(new MessageIdImpl(position.getLedgerId(), position.getEntryId(), partitionIndex)));
+    }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5493,7 +5493,7 @@ public class PersistentTopicsBase extends AdminResource {
     protected CompletableFuture<MessageId> internalGetMessageIDByIndexAsync(Long offset, boolean authoritative) {
         if (pulsar().getBrokerService().isBrokerEntryMetadataEnabled()) {
             return FutureUtil.failedFuture(new RestException(Status.PRECONDITION_FAILED,
-                    "GetMessageIDByIndex is not allowed when broker entry metadata is enabled"));
+                    "GetMessageIDByIndex is not allowed when broker entry metadata is disabled"));
         }
         int partitionIndex = topicName.getPartitionIndex();
         CompletableFuture<Void> future = validateTopicOperationAsync(topicName, TopicOperation.PEEK_MESSAGES);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5531,8 +5531,15 @@ public class PersistentTopicsBase extends AdminResource {
                                 "Get message id by index on a non-persistent topic is not allowed"));
                     }
                     ManagedLedger managedLedger = persistentTopic.getManagedLedger();
+                    Position lastPosition = managedLedger.getLastConfirmedEntry();
+                    Position firstPosition = managedLedger.getFirstPosition();
+                    if (firstPosition == null || lastPosition == null ||
+                            firstPosition.equals(lastPosition)) {
+                        return FutureUtil.failedFuture(new RestException(Status.NOT_FOUND,
+                                "No messages found in topic " + topicName));
+                    }
                     return findMessageIndexByPosition(
-                            PositionFactory.create(managedLedger.getFirstPosition().getLedgerId(), 0),
+                            PositionFactory.create(firstPosition.getLedgerId(), 0),
                             managedLedger)
                             .thenCompose(firstIndex -> {
                                 if (index < firstIndex) {
@@ -5557,8 +5564,7 @@ public class PersistentTopicsBase extends AdminResource {
                                     });
                                 }
                             }).thenCompose(position -> {
-                                Position lastPosition = managedLedger.getLastConfirmedEntry();
-                                if (position == null || position.compareTo(lastPosition) > 0) {
+                                if (position.compareTo(lastPosition) > 0) {
                                     return FutureUtil.failedFuture(new RestException(Status.NOT_FOUND,
                                             "Message not found for index " + index));
                                 } else {
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5540,19 +5540,18 @@ public class PersistentTopicsBase extends AdminResource {
                                 } else {
                                     return managedLedger.asyncFindPosition(entry -> {
                                         try {
-                                            BrokerEntryMetadata brokerEntryMetadata =
-                                                    Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
-                                            // Skip messages without index
-                                            if (brokerEntryMetadata == null) {
-                                                return true;
+                                            Long messageIndex = getIndexAndRelease(entry);
+                                            if (messageIndex == null) {
+                                                return false; // Skip messages without index
+                                            } else {
+                                                // If the message index is less than the requested index,
+                                                // we continue searching
+                                                return messageIndex < index;
                                             }
-                                            return brokerEntryMetadata.getIndex() < index;
-                                        } catch (Exception e) {
+                                        } catch (IOException e) {
                                             log.error("Error deserialize message for message position find", e);
-                                        } finally {
-                                            entry.release();
+                                            return false;
                                         }
-                                        return false;
                                     });
                                 }
                             }).thenCompose(position -> {
@@ -5573,19 +5572,20 @@ public class PersistentTopicsBase extends AdminResource {
         managedLedger.asyncReadEntry(position, new AsyncCallbacks.ReadEntryCallback() {
             @Override
             public void readEntryComplete(Entry entry, Object ctx) {
-                BrokerEntryMetadata brokerEntryMetadata =
-                        Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
-                if (brokerEntryMetadata == null) {
-                    indexFuture.completeExceptionally(new RestException(Status.PRECONDITION_FAILED,
-                            "Broker entry metadata is not present in the message"));
-                } else {
-                    long index = brokerEntryMetadata.getIndex();
-                    if (index < 0) {
+                try {
+                    Long index = getIndexAndRelease(entry);
+                    if (index == null) {
+                        indexFuture.completeExceptionally(new RestException(Status.PRECONDITION_FAILED,
+                                "Broker entry metadata is not present in the message"));
+                    } else if (index < 0) {
                         indexFuture.completeExceptionally(new RestException(Status.PRECONDITION_FAILED,
                                 "Invalid message index: " + index));
                     } else {
                         indexFuture.complete(index);
                     }
+                } catch (IOException e) {
+                    indexFuture.completeExceptionally(new RestException(Status.INTERNAL_SERVER_ERROR,
+                            "Failed to get index from entry: " + e.getMessage()));
                 }
             }
 
@@ -5599,4 +5599,19 @@ public class PersistentTopicsBase extends AdminResource {
         return indexFuture;
     }
 
+
+    private static Long getIndexAndRelease(Entry entry) throws IOException {
+        try {
+            final var brokerEntryMetadata = Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+            if (brokerEntryMetadata != null && brokerEntryMetadata.hasIndex()) {
+                return brokerEntryMetadata.getIndex();
+            } else {
+                return null;
+            }
+        } catch (Throwable throwable) {
+            throw new IOException(throwable);
+        } finally {
+            entry.release();
+        }
+    }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5555,7 +5555,7 @@ public class PersistentTopicsBase extends AdminResource {
                                                 // we continue searching
                                                 return messageIndex < index;
                                             }
-                                        } catch (IOException e) {
+                                        } catch (Throwable e) {
                                             log.error("Error deserialize message for message position find", e);
                                             return false;
                                         } finally {
@@ -5591,7 +5591,7 @@ public class PersistentTopicsBase extends AdminResource {
                     } else {
                         indexFuture.complete(index);
                     }
-                } catch (IOException e) {
+                } catch (Throwable e) {
                     indexFuture.completeExceptionally(new RestException(Status.INTERNAL_SERVER_ERROR,
                             "Failed to get index from entry: " + e.getMessage()));
                 } finally {
@@ -5610,16 +5610,12 @@ public class PersistentTopicsBase extends AdminResource {
     }
 
 
-    private static Long getIndexFromEntry(Entry entry) throws IOException {
-        try {
-            final var brokerEntryMetadata = Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
-            if (brokerEntryMetadata != null && brokerEntryMetadata.hasIndex()) {
-                return brokerEntryMetadata.getIndex();
-            } else {
-                return null;
-            }
-        } catch (Throwable throwable) {
-            throw new IOException(throwable);
+    private static Long getIndexFromEntry(Entry entry) {
+        final var brokerEntryMetadata = Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+        if (brokerEntryMetadata != null && brokerEntryMetadata.hasIndex()) {
+            return brokerEntryMetadata.getIndex();
+        } else {
+            return null;
         }
     }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5542,7 +5542,10 @@ public class PersistentTopicsBase extends AdminResource {
                                         try {
                                             BrokerEntryMetadata brokerEntryMetadata =
                                                     Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
-                                            assert brokerEntryMetadata != null;
+                                            // Skip messages without index
+                                            if (brokerEntryMetadata == null) {
+                                                return true;
+                                            }
                                             return brokerEntryMetadata.getIndex() < index;
                                         } catch (Exception e) {
                                             log.error("Error deserialize message for message position find", e);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5525,10 +5525,10 @@ public class PersistentTopicsBase extends AdminResource {
                 .thenCompose(__ -> getTopicReferenceAsync(topicName))
                 .thenCompose(topic -> {
                     if (!(topic instanceof PersistentTopic persistentTopic)) {
-                        log.error("[{}] Get message id by offset on a non-persistent topic {} is not allowed",
+                        log.error("[{}] Get message id by index on a non-persistent topic {} is not allowed",
                                 clientAppId(), topicName);
                         return FutureUtil.failedFuture(new RestException(Status.METHOD_NOT_ALLOWED,
-                                "Get message id by offset on a non-persistent topic is not allowed"));
+                                "Get message id by index on a non-persistent topic is not allowed"));
                     }
                     ManagedLedger managedLedger = persistentTopic.getManagedLedger();
                     return findMessageIndexByPosition(
@@ -5556,7 +5556,7 @@ public class PersistentTopicsBase extends AdminResource {
                                 Position lastPosition = managedLedger.getLastConfirmedEntry();
                                 if (position == null || position.compareTo(lastPosition) > 0) {
                                     return FutureUtil.failedFuture(new RestException(Status.NOT_FOUND,
-                                            "Message not found for offset " + index));
+                                            "Message not found for index " + index));
                                 } else {
                                     return CompletableFuture.completedFuture(position);
                                 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5490,7 +5490,7 @@ public class PersistentTopicsBase extends AdminResource {
                         }));
     }
 
-    protected CompletableFuture<MessageId> internalGetMessageIDByOffsetAsync(Long offset, boolean authoritative) {
+    protected CompletableFuture<MessageId> internalGetMessageIDByIndexAsync(Long offset, boolean authoritative) {
         int partitionIndex = topicName.getPartitionIndex();
         CompletableFuture<Void> future = validateTopicOperationAsync(topicName, TopicOperation.PEEK_MESSAGES);
         return future.thenCompose(__ -> {
@@ -5506,10 +5506,10 @@ public class PersistentTopicsBase extends AdminResource {
                         return getPartitionedTopicMetadataAsync(topicName, authoritative, false)
                                 .thenAccept(topicMetadata -> {
                                     if (topicMetadata.partitions > 0) {
-                                        log.warn("[{}] Not supported getMessageIdByOffset operation on "
+                                        log.warn("[{}] Not supported getMessageIdByIndex operation on "
                                                         + "partitioned-topic {}", clientAppId(), topicName);
                                         throw new RestException(Status.METHOD_NOT_ALLOWED,
-                                                "GetMessageIDByOffset is not allowed on partitioned-topic");
+                                                "GetMessageIDByIndex is not allowed on partitioned-topic");
                                     }
                                 });
                     }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -5490,10 +5490,14 @@ public class PersistentTopicsBase extends AdminResource {
                         }));
     }
 
-    protected CompletableFuture<MessageId> internalGetMessageIDByIndexAsync(Long offset, boolean authoritative) {
-        if (pulsar().getBrokerService().isBrokerEntryMetadataEnabled()) {
+    protected CompletableFuture<MessageId> internalGetMessageIDByIndexAsync(Long index, boolean authoritative) {
+        if (!pulsar().getBrokerService().isBrokerEntryMetadataEnabled()) {
             return FutureUtil.failedFuture(new RestException(Status.PRECONDITION_FAILED,
                     "GetMessageIDByIndex is not allowed when broker entry metadata is disabled"));
+        }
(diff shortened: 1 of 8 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,53 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                return FutureUtil.failedFuture(new RestException(Status.METHOD_NOT_ALLOWED,
+                        "Get message id by offset on a non-persistent topic is not allowed"));
+            }
+            ManagedLedger managedLedger = persistentTopic.getManagedLedger();
+            return managedLedger.asyncFindPosition(entry -> {
+                try {
+                    BrokerEntryMetadata brokerEntryMetadata =
+                            Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+                    return brokerEntryMetadata.getIndex() < offset;
+                } catch (Exception e) {
```
StevenLuMT, line 5531: Need to determine whether brokerEntryMetadata is empty?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,53 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                    log.error("Error deserialize message for message position find", e);
+                } finally {
+                    entry.release();
+                }
+                return false;
+            });
+        }).thenCompose(position -> CompletableFuture
+                .completedFuture(new MessageIdImpl(position.getLedgerId(), position.getEntryId(), partitionIndex)));
+    }
```
StevenLuMT, line 5540: `managedLedger.asyncFindPosition` throws an exception,

```
thenCompose(position -> CompletableFuture
.completedFuture(new MessageIdImpl(position.getLedgerId(), position.getEntryId(), partitionIndex)));
```
Will it still process and return normally? Should I change to `whenComplete` to receive the exception and do some special processing

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,53 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                        clientAppId(), topicName);
+                return FutureUtil.failedFuture(new RestException(Status.METHOD_NOT_ALLOWED,
+                        "Get message id by offset on a non-persistent topic is not allowed"));
+            }
+            ManagedLedger managedLedger = persistentTopic.getManagedLedger();
+            return managedLedger.asyncFindPosition(entry -> {
+                try {
+                    BrokerEntryMetadata brokerEntryMetadata =
+                            Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+                    return brokerEntryMetadata.getIndex() < offset;
```
AuroraTwinkle, line 5530: maybe `brokerEntryMetadata.getIndex() <= offset`?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,53 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
                             return null;
                         }));
     }
+
+    protected CompletableFuture<MessageId> internalGetMessageIDByOffsetAsync(Long offset, boolean authoritative) {
```
AuroraTwinkle, line 5493: what happen if the index not enabled.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: No, we need to get the first message after this offset (inclusive).

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: The `managedLedger.asyncFindPosition` maybe return a `null`,so there need to add a NPE check.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: IMO, It is enough to catch the exception.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: The api will return a 404 error.

AuroraTwinkle: It is better to check whether the index is enabled first, and then call this function to avoid unnecessary searches.

StevenLuMT (approved):
#24220 PIP has a lot of adjustments. 
@liangyepianzhou please ping me after you modify the code and I will review it again.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,111 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                            managedLedger)
+                            .thenCompose(firstIndex -> {
+                                if (index < firstIndex) {
+                                    return CompletableFuture.completedFuture(PositionFactory.EARLIEST);
+                                } else {
+                                    return managedLedger.asyncFindPosition(entry -> {
+                                        try {
+                                            BrokerEntryMetadata brokerEntryMetadata =
+                                                    Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+                                            assert brokerEntryMetadata != null;
```
BewareMyPower, line 5545: `assert` should not be used here. It should only work in tests via the `-ea` option.

In production, if `asyncFindPosition` encounters an entry without `BrokerEntryMetadata`, NPE will be thrown here. 
First, many unnecessary error logs will be printed.

Second, returning false does not make sense, it will search to the left side. If the topic includes entries like: XXXOO, where `X` is the entry without index and `O` is the entry with index, eventually 0 will be returned.

Here are some solutions to handle this case:
- Fail the search directly if the topic contains entries without broker entry metadata
- Add an option to control the behavior (return false to left or return true to right). But this option assumes the entries distribution is either XX...OO... or OO...XX....For entries like OXOXOX, it does not work. In addition, an extra option makes this API hard to understand.

BewareMyPower (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Good catch! Thanks.  

First, we have checked that the broker configuration `isBrokerEntryMetadataEnabled = true`, so the entries distribution should be **OOOOOO** or **XXXOOO**.  

Second, for the **XXXOOO** scenario, before the lookup completes, we cannot determine whether the user-provided index points to **X** or **O**. Directly failing the request might not be ideal. For example, if a user's message retention time is set to 7 days or even a month, they would be unable to use this API for an entire month after enabling `isBrokerEntryMetadataEnabled`, even if they want to query the latest message ID by index.  

The example in [https://github.com/streamnative/kop/pull/916](https://github.com/streamnative/kop/pull/916) is correct—we can skip messages without an index. Additionally, this option could be set to `true` by default. As you mentioned, adding an extra option would make this API harder to understand.

BewareMyPower: Then return `true` instead of `false` here will be the best solution.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/PersistentTopics.java
```diff
@@ -5015,5 +5015,38 @@ public void removeAutoSubscriptionCreation(
...
+                    + "it will return the message id of the later message.")
+    @ApiResponses(value = {
+            @ApiResponse(code = 307, message = "Current broker doesn't serve the namespace of this topic"),
+            @ApiResponse(code = 403, message = "Don't have admin permission"),
+            @ApiResponse(code = 404, message = "Namespace or partitioned topic does not exist, "
+                    + "or the index is invalid"),
+            @ApiResponse(code = 406, message = "The topic is not a persistent topic"),
+            @ApiResponse(code = 412, message = "The broker is not enable broker entry metadata"),
+    })
+    public void getMessageIDByIndexAndPartitionID(@Suspended final AsyncResponse asyncResponse,
```
BewareMyPower, line 5031: The `AndPartitionID` suffix in the method seems meaningless (and confusing)?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,114 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                            "Broker entry metadata is not present in the message"));
+                } else {
+                    long index = brokerEntryMetadata.getIndex();
+                    if (index < 0) {
+                        indexFuture.completeExceptionally(new RestException(Status.PRECONDITION_FAILED,
+                                "Invalid message index: " + index));
+                    } else {
+                        indexFuture.complete(index);
+                    }
+                }
```
BewareMyPower, line 5589: You must release `entry` here.

I think you can add a common method for both `asyncReadEntry` and `asyncFindPosition` to use. Here is an example:

```java
    private static Optional<Long> getIndexAndRelease(Entry entry) throws IOException {
        try {
            final var brokerEntryMetadata = Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
            if (brokerEntryMetadata != null && brokerEntryMetadata.hasIndex()) {
                return Optional.of(brokerEntryMetadata.getIndex());
            } else {
                return Optional.empty();
            }
        } catch (Throwable throwable) {
            throw new IOException(throwable);
        }
    }
```

The implementation above forces the caller side to handle the two exceptional cases:
- `Optional.empty()` is returned, which indicates the broker entry metadata does not exist or the index does not exist (your code does not take it into consideration)
- The buffer is corrupted, where `IOException` will be thrown

But you can implement your own method if you care much about the overhead of `Optional` or don't want to use checked exception.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,114 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                .thenCompose(topic -> {
+                    if (!(topic instanceof PersistentTopic persistentTopic)) {
+                        log.error("[{}] Get message id by index on a non-persistent topic {} is not allowed",
+                                clientAppId(), topicName);
+                        return FutureUtil.failedFuture(new RestException(Status.METHOD_NOT_ALLOWED,
+                                "Get message id by index on a non-persistent topic is not allowed"));
+                    }
+                    ManagedLedger managedLedger = persistentTopic.getManagedLedger();
+                    return findMessageIndexByPosition(
+                            PositionFactory.create(managedLedger.getFirstPosition().getLedgerId(), 0),
```
BewareMyPower, line 5535: Why not just pass `managedLedger.getFirstPosition()` here?

BewareMyPower (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/PersistentTopics.java
liangyepianzhou: You are right! The suffix is incorrect.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Thanks~, I add a common method to handle it.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: IMO, the method `findMessageIndexByPosition` may be reused in the future,  so it need a param `position` instead of only managerLedger

liangyepianzhou:
> Out of the scope of this PR, I think it would be better to add a method to `ManagedLedger` for it. Because the customized ML implementation (e.g. StreamNative's Ursa engine) might not depend on the broker entry metadata to get the message's index. It could be another PIP.

Yes, we should add a new manager ledger interface to support customized ML implementations.

Additionally, the current message index retrieval solution has suboptimal performance. We could improve this by adding a field in `LedgerInfo` to record the first entry's index, then compute subsequent indexes directly from message IDs.  

I've created an [issue](https://github.com/apache/pulsar/issues/24374) to track this enhancement.

dao-jun:
Overall LGTM, but the current impl will scan all entire ML, maybe you can optimize it via PIP-404

liangyepianzhou:
> Overall LGTM, but the current impl will scan all entire ML, maybe you can optimize it via PIP-404

Interesting—I originally proposed directly adding the first entry's index in `LedgerInfo` to optimize this issue. While the `properties` field you added could also serve the same purpose, it currently doesn’t record the first entry's index. We can address this optimization in a future PIP, as mentioned in [issue](https://github.com/apache/pulsar/issues/24374).

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,129 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                                            Long messageIndex = getIndexAndRelease(entry);
+                                            if (messageIndex == null) {
+                                                return false; // Skip messages without index
+                                            } else {
+                                                // If the message index is less than the requested index,
+                                                // we continue searching
+                                                return messageIndex < index;
+                                            }
+                                        } catch (IOException e) {
+                                            log.error("Error deserialize message for message position find", e);
```
StevenLuMT, line 5558: 1. When this place is thrown wrong, print out the `entry`
2. Should `IOException` be replaced with `Throwable` ?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,129 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                                            return false;
+                                        }
+                                    });
+                                }
+                            }).thenCompose(position -> {
+                                Position lastPosition = managedLedger.getLastConfirmedEntry();
+                                if (position == null || position.compareTo(lastPosition) > 0) {
+                                    return FutureUtil.failedFuture(new RestException(Status.NOT_FOUND,
```
StevenLuMT, line 5567: `position.compareTo(lastPosition) > 0`
When does this happen? Is it normal?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,129 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                        clientAppId(), position, topicName, exception);
+                indexFuture.completeExceptionally(exception);
+            }
+        }, null);
+        return indexFuture;
+    }
+
+
+    private static Long getIndexAndRelease(Entry entry) throws IOException {
+        try {
```
StevenLuMT: Do not nest `release` in the method to prevent others from using entry after this method. Please mention the method caller to implement `release`

liangyepianzhou: done

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Occurs when the user passes in incorrect parameters.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: The  `Throwable ` is catch in `getIndexFromEntry`. The `IOException` is enough here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
BewareMyPower: Sorry I've thought that `managedLedger.getFirstPosition()` is the same with `PositionFactory.create(managedLedger.getFirstPosition().getLedgerId(), 0)`, so I asked the question.

But now there is a new question, what if the ledger of the first position is empty (i.e. topic is empty)? In this case, `asyncReadEntry` could fail because the entry does not exist. Could you add a test to show the behavior when `getMessageIdByIndex` is called on an empty topic?

nodece: `findMessageIndexByPosition(managedLedger.getFirstPosition(), ...)` is better.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,131 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                                            if (messageIndex == null) {
+                                                return false; // Skip messages without index
+                                            } else {
+                                                // If the message index is less than the requested index,
+                                                // we continue searching
+                                                return messageIndex < index;
+                                            }
+                                        } catch (IOException e) {
+                                            log.error("Error deserialize message for message position find", e);
+                                            return false;
```
nodece, line 5559: It looks like you should not ignore this exception.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
nodece: Could you add a test to cover this case?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,131 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+    private static Long getIndexFromEntry(Entry entry) throws IOException {
+        try {
+            final var brokerEntryMetadata = Commands.parseBrokerEntryMetadataIfExist(entry.getDataBuffer());
+            if (brokerEntryMetadata != null && brokerEntryMetadata.hasIndex()) {
+                return brokerEntryMetadata.getIndex();
+            } else {
+                return null;
+            }
+        } catch (Throwable throwable) {
+            throw new IOException(throwable);
```
nodece: This `catch` block appears to be unnecessary since it simply wraps and rethrows the original `Throwable` as an `IOException` without any additional handling.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5489,4 +5489,131 @@ protected CompletableFuture<AutoSubscriptionCreationOverride> internalGetAutoSub
...
+                        return FutureUtil.failedFuture(new RestException(Status.METHOD_NOT_ALLOWED,
+                                "Get message id by index on a non-persistent topic is not allowed"));
+                    }
+                    ManagedLedger managedLedger = persistentTopic.getManagedLedger();
+                    return findMessageIndexByPosition(
+                            PositionFactory.create(managedLedger.getFirstPosition().getLedgerId(), 0),
+                            managedLedger)
+                            .thenCompose(firstIndex -> {
+                                if (index < firstIndex) {
+                                    return CompletableFuture.completedFuture(PositionFactory.EARLIEST);
```
nodece, line 5545: Or consider returning null?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
```diff
@@ -420,4 +424,74 @@ public void testManagedLedgerTotalSize() throws Exception {
...
+        long index2 = message2.getIndex().get();
+        // 2.1 test partitioned topic name with partition index
+        MessageIdImpl messageIdByIndex2 =
+                (MessageIdImpl) admin.topics().getMessageIdByIndex(partitionedTopicName, index2);
+        Assert.assertEquals(messageIdByIndex2, messageId2);
+        // 2.2 test partitioned topic name without partition index
+        try {
+            messageIdByIndex2 = (MessageIdImpl) admin.topics().getMessageIdByIndex(topicName2, index2);
+            fail("Should not be able to get messageId by index for partitioned topic without partition index");
+        } catch (PulsarAdminException e) {
```
nodece: `expectThrows` is a better way.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Why not?  If a message is corrupted, shouldn't we skip it?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Can you explain why you think null is better than earliest?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
liangyepianzhou: update

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: done

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: >Could you add a test to show the behavior when getMessageIdByIndex is called on an empty topic?

Done

liangyepianzhou: >findMessageIndexByPosition(managedLedger.getFirstPosition(), ...) is better.

@nodece `managedLedger.getFirstPosition()` will return [first ledger Id, -1], regardless of whether the topic has data

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: `Exception Translation` is a common practice. I think it is reasonable to just raise IOException here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
nodece: Explicit signal that no valid position exists for the given index.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
nodece: Wrapping all `Throwable` into an `IOException` here adds an extra layer without clear benefit and may obscure the original exception type. Consider removing the `try-catch` and letting exceptions propagate naturally.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Maybe you can get more context here
https://github.com/apache/pulsar/pull/24220/files#r2090841224

nodece: Ok, got. 

`PositionFactory.EARLIEST` name feels odd, as its value `(-1, -1)` doesn't correspond to a real message and may mislead developers into thinking it's a valid starting point.  Returning `null` would more clearly indicate that the index is out of range and no valid position exists when used as the public value.

It's just my opinion.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
dao-jun: return false is OK, same as KoP

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
dao-jun: Return earliest is OK, but do we pass the index to the clients? 
I seem to have some impression, but I'm not sure.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
dao-jun: It maybe throws MessageConrruptException, no need to wrap it to IOException. agree with @nodece

dao-jun (approved):
Just some small comments, Overall LGTM.
By the way, if you rely on the AppendMessageIndexInterceptor, I'd suggest you need to check whether the interceptor configured before start the broker.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
BewareMyPower: The benefit of checked exceptions is to force the caller to handle the exception, which should be a "known" exception. However, the `IOException` thrown in this method is just a wrapper on an "unknown" exception.

A normal practice of exception handling should be:
- Fail fast or propagate the exception for unknown exceptions, because the caller never knows how to handle it correctly
- Handle the known exception according to the method's API docs

Specifically, for unknown exceptions, if propagating it could bring unexpected result, we should catch `Throwable` and handle the case well, for example,

```java
@RequiredArgsConstructor
public class Main {

    private final Function<Integer, Integer> func;

    public CompletableFuture<Integer> f(int x) {
        final var future = new CompletableFuture<Integer>();
        new Thread(() -> future.complete(func.apply(x))).start();
        return future;
    }

    public static void main(String[] args) {
        final var main = new Main(x -> 1 / x);
        final var result = main.f(0).join();
        System.out.println(result);
    }
}
```

To prevent the future of `Main#f` from never being completed, we should add a catch clause for it:

```java
            try {
                future.complete(func.apply(x));
            } catch (Throwable throwable) {
                future.completeExceptionally(throwable);
            }
```

But adding an additional method with exception signature is bad:

```java
    public CompletableFuture<Integer> f(int x) {
        final var future = new CompletableFuture<Integer>();
        new Thread(() -> {
            try {
                future.complete(apply(x));
            } catch (IOException e) {
            }
        }).start();
        return future;
    }
    
    public int apply(int x) throws IOException {
        try {
            return func.apply(x);
        } catch (Throwable throwable) {
            throw new IOException(throwable);
        }
    }
```

In this case, it's better to keep `getIndexFromEntry` simple without the exception signature and catch the `Throwable` when calling it like the suggestion [here](https://github.com/apache/pulsar/pull/24222/files#r2125780470).

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
BewareMyPower: > (-1, -1) doesn't correspond to a real message and may mislead developers into thinking it's a valid starting point. 

It's already documented, so it's good.

>  but do we pass the index to the clients?

Do you mean the actual first available index? If so, I think the API could be unnecessarily complicated (a good API does not do multiple things)

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
liangyepianzhou: Removed exception signature

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
```diff
@@ -420,4 +427,109 @@ public void testManagedLedgerTotalSize() throws Exception {
...
+                found = true;
+                break;
+            }
+            cause = cause.getCause();
+        }
+
+        Assert.assertTrue(found);
+    }
+
+    private void assertThrowsWithCause(Executable executable,
```
nodece: Refactor by the chatgpt:

```
import static org.assertj.core.api.Assertions.*;

private void assertThrowsWithCause(Executable executable,
                                   Class<? extends Throwable> expectedException,
                                   Class<? extends Throwable> expectedCause) {
    assertThatThrownBy(executable::execute)
        .isInstanceOf(expectedException)
        .hasCauseInstanceOf(expectedCause);
}
```

liangyepianzhou: While ChatGPT's provided code wasn't directly applicable to our scenario, it pointed us in the right direction. After testing, confirmed the following implementation works:

```java
private void assertThrowsWithCause(Executable executable,
                                 Class<? extends Throwable> expectedException,
                                 Class<? extends Throwable> expectedCause) {
    assertThatThrownBy(executable::execute)
            .isInstanceOf(expectedException)
            .getCause()  // First layer: ResponseProcessingException
            .getCause()  // Second layer: NotFoundException
            .isInstanceOf(expectedCause);
}
```
or
```java
private void assertThrowsWithCause(Executable executable,
                                   Class<? extends Throwable> expectedException,
                                   Class<? extends Throwable> expectedCause) {
    assertThatThrownBy(executable::execute)
            .isInstanceOf(expectedException)
            .hasRootCauseInstanceOf(expectedCause);
}
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
```diff
@@ -420,4 +426,93 @@ public void testManagedLedgerTotalSize() throws Exception {
...
+                .create();
+        for (int i = 0; i < 100; i++) {
+            producer.send("msg-" + i);
+        }
+
+        assertThrowsWithCause(() -> admin.topics().getMessageIdByIndex(topicName, 1000),
+                PulsarAdminException.class, NotFoundException.class);
+    }
+
+    private void assertThrowsWithCause(Executable executable,
```
nodece:
```suggestion
    private void assertThrowsWithCause(Runnable executable,
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
```diff
@@ -41,6 +46,7 @@
 import org.apache.pulsar.common.util.FutureUtil;
 import org.assertj.core.util.Sets;
 import org.awaitility.Awaitility;
+import org.junit.jupiter.api.function.Executable;
```
nodece:
```suggestion
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
```diff
@@ -19,20 +19,25 @@
 package org.apache.pulsar.broker.service;
 
 import static org.apache.bookkeeper.mledger.proto.MLDataFormats.ManagedLedgerInfo.LedgerInfo;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
```
nodece:
```suggestion
import static org.assertj.core.api.Assertions.assertThatThrownBy;
```

nodece (approved):
LGTM

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
liangyepianzhou: The `run` method of `Runnable` does not throw checked exceptions, while our code throws `PulsarAdminException` (a checked exception).  

Therefore, the code will fail at compile time because we are throwing a checked exception in the `run` method of `Runnable` (inside a lambda expression), whereas the `run` method of `Runnable` does not declare any exceptions.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
liangyepianzhou: same with https://github.com/apache/pulsar/pull/24222#discussion_r2128055455

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
liangyepianzhou: The difference between these two imports lies in the fact that they come from different entry point classes of the AssertJ library, but their actual functionality is identical. However, using `Assertions` is indeed the more standard practice.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/BrokerEntryMetadataE2ETest.java
nodece: Got, but we should not use the junit API.

BewareMyPower (approved):

StevenLuMT:
@liangyepianzhou 
https://lists.apache.org/thread/j9vxmmsw1r85l27v3rgp9tcnllk48wyb
Please add two labels, and cherry-pick PIP-415 into `branch-4.0` and `branch-4.1`.

PR #24247, opened 2025-05-05
[feat][broker] add consumer first consume/flow timestamp

mattisonchao:
### Motivation

This PR supports two stats for the consumer side to help identify the first message dispatch latency.

### Modifications

- Add stats `firstConsumedFlowTimestamp`
- Add stats `firstMessagesSentTimestamp `

### Verifying this change

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
@@ -101,7 +101,7 @@ public class Consumer {
     private final LongAdder messageAckCounter;
     private final Rate messageAckRate;
 
-    private volatile long firstConsumedTimestamp;
+    private volatile long firstMessagesSentTimestamp;
     private volatile long lastConsumedTimestamp;
     private volatile long lastAckedTimestamp;
     private volatile long firstConsumedFlowTimestamp;
@@ -434,8 +434,8 @@ public class Consumer {
         writeAndFlushPromise.addListener(status -> {
             // only increment counters after the messages have been successfully written to the TCP/IP connection
             if (status.isSuccess()) {
-                if (firstConsumedTimestamp == 0) {
-                    firstConsumedTimestamp =  System.currentTimeMillis();
+                if (firstMessagesSentTimestamp == 0) {
+                    firstMessagesSentTimestamp =  System.currentTimeMillis();
                 }
                 msgOut.recordMultipleEvents(totalMessages, totalBytes);
                 msgOutCounter.add(totalMessages);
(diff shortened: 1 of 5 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -435,6 +437,9 @@ public Future<Void> sendMessages(final List<? extends Entry> entries,
         writeAndFlushPromise.addListener(status -> {
             // only increment counters after the messages have been successfully written to the TCP/IP connection
             if (status.isSuccess()) {
+                if (firstConsumedTimestamp == 0) {
+                    firstConsumedTimestamp =  System.currentTimeMillis();
+                }
```
lhotari, line 442: just wondering if this makes sense since `flowPermits` method would get called before messages get sent to a consumer?

lhotari: oh sorry, there are 2 separate fields. Just wondering where `lastConsumedTimestamp` is set. wouldn't it be set in a similar way as `firstConsumedTimestamp`?

mattisonchao: > oh sorry, there are 2 separate fields. Just wondering where lastConsumedTimestamp is set. wouldn't it be set in a similar way as firstConsumedTimestamp?

IMO, we should also move `lastConsumedTimestamp ` to success sent. :/

liangyepianzhou (approved):
LGTM

dao-jun (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
lhotari: > IMO, we should also move `lastConsumedTimestamp ` to success sent. That will be more accurate, but I don't want to make this PR more complex. :/

This is problematic when the definition is different between `firstConsumedTimestamp` and `lastConsumedTimestamp`. Renaming `firstConsumedTimestamp` would resolve this. I think that "consumed" is very vague and it would be better to make it more specific such as `firstMessagesSentTimestamp`, `firstMessagesWrittenTimestamp` or `firstMessagesFlushedTimestamp`.

mattisonchao: let me rename it to `firstMessagesSentTimestamp`.

lhotari (approved):

PR #24241, opened 2025-05-02
[improve][build] Upgrade Mockito to 5.17.0 and byte-buddy to 1.15.11

lhotari:
### Motivation & Modifications

- Upgrade Mockito to 5.17.0 and byte-buddy to 1.15.11 so that Java 24 byte code is supported.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24256, opened 2025-05-06
[fix][ml] Cancel Metadata Store listeners at shutdown to fix resource leaks in some tests

lhotari:
### Motivation

The current MetadataStore API and MetadataStoreExtended API doesn't provide a way to cancel (unregister) registered listeners. This is a problem in some tests where the metadata store instance is shared, but the brokers are restarted during the test. Notifications will keep on delivered to the broker instances that have already been shutdown.
In addition, to the above problem, the listeners will also keep a strong reference to instances that have already been shutdown, impacting memory usage in some tests.
One of the additional motivations is to cleanup the testing infrastructure so that it would be possible to migrate Pulsar to use the most recent Mockito version, 5.17.0. When mocks are invalidated in newer Mockito versions, they will throw exceptions upon execution. This change is made to ensure that these problems that are currently seen in PR #24241 aren't originating from Metadata Store listeners.

### Modifications

Add new methods:
- `registerCancellableListener` to MetadataStore
- `registerCancellableSessionListener` to MetadataStoreExtended
 
Make the API binary compatible by adding default methods `registerListener` and `registerSessionListener` with the previous method signatures so that plugins using the Metadata API won't break due to this change. Mark these methods deprecated.

Migrate the Pulsar code to use the cancellable methods and store the callback instance for cancellation and use this in the shutdown sequence of each component that registers listeners.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/extended/MetadataStoreExtended.java
```diff
@@ -72,8 +72,23 @@ CompletableFuture<Stat> put(String path, byte[] value, Optional<Long> expectedVe
      *
      * @param listener
      *            the session listener
+     * @return a runnable that can be used to unregister the listener
      */
-    void registerSessionListener(Consumer<SessionEvent> listener);
+    Runnable registerCancellableSessionListener(Consumer<SessionEvent> listener);
```
heesung-sn, line 77: can we keep the name as-is? `registerSessionListener`?

lhotari: That would break binary compatibility for Metadata Store API consumers such as plugins. There's a default method with the old name.
There are solutions for changing the return type while keeping binary compatibility, but that would require a compiler plugin. Some solutions are explained in this article: https://www.morling.dev/blog/refining-return-type-java-methods-without-breaking-backwards-compatibility/ and in the comments.

lhotari: oh sorry, I made a mistake and called it `addSessionListener`, I'll fix that.

lhotari: Fixed it.

heesung-sn: Can the parent class just remove all listeners when closed?

lhotari: > Can the parent class just remove all listeners when closed?

@heesung-sn What are you referring as the "parent class"? We don't currently have any dependency injection or application eventing solution in Pulsar Java code so creating such parent classes could easily make things much more complicated than what is done in this PR. Each class will keep track of what listeners it has registered and execute the callbacks when the instance is closed.

The metadata store cannot be the "parent class" since it doesn't know when an individual pulsar broker is stopped. A metadata store can be shared and used across multiple pulsar broker instances in tests. That's why this solution is needed in the first place.

heesung-sn: I was referring to `AbstractMetadataStore`. Can AbstractMetadataStore.close() remove the listeners?

heesung-sohn (approved):
LGTM, but maybe we need more approvals from others too.

jiazhai (approved):

liangyepianzhou (approved):

liangyepianzhou:
This test has failed five times, maybe you should look into it.
`org.apache.bookkeeper.mledger.impl.ManagedLedgerFactoryShutdownTest.openEncounteredShutdown`

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStore.java
```diff
@@ -156,12 +156,28 @@ default CompletableFuture<Void> deleteIfExists(String path, Optional<Long> expec
...
      *
-     * @param listener
-     *            a consumer of notifications
+     * @param listener A Consumer that handles notifications. The consumer will receive a {@link Notification}
+     *                 object containing details about what changed in the metadata store.
+     * @return A Runnable that can be used to stop receiving notifications and unregister the listener.
+     * Executing this Runnable will cancel the subscription.
      */
-    void registerListener(Consumer<Notification> listener);
+    Runnable registerCancellableListener(Consumer<Notification> listener);
```
nodece, line 166: @lhotari It looks like this change breaks compatibility with existing implementations of `registerListener`.
Currently, you always invoke `registerCancellableListener`, which causes a compile-time failure if the custom metadata store only implements the older `registerListener` method.
To maintain backward compatibility, `registerCancellableListener` should delegate to `registerListener` by default and return a noop runnable. For example:
```
default Runnable registerCancellableListener(Consumer<Notification> listener) {
    registerListener(listener);
    return () -> {
        // noop
    };
}

@Deprecated
default void registerListener(Consumer<Notification> listener) {
    // noop
}
```
This ensures compatibility while allowing newer implementations to override `registerCancellableListener` directly.

nodece: If you use this way, please add a testcase to cover this change.

dao-jun (approved):

PR #24268, opened 2025-05-07
[improve][build] Allow building and running tests on JDK 24 and upcoming JDK 25 LTS

lhotari:
### Motivation

As a preparation for JDK 25 LTS version, Pulsar build should allow running with JDK 24 and with JDK 25 once it becomes available so that we can start fixing the different compatibility issues.

One of the already detected issues is that the Security Manager has been removed in JDK 24 with "JEP 486: Permanently Disable the Security Manager". This breaks at least tiered storage since it uses Hadoop libraries which depend on Security Manager (issue: [HDFS-17778](https://issues.apache.org/jira/browse/HDFS-17778)). There are also failing tests in Pulsar IO RabbitMQ connector which use Qpid libraries which require Security Manager features.

### Modifications

- change enforcer plugin config to allow building with JDK 24 and JDK 25 in addition to JDK 17 and JDK 24.
- add required JVM arguments for JDK 24+
- fix previous issue where Pulsar CI Flaky tests aren't built and run with JDK 21 by default
- Add JDK 24 and JDK 25 choices for triggered Pulsar CI and Pulsar CI Flaky GitHub Actions workflows
- Run Shade integration tests with JDK 24 too
- Fix issue in building the Docker image for JDK 24

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/pulsar-ci-flaky.yaml
+++ b/.github/workflows/pulsar-ci-flaky.yaml
@@ -44,7 +44,7 @@ on:
         options:
           - '17'
           - '21'
-        default: '17'
+        default: '21'
       trace_test_resource_cleanup:
         description: 'Collect thread & heap information before exiting a test JVM. When set to "on", thread dump and heap histogram will be collected. When set to "full", a heap dump will also be collected.'
         required: true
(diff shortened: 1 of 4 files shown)


dao-jun (approved):
Goooood job！

PR #24274, opened 2025-05-08
[fix][misc] Fix ByteBuf leak in SchemaUtils

lhotari:
### Motivation

Netty leak detection reporting catched this leak. Since it's using a heap buffer, it's possible that there's not real impact of the leak other than leak detection nagging about it. 

### Modifications

Release the buffer that is no longer used.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


dao-jun (approved):
LGTM

PR #24276, opened 2025-05-08
[improve][pip] PIP-416: Add a new topic method to implement trigger offload by size threshold

JunFu0814:
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-416.md
+++ b/pip/pip-416.md
@@ -1,0 +1,75 @@
+# PIP-416: Add a new topic method to implement trigger offload by size threshold
+
+# Background Knowledge
+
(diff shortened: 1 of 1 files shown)


liangyepianzhou (changes requested):
This vote is closed with one +1(binding) and two +1(non-binding). Moreover, it has only been 11 days since the vote was initiated, so we should wait for more feedback from the PMC.
https://lists.apache.org/thread/mzl2mdbylhz4fb5l55pmgyqjszrhop6l

JunFu0814:
> Hi @lhotari , I've submitted a pip for adding a new Pulsar client method that can manually trigger a topic-level offload based on a size threshold. Can you review it when you have time? Looking forward to your suggestions. Thanks!

Looking forward to your suggestions for  this pip！ @lhotari

nodece (approved):
LGTM, this PIP has been passed with +1(binding).

@liangyepianzhou This PIP follows the vote rule:
```
Everyone is welcome to vote on the proposal, though only the vote of the PMC members will be considered binding. The requirement is to have at least one binding +1 vote from a lazy majority if no binding -1 votes have been cast on the PIP. The vote should stay open for at least 48 hours.
```

Issue #24245, opened 2025-05-05
[Task] Enable SpotBug checks disabled in 4.9.x upgrade and fix possible detected bugs


lhotari:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

In #24243 SpotBugs was upgraded from 4.7.3 to 4.9.3. Since new checks were introduced, the checks were disabled with the `omitVisitors` configuration to make the upgrade possible without addressing the issues in the upgrade PR. 

### Solution

As a follow up, it would be necessary to enable checks to see what the actual problems are and fix them one by one.

- [ ] [ConstructorThrow](https://spotbugs.readthedocs.io/en/stable/detectors.html#constructorthrow)
- [ ] [FindPublicAttributes](https://spotbugs.readthedocs.io/en/stable/detectors.html#findpublicattributes)
- [ ] [SharedVariableAtomicityDetector](https://spotbugs.readthedocs.io/en/stable/detectors.html#sharedvariableatomicitydetector)
- [ ] [UnreadFields](https://spotbugs.readthedocs.io/en/stable/detectors.html#unreadfields)
- [ ] [MultipleInstantiationsOfSingletons](https://spotbugs.readthedocs.io/en/stable/detectors.html#multipleinstantiationsofsingletons)
- [ ] [InitializeNonnullFieldsInConstructor](https://spotbugs.readthedocs.io/en/stable/detectors.html#initializenonnullfieldsinconstructor)
- [ ] [FindNullDeref](https://spotbugs.readthedocs.io/en/stable/detectors.html#findnullderef)

### Alternatives

The checks that are unnecessarily verbose for Pulsar's code style could be kept disabled so that we don't end up adding exclusions for each occurrence like we have done for EI_EXPOSE_REP/EI_EXPOSE_REP2 bug pattern.

For example, [AT: This write of this 64-bit primitive variable may not atomic (AT_NONATOMIC_64BIT_PRIMITIVE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#at-nonatomic-64bit-primitive) should be ignored in Pulsar source code since Pulsar required 64-bit JVM, #10311. 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

nisranjan:
jbjkb

PR #24285, opened 2025-05-11
[improve][build]enable FindPublicAttributes check of SpotBugs

3pacccccc:
Fixes https://github.com/apache/pulsar/issues/24245

### Motivation
enable FindPublicAttributes check of SpotBugs
### Modifications
as the FindPublicAttributes check of SpotBugs requires, change the Primitive field from public into private to limit external accessibility. 

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/6


pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/PersistentOfflineTopicStats.java
```diff
@@ -23,19 +23,27 @@
...
-    public long totalMessages;
+    @Getter
+    @Setter
+    private long totalMessages;
 
     /** Total backlog. */
-    public long messageBacklog;
+    @Getter
+    @Setter
+    private long messageBacklog;
```
lhotari, line 46: These are most likely a breaking change in the public API. In other stats classes there's a separate "Impl" class and a separate interface. I don't see a way how to preserve API compatibility and perhaps we should just break the API when making the change.

pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/PersistentOfflineTopicStats.java
```diff
@@ -23,19 +23,27 @@
...
 
     /** Broker host where this stat was generated. */
     public final String brokerName;
```
lhotari, line 49: These fields should be handled for consistency.

lhotari (commented):
Thanks for the contribution, @3pacccccc . I think that this is the correct way how to enable FindPublicAttributes check. For some reason, public fields have been used in Pulsar code base extensively, perhaps due to the coding style and possibly also due to a belief that setters would be causing addition overhead. Modern JVMs can eliminate the overhead, so it's not really useful to avoid using setters.

The main challenge in this change is to ensure that we don't break current public interfaces or internal interfaces that are are used when there's a mixed version Pulsar cluster running during a new version upgrade or downgrade.  Public interfaces extend also to Pulsar plugins and they aren't very well defined. 

Since there's at least one breaking change in the public API at the moment, it's worth starting a discussion on the dev mailing list about these changes and the FindPublicAttributes check of SpotBugs. It's usually necessary to write [a PIP](https://github.com/apache/pulsar/tree/master/pip#pulsar-improvement-proposal-pip) so that everyone becomes aware of this change and that we properly document the considerations and make a decision about it. It's also possible that we decide to keep on ignoring the `FindPublicAttributes` check in SpotBugs, but it's useful to make that decision as part of handling the PIP. That's why the PIP should be mainly about deciding whether to handle it or not and when would we possibly ignore it.

PR #24286, opened 2025-05-12
[fix][broker]Non-global topic policies and global topic policies overwrite each other

poorbarcode:
#### Issue 1: Non-global topic policies and global topic policies overwrite each other
- Set local topic policy: dispatch rate `2000`
- Set global topic policy: dispatch rate `1000`
- Trigger a compaction for the system topic `__change_events`
- Issue: get local topic policy `1000`

#### Issue 2: Global policy can not be removed successfully
- Set global topic policy: dispatch rate `1000`
- Delete topic, the global topic-level policies will never be deleted

### Modifications

Fix two issues

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
@@ -568,9 +568,10 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
     private void refreshTopicPoliciesCache(Message<PulsarEvent> msg) {
         // delete policies
         if (msg.getValue() == null) {
+            boolean isGlobalPolicy = TopicPoliciesService.isGlobalPolicy(msg);
             TopicName topicName = TopicName.get(TopicPoliciesService.unwrapEventKey(msg.getKey())
                     .getPartitionedTopicName());
-            if (hasReplicateTo(msg)) {
+            if (isGlobalPolicy) {
                 globalPoliciesCache.remove(topicName);
             } else {
                 policiesCache.remove(topicName);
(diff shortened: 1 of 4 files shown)


Technoboy- (approved):

gaoran10 (approved):

lhotari:
@poorbarcode It seems that this change should be covered with a PIP. Do you agree?

poorbarcode:
@lhotari 

> @poorbarcode It seems that this change should be covered with a PIP. Do you agree?

Could you explain why it is needed?

lhotari:
> @lhotari
> 
> > @poorbarcode It seems that this change should be covered with a PIP. Do you agree?
> 
> Could you explain why it is needed?

This PR changed the way how topic policies are stored. It would be useful to have such details documented in PIPs. This change was also cherry-picked to maintenance branches and it would be useful to document the backwards compatibility in the PIP document.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -218,6 +218,25 @@ private CompletableFuture<Void> sendTopicPolicyEvent(TopicName topicName, Action
...
+                                      @Nullable TopicPolicies policies) {
+        PulsarEvent event = getPulsarEvent(topicName, actionType, policies);
+        if (!ActionType.DELETE.equals(actionType)) {
+            return writer.writeAsync(getEventKey(event, policies != null && policies.isGlobalPolicies()), event);
+        }
+        // When a topic is deleting, delete both non-global and global topic-level policies.
+        CompletableFuture<MessageId> deletePolicies = writer.deleteAsync(getEventKey(event, true), event)
+            .thenCompose(__ -> {
+                return writer.deleteAsync(getEventKey(event, false), event);
+            });
```
Denovo1998, line 232: @poorbarcode @lhotari It's a bit strange, why isn't `CompletableFuture.allOf` used here?

poorbarcode: I think using `CompletableFuture.allOf` can benefit the performance. Thanks for mentioning this improvement

Denovo1998: Of course, you can submit a small PR later.

PR #24290, opened 2025-05-12
[improve][pip]pip-417:Enable SpotBugs FindPublicAttributes Check

3pacccccc:
PIP documentation.
Implementation PR: https://github.com/apache/pulsar/pull/24285

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-417.md
+++ b/pip/pip-417.md
@@ -1,0 +1,72 @@
+# PIP-417: Enable SpotBugs `FindPublicAttributes` Check
+
+## Background Knowledge
+
+As Pulsar has upgraded the SpotBugs version from 4.7.x to 4.9.x ([PR #24243](https://github.com/apache/pulsar/pull/24243)), the `FindPublicAttributes` check is currently disabled.
+
+To enable the `FindPublicAttributes` check, we need to refactor many classes by changing primitive fields from `public` to `private` to limit external accessibility. I have recently attempted this in [PR #24285](https://github.com/apache/pulsar/pull/24285).
+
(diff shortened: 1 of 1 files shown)


crossoverJie:
I also have doubts about extensively using public fields, but I feel the cost of making changes is not proportional to the benefits.

BewareMyPower:
> but I feel the cost of making changes is not proportional to the benefits.

+1 to this point. @3pacccccc Could you have a demonstration for how many interfaces and classes will be affected?

lhotari:
> > but I feel the cost of making changes is not proportional to the benefits.
> 
> +1 to this point. @3pacccccc Could you have a demonstration for how many interfaces and classes will be affected?

@BewareMyPower the changes can be seen in #24285

BewareMyPower:
I see. It LGTM now.

3pacccccc:
> I see. It LGTM now.
@BewareMyPower But I worry about this change will break current public interfaces or internal interfaces, any suggestions?

BewareMyPower:
IMO, it's okay if these interfaces changes only affect the plugins. Otherwise, it would be very complicated to maintain the compatibility and the code would become very ugly. I started a discussion before: https://lists.apache.org/thread/v4bswc3byqtvpl3h4cjpvzm8sbsy0wq9

And I agree with this point

> It would essentially prevent most
refactorings and improvements to Pulsar. We could choose to do so, but
it has consequences.
Each class should have a reason why it's part of the stable API. It
shouldn't be too hard to list what is used by custom plugins and why.
Custom plugin authors should provide this information and this could
be taken into account when the stable API annotations are added and
reviewed.

PR #24296, opened 2025-05-13
[improve][pip] PIP-418: Determine the behaviors for components that rely on BookKeeper when BookKeeper is not used

BewareMyPower:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-418.md
+++ b/pip/pip-418.md
@@ -1,0 +1,174 @@
+# PIP-418: Determine the behaviors for components that rely on BookKeeper when BookKeeper is not used
+
+# Background knowledge
+
+`PulsarService` has a method for other classes to access the BookKeeper client (`BookKeeper`):
+
+```java
+    public BookKeeper getBookKeeperClient() {
+        ManagedLedgerStorageClass defaultStorageClass = getManagedLedgerStorage().getDefaultStorageClass();
+        if (defaultStorageClass instanceof BookkeeperManagedLedgerStorageClass bkStorageClass) {
+            return bkStorageClass.getBookKeeperClient();
+        } else {
+            // TODO: Refactor code to support other than default bookkeeper based storage class
+            throw new UnsupportedOperationException("BookKeeper client is not available");
+        }
+    }
+```
+
+The BookKeeper client is obtained from the `BookkeeperManagedLedgerStorageClass` interface, which extends the `ManagedLedgerStorageClass` interface. This interface represents the actual storage layer of `ManagedLedgerStorage`.
+
+```java
+    Collection<ManagedLedgerStorageClass> getStorageClasses();
+```
+
+The following components share the same `BookKeeper` instance:
+- `PulsarCompactionServiceFactory`: The default compaction service configured by `compactionServiceFactoryClassName`
+- `EventTimeCompactionServiceFactory`: Another built-in compaction service introduced by PIP-215
+- `StrategicTwoPhaseCompactor`: The compactor used only for the `loadbalancer-service-unit-state` topic (introduced by PIP-352)
+
+This `BookKeeper` instance is also used in:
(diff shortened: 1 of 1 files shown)


pip/pip-418.md
```diff
@@ -0,0 +1,174 @@
...
+
+```java
+    public BookKeeperClientFactory getBookKeeperClientFactory() {
+        return bkClientFactory;
+    }
+```
+
+The following components leverages this factory to create their own `BookKeeper` instances:
+- `BookkeeperSchemaStorageFactory`: The default schema registry storage configured by `schemaRegistryStorageClassName`
+- `BookkeeperBucketSnapshotStorage`: A non-default built-in delayed delivery tracker configured by `delayedDeliveryTrackerFactoryClassName`
```
codelipenghui, line 44: I’m not entirely sure why separate BookKeeper instances are used for these two cases. While this doesn’t directly affect the goal of the proposal, should we consider addressing or revisiting this as part of a broader cleanup?

pip/pip-418.md
```diff
@@ -0,0 +1,174 @@
...
+
+# High Level Design
+
+When BookKeeper is not the default storage, 
+
+| Component | Behavior |
+| :- | :- |
+| Schema Registry | Fail fast |
+| Delayed Delivery Tracker | Fail fast if it's `BucketDelayedDeliveryTrackerFactory` |
+| Compaction Service | Fallback to a dummy implementation |
```
codelipenghui, line 70: These three items refer to Pulsar’s default implementations, correct? If users have provided their own plugin implementations, they shouldn’t be affected — is that understanding accurate?

pip/pip-418.md
```diff
@@ -0,0 +1,174 @@
...
+public interface ManagedLedgerStorage extends AutoCloseable {
+
+    void initialize(ServiceConfiguration conf,
+                    MetadataStoreExtended metadataStore,
+                    OpenTelemetry openTelemetry) throws Exception;
+```
+
+The reason to have these parameters is that `BookKeeperClientFactory` is maintained in `PulsarService` and creating a `BookKeeper` instance requires a `EventLoopGroup` object.
+
+However, these BookKeeper related parameters are not useful for a `ManagedLedgerStorage` implementation. Even if it's another BookKeeper based implementation, it should create its own BookKeeper client objects according to the configuration (`conf`). In addition, the `EventLoopGroup` instance passed to the method represents the `pulsar-io` thread pool, which is coupled with Pulsar's internal details. The custom implementation should create is own thread pool.
```
codelipenghui:
```suggestion
However, these BookKeeper related parameters are not useful for a `ManagedLedgerStorage` implementation. Even if it's another BookKeeper based implementation, it should create its own BookKeeper client objects according to the configuration (`conf`). In addition, the `EventLoopGroup` instance passed to the method represents the `pulsar-io` thread pool, which is coupled with Pulsar's internal details. The custom implementation should create its own thread pool.
```

pip/pip-418.md
```diff
@@ -0,0 +1,174 @@
...
+    return ManagedLedgerStorage.create(config, localMetadataStore, openTelemetry);
+}
+```
+
+Add a new method to `BookkeeperManagedLedgerStorageClass`:
+
+```java
+public interface BookkeeperManagedLedgerStorageClass extends ManagedLedgerStorageClass {
+    /* ... */
+    BookKeeperClientFactory getBookKeeperClientFactory();
```
codelipenghui, line 110: Is it designed to be used by `BookkeeperSchemaStorageFactory` and `BookkeeperBucketSnapshotStorage`?

I mean, BookkeeperManagedLedgerStorageClass already has `getBookKeeperClient`, and should all Pulsar's internal implementations use the same bookkeeper client? Maybe I missed something.

PR #22381, opened 2024-03-29
[Enhancement]Added a metric for geo replication for tracking replicated subscriptions snapshot timeouts

nikam14:
Fixes #21793

### Motivation
Geo replication replicated subscriptions (PIP-33) [snapshot creation might time out](https://github.com/apache/pulsar/wiki/PIP-33%3A-Replicated-subscriptions#constructing-a-cursor-snapshot).
The code contains a debug log message when this happens
When this happens, the subscription state won't be reflected on the remote side and a backlog would build up.
There's no metric to detect this situation.

### Modifications
Add a new metric `pulsar_replicated_subscriptions_snapshot_timeouts` which is a counter (that only resets when the broker restarts).

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #16651, opened 2022-07-18
Skip replicated subscription update if the previous update was for the same snapshot id

lhotari:
### Motivation

- Reduces unnecessary replicated subscription updates

### Modifications

- skip replicated subscription update if the previous update was for the same snapshot id
- Move replicated subscription advance markdelete position logic to proper code location
 - the operation is asynchronous and should be handled after the markdelete position
   updating operation completes

Issue #10054, opened 2021-03-26
[Geo-replicaiton] Subscription replication is not working across clusters


yabinmeng:
**Describe the bug**
I'm testing geo-replication with subscription enabled. In my testing environment, there are 2 clusters (ClusterA and ClusterB) with a global zookeeper on  a dedicated server host. 

Each cluster also has its own producer and consumer client application which is based on Pulsar (Java) producer API and consumer Sync API. 

Below is the summary of my testing steps:
1. ProducerA publishes some messages (e.g. 10 messages  0 ~ 9) to a topic that is geo-replication enabled.
2. ConsumerA receives the first half of the messages (e.g. msg 0 ~ 4) from the same topic and gets killed.
3. After a short while (> 10s), starts ConsumerB in ClusterB to receive messages from the same topic.

I'm expecting ConsumerB to receive the second half of the messages (e.g. msg 5 ~ 9) if subscription replication is working properly. But the results shows that for most time, ConsumerB receives all 10 messages (e.g. msg 0 ~ 9). There are also several cases where ConsumerB receives 6 or 7 messages (e.g. msg 3 ~ 9 or 4 ~ 9). 

My testing is based on Pulsar 2.6.3 with default configuration, with key geo-replication/subscription replication settings below:

**broker.conf**
`
# Enable tracking of replicated subscriptions state across clusters.
enableReplicatedSubscriptions=true

# Frequency of snapshots for replicated subscriptions tracking.
replicatedSubscriptionsSnapshotFrequencyMillis=1000
`

The consumer application also has the following code added:
`
.replicateSubscriptionState(true)
`

**To Reproduce**
See "Description" section

**Expected behavior**
See "Description" section

**Screenshots**
Below is one example of the testing results;

1.  **ProducerA** publishes 6 messages to a geo-replication enabled topic "georep/default/t1"
`
$ ./producerA.sh -n 6 -t georep/default/t1
2021-03-24 06:51:54.965
---------------------------------------------------------
Cluster Name: ClusterA
Producer    : [P]0090276c-8cd2-11eb-b297-d555334bf2d0
Topic       : persistent://georep/default/t1
Number of messages to publish: 6
---------------------------------------------------------
  message published: msg-key=[0] 20210324-065155-016, msg-payload=mN8etPUo9l0PDTW2cpa0
  message published: msg-key=[1] 20210324-065155-089, msg-payload=z4tGKNSz4vmnqWIwIKO6kiclyhhso
[...]

PR #24300, opened 2025-05-14
[improve][broker] optimize the problem of subscription snapshot cache not hitting

liudezhi2098:
### Motivation

- When message acknowledgment confirmation is slower than message consumption rate, subscription cursor synchronization fails to complete. This occurs because:

1. Current Behavior
   - With large receiver queues (e.g., `receiverQueueSize=1000`), the cursor never synchronizes
    ```
     Consumer<String> consumer = client.newConsumer(Schema.STRING)
               .topic(topic)
               .subscriptionName("sub")
               .receiverQueueSize(1000)
                .subscribe();
      while (true) {
                Message<String> msg = consumer.receive();
                 consumer.acknowledge(msg);
                 Thread.sleep(100);
            }
    
    ```
   - With small queues (e.g., `receiverQueueSize=1`), synchronization works properly
    ```
     Consumer<String> consumer = client.newConsumer(Schema.STRING)
               .topic(topic)
               .subscriptionName("sub")
               .receiverQueueSize(1)
                .subscribe();
      while (true) {
                Message<String> msg = consumer.receive();
                 consumer.acknowledge(msg);
                 Thread.sleep(100);
            }
    
    ```
2. Root Cause:
    - The SnapshotCache updates too aggressively
    - When advancedMarkDeletePosition executes, valid cache entries are frequently unavailable
    - Current eviction strategy doesn't account for periodic synchronization needs
  https://github.com/apache/pulsar/blob/7be22eb2b23057bd5e09c361a43d6ccdcc0c8afd/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java#L44C4-L59C6

### Modifications
1. Cache Update Strategy:
    - modified the cache to maintain mapping relationships for remote clusters.
2. Eviction Policy Enhancement:
    - When cache reaches capacity (maxSnapshotToCache):
        - Allow subsequent snapshots to be added through periodic dynamic adjustment
        - The latest snapshot is used to replace the intermediate snapshot of the cache, and the update becomes slower as the difference between the latest snapshot time and the Mark Delete Position time increases.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
@@ -18,6 +18,8 @@
  */
 package org.apache.pulsar.broker.service.persistent;
 
+import java.util.Iterator;
+import java.util.Map;
 import java.util.NavigableMap;
 import java.util.TreeMap;
 import lombok.extern.slf4j.Slf4j;
@@ -33,15 +35,20 @@ import org.apache.pulsar.common.api.proto.ReplicatedSubscriptionsSnapshot;
 public class ReplicatedSubscriptionSnapshotCache {
     private final String subscription;
     private final NavigableMap<Position, ReplicatedSubscriptionsSnapshot> snapshots;
+    private final NavigableMap<Position, Long> positionToTimestamp;
     private final int maxSnapshotToCache;
+    private final int snapshotFrequencyMillis;
 
-    public ReplicatedSubscriptionSnapshotCache(String subscription, int maxSnapshotToCache) {
+    public ReplicatedSubscriptionSnapshotCache(String subscription, int maxSnapshotToCache,
+                                               int snapshotFrequencyMillis) {
         this.subscription = subscription;
         this.snapshots = new TreeMap<>();
+        this.positionToTimestamp = new TreeMap<>();
         this.maxSnapshotToCache = maxSnapshotToCache;
+        this.snapshotFrequencyMillis = snapshotFrequencyMillis;
     }
 
-    public synchronized void addNewSnapshot(ReplicatedSubscriptionsSnapshot snapshot) {
+    public synchronized void addNewSnapshot(ReplicatedSubscriptionsSnapshot snapshot, long publishTime) {
         MarkersMessageIdData msgId = snapshot.getLocalMessageId();
         Position position = PositionFactory.create(msgId.getLedgerId(), msgId.getEntryId());
 
@@ -50,12 +57,60 @@ public class ReplicatedSubscriptionSnapshotCache {
                     snapshot.getSnapshotId());
         }
 
-        snapshots.put(position, snapshot);
-
-        // Prune the cache
-        while (snapshots.size() > maxSnapshotToCache) {
-            snapshots.pollFirstEntry();
+        if (positionToTimestamp.lastEntry() == null) {
+            snapshots.put(position, snapshot);
+            positionToTimestamp.put(position, publishTime);
+            return;
         }
+
+        final long timeSinceFirstSnapshot = publishTime - positionToTimestamp.firstEntry().getValue();
+        final long timeSinceLastSnapshot = publishTime - positionToTimestamp.lastEntry().getValue();
+        final long timeWindowPerSlot = timeSinceFirstSnapshot / snapshotFrequencyMillis / maxSnapshotToCache;
+        // reset cursor
+        if (position.compareTo(positionToTimestamp.firstKey()) < 0) {
+            positionToTimestamp.clear();
+            snapshots.clear();
+            snapshots.put(position, snapshot);
+            positionToTimestamp.put(position, publishTime);
+            return;
+        } else if (position.compareTo(positionToTimestamp.lastKey()) < 0) {
+            while (position.compareTo(positionToTimestamp.lastKey()) < 0) {
+                positionToTimestamp.pollLastEntry();
+                snapshots.pollLastEntry();
+            }
+            snapshots.put(position, snapshot);
+            positionToTimestamp.put(position, publishTime);
+        }
+
+        if (timeSinceLastSnapshot < snapshotFrequencyMillis || timeSinceLastSnapshot < timeWindowPerSlot) {
+            return;
+        }
+        if (snapshots.size() < maxSnapshotToCache) {
+            snapshots.put(position, snapshot);
+            positionToTimestamp.put(position, publishTime);
+        } else {
+            int medianIndex = maxSnapshotToCache / 2;
+            Position positionToRemove = findPositionByIndex(medianIndex);
+            if (positionToRemove != null) {
+                positionToTimestamp.remove(positionToRemove);
+                snapshots.remove(positionToRemove);
+            }
+            positionToTimestamp.put(position, publishTime);
+            snapshots.put(position, snapshot);
+        }
+    }
+
+    private Position findPositionByIndex(int targetIndex) {
+        Iterator<Map.Entry<Position, Long>> it = positionToTimestamp.entrySet().iterator();
+        int currentIndex = 0;
+        while (it.hasNext()) {
+            Map.Entry<Position, Long> entry = it.next();
+            if (currentIndex == targetIndex) {
+                return entry.getKey();
+            }
+            currentIndex++;
+        }
+        return null;
     }
 
     /**
@@ -73,6 +128,7 @@ public class ReplicatedSubscriptionSnapshotCache {
                 // This snapshot is potentially good. Continue the search for to see if there is a higher snapshot we
                 // can use
                 snapshot = snapshots.pollFirstEntry().getValue();
+                positionToTimestamp.pollFirstEntry();
             }
         }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java
@@ -211,7 +211,7 @@ public abstract class AbstractBaseDispatcher extends EntryFilterSupport implemen
 
                 if (Markers.isReplicatedSubscriptionSnapshotMarker(msgMetadata)) {
                     final int readerIndex = metadataAndPayload.readerIndex();
-                    processReplicatedSubscriptionSnapshot(pos, metadataAndPayload);
+                    processReplicatedSubscriptionSnapshot(pos, metadataAndPayload, msgMetadata.getPublishTime());
                     metadataAndPayload.readerIndex(readerIndex);
                 }
 
@@ -358,13 +358,13 @@ public abstract class AbstractBaseDispatcher extends EntryFilterSupport implemen
                 && maxConsumersPerSubscription <= consumerSize;
     }
 
-    private void processReplicatedSubscriptionSnapshot(Position pos, ByteBuf headersAndPayload) {
+    private void processReplicatedSubscriptionSnapshot(Position pos, ByteBuf headersAndPayload, long publishTime) {
         // Remove the protobuf headers
         Commands.skipMessageMetadata(headersAndPayload);
 
         try {
             ReplicatedSubscriptionsSnapshot snapshot = Markers.parseReplicatedSubscriptionsSnapshot(headersAndPayload);
-            subscription.processReplicatedSubscriptionSnapshot(snapshot);
+            subscription.processReplicatedSubscriptionSnapshot(snapshot, publishTime);
         } catch (Throwable t) {
             log.warn("Failed to process replicated subscription snapshot at {} -- {}", pos, t.getMessage(), t);
             return;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Subscription.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Subscription.java
@@ -106,7 +106,7 @@ public interface Subscription extends MessageExpirer {
 
     boolean isSubscriptionMigrated();
 
-    default void processReplicatedSubscriptionSnapshot(ReplicatedSubscriptionsSnapshot snapshot) {
+    default void processReplicatedSubscriptionSnapshot(ReplicatedSubscriptionsSnapshot snapshot, long publishTime) {
         // Default is no-op
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
@@ -210,7 +210,8 @@ public class PersistentSubscription extends AbstractSubscription {
             this.replicatedSubscriptionSnapshotCache = null;
         } else if (this.replicatedSubscriptionSnapshotCache == null) {
             this.replicatedSubscriptionSnapshotCache = new ReplicatedSubscriptionSnapshotCache(subName,
-                    config.getReplicatedSubscriptionsSnapshotMaxCachedPerSubscription());
+                    config.getReplicatedSubscriptionsSnapshotMaxCachedPerSubscription(),
+                    config.getReplicatedSubscriptionsSnapshotFrequencyMillis());
         }
 
         if (this.cursor != null) {
@@ -1523,10 +1524,10 @@ public class PersistentSubscription extends AbstractSubscription {
     }
 
     @Override
-    public void processReplicatedSubscriptionSnapshot(ReplicatedSubscriptionsSnapshot snapshot) {
+    public void processReplicatedSubscriptionSnapshot(ReplicatedSubscriptionsSnapshot snapshot, long publishTime) {
         ReplicatedSubscriptionSnapshotCache snapshotCache = this.replicatedSubscriptionSnapshotCache;
         if (snapshotCache != null) {
-            snapshotCache.addNewSnapshot(new ReplicatedSubscriptionsSnapshot().copyFrom(snapshot));
+            snapshotCache.addNewSnapshot(new ReplicatedSubscriptionsSnapshot().copyFrom(snapshot), publishTime);
         }
     }
 
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCacheTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCacheTest.java
@@ -28,10 +28,13 @@ import org.testng.annotations.Test;
 
 @Test(groups = "broker")
 public class ReplicatedSubscriptionSnapshotCacheTest {
+    int snapshotFrequencyMillis = 1000;
 
     @Test
     public void testSnapshotCache() {
-        ReplicatedSubscriptionSnapshotCache cache = new ReplicatedSubscriptionSnapshotCache("my-subscription", 10);
+
+        ReplicatedSubscriptionSnapshotCache cache =
+                new ReplicatedSubscriptionSnapshotCache("my-subscription", 10, snapshotFrequencyMillis);
 
         assertNull(cache.advancedMarkDeletePosition(PositionFactory.create(0, 0)));
         assertNull(cache.advancedMarkDeletePosition(PositionFactory.create(100, 0)));
@@ -51,11 +54,11 @@ public class ReplicatedSubscriptionSnapshotCacheTest {
         ReplicatedSubscriptionsSnapshot s7 = new ReplicatedSubscriptionsSnapshot()
                 .setSnapshotId("snapshot-7");
         s7.setLocalMessageId().setLedgerId(7 ).setEntryId(7);
-
-        cache.addNewSnapshot(s1);
-        cache.addNewSnapshot(s2);
-        cache.addNewSnapshot(s5);
-        cache.addNewSnapshot(s7);
+        long publishTime = System.currentTimeMillis();
+        cache.addNewSnapshot(s1, publishTime);
+        cache.addNewSnapshot(s2, publishTime + snapshotFrequencyMillis);
+        cache.addNewSnapshot(s5, publishTime + 2L * snapshotFrequencyMillis);
+        cache.addNewSnapshot(s7, publishTime + 3L * snapshotFrequencyMillis);
 
         assertNull(cache.advancedMarkDeletePosition(PositionFactory.create(0, 0)));
         assertNull(cache.advancedMarkDeletePosition(PositionFactory.create(1, 0)));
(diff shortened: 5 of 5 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -50,12 +57,60 @@ public synchronized void addNewSnapshot(ReplicatedSubscriptionsSnapshot snapshot
...
+            int medianIndex = maxSnapshotToCache / 2;
+            Position positionToRemove = findPositionByIndex(medianIndex);
+            if (positionToRemove != null) {
+                positionToTimestamp.remove(positionToRemove);
+                snapshots.remove(positionToRemove);
+            }
+            positionToTimestamp.put(position, publishTime);
+            snapshots.put(position, snapshot);
+        }
+    }
```
lhotari, line 101: Please add sufficient code comments so that it's possible to read the code and understand the algorithm. It might be worth extracting this logic to a method and add a comment on the method with a longer description.

lhotari:
> * When message acknowledgment confirmation is slower than message consumption rate, subscription cursor synchronization fails to complete.

@liudezhi2098 Regarding this scenario, is there a way to find out this happens?
Does the metric `pulsar_replicated_subscriptions_timedout_snapshots` added in #22381 help detecting problems?

lhotari (commented):
Great work @liudezhi2098! Added a comment about adding code comments. :)

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
liudezhi2098: Good idea

lhotari (changes requested):
Since the `publishTime` information comes from Pulsar client, the logic could be brittle. In Pulsar, there's also an optional `brokerPublishTime` which is not enabled by default (requires specific configuration in all brokers). 

Have you considered what could happen when `publishTime` values aren't in sync?

I've understood that the ["PIP-33: Replicated subscriptions"](https://github.com/apache/pulsar/wiki/PIP-33%3A-Replicated-subscriptions) algorithm rely on vector clocks so that clock sync doesn't become a problem. ([earlier discussion](https://github.com/apache/pulsar/discussions/22315#discussioncomment-8875503))
The snapshots are the way how the vector clocks are synchronized, at least that's how I interpret it from one view point.

The changes in this PR don't currently make sense to me.

I'd assume that in your problem scenario, the correct approach would be to tune `replicatedSubscriptionsSnapshotFrequencyMillis`, `replicatedSubscriptionsSnapshotTimeoutSeconds` and `replicatedSubscriptionsSnapshotMaxCachedPerSubscription` values.

https://github.com/apache/pulsar/blob/825b862d8674f32ffbc13f10c17611ba394bfcc9/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java#L1444-L1457

Have you already done this?

Currently it is a problem that it's necessary to tune the values to fix issues and it's also hard to notice that the problem is occurring.

It looks like future improvements are needed too.

lhotari:
> Current Behavior
> * With large receiver queues (e.g., `receiverQueueSize=1000`), the cursor never synchronizes

@liangyepianzhou Do you have a separate repro app where this could be observed with real brokers (let's say 2 Pulsar broker within docker-compose + some test app)? Creating a separate Git repo for such a repro would be one approach to share it. Having a runnable repro makes things easier for reviewers too.

lhotari:
> * The SnapshotCache updates too aggressively
> * When advancedMarkDeletePosition executes, valid cache entries are frequently unavailable

@liudezhi2098 One thought here is that perhaps there could be interaction between `ReplicatedSubscriptionsController` and all `ReplicatedSubscriptionSnapshotCache` instances? Could there be a solution that when "updates too aggressively" that there's a solution in place that a snapshot would be completed every `replicatedSubscriptionsSnapshotFrequencyMillis`. 
Since the `ReplicatedSubscriptionSnapshotCache` is an internal interface, we don't need to keep it as a "cache". It's possible that it doesn't make sense in the revisited solution. 
Do you have a chance to try something in this area instead since I don't think that using `publishTime` in the solution makes sense.

lhotari (changes requested):
> The generation of snapshots is completed through the exchange of snapshotRequest and snapshotResponse between two clusters. Ultimately, the ReplicatedSubscriptionsController writes Marker messages, and using publishTime is reliable because this behavior occurs within the same broker.
> 
> Of course, the topic may be transferred to another broker, but this is a low-frequency scenario, and its publishTime will not exhibit continuous jumps.
> 
> However, we can adopt a simpler approach that doesn't require using publishTime. Instead, we can record the current system time each time the snapshotCache is updated and use this timestamp for dynamic adjustments.

Thanks for explaining that. I missed that point that the marker messages are originated from the same broker. `publishTime` would be fine due to that detail. 

When looking at the current master branch code in ReplicatedSubscriptionSnapshotCache.addNewSnapshot, I'd assume that a potential solution to the problem could be that the current mark delete position is taken into account before purging entries. 

It looks like the problem arrises when there's isn't at least one entry that is older than the current mark delete position in the cache.  

I'd suggest to revisit the purging logic in this way:
- modify `addNewSnapshot` and add a 2nd parameter which is the current mark delete position
- always keep the newest entry that is before the current mark delete position when purging entries, all older entries can be purged
- if the cache remains full after after doing this, remove a single entry in the cache so that the new entry could be added. 
  - keep the position of the last removed entry so that it's possible to continue the purging algorithm in subsequent calls
  - if there's no previous last removed entry, purge the 2nd entry (assuming that the first entry is the newest entry before current mark delete position)
  - if there's a previous last removed entry, continue purging from next entry after the last removed position by first skipping one entry and then removing the 2nd entry
  - if there's no more entries to remove, start removing from the beginning.

This purging logic should always result in making it possible to add a new entry. Since every 2nd entry is removed, it will result in "sampling" so that when the mark delete position finally advances, it advances to the most recent position.

There's a possibility to increase `replicatedSubscriptionsSnapshotMaxCachedPerSubscription` parameter to improve the resolution, if that's desirable.

Perhaps the intention of your timestamp based approach is already to achieve something similar?

lhotari:
> The intention of timestamp based is to achieve this purpose，the key is when cache is full, how to update the cache, there is no perfect algorithm to solve this problem.
> 
> I recommend using Median-based eviction for simplicity, try to make the cache an arithmetic progression in time, because for the shared mode subscription, there will be individual unconfirmed messages, presenting a very jumpy situation.

You are right about this. The added comments in the code make it easier to understand the intention of the logic. My previous comment about taking the mark deletion position into account in adding snapshots didn't make much sense after rethinking.
I'll review again.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -33,29 +35,96 @@
...
+                                               int snapshotFrequencyMillis) {
         this.subscription = subscription;
         this.snapshots = new TreeMap<>();
+        this.positionToTimestamp = new TreeMap<>();
         this.maxSnapshotToCache = maxSnapshotToCache;
+        this.snapshotFrequencyMillis = snapshotFrequencyMillis;
     }
 
-    public synchronized void addNewSnapshot(ReplicatedSubscriptionsSnapshot snapshot) {
+    public synchronized void addNewSnapshot(ReplicatedSubscriptionsSnapshot snapshot, long publishTime) {
```
lhotari, line 52: Please add a javadoc comment to this method and explain that `publishTime` is the `publishTime` of the `ReplicatedSubscriptionsSnapshot` marker message (entry). Since the `publishTime` originates from the same broker or at least from the same cluster, it's assumed that the time is reliable and clock skew is not a problem.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -33,29 +35,96 @@
 public class ReplicatedSubscriptionSnapshotCache {
     private final String subscription;
     private final NavigableMap<Position, ReplicatedSubscriptionsSnapshot> snapshots;
+    // Used to record the timestamp of snapshots location, which will be used to adjust cache update frequency later.
+    private final NavigableMap<Position, Long> positionToTimestamp;
```
lhotari, line 39: instead of adding a separate data structure for timestamps, add a new Java record to hold `ReplicatedSubscriptionsSnapshot` and the timestamp. Modify the `snapshots` to use this type of value.

lhotari: After making this change, I'll review the code again since the current solution of having both `snapshots` and `positionToTimestamp` makes the code harder to understand. It will be easier to add more review comments after this is addressed.

lhotari: @liangyepianzhou You didn't seem to follow the advice about adding a Java record. Since ReplicatedSubscriptionsSnapshot is a protobuf message, you shouldn't be adding fields there since the `publishTime` is outside of it. Adding a field just to be able to solve holding the timestamp wouldn't be the correct approach.

Instead, add a Java `record` that is just private to this `ReplicatedSubscriptionSnapshotCache`:

```java
private record SnapshotEntry(ReplicatedSubscriptionsSnapshot snapshot, long timestamp) {
}
```

pulsar-common/src/main/proto/PulsarMarkers.proto
```diff
@@ -60,6 +60,7 @@ message ReplicatedSubscriptionsSnapshot {
     required string snapshot_id = 1;
     optional MarkersMessageIdData local_message_id = 2;
     repeated ClusterMessageId clusters = 3;
+    optional int64 markers_timestamp = 4;
```
lhotari, line 63: This doesn't make sense. Comment added to https://github.com/apache/pulsar/pull/24300#discussion_r2090386514

liudezhi2098: Is the above method simpler and more intuitive?

lhotari: It's changing the wireprotocol just to be able to add an implementation specific field, that doesn't make any sense. Please refactor to use a Java record so that we can proceed with the review. It's not much work to do that. See https://github.com/apache/pulsar/pull/24300#discussion_r2090386514

lhotari: In mixed version clusters after upgrades, there would be problems and addressing that in code will add more complexity since you'd have to check whether the field is set.

liudezhi2098: The versions are compatible, but what you said makes sense.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -32,32 +34,114 @@
...
+            }
+            snapshots.put(position, snapshotEntry);
         }
     }
 
+    /**
+     * Find the Position in NavigableMap according to the target index.
+     */
+    private Position findPositionByIndex(int targetIndex) {
+        Iterator<Map.Entry<Position, SnapshotEntry>> it = snapshots.entrySet().iterator();
```
lhotari, line 133: since this doesn't use the value, `snapshots.keySet().iterator()` could be used directly.

lhotari (commented):
Great work, mainly comments about comments. The 2nd rule is skipping to add entries, `timeSinceLastSnapshot < timeWindowPerSlot`, seems risky to add since it could have surprising consequences. I think it would be better to remove that.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -32,32 +34,114 @@
...
+        }
 
-        snapshots.put(position, snapshot);
+        // The time difference between the previous position and the earliest cache entry
+        final long timeSinceFirstSnapshot = publishTime - snapshots.firstEntry().getValue().timestamp();
+        // The time difference between the previous position and the lately cache entry
+        final long timeSinceLastSnapshot = publishTime - snapshots.lastEntry().getValue().timestamp();
+        // The time window length of each time slot, used for dynamic adjustment in the snapshot cache.
+        // The larger the time slot, the slower the update.
+        final long timeWindowPerSlot = timeSinceFirstSnapshot / snapshotFrequencyMillis / maxSnapshotToCache;
```
lhotari, line 95: this is a bit hard to grasp, what the "time slot" concept is here. 
Let's say, timeSinceFirstSnapshot is 25000 ms, snapshotFrequencyMillis is 1000ms and maxSnapshotToCache is 10, it would result in `2`. What's the point of this?
With low values, this would be close to 0, I guess. This is also why I think this is just unnecessary complexity.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -32,32 +34,114 @@
...
+        } else if (position.compareTo(snapshots.lastKey()) < 0) {
+            // Case 3: Reset cursor If the position is in the middle, delete the cache after that position
+            while (position.compareTo(snapshots.lastKey()) < 0) {
+                snapshots.pollLastEntry();
+            }
+            snapshots.put(position, snapshotEntry);
+        }
+        // Time-based eviction conditions
+        // timeSinceLastSnapshot < snapshotFrequencyMillis, keep the same frequency
+        // timeSinceLastSnapshot < timeWindowPerSlot, implementing dynamic adjustments
```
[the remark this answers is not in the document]
liudezhi2098: timeWindowPerSlot This is reasonable. For example, the mark-delete position is updated very slowly, but the Dispatcher is still pushing messages to the consumer. It is valuable to gradually reduce the cache update frequency.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
liudezhi2098: What if timeSinceFirstSnapshot is 25 minutes?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
```diff
@@ -32,32 +34,114 @@
...
+        final long timeSinceLastSnapshot = publishTime - snapshots.lastEntry().getValue().timestamp();
+        // The time window length of each time slot, used for dynamic adjustment in the snapshot cache.
+        // The larger the time slot, the slower the update.
+        final long timeWindowPerSlot = timeSinceFirstSnapshot / snapshotFrequencyMillis / maxSnapshotToCache;
 
-        // Prune the cache
-        while (snapshots.size() > maxSnapshotToCache) {
-            snapshots.pollFirstEntry();
+        if (position.compareTo(snapshots.firstKey()) < 0) {
+            // Case 2: Reset cursor if position precedes first entry
```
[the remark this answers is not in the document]
liudezhi2098: There will be no problem understanding it, refer to pulsar admin, which can explain it more directly
```
bin/pulsar-admin topics  reset-cursor  
    reset-cursor      Reset position for subscription to a position that is 
      Usage: reset-cursor [options] persistent://tenant/namespace/topic
          -e, --exclude-reset-position
            Exclude the reset position, start consume messages from the next 
            messageId to reset back to ('latest', 'earliest', or 
            Subscription to reset position on
            time in minutes to reset back to (or minutes, hours, days, weeks 
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
lhotari: > What if timeSinceFirstSnapshot is 25 minutes? The goal is that if timeSinceFirstSnapshot becomes longer, the update frequency should be lower.

This doesn't seem to be a realistic case. I disagree that the update frequency should become lower. That's exactly my point that if there's a long delay, it will be enforced going forward. I think it's easier to make progress in this PR by removing this rule and adding it later if there's a specific reason to do so.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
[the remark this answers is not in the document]
lhotari: Commented in https://github.com/apache/pulsar/pull/24300#discussion_r2092417336, please remove the 2nd rule.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionSnapshotCache.java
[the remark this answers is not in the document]
lhotari: Please rephrase the sentence in that case. The wording is now imperative" "reset cursor" that's like it's explaining that the code below is resetting the cursor. The way to address this in the comment is to add "After", "After cursor reset position precedes first entry", etc.

lhotari:
> I recommend using Median-based eviction for simplicity, try to make the cache an arithmetic progression in time, because for the shared mode subscription, there will be individual unconfirmed messages, presenting a very jumpy situation.

I don't see how the median based eviction could make sense. After the cache is filled up, when the median entry is removed and a new entry is added, and this repeats, the result will be that only entries after the median entry will be evicted (assuming no other events happen in between). Eventually there will be a large gap between the 2 entries in the middle.

Since time is already considered in the algorithm, it seems that an alternative approach would be to evict the entry with the shortest time distance to it's adjacent entry. Makes sense?

lhotari:
@liudezhi2098 Are you planning to continue working on this? I think that this is a really great improvement to address a long time issue with replicated subscriptions.

lhotari:
@liudezhi2098 There's also a long-standing issue #10054 which is addressed by #16651. I have updated PR 16651 description, rebased it and revisited it slightly. Please review

liudezhi2098:
> @liudezhi2098 Are you planning to continue working on this? I think that this is a really great improvement to address a long time issue with replicated subscriptions.

yes , I  will continue working on this

lhotari:
@liudezhi2098 Any updates on this PR?

lhotari:
@liudezhi2098 Any updates on this PR?

Ksnz:
It seems that the best way to store ReplicatedSubscriptionsSnapshots is in something like an increasing backoff queue:
the last one — the most recent,
the next — plus `snapshotFrequencyMillis`,
then `snapshotFrequencyMillis`^2,
then `snapshotFrequencyMillis`^3, and so on.
The oldest one (the head) should remain untouched until it is consumed.

It also seems that the solution could be simplified around `ReplicatedSubscriptionSnapshotCache`, and delayed messages would be handled much better.
If the PR is stale, I may reopen a new one :)

lhotari:
> It seems that the best way to store ReplicatedSubscriptionsSnapshots is in something like an increasing backoff queue: the last one — the most recent, the next — plus `snapshotFrequencyMillis`, then `snapshotFrequencyMillis`^2, then `snapshotFrequencyMillis`^3, and so on. The oldest one (the head) should remain untouched until it is consumed.
> 
> It also seems that the solution could be simplified around `ReplicatedSubscriptionSnapshotCache`, and delayed messages would be handled much better. If the PR is stale, I may reopen a new one :)

@Ksnz That could be useful if time would be the only contributing factor.

One detail that has been missed before (in the comments of this PR) is the number of entries between 2 different snapshots. Optimally, the cache would keep snapshots with equal number of entries between the snapshots. When there's a burst of messages to the topic, there could be a lot of messages between a single snapshot that happens once per second.
When the snapshot cache fills up, I'd assume that the snapshot entry with the shortest distance to it's adjacent entries should be evicted. WDYT?

Ksnz:
>WDYT? 

There may be several strategies to address this. And the cache-eviction policy should probably be aware of the namespace, topic, or even individual subscription.

The main problem in my current project is that  `markDelete` never advances on the delayed-message topic replica. In cluster A the `markDelete` cursor moves only rarely, at unpredictable moments, and over an unpredictable range. This causes the Cluster B replica backlog to grow.

To mitigate this, we want to keep at least  the oldest snapshot to ensure that we do not lose any `markDelete` update.

The following strategies come to mind:

1. **Equal time intervals** (current behavior) — evict from the head. May be used as a default, to not break expected behaviour.
2. **Equal distance intervals** (new strategy) — evict from the head. Requires extra code outside of `ReplicatedSubscriptionSnapshotCache,` because `Position` itself is not trusted source to calculate distance.
3. **Increasing time intervals (from tail to head)** (new strategy) — evict from the middle, keep the head. 
4. **Increasing distance intervals (from tail to head)** (new strategy) — evict from the middle, keep the head. Again extra code.
5. **Increasing time intervals (from head to tail)** (new strategy) — evict from the middle, keep the head. 
6. **Increasing distance intervals (from head to tail)** (new strategy) — evict from the middle, keep the head. Requires additional code .

Policies 5 or 6 fit our needs for delayed-topics. But any distance-based strategies require a lot changes to be implemented. 

For me the simple way is to implement 3 and 5 strategies. 
3 for fast-consumers but sometimes get stuck due to burst of messages. To not loss sync at all.
5 for delayed-messages consumers, where `markDelete` moves intermittently and unpredictably.

And an option to pick strategy per namespace. 
Ability to pick strategy on consumer configuration requires a cascade of code changes.

lhotari:
> The main problem in my current project is that `markDelete` never advances on the delayed-message topic replica. In cluster A the `markDelete` cursor moves only rarely, at unpredictable moments, and over an unpredictable range. This causes the Cluster B replica backlog to grow.

@Ksnz Yes, this seems like a problem case where the issue is caused by snapshot cache overflowing.

> 2. **Equal distance intervals** (new strategy) — evict from the head. Requires extra code outside of `ReplicatedSubscriptionSnapshotCache,` because `Position` itself is not trusted source to calculate distance.

There's existing org.apache.bookkeeper.mledger.ManagedLedger#getNumberOfEntries(com.google.common.collect.Range<org.apache.bookkeeper.mledger.Position>) available for the calculation. The reason why I think it's more useful than time based strategies is that the rate of consuming usually remains about the same while producing can have high bursts.

> 3. **Increasing time intervals (from tail to head)** (new strategy) — evict from the middle, keep the head.
> 4. **Increasing distance intervals (from tail to head)** (new strategy) — evict from the middle, keep the head. Again extra code.
> 5. **Increasing time intervals (from head to tail)** (new strategy) — evict from the middle, keep the head.
> 6. **Increasing distance intervals (from head to tail)** (new strategy) — evict from the middle, keep the head. Requires additional code .

"evict from the middle" sounds great, but it's not practical due to the variable producing speed. It could work for topics where producing speed remains constant.

> And an option to pick strategy per namespace. Ability to pick strategy on consumer configuration requires a cascade of code changes.

Having too many options to pick from is not a great idea. It should be possible to solve this snapshot caching issue in a generic way that works for all cases sufficiently. Based on the previous experiences, I believe that it should be based on distances between positions and the cache should optimally keep entries with equal distances between the entries. 
In addition, there might be a need to simply increase the default snapshot cache size, snapshot timeout time and/or make it configurable with namespace level policies.

btw. I've previously created a mermaidjs diagram to show the main interactions in replicated subscriptions: https://gist.github.com/lhotari/96fda511a70d7de93744d868b4472b92
That might be helpful to understand the flow. However, it seems that you have already found out how it works. :)

Ksnz:
We could introduce an additional option similar to `replicatedSubscriptionsSnapshotFrequencyMillis`, but for message count – for example  `replicatedSubscriptionsSnapshotFrequencyMessageGap` (the naming can be discussed later, along with the semantics of how the two properties interact), and send `ReplicatedSubscriptionsSnapshotRequest` not only based on time but also after a certain number of messages have been received.
This way, we could achieve intervals between snapshots that are equal in terms of message count.

However, I still don’t see how this would help when `markDelete` can be far beyond the horizon of the current cache.

An additional option to solution would be to store the `ReplicatedSubscriptionSnapshotCache` somewhere in an **off-heap storage**, for example in **RocksDB**, evicting entries that are older than the `markDelete` horizon. 

This is the generic approach to keep both slow/delayed message consumers and fast backlog-tip consumers subscriptions aligned across clusters and to minimise inconsistency lag.

lhotari:
> We could introduce an additional option similar to `replicatedSubscriptionsSnapshotFrequencyMillis`, but for message count – for example `replicatedSubscriptionsSnapshotFrequencyMessageGap` (the naming can be discussed later, along with the semantics of how the two properties interact), and send `ReplicatedSubscriptionsSnapshotRequest` not only based on time but also after a certain number of messages have been received. This way, we could achieve intervals between snapshots that are equal in terms of message count.

Yes, a position interval based approach could be useful in ensuring that there would be more frequent snapshots when a lot of messages have been produced. One detail is that the snapshotting isn't deterministic so it wouldn't result in equal intervals between the snapshots. 
* Since the snapshot require a request-response flow, it's not deterministic. The remote side might be offline or the geo-replication processing of the messages might be slow or vary. There wouldn't be a stable interval between the completed snapshots because of these differences.
* There could also be more than 2 clusters in geo-replication and that requires [2 rounds of request-response](https://github.com/apache/pulsar/blob/1ce7855c9424b23ac357cfd1cfe89bdb6e22ea57/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsSnapshotBuilder.java#L109-L117). This increases variance. 

> However, I still don’t see how this would help when `markDelete` can be far beyond the horizon of the current cache.

The cache would keep snapshots with equal position intervals. I don't see why it wouldn't help. 
Increasing the snapshot cache size would also help. It's not that much memory that it requires. Optimizing the memory usage further would be useful. One optimization target would be to de-duplicate the java.lang.String instances of the cluster ids that are part of the snapshots. `org.apache.pulsar.common.util.StringInterner` class can be used for deduplication.

> An additional option to solution would be to store the `ReplicatedSubscriptionSnapshotCache` somewhere in an **off-heap storage**, for example in **RocksDB**, evicting entries that are older than the `markDelete` horizon.

This shouldn't be needed since the amount of memory that the snapshots consume can be optimized. It should be possible to store millions of snapshots with fairly low heap memory usage after optimizing the data structures.

Ksnz:
Okay, let’s go back to the square one. We have this description:

>  When you enable replicated subscriptions, you're creating a consistent distributed snapshot to establish an association between message ids from different clusters. The snapshots are taken periodically. The default value is `1 second`. It means that a consumer failing over to a different cluster can potentially receive 1 second of duplicates. You can also configure the frequency of the snapshot in the `broker.conf` file.


In my opinion, a one-second window is an acceptable limitation, even if a burst of messages occurs during that second it is still **one second**.

The problem is that we evict old snapshots from the head of the queue (the oldest positions), so **slow consumers** can no longer move their `markDelete` position because the required snapshot is already gone.

**_Perhaps_** if we added **an option** to make the cache act as a buffer (skipping all new entries once it’s full) it would help slow subscriptions stay in sync. Moreover, it would be **_generic approach_**

After a successful advanced `markDelete` position, the buffer would be cleared up to some point and refilled with new `ReplicatedSubscriptionsSnapshots`.

The queue that is aligned to a `markDelete` position is more useful for synchronization than a queue that follows the tip of the topic. 

A remaining issue is that a slow consumer can suddenly advance its `markDelete` position by a large amount, wiping the entire buffer. The next snapshots will then be captured far ahead of the consumer’s new `markDelete`, leaving a gap.

But we can combine the best of both worlds. By adding the
`private final NavigableMap<Position, ReplicatedSubscriptionsSnapshot> tailSnapshots; `
and 
`private final NavigableMap<Position, ReplicatedSubscriptionsSnapshot> headSnapshots; `
to a `ReplicatedSubscriptionSnapshotCache`

It's not a perfect solution. But it's less evil than rewriting everything and still better than the current one.

lhotari:
> In my opinion, a one-second window is an acceptable limitation, even if a burst of messages occurs during that second it is still **one second**.

That's true, but the end result isn't that the distance between 2 different snapshots would be with 1 second difference in rate since it's not deterministic due to the request-response required for creating the snapshot.

> The problem is that we evict old snapshots from the head of the queue (the oldest positions), so **slow consumers** can no longer move their `markDelete` position because the required snapshot is already gone.

I agree that it doesn't make sense to evict the head of the queue.

> **_Perhaps_** if we added **an option** to make the cache act as a buffer (skipping all new entries once it’s full) it would help slow subscriptions stay in sync. Moreover, it would be **_generic approach_**

The problem with that approach is that there would later be problems in updating the mark delete position since there would be a long duration where there wouldn't be snapshots at all. That's the reason why I think that it makes more sense to have an eviction algorithm that keeps snapshots with similar distances between the first entry and the last entry in the queue. It's not very hard to implement such an eviction algorithm. Together with a much larger maximum cache size (or even dynamic max size), it's possible to address the problem that has been caused by the snapshot cache overflowing.

> After a successful advanced `markDelete` position, the buffer would be cleared up to some point and refilled with new `ReplicatedSubscriptionsSnapshots`.

Yes, it will eventually catch up as long as the head of the queue isn't deleted. However, there's a potential to fix the complete problem of having a long lag in the update when the eviction is improved and there are a lot more snapshots available.

> The queue that is aligned to a `markDelete` position is more useful for synchronization than a queue that follows the tip of the topic.

yes

> A remaining issue is that a slow consumer can suddenly advance its `markDelete` position by a large amount, wiping the entire buffer. The next snapshots will then be captured far ahead of the consumer’s new `markDelete`, leaving a gap.

yes. that's what could be addressed with the improved eviction algorithm that would keep snapshots with about equal distances from the head to the tail.

> But we can combine the best of both worlds. By adding the `private final NavigableMap<Position, ReplicatedSubscriptionsSnapshot> tailSnapshots; ` and `private final NavigableMap<Position, ReplicatedSubscriptionsSnapshot> headSnapshots; ` to a `ReplicatedSubscriptionSnapshotCache`
> 
> It's not a perfect solution. But it's less evil than rewriting everything and still better than the current one.

true

lhotari:
@Ksnz I have created draft PR #25044 based on my previous comments. It's easier to continue the discussion on that PR when there's concrete code. I haven't yet done any performance optimization and it's not great yet. Adding 1 million snapshots in the unit test takes about 18 seconds on Mac M3, so there's a need to optimize it. I'll take a closer look later by adding a JMH benchmark.

lhotari:
Replaced by #25044

PR #8544, opened 2020-11-12
Consumer filtering

andrekramer1:
Fixes #3302 (a request for consumer filtering in the Pulsar broker)

### Motivation

At Software AG, we have prototyped property (key / value) based message filtering on consumer subscriptions in the Pulsar broker. This feature was also requested as Issue 3302 but the mitigation suggestion of using Pulsar Functions, which imply management and runtime overheads and an extra topic, or the alternative of publishing to multiple (sub)topics are not always adequate mitigations to lack of simple consumer subscription filtering.

In Pulsar, consumers receive all messages published on a topic, even if only interested in messages with certain properties. Always delivering all messages and having the consumer client application filter means all messages must be sent over the network, processed and acknowledged remotely. Other message brokers and messaging standards (such as Java's JMS message selectors) typically allow a filter to be specified on client subscription, allowing the message broker to filter on behalf of the client subscriber and deliver a subset of the messages over the wire.
In some cases, a separate topic can be used for each subscription filter, but if there are a very large number of possible filters then capacity as well as the management and processing overheads of topics (or any Pulsar Functions) are an issue. It therefore seems sensible to extend consumer subscriptions to perform an additional level of selection that is not directly driven by the publisher and does not require additional Functions and/or topics to be created for consumers.

Our proposal is to allow the publisher to "tag" messages with one or more additional message properties that can be used to selectively filter by. The consumer (on subscription) should be able to specify which tags to filter on, specified by special consumer “meta data” (using the meta data key value pairs in consumer properties).

The implementation is itself configurable, allowing both for no filtering and for alternative property-based schemes to be implemented and configured in Pulsar brokers, using a new configuration time "ConsumerFilter" dynamic factory mechanism. Our default "tag" based scheme, is explicit about which properties are used for filtering so as to more safely use existing message properties and consumer meta data. Two types of meta data tags are provided to allow both simple AND (all match) as well as OR (any match) filter expressions on message properties.

Our implemented design enables broker side "tag" based filtering of messages on selected message properties by allowing consumers to specify a subset of message properties as being “tags” to filter on, using existing consumer (meta data) property API (and underlying protocol). Message properties which have a key beginning with the string "tag" are matched by value against the consumer subscription properties that are specified on subscription with meta data string keys starting with either "anytag" or "alltag". What follows “tag”, “anytag”, “alltag” textually is immaterial except to allow multiple key /  values to be enumerated (e.g. “tag1”, “tag2” etc).

[...]

PR #7629, opened 2020-07-22
[][broker/client] Initial implementation of message filtering

seeday:
### Motivation

Broker-side filtering of messages can reduce the network load significantly when running certain types of highly-filtered jobs. Example usecases are flink filter-pushdown, displaying data over websocket, and other network-limited actions.

So far, the list of available filters is pretty small, but this framework should make writing new ones relatively easy.

### Modifications

I've modified the pulsar protocol such that clients can send a spec for filtering to the broker. The broker's Consumer class builds a filter from the arguments sent as well as the class name. I've gone with reflection because that makes for easy extension and customization of the available filters. 

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change added tests and can be verified as follows:
 - the new tests in MessageFilteringTest/SimpleTypedMessageProducerConsumerTest should pass
 - all other tests should be unaffected by the changes

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API: (**yes** / no)
      - a new function is added to the consumer to allow for constructing a message filter
  - The schema: (yes / **no** / don't know)
  - The default values of configurations: (yes / **no**)
  - The wire protocol: (**yes** / no)
      - a new field was added to commandsubscribe to send info about the message filter
  - The rest endpoints: (yes / **no**)
  - The admin cli options: (yes / **no**)
  - Anything that affects deployment: (yes / **no** / don't know)

### Documentation

  - Does this pull request introduce a new feature? (**yes** / no)
  - If yes, how is the feature documented? (not applicable / docs / **JavaDocs** / not documented)
    - once this is in a more stable state, an entry in the actual docs is probably necessary. 
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

Issue #23466, opened 2024-10-16
[Enchancement] Improve Pulsar Broker cache defaults to get better out-of-the-box performance


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Pulsar's defaults for broker cache aren't good for performance. This leads to wasted CPU and unnecessary network transfer costs in cloud environments.

### Solution

Here are some settings which would be better defaults.

- `maxMessagePublishBufferSizeInMB` - not broker cache related, but it's necessary to set it to an explicit value when fine-tuning broker cache settings so that direct memory OOM can be avoided. Default is 50% of direct memory. Set to `500`
- `managedLedgerCacheSizeMB` - the default is 20% of direct memory. It's better to set it to an explicit value to avoid direct memory OOM. Set to `512`
- `managedLedgerMaxReadsInFlightSizeInMB` - this feature is disabled by default. It's useful for avoiding direct memory OOM. Set to `100`
- `managedLedgerCacheEvictionTimeThresholdMillis` - the default `1000` is too low. Set to `10000`
- `managedLedgerCacheEvictionIntervalMs` - the default `10` is too low. Set to `5000` to avoid spending a lot of CPU with cache eviction.
- `managedLedgerMinimumBacklogCursorsForCaching` - the default `0` disables backlog cursors (catch-up read) caching. Set to `3`
- `managedLedgerMinimumBacklogEntriesForCaching` - the default `1000` is way too high. Set to `1`
- `managedLedgerMaxBacklogBetweenCursorsForCaching` - the default `10000` is way too low. Set to `2147483647` to disable the limit completely.

Sample settings for broker cache tuning:

yaml format:
```yaml
  maxMessagePublishBufferSizeInMB: 500
  managedLedgerCacheSizeMB: 512
  managedLedgerMaxReadsInFlightSizeInMB: 100
  managedLedgerCacheEvictionTimeThresholdMillis: 10000
  managedLedgerCacheEvictionIntervalMs: 5000
  managedLedgerMinimumBacklogCursorsForCaching: 3
  managedLedgerMinimumBacklogEntriesForCaching: 1
  managedLedgerMaxBacklogBetweenCursorsForCaching: 2147483647
```

`broker.conf` format:
```properties
maxMessagePublishBufferSizeInMB=500
managedLedgerCacheSizeMB=512
managedLedgerMaxReadsInFlightSizeInMB=100
[...]

PR #24305, opened 2025-05-15
[feat][pip] PIP-419: Support consumer message filter on broker-side for specific subscription

AuroraTwinkle:
Fixes #xyz

Main Issue: https://github.com/apache/pulsar/pull/8544
https://github.com/apache/pulsar/pull/7629

PIP: #xyz 

### Motivation
Currently, Pulsar does not have the ability to filter messages by subscription, this pip try to implement message filtering on the pulsar broker side, enrich the pulsar ecosystem, save bandwidth, and help RocketMQ users switch to Pulsar more smoothly. And, I've noticed two similar PRs https://github.com/apache/pulsar/pull/8544 and https://github.com/apache/pulsar/pull/7629, but neither was approved, so I want to try to get this work implemented.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: 
https://github.com/AuroraTwinkle/pulsar/pull/6

### DIFF
--- a/pip/pip-419.md
+++ b/pip/pip-419.md
@@ -163,7 +163,7 @@ public class ConsumerExample {
 
 ### Public API
 
-1. Add an optional config `filterExpression` in `ConsumerBuilder`
+1. Add an optional config `filterExpression` in [ConsumerBuilder](https://github.com/apache/pulsar/blob/965ef5c14c93ca896ef4c8f34520066285fcf047/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java)
 
 ```java
 ConsumerBuilder<T> filterExpression(FilterExpression filterExpression);
--- a/pip/pip-419.md
+++ b/pip/pip-419.md
@@ -1,0 +1,184 @@
+# 消息过滤
+
+# PIP-419: Support consumer message filter on broker-side for specific subscription
+
+Implementation PR: 
+
+# Background knowledge
+
+In my company, we developed a centralized message broker service, which starts multiple subscriptions to consume messages from multiple topics, and then forwards them to business clients after certain processing. Our users have message filtering requirements, that is, when messages are written to topics with different tags, multiple different subscriptions will be started when consuming, and each subscription will only consume messages with the tags it cares about.
+
+What we are doing is replacing all RocketMQ with Pulsar,  but I found that consumers cannot filter messages based on certain message tags or properties, just like [message filtering](https://rocketmq.apache.org/docs/featureBehavior/07messagefilter) in RocketMQ.
+
+From the user's perspective, when using a Pulsar Consumer, we have two main options to consume messages:
+
+1. Pull mode, by calling `consumer.receive()`(or `consumer.receiveAsync()`)
+
+```java
+public class ConsumerExample {
+    public static void main(String[] args) throws PulsarClientException {
+        PulsarClient pulsarClient = PulsarClient.builder()
+                .serviceUrl("pulsar://localhost:6650")
+                .build();
+        Consumer<Long> consumer = pulsarClient.newConsumer(Schema.INT64)
+                .topic("persistent://public/default/my-topic")
+                .subscriptionName("my-subscription")
+                .subscribe();
+        do {
+            Message<Long> message = consumer.receive();
+            consumer.acknowledge(message);
+        } while (true);
+
+    }
+}
+
+```
+
+1. Push mode, by registering a `MessageListener` interface, when building the Consumer.
+
+```java
+public class ConsumerExample {
+    public static void main(String[] args) throws PulsarClientException {
+        PulsarClient pulsarClient = PulsarClient.builder()
+                .serviceUrl("pulsar://localhost:6650")
+                .build();
+        Consumer<Long> consumer = pulsarClient.newConsumer(Schema.INT64)
+                .topic("persistent://public/default/my-topic")
+                .subscriptionName("my-subscription2")
+                .messageListener((consumer, message) -> {
+                    // process message
+                    consumer.acknowledgeAsync(message);
+                })
+                .subscribe();
+    }
+}
+
+```
+
+However, I found that they do not have similar message filtering configurations, which means that a subscription will receive all messages from a topic and then filter them at the business layer if necessary. These meaningless messages may cause a lot of bandwidth waste. **Therefore, I think it is very meaningful to implement message filtering on the pulsar broker side.**
+
+# Motivation
+
+As [Background knowledge](https://www.notion.so/1f38504c2b898043982ce46d7d903c64?pvs=21) mentioned, implement message filtering on the pulsar broker side, enrich the pulsar ecosystem, save bandwidth, and help RocketMQ users switch to Pulsar more smoothly. And, I've noticed two similar PRshttps://github.com/apache/pulsar/pull/8544 and https://github.com/apache/pulsar/pull/7629, but neither was approved, so I want to try to get this work implemented.
+
+# Goals
+
+1. **Implement message filtering on the pulsar broker side.**
+
+## In Scope
+
+If this PIP is accepted, it will help users to easily filter messages when consumption. 
+
+# Detailed Design
+
+## Design & Implementation Details
+
+**The general idea is to filter based on the properties of the message at pulsar broker.** 
+
+In the following scenario, the producer adds two properties, messageType and messageSource, when writing messages. For example:
+
+```java
+ // message1
+ TypedMessageBuilder<byte[]> messageBuilder1 = producer.newMessage();
+ messageBuilder1.property("messageType", "file");
+ messageBuilder1.property("messageSource", "QQ");
+ 
+  // message2
+ TypedMessageBuilder<byte[]> messageBuilder2 = producer.newMessage();
+ messageBuilder2.property("messageType", "audio");
+ messageBuilder2.property("messageSource", "wechat");
+```
+
+If a subscription only wants to consume *audio* messages from w*echat*, how should it be implemented? 
+
+We can implement message filtering logic on pulsar broker at `org.apache.pulsar.broker.service.AbstractBaseDispatcher.filterEntriesForConsumer`
+
+![image.png](%E6%B6%88%E6%81%AF%E8%BF%87%E6%BB%A4%201f38504c2b898043982ce46d7d903c64/image.png)
+
+The filtering engine can choose the following two solutions. **In general, I prefer the SQL-92 solution, which should have better performance. Groovy may have slightly worse performance due to its dynamic characteristics.**
+
+### Solution 1 SQL-92
+
+Message filtering using [SQL-92](https://en.wikipedia.org/wiki/SQL-92) syntax mentioned in https://github.com/apache/pulsar/pull/8544#issuecomment-1064619792
+
+We can write the following SQL expression to filter messages, the pseudo code is as follow:
+
+```java
+// The filter expression passed by the pulsar consumer client to the pulsar broker
+String sql = "messageSource='wechat' AND messageType='audio'"
+
+// The pulsar broker uses the expression passed by the client to filter the messages and returns the messages that meet the conditions to the client.
+List filtedMessages = messages.stream()
+                    .filter(message -> {
+                        return SQL92Expression.evaluate(message)
+                    })
+                    .collect(Collectors.toList());
+```
+
+### Solution 2 Groovy
+
+> [Groovy](https://groovy-lang.org/) is a JVM-based object-oriented programming language under Apache. It can be used for object-oriented programming and can also be used as a pure scripting language to call Groovy scripts in Java.
+> 
+
+We can write the following Groovy expression to filter messages, the pseudo code is as follows:
+
+```java
+// The filter expression passed by the pulsar consumer client to the pulsar broker
+String groovyExpression = "it.properties.messageType.equals('audio') && it.properties.messageSource.equals('wechat')"
+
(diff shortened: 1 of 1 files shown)


AuroraTwinkle:
@lhotari  @StevenLuMT @liangyepianzhou I have submitted a pip about message filtering. Would you please review it when you have time? Looking forward to your suggestions. Thanks!

pip/pip-419.md
```diff
@@ -0,0 +1,182 @@
...
+```java
+public class ConsumerExample {
+    public static void main(String[] args) throws PulsarClientException {
+        PulsarClient pulsarClient = PulsarClient.builder()
+                .serviceUrl("pulsar://localhost:6650")
+                .build();
+        Consumer<Long> consumer = pulsarClient.newConsumer(Schema.INT64)
+                .topic("persistent://public/default/my-topic")
+                .subscriptionName("my-subscription")
+                .filterExpression(new FilterExpression("messageSource='wechat' AND messageType='audio'", FilterExpressionType.SQL92))
```
liangyepianzhou, line 148: Excellent work! I think this feature is indeed useful. However, I suggest that the filter expressions should be configured, modified, and deleted through the management/control plane (Pulsar Admin/CLI), rather than being set during consumer initialization. My reasoning is as follows:
1. **Practical Operational Considerations**: When maintaining a messaging service, we typically provide users with a management platform to configure subscription settings. If filter conditions are set during consumer initialization, any configuration changes would require propagating updates to user clients and forcing client reinitialization to take effect. This prevents direct updates via admin operations.

2. **Service Stability**: Broker-side configuration changes requiring client reinitialization could introduce consumption service instability.

3. **Subscription Consistency**: A single subscription may have multiple consumer instances. Setting filter conditions at consumer initialization risks inconsistent filtering rules across consumers under the same subscription, potentially causing rule overwrites (e.g., Consumer A sets Filter X while Consumer B sets Filter Y, leading to conflicts).

4. **UX Clarity**: Since clients already have a client-side filter mechanism (`ConsumerBuilder#messageFilter`), adding another `filterExpression` parameter might confuse users.

**Regarding point 3**: This mirrors real-world issues observed in RocketMQ's [message filtering](https://rocketmq.apache.org/docs/featureBehavior/07messagefilter). I've encountered multiple cases where tag-based filtering failed because different consumers in the same consumer group subscribed to conflicting tags, resulting in rule overrides. 

**Recommendation**: Store filter rules as subscription metadata (managed via Admin API/CLI) rather than per-consumer settings. This would:
- Ensure consistent filtering across all consumers in a subscription
- Allow dynamic rule updates without client restarts
- Align with existing subscription-level configurations
- Prevent rule conflicts between consumers

AuroraTwinkle: It is a good idea to configure the filtering rules through admin API. This can avoid the problems you mentioned, which is also the current problem of RocketMQ. I will refactor it together after receiving more comments.

lhotari:
> And, I've noticed two similar PRs #8544 and #7629, but neither was approved, so I want to try to get this work implemented.

@AuroraTwinkle For former context why such feature has been rejected in the past, this email thread is a good read: https://lists.apache.org/thread/k0hb33nz42f3603hvs03p7lmltjgj40z .

There simply hasn't been consensus that we'd like to move Pulsar together as a community to this direction.  
As a compromise, it was made possible to do broker side filtering using custom plugins with ["PIP 105: Support pluggable entry filter in Dispatcher".](https://github.com/apache/pulsar/issues/12269) and related subscription properties https://github.com/apache/pulsar/pull/15503. There's an implementation in Pulsar JMS using this approach: https://github.com/datastax/pulsar-jms/blob/master/pulsar-jms-filters/src/main/java/com/datastax/oss/pulsar/jms/selectors/JMSFilter.java

There are some performance challenges currently when using broker side filtering. However, many of the problems have been addressed over time and if we'd introduce the feature as a top level feature, there would be more motivation to address more of the challenges such as ones of the gaps in broker cache and handling of "ack holes" in Pulsar efficiently. 

The "ack hole" challenge is similar for delayed message delivery at large scale. Ack hole handling on broker side has been optimized with changes such as #9292, (must be enabled with `managedLedgerPersistIndividualAckAsLongArray=true` since default is `false` after #23759), increasing `managedLedgerMaxUnackedRangesToPersist`, setting `managedCursorInfoCompressionType=LZ4`, etc.

For broker cache, there's lot more to do. https://github.com/apache/pulsar/issues/23466 . 
One of the workarounds is `cacheEvictionByMarkDeletedPosition=true` for message filtering. I've have a WIP branch to address many of the problems, but haven't had a chance to push that forward.
It's just an example that once there's a requirement to address performance issues, those would get addressed over time. That's why I'd like to see message filtering being added as a top level feature in Pulsar in the long term. 

One of the possible compromises could be to make it an optional feature. In the Pulsar Java client API, we'd need to have a way to handle optional features in the builder interface itself so that the main builder interface for consumers doesn't get distracted by options that are only supported in certain configurations. One way to do this is with an "extensions" pattern. Where the builder would have a method `<T extends ConsumerExtensionBuilder> extension(Class<T> consumerExtensionBuilderClass)` which continues configuring the extension in a "nested" builder.

pip/pip-419.md
lhotari: It's already possible to handle filtering rules through the Admin API. The pulsar-admin CLI tool has a plugin model that allows adding also new commands. This approach is used in Pulsar JMS: https://github.com/datastax/pulsar-jms/blob/master/pulsar-jms-admin-ext/src/main/java/com/datastax/oss/pulsar/jms/cli/SubscriptionBaseCommand.java (more context in my other comment, https://github.com/apache/pulsar/pull/24305#issuecomment-2882812437)

lhotari: In Pulsar JMS, you could also provide the filter at a consumer level with consumer properties. If we were to add message filtering as a top level feature, it would be great to have full support in the clients too. In my other comment, I commented about this aspect

> One of the possible compromises could be to make it an optional feature. In the Pulsar Java client API, we'd need to have a way to handle optional features in the builder interface itself so that the main builder interface for consumers doesn't get distracted by options that are only supported in certain configurations. One way to do this is with an "extensions" pattern. Where the builder would have a method <T extends ConsumerExtensionBuilder> extension(Class<T> consumerExtensionBuilderClass) which continues configuring the extension in a "nested" builder.

pip/pip-419.md
```diff
@@ -0,0 +1,180 @@
...
+
+We can implement message filtering logic on pulsar broker at [org.apache.pulsar.broker.service.AbstractBaseDispatcher.filterEntriesForConsumer](https://github.com/apache/pulsar/blob/965ef5c14c93ca896ef4c8f34520066285fcf047/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java#L108)
+
+The filtering engine can choose the following two solutions. **In general, I prefer the SQL-92 solution, which should have better performance. Groovy may have slightly worse performance due to its dynamic characteristics.**
+
+### Solution 1 SQL-92
+
+Message filtering using [SQL-92](https://en.wikipedia.org/wiki/SQL-92) syntax mentioned in https://github.com/apache/pulsar/pull/8544#issuecomment-1064619792
+
+We can write the following SQL expression to filter messages, the pseudo code is as follow:
```
lhotari, line 100: Pulsar JMS reuses ActiveMQ's SQL-92 parser for filtering messages
https://github.com/datastax/pulsar-jms/blob/master/pulsar-jms-filters-common/src/main/java/com/datastax/oss/pulsar/jms/selectors/SelectorSupport.java

lhotari: The project has forked the filtering logic from ActiveMQ in this module: https://github.com/datastax/pulsar-jms/tree/master/activemq-filters

pip/pip-419.md
```diff
@@ -0,0 +1,180 @@
...
+// The pulsar broker uses the expression passed by the client to filter the messages and returns the messages that meet the conditions to the client.
+GroovyShell shell = new GroovyShell();
+Script script = shell.parse("return " + groovyExpression);
+List filtedMessages = messages.stream()
+                    .filter(message -> {
+                        binding.setVariable("it", message);
+                        script.setBinding(binding);
+                        return (Boolean) script.run();
+                    })
+                    .collect(Collectors.toList());
```
lhotari, line 134: This would be terrible from security perspective. Something like [Google CEL](https://github.com/google/cel-java) or [Spring Expression Language (SpEL)](https://docs.spring.io/spring-framework/reference/core/expressions.html) would be a better choice.

dao-jun:
I’m +1 with the motivation.

even though we have `EntryFilter` already, but it's not ready to use right out of the box.

In the details, maybe we can impl `EntryFilter`, and manage the rules via pulsar admin?

SpEL is a better choice than groovy scripts, it can easily lead to metadata area OOM because of caching classloader, I had encountered this problem in our business systems.

Besides, I hope you can consider that the ack hole problems, and it could comes incorrect stats/metrics(such as backlog).

AuroraTwinkle:
> > And, I've noticed two similar PRs #8544 and #7629, but neither was approved, so I want to try to get this work implemented.
> 
> @AuroraTwinkle For former context why such feature has been rejected in the past, this email thread is a good read: https://lists.apache.org/thread/k0hb33nz42f3603hvs03p7lmltjgj40z .
> 
> There simply hasn't been consensus that we'd like to move Pulsar together as a community to this direction. As a compromise, it was made possible to do broker side filtering using custom plugins with ["PIP 105: Support pluggable entry filter in Dispatcher".](https://github.com/apache/pulsar/issues/12269) and related subscription properties #15503. There's an implementation in Pulsar JMS using this approach: https://github.com/datastax/pulsar-jms/blob/master/pulsar-jms-filters/src/main/java/com/datastax/oss/pulsar/jms/selectors/JMSFilter.java
> 
> There are some performance challenges currently when using broker side filtering. However, many of the problems have been addressed over time and if we'd introduce the feature as a top level feature, there would be more motivation to address more of the challenges such as ones of the gaps in broker cache and handling of "ack holes" in Pulsar efficiently.
> 
> The "ack hole" challenge is similar for delayed message delivery at large scale. Ack hole handling on broker side has been optimized with changes such as #9292, (must be enabled with `managedLedgerPersistIndividualAckAsLongArray=true` since default is `false` after #23759), increasing `managedLedgerMaxUnackedRangesToPersist`, setting `managedCursorInfoCompressionType=LZ4`, etc.
> 
> For broker cache, there's lot more to do. #23466 . One of the workarounds is `cacheEvictionByMarkDeletedPosition=true` for message filtering. I've have a WIP branch to address many of the problems, but haven't had a chance to push that forward. It's just an example that once there's a requirement to address performance issues, those would get addressed over time. That's why I'd like to see message filtering being added as a top level feature in Pulsar in the long term.
> 
> One of the possible compromises could be to make it an optional feature. In the Pulsar Java client API, we'd need to have a way to handle optional features in the builder interface itself so that the main builder interface for consumers doesn't get distracted by options that are only supported in certain configurations. One way to do this is with an "extensions" pattern. Where the builder would have a method `<T extends ConsumerExtensionBuilder> extension(Class<T> consumerExtensionBuilderClass)` which continues configuring the extension in a "nested" builder.

Thank you Lari for your rigorous and detailed suggestions. Please give me some time to understand them carefully.

AuroraTwinkle:
> I’m +1 with the motivation.
> 
> even though we have `EntryFilter` already, but it's not ready to use right out of the box.
> 
> In the details, maybe we can impl `EntryFilter`, and manage the rules via pulsar admin?
> 
> SpEL is a better choice than groovy scripts, it can easily lead to metadata area OOM because of caching classloader, I had encountered this problem in our business systems.
> 
> Besides, I hope you can consider that the ack hole problems, and it could comes incorrect stats/metrics(such as backlog).

Thanks!

> I’m +1 with the motivation.
> 
> even though we have `EntryFilter` already, but it's not ready to use right out of the box.
> 
> In the details, maybe we can impl `EntryFilter`, and manage the rules via pulsar admin?
> 
> SpEL is a better choice than groovy scripts, it can easily lead to metadata area OOM because of caching classloader, I had encountered this problem in our business systems.
> 
> Besides, I hope you can consider that the ack hole problems, and it could comes incorrect stats/metrics(such as backlog).

Thanks.I will consider SpEL and ack hole problems.

lhotari:
> Thanks.I will consider SpEL and ack hole problems.

Besides SpEL, [Google's CEL](https://github.com/google/cel-java) would be a great option. DataStax Pulsar JMS [reuses ActiveMQ's SQL92 selectors for filtering](https://github.com/datastax/pulsar-jms/blob/master/pulsar-jms-filters-common/src/main/java/com/datastax/oss/pulsar/jms/selectors/SelectorSupport.java). I'd prefer Google's CEL for it's simplicity.

AuroraTwinkle:
> > Thanks.I will consider SpEL and ack hole problems.
> 
> Besides SpEL, [Google's CEL](https://github.com/google/cel-java) would be a great option. DataStax Pulsar JMS [reuses ActiveMQ's SQL92 selectors for filtering](https://github.com/datastax/pulsar-jms/blob/master/pulsar-jms-filters-common/src/main/java/com/datastax/oss/pulsar/jms/selectors/SelectorSupport.java). I'd prefer Google's CEL for it's simplicity.

OK, I'll do some research when I have free time. Thanks!

Denovo1998:
@AuroraTwinkle This feature looks interesting, are there any more updates?

PR #24313, opened 2025-05-16
[fix][broker] Fix potential deadlock when creating partitioned topic

nodece:
### Motivation

There is still a call to the synchronous method in `internalCreatePartitionedTopic`, which can lead to a deadlock under certain conditions.

### Modifications

- Replaced synchronous call to `getNamespacePolicies()` with `getNamespacePoliciesAsync()`.
- Replaced synchronous call to `getTopicPartitionList()` with `getTopicPartitionListAsync()`.
- Updated `internalCreatePartitionedTopic` to propagate the async flow properly.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
@@ -310,6 +310,10 @@ public abstract class AdminResource extends PulsarWebResource {
         }
     }
 
+    /**
+     * @deprecated Use {@link #getNamespacePoliciesAsync(NamespaceName)} instead.
+     */
+    @Deprecated
     protected Policies getNamespacePolicies(NamespaceName namespaceName) {
         try {
             Policies policies = namespaceResources().getPolicies(namespaceName)
@@ -538,6 +542,10 @@ public abstract class AdminResource extends PulsarWebResource {
                 .listPartitionedTopicsAsync(namespaceName, topicDomain);
     }
 
+    /**
+     * @deprecated Use {@link #getTopicPartitionListAsync()} instead.
+     */
+    @Deprecated
     protected List<String> getTopicPartitionList(TopicDomain topicDomain) {
         try {
             return getPulsarResources().getTopicResources().getExistingPartitions(topicName)
@@ -547,6 +555,10 @@ public abstract class AdminResource extends PulsarWebResource {
                     namespaceName.toString(), e);
             throw new RestException(e);
         }
+    }
+
+    protected CompletableFuture<List<String>> getTopicPartitionListAsync() {
+        return getPulsarResources().getTopicResources().getExistingPartitions(topicName);
     }
 
     protected void internalCreatePartitionedTopic(AsyncResponse asyncResponse, int numPartitions,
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
```diff
@@ -568,33 +580,38 @@ protected void internalCreatePartitionedTopic(AsyncResponse asyncResponse, int n
...
-                    try {
-                        policies = getNamespacePolicies(namespaceName);
-                    } catch (RestException e) {
-                        if (e.getResponse().getStatus() != Status.NOT_FOUND.getStatusCode()) {
-                            throw e;
+                .thenCompose((__) -> getNamespacePoliciesAsync(namespaceName).exceptionally(ex -> {
+                    Throwable unwrapped = FutureUtil.unwrapCompletionException(ex);
+                    if (unwrapped instanceof RestException) {
+                        RestException re = (RestException) unwrapped;
+                        if (re.getResponse().getStatus() == Status.NOT_FOUND.getStatusCode()) {
```
lhotari, line 587: this could use the newer Java syntax, `unwrapped instanceof RestException re && re.getResponse().getStatus() == Status.NOT_FOUND.getStatusCode()`

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
```diff
@@ -568,33 +580,37 @@ protected void internalCreatePartitionedTopic(AsyncResponse asyncResponse, int n
...
-                        long topicsCount =
-                                partitionedTopics.stream().filter(t ->
-                                        !pulsar().getBrokerService().isSystemTopic(TopicName.get(t))).count();
-                        if (topicsCount + numPartitions > maxTopicsPerNamespace) {
-                            log.error("[{}] Failed to create partitioned topic {}, "
-                                    + "exceed maximum number of topics in namespace", clientAppId(), topicName);
-                            throw new RestException(Status.PRECONDITION_FAILED,
-                                    "Exceed maximum number of topics in namespace.");
-                        }
+                        return getTopicPartitionListAsync().thenCompose(partitionedTopics -> {
```
lhotari, line 598: instead of `.thenCompose`, `.thenRun` could be used here since there's no asynchronous downstream calls beyond this point. A sign of this is `return CompletableFuture.completedFuture(null)`

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
```diff
@@ -568,33 +580,37 @@ protected void internalCreatePartitionedTopic(AsyncResponse asyncResponse, int n
...
-                            throw e;
+                .thenCompose((__) -> getNamespacePoliciesAsync(namespaceName).exceptionally(ex -> {
+                    Throwable unwrapped = FutureUtil.unwrapCompletionException(ex);
+                    if (unwrapped instanceof RestException re) {
+                        if (re.getResponse().getStatus() == Status.NOT_FOUND.getStatusCode()) {
+                            return null;
                         }
                     }
-
+                    throw new CompletionException(unwrapped);
```
lhotari, line 590: could use `throw FutureUtil.wrapToCompletionException(ex);` instead. There wouldn't be a need to create a new `CompletionException` when it's already a CompletionException.

lhotari (approved):

PR #24318, opened 2025-05-20
[fix][test] Fix flaky MultiTopicsReaderTest.testMultiNonPartitionedTopicWithRollbackDuration

hanmz:
### Motivation
image

The root cause of this problem is that the `reader` is not cleaned up in the unit test `shouldSupportCancellingReadNextAsync`

### Modifications
Add @Cleanup on `reader` in test `shouldSupportCancellingReadNextAsync`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari (approved):
LGTM

PR #24324, opened 2025-05-21
[fix][test] Fix flaky AutoScaledReceiverQueueSizeTest.testNegativeClientMemory

hanmz:
### Motivation
image

Since not every unit test is cleaned up before execution, the previous unit test may generate some dirty data, affecting the execution results of subsequent test cases.

### Modifications 
`setup` and `cleanup` is performed before each unit test is executed

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari (approved):
LGTM

PR #24325, opened 2025-05-21
[improve][build] Use project version with commit hash as docker tag

nodece:
### Motivation

The PR [#24168](https://github.com/apache/pulsar/pull/24168) introduced a regression where the `pulsar:${project.version}-${git.commit.id.abbrev}` Docker image cannot be pulled in the `pulsar-all` Dockerfile.

We consistently push both the `latest` and `${project.version}-${git.commit.id.abbrev}` tags to DockerHub. However, the only tag we require is the uniquely identifiable `${project.version}-${git.commit.id.abbrev}`. The `latest` tag can be ignored for our use case.

We recommend using the `${project.version}-${git.commit.id.abbrev}` format instead of `latest`, providing clearer traceability and avoiding ambiguity.

### Modifications

- Removed the custom Docker tag configuration, and `-Ddocker.tag=custom-tag` is the right path.
- Changed `docker.tag` value from `latest` to`${project.version}-${git.commit.id.abbrev}`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
Please check that test image builds are also consistent.

nodece:
@lhotari Done.

lhotari:
@nodece I remember having some problems with git-commit-id-plugin configuration in the past when someone builds the source code directly without git. Since docker.tag is now set at the top level, it would also be necessary to revisit the `git-commit-id-no-git` profiles in docker projects and possibly also revisit git-commit-id-plugin configuration.
Please test building the docker image directly from the source package to ensure that it also works.

nodece:
@lhotari Both are working fine.
```
# git env
mvn install -pl docker/pulsar,docker/pulsar-all -DskipTests \
    -Pmain,docker,docker-push \
    -Ddocker.platforms=linux/amd64,linux/arm64 \
    -Ddocker.organization=nodece \
    -Ddocker.image=pulsar

# no-git env
mvn install -pl docker/pulsar,docker/pulsar-all -DskipTests \
    -Pmain,docker,docker-push,git-commit-id-no-git \
    -Ddocker.platforms=linux/amd64,linux/arm64 \
    -Ddocker.organization=xxxx \
    -Ddocker.image=pulsar
```

lhotari:
> @lhotari Both are working fine.
> 
> ```
> # git env
> mvn install -pl docker/pulsar,docker/pulsar-all -DskipTests \
>     -Pmain,docker,docker-push \
>     -Ddocker.platforms=linux/amd64,linux/arm64 \
>     -Ddocker.organization=nodece \
>     -Ddocker.image=pulsar
> 
> # no-git env
> mvn install -pl docker/pulsar,docker/pulsar-all -DskipTests \
>     -Pmain,docker,docker-push,git-commit-id-no-git \
>     -Ddocker.platforms=linux/amd64,linux/arm64 \
>     -Ddocker.organization=xxxx \
>     -Ddocker.image=pulsar
> ```

What I mean is that the current `git-commit-id-no-git` profile configuration no longer makes sense when it's in the sub projects. It should automatically get activated at the top level so that it gets properly resolved in the docker.tag property value. I don't think that manually activating the profile would create a proper tag. It's also possible that git-commit-id-no-git profile isn't needed any more and there's a better way to configure the fallback value.

nodece:
@lhotari `git-commit-id-no-git` has been removed. Makes sense.

lhotari:
> @lhotari `git-commit-id-no-git` has been removed. Makes sense.

What is the default tag in this case?

lhotari:
btw. We could consider upgrading the deprecated git-commit-id-plugin to https://github.com/git-commit-id/git-commit-id-maven-plugin.

lhotari:
> btw. We could consider upgrading the deprecated `pl.project13.maven:git-commit-id-plugin` to `io.github.git-commit-id:git-commit-id-maven-plugin` https://github.com/git-commit-id/git-commit-id-maven-plugin.

The plugin is Java 11+, so for Maven Shade Java 8 tests to work, the plugin would have to be disabled when running on Java 8 in the existing `integration-test-java8` profile.

nodece:
> @lhotari
> 
> > What is the default docker.tag in this case?
> 
> `${project.version}-${git.commit.id.abbrev}` is default value.
> 
> > Please test the docker build from the source package so that we don't break that.
> 
> If no `.git`, I suggest that you append `-Ddocker.tag=${project.version}-custom`.
> 
> Do you have any idea? or keep `git-commit-id-no-git`?
> 
> > btw. We could consider upgrading the deprecated `pl.project13.maven:git-commit-id-plugin` to `io.github.git-commit-id:git-commit-id-maven-plugin` https://github.com/git-commit-id/git-commit-id-maven-plugin.
> 
> Ok, I will open a new PR to upgrade this plugin.

nodece:
@lhotari Ping

lhotari:
> Do you have any idea? Or keep `git-commit-id-no-git`? I prefer `-Ddocker.tag`.

we could move git-commit-id-no-git profile to the main pom.xml so that things would work also in the case where the build is run for the source distribution. That would be the most obvious solution. The profile should get activated only if docker.tag isn't set.

nodece:
@lhotari Done.

PR #23794, opened 2024-12-30
[improve][client] Enhance dynamic topic subscription management in PatternMultiTopicsConsumer

Denovo1998:
Main Issue: #xyz

### Motivation

When subscribing to topics using regex patterns, it’s possible to encounter scenarios where some topics process messages slower than others. This can lead to significant performance bottlenecks, as slower topics might block others. There’s a need to manage such situations dynamically to ensure efficient processing and fair resource allocation among topics.

Related discussions can be found in [this Reddit thread](https://www.reddit.com/r/ApachePulsar/comments/1fssbbn/roundrobin_between_wildcard_topics/), which highlights issues similar to what we aim to address.

When using `MultiTopicsConsumerImpl`, we can call this.

https://github.com/apache/pulsar/blob/5a3a1f169a7f90181bd5c213c8e9f479bc74f0f2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java#L1246-L1258

There is a test case that explains how to unsubscribe from a single topic.

https://github.com/apache/pulsar/blob/5a3a1f169a7f90181bd5c213c8e9f479bc74f0f2/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/TopicsConsumerImplTest.java#L660-L663

**This proposal seeks to integrate similar functionality into PatternMultiTopicsConsumerImpl, allowing dynamically remove or add specified topics.**

### Modifications

1. UnBlockTopics and BlockTopics Methods: In PatternMultiTopicsConsumerImpl, we introduce two methods: unBlockTopics and blockTopics.

- BlockTopics: Unsubscribe from topics by invoking onTopicsRemoved from TopicsChangedListener.

- UnBlockTopics: Resubscribe to topics by invoking onTopicsAdded from the same listener.

These changes will enable dynamic topic management, greatly improving the consumer’s ability to adapt to changing topic loads and enhancing overall system resilience against individual slow topics.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
[...]

PR #24328, opened 2025-05-21
[feat][pip] PIP-420: Provide ability for Pulsar clients to integrate with third-party schema registry service

gaoran10:
### Motivation

The Pulsar client is better has the ability to access third-party schema registry service to manage the schema (register schema, get schema, validate schema, etc.). The schema registry service can be an independent service, if using an external schema registry service, the Pulsar broker doesn't need to care about the schema of the messages while creating producer or adding consumer subscription.

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-420.md
+++ b/pip/pip-420.md
@@ -92,14 +92,13 @@ To integrate with external schema registries, users can:
 Unlike Pulsar, which uses **schema version** to identify schemas, many external schema registry systems use **schema ID** as the primary schema identifier.
 
 When integrating with external schema registries:
-- The `schemaVersion` filed in Pulsar message metadata is used in some places, **set to `-1` to flag the message is using external schema systems**.
-- The Pulsar client **needs to set the schema version to -1 and store in the message metadata**.
+- The `schemaVersion` will not indicate the read schema version, it points an external schema.
 - Instead, the external schema implementation can manage schema ID handling internally.
 - The schema ID can be embedded directly into the message payload by the external schema’s `encode` and `decode` methods.
 
 This approach allows external schema systems to fully control schema evolution and versioning without being constrained by Pulsar’s native schema versioning mechanism.
-This may impact some components that rely on schema version to deserialize messages, such as Pulsar Functions and Pulsar SQL,
-they will need to be updated to support the new schema type and handle external schemas appropriately.
+This may impact some components that rely on schema version to deserialize messages, such as Pulsar Functions and Pulsar SQL;
+they need to be updated to support setting schema properties, using the new schema type and handle external schemas appropriately.
 
 #### Example usage
 
--- a/pip/pip-420.md
+++ b/pip/pip-420.md
@@ -25,13 +25,14 @@ and governance workflows already centralized in external registries.
 ## Out Scope
 
 - Providing built-in implementations for third-party schemas.
+- Support `AutoProduceBytesSchema` and `AutoConsumeSchema`.
 - Migrating existing Pulsar-managed schemas to external schema registries.
 
 # High Level Design
 
 - Provide a mechanism to configure the Pulsar client to use either:
   - The existing Pulsar schema registry (default)
-  - A third-party schema registry implementation
+  - Third-party schema registry implementations
 
 # Detailed Design
 
@@ -46,8 +47,7 @@ The Pulsar broker will no longer manage schema data for topics using external sc
 Pulsar will introduce a new schema type: **SchemaType.EXTERNAL**.
 
 - All schemas that integrate with external schema registries must declare `SchemaType.EXTERNAL`.
-- When using `EXTERNAL` schema type, the Pulsar client will provide an empty schema definition to the broker.
-- If the `SchemaInfoProvider` is external but the schema type is not `EXTERNAL`, throw an ExternalSchemaException.
+- When using `EXTERNAL` schema type, the Pulsar client will provide empty schema data to the broker.
 - The broker will only record the schema type for topics.
 - Compatibility restrictions:
   - Introduce a new compatibility check in broker side.
@@ -61,7 +61,6 @@ This design isolates external schema management and protects existing topics usi
 
 To integrate with external schema registries, users can:
 - Implement the `Schema` interface to define custom schema encoding and decoding logic.
-- Implement the `SchemaInfoProvider` interface to initialize "external" schemas.
 
 #### Key `Schema` Interface Methods:
 - byte[] encode(T message)
@@ -72,16 +71,18 @@ To integrate with external schema registries, users can:
   - Deserialize the message using the external schema.
   - Users should handle exceptions when get value by themselves.
 
-- close()
+- void setSchemaInfoProvider(SchemaInfoProvider schemaInfoProvider)
+  - Call this method when creating schema
+  - External schema can be initialized when calling this method
 
-  **(New addition)**
+- close() **(New addition)**
   - Called when the producer or consumer is closed.
   - Allows external schema implementations to release resources, such as schema registry connections or caches.
 
 #### Example Workflow:
 
 - During producer or consumer initialization:
-  The custom schema can register schemas or fetch schema metadata from the external registry.
+  The external schema info will be registered to Pulsar schema storage.
 
 - During message send or receive:
   The `encode` and `decode` methods handle the schema-aware serialization and deserialization using the external schema registry.
@@ -107,34 +108,37 @@ public void workWithExternalSchemaRegistry() throws Exception {
     Map<String, String> srConfig = new HashedMap<>();
     srConfig.put("schema.registry.url", "http://localhost:8001");
 
-    PulsarClient client = PulsarClient.builder()
-            .serviceUrl("pulsar://localhost:6650")
-            .schemaInfoProviderFactory(new KafkaSchemaInfoProviderFactory(srConfig))
-            .build();
-
-    String topic = "t1";
+    String topic = "testExternalJsonSchema";
+  
     Schema<User> schema = KafkaSchemas.JSON(User.class);
-
-    Producer<User> producer = client.newProducer(schema)
+  
+    @Cleanup
+    PulsarClient pulsarClient = PulsarClient.builder()
+            .serviceUrl("pulsar://localhost:" + getBrokerServicePort())
+            .schemaProperties(srConfig)
+            .build();
+  
+    @Cleanup
+    Producer<User> producer = pulsarClient.newProducer(schema)
             .topic(topic)
             .create();
-
-    Consumer<User> consumer = client.newConsumer(schema)
+  
+    @Cleanup
+    Consumer<User> consumer = pulsarClient.newConsumer(schema)
             .topic(topic)
             .subscriptionName("sub")
             .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
             .subscribe();
-
+  
     for (int i = 0; i < 10; i++) {
-        producer.send(new User("name-" + i, 10 + i));
+      producer.send(new User("name-" + i, 10 + i));
     }
-
+  
     for (int i = 0; i < 10; i++) {
-        Message<User> message = consumer.receive();
-        consumer.acknowledge(message);
+      Message<User> message = consumer.receive();
+      consumer.acknowledge(message);
+      assertEquals(message.getValue().getName(), "name-" + i);
     }
-
-    client.close();
 }
 ```
 
@@ -151,14 +155,30 @@ public enum SchemaType {
      * External schema type is not compatible with any other schema type.
      * </p>
      */
-    EXTERNAL(-5)
+    EXTERNAL(21)
 
 }
 ```
 
-Add new methods for `SchemaInfoProvider` interface.
-The `SchemaInfoProvider` provides necessary params for connecting to the external schema registry service with the method `getConfigs`.
-If the schema info provider is external, the new producer command, consumer subscribe command will treat the schema as bytes schema, the broker will ignore schema validation.
+```protobuf
+// File `SchemaRegistryFormat.proto`
+message SchemaInfo {
+    enum SchemaType {
+        EXTERNAL = 22;
+    }
+}
+```
+
+```protobuf
+// File `PulsarApi.proto`
+message Schema {
+    enum Type {
+        External = 22;
+    }
+}
+```
+
+Add a new method `getConfigs` for `SchemaInfoProvider` interface to provide necessary params for external schemas connect to the schema registry service.
 ```java
 public interface SchemaInfoProvider {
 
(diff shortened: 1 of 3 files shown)


Denovo1998 (changes requested):
This PIP looks meaningful.

pip/pip-420.md
```diff
@@ -0,0 +1,253 @@
+# PIP-420: Provide ability for Pulsar clients to integrate with third-party schema registry service
+
+# Background knowledge
+
+Schema is an important feature for messaging systems. Pulsar integrates schema manager into the Pulsar broker.
+The current implementation in Pulsar clients couples schema management with some protocols (creating producer, adding consumer subscription).
+This increases the Pulsar protocol complexity and users can’t leverage third-party schema registry services in Pulsar client.
```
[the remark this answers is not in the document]
gaoran10: Thanks for reminding, the `SchemaStorage` is designed to manage schema on the server side. This PIP mainly provides the ability to access a third-party schema registry service on the Pulsar client side, so it's not an alternative implementation.

pip/pip-420.md
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+}
+```
+
+## Public-facing Changes
+
+Add new methods for `SchemaInfoProvider` interface.
+The `SchemaInfoProvider` provides necessary params for connecting to the external schema registry service with the method `getConfigs`.
+If the schema info provider is external, the new producer command, consumer subscribe command will treat the schema as bytes schema, the broker will ignore schema validation.
```
BewareMyPower: We should tell the broker that the producer or consumer is using an external schema registry to access this topic. Otherwise, if a Pulsar producer with built-in schema produces messages to this topic, it will succeed.

BewareMyPower:
BTW, you can leverage AI tool (e.g. Copilot) to improve the documents with more professional expression.

image

Denovo1998:
> BTW, you can leverage AI tool (e.g. Copilot) to improve the documents with more professional expression.
> 
> image

@BewareMyPower @gaoran10 

**Before this, a small trick by @lhotari inspired me.(https://github.com/apache/pulsar/pull/23794#issuecomment-2569093501)**

**Now, I wrote PIP documentation through LLM.**

---

CleanShot 2025-06-13 at 20 05 18@2x

---

- **First, we need a good system prompt that can directly generate content output that meets the format requirements based on the article title in the PIP document.**
- **We need to provide sufficient contextual information (e.g., code patches, the motivation for writing this PIP), so that the generated draft can reach a basic usable level. After multiple rounds of dialogue adjustments or manual modifications, the document will become very professional.** 

**This approach saves a lot of time, allowing us to focus on code development.**

**The system prompt is as follows:**

https://gist.github.com/Denovo1998/163e55b3a612873364a00cf0df5a1b95

pip/pip-420.md
gaoran10: Introduce a new schema type `EXTERNAL` and add a section for schema type change.

pip/pip-420.md
```diff
@@ -0,0 +1,253 @@
...
+      * @return a {@code SchemaInfoProvider} representing the schema info provider
+      */
+    default SchemaInfoProvider getSchemaInfoProvider() {
+        return null;
+    }
+
+}
+```
+
+# Pulsar Function
```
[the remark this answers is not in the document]
gaoran10: Introduce a new schema type `EXTERNAL`; schemas that use the external schema registry should use this schema type. Please check the schema type section.

pip/pip-420.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Design & Implementation Details
+
+This PIP's target is decoupling schema management from Pulsar messaging protocols,
+and making Pulsar client have the ability to leverage external schema registry service to manage schema.
+The external schema registry is responsible for managing the schema, the broker doesn't care about the messaging schema.
+The Pulsar client should ignore the schema information when creating producer and adding consumer subscription.
+
+Users can implement the `SchemaInfoProvider` interface and `Schema` interface to access external schema registry service.
+The `Schema` interface has mainly two methods `encode` and `decode`, the customized schemas can register schema or get schema with these methods.
```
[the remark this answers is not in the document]
gaoran10: Good suggestions.

When encoding, the external schema implementations should throw `SchemaSerializationException` if they encounter any exceptions, then the client can catch the exception and return it to the users. Users should handle exceptions when getting the value by themselves. I added to the "Key `Schema` Interface Methods" section.

pip/pip-420.md
```diff
@@ -0,0 +1,253 @@
...
+
+For supporting using third-party schema registry service in Pulsar Function,
+the function configurations should support setting the `SchemaInfoProviderFactory` and schema registry auth configurations while initializing the Pulsar client.
+
+# Security Considerations
+Users can provide security configuration while generating the `SchemaInfoProviderFactory` instance,
+the factory can transfer the security configuration to the `SchemaInfoProvider` instance.
+
+# Pulsar-GEO replication
+If users can use third-party schema registry service, it provides a new way to manage scheme for geo-replicated topics.
```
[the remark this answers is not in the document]
gaoran10: Add more details.

pip/pip-420.md
```diff
@@ -0,0 +1,253 @@
...
+
+Schema is an important feature for messaging systems. Pulsar integrates schema manager into the Pulsar broker.
+The current implementation in Pulsar clients couples schema management with some protocols (creating producer, adding consumer subscription).
+This increases the Pulsar protocol complexity and users can’t leverage third-party schema registry services in Pulsar client.
+
+# Motivation
+
+The Pulsar client is better has the ability to access third-party schema registry service to manage the schema (register schema,
+get schema, validate schema, etc.). The schema registry service can be an independent service, if using third-party schema registry service,
+the Pulsar broker doesn't need to care about the schema of the messages.
```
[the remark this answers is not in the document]
gaoran10: Update the motivation.

gaoran10:
> @BewareMyPower @gaoran10
> 
> Before this, a small trick by @lhotari inspired me.(https://github.com/apache/pulsar/pull/23794#issuecomment-2569093501)
> 
> Now, I wrote PIP documentation through LLM.

@Denovo1998 Nice work, I also improved the document with AI tools. /cc @BewareMyPower

pip/pip-420.md
```diff
@@ -0,0 +1,284 @@
...
+- Support setting the `SchemaInfoProviderFactory` while initializing the Pulsar client
+- Support the schema registry auth configurations while initializing the Pulsar client
+- Support the `SchemaType.EXTERNAL` schema type in Pulsar Function
+
+# Security Considerations
+
+Users can provide security configuration while generating the `SchemaInfoProviderFactory` instance,
+the factory can transfer the security configuration to the `SchemaInfoProvider` instance.
+
+# Pulsar-GEO replication impact
```
gaoran10, line 275: Hi, @poorbarcode , could you help review this section?

pip/pip-420.md
```diff
@@ -0,0 +1,284 @@
...
+
+- All schemas that integrate with external schema registries must declare `SchemaType.EXTERNAL`.
+- When using `EXTERNAL` schema type, the Pulsar client will provide an empty schema definition to the broker.
+- If the `SchemaInfoProvider` is external but the schema type is not `EXTERNAL`, throw an ExternalSchemaException.
+- The broker will only record the schema type for topics.
+- Compatibility restrictions:
+  - Introduce a new compatibility check in broker side.
+  - The schema type `SchemaType.EXTERNAL` can't be compatible with other Pulsar schemas
+  - This prevents accidental data corruption or schema conflicts between internal and external schema management systems.
+- Pulsar Geo replicator needs to transfer the schema type `SchemaType.EXTERNAL` to the remote cluster.
```
gaoran10: Hi, @poorbarcode , could you help review this PIP at your convenience?

codelipenghui (approved):

BewareMyPower (approved):

pip/pip-420.md
```diff
@@ -0,0 +1,277 @@
...
+
+- During message send or receive:
+  The `encode` and `decode` methods handle the schema-aware serialization and deserialization using the external schema registry.
+
+#### Schema ID & Schema Version
+
+Unlike Pulsar, which uses **schema version** to identify schemas, many external schema registry systems use **schema ID** as the primary schema identifier.
+
+When integrating with external schema registries:
+- The `schemaVersion` filed in Pulsar message metadata is used in some places, **set to `-1` to flag the message is using external schema systems**.
```
BewareMyPower:
```protobuf
    optional bytes schema_version = 16;
```

The type of schema version is `bytes`. It's better to clearly define the specific bytes that represent external schema registry.

In addition, could you explain more about when is the schema version used? I took a quick look and found the following API should be affected:

```java
public interface LookupService extends AutoCloseable {

    CompletableFuture<Optional<SchemaInfo>> getSchema(TopicName topicName, byte[] version);
```

It's used in `MultiVersionSchemaInfoProvider#loadSchema`.

It's better to add a **background** section to explain all related details, otherwise setting -1 as the schema version might be confusing. For example, why not use an empty byte array as the schema version?

gaoran10: Oh, this is for the previous design that used the bytes schema. The current design introduces a new schema type, EXTERNAL, and Pulsar will store the schema information. I updated the doc.

BewareMyPower (approved):

pip/pip-420.md
poorbarcode: The same schema crossing both clusters has a different version. How does the third party service confirm the compatibility?

pip/pip-420.md
poorbarcode: Answered [here](https://github.com/apache/pulsar/pull/24328/files#r2196597452)

pip/pip-420.md
gaoran10: If using the schema type `EXTERNAL`, the schema version will always be 1, as it's not compatible with other schema types. Therefore, when crossing both clusters, they should have the same schema version.

If using the schema type, the schema compatibility check is handled by the third-party schema registry service. They can retrieve the schema ID from the message payload.

poorbarcode: @gaoran10 

Thanks, I have reviewed all of the doc.

> During producer or consumer initialization: The external schema info will be registered to Pulsar schema storage.
> During message send or receive: The encode and decode methods handle the schema-aware serialization and deserialization using the external schema registry.

producer/consumer registries schemas to the external schema store directly, will not pass by the broker. Since the client knows how to encode and decode the message, how about using `BYTES_SCHEMA`, which avoids involving the broker?

poorbarcode: Setting schema compatibility to `always incompatible` can also make it possible, which does not involve the broker. I think adding an optional parameter that indicates auto-set schema compatibility to `always incompatible` into the creating a topic is a better solution.
- Or introduce a new compatibility strategy named `always incompatible because using an external store`

Adding a new schema type, you need to change the compatibility of the schema service of the broker, but the broker does not touch any schema that is stored externally, and the producer/consumer registers a dummy schema into the schema service, which is implemented by the broker, but it does nothing, and it occupies an empty ledger that relates to the dummy schema

pip/pip-420.md
```diff
@@ -0,0 +1,276 @@
...
+
+# Pulsar Function
+
+For support using third-party schema registry service in Pulsar Function,
+- Support setting the `schemaProperties` while initializing the Pulsar client
+- Support the `SchemaType.EXTERNAL` schema type in Pulsar Function
+
+# Security Considerations
+
+Users can provide schema registry security configuration in the `schemaProperties`.
```
poorbarcode: Since the client directly connects to the external schema store, does this proposal involve the definition of the authentication and authorization interface?

pip/pip-420.md
```diff
@@ -0,0 +1,276 @@
...
+  
+    for (int i = 0; i < 10; i++) {
+      Message<User> message = consumer.receive();
+      consumer.acknowledge(message);
+      assertEquals(message.getValue().getName(), "name-" + i);
+    }
+}
+```
+
+## Public-facing Changes
```
poorbarcode, line 152: How could users query an existing schema by a `schema id`? Is there a rule used to define how to generate a schema ID? How could a user get related schema IDs to a topic?

pip/pip-420.md
```diff
@@ -0,0 +1,276 @@
...
+
+Introduce a new SchemaType `EXTERNAL` to represent the schema types that work with external schema registry.
```
poorbarcode: It seems that we need a definition of the interface `external schema store client`, because we need to call it in the Pulsar client.

pip/pip-420.md
gaoran10: Initially, I tried using the BYTES schema, but it was difficult to perform compatibility checks against other schema types.

We could define the EXTERNAL schema type as universally incompatible with all other schema types, independent of the selected compatibility strategy. WDYT?

pip/pip-420.md
gaoran10: The schema ID will be generated by the third-party schema registry, and users can retrieve schema information via the third-party registry's service.

pip/pip-420.md
gaoran10: Because the other schema system may integrate register and retrieve schema operations into the serialization and deserialization process, such as Kafka JSON, Avro serializer and deserializer, Pulsar doesn't need to touch that underlying operations; otherwise, we need to rewrite a lot of logic in custom external schemas.

pip/pip-420.md
gaoran10: Users only need to provide security configurations in schemaProperties; these are used to initialize the external schema registry client.

poorbarcode (approved):

pip/pip-420.md
```diff
@@ -0,0 +1,276 @@
...
+
+Pulsar will introduce a new schema type: **SchemaType.EXTERNAL**.
+
+- All schemas that integrate with external schema registries must declare `SchemaType.EXTERNAL`.
+- When using `EXTERNAL` schema type, the Pulsar client will provide empty schema data to the broker.
+- The broker will only record the schema type for topics.
+- Compatibility restrictions:
+  - Introduce a new compatibility check in broker side.
+  - The schema type `SchemaType.EXTERNAL` can't be compatible with other Pulsar schemas
+  - This prevents accidental data corruption or schema conflicts between internal and external schema management systems.
```
RobertIndie, line 55: Could we talk about how the external compatibility checker work with the compatibility check strategy in this PIP? https://pulsar.apache.org/docs/next/schema-understand/#schema-compatibility-check-strategy

Like what the behavior iof the external checker is when users set it to each different strategy.

BewareMyPower (approved):

pip/pip-420.md
```diff
@@ -0,0 +1,304 @@
...
+
+# General Notes
+
+# Links
+
+<!--
+Updated afterwards
+-->
+* Mailing List discussion thread: https://lists.apache.org/thread/olx4xm8cdy43omp5c0jm44sj1gp0grcr
+* Mailing List voting thread: https://lists.apache.org/thread/vhq6ox4nh2rx59yoxowftqzv8f9lnm4q
```
BewareMyPower: This link is wrong, please update the correct link

BewareMyPower (changes requested):

pip/pip-420.md
BewareMyPower: The discussion thread is also wrong. They are all related to PIP-401

RobertIndie (approved):

BewareMyPower (approved):

pip/pip-420.md
gaoran10: update

PR #24331, opened 2025-05-21
[fix][offload] Complete the future outside of the reading loop in BlobStoreBackedReadHandleImplV2.readAsync

lhotari:
### Motivation

There's a bug in `BlobStoreBackedReadHandleImplV2.readAsync` which can be reproduced with 
the existing test `BlobStoreManagedLedgerOffloaderStreamingTest#testReadAndWriteAcrossSegment`. 
The test is flaky and could result in a `ConcurrentModificationException`.

```
java.util.ConcurrentModificationException
        at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1095)
        at java.base/java.util.ArrayList$Itr.next(ArrayList.java:1049)
        at org.apache.bookkeeper.mledger.offload.jcloud.impl.BlobStoreManagedLedgerOffloaderStreamingTest.testReadAndWriteAcrossSegment(BlobStoreManagedLedgerOffloaderStreamingTest.java:326)
```

While investigating this, I noticed that this is a real bug in production code.

It looks like this only impacts "streaming offload" implementation, ["PIP 76: Streaming Offload"](https://github.com/apache/pulsar/wiki/PIP-76:-Streaming-Offload).
It's a bit confusing since it looks like PIP-76 was never fully completed. I don't see any calls to `org.apache.bookkeeper.mledger.LedgerOffloader#streamingOffload` method in production code. Similarly, there's no `org.apache.bookkeeper.mledger.LedgerOffloader#readOffloaded` method calls in production code.

### Modifications

- complete the future outside of the reading loop in BlobStoreBackedReadHandleImplV2.readAsync

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
+++ b/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
@@ -233,8 +233,8 @@ public class BlobStoreBackedReadHandleImplV2 implements ReadHandle {
                     entries.forEach(LedgerEntry::close);
                 }
 
-                promise.complete(LedgerEntriesImpl.create(entries));
             }
+            promise.complete(LedgerEntriesImpl.create(entries));
         });
         return promise;
     }


tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
```diff
@@ -233,8 +233,8 @@ public CompletableFuture<LedgerEntries> readAsync(long firstEntry, long lastEntr
                     entries.forEach(LedgerEntry::close);
                 }
 
-                promise.complete(LedgerEntriesImpl.create(entries));
```
heesung-sn, line 236: This seems to be a critical bug. Do we have any test to cover this logic? (Seems not). maybe we should ask the code author to add a test for this logic. (as a separate PR).

heesung-sn (approved):

nodece (approved):

dao-jun (approved):

tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
lhotari: There's an existing test BlobStoreManagedLedgerOffloaderStreamingTest.testReadAndWriteAcrossSegment to cover this. It fails for me consistently when running on command line locally. It seems that it requires a fast CPU to reproduce and this test has rarely failed in CI.

lhotari: @heesung-sn It looks like this code never went to production, PIP-76 hasn't been completed. I added details to the description.

PR #24341, opened 2025-05-22
[fix][test] Fix flaky PulsarBrokerStatsClientTest.testTopicInternalStats

hanmz:
### Motivation
Clipboard_Screenshot_1747912362

This unit test `assertTrue(cursor.totalNonContiguousDeletedMessagesRange > 0 && (cursor.totalNonContiguousDeletedMessagesRange) < numberOfMsgs / 2)` reports an error because the consumer's acknowledge method is implemented one-way and the server may not have processed the ack request yet before `topic.getInternalStats` , so `totalNonContiguousDeletedMessagesRange` is sometimes equal to 0.
https://github.com/apache/pulsar/blob/bdf6277456db9df1618adcd07b588f3a94714796/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java#L616

At the same time, `assertFalse(cursor.subscriptionHavePendingRead)` sometimes goes wrong unless we close the consumer first.

### Modifications 
Optimize assert logic to ensure the robustness of unit tests.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/stats/client/PulsarBrokerStatsClientTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/stats/client/PulsarBrokerStatsClientTest.java
@@ -133,6 +133,7 @@ public class PulsarBrokerStatsClientTest extends ProducerConsumerBase {
         assertTrue(cursor.active);
 
         producer.close();
+        // Call close to ensure that all acknowledge requests are executed.
         consumer.close();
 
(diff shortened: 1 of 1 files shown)


dao-jun:
Good job, I tested locally 4 times and it all passed

dao-jun (approved):

PR #24342, opened 2025-05-23
[fix][test] Fix flaky NegativeAcksTest.testNegativeAcksWithBackoff

hanmz:
### Motivation
Clipboard_Screenshot_1747985208

This test sometimes blocks at `consumer.receive()` at line 261. The problem is introduced in this PR https://github.com/apache/pulsar/pull/23986. For batch sending scenarios, the current way of creating producers cannot guarantee that all messages are sent out in one batch. Therefore, for batch messages, executing `consumer。negativeAcknowledge` only once is not enough.

### Modifications 
By adding `batchingMaxPublishDelay` when creating the producer, all messages are guaranteed to be sent together.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
@@ -240,6 +240,7 @@ public class NegativeAcksTest extends ProducerConsumerBase {
         Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
                 .topic(topic)
                 .enableBatching(batching)
+                .batchingMaxPublishDelay(30, TimeUnit.SECONDS)
                 .create();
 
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java
```diff
@@ -259,7 +260,7 @@ public void testNegativeAcksWithBackoff(boolean batching, boolean usePartitions,
             Message<String> msg = null;
             for (int j = 0; j < N; j++) {
                 msg = consumer.receive();
```
liangyepianzhou, line 262: It's better to add receive timeout here.

lhotari (approved):

PR #22560, opened 2024-04-23
[improve][ml] Optimize read entries with skipCondition

dao-jun:
### Motivation

In https://github.com/apache/pulsar/pull/19035 we introduced `skipCondition` to filter-out delay delivery messages before read entries from Bookkeeper, and in https://github.com/apache/pulsar/pull/21739, we also filter-out deleted(individual acked messages) before read entries.

However, it will lead to one situation:  one single segment can be spit into segments. For example:
entries to be filter-out: [3, 5, 7]
entries to read: [1, 10]
then, it will be split into: [[1,2],[4], [6], [8,10]].

In the current implementation, after read [1,2] finished, then begin to read [4], after read [4] finished, then start to read [6]...
It will lead to latency increasing, memory(allocated for entries) will also be retained for a longer period of time, and affect the throughput of the system. 

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24346, opened 2025-05-24
[improve][broker] Enable concurrent processing of pending read Entries to avoid duplicate Reads

berg223:
### Motivation

When I read source code about "skip send duplicate read to BK", I found it's not concurrent. 
Firstly, it would be faster if it's concurrrent. Secondly, it may still send duplicate read to BK.
For example:

1. read  [10, 70]
2. read [80, 100]
3. read [10, 100]

The third read will be splitted into [10,70], [71, 79], [80, 100]. The three reads will be performed in sequence. If [80, 100] pending reads is done before [10, 70] , there will be another [80, 100] read send to BookKeeper later.  At the same time, it will be faster if we send [71, 79] read to BK earlier.

### Modifications

I have add three futures in `org.apache.bookkeeper.mledger.impl.cache.PendingReadsManager.PendingRead#readEntriesComplete`. Each future
will read a  range of entry concurrently. At last aggregate all the results of future.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/PendingReadsManager.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/PendingReadsManager.java
@@ -21,6 +21,7 @@ package org.apache.bookkeeper.mledger.impl.cache;
 import static org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.createManagedLedgerException;
 import io.prometheus.client.Counter;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
@@ -268,7 +269,7 @@ public class PendingReadsManager {
 
         // this method isn't synchronized since that could lead to deadlocks
         private void readEntriesComplete(List<ReadEntriesCallbackWithContext> callbacks,
-                                                List<EntryImpl> entriesToReturn) {
+                                         List<EntryImpl> entriesToReturn) {
             if (callbacks.size() == 1) {
                 ReadEntriesCallbackWithContext first = callbacks.get(0);
                 if (first.startEntry == key.startEntry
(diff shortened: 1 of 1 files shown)


berg223:
@lhotari Thanks for expanding my horizons. I would be very happy if there is anything I can do.

> @berg223 There's also #22560 in this area by @dao-jun .

lhotari:
> @lhotari Thanks for expanding my horizons. I would be very happy if there is anything I can do.
> 
> > @berg223 There's also #22560 in this area by @dao-jun .

Any feedback on #22560 or https://github.com/lhotari/pulsar/pull/207 is appreciated. I just finished rebasing https://github.com/lhotari/pulsar/pull/207 which is something that I have been working on since October 2024. I'd like to get those changes into a better state.

It would be useful to have a benchmark for Pulsar where the main goal would be to minimize the number of reads to bookkeeper so that it would be possible to show how broker cache improvements or other improvements on Pulsar side such as this change or #22560 would be beneficial. There aren't any workload simulation tests in Pulsar currently. They wouldn't necessarily need to be something that does a massive load test since the efficiency could be shown also in relatively small scale.

lhotari:
@BewareMyPower 
> Performing read operations concurrently could increase the CPU pressure from Pulsar side, because there will be more inflight requests.

Pulsar has a setting managedLedgerMaxReadsInFlightSizeInMB which serves as back pressure for inflight requests.

> In addition, this PR is merged with no tests added. It's very risky from my perspective. I didn't look into the details at the moment, but not sure if it could affect the calling order of original `readEntriesComplete` callbacks. I will take a look soon.

The existing test coverage covers the functional testing of these changes.

Reviewing PRs in a timely manner is important for keeping our contributors engaged. Hopefully more Pulsar committers will pay attention to PRs going forward. This PR has a clear argumentation and is good quality. I hope we receive more of these going forward.

BewareMyPower:
I tried to understand the motivation by adding a test to `PendingReadsManagerTest`:

```java
        final var pendingReadsManager = new PendingReadsManager(rangeEntryCache);
        final var readFutures = new ArrayList<CapturingReadEntriesCallback>();
        final BiConsumer<Long, Long> readEntries = (firstEntry, lastEntry) -> {
            final var callback = new CapturingReadEntriesCallback();
            pendingReadsManager.readEntries(lh, firstEntry, lastEntry, false, callback, CTX);
            readFutures.add(callback);
        };
        final BiFunction<Long, Long, PreparedReadFromStorage> mockReadFromStorage = (firstEntry, lastEntry) ->
                prepareReadFromStorage(lh, rangeEntryCache, firstEntry, lastEntry, false);

        final var read0 = mockReadFromStorage.apply(10L, 70L);
        readEntries.accept(10L, 70L);
        final var read1 = mockReadFromStorage.apply(80L, 100L);
        readEntries.accept(80L, 100L);
        final var read2 = mockReadFromStorage.apply(71L, 100L); // NOTE: it's not [71, 79]
        readEntries.accept(10L, 100L);

        read1.storageReadCompleted();
        read0.storageReadCompleted();
        read2.storageReadCompleted();
        FutureUtil.waitForAll(readFutures).get();
```

See the `NOTE` comment, when I debugged the test, I found this description is wrong:

> The third read will be splitted into [10,70], [71, 79], [80, 100].

Instead, the 3rd read will just be split into:
- A pending read on `[10, 70]`
- A new read on `[71, 100]`

The reason is for simplicity, only 1 `PendingRead` can be reused for a new `PendingReadsManager#readEntries` call.

```java
    private record FindPendingReadOutcome(PendingRead pendingRead, // NOTE: it's a single PendingRead
                                          PendingReadKey missingOnLeft, PendingReadKey missingOnRight) {
```

Basically, the motivation is wrong, how can you say it's a clear argumentation? The root cause is that **there is no test or code example**, so the explanation might not match the actual behavior.

BewareMyPower:
Oh I realized my test is wrong.

I will open a new PR to add the missed test for this PR.

BewareMyPower:
Here it is: https://github.com/apache/pulsar/pull/24519

lhotari:
> Basically, the motivation is wrong, how can you say it's a clear argumentation? The root cause is that **there is no test or code example**, so the explanation might not match the actual behavior.

@BewareMyPower Sure, that's true. There's room for improvement. Thanks for looking into this.

Another way to describe the motivation is to avoid extra latency due to sequential reads.
One different way to look at this PR is that it will fix a performance regression that PendingReadsManager could introduce. due to the sequential reads when concurrent reads overlap. Before PendingReadsManager and InflightReadsLimiter were introduced, the reads would go directly to BookKeeper without additional backpressure from what dispatcher does. That's why I'd consider the risk of this change causing anything but positive impacts by reducing latency.

Issue #24312, opened 2025-05-16
[Bug] pulsar 4.0.4 Unable to start broker


HGHNice:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

pulsar 4.0.4

### Issue Description

![Image](https://github.com/user-attachments/assets/da202171-5052-43d0-bae2-6f397b2fc5cf)

### Error messages

```text
unable get broker  id
```

### Reproducing the issue

The apache-pulsar-4.0.4-bin.tar.gz used, and it was found that the deployment was completed. The last step was to start the broker and it could not start normally.

The reproduction step is: execute it once according to the document. Start broker and report an error in the last step.

### Additional information

null

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

HGHNice:
I've used bin-pulsar initialize-cluster-metadata normally for initialization

Shawyeok:
The `brokerId` is initialized after the BrokerService has started.

https://github.com/apache/pulsar/blob/eccc6b647e9cb07a18901471de1b2f8fafa88417/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java#L940-L942

So there's a chance the broker could receive a metadata change event before `brokerId` is initialized.

https://github.com/apache/pulsar/blob/eccc6b647e9cb07a18901471de1b2f8fafa88417/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L2535-L2539

To avoid this race condition, we could initialize the `brokerId` earlier in the startup process, Would you be interested in opening a PR to address this?

@HGHNice

HGHNice:
> The `brokerId` is initialized after the BrokerService has started.
> 
> [pulsar/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java](https://github.com/apache/pulsar/blob/eccc6b647e9cb07a18901471de1b2f8fafa88417/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java#L940-L942)
> 
[...]

PR #24352, opened 2025-05-27
[fix][broker] Ignore metadata changes when broker is not in the Started state

nodece:
Fixes https://github.com/apache/pulsar/issues/24312

### Motivation

The `brokerId` was previously initialized in the `start` method, after the `BrokerService` had already started. This could lead to a `NullPointerException` if the `brokerId` is accessed during the early stages of startup. 

### Modifications

- Only handle the metadata changes when broker is fully started. 
- Failed to retrieve `brokerId`, please fall back to `unknown`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
Commented about another way to address the issue: https://github.com/apache/pulsar/issues/24312#issuecomment-2913163315

lhotari (approved):
LGTM

PR #24354, opened 2025-05-27
[fix][io] Acknowledge RabbitMQ message after processing the message successfully

### LINKED ISSUE #24353: [Bug] Pulsar IO RabbitMQ source connector might loose messages since it acks to RabbitMQ before message has been handled
### Discussed in https://github.com/apache/pulsar/discussions/23677

<div type='discussions-op-text'>

<sup>Originally posted by **fabrizziocht** December  4, 2024</sup>
I was exploring the rabbitMQ source connector (https://github.com/apache/pulsar/blob/v4.0.1/pulsar-io/rabbitmq/src/main/java/org/apache/pulsar/io/rabbitmq/RabbitMQSource.java) that implements the class PushSource (it extends AbstractPushSource class), 

I can see that each time a message is collected by the RabbitMQConsumer class using the handleDelivery method it invokes the RabbitMQSource consume:
https://github.com/apache/pulsar/blob/v4.0.1/pulsar-io/rabbitmq/src/main/java/org/apache/pulsar/io/rabbitmq/RabbitMQSource.java#L99

Then it  stores only in memory in a LinkedBlockingQueue and the ack is sent to RabbitMQ:
https://github.com/apache/pulsar/blob/master/pulsar-io/core/src/main/java/org/apache/pulsar/io/core/AbstractPushSource.java#L76

What happens if the pulsar IO service is restarted? I can see that the data stored in the LinkedBlockingQueue will be lost, because the message was already acked and removed in rabbitMQ and also was not still written in the pulsar topic. Is it a limitation of the rabbitmq source connector? or all connectors that implements the PushSource has the problem? Exists another interface/class that can guarantee that the message was written first in the pulsar topic before send the ack to rabbitMQ?</div>

lhotari:
Fixes #24353

### Motivation

See #24346 

### Modifications

- add `ackFunction` and `failFunction` to the internal `RabbitMQRecord` class and use these functions in the `ack` and `fail` implementations of the `Record` interface.

- switch to use individual acks towards RabbitMQ since there's no benefit of using "multi" acks when every message will get acked in any case

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-io/rabbitmq/src/main/java/org/apache/pulsar/io/rabbitmq/RabbitMQSource.java
+++ b/pulsar-io/rabbitmq/src/main/java/org/apache/pulsar/io/rabbitmq/RabbitMQSource.java
@@ -96,11 +96,23 @@ public class RabbitMQSource extends PushSource<byte[]> {
         @Override
         public void handleDelivery(String consumerTag, Envelope envelope, AMQP.BasicProperties properties, byte[] body)
                 throws IOException {
-            source.consume(new RabbitMQRecord(Optional.ofNullable(envelope.getRoutingKey()), body));
             long deliveryTag = envelope.getDeliveryTag();
-            // positively acknowledge all deliveries up to this delivery tag to reduce network traffic
-            // since manual message acknowledgments are turned on by default
-            this.getChannel().basicAck(deliveryTag, true);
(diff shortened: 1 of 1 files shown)


david-streamlio (approved):
LGTM

PR #24356, opened 2025-05-28
[fix][client] Fix consumer not returning encrypted messages on decryption failure with compression enabled

RobertIndie:
## Motivation

There is a problem when a consumer tries to decrypt a compressed message. If a consumer fails to decrypt a compressed message, it also fails to decompress it, leading to the message being discarded. The user won't receive the message, even if using `CryptoFailureAction.CONSUME`, because the failure occurs during decompression, not decryption. This issue could be easily reproduced by the test added in this PR: `testE2EEncryptionWithCompression `.

We already have logic to skip decompression for messages that can't be decrypted: https://github.com/apache/pulsar/blob/397b0211cf1c9069d80daa691ca530e09bc37999/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1439-L1441

The main issue is if the consumer fails to decrypt the message here: https://github.com/apache/pulsar/blob/397b0211cf1c9069d80daa691ca530e09bc37999/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1947, it isn't treated as undecryptable and still tries to decompress.

## Modification

- Refactor `decryptPayloadIfNeeded` to return a clearer result, `DecryptResult`, to guide the logic on whether the message was decrypted successfully or should be discarded.
- If decryption fails, treat the message as undecryptable and skip decompression.

### Verifying this change

This change added tests

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1923,6 +1923,9 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         return connectionHandler.lastConnectionClosedTimestamp;
     }
 
+    /**
+     * Represents the outcome of a message decryption attempt for the consumer.
+     */
     private static class DecryptResult {
         private final boolean success;
         private final ByteBuf decryptedPayload;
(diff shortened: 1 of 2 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1921,11 +1923,53 @@ public long getLastDisconnectedTimestamp() {
...
     }
 
-    private ByteBuf decryptPayloadIfNeeded(MessageIdData messageId, int redeliveryCount, MessageMetadata msgMetadata,
-                                           ByteBuf payload, ClientCnx currentCnx) {
+    /**
+     * Represents the outcome of a message decryption attempt for the consumer.
+     */
+    private static class DecryptResult {
+        private final boolean success;
+        private final ByteBuf decryptedPayload;
```
codelipenghui: Is it better to use payload instead of decryptedPayload? Since it could be raw data when `success == true`

RobertIndie: +1

dao-jun (approved):

lhotari:
@RobertIndie Would there be a need to cherry-pick to branch-3.0 ?

RobertIndie:
> @RobertIndie Would there be a need to cherry-pick to branch-3.0 ?

I think we don't need to. It's not a critical bug. We later cherry-pick if someone needs it.

PR #24359, opened 2025-05-28
[improve][broker]Improve the log when encountered in-flight read limitation

poorbarcode:

- When reaching the pending read queue size limitation
  - Highlight the config `managedLedgerMaxReadsInFlightPermitsAcquireQueueSize` when the broker reaches the limitation that the pending read queue is full.
  - Instead of printing an error-level log,  print a warning-level log.
- Additional print queue size and the limit threshold when the pending read queue is full
- Print warning-level log if reached in-flight messages size is reached, instead of debug-level.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
@@ -179,7 +179,9 @@ public class InflightReadsLimiter implements AutoCloseable {
             if (queuedHandles.size() >= maxReadsInFlightAcquireQueueSize) {
                 log.warn("Failed to queue handle for acquiring permits: {}, creationTime: {}, remainingBytes:{},"
                     + " maxReadsInFlightAcquireQueueSize:{}, pending-queue-size: {}, please increase broker"
-                    + " config managedLedgerMaxReadsInFlightPermitsAcquireQueueSize",
+                    + " config managedLedgerMaxReadsInFlightPermitsAcquireQueueSize and confirm the configuration of"
+                    + " managedLedgerMaxReadsInFlightSizeInMB and"
+                    + " managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis are suitable.",
                     permits, handle.creationTime, remainingBytes, maxReadsInFlightAcquireQueueSize, queuedHandles.size());
                 return Optional.of(new Handle(0, handle.creationTime, false));
             } else {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
@@ -182,7 +182,8 @@ public class InflightReadsLimiter implements AutoCloseable {
                     + " config managedLedgerMaxReadsInFlightPermitsAcquireQueueSize and confirm the configuration of"
                     + " managedLedgerMaxReadsInFlightSizeInMB and"
                     + " managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis are suitable.",
-                    permits, handle.creationTime, remainingBytes, maxReadsInFlightAcquireQueueSize, queuedHandles.size());
+                    permits, handle.creationTime, remainingBytes, maxReadsInFlightAcquireQueueSize,
+                    queuedHandles.size());
                 return Optional.of(new Handle(0, handle.creationTime, false));
             } else {
                 queuedHandles.offer(new QueuedHandle(handle, callback));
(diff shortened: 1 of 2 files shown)


shibd (approved):

dao-jun (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
```diff
@@ -332,14 +332,12 @@ void doAsyncReadEntriesWithAcquiredPermits(ReadHandle lh, Position firstPosition
...
-                            + "%d, %s, estimated read size %d bytes for %d entries (check "
-                            + "managedLedgerMaxReadsInFlightSizeInMB, "
-                            + "managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis and "
-                            + "managedLedgerMaxReadsInFlightPermitsAcquireQueueSize)", lh.getId(), getName(),
-                    estimatedReadSize, numberOfEntries);
-            log.error(message);
+            String message = String.format("Couldn't acquire enough permits on the max reads in flight limiter to read"
+                + " from ledger %d, %s, estimated read size %d bytes for %d entries. because the pending read requests"
+                + " queue is full, please increase the broker config"
+                + " managedLedgerMaxReadsInFlightPermitsAcquireQueueSize.",
```
lhotari, line 338: This tuning recommendation isn't so straight forward since `managedLedgerMaxReadsInFlightSizeInMB` and `managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis` settings also impact the situation.

If the `managedLedgerMaxReadsInFlightSizeInMB` value is higher, the queuing wouldn't happen in the first place. The general tuning recommendation should be to primarily  increase `managedLedgerMaxReadsInFlightSizeInMB`, make more direct memory for the broker  and possibly also increase the queue size. 

When the timeout `managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis` is shorter, the queue length would eventually be shorter when reads are rejected due to overload. However, reducing the timeout isn't very helpful since retries would happen eventually and that would just add unnecessary load to the system.

poorbarcode: Reverted and added a highlight to the config queue size

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
```diff
@@ -223,15 +225,17 @@ private synchronized void timeoutCheck() {
...
     private void handleTimeout(QueuedHandle queuedHandle) {
-        if (log.isDebugEnabled()) {
-            log.debug("timed out queued permits: {}, creationTime: {}, remainingBytes:{}",
-                    queuedHandle.handle.permits, queuedHandle.handle.creationTime, remainingBytes);
-        }
+        log.warn("timed out queued permits: {}, creationTime: {}, remainingBytes:{}, acquireTimeoutMillis: {}. Please"
+                + " review whether the BK read requests is fast enough or broker config"
+                + " managedLedgerMaxReadsInFlightSizeInMB and managedLedgerMaxReadsInFlightPermitsAcquireTimeoutMillis"
+                + " are suitable",
+                queuedHandle.handle.permits, queuedHandle.handle.creationTime, remainingBytes, acquireTimeoutMillis);
```
lhotari, line 232: Does this just cause duplicate logging when it's at warn level since RangeEntryCacheImpl will log the failure too?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/InflightReadsLimiter.java
```diff
@@ -177,8 +177,10 @@ private synchronized Optional<Handle> internalAcquire(long permits, Consumer<Han
             return Optional.of(new Handle(maxReadsInFlightSize, handle.creationTime, true));
         } else {
             if (queuedHandles.size() >= maxReadsInFlightAcquireQueueSize) {
-                log.warn("Failed to queue handle for acquiring permits: {}, creationTime: {}, remainingBytes:{}",
-                        permits, handle.creationTime, remainingBytes);
+                log.warn("Failed to queue handle for acquiring permits: {}, creationTime: {}, remainingBytes:{},"
+                    + " maxReadsInFlightAcquireQueueSize:{}, pending-queue-size: {}, please increase broker"
+                    + " config managedLedgerMaxReadsInFlightPermitsAcquireQueueSize",
```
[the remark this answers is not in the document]
poorbarcode: Removed the `RangeEntryCacheImpl`'s one, because the current log contains `queue size`, but `RangeEntryCacheImpl`'s does not contain

lhotari (approved):
LGTM

PR #24363, opened 2025-05-29
[refactor][ml] Replace cache eviction algorithm with centralized removal queue and job

lhotari:
## Motivation

This PR fixes fundamental inefficiencies and correctness issues in the current Pulsar **broker entry cache** eviction algorithm. The current implementation has flawed size-based eviction that doesn't remove the oldest entries and incorrect timestamp-based eviction with high CPU overhead. These fixes ensure that size-based eviction properly removes the oldest entries and timestamp-based eviction works correctly. Additionally, this PR serves as a foundation for future improvements to efficiently handle catch-up reads and Key_Shared subscription scenarios.

Mailing list discussion about this PR: https://lists.apache.org/thread/ddzzc17b0c218ozq9tx0r3rx5sgljfb0

UPDATE: This change is covered in "PIP-430: Pulsar Broker cache improvements: refactoring eviction and adding a new cache strategy based on expected read count", #24444

### Problems with the Current Broker Entry Cache Implementation

1. **Size-Based Eviction doesn't remove oldest entries**: The existing `EntryCacheDefaultEvictionPolicy` uses an algorithm for keeping the cache size under the limit but cannot guarantee removal of the oldest entries from the cache. The algorithm:
   - Sorts caches by size in descending order
   - Selects caches representing a percentage of total size (`PercentOfSizeToConsiderForEviction`, default `0.5`)
   - Attempts to evict proportional amounts from each selected cache
   - This approach doesn't ensure that the oldest entries are removed, leading to inefficient cache utilization

2. **Inefficient and Incorrect Timestamp-Based Eviction**: The current timestamp eviction has both performance and correctness issues:
   - **Performance problems**:
     - Iterates through ALL cache instances (there's an instance for each persistent topic / ManagedLedgerImpl)
     - Causes remarkable CPU and memory pressure due to constant frequent iterations when there's a high number of topics running with high throughput.
     - Runs every 10 milliseconds by default (`managedLedgerCacheEvictionIntervalMs=10`) - 100 times per second!
   - **Correctness problems**:
     - Assumes entries are ordered by both position and timestamp, which breaks when:
       - Cache is used for catch-up reads (backlogged cursors)
       - Individual entries are cached out of order
       - Multiple read patterns (tailing + catch-up) access the same cache simultaneously

3. **Limited Cache Scope**: The original `RangeCache` was designed for tailing reads. Later changes added support for backlogged cursors, but the eviction algorithms weren't updated to handle mixed read patterns effectively.

4. **Unnecessary Complexity**: Generic type parameters in `RangeCache` add complexity without providing value, as the cache is only used for entry storage.

### 1. Centralized Removal Queue (`RangeCacheRemovalQueue`)

- **Single-threaded eviction**: All cache evictions are handled by one thread to avoid contention
- **Insertion-order tracking**: Uses `MpscUnboundedArrayQueue` to maintain entry insertion order
- **Accurate timestamp eviction**: Entries are processed in insertion order, making timestamp-based eviction reliable
- **Efficient size-based eviction**: Oldest entries are removed first when freeing cache space

### 2. Simplified Cache Implementation

- **Removed generics**: Dropped unnecessary type parameters from `RangeCache` to reduce complexity
- **Unified eviction handling**: All cache instances use the same central removal queue
- **Improved consistency**: Race conditions in eviction are minimized through centralized processing

### 3. Foundation for Future Improvements

The existing broker cache has limitations:
- Unnecessary BookKeeper reads during catch-up read scenarios
  - Causes increased network costs and resource usage on BookKeeper nodes
  - Cascading performance issues under high fan-out catch-up reads
  - Current backlogged cursors caching solution has multiple gaps
- Poor cache hit rates for Key_Shared subscriptions with slow consumers since entries get put into the replay queue and once the consumer has sent permits, these entries are read from BookKeeper (unless `cacheEvictionByMarkDeletedPosition=true`)

This refactoring prepares the cache system for:
- Enhanced catch-up read optimization
- Efficient replay queue caching for Key_Shared subscriptions

### Before (EntryCacheDefaultEvictionPolicy)

Size Based Eviction

1. Sort all caches by size (largest first)
2. Select caches until reaching PercentOfSizeToConsiderForEviction (0.5)
3. For each selected cache:
   - Calculate proportional eviction amount
   - Remove entries (no guarantee of age-based removal)
4. Problem: May remove newer entries while keeping older ones

https://github.com/apache/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/EntryCacheDefaultEvictionPolicy.java#L41-L93

Timestamp eviction

1. Iterate all caches
3. For each cache:
   - Start from the first position in the cache
   - Remove entries from the cache until the cache is empty or there's a valid entry that hasn't yet been expired
4. Problem: Iterating all caches and entries cause a lot of unnecessary CPU and memory pressure due to iterations. By default, this is performed every 10 milliseconds, 100 times per second. (`managedLedgerCacheEvictionIntervalMs=10`)

https://github.com/apache/pulsar/blob/eccc6b647e9cb07a18901471de1b2f8fafa88417/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java#L306-L317

### After (RangeCacheRemovalQueue)

1. All entries added to insertion-order queue when cached
2. For timestamp eviction:
   - Process queue from oldest to newest
   - Remove entries older than threshold
   - Stop when hitting newer entry (leverages insertion order)
3. For size eviction:
   - Process queue from oldest to newest
   - Remove entries until target size freed
   - Guarantees oldest entries are removed first

https://github.com/apache/pulsar/blob/b72bc4ff3aa5c9c45d9233d2d000429b3cf0ce1a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeCacheRemovalQueue.java

Note: There's a single shared removal queue for all ManagedLedgerImpl instances instead of having to do the check in multiple instances.

## Verifying this change

This change is already covered by existing tests:
- All existing cache-related tests continue to pass
- `RangeCacheTest` validates the new removal queue functionality
- `EntryCacheManagerTest` verifies eviction behavior remains correct
- Integration tests ensure no regression in cache performance

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -56,7 +56,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -307,7 +307,7 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
     private final OrderedScheduler scheduledExecutor;
 
     @Getter
-    protected final Executor executor;
+    protected final ExecutorService executor;
 
     @Getter
     private final ManagedLedgerFactoryImpl factory;
@@ -4836,5 +4836,26 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
     @Override
     public long getMetadataCreationTimestamp() {
         return ledgersStat != null ? ledgersStat.getCreationTimestamp() : 0;
+    }
+
+    /**
+     * Waits for Managed Ledger level pending cache evictions to complete. This doesn't wait for the size or time
+     * based evictions to complete. Use the {@link ManagedLedgerFactoryImpl#waitForPendingCacheEvictions()} for that
+     * purpose.
+     * This is for testing purposes only, so that we can ensure all cache evictions are done before proceeding with
+     * further operations.
+     */
+    @VisibleForTesting
+    public void waitForPendingCacheEvictions() {
+        try {
+            // currently it's sufficient to just submit a no-op task to the executor and wait for its completion.
+            executor.submit(() -> {
+                // no-op
+            }).get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e);
+        }
     }
 }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java
@@ -324,9 +324,11 @@ public class ManagedLedgerFactoryImpl implements ManagedLedgerFactory {
     }
 
     /**
-     * Waits for all pending cache evictions to complete.
+     * Waits for all pending cache evictions based on total cache size or entry TTL to complete.
      * This is for testing purposes only, so that we can ensure all cache evictions are done before proceeding with
-     * further operations.
+     * further operations. Please notice that Managed Ledger level cache eviction is not handled here. There's
+     * a similar method {@link ManagedLedgerImpl#waitForPendingCacheEvictions()} that waits for pending cache evictions
+     * at the Managed Ledger level.
      */
     @VisibleForTesting
     public void waitForPendingCacheEvictions() {
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
@@ -1809,40 +1809,40 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
         c2.setReadPosition(p3);
 
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 4);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
         c1.setReadPosition(p2);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 3);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
         c1.setReadPosition(p3);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 2);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
 
         ledger.deactivateCursor(c1);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 2); // as c2.readPosition=p3 => Cache contains p3,p4
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
 
         c2.setReadPosition(p4);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
 
         ledger.deactivateCursor(c2);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 0);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
     }
@@ -1875,7 +1875,7 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
         c2.setReadPosition(p4);
         c2.markDelete(p3);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 4);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
@@ -1883,7 +1883,7 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
         c1.setReadPosition(p3);
         c1.markDelete(p2);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 3);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
@@ -1891,26 +1891,26 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
         c1.setReadPosition(p4);
         c1.markDelete(p3);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 2);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
 
         ledger.deactivateCursor(c1);
 
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7 * 2);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
 
         c2.markDelete(p4);
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 7);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
 
 
         ledger.deactivateCursor(c2);
-        factory.waitForPendingCacheEvictions();
+        ledger.waitForPendingCacheEvictions();
         assertEquals(entryCache.getSize(), 0);
         assertEquals(cacheManager.getSize(), entryCache.getSize());
     }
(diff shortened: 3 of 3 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeCache.java
```diff
@@ -0,0 +1,287 @@
...
+ * Removing an entry multiple times could result in the entries of the cache being released multiple times,
+ * even while they are still in use. This is prevented by using a custom wrapper around the value to store in the map
+ * that ensures that the value is removed from the map only if the exact same instance is present in the map.
+ * There's also a check that ensures that the value matches the key. This is used to detect races without impacting
+ * consistency.
+ */
+@Slf4j
+class RangeCache {
+    private final ConcurrentNavigableMap<Position, RangeCacheEntryWrapper> entries;
+    private final RangeCacheRemovalQueue removalQueue;
```
heesung-sn, line 47: Is removal O(1)? Do we need to scan items in the queue to remove an item?

heesung-sn: 1. what happens if some entries are reused repeatedly? Do we move them to the tail of the removal queue? Shouldn't we deprioritize evicting those entries?(e.g large fanout cases, subscriptions with many consumers)

2. Dont we still need to scan entire list when all items timestamp are less than the threshold?(we dont do early stop here)

3. Is this cache per ledger? If yes, Why not a single global cache shared with all ledgers?

lhotari: > 3\. Is this cache per ledger? If yes, Why not a single global cache shared with all ledgers?

this PR doesn't change that. I think it's an efficient solution that caches are per ledger for the broker use cases. I don't think that a single global cache would bring any benefits over the current solution.

This PR addresses the problems in the existing implementation about evicting expired entries and keeping the total global cache size under a limit. That has been a problem in having the individual caches, but after this PR, the problem is solved.

lhotari: > 4\. Do we need to worry about the size limit of the removal queue from too many empty wrapper entries?

The RangeCacheEntryWrapper causes an overhead about 100 bytes per entry currently. With 100,000 entries in the queue, that would be about 10 MB of heap. That shouldn't be a problem.

More details about RangeCacheEntryWrapper and its background:
The wrapper entries are recycled so the wrapper entries don't add GC overhead. These wrappers are already used in the RangeCache to ensure consistency while using StampedLocks to avoid synchronization and blocking. The wrapper solution was added to RangeCache in #22789, with follow ups in #22814, #22818, #23903 and #23955. Without the wrapper there were issues where the entry got evicted and recycled while it was getting used. It's explained in the PR descriptions and there are also test cases to reproduce the problems. 
The design of the RangeCache is intentionally "racy" so that we wouldn't need to synchronize and block. StampedLock provides the necessary consistency together with the reference counting. The RangeCache's EntryWrapper is also a foundation for this removal queue work since it makes it possible to have the separate EntryCache instances per ledger and centralize the expiration and size based eviction to the queue.

heesung-sohn: Thank you for answering my questions. I think this PR and direction looks good to me with the future improvements. 

But I guess this requires more discussions in the Pulsar community, too.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -3046,7 +3056,9 @@ protected void doDeleteLedgers(List<LedgerInfo> ledgersToDelete) {
             NUMBER_OF_ENTRIES_UPDATER.addAndGet(this, -ls.getEntries());
             TOTAL_SIZE_UPDATER.addAndGet(this, -ls.getSize());
 
-            entryCache.invalidateAllEntries(ls.getLedgerId());
+            getFactory().getCacheEvictionExecutor().execute(() -> {
```
merlimat: Why do we need to evict from separate thread? Could we do it inside the entry cache implementation?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/CachedEntry.java
```diff
@@ -18,20 +18,13 @@
...
-     * Perform the cache eviction of at least sizeToFree bytes on the supplied list of caches.
-     *
-     * @param caches
-     *            the list of caches to consider
-     * @param sizeToFree
-     *            the minimum size in bytes to be freed
-     */
-    void doEviction(List<EntryCache> caches, long sizeToFree);
+public interface CachedEntry extends Entry, ReferenceCounted {
+    boolean matchesKey(Position key);
```
merlimat: what does the matching mean? is it equality or lower than?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheManagerImpl.java
```diff
@@ -174,33 +204,22 @@ public double getCacheEvictionWatermark() {
...
-    public void clear() {
-        caches.values().forEach(EntryCache::clear);
-    }
-
-    public static Entry create(long ledgerId, long entryId, ByteBuf data) {
-        return EntryImpl.create(ledgerId, entryId, data);
+    public void doCacheEviction(long maxTimestamp) {
+        // this method is expected to be called from the cache eviction executor
+        CompletableFuture<Void> evictionCompletionFuture = new CompletableFuture<>();
+        evictionInProgress.set(evictionCompletionFuture);
```
merlimat, line 210: should we check here if there's already a pending eviction in progress?

merlimat (commented):
My only question is about the eviction from the stash queue: since the insertion might not be in order, when there is memory pressure, how is the eviction going to happen? 

Or, is the size of this stash already controlled in size such that it would only require time-based expiration?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/CachedEntry.java
lhotari: It means that the cached entry has a key that matches the given key. Since `Position` is an interface, it's not exactly equality.

lhotari: an alternative method name could be `hasKey`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
lhotari: Correct. In this case, there's no need for doing that in a separate thread.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheManagerImpl.java
lhotari: This method is called from the cacheEvictionExecutor which is single threaded. There could never be a pending eviction in progress when the doCacheEviction method is called.

the `evictionInProgress` field is used in the `triggerEvictionWhenNeeded` to find out whether there's a pending eviction in progress so that it doesn't trigger a new eviction when one is already in the queue. I now noticed that there's some leftover code from the solution I needed for fixing tests that assert cache evictions (that now happen asynchronously in the cacheEvictionExecutor) and I found a simpler solution for that without the need to have a CompletableFuture available for the eviction that is in progress. 
I'll replace the `evictionInProgress` solution something different that addresses the purpose without extra overhead.

lhotari:
The merge conflicts will be resolved by first merging #24552 to master branch.

lhotari:
> My only question is about the eviction from the stash queue: since the insertion might not be in order, when there is memory pressure, how is the eviction going to happen?
> Or, is the size of this stash already controlled in size such that it would only require time-based expiration?

For the implementation in this PR all entries are added in order to the removal queue and the timestamp is set when the entry gets added to the cache and the removal queue. Therefore, the timestamps are in order. 
When evicting to reduce the cache size under the watermark to handle "memory pressure", oldest entries are evicted from the queue. Time based expiration is done by evicting from the head of the queue until there's a newer timestamp on the entry.

The eviction will become more complicated later. In PIP-430, there's a need to put entries aside and keep them in the cache for a longer period of time. Currently that is addressed in [PIP-430 by having 2 separate settings for expiration](https://github.com/lhotari/pulsar/blob/lh-pip-430/pip/pip-430.md#2-expected-read-count-cache-strategy) and that will simplify the solution. The maximum TTL for cache entries will continue to be in order in the removal queue, but another datastructure is needed when size based eviction would remove the entry, but it's prioritized to be skipped and kept until the entry expires. Handling that challenge is out-of-scope for this PR. After this current PR has been merged, it will be possible to build upon this and expand the solution further to implement PIP-430 (I have a WIP PR in my own fork, but it's not matching the PIP high-level design).

One more detail about eviction handling that is present in this PR. The central queue is used for eviction by size (to keep all cached items under the limit) or by TTL. When entries are evicted directly from the RangeCache (for a complete ledger or for up to the markdelete position), the entry wrapper held in the range cache will be cleared and the same instance will remain in the queue until it gets processed. Empty entry wrappers will be held in the queue at most for the TTL in worst case. The entry wrapper gets recycled when it gets processed from the `RangeCacheRemovalQueue`.

I hope this answer covers your question @merlimat

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/CachedEntry.java
lhotari: I ended up with `matchesPosition` directly in the Entry interface while providing a default implementation. There's already a `matchesKey` in EntryImpl so that has been renamed. 
I also ended up eliminating the need for a separate "CachedEntry". Instead, there's now `ReferenceCountedEntry` which combines `Entry` and `ReferenceCounted`. `EntryImpl` already implements this.

merlimat (approved):

lhotari:
PIP-430 vote has passed. I'll merge this PR to master.

Issue #24148, opened 2025-04-01
[Bug] Compaction of __change_events topic is blocking forceful namespace/topic deletion


pdolif:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

Latest master

### Minimal reproduce step

I was investigating #22736. Sometimes, tests fail because `MockedPulsarServiceBaseTest.deleteNamespaceWithRetry` times out. This seems related to a compaction of the `persistent://public/default/__change_events` topic.

To reproduce the timeout of `deleteNamespaceWithRetry`, add the following test to `NamespaceAuthZTest`:

```java
@Test
public void reproduce() throws Exception {
    final String namespace = "public/default";
    final String topic = "persistent://" + namespace + "/" + UUID.randomUUID();

    superUserAdmin.topics().createNonPartitionedTopic(topic);
    superUserAdmin.topics().delete(topic, true);

    final AtomicInteger succeeded = new AtomicInteger(0);
    
    new Thread(() -> {
        try {
            deleteNamespaceWithRetry(namespace, true, superUserAdmin);
            succeeded.incrementAndGet();
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
    }).start();

    new Thread(() -> {
        try {
            Thread.sleep(400);
            superUserAdmin.topics().triggerCompaction("persistent://public/default/__change_events");
            succeeded.incrementAndGet();
        } catch (PulsarAdminException | InterruptedException e) {
            throw new RuntimeException(e);
        }

    }).start();

    await().atMost(60, TimeUnit.SECONDS).untilAsserted(() -> assertEquals(succeeded.get(), 2));
}
```

The test calls `deleteNamespaceWithRetry` in one thread and triggers a compaction of the `persistent://public/default/__change_events` topic shortly after the namespace deletion was requested. For me, a delay of 400ms works well and every test run reproduces the issue.
To avoid confusion, remove the call of `deleteNamespaceWithRetry` from `after() `method in the test class.

### What did you expect to see?

I would expect the above test to pass.

### What did you see instead?

Test fails.

### Anything else?

From my understanding, the problem is the following:
[...]

PR #23462, opened 2024-10-15
[improve] [pip] PIP-387: Modify interface TopicCompactionService to support cancelling an in-progress compaction task

poorbarcode:
### Motivation

- [DIscussion](https://lists.apache.org/thread/brh9816o0xnkkxqbpndj69c25opg39lo)
- [Vote]

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24366, opened 2025-05-30
[fix][broker]Fix deadlock when compaction and topic deletion execute concurrently

poorbarcode:
#### **Issue 1**: deadlock when compaction and topic deletion execute concurrently
Reproduction steps of the deadlock( see also the new test `testConcurrentCompactionAndTopicDelete` )
1. `thread-compaction`: create a raw reader.
2. `thread-topic-deleting`: mark the topic as deleted and disconnect clients, including the compaction reader.
3. `thread-compaction`: The raw reader attempts to reconnect due to the disconnection in step "2-1".
4. `thread-topic-deleting`: unsubscribe all subscriptions that contain "__compaction".
5. `thread-topic-deleting`: deleting "__compaction" cursor will wait for the in-progress compaction task.
6. `thread-compaction`: the raw read can not connect successfully anymore because the topic was marked as  "fenced && isDeleting".
7. `deadlock`: "thread-topic-deleting" waiting for compaction task being finished, thread-compaction continuously reconnects.

Note:
- If the `thread-compaction` is stuck when the first phase of compaction calling `reader.readNextAsync()` at step 6, the deadlock can be solved after `{brokerServiceCompactionPhaseOneLoopTimeInSeconds)` seconds
- But regarding other scenarios, such as `reader.getLastMessageIdAsync`... the deadlock will persist forever.
 
You can reproduce the issue by the new test `testConcurrentCompactionAndTopicDelete`

---

#### **Issue 2-1**: The responded compostable future of `Consumer.receiveAsync()` never complete if the topic is deleted

**Background**: 
- Enabling `topic auto creation`: Client triggers a new topic creation if users delete a topic with `-f`.
- Disabling `topic auto creation`: Client stops retrying reconnection and sets the consumer's state to `Failed`.
- Issue: regarding the scenario "Disabling topic-auto-creation", the client set the consumer's state to `Failed`, but it does not trigger a completion for the pending receive requests, leading to the pending receive future never completing. See https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L988-L999

```java
setState(State.Failed);
closeConsumerTasks();
client.cleanupConsumer(this);
// Issue: it should call "failPendingReceive()" also
```

You can reproduce the issue by the new test `testReceiveWillDoneAfterTopicDeleted`

#### **Issue 2-2**: resource leak that contains a dead letter producer and a retry letter producer.
Regarding `issue 2-1`, the client also forgot to release the related dead letter producer and the retry letter producer, which also caused a resource leak. Since there are other issues that prevent the new test from passing, I will write a separate PR to add the test to cover this case, which also contains another fix.

---

#### **Issue 3**: Even though the client received a `TopicNotFound` error, it continuously retries.
Root cause: Pulsar client prevents retrying when it receives a `TopicNotFound` error from the `Subscribe` command, but the `TopicNotFound` can also be received from the process `get broker connections of the target topic`.

You can reproduce the issue by the new test `testReceiveWillDoneAfterTopicDeleted`

---

### Modifications

- Regarding issue 1 - a deadlock error, closes the RawReader if the compaction task received a ServiceNotReady error, which happens when a topic/subscription is fenced, or the topic/namespace is loading/unloading. Since the RawReader is an internal API, this change is safe.
- Regarding issue 2 - consumer forgets to complete the pending read request when it stops reconnecting, calls `consumer.close()` instead of calling a part of the stats modifying, which is more complete.
- Regarding issue 3 - continues to reconnect even though it received a TopicNotFound error, stops reconnecting after receiving an unrecovered error from grabbing new connections, and closes consumers/producers.

---

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

---

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -1511,28 +1511,6 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             }
 
             fenceTopicToCloseOrDelete(); // Avoid clients reconnections while deleting
-            // Wait compaction to avoid a deadlock below:
-            // 1. thread-compaction : create a raw reader.
-            // 2. thread-topic-deleting : mark topic as deleting, and disconnect clients that includes compaction
-            //      reader.
-            // 3. thread-compaction: the raw reader attempts to reconnect due to the disconnection in step "2-1".
-            // 4. thread-topic-deleting: unsubscribe all subscriptions, which contains "__compaction".
-            // 5. thread-topic-deleting: deleting "__compaction" cursor will wait for the in-progress compaction task.
-            // 6. thread-compaction: the raw read can not connect successfully anymore because the topic was marked as
-            //    "fenced && isDeleting".
-            // 7. deadlock: "thread-topic-deleting" waiting for compaction task being finished, thread-compaction
-            //    continuously reconnects.
-            // To solve the issue, let "thread-topic-deleting" release the fencing&deleting state, wait for the
-            // compaction task being finished. Since "thread-topic-deleting" never release "lock.writeLock", the broker
-            // will not trigger a new compaction task.
-            if (currentCompaction != null && !currentCompaction.isDone()) {
-                unfenceTopicToResume();
-                return currentCompaction
-                    .thenCompose(__ -> delete(failIfHasSubscriptions, failIfHasBacklogs, closeIfClientsConnected))
-                    .whenComplete((__, ex) -> {
-                        lock.writeLock().unlock();
-                    });
-            }
             // Mark the progress of close to prevent close calling concurrently.
             this.closeFutures =
                     new CloseFutures(new CompletableFuture(), new CompletableFuture(), new CompletableFuture());
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -270,6 +270,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
     private volatile double lastUpdatedAvgPublishRateInMsg = 0;
     private volatile double lastUpdatedAvgPublishRateInByte = 0;
 
+    @Getter
     private volatile boolean isClosingOrDeleting = false;
 
     private ScheduledFuture<?> fencedTopicMonitoringTask = null;
@@ -1510,6 +1511,28 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             }
 
             fenceTopicToCloseOrDelete(); // Avoid clients reconnections while deleting
+            // Wait compaction to avoid a deadlock below:
+            // 1. thread-compaction : create a raw reader.
+            // 2. thread-topic-deleting : mark topic as deleting, and disconnect clients that includes compaction
+            //      reader.
+            // 3. thread-compaction: the raw reader attempts to reconnect due to the disconnection in step "2-1".
+            // 4. thread-topic-deleting: unsubscribe all subscriptions, which contains "__compaction".
+            // 5. thread-topic-deleting: deleting "__compaction" cursor will wait for the in-progress compaction task.
+            // 6. thread-compaction: the raw read can not connect successfully anymore because the topic was marked as
+            //    "fenced && isDeleting".
+            // 7. deadlock: "thread-topic-deleting" waiting for compaction task being finished, thread-compaction
+            //    continuously reconnects.
+            // To solve the issue, let "thread-topic-deleting" release the fencing&deleting state, wait for the
+            // compaction task being finished. Since "thread-topic-deleting" never release "lock.writeLock", the broker
+            // will not trigger a new compaction task.
+            if (currentCompaction != null && !currentCompaction.isDone()) {
+                unfenceTopicToResume();
+                return currentCompaction
+                    .thenCompose(__ -> delete(failIfHasSubscriptions, failIfHasBacklogs, closeIfClientsConnected))
+                    .whenComplete((__, ex) -> {
+                        lock.writeLock().unlock();
+                    });
+            }
             // Mark the progress of close to prevent close calling concurrently.
             this.closeFutures =
                     new CloseFutures(new CompletableFuture(), new CompletableFuture(), new CompletableFuture());
--- a/pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawReaderImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawReaderImpl.java
@@ -53,7 +53,7 @@ public class RawReaderImpl implements RawReader {
 
     public RawReaderImpl(PulsarClientImpl client, String topic, String subscription,
                          CompletableFuture<Consumer<byte[]>> consumerFuture,
-                         boolean createTopicIfDoesNotExist) {
+                         boolean createTopicIfDoesNotExist, boolean retryOnRecoverableErrors) {
         consumerConfiguration = new ConsumerConfigurationData<>();
         consumerConfiguration.getTopicNames().add(topic);
         consumerConfiguration.setSubscriptionName(subscription);
@@ -63,14 +63,16 @@ public class RawReaderImpl implements RawReader {
         consumerConfiguration.setSubscriptionInitialPosition(SubscriptionInitialPosition.Earliest);
         consumerConfiguration.setAckReceiptEnabled(true);
 
-        consumer = new RawConsumerImpl(client, consumerConfiguration, consumerFuture, createTopicIfDoesNotExist);
+        consumer = new RawConsumerImpl(client, consumerConfiguration, consumerFuture, createTopicIfDoesNotExist,
+                retryOnRecoverableErrors);
     }
 
     public RawReaderImpl(PulsarClientImpl client, ConsumerConfigurationData<byte[]> consumerConfiguration,
                          CompletableFuture<Consumer<byte[]>> consumerFuture,
-                         boolean createTopicIfDoesNotExist) {
+                         boolean createTopicIfDoesNotExist, boolean retryOnRecoverableErrors) {
         this.consumerConfiguration = consumerConfiguration;
-        consumer = new RawConsumerImpl(client, consumerConfiguration, consumerFuture, createTopicIfDoesNotExist);
+        consumer = new RawConsumerImpl(client, consumerConfiguration, consumerFuture, createTopicIfDoesNotExist,
+                retryOnRecoverableErrors);
     }
 
 
@@ -118,9 +120,11 @@ public class RawReaderImpl implements RawReader {
     static class RawConsumerImpl extends ConsumerImpl<byte[]> {
         final BlockingQueue<RawMessageAndCnx> incomingRawMessages;
         final Queue<CompletableFuture<RawMessage>> pendingRawReceives;
+        final boolean retryOnRecoverableErrors;
 
         RawConsumerImpl(PulsarClientImpl client, ConsumerConfigurationData<byte[]> conf,
-                CompletableFuture<Consumer<byte[]>> consumerFuture, boolean createTopicIfDoesNotExist) {
+                CompletableFuture<Consumer<byte[]>> consumerFuture, boolean createTopicIfDoesNotExist,
+                boolean retryOnRecoverableErrors) {
             super(client,
                     conf.getSingleTopic(),
                     conf,
@@ -136,11 +140,12 @@ public class RawReaderImpl implements RawReader {
             );
             incomingRawMessages = new GrowableArrayBlockingQueue<>();
             pendingRawReceives = new ConcurrentLinkedQueue<>();
+            this.retryOnRecoverableErrors = retryOnRecoverableErrors;
         }
 
         protected boolean isUnrecoverableError(Throwable t) {
-            if (t instanceof PulsarClientException.ServiceNotReadyException) {
-                return false;
+            if (!retryOnRecoverableErrors && (t instanceof PulsarClientException.ServiceNotReadyException)) {
+                return true;
             }
             return super.isUnrecoverableError(t);
         }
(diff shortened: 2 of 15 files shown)


lhotari:
@poorbarcode Does this resolve #24148?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1510,6 +1511,28 @@ private CompletableFuture<Void> delete(boolean failIfHasSubscriptions,
...
+            //    "fenced && isDeleting".
+            // 7. deadlock: "thread-topic-deleting" waiting for compaction task being finished, thread-compaction
+            //    continuously reconnects.
+            // To solve the issue, let "thread-topic-deleting" release the fencing&deleting state, wait for the
+            // compaction task being finished. Since "thread-topic-deleting" never release "lock.writeLock", the broker
+            // will not trigger a new compaction task.
+            if (currentCompaction != null && !currentCompaction.isDone()) {
+                unfenceTopicToResume();
+                return currentCompaction
+                    .thenCompose(__ -> delete(failIfHasSubscriptions, failIfHasBacklogs, closeIfClientsConnected))
```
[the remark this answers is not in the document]
lhotari: If the compaction isn't yet finished, it would execute `delete` on the compaction executor and deadlock.

https://github.com/apache/pulsar/blob/12806727c8eb8cf17fded4ef11e786718b432444/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java#L1620-L1626

If it's already finished, there would be no deadlock.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1510,6 +1511,28 @@ private CompletableFuture<Void> delete(boolean failIfHasSubscriptions,
...
+            // 5. thread-topic-deleting: deleting "__compaction" cursor will wait for the in-progress compaction task.
+            // 6. thread-compaction: the raw read can not connect successfully anymore because the topic was marked as
+            //    "fenced && isDeleting".
+            // 7. deadlock: "thread-topic-deleting" waiting for compaction task being finished, thread-compaction
+            //    continuously reconnects.
+            // To solve the issue, let "thread-topic-deleting" release the fencing&deleting state, wait for the
+            // compaction task being finished. Since "thread-topic-deleting" never release "lock.writeLock", the broker
+            // will not trigger a new compaction task.
+            if (currentCompaction != null && !currentCompaction.isDone()) {
+                unfenceTopicToResume();
```
lhotari, line 1529: on a few lines above there's `fenceTopicToCloseOrDelete()`. Would it be possible to avoid fencing if it's going to be unfenced here?

poorbarcode: The current task will set `isFenced` and `isClosingOrDeleting` to `true` here(https://github.com/apache/pulsar/pull/24366/files#diff-5edf14cc6f25857d0cfdd26b2d3b3141230ecfb0dfa95aebf7583fd76ede4c4bR1513), the method `unfenceTopicToResume` is the negative method of `fenceTopicToCloseOrDelete`.

There is no protection of guaranteeing `isFenced` and `isClosingOrDeleting` atomically modifying, but it was not introduced in the current PR. We may need to merge `transferring`, `isFenced`, and `isClosingOrDeleting` into one in a separate PR to guarantee atomicity

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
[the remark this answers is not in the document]
poorbarcode: > Does this code get executed in tests?

Yes, it does.

> If the compaction isn't yet finished, it would execute delete on the compaction executor and deadlock.

Could you explain the details of the deadlock you mentioned?

poorbarcode:
@codelipenghui 

> Actually, we are safe to give up the reconnect from the compaction reader if the topic is deleting.
> Now, the solution is to unfence the topic if there is ongoing compaction task. But it unfencing will also make producers/consumers from user side available to connect to this topic which should not be expected.
> So, I think the correct fix should be find a way to expose the "Topic is deleting Exception" to Raw reader and the reader can optionally give up the retry.

The solution you mentioned is what [PIP-387: Modify interface TopicCompactionService to support cancelling an in-progress compaction task](https://github.com/apache/pulsar/pull/23462), which has been canceled 😂

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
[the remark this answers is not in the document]
lhotari: > Could you explain the details of the deadlock you mentioned?

already explained in https://github.com/apache/pulsar/pull/24366#discussion_r2115906863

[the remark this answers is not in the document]
poorbarcode: > already explained in https://github.com/apache/pulsar/pull/24366#discussion_r2115906863
>> If the compaction isn't yet finished, it would execute delete on the compaction executor and deadlock.

There is no deadlock if tasks execute as you mentioned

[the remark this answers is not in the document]
lhotari: > There is no deadlock if tasks execute as you mentioned

how come since `delete` would require a lock in a different thread and the lock is currently held by another thread.

gaoran10:
We'd better make the compaction task aware of the topic deletion operation and cancel the task.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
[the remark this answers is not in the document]
poorbarcode: After you list the steps that you mentioned, the scenarioof  how deadlock occurs, you will know that there is no deadlock

poorbarcode:
@codelipenghui @gaoran10 @lhotari I have used a new solution to solve the issue: cancel the pending compaction when deleting the topic, please review again

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java
```diff
@@ -64,7 +64,8 @@ interface Connection {
          * @apiNote If the returned future is completed exceptionally, reconnectLater will be called.
          */
         CompletableFuture<Void> connectionOpened(ClientCnx cnx);
-        default void connectionFailed(PulsarClientException e) {
+        default boolean connectionFailed(PulsarClientException e) {
```
lhotari, line 74: would it be possible to add a javadoc for this method so that the purpose of the method is explained and the return value is explained?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TopicListWatcher.java
```diff
@@ -89,7 +89,7 @@ public TopicListWatcher(PatternConsumerUpdateQueue patternConsumerUpdateQueue,
     }
 
     @Override
-    public void connectionFailed(PulsarClientException exception) {
+    public boolean connectionFailed(PulsarClientException exception) {
         boolean nonRetriableError = !PulsarClientException.isRetriableError(exception);
         if (nonRetriableError) {
```
lhotari, line 94: just wondering why there isn't `return false` in the end of this code block

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TransactionMetaStoreHandler.java
```diff
@@ -122,7 +122,7 @@ public void start() {
     }
 
     @Override
-    public void connectionFailed(PulsarClientException exception) {
+    public boolean connectionFailed(PulsarClientException exception) {
         boolean nonRetriableError = !PulsarClientException.isRetriableError(exception);
         boolean timeout = System.currentTimeMillis() > lookupDeadline;
         if (nonRetriableError || timeout) {
```
lhotari, line 128: just wondering why there isn't `return false` in the end of this code block

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java
poorbarcode: Added a doc

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TransactionMetaStoreHandler.java
poorbarcode: Fixed, thanks for mentioning

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TopicListWatcher.java
poorbarcode: Fixed, thanks for mentioning

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawReaderImpl.java
```diff
@@ -137,6 +138,13 @@ static class RawConsumerImpl extends ConsumerImpl<byte[]> {
             pendingRawReceives = new ConcurrentLinkedQueue<>();
         }
 
+        protected boolean isUnrecoverableError(Throwable t) {
+            if (t instanceof PulsarClientException.ServiceNotReadyException) {
+                return false;
+            }
+            return super.isUnrecoverableError(t);
+        }
+
```
codelipenghui: Can we add an option to the Raw reader such as "retryOnRecoverableErrors" and keep it as true by default?

- Do not break any cases might build based on the Raw Reader
- Compaction service and set to false explicitly to make the code easy to understand

pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
```diff
@@ -2320,6 +2328,65 @@ public void testAcknowledgeWithReconnection() throws Exception {
...
+        consumer.close();
+        CompletableFuture<Void> deleteTopicFuture = persistentTopic.deleteForcefully();
+
+        // Remove the injection after 3s.
+        Thread.sleep(3000);
+        delayReadSignal.countDown();
+
+        // Verify: topic deletion is successfully executed.
+        Awaitility.await().atMost(15, TimeUnit.SECONDS).untilAsserted(() -> {
+            assertTrue(deleteTopicFuture.isDone());
```
codelipenghui, line 2386: It seems the test failed here, please check.

pulsar-broker/src/main/java/org/apache/pulsar/client/impl/RawReaderImpl.java
poorbarcode: - Added a new param named `retryOnRecoverableErrors`
- And added a new test named `RawReaderTest.testReconnectsWhenServiceNotReady` to cover the case that the raw-read continuously reconnects if `retryOnRecoverableErrors ` is `true`

pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
poorbarcode: Fixed

codelipenghui (approved):

lhotari:
@poorbarcode Please handle cherry-picking to branch-3.0 and branch-3.3 since there are merge conflicts.

lhotari:
> @poorbarcode Please handle cherry-picking to branch-3.0 and branch-3.3 since there are merge conflicts.

@poorbarcode branch-3.0 compilation fails in checkstyle.
```
[INFO] There are 5 errors reported by Checkstyle 10.14.2 with ../buildtools/src/main/resources/pulsar/checkstyle.xml ruleset.
[ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[53,8] (imports) UnusedImports: Unused import: org.apache.pulsar.client.api.SubscriptionInitialPosition.
[ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[54,8] (imports) UnusedImports: Unused import: org.apache.pulsar.client.api.SubscriptionType.
[ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[55,8] (imports) UnusedImports: Unused import: org.apache.pulsar.client.impl.conf.ConsumerConfigurationData.
[ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[56,8] (imports) UnusedImports: Unused import: org.apache.pulsar.common.api.proto.BrokerEntryMetadata.
[ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[70,15] (imports) UnusedImports: Unused import: org.apache.pulsar.client.impl.RawReaderImpl.DEFAULT_RECEIVER_QUEUE_SIZE.
```
After cherry-picking, please make sure that checkstyle passes and the project builds.
A quick way to validate is to run this command:
```
mvn -Pcore-modules,-main -T 1C clean install -DskipTests -Dspotbugs.skip=true -DnarPluginPhase=none
```

lhotari:
> > @poorbarcode Please handle cherry-picking to branch-3.0 and branch-3.3 since there are merge conflicts.
> 
> @poorbarcode branch-3.0 compilation fails in checkstyle.
> 
> ```
> [INFO] There are 5 errors reported by Checkstyle 10.14.2 with ../buildtools/src/main/resources/pulsar/checkstyle.xml ruleset.
> [ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[53,8] (imports) UnusedImports: Unused import: org.apache.pulsar.client.api.SubscriptionInitialPosition.
> [ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[54,8] (imports) UnusedImports: Unused import: org.apache.pulsar.client.api.SubscriptionType.
> [ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[55,8] (imports) UnusedImports: Unused import: org.apache.pulsar.client.impl.conf.ConsumerConfigurationData.
> [ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[56,8] (imports) UnusedImports: Unused import: org.apache.pulsar.common.api.proto.BrokerEntryMetadata.
> [ERROR] src/test/java/org/apache/pulsar/client/impl/RawReaderTest.java:[70,15] (imports) UnusedImports: Unused import: org.apache.pulsar.client.impl.RawReaderImpl.DEFAULT_RECEIVER_QUEUE_SIZE.
> ```
> 
> After cherry-picking, please make sure that checkstyle passes and the project builds. A quick way to validate is to run this command:
> 
> ```
> mvn -Pcore-modules,-main -T 1C clean install -DskipTests -Dspotbugs.skip=true -DnarPluginPhase=none
> ```

@poorbarcode I fixed it in https://github.com/apache/pulsar/commit/b92af789

PR #24368, opened 2025-05-30
[improve][pip]PIP-422 Support global topic-level policy: replicated clusters and new API to delete topic-level policies

poorbarcode:
- [DIscussion](https://lists.apache.org/thread/5rh536dqm2o35tzop5jl7b5x7r291pv2)
- [Vote](https://lists.apache.org/thread/p0wmnfr3jkmfzh69jfh7jv10184g1vlm)

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pip/pip-422.md
+++ b/pip/pip-422.md
@@ -98,4 +98,4 @@ Passed solutions:
 # Links
 
 * Mailing List discussion thread: https://lists.apache.org/thread/5rh536dqm2o35tzop5jl7b5x7r291pv2
-* Mailing List voting thread:
+* Mailing List voting thread: https://lists.apache.org/thread/p0wmnfr3jkmfzh69jfh7jv10184g1vlm
--- a/pip/pip-422.md
+++ b/pip/pip-422.md
@@ -59,7 +59,7 @@ B. Delete a partitioned topic under a specify cluster.
 ### Public API
 
 **new Pulsar Admin API**
-- `pulsar-admin topicPolicies --clusters <clusters> --global <topic>`
+- `pulsar-admin topicPolicies --set-replication-cluster <clusters> --global <topic>`
 - `pulsar-admin topicPolicies delete <topic>`
 
 ### Binary protocol
--- a/pip/pip-422.md
+++ b/pip/pip-422.md
@@ -30,6 +30,10 @@ Add two new APIs
 - `pulsar-admin topicPolicies delete <topic>`
   - To delete topic level policies, including global and local policies, it can be called even if the topic has been deleted.
 
+Old Apis
+- Prohibits users from using local policies to delete the local cluster.
+
+
 The new APIs will be used for the following scenarios:
 
 A. Delete a partitioned topic under all clusters, if enabled Geo-Replication with a shared config metadata store.
@@ -57,6 +61,7 @@ B. Delete a partitioned topic under a specified cluster.
 **new Pulsar Admin API**
 - `pulsar-admin topicPolicies --set-replication-cluster <clusters> --global <topic>`
 - `pulsar-admin topicPolicies delete <topic>`
+- The API `pulsar-admin topics --set-replication-cluster <clusters> <topic>` never supports to delete the local cluster anymore.
 
 ### Binary protocol
 Nothing.
--- a/pip/pip-422.md
+++ b/pip/pip-422.md
@@ -2,25 +2,25 @@
 
 # Background knowledge
 
-Users have two choices of config metadata store when using the feature Geo-Replication: shared or not.
-All clusters share data that are stored in global config metadata store if users choose to use a shared metadata store.
-By the way, the share config metadata store often is build crossing multi zone/region to ensure fault tolerance.
+Users have two choices when configuring the metadata store when using the feature Geo-Replication: shared or not.
+All clusters share data that are stored in the global config metadata store if users choose to use a shared metadata store.
+By the way, the share config metadata store is often built across multiple zones/regions to ensure fault tolerance.
 
-Since the data that are stored in the shared metadata store contains the following pulsar resources: tenants, namespaces, partitioned topic,
+Since the data that is stored in the shared metadata store contains the following pulsar resources: tenants, namespaces, partitioned topics,
 the Admin API can not be used to delete these pulsar resources for a single cluster.
-Pulsar has a feature that used to delete namespace level resources for a specify cluster when using a shared metadata store, it works as follows:
+Pulsar has a feature that is used to delete namespace-level resources for a specific cluster when using a shared metadata store, it works as follows:
 - The original namespace-level policy that named `replicated clusters` contains two clusters `c1` and `c2`.
 - The topics under the cluster `c1` will be deleted automatically when users remove `c1` from the namespace-level policy.
 
-The feature above very useful at the following scenarios:
-- Transfer a namespace from a old cluster to a new cluster: at last, the namespace should be deleted from the old cluster.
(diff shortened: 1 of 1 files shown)


pip/pip-422.md
```diff
@@ -0,0 +1,103 @@
+# PIP-422: Support global topic-level policy: replicated clusters and new API to delete topic-level policies
+
+# Background knowledge
+
+Users have two choices of config metadata store when using the feature Geo-Replication: shared or not.
+All clusters share data that are stored in global config metadata store if users choose to use a shared metadata store.
+By the way, the share config metadata store often is build crossing multi zone/region to confirm fault tolerance.
```
codelipenghui:
```suggestion
By the way, the share config metadata store often is build crossing multi zone/region to ensure fault tolerance.
```

pip/pip-422.md
```diff
@@ -0,0 +1,103 @@
...
+
+The feature above very useful at the following scenarios:
+- Transfer a namespace from a old cluster to a new cluster: at last, the namespace should be deleted from the old cluster.
+- Delete topics: since Pulsar does not allow deleting topics when enabled Geo-Replication, users should disable replication first.
+
+# Motivation
+
+When using shared metadata store and enabled Geo-Replication, there is no mechanism for the following scenarios:
+- Transfer a partitioned topic from a old cluster to a new cluster.
+- Delete a partitioned topic form a specify cluster.
```
codelipenghui:
```suggestion
- Delete a partitioned topic from a specify cluster.
```

pip/pip-422.md
```diff
@@ -0,0 +1,103 @@
...
+  - Once you want to delete the topic under all clusters, please use the solution A to remove the global topic-level policy left.
+
+**Question**: why not to add a mechanism that brokers clear global topic-level policies that the related topic was deleted regularly, which avoids to call `pulsar-admin topicPolicies delete <topic>`?
+
+**Answer**: When users quickly delete and create topics, it will result in the previous policies not being deleted by the scheduled cleaning mechanism, thereby causing confusion.
+
+### Public API
+
+**new Pulsar Admin API**
+- `pulsar-admin topicPolicies --clusters <clusters> --global <topic>`
```
codelipenghui: We already have `bin/pulsar-admin topics set-replication-cluster`, it seems we can only add an option instead of a new command.

pip/pip-422.md
```diff
@@ -0,0 +1,103 @@
...
+# Motivation
+
+When using shared metadata store and enabled Geo-Replication, there is no mechanism for the following scenarios:
+- Transfer a partitioned topic from a old cluster to a new cluster.
+- Delete a partitioned topic form a specify cluster.
+
+Passed solutions:
+1. Disable namespace-level Geo-Replication then delete the partitioned topic: this will trigger a deletion for other topics, which is not expected.
+2. Disable topic-level Geo-Replication under each cluster, and delete the partitioned topic by Admin API: it will delete the shared partitioned topic metadata, which was also deleted from another cluster.
+3. Disable topic-level Geo-Replication under each cluster, and delete the sub-topics manually by Admin API: this is the best solution without the current PIP, the defect is that the schema data and topic-level policies are still there.
```
codelipenghui: Is it better to move to `# Alternatives`?

pip/pip-422.md
```diff
@@ -0,0 +1,103 @@
...
+- `pulsar-admin topicPolicies -c <clusters> --g <topic>`
+  -  Pulsar only support setting a local topic-level `replicated clusters`, does not support global policies so far.
+- `pulsar-admin topicPolicies delete <topic>`
+  - To delete topic level policies, including global and local policies, it can be called even if the topic has been deleted.
+
+The new APIs will be used for the following scenarios:
+
+A. Delete a partitioned topic under all clusters, if enabled Geo-Replication with a shared config metadata store.
+- Disable global topic level `replicated clusters`, in other words, set `replicated clusters` with a single list.
+  - The clusters are not contained in the `replicated clusters` will delete the sub-topics automatically, such as the namespace-level implementation.
```
codelipenghui: If there is any active producers or consumers connected the cluster which been removed from the replicated cluster list, the topic will be create again if topic auto-creation is enabled?

Or it will rely on the global topic policy which defined the replication_cluster. All the producers and consumers will be rejected if the connected cluster is not listed in replicated_clusters?

pip/pip-422.md
codelipenghui: And for the `set-replication-cluster` without `--global = true`. We should only allow to remove the non-local cluster?

pip/pip-422.md
poorbarcode: > Or it will rely on the global topic policy which defined the replication_cluster. All the producers and consumers will be rejected if the connected cluster is not listed in replicated_clusters?

This one. I have added to the proposal. https://github.com/apache/pulsar/pull/24368/files#diff-fca5852dd938f5905286f6ed91af543d8037c02e12726389f9f2ade9ae427c14R46

pip/pip-422.md
poorbarcode: Moved them into `#Alternatives`

pip/pip-422.md
poorbarcode: > We already have bin/pulsar-admin topics set-replication-cluster, it seems we can only add an option instead of a new command.

Only the command `topicPolicies` supports the param `--global` so far, add the new command into `topicPolicies` is better the `topics`.

Renamed the sub command name `--clusters` -> `set-replication-cluster`

---

> And for the set-replication-cluster without --global = true. We should only allow to remove the non-local cluster?

Users can set a global policy from the remote side, then it will be replicated to the local cluster. After we added this limitation, users only can set a global policy on the local cluster

codelipenghui: Ok, we might need to consider to hidden the `bin/pulsar-admin topics set-replication-cluster` command since new command will be add to topicPolicies.

> Users can set a global policy from the remote side, then it will be replicated to the local cluster. After we added this limitation, users only can set a global policy on the local cluster


It should be different cases. I mean if user just use `set-replication-cluster` without `--global` but try to remove the local cluster from the cluster list, do we need to reject this request?

pip/pip-422.md
```diff
@@ -0,0 +1,101 @@
...
+# Motivation
+
+When using a shared metadata store and enabling Geo-Replication, there is no mechanism for the following scenarios:
+- Transfer a partitioned topic from an old cluster to a new cluster.
+- Delete a partitioned topic from a specified cluster.
+
+# Goals & Designs
+
+Add two new APIs
+- `pulsar-admin topicPolicies -c <clusters> --g <topic>`
```
codelipenghui:
```suggestion
- `pulsar-admin topicPolicies --set-replication-cluster -c <clusters> --g <topic>`
```

pip/pip-422.md
poorbarcode: Added to the proposal
- https://github.com/apache/pulsar/pull/24368/files#diff-fca5852dd938f5905286f6ed91af543d8037c02e12726389f9f2ade9ae427c14R34
- https://github.com/apache/pulsar/pull/24368/files#diff-fca5852dd938f5905286f6ed91af543d8037c02e12726389f9f2ade9ae427c14R64

poorbarcode: The PR of this part has been pushed out
- [ ] https://github.com/apache/pulsar/pull/24351

PR #24370, opened 2025-06-01
[improve][pip] PIP-423: Add a new admin API to acknowledge a single message

Denovo1998:
### Motivation

Operators and SREs occasionally need to intervene in a topic's backlog to handle problematic messages or adapt to changing business requirements. For instance:

*   **Cancelling Scheduled Actions:** A delayed message representing a future task (e.g., a scheduled report or a notification) may become obsolete. The most efficient way to handle this is to prevent its delivery entirely by acknowledging it pre-emptively.
*   **Removing Backlogs:** A specific message in a backlog might have a malformed payload that causes consumer applications to crash repeatedly. Removing this single "poison pill" message without affecting valid messages around it is a critical operational capability. This also applies to removing a single bad message from within a larger batch.
*   **Manual Business Logic Correction:** An event may have been sent that is later determined to be invalid due to external factors. An administrator needs a precise tool to remove this specific event from a subscription's delivery queue.

The existing `skip(numMessages)` API is a blunt instrument, ill-suited for these precise, targeted operations. This proposal introduces an administrative API to skip messages by their specific `MessageId` (including `ledgerId`, `entryId`, and optional `batchIndex`), providing a robust and reliable way to remove any individual message—delayed or not—from a subscription's backlog.

**Implementation PR:** https://github.com/apache/pulsar/pull/23907

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/Denovo1998/pulsar/pull/8

### DIFF
--- a/pip/pip-423.md
+++ b/pip/pip-423.md
@@ -3,10 +3,10 @@
 # Background knowledge
 
 *   **Message Identification (MessageId):** In Pulsar, every message is uniquely identified by its `MessageId`, which encapsulates a `ledgerId`, an `entryId`, and optionally a partition index. The combination of `ledgerId` and `entryId` (often represented as a `Position`) uniquely identifies a message's physical storage location within a topic. A producer receives a `MessageId` for each message it successfully sends.
-*   **Subscriptions and Cursors:** Each subscription on a topic maintains its own "cursor" which tracks consumption progress. It maintains a `mark-delete position`, indicating that all messages up to that point have been acknowledged.
-*   **Individual Acknowledgement:** Pulsar subscriptions support different acknowledgement modes. While cumulative acknowledgement is the default for some types, `Shared` and `Key_Shared` subscriptions heavily rely on **individual acknowledgement**. This allows a consumer to acknowledge a single message, even if it's out of order. When a message is acknowledged individually ahead of the `mark-delete position`, it creates an "acknowledgement hole". The broker's `ManagedCursor` must persistently track these holes to ensure that acknowledged messages are not redelivered after a broker restart.
-*   **Delayed Messages:** Pulsar supports scheduling messages for future delivery. This is a primary use case that benefits from this proposal, as it allows a scheduled message to be effectively "cancelled" before its delivery time.
-*   **Existing `skip-messages` API:** Pulsar has an admin API to `skip` a specified *number* of messages for a subscription. This is useful for bulk-skipping, but it lacks the precision to target a single, specific message within a large backlog, especially if its exact sequence is unknown.
+*   **Subscriptions and Cursors:** Each subscription on a topic maintains its own "cursor" which tracks consumption progress. For subscription types like `Shared` or `Key_Shared`, the cursor can track individually acknowledged messages, even if they are out of order relative to the main consumption progress marker (`mark-delete position`).
+*   **Individual Acknowledgement:** Pulsar subscriptions support different acknowledgement modes. `Shared` and `Key_Shared` subscriptions heavily rely on **individual acknowledgement**. This allows a consumer to acknowledge a single message. When a message is acknowledged individually ahead of the `mark-delete position`, the broker's `ManagedCursor` persistently tracks this to ensure that acknowledged messages are not redelivered after a broker or consumer restart.
+*   **Delayed Messages:** Pulsar supports scheduling messages for future delivery. A primary use case for this proposal is to allow a scheduled message to be effectively "cancelled" by acknowledging it before its delivery time.
+*   **Existing `skipMessages` API:** Pulsar has an admin API to `skip` a specified *number* of messages for a subscription. This is useful for bulk-skipping but lacks the precision to target a single, specific message within a large backlog, especially if its exact sequence is unknown. This proposal extends this existing API to add the ability to skip messages by their specific `MessageId`.
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


Denovo1998:
Currently, there are some issues with this implementation. 
(Note that this PIP only discusses the implementation of BucketDelayedDeliveryTracker)

If a delayed message in the LastMutableBucket has not been flushed to Bookie yet, and a failure occurs, the data in the LastMutableBucket will be lost. However, this will not have any impact, as after restarting, messages will still be read from the MackDelete position and the Bucket will be rebuilt.

However, our delayed message deletion command is different. If there is a LastMutableBucket that has not been successfully persisted (sealBucketAndAsyncPersistent), and the Broker crashes, this command will be lost. We cannot wait for the Seal Bucket condition to be triggered before returning the success of canceling the delayed message command, because we do not know how long it will take.

This part is important and needs to be resolved. I will think about how to solve this part, and everyone is also welcome to discuss with me.

I have some solutions, let me try testing them out first.

Denovo1998:
The document has been updated, we will implement the delayed message cancellation function through acknowledge message.

pip/pip-423.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+## Public-facing Changes
+The existing `skipMessages` API is modified to accept a POST body containing message IDs.
+
+### Public API
+The REST endpoint for skipping messages is updated. To skip by message ID, a client must send a `POST` request with `numMessages` set to `0` in the path and provide a map of message IDs in the JSON body.
+
+*   **Path (v2):** `POST /admin/v2/persistent/{tenant}/{namespace}/{topic}/subscription/{subName}/skip/{numMessages}`
+*   **Path (v1):** `POST /admin/v1/persistent/{property}/{cluster}/{namespace}/{topic}/subscription/{subName}/skip/{numMessages}`
```
codelipenghui: Actually we can add a new path e.g. skipByMessageIds to avoid the confusing 0 for numMessages is required here

Denovo1998: @codelipenghui Already modified, are there any other areas that need to be adjusted?

pip/pip-423.md
```diff
@@ -0,0 +1,254 @@
...
+### Public API
+
+A new REST endpoint is added for skipping specific messages.
+
+*   **Path (v2):** `POST /admin/v2/persistent/{tenant}/{namespace}/{topic}/subscription/{subName}/skipByMessageIds`
+*   **Path (v1):** `POST /admin/v1/persistent/{property}/{cluster}/{namespace}/{topic}/subscription/{subName}/skipByMessageIds`
+*   **Path Parameters:**
+    *   All path parameters identify the target subscription.
+*   **HTTP Body Parameters (JSON):** A map where keys are `ledgerId` as a string and values are `entryId` as a string.
+    *   **Example Body:** `{"12345": "100", "12346": "200"}`
```
lhotari: The serialization format should be changed. The problem with this format is that it cannot be extended to cover batch messages. 

besides ledgerId and entryId, there's also batchIndex, which is part of the message id.

One of the design principles has been to hide the implementation details of the message id in Pulsar and not leak it. That's why there are org.apache.pulsar.client.api.MessageId#toByteArray and org.apache.pulsar.client.api.MessageId#fromByteArray .

It would be useful to support multiple formats for the message id. For consistency with the MessageId.toByteArray() API, there should be support for accepting a list of base64 encoded byte array representations of the MessageId. 
One way to solve this is to have the body map to an object that is mapped with Jackson using polymorphism configured with `@JsonTypeInfo` and `@JsonSubTypes` annotations. That would allow flexibility for adding various formats for the message id list.

the default format could be the byteArrays encoded as base64:

```json
{"type": "byteArray", "messageIds": ["CLlgEAQwAA==", "CLlgEAYwAA==", "CLlgEJ4BMAA=", "CLlgEKQCMAA=", "CLlgEPQCMAA=", "CLlgECcwAA==", "CLlgEIQDMAA=", "CLlgEKEDMAA=", "CLlgEAQwAA==", "CLlgEI4EMAA="]}
```

since this could be made the default, the `"type": "byteArray"` could be omitted:
```json
{"messageIds": ["CLlgEAQwAA==", "CLlgEAYwAA==", "CLlgEJ4BMAA=", "CLlgEKQCMAA=", "CLlgEPQCMAA=", "CLlgECcwAA==", "CLlgEIQDMAA=", "CLlgEKEDMAA=", "CLlgEAQwAA==", "CLlgEI4EMAA="]}
```

another could be the ledgerId:entryId pairs (without support for batchIndex) if this is absolutely necessary:

```json
{"type": "map_of_ledgerId_entryId", "messageIds": {12345: 100, 12346: 200}}
```

Instead of that, it could be useful to have a format that maps directly:

```json
{"type": "messageId", "messageIds": [{"ledgerId": 12345, "entryId": 100}, {"ledgerId": 12346, "entryId": 200}]}
```

this format could also support the batch index.
```json
{"type": "messageId", "messageIds": [{"ledgerId": 12345, "entryId": 100, "batchIndex": 5}, {"ledgerId": 12346, "entryId": 200, "batchIndex": 10}]}
```

lhotari (changes requested):

pip/pip-423.md
```diff
@@ -0,0 +1,254 @@
...
+### CLI
+
+A new CLI command is added to `pulsar-admin topics`.
+
+*   **Command:** `skip-messages`
+*   **Description:** Skip some messages for a subscription by their message IDs.
+*   **Usage:** `pulsar-admin topics skip-messages <topic-name> [options]`
+*   **Options:**
+    *   `-s, --subscription <subName>` (required): The subscription to skip messages on.
+    *   `-m, --messageId <ledgerId=entryId>` (required): The message ID to skip. This option can be specified multiple times.
```
lhotari: it should be possible to also specify the batch index.
A better format is `ledgerId:entryId[:batchIndex]` where the separator is `:` and the batch index is optional.

pip/pip-423.md
```diff
@@ -0,0 +1,254 @@
...
+1.  **Subscription Interface Extension:**
+    The `Subscription` interface is extended with a new method to handle skipping by message IDs.
+
+```java
+// in pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Subscription.java
+public interface Subscription extends MessageExpirer {
+    // ... existing methods
+    CompletableFuture<Void> skipMessages(int numMessagesToSkip);
+
+    CompletableFuture<Void> skipMessages(Map<String, String> messageIds);
```
lhotari: should be `List<MessageIdAdv>` in this interface.

lhotari: I'm not actually sure if `List<MessageIdAdv>` is the best option here. I guess we'd need a new type (class) that captures Position and batchIndex. 
client's MessageId and MessageIdAdv are slightly different and contain more state that's not relevant here.

pip/pip-423.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+2.  **PersistentSubscription Implementation:**
+
+The `PersistentSubscription` class provides the concrete implementation. It converts the message ID map into `Position` objects and uses the standard individual acknowledge flow.
+
+```java
+// in pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
+@Override
+public CompletableFuture<Void> skipMessages(Map<String, String> messageIds) {
```
lhotari: should be List of message ids with (ledgerId, entryId and batchIndex information)

pip/pip-423.md
```diff
@@ -0,0 +1,254 @@
...
+            long entryId = Long.parseLong(entry.getValue());
+            Position position = PositionFactory.create(ledgerId, entryId);
+            positions.add(position);
+        } catch (Exception e) {
+            return CompletableFuture.failedFuture(new NotAllowedException("Invalid message ID."));
+        }
+    }
+
+    Map<String, Long> properties = Collections.emptyMap();
+    acknowledgeMessage(positions, AckType.Individual, properties);
```
lhotari: this logic should cover handling the case when the batchIndex is passed. If no batchIndex is passed, it should acknowledge the complete entry.

Denovo1998:
@lhotari @codelipenghui 
bump;

PR #24376, opened 2025-06-03
[fix][broker] fix unacked message count is zero when using exclusive subscription

### LINKED ISSUE #24159: [Bug] unacked message count is zero when using exclusive subscription.
### Search before asking

### Read release policy

### Version

Pulsar: master

### Minimal reproduce step

```java
@Test
    public void testUnackedMessages() throws Exception {
        String topicName = "testUnackedMessages";
        String subscriptionName = "my-sub";
        String serviceUrl = "pulsar://localhost:6650";

        @Cleanup PulsarClient pulsarClient = PulsarClient.builder().serviceUrl(serviceUrl).build();

        @Cleanup Consumer<byte[]> consumer =
                pulsarClient.newConsumer().topic(topicName).subscriptionName(subscriptionName)
                        .subscriptionType(SubscriptionType.Exclusive).messageListener((c, message) -> {
                            // noop
                        }).subscribe();

        @Cleanup Producer<byte[]> producer = pulsarClient.newProducer().topic(topicName).create();
        producer.send("1".getBytes());

        TopicStats stats = admin.topics().getStats(topicName);
        assertThat(stats.getSubscriptions().get(subscriptionName).getUnackedMessages()).isNotEqualTo(0);
    }
```

unackedMessages is not equal to 0.

### What did you see instead?

unackedMessages is equal to 0.

### Anything else?

_No response_

berg223:
Fixes #24159

Main Issue: #24159

### Motivation

The unacked message count is not correct when  using exclusive or failover subscription. We need to fix it.

### Modifications

1. `addAndGetUnAckedMsgs` for all subscription type in `Consumer`
2. `addAndGetUnAckedMsgs` for all subscription type  when handle ack.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [X] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


berg223:
I find it's not work when using Cumulative ACK in EXLUSIVE mode. Any feedback or advice is appreciated !

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -593,10 +593,9 @@ private CompletableFuture<Long> individualAckNormal(CommandAck ack, Map<String,
             } else {
                 position = PositionFactory.create(msgId.getLedgerId(), msgId.getEntryId());
                 ackedCount = getAckedCountForMsgIdNoAckSets(batchSize, position, ackOwnerConsumer);
-                if (checkCanRemovePendingAcksAndHandle(ackOwnerConsumer, position, msgId)) {
-                    addAndGetUnAckedMsgs(ackOwnerConsumer, -(int) ackedCount);
-                    updateBlockedConsumerOnUnackedMsgs(ackOwnerConsumer);
-                }
+                checkCanRemovePendingAcksAndHandle(ackOwnerConsumer, position, msgId);
```
nodece: You should check the return value, I'm concerned that the ack count is incorrect when using key_shared or shared.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -1315,7 +1315,9 @@ public CompletableFuture<SubscriptionStatsImpl> getStatsAsync(GetStatsOptions ge
         SubType subType = getType();
         subStats.type = getTypeString();
         if (dispatcher instanceof PersistentDispatcherSingleActiveConsumer) {
-            Consumer activeConsumer = ((PersistentDispatcherSingleActiveConsumer) dispatcher).getActiveConsumer();
+            PersistentDispatcherSingleActiveConsumer d = ((PersistentDispatcherSingleActiveConsumer) dispatcher);
+            subStats.unackedMessages = d.getUnackedMessages();
```
nodece, line 1319: https://github.com/apache/pulsar/pull/24376/files#diff-1d721da3aaff7689077d33bf3e55fd8d6a5c9184549857a6120d8050c5d67798R1274 has been handled `unackedMessages`. Could you check this logic?

nodece (commented):
Great work! I left my review result, please check.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
berg223: Yes. Good catch! I have rollback here and add a else branch to handle acks for Exlusive or Failover mode.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
berg223: I have checked the logic before and sure we can override the value here. Because PersistentDispatcherSingleActiveConsumer is for Exlusive or Failover mode, which has only one active consumer. In Exlusive or Failover mode, the unacked message of active consumer should be the result.

berg223:
> Great work! I left my review result, please check.

@nodece Thanks for kind review on it! I've been thinking about it for a while about Cumulative ACK.  Seems like it still need more work to fix unacked message count when using Cumulative ACK. Any feedback or advice is appreciated !

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -448,8 +448,8 @@ public Future<Void> sendMessages(final List<? extends Entry> entries,
     }
 
     private void incrementUnackedMessages(int unackedMessages) {
-        if (Subscription.isIndividualAckMode(subType)
-                && addAndGetUnAckedMsgs(this, unackedMessages) >= getMaxUnackedMessages()
+        if (addAndGetUnAckedMsgs(this, unackedMessages) >= getMaxUnackedMessages()
```
lhotari: Making state mutations outside of if conditions makes it easier to read and maintain the code in the future.
```suggestion
        int updatedUnackedMessages = addAndGetUnAckedMsgs(this, unackedMessages);
        if (updatedUnackedMessages >= getMaxUnackedMessages()
```

berg223:
I have another WIP https://github.com/berg223/pulsar/pull/1 to solve the issue further when the exlusive or failover consumer using the cumulative ack mode. I believe that it's complete now. Because there is another related flow control issue, I'm not sure whether to merge the WIP into this PR together. 

The basic idea is that we use the `pendingAcks` in Consumer to track unacked message for exlusive or failover subscription.However, `pendingAcks` is limited to individualAckMode before the WIP. I'm not sure whether the WIP break the origin design. So I have submitted a WIP there. 

If you think the WIP and this PR should be merged together. Please let me know and I will do it!

@lhotari @nodece

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
nodece: @berg223 It seems unnecessary to set the `unackedMessages` here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -596,6 +597,8 @@ private CompletableFuture<Long> individualAckNormal(CommandAck ack, Map<String,
                 if (checkCanRemovePendingAcksAndHandle(ackOwnerConsumer, position, msgId)) {
                     addAndGetUnAckedMsgs(ackOwnerConsumer, -(int) ackedCount);
                     updateBlockedConsumerOnUnackedMsgs(ackOwnerConsumer);
+                } else if (Subscription.isCumulativeAckMode(subType)) {
```
nodece: This method only handles the individual ack, so you can remove this check.

I think you should change `checkCanRemovePendingAcksAndHandle`:
```
    private boolean checkCanRemovePendingAcksAndHandle(PositionImpl position, MessageIdData msgId) {
        if (msgId.getAckSetsCount() == 0) {
            if (Subscription.isIndividualAckMode(subType)) {
                return removePendingAcks(position);
            }
            // Exclusive model
            return true;
        }
        return false;
    }
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumerTest.java
```diff
@@ -124,4 +130,47 @@ public void testSkipReadEntriesFromCloseCursor() throws Exception {
         // Verify: the topic can be deleted successfully.
         admin.topics().delete(topicName, false);
     }
+
+    @Test
+    public void testUnackedMessages() throws Exception {
```
nodece, line 137: Could you consider the batching model?

nodece:
@berg223 https://github.com/berg223/pulsar/pull/1 looks is an improvement, you can make a new PR to handle that after this PR is merged.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
berg223: It will be more clear if we set the value here. Otherwise, we will find it has sum `unackedMessages` for all consumers in fail over subscription.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
berg223: I got your concerns and  the else branch will be removed in WIP. It's not conformed to the name of the method if we return true for exclusive model. So may I leave a else branch here for now?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumerTest.java
berg223: Sure! I will add a batching model test. Thanks for kind review!

berg223:
> @berg223 [berg223#1](https://github.com/berg223/pulsar/pull/1) looks is an improvement, you can make a new PR to handle that after this PR is merged.

Thanks for the kind guide! I will make a new PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
nodece: > I got your concerns and the else branch will be removed in WIP.

You can cleanup in WIP.

> It's not conformed to the name of the method if we return true for exclusive model.  So may I leave a else branch here for now?

~I think we can do that.~

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumerTest.java
berg223: The test is changed to batching model for now!

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
nodece: @berg223 Apologies for the earlier incorrect review.

I'm concerned that if we don't improve `checkCanRemovePendingAcksAndHandle`, the unacknowledged message count may become inaccurate.

> It's not conformed to the name of the method if we return true for exclusive model.

That's a fair point—no worries.

> So may I leave an else branch here for now?

Please remove the `else` branch, as that code path is unreachable and could lead to confusion.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumerTest.java
```diff
@@ -124,4 +132,74 @@ public void testSkipReadEntriesFromCloseCursor() throws Exception {
...
+                    .negativeAckRedeliveryDelay(0, TimeUnit.SECONDS)
+                    .acknowledgmentGroupTime(0, TimeUnit.SECONDS)
+                    .maxAcknowledgmentGroupSize(1)
+                    .subscribe();
+
+            @Cleanup
+            Producer<String> producer = pulsarClient.newProducer(Schema.STRING).topic(topicName).create();
+
+            for (int i = 0; i < totalProducedMessage; ++i) {
+                producer.send(Integer.toString(i));
```
nodece, line 162: Batching mode is enabled by default, but calling `send` does not trigger the batching logic.
Please use `sendAsync` instead, as it properly handles the batching mechanism.

Also, make sure to account for batching conditions, such as message count, bytes, or batching delay.

berg223: Unfortunately, the batch mode is not supportted in this PR. Because we cannot [get the batch size](https://github.com/apache/pulsar/blob/cad9a84da3072010b2543e65c9a465f9b5ecd8a7/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java#L573) of the pending ack. However, this can be resolved in WIP.

berg223: I'm considering close this PR because it cann't support batching mode and cumulative ACK. On the other side, the changes has no side effect. What do you think about it? Do you think we should fix only the individual ack scenario for now?

berg223:
Since this PR only support individual ack in non–batching mode, I have submitted a new PR #24396 to further support features about:
1. batching mode
2. cumulative ack
3. transaction

Any feedback is appreciated @nodece

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumerTest.java
```diff
@@ -124,4 +132,74 @@ public void testSkipReadEntriesFromCloseCursor() throws Exception {
...
+
+            producer.send("1");
+            Message<String> msg = consumer.receive(5, TimeUnit.SECONDS);
+
+            consumer.negativeAcknowledge(msg);
+
+            // sleep for a while to wait for the negative ack to be processed
+            Thread.sleep(1000);
+            topicStats = admin.topics().getStats(topicName);
+            assertThat(topicStats.getSubscriptions().get(subscriptionName).getUnackedMessages()).isEqualTo(1);
```
nodece:
```suggestion
            assertUnackedMessage(topicName, subscriptionName, 1);
```

nodece (approved):

nodece:
I will help you fix the CI:

- Please remove `Subscription.isIndividualAckMode(subType)` check from `Consumer.shouldBlockConsumerOnUnackMsgs` to fix `AdminApi2Test#testConsumerStatsLastTimestamp`.

berg223:
> I will help you fix the CI:
> 
> * Please remove `Subscription.isIndividualAckMode(subType)` check from `Consumer.shouldBlockConsumerOnUnackMsgs` to fix `AdminApi2Test#testConsumerStatsLastTimestamp`.

Thanks for kind remind. But we cannot do this. Exclusive consumer shouldn't be blocked because it has no chance to be unblock when ack message. I have considered the block behavior carefully when coding and it's reasonable to keep it never be blocked the same as before. 

I have read the unit test in detail. And I think the  `assertEquals(lastConsumedFlowTimestamp, consumedFlowTimestamp);` should be changed to `assertTrue(ackedTimestampInSubStats < lastAckedTimestampInSubStats);`. A second flowPermit() will be triggered for Exclusive mode. The flow behavior is different with before and it's reasonable. Because after we acked message, the exlusive consumer will have more resource to consume messages. And we can give it a chance to `flowPermit()`.  It beahaves same to individual mode ack of shared consumer.

Sorry for the CI breaks.

codelipenghui (changes requested):
Cumulative acknowledgment is supported for both Inclusive and Failover subscription types, which is likely the main reason why unacknowledged message tracking hasn’t been supported previously.

Tracking unacked messages at the consumer level is quite difficult for Inclusive and Failover subscriptions, as the broker cannot determine how many individual messages were acknowledged in each cumulative ack.

While Shared subscriptions maintain pending acknowledgments with batch_size metadata per entry, I don’t think we should introduce similar tracking for Failover or Inclusive subscriptions.

In practice, users can already estimate unacknowledged messages using internal topic stats, since there’s only one active consumer for Failover and Inclusive subscriptions.

IMO, it's no a simple bug fix here. But we can move to the mail list   for the detailed discussion.

berg223:
IMO, since there will be a complete solution in PIP 426 #24400 , we can just close this PR if our user has no urgent need to fix it immediately. And we can leave comments in that PIP.  @codelipenghui  @nodece

berg223:
> Cumulative acknowledgment is supported for both Inclusive and Failover subscription types, which is likely the main reason why unacknowledged message tracking hasn’t been supported previously.

Sorry for that I'm confused by this. Why it's a reason? 

> Tracking unacked messages at the consumer level is quite difficult for Inclusive and Failover subscriptions, as the broker cannot determine how many individual messages were acknowledged in each cumulative ack.

A more complete implemention #24396 get the acknowledged number by `removeAllUpTo` method. Do you have concerns about that? Is there some cases haven't been covered?

> While Shared subscriptions maintain pending acknowledgments with batch_size metadata per entry, I don’t think we should introduce similar tracking for Failover or Inclusive subscriptions.

Do you have any worry about increasing memory usage here and that's the reason that you think it's negative to do that? Since consumer maybe acknowledge at any message,  we need to record  pending acknowledgments with batch_size metadata per entry.  Theoretically, it hasn't no difference with shared or key_shared subscription and it will has same impact on increasing memory. We can still add a feature flag here to limit it.

On the other hand, the flowcontrol of exlusive consumer is not work for now. And that functionality has a depend on accurate unacked message count. 

There are some issue like #15189 indicates that our failover consumer lack fine grained limit on pending ackownledge. We will have a better infrastructure to solve issues like that if we introduce similar tracking for Failover or EXclusive subscriptions.


> In practice, users can already estimate unacknowledged messages using internal topic stats, since there’s only one active consumer for Failover and Inclusive subscriptions.

Could you please share us the way to estimate unacknowledged messages using internal topic stats. Is that not a hack way and friendly to our users? Do you think we still need to make the count accurate?

> IMO, it's no a simple bug fix here. But we can move to the mail list for the detailed discussion.

Yes! We have a PIP 426 #24400 . We can discuss there!

@codelipenghui

codelipenghui:
@berg223 Thanks for sharing the proposal, let talk on the PIP and I will also close this PR.

PR #24381, opened 2025-06-05
[improve][ml]Release idle offloaded read handle only the ref count is 0

poorbarcode:
### Motivation

https://github.com/apache/pulsar/pull/19783 introduced a feature: release idle offloaded ledger handle to release the direct memory occupied. But it left a potential issue, which has been discussed [here](https://github.com/apache/pulsar/pull/19783#discussion_r2086201906), and it can happen this way

| time/thread | release idle ledger handle | `ledgerHandle.readAsync` |
| --- | --- | --- |
| 1 | | `touch()`, update `lastAccessTimestamp` |
| 2 | | append the reading request into `executor.queue` |
| 3 | | the `executor` is very busy, leading to delayed executions of `ledgerHandle.readAsync` |
| 4 | Check how long the ledger handle has been idled |
| 5 | release the ledger handle |
| 6 | | `ledgerHandle.readAsync` execute eventually, but the ledger handle has been closed |

Since users can modify the threshold of marking the ledger handle as `idle` by the configuration `managedLedgerInactiveOffloadedLedgerEvictionTimeSeconds`, it may increase the probability of the issue happening

### Modifications

- Added a ref count named `pendingRead` to record how many pending reading requests are in progress, Pulsar release ledger handle when `pendingRead` is `0,` and has idled for enough time
- Instead of updating "lastAccessTimestamp" when a reading is starting, update it when a reading task is finished

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24384, opened 2025-06-05
[fix][ml]Revert a behavior change of releasing idle offloaded ledger handle: only release idle BlobStoreBackedReadHandle

poorbarcode:
### Motivation

https://github.com/apache/pulsar/pull/24381 changed the behavior below:
- original: Pulsar only releases the offloaded ledger handle that is  typed `BlobStoreBackedReadHandle`
- after #24381, Pulsar will release the offloaded ledger handle immediately if its type is not `BlobStoreBackedReadHandle`, because the method `lastAccessTimestamp()` always returns `-1` if its type is not `BlobStoreBackedReadHandle`

### Modifications
- Revert the behavior change.
- Modify `lastAccessTimestamp` first, and modify `pendingReading` second, to avoid getting a new `pendingReading` and an older `lastAccessTimestamp`, which leads to an incorrect ledger handle invalidation.

### Potential issue

The following issue still exists. I will push a new PR to fix it in the future.

| time/thread | release idle ledger handle | `ledgerHandle.readAsync` |
| --- | --- | --- |
| 1 | Ledger handle has been idle for a long time |
| 2 | Start to release ledger handle |
| 3 | | a new reading request comes in |
| 4 | closed ledger handle |
| 5 | | get a ledger closed error |

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -2753,7 +2753,8 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
                     ReadHandle readHandle = ledger.join();
                     if (readHandle instanceof OffloadedLedgerHandle offloadedLedgerHandle) {
                         int pendingRead = offloadedLedgerHandle.getPendingRead();
-                        if (pendingRead == 0) {
+                        long lastAccessTimestamp = offloadedLedgerHandle.lastAccessTimestamp();
+                        if (lastAccessTimestamp >= 0 && pendingRead == 0) {
                             long delta = now - offloadedLedgerHandle.lastAccessTimestamp();
                             if (delta >= inactiveOffloadedLedgerEvictionTimeMs) {
                                 log.info("[{}] Offloaded ledger {} can be released ({} ms elapsed since last access)",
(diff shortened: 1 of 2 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2753,7 +2753,8 @@ synchronized List<Long> internalEvictOffloadedLedgers() {
                     ReadHandle readHandle = ledger.join();
                     if (readHandle instanceof OffloadedLedgerHandle offloadedLedgerHandle) {
```
lhotari, line 2754: `BlobStoreBackedReadHandleImpl` is the only production code class that currently implements `OffloadedLedgerHandle`.

lhotari (approved):
LGTM

lhotari:
> * original: Pulsar only releases the offloaded ledger handle that is  typed `BlobStoreBackedReadHandle`
> * after [[improve][ml]Release idle offloaded read handle only the ref count is 0 #24381](https://github.com/apache/pulsar/pull/24381), Pulsar will release the offloaded ledger handle immediately if its type is not `BlobStoreBackedReadHandle`, because the method `lastAccessTimestamp()` always returns `-1` if its type is not `BlobStoreBackedReadHandle`

the title "[fix][ml]Revert a behavior change of releasing idle offloaded ledger handle: only release idle BlobStoreBackedReadHandle " is slightly misleading since even before it would only release idle BlobStoreBackedReadHandleImpls since that's the only implementation of OffloadedLedgerHandle.
Revisiting the PR title to focus on an improvement to avoid a remaining race condition could be better.

Issue #22934, opened 2024-06-18
[improvement]Perform health checks on the instance passed in by serviceUrl


AuroraTwinkle:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Reference Links: https://github.com/apache/pulsar/discussions/22933

### Solution

 In the implementation of the `ServiceNameResolver` interface, we can perform periodic node health check and corresponding removal or recovery operations to avoid selecting unhealthy nodes when resolvingHost!

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

lhotari:
@AuroraTwinkle This issue doesn't describe any problem, only a solution. Please add more context about what solution you are trying to solve.

lhotari:
> In the implementation of the `ServiceNameResolver` interface, we can perform periodic node health check and corresponding removal or recovery operations to avoid selecting unhealthy nodes when resolvingHost!

@AuroraTwinkle What problem occurs if an unhealthy node is selected? I would assume that in the current design, the next node would be selected after a connection attempt. Why would there be a need to change this?

lhotari:
@AuroraTwinkle I see the referenced discussion https://github.com/apache/pulsar/discussions/22933. That's a good problem description. 
I think it's a bug when the next available serviceUrl isn't attempted. That's the whole point of having multiple endpoints. However the solution in #22935 has it's issues as I've commented on the PR. The general approach in the client should be that the connection is attempted and if it fails, it will pick the next endpoint.

I took a look at the code and it seems that 
`client.getCnxPool().getConnection(socketAddress)` calls should be using `client.getCnxPool().getConnection(serviceNameResolver)`.

`org.apache.pulsar.client.impl.ConnectionPool` would need a new `getConnection` method that accepts the `org.apache.pulsar.client.impl.ServiceNameResolver` instance. The ServiceNameResolver interface will need some changes. When there's a failure for one address, there should be a method that can be called by ConnectionPool to provide feedback that the connection failed for that address. This way the ServiceNameResolver can remove it from the main pool of addresses for a configurable period of time. If all main pool addresses are offline, the resolver implementation could start getting random addresses from the offline pool. There would also be a need to have a success feedback method so that ConnectionPool could call it each time a connection was successful. 
Would you like to implement such solution instead of #22935?

AuroraTwinkle:
[...]

PR #22935, opened 2024-06-18
[improve][client]:Perform health checks on the endpoints passed in by serviceUrl when building PulsarClient

AuroraTwinkle:
Main Issue: https://github.com/apache/pulsar/issues/22934

### Motivation

Refer to issue: https://github.com/apache/pulsar/issues/22934

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24387, opened 2025-06-05
[improve][client]PIP-425:Support connecting with next available endpoint for multi-endpoint serviceUrls

AuroraTwinkle:
Fixes https://github.com/apache/pulsar/issues/22934#issuecomment-2943690398

Main Issue: https://github.com/apache/pulsar/issues/22934#issuecomment-2943690398

PIP: https://github.com/apache/pulsar/pull/24394

### Motivation
As https://github.com/apache/pulsar/issues/22934 and https://github.com/apache/pulsar/discussions/22933 mentioned, when most of the nodes in serviceurl are down (but there is at least one available node), creating consumers and producers through PulsarClient will most likely fail. I think this is not as expected. If the code is robust enough, as long as there is one available node, it should be accessible normally. Therefore, this PR is going to optimize the code logic, **remove unavailable nodes through the feedback mechanism**, and improve the success rate of PulsarClient requests.

By the way, https://github.com/apache/pulsar/pull/22935 removes faulty nodes through a regular health check mechanism, but this brings new problems (frequent creation of connections and increased system load), so this solution is abandoned. See https://github.com/apache/pulsar/issues/22934#issuecomment-2943690398 for more details!

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: 
https://github.com/AuroraTwinkle/pulsar/pull/8

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
@@ -20,7 +20,6 @@ package org.apache.pulsar.client.impl;
 
 import static com.google.common.base.Preconditions.checkState;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableSet;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -135,7 +134,7 @@ public class PulsarServiceNameResolver implements ServiceNameResolver {
             }
         }
         this.allAddressList = addresses;
-        this.allAddressSet = ImmutableSet.copyOf(addresses);
+        this.allAddressSet = Set.copyOf(addresses);
         this.serviceUrl = serviceUrl;
         this.serviceUri = uri;
         this.currentIndex = randomIndex(addresses.size());


AuroraTwinkle:
The CI tests have passed in the fork repository https://github.com/AuroraTwinkle/pulsar/pull/8

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -40,11 +50,37 @@ public class PulsarServiceNameResolver implements ServiceNameResolver {
...
+    private volatile List<InetSocketAddress> availableAddressList;
+    private final Map<InetSocketAddress, EndpointStatus> hostAvailabilityMap = new HashMap<>();
+    private final long serviceUrlQuarantineInitDurationMs;
+    private final long serviceUrlQuarantineMaxDurationMs;
+    private final boolean enableServiceUrlQuarantine;
+
+    public PulsarServiceNameResolver() {
+        this.serviceUrlQuarantineInitDurationMs = 0;
+        this.serviceUrlQuarantineMaxDurationMs = 0;
+        this.enableServiceUrlQuarantine = false;
```
nodece:
```suggestion
        this(0, 0);
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +132,129 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
-        this.addressList = addresses;
+        this.allAddressList = addresses;
+        this.allAddressSet = new HashSet<>(addresses);
         this.serviceUrl = serviceUrl;
         this.serviceUri = uri;
         this.currentIndex = randomIndex(addresses.size());
+        if (enableServiceUrlQuarantine) {
+            this.availableAddressList = new ArrayList<>(addresses);
+            hostAvailabilityMap.keySet().removeIf(host -> !allAddressSet.contains(host));
+            addresses.forEach(address -> hostAvailabilityMap.putIfAbsent(address, createEndpointStatus(true, address)));
```
nodece: It looks like you should use `allAddressSet` instead of `address`.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +132,129 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+    private boolean computeEndpointStatus(boolean newIsAvailable, EndpointStatus status) {
+        if (!newIsAvailable) {
+            if (!status.isAvailable()) {
+                // from unavailable to unavailable, check if we need to try to recover
+                long elapsedTimeMsSinceLast = System.currentTimeMillis() - status.getLastUpdateTimeStampMs();
+                boolean needTryRecover = elapsedTimeMsSinceLast >= status.getNextDelayMsToRecover();
+                if (needTryRecover) {
+                    log.info("service name resolver try to recover host {} after {}", status.getSocketAddress(),
+                            Duration.ofMillis(elapsedTimeMsSinceLast));
+                    status.setAvailable(true);
```
nodece, line 257: `false`?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +132,129 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+                    availableHostsChanged.set(changed);
+                }
+            }
+        });
+
+        if (availableHostsChanged.get()) {
+            availableAddressList = hostAvailabilityMap.entrySet()
+                    .stream()
+                    .filter(entry -> entry.getValue().isAvailable())
+                    .filter(entry -> allAddressSet.contains(entry.getKey()))
```
nodece:
```suggestion
                    .filter(entry -> entry.getValue().isAvailable() && allAddressSet.contains(entry.getKey()))
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +132,129 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+        Backoff backoff = new BackoffBuilder()
+                .setInitialTime(serviceUrlQuarantineInitDurationMs, TimeUnit.MILLISECONDS)
+                .setMax(serviceUrlQuarantineMaxDurationMs, TimeUnit.MILLISECONDS)
+                .create();
+        EndpointStatus endpointStatus =
+                new EndpointStatus(inetSocketAddress, backoff, System.currentTimeMillis(), 0,
+                        isAvailable);
+        if (!isAvailable) {
+            endpointStatus.setNextDelayMsToRecover(endpointStatus.getQuarantineBackoff().next());
+        }
```
nodece: Perhaps you can quickly compute this status by `computeEndpointStatus`.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +132,129 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+            }
+            computeEndpointStatus(isAvailable, oldStatus);
+            return oldStatus;
+        });
+
+        hostAvailabilityMap.forEach((__, endpointStatus) -> {
+            if (!endpointStatus.isAvailable()) {
+                boolean changed = computeEndpointStatus(false, endpointStatus);
+                if (!availableHostsChanged.get()) {
+                    availableHostsChanged.set(changed);
```
nodece:
```suggestion
                    availableHostsChanged.set(true);
```

nodece: If applied, please make computeEndpointStatus return void.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -40,11 +50,37 @@ public class PulsarServiceNameResolver implements ServiceNameResolver {
     private static final AtomicIntegerFieldUpdater<PulsarServiceNameResolver> CURRENT_INDEX_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(PulsarServiceNameResolver.class, "currentIndex");
     private volatile int currentIndex;
-    private volatile List<InetSocketAddress> addressList;
+    private volatile List<InetSocketAddress> allAddressList;
+    private volatile Set<InetSocketAddress> allAddressSet;
+    private volatile List<InetSocketAddress> availableAddressList;
+    private final Map<InetSocketAddress, EndpointStatus> hostAvailabilityMap = new HashMap<>();
```
nodece: Please use a thread-safe structure.
```suggestion
    private final Map<InetSocketAddress, EndpointStatus> hostAvailabilityMap = new ConcurrentHashMap<>();
```

nodece: Set and List also have the same issue.

nodece (commented):
LGTM, I've completed a partial review of the code.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: This means that the quarantine time has expired and the address needs to be set to an available state, so it is true.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: ok

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
nodece: Ok, got. You want to recover this address from unavailable to available.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +130,125 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+            if (!status.isAvailable()) {
+                // from unavailable to unavailable, check if we need to try to recover
+                long elapsedTimeMsSinceLast = System.currentTimeMillis() - status.getLastUpdateTimeStampMs();
+                boolean needTryRecover = elapsedTimeMsSinceLast >= status.getNextDelayMsToRecover();
+                if (needTryRecover) {
+                    log.info("service name resolver try to recover host {} after {}", status.getSocketAddress(),
+                            Duration.ofMillis(elapsedTimeMsSinceLast));
+                    status.setAvailable(true);
+                    status.setLastUpdateTimeStampMs(System.currentTimeMillis());
+                    status.setNextDelayMsToRecover(status.getQuarantineBackoff().next());
```
nodece, line 259:
```suggestion
                    status.setNextDelayMsToRecover(0);
                    status.getQuarantineBackoff().reset();
```
If set to `true`, you should reset the delay and backoff.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +130,125 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+                new EndpointStatus(inetSocketAddress, backoff, System.currentTimeMillis(), 0,
+                        isAvailable);
+        if (!isAvailable) {
+            computeEndpointStatus(false, endpointStatus);
+        }
+        return endpointStatus;
+    }
+
+    /**
+     * Compute the endpoint status based on the new availability status.
```
nodece: I suggest you should update this javadoc, because this method can recover the availability status when `false`:

```
/**
 * Updates the endpoint's availability status based on the given input flag and internal timing logic.
 *
 * <p>This method applies the input flag directly, and includes a time-based self-healing mechanism: if the 
 * endpoint has been marked unavailable for long enough, it may automatically transition back to {@code available} 
 * even when {@code newIsAvailable} is {@code false}.
 *
 * <p>This dual behavior—accepting input and triggering recovery based on cooldown time—can be non-obvious.
 * Callers should be aware that the availability status may change independently of the input flag.
 */
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: Here is the recovery action performed when the quarantine time expired. If the next call still fails, the quarantine time of the address needs to continue to increase exponentially. For example, suppose host1 fails to call for the first time, available is set to false, and the initial isolation time is 1 minute. After one minute, the quarantine time expires and the address is recovery. If the first call after recovery fails again, the quarantine time should increase exponentially to 2 minutes, and so on.
**Only when the first call after the address is recovery is successful, `quarantineBackoff` will be reset to the initial state.**

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: ok, good suggesstions!

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
nodece: I have understood this design.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +130,134 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+     * <p>This method applies the input flag directly, and includes a time-based self-healing mechanism: if the
+     * endpoint has been marked unavailable for long enough, it may automatically transition back to {@code available}
+     * even when {@code newIsAvailable} is {@code false}.
+     *
+     * <p>This dual behavior—accepting input and triggering recovery based on cooldown time—can be non-obvious.
+     * Callers should be aware that the availability status may change independently of the input flag.
+     *
+     * @param newIsAvailable the new availability status of the endpoint
+     * @param status         the current status of the endpoint
+     */
```
nodece:
```suggestion
    /**
     * Updates the endpoint's availability status based on the given input flag and internal quarantine logic.
     *
     * <p>This method applies the input flag directly, and includes a time-based self-healing mechanism: if the
     * endpoint has been marked unavailable for a sufficient cooldown period (quarantine), it automatically transitions back to available even when {@code newIsAvailable} is {@code false}.
     *
     * <p>This allows the system to retry endpoints that were previously marked as unavailable.
     * If the endpoint fails again after recovery, it is marked unavailable and re-enters quarantine
     * with an exponentially increased delay before the next recovery attempt.
     *
     * <p>The backoff is only reset when a successful update occurs after recovery
     * (i.e., when {@code newIsAvailable} is {@code true} and the endpoint was previously marked unavailable).
     *
     * @param newIsAvailable the new availability status of the endpoint
     * @param status         the current status of the endpoint
     */
```

nodece: I want to add the working principle to this Javadoc for clarity. Please feel free to refine or enhance the wording as you see fit.

AuroraTwinkle: Very detailed comments, great!

pulsar-client/src/test/java/org/apache/pulsar/client/impl/PulsarServiceNameResolverTest.java
```diff
@@ -129,4 +130,54 @@ public void testMultipleHostsTlsUrl() throws Exception {
             assertTrue(expectedHostUrls.contains(resolver.resolveHostUri()));
         }
     }
+
+    @Test
+    public void testRemoveUnavailableHost() throws InvalidServiceURL {
+        String serviceUrl = "pulsar+ssl://host1:6651,host2:6651,host3:6651";
+        resolver.updateServiceUrl(serviceUrl);
+        assertEquals(serviceUrl, resolver.getServiceUrl());
+        assertEquals(ServiceURI.create(serviceUrl), resolver.getServiceUri());
```
nodece: IDEA: Arguments to 'assertEquals()' in wrong order 

```suggestion
        assertEquals(resolver.getServiceUrl(), serviceUrl);
        assertEquals(resolver.getServiceUri(), ServiceURI.create(serviceUrl));
```

pulsar-client/src/test/java/org/apache/pulsar/client/impl/PulsarServiceNameResolverTest.java
```diff
@@ -129,4 +130,54 @@ public void testMultipleHostsTlsUrl() throws Exception {
...
+        expectedHostUrls.add(URI.create("pulsar+ssl://host3:6651"));
+        Set<InetSocketAddress> allOriginAddresses = new HashSet<>(expectedAddresses);
+        allOriginAddresses.add(InetSocketAddress.createUnresolved("host1", 6651));
+
+        // Mark host1 as unavailable
+        resolver.markHostAvailability(InetSocketAddress.createUnresolved("host1", 6651), false);
+        // Now host1 should be removed from the available hosts
+        for (int i = 0; i < 10; i++) {
+            assertTrue(expectedAddresses.contains(resolver.resolveHost()));
+            assertTrue(expectedHostUrls.contains(resolver.resolveHostUri()));
```
nodece: Please check `host1`.
```suggestion
            InetSocketAddress inetSocketAddress = resolver.resolveHost();
            assertNotEquals(inetSocketAddress.getHostName(), "host1");
            assertTrue(expectedAddresses.contains(inetSocketAddress));

            URI uri = resolver.resolveHostUri();
            assertNotEquals(uri.getHost(), "host1");
            assertTrue(expectedHostUrls.contains(uri));
```

pulsar-client/src/test/java/org/apache/pulsar/client/impl/PulsarServiceNameResolverTest.java
```diff
@@ -129,4 +130,54 @@ public void testMultipleHostsTlsUrl() throws Exception {
...
+        }
+        assertEquals(resolverAddresses, allOriginAddresses);
+
+        resolverAddresses.clear();
+        // Mark all hosts as unavailable
+        resolver.markHostAvailability(InetSocketAddress.createUnresolved("host1", 6651), false);
+        resolver.markHostAvailability(InetSocketAddress.createUnresolved("host2", 6651), false);
+        resolver.markHostAvailability(InetSocketAddress.createUnresolved("host3", 6651), false);
+
+        // After marking all hosts as unavailable, resolver should fall back to select from all origin host
```
nodece, line 182:
```suggestion
        // After marking all hosts as unavailable, resolver should fall back to select from all origin host
        assertTrue(resolver.getAvailableAddressList().isEmpty());
```

You need to add the following code to the `PulsarServiceNameResolver`:
```
import com.google.common.annotations.VisibleForTesting;

public class PulsarServiceNameResolver implements ServiceNameResolver {
    @VisibleForTesting
    List<InetSocketAddress> getAvailableAddressList() {
        return availableAddressList;
    }
}
```

Please make sure `availableAddressList` is empty.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
```diff
@@ -0,0 +1,337 @@
...
+                    .subscribeAsync()
+                    .thenAccept(Consumer::closeAsync);
+        }
+        Uninterruptibles.sleepUninterruptibly(10, java.util.concurrent.TimeUnit.SECONDS);
+        // check if the unhealthy address is removed
+        Set<InetSocketAddress> expectedHealthyAddresses = new HashSet<>();
+        expectedHealthyAddresses.add(healthyAddress);
+
+        Set<InetSocketAddress> resolvedAddresses = new HashSet<>();
+        for (int i = 0; i < 100; i++) {
```
nodece:
```suggestion
        for (int i = 0; i < hosts.length; i++) {
```
?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
```diff
@@ -0,0 +1,337 @@
...
+        String topic = "persistent://my-property/my-ns/topic" + UUID.randomUUID();
+        admin.topics().createNonPartitionedTopic(topic);
+        // Create consumers to trigger unhealthy address removal
+        for (int i = 0; i < originAllAddresses.size(); i++) {
+            String subName = "my-sub" + UUID.randomUUID();
+            pulsarClient.newConsumer()
+                    .subscriptionMode(SubscriptionMode.Durable)
+                    .topic(topic).receiverQueueSize(1).subscriptionName(subName)
+                    .subscribeAsync()
+                    .thenAccept(Consumer::closeAsync);
```
nodece: Please use sync call instead of async call.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
```diff
@@ -0,0 +1,337 @@
...
+
+    @DataProvider(name = "ackReceiptEnabledAndSubscriptionTypes")
+    public Object[][] ackReceiptEnabledAndSubscriptionTypes() {
+        return new Object[][]{
+                {true, SubscriptionType.Shared},
+                {true, SubscriptionType.Key_Shared},
+                {false, SubscriptionType.Shared},
+                {false, SubscriptionType.Key_Shared},
+        };
+    }
```
nodece:
```suggestion
```

Useless code.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
```diff
@@ -0,0 +1,337 @@
...
+                successCount++;
+            } catch (Exception e) {
+                log.warn("Failed to create consumer and producer {} for topic {}: {}", subName, topic, e.getMessage());
+            }
+        }
+        return successCount;
+    }
+
+    @Test
+    public void testServiceUrlHealthCheck() throws Exception {
```
nodece: This test always fails, please check.

nodece (changes requested):

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
```diff
@@ -0,0 +1,337 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+@Test(groups = "broker-api")
+public class CreateConsumerProducerTest extends ProducerConsumerBase {
```
nodece: ServiceUrlQuarantineTest?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
AuroraTwinkle: fixed

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
AuroraTwinkle: The CI tests have passed at https://github.com/AuroraTwinkle/pulsar/pull/8

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/CreateConsumerProducerTest.java
AuroraTwinkle: fixed

pulsar-client/src/test/java/org/apache/pulsar/client/impl/PulsarServiceNameResolverTest.java
AuroraTwinkle: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +131,143 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
                 throw new InvalidServiceURL(e);
             }
         }
-        this.addressList = addresses;
+        this.allAddressList = addresses;
+        this.allAddressSet = new HashSet<>(addresses);
```
nodece: I think you should use `CopyOnWriteArraySet` instead of `HashSet` to ensure thread safety, because the `markHostAvailability` also accesses this data.

AuroraTwinkle: How about using `ImmutableSet`? It should be immutable.

nodece (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -40,11 +51,35 @@ public class PulsarServiceNameResolver implements ServiceNameResolver {
...
     @Override
     public InetSocketAddress resolveHost() {
-        List<InetSocketAddress> list = addressList;
+        final List<InetSocketAddress> list;
+        List<InetSocketAddress> availableAddresses = availableAddressList;
+        if (availableAddresses != null && !availableAddresses.isEmpty()) {
+            list = availableAddresses;
+        } else {
+            // if no available address, use the original address list
+            list = allAddressList;
```
liangyepianzhou, line 80: Maybe a warning log is required here.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +131,143 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
             }
         }
-        this.addressList = addresses;
+        this.allAddressList = addresses;
+        this.allAddressSet = ImmutableSet.copyOf(addresses);
         this.serviceUrl = serviceUrl;
         this.serviceUri = uri;
         this.currentIndex = randomIndex(addresses.size());
+        if (enableServiceUrlQuarantine) {
+            this.availableAddressList = new ArrayList<>(addresses);
```
315157973: Should `availableAddressList  = hostAvailabilityMap.keySet()` ?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +131,143 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+            hostAvailabilityMap.keySet().removeIf(host -> !allAddressSet.contains(host));
```
315157973: `hostAvailabilityMap.keySet().retainAll(allAddressSet);`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
```diff
@@ -96,15 +131,143 @@ public void updateServiceUrl(String serviceUrl) throws InvalidServiceURL {
...
+            if (!endpointStatus.isAvailable()) {
+                computeEndpointStatus(false, endpointStatus);
+                if (!availableHostsChanged.get() && endpointStatus.isAvailable()) {
+                    availableHostsChanged.set(true);
+                }
+            }
+        });
+
+        if (availableHostsChanged.get()) {
+            availableAddressList = hostAvailabilityMap.entrySet()
```
315157973, line 196: Can immutable collections replace synchronized?

```
availableAddressList = hostAvailabilityMap.entrySet().stream()
        .filter(e -> e.getValue().isAvailable())
        .map(Map.Entry::getKey)
        .collect(Collectors.collectingAndThen(
            Collectors.toList(),
            Collections::unmodifiableList
        ));
```

AuroraTwinkle: The main purpose of `synchronized` is to ensure the thread safety of `EndpointStatus` calculation, using only immutable collections will not achieve the goal. How do you think about it?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java
AuroraTwinkle: fixed

PR #24394, opened 2025-06-07
[improve][pip] PIP-425: fix problem that consumer or producer create failed when many nodes  in PulsarClient serviceUrl become unavailable

AuroraTwinkle:
…failed when many nodes  in PulsarClient serviceUrl become unavailable

Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-425.md
+++ b/pip/pip-425.md
@@ -1,0 +1,127 @@
+# PIP-425: fix problem that consumer or producer create failed when many nodes in PulsarClient serviceUrl become unavailable
+Implementation PR: [#24387](https://github.com/apache/pulsar/pull/24387)
+
+# Background knowledge
+In the current Pulsar client versions, consumer or producer creation is the responsibility of PulsarClient, we can build a PulsarClient by pass serviceUrl param:
+
+```java
+public class Example {
+    public static void main(String[] args) throws PulsarClientException {
+        PulsarClient pulsarClient = PulsarClient.builder()
+                .serviceUrl("pulsar://localhost1:6650,localhost2:6650,localhost3:6650")
+                .build();
+        Consumer<Long> consumer = pulsarClient.newConsumer(Schema.INT64)
+                .topic("persistent://public/default/my-topic")
+                .subscriptionName("my-subscription")
+                .subscribe();
+
+        Producer<Long> producer = pulsarClient.newProducer(Schema.INT64)
+                .topic("persistent://public/default/my-topic")
+                .create();   
+
+        .....
+
+    }
+}
+
+```
+
+# Motivation
+
+As [#22934 ](https://github.com/apache/pulsar/issues/22934) and [#22933](https://github.com/apache/pulsar/discussions/22933) mentioned, when most of the nodes in serviceurl are down (but there is at least one available node), creating consumers and producers through PulsarClient will most likely fail. I think this is not as expected. If the code is robust enough, as long as there is one available node, it should be accessible normally. Therefore, this pip is going to optimize the code logic, remove unavailable nodes through the feedback mechanism, and improve the success rate of PulsarClient requests.
+
+By the way, [#22935](https://github.com/apache/pulsar/pull/22935) removes faulty nodes through a regular health check mechanism, but this brings new problems (frequent creation of connections and increased system load), so this solution is abandoned. See #22934 (comment) for more details!
+
+
+
+# Goals
+Optimize the PulsarClient code logic, remove unavailable serviceUrl nodes through the feedback mechanism, and improve the success rate of PulsarClient requests.
+
+## In Scope
+If this pip is accepted, it will help improve the PulsarClient success rate of consumer or producer creation, even if most of the nodes in serviceurl are no longer available.
+
+# Detailed Design
+**Explanation of nouns:**
+- `availableAddressPool`: a pool containing all available addresses, subset of serviceUrl.
+- `originAddressPool`: a pool containing all addresses in serviceUrl.
+
+## Design & Implementation Details
+As [#22934](https://github.com/apache/pulsar/issues/22934#issuecomment-2943690398) mentioned, add `markHostAvailability` method in `ServiceNameResolver` interface like:
+```java
+    /**
+     * Mark the availability of a host.
+     * @param address the host address to mark availability for
+     * @param isAvailable true if the host is available, false otherwise
+     */
+    default void markHostAvailability(InetSocketAddress address, boolean isAvailable){
+        // Default implementation does nothing
+        // Subclass can override this method to implement host availability tracking
+    }
+```
+When resolve one address in serviceUrl, the `ConnectionPool` will invoke `markHostAvailability` to tell availability about that address, so the `ServiceNameResolver` can remove it from `availableAddressPool` if unavailable (or recover back to `availableAddressPool` if available). If all addressed in  `availableAddressPool` are unavailable, the resolver implementation could start getting random addresses from the `originAddressPool`.
+
+For addresses removed from `availableAddressPool`, will retry according to the exponential avoidance interval strategy, avoid the problem that the failed address cannot be resolved again after recovery. The retry interval can be customized via new optional configurations `serviceUrlRecoveryInitBackoffInterval` and `serviceUrlRecoveryMaxBackoffInterval` in `ClientBuilder`
+
+
+
+### Usage Example
+```java
+    private void example() throws PulsarClientException {
+        // The failed node will be removed and retried after a period of time, with an initial interval of 30 seconds increasing exponentially. After reaching the maximum interval of 1 hour, the retry will be maintained once an hour until a certain retry succeeds, and the retry interval is reset to 0.
+        PulsarClient pulsarClient = PulsarClient.builder()
+                .serviceUrl("pulsar://localhost1:6650,localhost2:6650,localhost3:6650")
+                .serviceUrlRecoveryInitBackoffInterval(30000, TimeUnit.MILLISECONDS)
+                .serviceUrlRecoveryMaxBackoffInterval(1, TimeUnit.HOURS)
+                .build();    
+
+        // disable removing unavailable serviceUrl nodes by pass 0 to serviceUrlRecoveryInitBackoffInterval and serviceUrlRecoveryMaxBackoffInterval
+        PulsarClient pulsarClientDisableReovery = PulsarClient.builder()
+                .serviceUrl("pulsar://localhost1:6650,localhost2:6650,localhost3:6650")
+                .serviceUrlRecoveryInitBackoffInterval(0, TimeUnit.MILLISECONDS)
+                .serviceUrlRecoveryMaxBackoffInterval(0, TimeUnit.MILLISECONDS)
+                .build();                
+}
+
+```
+## Public-facing Changes
+
+### Public API
+
+1. Add two optional configs `serviceUrlRecoveryInitBackoffInterval` and `serviceUrlRecoveryMaxBackoffInterval` in `ClientBuilder`
+```java
+    /**
+     * Configure the service URL recovery init backoff intervals.
+     *
+     * <p>When the client is unable to connect to the service URL, it will wait for a certain amount of time before
+     * trying to recover the service URL. The init backoff intervals can be configured using this method.
+     *
+     * <p>
+     * A value of 0 means don't need wait before retrying to connect to the failed service URL.
+     * @param serviceUrlRecoveryInitBackoffInterval the initial backoff interval for service URL recovery
+     * @param unit the time unit for the backoff interval
+     * @return the client builder instance
+     */
+    ClientBuilder serviceUrlRecoveryInitBackoffInterval(long serviceUrlRecoveryInitBackoffInterval, TimeUnit unit);
+
+    /**
+     * Configure the service URL recovery max backoff interval.
+     *
+     * <p>When the client is unable to connect to the service URL, it will wait for a certain amount of time before
+     * trying to recover the service URL. The max backoff interval can be configured using this method.
+     *
+     * <p>
+     * A value of 0 means don't need wait before retrying to connect to the failed service URL.
+     * @param serviceUrlRecoveryMaxBackoffInterval the maximum backoff interval for service URL recovery
+     * @param unit the time unit for the backoff interval
+     * @return the client builder instance
+     */
+    ClientBuilder serviceUrlRecoveryMaxBackoffInterval(long serviceUrlRecoveryMaxBackoffInterval, TimeUnit unit);
+```
+
+# Backward & Forward Compatibility
+You can do upgrading or reverting normally, no specified steps are needed to do.
+
+# Links
+
+* Mailing List discussion thread: 
+* Mailing List voting thread: 


pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
+# PIP-425: fix problem that consumer or producer create failed when many nodes in PulsarClient serviceUrl become unavailable
```
lhotari: the title of the PIP could be simpler than "fix problem that consumer or producer create failed when many nodes in PulsarClient serviceUrl become unavailable".
Something like "Support connecting with next available endpoint for multi-endpoint serviceUrls" etc. would be a better description of the feature. Iterating on the title with the help of LLM could be useful.

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+Implementation PR: [#24387](https://github.com/apache/pulsar/pull/24387)
+
+# Background knowledge
+In the current Pulsar client versions, consumer or producer creation is the responsibility of PulsarClient, we can build a PulsarClient by pass serviceUrl param:
```
lhotari: Explaining that the serviceUrl supports multiple endpoints with the syntax `pulsar://host1:port1,host2:port2,...` or `pulsar+ssl://host1:port1,host2:port2,...` would be useful context.

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+```
+
+# Motivation
+
+As [#22934 ](https://github.com/apache/pulsar/issues/22934) and [#22933](https://github.com/apache/pulsar/discussions/22933) mentioned, when most of the nodes in serviceurl are down (but there is at least one available node), creating consumers and producers through PulsarClient will most likely fail. I think this is not as expected. If the code is robust enough, as long as there is one available node, it should be accessible normally. Therefore, this pip is going to optimize the code logic, remove unavailable nodes through the feedback mechanism, and improve the success rate of PulsarClient requests.
+
+By the way, [#22935](https://github.com/apache/pulsar/pull/22935) removes faulty nodes through a regular health check mechanism, but this brings new problems (frequent creation of connections and increased system load), so this solution is abandoned. See #22934 (comment) for more details!
```
lhotari: Instead of referring multiple external references, explain the problem in the current multi-endpoint support: when an endpoint chosen to be used for the connection is unavailable, creating consumers and producers will fail. It would be expected that the next available endpoint is attempted for connecting when there's a failure.

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+# Goals
+Optimize the PulsarClient code logic, remove unavailable serviceUrl nodes through the feedback mechanism, and improve the success rate of PulsarClient requests.
+
+## In Scope
+If this pip is accepted, it will help improve the PulsarClient success rate of consumer or producer creation, even if most of the nodes in serviceurl are no longer available.
+
+# Detailed Design
+**Explanation of nouns:**
+- `availableAddressPool`: a pool containing all available addresses, subset of serviceUrl.
+- `originAddressPool`: a pool containing all addresses in serviceUrl.
```
lhotari: it would be more consistent to call this `allAddressPool`

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Public API
+
+1. Add two optional configs `serviceUrlRecoveryInitBackoffInterval` and `serviceUrlRecoveryMaxBackoffInterval` in `ClientBuilder`
+```java
+    /**
+     * Configure the service URL recovery init backoff intervals.
+     *
+     * <p>When the client is unable to connect to the service URL, it will wait for a certain amount of time before
+     * trying to recover the service URL. The init backoff intervals can be configured using this method.
```
lhotari: this description isn't accurate and is slightly misleading. I guess the problem is calling this "backoff". 
"Backoff" is usually about retries. 

This is really about quarantining instead of backoff and retries. I'd suggest renaming "backoff interval" with "quarantine duration".

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+     */
+    default void markHostAvailability(InetSocketAddress address, boolean isAvailable){
+        // Default implementation does nothing
+        // Subclass can override this method to implement host availability tracking
+    }
+```
+When resolve one address in serviceUrl, the `ConnectionPool` will invoke `markHostAvailability` to tell availability about that address, so the `ServiceNameResolver` can remove it from `availableAddressPool` if unavailable (or recover back to `availableAddressPool` if available). If all addressed in  `availableAddressPool` are unavailable, the resolver implementation could start getting random addresses from the `originAddressPool`.
+
+For addresses removed from `availableAddressPool`, will retry according to the exponential avoidance interval strategy, avoid the problem that the failed address cannot be resolved again after recovery. The retry interval can be customized via new optional configurations `serviceUrlRecoveryInitBackoffInterval` and `serviceUrlRecoveryMaxBackoffInterval` in `ClientBuilder`
+
```
lhotari: I'd suggest renaming "backoff interval" with "quarantine duration". Backoff is about retries and this is about quarantining from the available pool. The duration can be exponential.

AuroraTwinkle:
> Some initial feedback

@lhotari All good suggestions, I will fix them and update later. Thanks for your help!

AuroraTwinkle:
> Some initial feedback

Updated!

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
+# PIP-425: Support connecting with next available endpoint for multi-endpoint serviceUrls
+Implementation PR: [#24387](https://github.com/apache/pulsar/pull/24387)
+
+# Background knowledge
+In the current Pulsar client versions, consumer or producer creation is the responsibility of PulsarClient, we can build a PulsarClient by pass serviceUrl param. The serviceUrl param supports multiple endpoints with the syntax `pulsar://localhost1:6650,localhost2:6650,localhost3:6650,...` or `pulsar+ssl://localhost1:6650,localhost2:6650,localhost3:6650,...`. What happens if some of the endpoints are unavailable? Yes, in the current version, it may cause the consumer or producer to fail to create, which is also the optimization goal of this pip.
```
lhotari: drop "local" prefix in the examples to make them more generic. For pulsar+ssl, the example port should be 6651.

lhotari: Could this be explained more directly? "Currently when an unavailable endpoint is used from the list of multiple endpoint, the client will fail to create a consumer or producer. " And also explaining whether this error is transient or persistent.

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+        .....
+
+    }
+}
+
+```
+
+# Motivation
+
+As [#22934 ](https://github.com/apache/pulsar/issues/22934) and [#22933](https://github.com/apache/pulsar/discussions/22933) mentioned, when most of the endpoints in serviceurl are unavilable (but there is at least one available endpoint), creating consumers and producers through PulsarClient will most likely fail. I think this is not as expected. If the code is robust enough, as long as there is one available endpoint, it should be accessible normally. Obviously, we can remove unavailable endpoints through some mechanism, and try to ensure the next connection to an available endpoint.
```
lhotari: It's better to avoid references to other sources in the PIP motivation part. You might combine the "background knowledge" and "motivation" sections into a single section.

In the writing style it's better to avoid personal style, instead of "I think this is not expected", you could just make the claim which some rationale, for example "this is unexpected behavior for Pulsar users since when multiple endpoints are provided, it would be expected that when an endpoint is unavailable, an available endpoint would be used", etc.

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+                .serviceUrlMaxFaultQuarantineDuration(0, TimeUnit.MILLISECONDS)
+                .build();                
+}
+
+```
+## Public-facing Changes
+
+### Public API
+
+1. Add two optional configs `serviceUrlInitFaultQuarantineDuration` and `serviceUrlMaxFaultQuarantineDuration` in `ClientBuilder`
```
lhotari: I don't think that `fault` is needed at all as part of the parameter name. `quarantine` already implies a fault.
`serviceUrlQuarantineInitDuration` and `serviceUrlQuarantineMaxDuration` would be easier to read and understand.

pip/pip-425.md
```diff
@@ -0,0 +1,127 @@
...
+```java
+    /**
+     * Configure the service URL init fault quarantine duration.
+     *
+     * <p>When the client is unable to connect to the service URL, it will wait for a certain amount of time before
+     * trying to recover the service URL endpoint. The init quarantine duration can be configured using this method.
```
lhotari: this description isn't correct. There's not recovery happening. It's simply about quarantining the endpoint url for a specific duration that is determined in a certain emponential way. A successful usage of the endpoint url will reset the duration to the initial value and move it back to the available addresses pool.

lhotari (commented):
good progress! added some follow up comments

AuroraTwinkle:
> good progress! added some follow up comments

@lhotari Modified and updated as suggested. Thanks!

pip/pip-425.md
AuroraTwinkle: Good suggestions!I will continuously improve my writing skills, thank you!

pip/pip-425.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Public API
+
+1. Add two optional configs `serviceUrlQuarantineInitDuration` and `serviceUrlQuarantineMaxDuration` in `ClientBuilder`
+```java
+    /**
+     * Configure the service URL init quarantine duration.
+     *
+     * <p>When the client is unable to connect to the a endpint from serviceUrl, that endpoint
+     *  will be quarantined for a specific duration that is determined in a certain
```
lhotari: There should be explanation about this specifically applying to cases where a serviceUrl with multiple hosts is specified. For single host serviceUrl, this setting has no effect.

AuroraTwinkle: fixed

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+        .....
+
+    }
+}
+
+```
+
+# Motivation
+
+When most of the endpoints in serviceurl are unavilable (but there is at least one available endpoint), creating consumers and producers through PulsarClient will most likely fail. This is an unexpected behavior for Pulsar users since when multiple endpoints are provided, it would be expected that when an endpoint is unavailable, an available endpoint would be used. To solve this problem, we can quarantine unavailable endpoints through some mechanism, and try to ensure the next connection to an available endpoint.
```
codelipenghui:
```suggestion
When most of the endpoints in service url are unavailable (but there is at least one available endpoint), creating consumers and producers through PulsarClient will most likely fail. This is an unexpected behavior for Pulsar users since when multiple endpoints are provided, it would be expected that when an endpoint is unavailable, an available endpoint would be used. This behavior is counter-intuitive for users who expect the client to transparently fail over to a working endpoint. To address this, we propose an intelligent feedback mechanism that allows the client to identify, quarantine, and eventually retry unavailable endpoints, ensuring that new connections are routed to healthy brokers whenever possible.
```

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+
+
+# Goals
+Optimize the PulsarClient code logic, remove unavailable serviceUrl endpoints through the feedback mechanism, and improve the success rate of PulsarClient requests.
+
+## In Scope
+If this pip is accepted, it will help improve the PulsarClient success rate of consumer or producer creation, even if most of the endpoints in serviceurl are no longer available.
+
+## Out of Scope
+[#22935](https://github.com/apache/pulsar/pull/22935) removes unavailable endpoints through a regular health check mechanism, but this brings new problems (frequent creation of connections and increased system load). So this pip will not use the health check solution.
```
codelipenghui: It should be moved to the Alternatives section

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
+# PIP-425: Support connecting with next available endpoint for multi-endpoint serviceUrls
+Implementation PR: [#24387](https://github.com/apache/pulsar/pull/24387)
+
```
codelipenghui:
```suggestion
# Abstract
This proposal introduces a mechanism for the Pulsar client to gracefully handle unavailable service endpoints. When configured with a multi-endpoint service URL, the client will temporarily quarantine unresponsive endpoints and automatically attempt to connect to the next available one. This change introduces an intelligent failover strategy with exponential backoff, significantly improving the client's connection reliability and resilience in environments where broker availability can fluctuate.
```

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+# Background knowledge
+In the current Pulsar client versions, consumer or producer creation is the responsibility of PulsarClient, we can build a PulsarClient by pass serviceUrl param. The serviceUrl param supports multiple endpoints with the syntax `pulsar://host1:6650,host2:6650,host3:6650,...` or `pulsar+ssl://host1:6651,host2:6651,host3:6651,...`. Currently when an unavailable endpoint is used from the list of multiple endpoints, the PulsarClient will fail to create a consumer or producer. Unfortunately it will continue to fail unless the unavailable endpoint becomes available.
```
codelipenghui:
```suggestion
In the current Pulsar client versions, consumer or producer creation is the responsibility of PulsarClient, we can build a PulsarClient by pass serviceUrl param. The serviceUrl param supports multiple endpoints with the syntax `pulsar://host1:6650,host2:6650,host3:6650,...` or `pulsar+ssl://host1:6651,host2:6651,host3:6651,...`. When a Pulsar client tries to create a producer or consumer, it may pick an endpoint from the list that is temporarily unavailable. The creation will fail, but worse, subsequent attempts may continue to pick the same unavailable endpoint, leading to repeated failures. This occurs even when other endpoints in the list are perfectly healthy and available.
```

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+# Motivation
+
+When most of the endpoints in serviceurl are unavilable (but there is at least one available endpoint), creating consumers and producers through PulsarClient will most likely fail. This is an unexpected behavior for Pulsar users since when multiple endpoints are provided, it would be expected that when an endpoint is unavailable, an available endpoint would be used. To solve this problem, we can quarantine unavailable endpoints through some mechanism, and try to ensure the next connection to an available endpoint.
+
+
+# Goals
+Optimize the PulsarClient code logic, remove unavailable serviceUrl endpoints through the feedback mechanism, and improve the success rate of PulsarClient requests.
+
+## In Scope
+If this pip is accepted, it will help improve the PulsarClient success rate of consumer or producer creation, even if most of the endpoints in serviceurl are no longer available.
```
codelipenghui:
```suggestion
- Improve Connection Resilience: Modify the PulsarClient to avoid repeatedly trying to connect to unavailable endpoints.
- Introduce Endpoint Quarantining: When a connection to an endpoint fails, temporarily remove it from the pool of available endpoints.
- Implement Exponential Backoff: Re-introduce quarantined endpoints for connection attempts after a dynamically increasing delay, allowing them to recover without overwhelming the system.
```

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Out of Scope
+[#22935](https://github.com/apache/pulsar/pull/22935) removes unavailable endpoints through a regular health check mechanism, but this brings new problems (frequent creation of connections and increased system load). So this pip will not use the health check solution.
+
+# Detailed Design
```
codelipenghui, line 43:
```suggestion
# Detailed Design
We will introduce an endpoint quarantining strategy based on passive feedback from connection attempts.
```

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+**Explanation of nouns:**
+- `availableAddressPool`: a pool containing all available addresses, subset of serviceUrl.
+- `allAddressPool`: a pool containing all addresses in serviceUrl.
+
+## Design & Implementation Details
```
codelipenghui:
``` suggestion
By default, the client will only try to connect to endpoints from the availableAddressPool.

- On Connection Failure: If the client fails to connect to an endpoint, it will be temporarily removed from the availableAddressPool and placed into "quarantine".
- Exponential Backoff: The quarantine duration will increase exponentially upon consecutive failures.
- On Connection Success: A successful connection to an endpoint proves it is available. If the endpoint was previously failing, its quarantine backoff state will be reset. It will be immediately considered part of the availableAddressPool.
- Fallback: If the availableAddressPool ever becomes empty, the client will fall back to trying a random endpoint from the allAddressPool. This prevents a deadlock where the client cannot connect because all known endpoints are simultaneously in quarantine.
```

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+     *  will be quarantined for a specific duration that is determined in a certain emponential way. 
+     * The init value of a single quarantine duration is set by
+     * @param serviceUrlQuarantineInitDuration. A successful usage of the endpoint will reset the
+     * duration to the initial value and move it back to the available addresses pool.
+     *
+     * <p>
+     * A value of 0 means don't quarantine any endpoints even if they fail.
+     * @param serviceUrlQuarantineInitDuration the initial quarantine duration for unavailable endpoint
+     * @param unit the time unit for the quarantine duration
+     * @return the client builder instance
```
codelipenghui: Is it also better to add the default value?

pip/pip-425.md
```diff
@@ -0,0 +1,135 @@
...
+     * will be quarantined for a specific duration that is determined in a certain emponential way. 
+     * The max value of a single quarantine duration is set by
+     * @param serviceUrlQuarantineMaxDuration. A successful usage of the endpoint will reset the
+     * duration to the initial value and move it back to the available addresses pool.
+     *
+     * <p>
+     * A value of 0 means don't quarantine any endpoints even if they fail.
+     * @param serviceUrlQuarantineMaxDuration the maximum quarantine duration for unavailable endpoint
+     * @param unit the time unit for the quarantine duration
+     * @return the client builder instance
```
codelipenghui: Is it also better to add the default value?

AuroraTwinkle:
> Hi @AuroraTwinkle
> 
> The proposal looks good to me. I’ve added a few comments to help clarify the problem and the proposed solution, making it easier to understand.

Ok, I will fix them later, Thanks!

AuroraTwinkle:
> Hi @AuroraTwinkle
> 
> The proposal looks good to me. I’ve added a few comments to help clarify the problem and the proposed solution, making it easier to understand.

@codelipenghui Very interesting and detailed suggestions, I have fixed them, thank you again!

codelipenghui (approved):

pip/pip-425.md
```diff
@@ -0,0 +1,147 @@
...
+     * Mark the availability of a host.
+     * @param address the host address to mark availability for
+     * @param isAvailable true if the host is available, false otherwise
+     */
+    default void markHostAvailability(InetSocketAddress address, boolean isAvailable){
+        // Default implementation does nothing
+        // Subclass can override this method to implement host availability tracking
+    }
+```
+When resolve one address in serviceUrl, the `ConnectionPool` will invoke `markHostAvailability` to tell availability about that address, so the `ServiceNameResolver` can remove it from `availableAddressPool` if unavailable (or recover back to `availableAddressPool` if available). If all addressed in  `availableAddressPool` are unavailable, the resolver implementation could start getting random addresses from the `allAddressPool`.
```
315157973: It seems we need to add a circuit breaker mechanism. Can we simply address your problem by:
1. Changing from establishing connections by randomly selecting nodes to using a round-robin approach.
2. To prevent hotspots caused by multiple clients connecting to the same node , shuffling the list of IPs used for round-robin.

pip/pip-425.md
```diff
@@ -0,0 +1,147 @@
...
+**Explanation of nouns:**
+- `availableAddressPool`: a pool containing all available addresses, subset of serviceUrl.
+- `allAddressPool`: a pool containing all addresses in serviceUrl.
+
+By default, the client will only try to connect to endpoints from the availableAddressPool.
+
+- On Connection Failure: If the client fails to connect to an endpoint, it will be temporarily removed from the availableAddressPool and placed into "quarantine".
+- Exponential Backoff: The quarantine duration will increase exponentially upon consecutive failures.
+- On Connection Success: A successful connection to an endpoint proves it is available. If the endpoint was previously failing, its quarantine backoff state will be reset. It will be immediately considered part of the availableAddressPool.
+- Fallback: If the availableAddressPool ever becomes empty, the client will fall back to trying a random endpoint from the allAddressPool. This prevents a deadlock where the client cannot connect because all known endpoints are simultaneously in quarantine.
```
315157973: If I restart the entire cluster at once, and the available IP pools are all empty at that moment, how should the client handle it?

AuroraTwinkle: If the available IP pools are all empty, the client will fall back to select a random endpoint from the allAddressPool.Make sure at least one endpoint can be selected.

pip/pip-425.md
AuroraTwinkle: OK, that was a typo, what I meant was to use a round-robin approach to select endpoint.

315157973: If we use round-robin, do we still need all the interfaces you've listed, or do we only need to keep track of the current round-robin index?

liangyepianzhou (approved):

lhotari (approved):
LGTM, good work @AuroraTwinkle

pip/pip-425.md
AuroraTwinkle: https://github.com/apache/pulsar/blob/5841238d49695e29f557e22ce964d5b7470a354a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java#L46-L58

@315157973 OKay, maybe I didn't explain it clearly enough.
This pip will not change the algorithm for selecting address, and will still maintain the original round-robin algorithm. What we change is the value of the list variable(at line 47) in the `resolveHost` function. When there are available addresses, an address is selected from the available address list. When there are no available addresses,  fall back to the original list (all address passed in by serviceUrl) for selection.

315157973 (approved):

liangyepianzhou:
@lhotari @codelipenghui @315157973 We need more votes on the mailing list to close this PIP. Could you please help vote when you have a moment?
https://lists.apache.org/thread/c2zvjwf7bqp8nc2rpzbxd4kdtztk23xp

Technoboy- (approved):

PR #24395, opened 2025-06-08
[fix][broker] fix the direct memory OOM by parallel pattern subscription

mattisonchao:
### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -2503,7 +2503,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         final NamespaceName namespaceName = NamespaceName.get(namespace);
 
         final Semaphore lookupSemaphore = service.getLookupRequestSemaphore();
-        if (lookupSemaphore.tryAcquire()) {
+        if (lookupSemaphore.tryAcquire() || !ctx.channel().isWritable()) {
             isNamespaceOperationAllowed(namespaceName, NamespaceOperation.GET_TOPICS).thenApply(isAuthorized -> {
                 if (isAuthorized) {
                     getBrokerService().pulsar().getNamespaceService().getListOfUserTopics(namespaceName, mode)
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2531,17 +2531,30 @@ protected void handleGetTopicsOfNamespace(CommandGetTopicsOfNamespace commandGet
...
-                            commandSender.sendGetTopicsOfNamespaceResponse(filteredTopics, hash, filterTopics,
-                                    !hashUnchanged, requestId);
+                            if (ctx.channel().isWritable()) {
+                                commandSender.sendGetTopicsOfNamespaceResponse(filteredTopics, hash, filterTopics,
+                                        !hashUnchanged, requestId);
+                            } else {
+                                /*
+                                    To avoid direct OutOfMemoryError (OOM), we should refrain from appending a large
+                                    volume of topic lists to the Netty ChannelOutboundBuffer's pending queue when the
+                                    channel is in an unwritable state.
```
lhotari, line 2541: How large is the data volume that is sent out? Does this solution really help in addressing the problem?
The same channel might be used for other purposes at the same time, for sending messages to consumers and the channel might not be writable even after retries.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2531,17 +2531,30 @@ protected void handleGetTopicsOfNamespace(CommandGetTopicsOfNamespace commandGet
                                         "[{}] Received CommandGetTopicsOfNamespace for namespace [//{}] by {}, size:{}",
                                         remoteAddress, namespace, requestId, topics.size());
                             }
-                            commandSender.sendGetTopicsOfNamespaceResponse(filteredTopics, hash, filterTopics,
```
lhotari, line 2534: If a very large buffer is requested, a reason for OOM could be simply that the memory space is fragmented and there isn't an available non-continuos block of memory to be used. 
A workaround for this problem is to allocate a composite direct buffer. This is the approach used in PrometheusMetricsGenerator: https://github.com/apache/pulsar/blob/9b79d701a2c1067693508515915b5a4504f60916/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/PrometheusMetricsGenerator.java#L368-L388

lhotari:
In addition to https://github.com/apache/pulsar/pull/24395#discussion_r2134770661, configuring `maxMessagePublishBufferSizeInMB`, `managedLedgerMaxReadsInFlightSizeInMB` and `managedLedgerCacheSizeMB` in the broker should help avoiding OOM.

PR #24399, opened 2025-06-09
[improve][ci] Use pulsar-test-latest-version-image instead of java-test-image

nodece:
### Motivation

The `apachepulsar/pulsar-test-latest-version:latest` image already includes all the features provided by `apachepulsar/java-test-image:latest`. Therefore, building the `java-test-image` module is redundant.

### Modifications

- Removed the `java-test-image` module.
- Replaced `apachepulsar/java-test-image:latest` with `apachepulsar/pulsar-test-latest-version:latest` in the CI.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


nodece:
Maybe we can provide a flag to build a minimal image based on pulsar-test-latest-version-image.

lhotari:
> Maybe we can provide a flag to build a minimal image based on pulsar-test-latest-version-image.

The maven build doesn't couple to java-test-image. What problem does java-test-image cause in practice in your case?
It's more minimal when java-test-image is used since each build job doesn't have to download so much to run the tests. It's actually saving resources when java-test-image is used. That's why I don't see why this PR would make sense.

nodece:
Thanks for the feedback — I understand that `java-test-image` is more minimal and helps save resources. The intention of this PR is purely to remove redundancy, since `pulsar-test-latest-version` already includes everything from `java-test-image`. That said, I'm open to closing this if keeping the smaller image is preferred for efficiency.

lhotari:
> Thanks for the feedback — I understand that `java-test-image` is more minimal and helps save resources. The intention of this PR is purely to remove redundancy, since `pulsar-test-latest-version` already includes everything from `java-test-image`. That said, I'm open to closing this if keeping the smaller image is preferred for efficiency.

I'll close this PR since it wouldn't improve Pulsar CI due to the reasons explained in earlier comments.

PR #24400, opened 2025-06-09
PIP-426: Enhanced Consumer Throttling and Unacknowledged Message Tracking for Exclusive and Failover Subscriptions

berg223:
Fixes https://github.com/apache/pulsar/issues/24159

Main Issue: https://github.com/apache/pulsar/issues/24159

### Motivation

1. I have tried to fix the issue by https://github.com/apache/pulsar/pull/24376. However, that PR doesn't support features about cumulative ack, batching mode, and transaction. So I want to improve it further.
2. I have found another issue that flowcontrol of exclusive or failover consumer is not work. Since the issues are highly correlated, I want to fix them at the same time. 

### Modifications

1. Remove the `Subscription.isIndividualAckMode()` restriction that limits `pendingAcks` usage to shared subscriptions.
2. Extend `PendingAckHandleImpl` for exclusive/failover consumers.
3. Enhance flow control in `PersistentDispatcherSingleActiveConsumer`

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [X] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-426.md
+++ b/pip/pip-426.md
@@ -1,0 +1,163 @@
+
+# PIP-426: Enhanced Consumer Throttling and Unacknowledged Message Tracking for Exclusive and Failover Subscriptions
+
+# Background knowledge
+
+<!--
+Describes all the knowledge you need to know in order to understand all the other sections in this PIP
+
+* Give a high level explanation on all concepts you will be using throughout this document. For example, if you want to talk about Persistent Subscriptions, explain briefly (1 paragraph) what this is. If you're going to talk about Transaction Buffer, explain briefly what this is. 
+  If you're going to change something specific, then go into more detail about it and how it works. 
+* Provide links where possible if a person wants to dig deeper into the background information. 
+
+DON'T
+* Do not include links *instead* explanation. Do provide links for further explanation.
+
+EXAMPLES
+* See [PIP-248](https://github.com/apache/pulsar/issues/19601), Background section to get an understanding on how you add the background knowledge needed.
+  (They also included the motivation there, but ignore it as we place that in Motivation section explicitly).
+-->
+
+Apache Pulsar currently lacks full support for enforcing unacknowledged message limits and consumer-side flow control in exclusive and failover subscriptions. While these mechanisms function correctly for shared subscriptions, their absence in exclusive/failover modes causes critical limitations:
+
+1. **Throttling Inaccuracies**  
+   Consumers can exceed configured `maxUnackedMessagesOnConsumer` limits, risking resource exhaustion.
+
+2. **Tracking Deficiencies**  
+   `pendingAcks` tracking is disabled or partially functional, compromising visibility into message processing.
+
+3. **Feature Incompatibilities**  
+   Current implementation lacks support for:
+   - Cumulative acknowledgements
+   - Message batching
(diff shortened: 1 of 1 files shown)


pip/pip-426.md
```diff
@@ -0,0 +1,163 @@
...
+DON'T
+* Do not include links *instead* explanation. Do provide links for further explanation.
+
+EXAMPLES
+* See [PIP-248](https://github.com/apache/pulsar/issues/19601), Background section to get an understanding on how you add the background knowledge needed.
+  (They also included the motivation there, but ignore it as we place that in Motivation section explicitly).
+-->
+
+Apache Pulsar currently lacks full support for enforcing unacknowledged message limits and consumer-side flow control in exclusive and failover subscriptions. While these mechanisms function correctly for shared subscriptions, their absence in exclusive/failover modes causes critical limitations:
+
```
lhotari, line 11: Yes, I agree with this.

After having "pending acks" also for failover subscriptions, it will be easier to come up with a proper solution for https://github.com/apache/pulsar/issues/15189 . When a failover subscription changes the active consumer, it causes duplicate processing which is problematic and surprising for many applications. Failover subscription active consumer changing should be potentially handled by first "draining" the previous consumer before sending messages to the new active consumer. (The draining approach is used for Key_Shared subscriptions since Pulsar 4.0 / PIP-379)

berg223: Yes. It's also benefit to solve similiar issues to #15189.  The draining approach could also be implemented .  Your comments always open my mind and point me in the right direction !

pip/pip-426.md
```diff
@@ -0,0 +1,198 @@
...
+# High Level Design
+
+1. Remove the `Subscription.isIndividualAckMode()` restriction that limits `pendingAcks` usage to shared subscriptions.
+2. Extend `PersistentAcknowledgmentsGroupingTracker` for exclusive/failover consumers.
+3. Enhance flow control in `PersistentDispatcherSingleActiveConsumer`
+
+# Detailed Design
+
+## Design & Implementation Details
```
codelipenghui, line 56: Indeed, each message in a Pulsar topic can have a persistent, monotonically increasing local index, which functions much like a Kafka offset. This feature was officially introduced by Pull Request #9039, which implemented PIP-70 ("Introduce lightweight broker entry metadata").

With this local index, the number of unacknowledged messages for a consumer can be calculated with a simple subtraction:

unacked_messages = last_delivered_index - mark_delete_position

Given its reliability, considering its proven value, enabling this index by default in a future Apache Pulsar release would be reasonable.

berg223: Thanks for more background knowledge! It's really helpful to solve origin issue #24159 . Since there is no need to add flow permits to exlusive or failover subscription (details discussion see: https://lists.apache.org/thread/glvd8lrvyll9mdtp62d34x1k38swrls1), there is no motivation except fix the origin issue. IMO, we can close this pip to keep the pip from too much useless information.

PR #24405, opened 2025-06-12
[fix][ml]Received more than once callback when calling cursor.delete

poorbarcode:
### Motivation

**About cursor deleting, there are two cases**
- `scenario-A`: Acked nothing, because all positions have been acknowledged before.
- `scenario-B`: Acked some position, then it will try to `mark-deleted`.
  - `scenario-B1`:  Acked some positions, but `individualDeletedMessages` became empty after combining ranges in the method `setAcknowledgedPosition`, 
  - `scenario-B2`: Acked some positions, and `individualDeletedMessages` was still not empty.

**Triggering callbacks**
- Regarding the `scenario-A`, Cursor will skip calling `mark-deleted`, because it is meaningless.
- Regarding the `scenario-B`, Cursor will call `mark-deleted`, which will trigger a `callback`.

**Issue**
Regarding the `scenario-B1`, the Cursor triggers twice the callbacks.
- The first time: calls the callback if `individualDeletedMessages` is empty. https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L2479

```java
boolean empty = individualDeletedMessages.isEmpty();
if (empty) {
    callback.deleteComplete(ctx);
}
```
- The first time: calls the callback after calling `mark-deleted`. https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L2499

```java
internalAsyncMarkDelete(newMarkDeletePosition, properties, new MarkDeleteCallback() {
        @Override
        public void markDeleteComplete(Object ctx) {
            callback.deleteComplete(ctx);
        }
}
```

### Modifications

To confirm the scenario matches `scenario-A`, rather than confirming `individualDeletedMessages.isEmpty`, use a specific variable to record it.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -103,6 +103,7 @@ import org.apache.bookkeeper.mledger.proto.MLDataFormats.MessageRange;
 import org.apache.bookkeeper.mledger.proto.MLDataFormats.PositionInfo;
 import org.apache.bookkeeper.mledger.proto.MLDataFormats.PositionInfo.Builder;
 import org.apache.bookkeeper.mledger.proto.MLDataFormats.StringProperty;
+import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.pulsar.common.policies.data.ManagedLedgerInternalStats;
@@ -2361,7 +2362,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         Position newMarkDeletePosition = null;
 
         lock.writeLock().lock();
-
+        final MutableBoolean cbHasExecuted = new MutableBoolean(false);
         try {
             if (log.isDebugEnabled()) {
                 log.debug("[{}] [{}] Deleting individual messages at {}. Current status: {} - md-position: {}",
@@ -2477,6 +2478,7 @@ public class ManagedCursorImpl implements ManagedCursor {
             boolean empty = individualDeletedMessages.isEmpty();
             lock.writeLock().unlock();
             if (empty) {
+                cbHasExecuted.setTrue();
                 callback.deleteComplete(ctx);
             }
         }
@@ -2485,7 +2487,9 @@ public class ManagedCursorImpl implements ManagedCursor {
         if (markDeleteLimiter != null && !markDeleteLimiter.tryAcquire()) {
             isDirty = true;
             updateLastMarkDeleteEntryToLatest(newMarkDeletePosition, null);
-            callback.deleteComplete(ctx);
+            if (!cbHasExecuted.booleanValue()) {
+                callback.deleteComplete(ctx);
+            }
             return;
         }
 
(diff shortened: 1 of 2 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2496,12 +2500,18 @@ public void asyncDelete(Iterable<Position> positions, AsyncCallbacks.DeleteCallb
             internalAsyncMarkDelete(newMarkDeletePosition, properties, new MarkDeleteCallback() {
                 @Override
                 public void markDeleteComplete(Object ctx) {
-                    callback.deleteComplete(ctx);
+                    if (!cbHasExecuted.booleanValue()) {
+                        cbHasExecuted.setTrue();
+                        callback.deleteComplete(ctx);
+                    }
```
3pacccccc: considering switch the order of ```cbHasExecuted.setTrue();``` and ```callback.deleteComplete(ctx);``` ?
if there is any excptions in ```callback.deleteComplete(ctx);``` happened, cbHasExecuted is already been set to true, and there is no callback will be called.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2361,7 +2362,7 @@ public void asyncDelete(Iterable<Position> positions, AsyncCallbacks.DeleteCallb
         Position newMarkDeletePosition = null;
 
         lock.writeLock().lock();
-
+        final MutableBoolean cbHasExecuted = new MutableBoolean(false);
```
codelipenghui: Is it better to add to `DeleteCallback`?

Something like

```
if (!callback.isCompleted()) {
    callback.deleteComplete(ctx);
}
```

codelipenghui:
@poorbarcode Could you please write down the reason that why the callback get called multiple times? If there is anything that we cannot control (network failures or something). I'm good to add a defensive solution to avoid multiple calls. If the issue which we can control, I would like to fix the issue at the first place.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2477,6 +2478,7 @@ public void asyncDelete(Iterable<Position> positions, AsyncCallbacks.DeleteCallb
             boolean empty = individualDeletedMessages.isEmpty();
             lock.writeLock().unlock();
             if (empty) {
+                cbHasExecuted.setTrue();
                 callback.deleteComplete(ctx);
```
Technoboy-, line 2481: I think the bug is here. 
If the callback executed here, then it may be executed at line 2488 or 2499

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Rewrote the Motivation and the Implementation, please take a look

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Rewrote the Motivation and the Implementation, please take a look

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Yes, that's the root cause. I rewrote the Motivation and the Implementation, please take a look

Technoboy- (approved):

codelipenghui (approved):

PR #24406, opened 2025-06-12
[fix][ml] Cursor ignores the position that has an empty ack-set if disabled deletionAtBatchIndexLevelEnabled

poorbarcode:
### Motivation

After disabling `deletionAtBatchIndexLevelEnabled`, `ManagedCursorImpl` ignores positions that have an empty `ackSet` when calling `cursor.delete`, which leads to the caller acknowledging nothing.

And the Transaction task will acknowledge positions that have an empty `ackSet` after all single messages of a batch were acknowledged, 

- https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/PositionAckSetUtil.java#L52 
```java
    public static void andAckSet(Position currentPosition, Position otherPosition) {
        if (currentPosition == null || otherPosition == null) {
            return;
        }
        AckSetState currentAckSetState = AckSetStateUtil.getAckSetState(currentPosition);
        AckSetState otherAckSetState = AckSetStateUtil.getAckSetState(otherPosition);
        // The ackSet will be an empty long array after the latest single message of a batch was acked.
        currentAckSetState.setAckSet(andAckSet(currentAckSetState.getAckSet(), otherAckSetState.getAckSet()));
    }
```

- https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/pendingack/impl/PendingAckHandleImpl.java#L804

```java
private void handleIndividualAck(TxnID txnID, List<MutablePair<Position, Integer>> positions) {
     ...
     if (pendingAckMessageForCurrentTxn.containsKey(position)) {
          PositionAckSetUtil.andAckSet(pendingAckMessageForCurrentTxn.get(position), position);
      }
     ...
}
```

### Modifications

Fix the bug

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -2390,7 +2390,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                     continue;
                 }
                 long[] ackSet = AckSetStateUtil.getAckSetArrayOrNull(position);
-                if (ackSet == null) {
+                if (ackSet == null || ackSet.length == 0) {
                     if (batchDeletedIndexes != null) {
                         batchDeletedIndexes.remove(position);
                     }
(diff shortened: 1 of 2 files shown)


codelipenghui (approved):

dao-jun (approved):

lhotari:
@poorbarcode What Pulsar versions are impacted by this bug? In other words, when did we introduce this regression?

PR #24365, opened 2025-05-30
[improve] Enable metrics for all broker caches

merlimat:
### Motivation

Several broker caches metrics are not exposed in the dashboards. We need to manually register them for the metrics to be collected.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24407, opened 2025-06-13
[Fix][broker] Fix to gather non-zero statistics

Technoboy-:
### Motivation

It's a follow up fix for #24365. Because we need to call `recordStats` to gather non-zero statistics, otherwise the metrics will be 
```
caffeine_cache_eviction_weight{cluster="test",cache="-[map type; class org.apache.pulsar.common.policies.data.BookiesRackConfiguration, [simple type, class java.lang.String] -> [map type; class java.util.Map, [simple type, class java.lang.String] -> [simple type, class org.apache.pulsar.common.policies.data.BookieInfo]]]"} 0.0
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


codelipenghui (commented):
@Technoboy- Is it possible to add a test for verifying the final metrics labels value?

merlimat (approved):

dao-jun (approved):

PR #24420, opened 2025-06-17
[improve][admin] PIP-416: Add a new topic method to implement trigger offload by size threshold

JunFu0814:
Main Issue: #24276 

### Motivation

For pip #24276 ,  add a new admin api for trigger offload with size threshold.

### Modifications

1. Add new admin apis for trigger offload with size threshold.
2. Optimize formatting.
3. Enhance testing.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
+++ b/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
@@ -2112,7 +2112,7 @@ public interface Topics {
      * Trigger offloading messages in topic to longterm storage.
      *
      * @param topic the topic to offload
-     * @param sizeThreshold maximum amount of data (bytes) that should be retained in longterm storage
+     * @param sizeThreshold maximum amount of data (bytes) that should be retained in bookkeeper
      */
     void triggerOffload(String topic, long sizeThreshold) throws PulsarAdminException;
 
@@ -2129,7 +2129,7 @@ public interface Topics {
      * Trigger offloading messages in topic to longterm storage asynchronously.
      *
      * @param topic the topic to offload
-     * @param sizeThreshold maximum amount of data (bytes) that should be retained in longterm storage
+     * @param sizeThreshold maximum amount of data (bytes) that should be retained in bookkeeper
      */
     CompletableFuture<Void> triggerOffloadAsync(String topic, long sizeThreshold);
 
--- a/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
+++ b/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
@@ -2109,12 +2109,29 @@ public interface Topics {
     void triggerOffload(String topic, MessageId messageId) throws PulsarAdminException;
 
     /**
+     * Trigger offloading messages in topic to longterm storage.
+     *
+     * @param topic the topic to offload
+     * @param sizeThreshold maximum amount of data (bytes) that should be retained in longterm storage
+     */
+    void triggerOffload(String topic, long sizeThreshold) throws PulsarAdminException;
+
+    /**
      * Trigger offloading messages in topic to longterm storage asynchronously.
      *
      * @param topic the topic to offload
      * @param messageId ID of maximum message which should be offloaded
      */
     CompletableFuture<Void> triggerOffloadAsync(String topic, MessageId messageId);
+
+
+    /**
+     * Trigger offloading messages in topic to longterm storage asynchronously.
+     *
+     * @param topic the topic to offload
+     * @param sizeThreshold maximum amount of data (bytes) that should be retained in longterm storage
+     */
+    CompletableFuture<Void> triggerOffloadAsync(String topic, long sizeThreshold);
 
     /**
      * Check the status of an ongoing offloading operation for a topic.
--- a/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
+++ b/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
@@ -19,6 +19,7 @@
 package org.apache.pulsar.client.admin.internal;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
@@ -27,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -1208,6 +1210,11 @@ public class TopicsImpl extends BaseResource implements Topics {
     }
 
     @Override
+    public void triggerOffload(String topic, long sizeThreshold) throws PulsarAdminException {
+        sync(() -> triggerOffloadAsync(topic, sizeThreshold));
+    }
+
+    @Override
     public CompletableFuture<Void> triggerOffloadAsync(String topic, MessageId messageId) {
         TopicName tn = validateTopic(topic);
         WebTarget path = topicPath(tn, "offload");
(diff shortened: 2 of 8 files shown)


pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
```diff
@@ -1231,6 +1238,42 @@ public void failed(Throwable throwable) {
         return future;
     }
 
+    @Override
+    public CompletableFuture<Void> triggerOffloadAsync(String topic, long sizeThreshold) {
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        try {
+            PersistentTopicInternalStats stats = getInternalStats(topic);
```
nodece: Please use `getInternalStatsAsync` instead of `getInternalStats` to avoid thread blocking.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
```diff
@@ -2108,6 +2108,14 @@ CompletableFuture<Void>  updateSubscriptionPropertiesAsync(String topic, String
      */
     void triggerOffload(String topic, MessageId messageId) throws PulsarAdminException;
 
+    /**
+     * Trigger offloading messages in topic to longterm storage.
+     *
+     * @param topic the topic to offload
+     * @param sizeThreshold maximum amount of data (bytes) that should be retained in longterm storage
```
codelipenghui: Do you mean 

@param sizeThreshold maximum amount of data (bytes) that should be retained in bookkeeper?

longterm storage will be a little confusing, not sure it means bookkeeper or tiered storage.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
```diff
@@ -1231,6 +1238,42 @@ public void failed(Throwable throwable) {
...
+        long previousLedger = ledgers.get(0).ledgerId;
+        for (PersistentTopicInternalStats.LedgerInfo l : ledgers) {
+            suffixSize += l.size;
+            if (suffixSize > sizeThreshold) {
+                return new MessageIdImpl(previousLedger, 0L, -1);
+            }
+            previousLedger = l.ledgerId;
+        }
+        return null;
+    }
```
codelipenghui: It's better to move to the broker side which can provide consistent behavior from the admin CLI and the admin REST API.

codelipenghui (changes requested):

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
JunFu0814: nice suggestion

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
JunFu0814: Indeed, the size here refers to the data retained in Bookkeeper. The description here is a bit confusing, I will give a more precise explanation

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
JunFu0814: @nodece please review again bc9d73554a195d38e5d5704bdbae7e1c9eb0cfc0

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
JunFu0814: hi @codelipenghui , It is a better way to move to the broker side, which can ensure consistent behavior of cli and clients of any language type, but of course this will also bring more workload. In the current pip, I will first ensure that cli and java client use the same `findFirstLedgerWithinThreshold` logic.  Please help review on this https://github.com/apache/pulsar/pull/24420/commits/ff002be87098f63bfe61121bfc81c5cf6a52b6b3.

codelipenghui: > but of course this will also bring more workload

While both approaches require iterating over ledgers, the scope of that iteration differs:

Client-side: You'd have to iterate over all ledgers to determine what needs offloading.

Broker-side: The broker only needs to iterate over the specific ledgers that are being offloaded, directly accessing data already in memory.

Even if the client pre-calculates message IDs and sends them to the broker, the broker still needs to iterate the ledger map and decide ledgers should be offloaded. The key is that the broker can do this much more efficiently, leveraging its existing in-memory structures without the constant creation and disposal of new objects.

codelipenghui: BTW, the broker side already has most of the implementation

https://github.com/apache/pulsar/blob/73a4ae4f886936ec33e5c8086b6c0c0151ea7f76/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2619-L2696

There is no need to have duplicated codes for like `findFirstLedgerWithinThreshold`.

JunFu0814: @codelipenghui Thank you for your suggestion, I will understand the logic here in depth frist

dao-jun (approved):

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
JunFu0814: HI @codelipenghui , I have gained an understanding of the offload-related logic. As you mentioned, regardless of whether the offload is executed automatically or the offload operation is manually triggered, the logic on the broker side will ultimately be executed in the `offloadLoop` code, and `offloadThresholdInBytes` is a crucial condition. Therefore, we can provide a unified API based on offloadThresholdInBytes on the broker side, enabling the admin client and CLI to no longer need to be concerned about the `findFirstLedgerWithinThreshold` logic. Regarding the changes here, I suggest introducing a new PIP for discussion and to finalize the optimization solution. In PIP-416 #24276 , our goal is merely to provide a means in the Java admin client to manually trigger topic offload based on the offloadThresholdInBytes condition (since the CLI already offers such a method), thereby enhancing the user experience. I believe these two approaches are not conflicting. https://github.com/apache/pulsar/blob/73a4ae4f886936ec33e5c8086b6c0c0151ea7f76/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3436-L3449

JunFu0814: @nodece What suggestions do you have for this?

nodece: The broker already supports automatic offload, and users can manually trigger offload using tools like the pulsar-admin CLI. Currently, @codelipenghui is proposing to enhance offload handling directly on the broker side, possibly by adding new methods or APIs to allow the broker to make offload decisions.

While I understand the motivation for this, adding such logic to the broker would increase complexity and require additional API surface. An alternative: handling the logic on the client side. In this approach, the client would fetch relevant metadata from the broker, apply the offload decision logic, and trigger offload manually through existing interfaces. This avoids modifying core broker logic and offers more flexibility for customized behavior.

That said, we could still consider implementing manual triggering directly on the broker side in the future as a built-in capability. But for now, I think starting with the client-side logic is a simpler and more incremental step.

nodece (approved):

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
codelipenghui: @nodece 

> adding such logic to the broker would increase complexity and require additional API surface

Could you please provide more details of where increased the complexity? new REST API?
As I understand, moving to the client side will definitely adding complexity to the system

- Duplicated codes added for finding the ledger
- Inconsistent REST API and admin CLI (If other people want to trigger the offset via REST API, how can we support them?)
- Pulsarctl or other CLI tools need to catchup the feature with adding duplicated codes

> This avoids modifying core broker logic and offers more flexibility for customized behavior.

And I don't think it needs to modifying the core broker logic, it just expose the existing internal capability(tigger data offloading) to users.

> That said, we could still consider implementing manual triggering directly on the broker side in the future as a built-in capability. But for now, I think starting with the client-side logic is a simpler and more incremental step.

And it definitely not a more incremental step. Once we supported REST API to trigger offload by size threshold, all the CLI tools need to remove the existing implementation and move to the new REST API.

PR #24423, opened 2025-06-17
[fix][broker]Broker OOM because ServerCnx handling too many pending write, when the client can not handle response quick enough

poorbarcode:
### Motivation

Thanks @mattisonchao for sharing the issue context and the code for reproducing the issue.

- There is a write buffer that Netty provided is `outboundBuffer = io.netty.channel.ChannelOutboundBuffer`, it is used to cache data that should be written out, and batch flush to benefit performance.
- When `ChannelOutboundBuffer` caches too much data, it will mark `channel` as `unwritable`, but it continuously caches the data, if the user continuously calls `channel.write`, eventually it leads to an OOM
- You can reproduce the issue with the new test 

Screenshot 2025-06-17 at 23 30 03

### Modifications

- Add a broker param: `connectionMaxPendingWriteBytes`, pause handling requests when `ChannelOutboundBuffer`'s pending write bytes length is larger than the threshold. Resume when the channel is `writable` again.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -920,6 +920,13 @@ public class ServiceConfiguration implements PulsarConfiguration {
 
     @FieldContext(
         category = CATEGORY_POLICIES,
+        doc = "The maximum number of pending write bytes of per connection when the connection is unwritable because"
+            + " the client can not handle too many response. Using a value of -1, is disabling it."
+    )
+    private int connectionMaxPendingWriteBytes = -1;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
         doc = "The maximum number of connections per IP. If it exceeds, new connections are rejected."
     )
     private int brokerMaxConnectionsPerIp = 0;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -315,6 +315,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         // the null check is a workaround for #13620
         super(pulsar.getBrokerService() != null ? pulsar.getBrokerService().getKeepAliveIntervalSeconds() : 0,
                 TimeUnit.SECONDS);
+        super.maxPendingWriteBytes = pulsar.getConfig().getConnectionMaxPendingWriteBytes();
         this.service = pulsar.getBrokerService();
         this.schemaService = pulsar.getSchemaRegistryService();
         this.listenerName = listenerName;
@@ -437,13 +438,6 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         // complete possible pending connection check future
         if (connectionCheckInProgress != null && !connectionCheckInProgress.isDone()) {
             connectionCheckInProgress.complete(Optional.of(false));
-        }
-    }
-
-    @Override
-    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
-        if (log.isDebugEnabled()) {
-            log.debug("Channel writability has changed to: {}", ctx.channel().isWritable());
         }
     }
 
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
@@ -22,8 +22,10 @@ import static com.google.common.base.Preconditions.checkArgument;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelOutboundBuffer;
 import io.netty.channel.ChannelOutboundInvoker;
 import io.netty.handler.codec.haproxy.HAProxyMessage;
+import java.util.LinkedList;
 import org.apache.pulsar.common.api.proto.BaseCommand;
 import org.apache.pulsar.common.api.proto.CommandAck;
 import org.apache.pulsar.common.api.proto.CommandAckResponse;
@@ -106,6 +108,50 @@ public abstract class PulsarDecoder extends ChannelInboundHandlerAdapter {
 
     private final BaseCommand cmd = new BaseCommand();
 
+    protected int maxPendingWriteBytes = -1;
+    private boolean isPendingCmd;
+    private final LinkedList<ByteBuf> pendingCmds = new LinkedList<>();
+
+    private boolean getOrUpdatePendingCmdState(ChannelHandlerContext ctx) {
+        if (maxPendingWriteBytes < 0) {
+            return false;
+        }
+        if (isPendingCmd) {
+            return true;
+        }
+        final ChannelOutboundBuffer outboundBuffer = ctx.channel().unsafe().outboundBuffer();
+        final BaseCommand.Type cmdType = cmd.getType();
+        if (!ctx.channel().isWritable() && cmdType != BaseCommand.Type.PONG
+                && outboundBuffer != null && outboundBuffer.totalPendingWriteBytes() > maxPendingWriteBytes) {
+            isPendingCmd = true;
+            log.warn("[{}] is not writable, pending output bytes: {}", this, outboundBuffer.totalPendingWriteBytes());
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
+        if (isPendingCmd && ctx.channel().isWritable()) {
+            isPendingCmd = false;
+            // Maybe there are too many pending commands, eventually causing the second time of switching state to
+            // "pending". So we only loop "queueSize" times, which also guarantees the order of commands handling.
+            int queueSize = pendingCmds.size();
+            log.info("[{}] is writable now, flushing pending commands {}", this, queueSize);
+            for (int i = 0; i < queueSize; i++) {
+                ByteBuf cmd = pendingCmds.pop();
+                // Since "ctx" and "handler" have a one-to-one relationship, such a call is safe. And ctx will handle
+                // the exception.
+                try {
+                    channelRead(ctx, cmd);
+                } catch (Exception e) {
+                    ctx.fireExceptionCaught(e);
+                }
+            }
+        }
+        ctx.fireChannelWritabilityChanged();
+    }
+
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         if (msg instanceof HAProxyMessage) {
@@ -116,6 +162,7 @@ public abstract class PulsarDecoder extends ChannelInboundHandlerAdapter {
         }
         // Get a buffer that contains the full frame
         ByteBuf buffer = (ByteBuf) msg;
+        int originalReaderIndex = buffer.readerIndex();
         try {
             // De-serialize the command
             int cmdSize = (int) buffer.readUnsignedInt();
@@ -125,6 +172,18 @@ public abstract class PulsarDecoder extends ChannelInboundHandlerAdapter {
                 log.debug("[{}] Received cmd {}", ctx.channel(), cmd.getType());
             }
             messageReceived();
+
+            if (getOrUpdatePendingCmdState(ctx)) {
+                buffer.readerIndex(originalReaderIndex);
+                // Since there is a "buffer.release()" in the finally block, we retain once to avoid over-release.
+                buffer.retain();
+                pendingCmds.add(buffer);
+                if (log.isDebugEnabled()) {
+                    log.debug("[{}] Since the connection is pending, append cmd {} into pending queue, queue-size: {}",
+                            ctx.channel(), cmd.getType(), pendingCmds.size());
+                }
+                return;
+            }
 
             switch (cmd.getType()) {
             case PARTITIONED_METADATA:
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerBackPressureTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerBackPressureTest.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.client.api;
+
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.broker.auth.MockedPulsarServiceBaseTest;
+import org.apache.pulsar.client.admin.PulsarAdminException;
+import org.apache.pulsar.client.impl.PulsarClientImpl;
+import org.apache.pulsar.common.api.proto.CommandGetTopicsOfNamespace;
+import org.apache.pulsar.common.naming.NamespaceName;
+import org.testng.Assert;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+@Slf4j
+@Test(groups = "broker-impl")
+public class PatternConsumerBackPressureTest extends MockedPulsarServiceBaseTest {
+
+    @Override
+    @BeforeMethod
+    protected void setup() throws Exception {
+        isTcpLookup = true;
+        conf.setEnableBrokerSideSubscriptionPatternEvaluation(false);
+        super.internalSetup();
+        setupDefaultTenantAndNamespace();
+    }
+
+    @Override
+    @AfterMethod(alwaysRun = true)
+    protected void cleanup() throws Exception {
+        super.internalCleanup();
+    }
+
+    @Override
+    protected void doInitConf() throws Exception {
+        conf.setConnectionMaxPendingWriteBytes(5 * 1024 * 1024);
+    }
+
+    @Test(timeOut = 60 * 1000)
+    public void testBrokerOOM() throws PulsarAdminException, InterruptedException {
+        final int topicCount = 8192;
+        final int requests = 10_000;
+        final String topicName = UUID.randomUUID().toString();
+        admin.topics().createPartitionedTopic(topicName, topicCount);
+        final ExecutorService executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
+
+        final PulsarClientImpl pulsarClientImpl = (PulsarClientImpl) pulsarClient;
+        final AtomicInteger success = new AtomicInteger(0);
+        final CountDownLatch latch = new CountDownLatch(requests);
+        for (int i = 0; i < requests; i++) {
+            executorService.execute(() -> {
+                pulsarClientImpl.getLookup()
+                    .getTopicsUnderNamespace(NamespaceName.get("public", "default"),
+                            CommandGetTopicsOfNamespace.Mode.PERSISTENT, ".*", "")
+                    .whenComplete((result, ex) -> {
+                        if (ex == null) {
+                            success.incrementAndGet();
+                        }
+                        log.info("latch-count: {}, succeed: {}", latch.getCount(), success.get());
+                        latch.countDown();
+                    });
+            });
+        }
+        latch.await();
+        Assert.assertEquals(success.get(), requests);
+    }
+}


lhotari (changes requested):
Pulsar Broker already contains multiple solutions for backpressure. 

The channel writability is controlled with https://netty.io/4.1/api/io/netty/channel/WriteBufferWaterMark.html settings which we don't currently expose, since Pulsar hasn't been using writability for backpressure (excluding the Proxy).

For consumers, there's already a backpressure solution for outbound entries with `managedLedgerMaxReadsInFlightSizeInMB`. 
In this case, based on the test case, this is specifically about returning topic pattern listing responses.

The problem with this PR is that the limit is per connection to the broker. That's not very effective in limiting the broker memory usage. In Netty, the standard way to handle this is to change autoread to false when writability changes to false. However, that could cause performance issues for Pulsar broker.

Instead, a more effective solution would be to have a rate limiter or semaphore in place that specifically targets the topic patter listing responses and watchers. When this is handled asynchronously, the solution could add delay to responding to requests when limits are reached.

Besides GET_TOPICS_OF_NAMESPACE/GET_TOPICS_OF_NAMESPACE_RESPONSE, there's also WATCH_TOPIC_LIST with WATCH_TOPIC_LIST_SUCCESS and WATCH_TOPIC_UPDATE responses. That's why it makes more sense to design a proper backpressure fore topic listings.

poorbarcode:
@lhotari 

> The problem with this PR is that the limit is per connection to the broker. That's not very effective in limiting the broker memory usage. In Netty, the standard way to handle this is to change autoread to false when writability changes to false. However, that could cause performance issues for Pulsar broker.

Changed the implementation to setting channel auto-read to `fasle`, but the drawback is that it also affects connection health check.

poorbarcode:
@lhotari 

> The channel writability is controlled with https://netty.io/4.1/api/io/netty/channel/WriteBufferWaterMark.html settings which we don't currently expose, since Pulsar hasn't been using writability for backpressure (excluding the Proxy).
> For consumers, there's already a backpressure solution for outbound entries with managedLedgerMaxReadsInFlightSizeInMB.
In this case, based on the test case, this is specifically about returning topic pattern listing responses.

Added a motivation section to the Motivation

> This fix is very meaningful for the following scenario
> - thousands of topics per namespace
> - thousands of regexp consumers per namespace, and they have different regexp expressions
> - most consumers start at the same time, which will cause huge request of getting topics of namespaces

poorbarcode:
> Instead, a more effective solution would be to have a rate limiter or semaphore in place that specifically targets the topic patter listing responses and watchers. When this is handled asynchronously, the solution could add delay to responding to requests when limits are reached.
> Besides GET_TOPICS_OF_NAMESPACE/GET_TOPICS_OF_NAMESPACE_RESPONSE, there's also WATCH_TOPIC_LIST with WATCH_TOPIC_LIST_SUCCESS and WATCH_TOPIC_UPDATE responses. That's why it makes more sense to design a proper backpressure fore topic listings.

For the scenario that I described [here](https://github.com/apache/pulsar/pull/24423#issuecomment-2982735382), for different namespaces, which have different quantities of topics, we can hardly set an appropriate value for the rate-limiter. So the current implementation is better.

dao-jun:
The root cause is similar with https://github.com/apache/bookkeeper/pull/4556. I think maybe we can consider close the channel if it has too many pending buffer to write.

dao-jun:
Agree with lari, I think the current impl can not solve the root cause.
If there are mannnnnny connections connected to the broker, even though we set `connectionMaxPendingWriteBytes` to a small value, the broker can also OOM

lhotari:
> The root cause is similar with [apache/bookkeeper#4556](https://github.com/apache/bookkeeper/pull/4556). I think maybe we can consider close the channel if it has too many pending buffer to write.

I don't think that there's a need to close the channel. For Pulsar brokers, the memory is already more or less bounded with `maxMessagePublishBufferSizeInMB`, `managedLedgerMaxReadsInFlightSizeInMB` and `managedLedgerCacheSizeMB` settings. The only major gap seems to be the GET_TOPICS_OF_NAMESPACE/GET_TOPICS_OF_NAMESPACE_RESPONSE and  WATCH_TOPIC_LIST/WATCH_TOPIC_LIST_SUCCESS & WATCH_TOPIC_UPDATE handling.

lhotari:
Regarding the TopicListWatcher, it will consume a significant amount of heap memory if the topic names aren't deduplicated String instances: https://github.com/apache/pulsar/blob/3bae1d1648edcedd03973476ef281715499a549e/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java#L47
Guava's Interner class is a good way to deduplicate String instances. Example from Gradle source code: https://github.com/gradle/gradle/blob/master/platforms/core-runtime/base-services/src/main/java/org/gradle/api/internal/cache/StringInterner.java . The usage of `java.lang.String.intern` could be problematic since the GC of the instances isn't well defined.

dao-jun:
> > The root cause is similar with [apache/bookkeeper#4556](https://github.com/apache/bookkeeper/pull/4556). I think maybe we can consider close the channel if it has too many pending buffer to write.
> 
> I don't think that there's a need to close the channel. For Pulsar brokers, the memory is already more or less bounded with `maxMessagePublishBufferSizeInMB`, `managedLedgerMaxReadsInFlightSizeInMB` and `managedLedgerCacheSizeMB` settings. The only major gap seems to be the GET_TOPICS_OF_NAMESPACE/GET_TOPICS_OF_NAMESPACE_RESPONSE and WATCH_TOPIC_LIST/WATCH_TOPIC_LIST_SUCCESS & WATCH_TOPIC_UPDATE handling.

Yes, but maybe we should call `cnx.completedSendOperation` after response send finished.
like here 
image

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
```diff
@@ -106,6 +107,36 @@ public abstract class PulsarDecoder extends ChannelInboundHandlerAdapter {
...
+    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
+        // To avoid the effect of producer limitation, which also will disable channel auto-read, we check
+        // "pauseAutoReadDueToChannelWriteBufFull" here.
+        if (ctx.channel().isWritable() && pauseAutoReadDueToChannelWriteBufFull) {
+            log.warn("[{}] is not writable, turn on channel auto-read", this);
+            pauseAutoReadDueToChannelWriteBufFull = false;
+            ctx.channel().config().setAutoRead(true);
+        }
+        ctx.fireChannelWritabilityChanged();
+    }
```
lhotari: This solution uses an incorrect way for
backpressuring based on outboundBuffer's totalPendingWriteBytes.
Instead of having a new setting `connectionMaxPendingWriteBytes` the
correct way would be to expose and configure WriteBufferWaterMark high
and low settings for the child channel
(`ChannelOption.WRITE_BUFFER_WATER_MARK`) and rely on
`channelWritabilityChanged` and use `ServerCnxThrottleTracker`'s
`incrementThrottleCount`/`decrementThrottleCount` there.

lhotari (changes requested):
I'd be fine with this solution if it would be based on
`ChannelOption.WRITE_BUFFER_WATER_MARK` settings and handle autoread
toggling with `ServerCnxThrottleTracker`'s
`incrementThrottleCount`/`decrementThrottleCount` in the
`channelWritabilityChanged` callback method. There should be a separate
setting to enable it, since configuring `WriteBufferWaterMark` high and
low values are separate from the throttling based on channel
writability. This solution would be useful with suitable setting, but it wouldn't be the correct way to fix the actual problem of broker OOM.
More details in email response, https://lists.apache.org/thread/2nknhw2rbspjbb4y2ocsgqh15yj7lzyh (previous one: https://lists.apache.org/thread/2qf12h5nh1m0n6y4t7szxz061dljw90f).

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
poorbarcode: Answered in mail list

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -437,11 +449,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+            ctx.channel().config().setAutoRead(false);
+            // Resume after 1 second.
+            ctx.channel().eventLoop().schedule(() -> {
+                if (pausedDueToRateLimitation) {
+                    log.info("[{}] Resuming connection after rate limitation", this);
+                    ctx.channel().config().setAutoRead(true);
+                    pausedDueToRateLimitation = false;
+                }
+            }, 1, TimeUnit.SECONDS);
+        }
```
lhotari: In my PIP-434 comments, I've explained that auto read shouldn't be toggled directly. You must use  ServerCnxThrottleTracker instead.
The channel must be throttled if any of the conditions for throttling are present.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -437,11 +449,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
     public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
-        if (log.isDebugEnabled()) {
-            log.debug("Channel writability has changed to: {}", ctx.channel().isWritable());
+        if (pauseReceivingRequestsIfUnwritable && ctx.channel().isWritable()) {
+            log.info("[{}] is writable, turn on channel auto-read", this);
+            ctx.channel().config().setAutoRead(true);
+            requestRateLimiter.timingOpen(rateLimitingSecondsAfterResumeFromUnreadable, TimeUnit.SECONDS);
+        } else if (pauseReceivingRequestsIfUnwritable && !ctx.channel().isWritable()) {
+            log.info("[{}] is not writable, turn off channel auto-read", this);
+            ctx.channel().config().setAutoRead(false);
```
lhotari, line 510: In my PIP-434 comments, I've explained that auto read shouldn't be toggled directly. You must use  ServerCnxThrottleTracker instead.
The channel must be throttled if any of the conditions for throttling are present.

lhotari (changes requested):
In my PIP-434 comments, I've explained that auto read shouldn't be toggled directly. You must use  ServerCnxThrottleTracker instead.
The channel must be throttled if any of the conditions for throttling are present, otherwise existing backpressure would break for rate limits or publish limits.
Please address this in the PIP-434 design doc: https://github.com/apache/pulsar/pull/24510#discussion_r2262580121

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Since `ServerCnxThrottleTracker` was introduced with `pulsar-3.2`, to guarantee the PR can be cherry-picked more easily, I will submit a separate PR to do this

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Since ServerCnxThrottleTracker was introduced with pulsar-3.2, to guarantee the PR can be cherry-picked more easily, I will submit a separate PR to do this

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: This PR is targeting master branch. You should use `ServerCnxThrottleTracker` in master branch PRs. A separate backport can be made for Pulsar 3.0 if this PIP is also for Pulsar 3.0.x.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: This PR is targeting master branch. You should use `ServerCnxThrottleTracker` in master branch PRs. A separate backport can be made for Pulsar 3.0 if this PIP is also for Pulsar 3.0.x.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Added document here https://github.com/apache/pulsar/pull/24510/files#diff-b2fc1ccd29f3f3e1429b9b39c13808f7a85e67d27f2c332189ecbc4c3356e49aR55-R57

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -437,11 +449,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         }
     }
 
+    protected void checkRateLimit(BaseCommand cmd) {
+        if (cmd.getType() == BaseCommand.Type.PONG && cmd.getType() == BaseCommand.Type.PING) {
+            return;
+        }
+        if (requestRateLimiter.acquire(1) == 0) {
```
gaoran10: Could we also check `pausedDueToRateLimitation`? If there are too many requests, numerous warning logs and scheduled tasks will be generated.

poorbarcode: Great suggestion, I have modified it

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,46 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    private int pulsarChannelRateLimitingRateAfterResumeFromUnreadable = 1000;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+            doc = "After the connection is recovered from an unreadable state, the channel will be rate-limited for a"
+                    + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines how"
+                    + " long the rate limiting should last, in seconds. Once the bytes that are waiting to be sent out"
+                    + " reach the \"pulsarChannelWriteBufferHighWaterMark\"， the timer will be reset."
+    )
+    private int pulsarChannelRateLimitingSecondsAfterResumeFromUnreadable = 5;
```
lhotari: There should be a way to disable this rate limiting while throttling ("pause receiving") is in use. Please document the values to disable it.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -437,11 +449,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
 
+    protected void checkRateLimit(BaseCommand cmd) {
+        if (cmd.getType() == BaseCommand.Type.PONG && cmd.getType() == BaseCommand.Type.PING) {
+            return;
+        }
+        if (requestRateLimiter.acquire(1) == 0 && !pausedDueToRateLimitation) {
+            log.warn("[{}] Reached rate limitation", this);
+            // Stop receiving requests.
+            pausedDueToRateLimitation = true;
+            ctx.channel().config().setAutoRead(false);
```
lhotari, line 480: The channel's auto read shouldn't be toggled directly. You must use `org.apache.pulsar.broker.service.ServerCnxThrottleTracker#incrementThrottleCount`/`org.apache.pulsar.broker.service.ServerCnxThrottleTracker#decrementThrottleCount` for toggling or modify `ServerCnxThrottleTracker` to have it's own boolean property for this condition (`publishBufferLimiting` is an example of this, `org.apache.pulsar.broker.service.ServerCnxThrottleTracker#setPublishBufferLimiting`).
If you are planning to backport to `branch-3.0`, you can handle those changes while you make the backport to `branch-3.0`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -437,11 +449,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+        }
+    }
+
     @Override
     public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
-        if (log.isDebugEnabled()) {
-            log.debug("Channel writability has changed to: {}", ctx.channel().isWritable());
+        if (pauseReceivingRequestsIfUnwritable && ctx.channel().isWritable()) {
+            log.info("[{}] is writable, turn on channel auto-read", this);
+            ctx.channel().config().setAutoRead(true);
```
lhotari, line 496: Please see previous comment about using ServerCnxThrottleTracker

lhotari (changes requested):
The channel's auto read shouldn't be toggled directly since PIP-322 has been introduced in Pulsar 3.2.

You must use `org.apache.pulsar.broker.service.ServerCnxThrottleTracker#incrementThrottleCount`/`org.apache.pulsar.broker.service.ServerCnxThrottleTracker#decrementThrottleCount` for toggling or modify `ServerCnxThrottleTracker` to have it's own boolean property for this new condition (`publishBufferLimiting` is an example of this, `org.apache.pulsar.broker.service.ServerCnxThrottleTracker#setPublishBufferLimiting`).

This is a very minor change to the PR and this guidance is easy to follow.
If you are planning to backport to `branch-3.0`, you can handle those changes while you make the backport to `branch-3.0`. It's easy to switch to use toggling of autoread directly in that case, especially if you have a single method where you switch "pausing" on and off for this PIP-434 implementation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: I will push a separate PR to use `ServerCnxThrottleTracker`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: I will push a separate PR to use `ServerCnxThrottleTracker`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: Optimally each PR should keep the project in releasable state. That's why I don't think it's acceptable to break the rate limiting in this PR and then fix it in another. 
Please handle it in this PR since this targets the master branch. I shared in my previous comment a suggestion of how you can handle backporting since that seems to be your reason for ignoring my review guidance.

> This is a very minor change to the PR and this guidance is easy to follow.
If you are planning to backport to branch-3.0, you can handle those changes while you make the backport to branch-3.0. It's easy to switch to use toggling of autoread directly in that case, especially if you have a single method where you switch "pausing" on and off for this PIP-434 implementation.

Is it hard to follow this advice?

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Added

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,47 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    private boolean pulsarChannelPauseReceivingRequestsIfUnwritable = false;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from an unreadable state, the channel will be rate-limited for a"
+                + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines how"
+                + " many requests should be allowed in the rate limiting period."
+
+    )
+    private int pulsarChannelRateLimitingRateAfterResumeFromUnreadable = 1000;
```
codelipenghui:
```suggestion
    private int pulsarChannelRateLimitingRequestsAfterResumeFromUnreadable = 1000;
```

Is this a better name? Just checked based on the description.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: > Optimally each PR should keep the project in releasable state. That's why I don't think it's acceptable to break the rate limiting in this PR and then fix it in another.

Yes, after the current PR, there is no bug. Which is releaable, it is just a state that a PIP is in progress, but every function is fine.

BTW, when you pushed [the part-1 of PIP-379](https://github.com/apache/pulsar/pull/23352), you left bugs there and just created issues to trace them, which is not releasable
- https://github.com/apache/pulsar/pull/23352#discussion_r1792098771
- https://github.com/apache/pulsar/pull/23352#discussion_r1792099326

> Is it hard to follow this advice?

I had said that I would push a separate PR to do it.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Renamed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: And I have said I will push 2 independent PRs to finish under the PIP [here](https://github.com/apache/pulsar/pull/24510#discussion_r2331839209), you just marked the comment as solved, and never argued it.

lhotari: Fair enough, I don't want to argue here. My opinion remains that in this case there isn't a need for 2 separate PRs and that this PR should target master branch and follow the architecture that is present after [PIP-322](https://github.com/apache/pulsar/blob/master/pip/pip-322.md) for controlling the auto read state of ServerCtx inbound channel. https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
is the class that controls auto read state.

lhotari: Feel free to mark this comment resolved.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari, line 1: Please add all new settings to `conf/broker.conf` and `conf/standalone.conf` too so that they are visible and wouldn't need the `PULSAR_PREFIX_` for setting them in k8s deployments.

pulsar-broker/src/main/java/org/apache/pulsar/utils/TimedSingleThreadRateLimiter.java
lhotari, line 1: There shouldn't be a need to put this directly in `pulsar-common`. I'd recommend keeping this in `pulsar-broker`. If it's needed in other server side components (such as `pulsar-proxy`), `pulsar-broker-common` should be preferred over `pulsar-common`. The reason for this is that `pulsar-common` is meant for classes that are shared with the Pulsar client too. In Pulsar 3.0.x, that is Java 8 only.

Please move to `pulsar-broker` for now and if there's a need to reuse, it can be moved later.

Are there any tests for this class?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -438,11 +450,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         }
     }
 
+    protected void checkRateLimit(BaseCommand cmd) {
+        if (cmd.getType() == BaseCommand.Type.PONG && cmd.getType() == BaseCommand.Type.PING) {
+            return;
+        }
+        if (requestRateLimiter.acquire(1) == 0 && !pausedDueToRateLimitation) {
```
lhotari, line 476: It doesn't look like there's a way to disable this completely. Please ensure that this is a no-op when this feature is disabled.

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
```diff
@@ -121,6 +121,8 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
             int cmdSize = (int) buffer.readUnsignedInt();
             cmd.parseFrom(buffer, cmdSize);
 
+            checkRateLimit(cmd);
```
lhotari: this method `checkRateLimit` is very specific in an abstract class. One possible solution would be to remove this here and instead change the `messageReceived` method to accept `cmd` as a parameter. The default implementation would just delegate to `messageReceived()` without parameters. This would keep the abstract class clean while serving the changes you are intended to make.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: I will add them with the third PR of this PIP, which leads to easier to review

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Setting with a negative value means disabling, I have improved the doc.

pulsar-broker/src/main/java/org/apache/pulsar/utils/TimedSingleThreadRateLimiter.java
poorbarcode: - Moved
- Added a test

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
poorbarcode: Good suggestion, improved

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: It doesn't look like this would be a no-op at the moment. I'd expect that the condition includes `rateLimitingSecondsAfterResumeFromUnreadable > 0 &&`

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,47 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+        doc = "If enabled, the channel will pause to deal with new request once the writer buffer is full, until it"
+                + " is changed to writable."
+    )
+    private boolean pulsarChannelPauseReceivingRequestsIfUnwritable = false;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from an unreadable state, the channel will be rate-limited for a"
+                + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines how"
+                + " many requests should be allowed in the rate limiting period."
```
lhotari: "in the rate limiting period" doesn't seem to be accurate here since it's fixed to 1 second in the implementation. It seems that it should be mentioned that the unit is requests per second.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -438,11 +451,53 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         }
     }
 
+    private void checkRateLimit(BaseCommand cmd) {
```
lhotari: rename this method to be more descriptive so that the name would make it obvious that this is related to the `pulsarChannelPauseReceivingRequestsIfUnwritable` feature's `pulsarChannelRateLimitingRequestsAfterResumeFromUnreadable`.

A more descriptive name would be `checkPauseReceivingRequestsAfterResumeRateLimit`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -438,11 +451,53 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+        if (cmd.getType() == BaseCommand.Type.PONG && cmd.getType() == BaseCommand.Type.PING) {
```
lhotari: fix `&&` -> `||` and also return if `!pauseReceivingRequestsIfUnwritable` or `rateLimitingSecondsAfterResumeFromUnreadable <= 0`
```suggestion
        if (!pauseReceivingRequestsIfUnwritable || rateLimitingSecondsAfterResumeFromUnreadable <= 0 || cmd.getType() == BaseCommand.Type.PONG || cmd.getType() == BaseCommand.Type.PING) {
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: I added another comment earlier in this method where the method would return if `rateLimitingSecondsAfterResumeFromUnreadable <= 0`

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,47 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from an unreadable state, the channel will be rate-limited for a"
+            + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines how"
+            + " long the rate limiting should last, in seconds. Once the bytes that are waiting to be sent out"
+            + " reach the \"pulsarChannelWriteBufferHighWaterMark\"， the timer will be reset. Setting a negative"
+            + " value will disable the rate limiting."
+    )
+    private int pulsarChannelRateLimitingSecondsAfterResumeFromUnreadable = 5;
```
lhotari: milliseconds as an unit would provide better accuracy for later tuning. seconds might be too granular.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari: I just hope that this doesn't result in cherry-picking parts separately since that would break functionality in at least branch-3.3, branch-4.0 and branch-4.1 if all parts of the implementation aren't completed and cherry-picked together once they are completed and properly addressing for example the autoread detail that I have so many times reminded about and which you are aware of. Since releases 4.1.1, 4.0.7 and 3.3.9 are about to start, please don't cherry-pick partial implementations to those branches along the way. In branch-3.0 the rate limiting is broken in any case since there's no coordination across autoread toggling so it doesn't really matter for branch-3.0. As long as the feature is disabled by default and a no-op, it's fine for branch-3.0 even if there wouldn't be all parts of the implementation when it lands there.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Improved at https://github.com/apache/pulsar/pull/24423#discussion_r2371536247

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: > "in the rate limiting period" doesn't seem to be accurate here since it's fixed to 1 second in the implementation. It seems that it should be mentioned that the unit is requests per second.

Improved here https://github.com/apache/pulsar/pull/24423#discussion_r2371536087

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Renamed

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: - `pulsarChannelRateLimitingSecondsAfterResumeFromUnreadable` means how long the rate limiter will work for
- Added a new parameter to define the period: `pulsarChannelRateLimitingRequestsPeriodAfterResumeFromUnreadable`

---

Renamed the `3` configurations
- `pulsarChannelRateLimitingSecondsAfterResumeFromUnreadable` -> `pulsarChannelUnReadableCooldownSeconds`
- `pulsarChannelRateLimitingRequestsAfterResumeFromUnreadable` -> `pulsarChannelUnReadableCooldownLimitRate`
- `pulsarChannelRateLimitingRequestsPeriodAfterResumeFromUnreadable` -> `pulsarChannelUnReadableCooldownRateLimitPeriod`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Improved

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,56 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    private int pulsarChannelUnReadableCooldownLimitRate = 5;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from an unreadable state, the channel will be rate-limited for a"
+            + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines the"
+            + " period in milliseconds."
+
+    )
+    private int pulsarChannelUnReadableCooldownRateLimitPeriod = 10;
```
lhotari: there's now multiple concepts for the same concept. "pause receiving" is called "unreadable" and there's "CooldownLimitRate" and "CooldownRateLimitPeriod". Would it be possible to make the concepts more aligned? 

The documentation there is says "after the connection is recovered from an unreadable state" is confusing. It's better to keep the concepts around "pause receiving", since that's being referenced before.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,56 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+        doc = "It relates to configuration \"WriteBufferLowWaterMark\" of Netty Channel Config. If the number of bytes"
+                + " queued in the write buffer is smaller than this value, channel writable state will start to return"
+                + " \"true\"."
+    )
+    private int pulsarChannelWriteBufferLowWaterMark = 32 * 1024;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "If enabled, the channel will pause to deal with new request once the writer buffer is full, until it"
+                + " is changed to writable."
```
lhotari: "the channel will pause" isn't accurate. It's better to be more specific about this. "the broker will pause reading from the channel" etc.

Passing this feedback with relevant context to LLM could help come up with more clear definitions for these configuration options.

poorbarcode: Modified

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Modified

- `pulsarChannelUnReadableCooldownSeconds` -> `pulsarChannelPauseReceivingCooldownSeconds`
- `pulsarChannelUnReadableCooldownLimitRate ` -> `pulsarChannelPauseReceivingCooldownLimitRate`
- `pulsarChannelUnReadableCooldownRateLimitPeriod` -> `pulsarChannelPauseReceivingCooldownRateLimitPeriod`
->  "after the connection is recovered from an unreadable state"  -> after the connection is recovered from a pause receiving state

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -438,11 +452,56 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         }
     }
 
+    private void checkPauseReceivingRequestsAfterResumeRateLimit(BaseCommand cmd) {
+        if (rateLimitingSecondsAfterResumeFromUnreadable <= 0 || cmd.getType() == BaseCommand.Type.PONG
```
lhotari: The later comment about "Please ensure that this is a no-op when this feature is disabled" hasn't been addressed yet. I would assume that it could be addressed by adding  `!pulsarChannelPauseReceivingRequestsIfUnwritable ||` here. 
```suggestion
        if (!pulsarChannelPauseReceivingRequestsIfUnwritable || rateLimitingSecondsAfterResumeFromUnreadable <= 0 || cmd.getType() == BaseCommand.Type.PONG
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: The comment about "Please ensure that this is a no-op when this feature is disabled" hasn't been addressed yet. I also added a new comment about 15 lines above.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,56 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+            + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines how"
+            + " many requests should be allowed in the rate limiting period."
+
+    )
+    private int pulsarChannelPauseReceivingCooldownLimitRate = 5;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from a pause receiving state, the channel will be rate-limited for a"
+            + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines the"
```
lhotari: "period of time" -> "period of time defined by pulsarChannelPauseReceivingCooldownSeconds

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,56 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+            + " period of the rate limiter in milliseconds."
```
lhotari: "period of the rate limiter" remains a concept that is hard to understand. There should be a way to explain it. One possibility is to provide an example to clarify it. "If the rate limit period is set to 1000, then the unit is requests per second. When it's 10, the unit is requests per every 10ms"

Cross referencing the "*RateLimitPermits" (I suggested a rename) config could also be useful since it's a "pair" for this config.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,56 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    private int pulsarChannelPauseReceivingCooldownSeconds = 5;
+
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from a pause receiving state, the channel will be rate-limited for a"
+            + " period of time to avoid overwhelming due to the backlog of requests. This parameter defines how"
+            + " many requests should be allowed in the rate limiting period."
+
+    )
+    private int pulsarChannelPauseReceivingCooldownLimitRate = 5;
```
lhotari: Why is this called "*LimitRate" and the matching pair of this is "*RateLimitPeriod"?
"*RateLimitPermits" and "*RateLimitPeriod" could possibly be more descriptive.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,56 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+
+    @FieldContext(
+            category = CATEGORY_POLICIES,
+            doc = "After the connection is recovered from an pause receiving state, the channel will be rate-limited"
+                + " for a of time window to avoid overwhelming due to the backlog of requests. This parameter defines"
+                + " how long the rate limiting should last, in seconds. Once the bytes that are waiting to be sent out"
+                + " reach the \"pulsarChannelWriteBufferHighWaterMark\"， the timer will be reset. Setting a negative"
+                + " value will disable the rate limiting."
+    )
+    private int pulsarChannelPauseReceivingCooldownSeconds = 5;
```
lhotari: I think I added a comment about the granularity of this setting, but I no longer see the previous comment due to changes. seconds might be too granular for tuning purposes. using milliseconds from the beginning would resolve that.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -438,11 +452,56 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+            ctx.channel().config().setAutoRead(true);
+            requestRateLimiter.timingOpen(rateLimitingSecondsAfterResumeFromUnreadable, TimeUnit.SECONDS);
+        } else if (pauseReceivingRequestsIfUnwritable && !ctx.channel().isWritable()) {
+            final ChannelOutboundBuffer outboundBuffer = ctx.channel().unsafe().outboundBuffer();
+            if (outboundBuffer != null) {
+                log.warn("[{}] is not writable, turn off channel auto-read, totalPendingWriteBytes: {}",
+                        this, outboundBuffer.totalPendingWriteBytes());
+            } else {
+                log.warn("[{}] is not writable, turn off channel auto-read", this);
+            }
```
lhotari, line 509: I think that this will cause very verbose logging. 
Channel writability changes are completely normal in a Netty application. No warning should be printed when the writability changes since it's not an error. There will always be cases where the broker will be filling the outbound buffer with very high amounts until there's a different type of flow control for dispatching to consumers. The dispatcher doesn't currently care if the channel is writable or not and that's also why Pulsar is really fast at the moment. It will buffer data and it can utilize the network very efficiently because of this. The flow control is currently based on `managedLedgerMaxReadsInFlightSizeInMB`. The main weakness is that the solution doesn't have a concept of "fairness". A single slow consumer can hog the resources even when it's not making progress. That's why writability should also be used in dispatching, but the high water mark / low water marks should be relatively high so that sufficient buffering is happening. Since water marks could be adjusted separately for each channel instance, it should possibly be later some sort of policy setting so that client connections that require high throughput could be configured with higher buffering.
This is why I think that adding PIP-434 is useful and we will learn the impacts of it along the way when we start experimenting how it behaves in practice. At this moment, I don't believe that the cooldown period is useful since channel writability can be switching back and forth constantly in a completely healthy situation. Flow control in Netty is like that when the server is pushing a lot of data (and the client cannot keep up).

poorbarcode: Changed the log level to `info`(because it is really helpful for troubleshooting). If it really prints too many, I will push a PR to change it to the `debug` level

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Improved

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Renamed

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Improved

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Improved

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Improved

BTW, the `requestRateLimiter` will skip to limit requests if `!pulsarChannelPauseReceivingRequestsIfUnwritable || rateLimitingSecondsAfterResumeFromUnreadable <= 0`, this place is a repeated check

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: > Changed the log level to info(because it is really helpful for troubleshooting).

As explained above, there is no issue when the channel switches to unwritable. There's nothing to troubleshoot. It will be flooding the logs, especially with the current default `pulsarChannelWriteBufferHighWaterMark` of 64kB. It doesn't make much sense to keep it at info level.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,58 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+
+    @FieldContext(
+            category = CATEGORY_POLICIES,
+            doc = "After the connection is recovered from an pause receiving state, the channel will be rate-limited"
+                + " for a of time window to avoid overwhelming due to the backlog of requests. This parameter defines"
+                + " how long the rate limiting should last, in seconds. Once the bytes that are waiting to be sent out"
+                + " reach the \"pulsarChannelWriteBufferHighWaterMark\"， the timer will be reset. Setting a negative"
+                + " value will disable the rate limiting."
+    )
+    private int pulsarChannelPauseReceivingCooldownMilliSeconds = 5000;
```
lhotari: the comment has "in seconds". Please update that.

For most existing configuration settings using milliseconds, we use the `Ms` or `Millis` suffix, so it would be useful to make this also consistent with that naming.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: Thanks for resolving. I don't think it's a repeated check since it's an internal implementation detail of the ratelimiter. It's not completely a no-op since it calls `System.currentTimeMillis()`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: * change log level to debug and add a `log.isDebugEnabled()` guard
* if there's a need to troubleshoot in some environment, that could be handled with logging configuration with a unique logger name and enabling debug logging for just that logger. It's possible a sub-loggers with the name `org.apache.pulsar.broker.service.ServerCnx.pauseReceiving` so that debug logs for that feature could be enabled separately from ServerCnx debug logs.

lhotari: example of the sub-logger: `private static final Logger PAUSE_RECEIVING_LOG = LoggerFactory.getLogger(ServerCnx.class.getName() + ".pauseReceiving");`

poorbarcode: Improved, thanks for your suggestion

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Corrected the doc.

lhotari:
@poorbarcode checkstyle:
```
Error:  src/main/java/org/apache/pulsar/broker/service/ServerCnx.java:[506] (sizes) LineLength: Line is longer than 120 characters (found 125).
```
A quick way to run checkstyle before pushing is `mvn -T 1C initialize checkstyle:check`

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -919,6 +919,58 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        doc = "After the connection is recovered from a pause receiving state, the channel will be rate-limited for a"
+            + " period of time defined by pulsarChannelPauseReceivingCooldownSeconds to avoid overwhelming due to the"
+            + " backlog of requests. This parameter defines the period of the rate limiter in milliseconds. If the rate"
+            + " limit period is set to 1000, then the unit is requests per second. When it's 10, the unit is requests"
+            + " per every 10ms."
+
+    )
+    private int pulsarChannelPauseReceivingCooldownRateLimitPeriod = 10;
```
lhotari: One detail here. `pulsarChannelPauseReceivingCooldownSeconds` -> `pulsarChannelPauseReceivingCooldownMs`.
In addition, add the `Ms` suffix to make it `pulsarChannelPauseReceivingCooldownRateLimitPeriodMs`.

poorbarcode: Renamed

lhotari (approved):

lhotari:
There's a flaky test, #24827. @poorbarcode Please check.

lhotari:
removing 3.0.x and 3.3.x release labels

PR #24427, opened 2025-06-18
[fix][broker] PIP-428: Fix corrupted topic policies issues with sequential topic policy updates

### LINKED ISSUE #21303: [Bug] Thread safety issues with topic policy mutations
### Search before asking

### Version

Since 2.11

### Minimal reproduce step

Topic policies are stored in system topics and stored as PulsarEvent instances which are serialized using Avro. The PulsarEvent can contain a TopicPoliciesEvent when it's eventType is TOPIC_POLICY. The TopicPoliciesEvent contains the TopicPolicies instance.

The problem is that TopicPolicies instances are cached and these cached instances are mutable in a non-thread safe way.

TopicPolicies handling should be thread safe.

### What did you see instead?

Thread safety concerns aren't covered in TopicPolicies.

### Anything else?

There was a similar issue in mutating namespace policies, #9711. That was fixed with an approach where the shared instances are never mutated. The shared/cached instance is cloned and the clone is mutated.

### LINKED ISSUE #24393: [Bug] Applying Pulsar Topic Policies fails when multiple policies are updated in a short period of time
### Search before reporting

### Read release policy

### User environment

Pulsar Version = 4.0.3
OS = Linux
Java Version =  21
Client Library Version =  4.0.3 (Pulsar-Admin)
Client OS = Darwin (Mac)
Client Java Version = 21

### Issue Description

I have a function to create topics and apply policies on that topic. 
```java
admin.topicPolicies().setInactiveTopicPolicies(pulsarTopic, policies.inactive_topic_policies);

admin.topics().setReplicationClusters(pulsarTopic, List.of("cluster-a", "cluster-b"));

admin.topicPolicies().setSubscriptionTypesEnabledAsync(pulsarTopic, Set.of(SubscriptionType.Shared));

admin.topics().updateProperties(pulsarTopic, policies.properties);

```
However, after applying these, when I verify the results I get mismatching results:
1. replication clusters is null (if I do with "applied"=true, I only get 1 cluster name)
2. subscriptionTypesEnabled = [].

If I introduce a 1 second delay between each of the commands using Thread.sleep(1000), then I don't get these errors.

Also the following order of commands works as well :

```java
admin.topics().setReplicationClusters(pulsarTopic, List.of("cluster-a", "cluster-b"));

admin.topicPolicies().setInactiveTopicPolicies(pulsarTopic, policies.inactive_topic_policies);

admin.topics().updateProperties(pulsarTopic, policies.properties);

admin.topicPolicies().setSubscriptionTypesEnabledAsync(pulsarTopic, Set.of(SubscriptionType.Shared));

```
I believe this might be a result of Async policy application on the broker.

### Error messages

```text
No error messages are observed. Each  POST call succeeds with a 204.
```

### Reproducing the issue

You can run the two method call orderings and verify the results.

### Additional information

_No response_

lhotari:
Fixes #24393
Fixes #21303

### Motivation

There's a long time issue where topic policies get corrupted as a result of sequential or concurrent topic policy updates.
This PR contains the implementation of 
"PIP-428: Change TopicPoliciesService interface to fix consistency issues". Please refer to the PIP document at https://github.com/apache/pulsar/blob/master/pip/pip-428.md for the broader context.

### Modifications

- Check the [PIP-428 document](https://github.com/apache/pulsar/blob/master/pip/pip-428.md) for an overview.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -178,92 +190,248 @@ public CompletableFuture<Void> deleteTopicPoliciesAsync(TopicName topicName,
...
+    /**
+     * This tracks the last handled message IDs for each partition of the topic policies topic and
+     * pending futures for topic policy messages. Each namespace has its own tracker instance since
+     * this is tracking the per-namespace __change_events topic.
+     * The purpose for this tracker is to ensure that write operations on topic policies don't complete before the topic
+     * policies message has been read by the reader and effective.
+     */
+    private static class TopicPolicyMessageHandlerTracker implements AutoCloseable {
+        private List<MessageIdAdv> lastHandledMessageIds = new ArrayList<>();
+        private List<PriorityQueue<PendingMessageFuture>> pendingFutures = new ArrayList<>();
```
BewareMyPower:
```suggestion
        private final List<MessageIdAdv> lastHandledMessageIds = new ArrayList<>();
        private final List<PriorityQueue<PendingMessageFuture>> pendingFutures = new ArrayList<>();
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -178,92 +190,248 @@ public CompletableFuture<Void> deleteTopicPoliciesAsync(TopicName topicName,
...
+            }
+            while (lastHandledMessageIds.size() <= partitionIndex) {
+                lastHandledMessageIds.add(null);
+            }
+            lastHandledMessageIds.set(partitionIndex, messageId);
+            if (pendingFutures.size() > partitionIndex) {
+                PriorityQueue<PendingMessageFuture> pq = pendingFutures.get(partitionIndex);
+                while (!pq.isEmpty() && pq.peek().messageId.compareTo(messageId) <= 0) {
+                    PendingMessageFuture pendingFuture = pq.poll();
+                    pendingFuture.future.complete(null);
```
BewareMyPower: Completing the future in a synchronized block is dangerous because all callbacks registered on the future will be executed when holding the lock.

It would be better to use a separated thread to complete the pending future

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -561,14 +741,16 @@ private void cleanCacheAndCloseReader(@NonNull NamespaceName namespace, boolean
...
         if (closed.get()) {
-            cleanCacheAndCloseReader(reader.getSystemTopic().getTopicName().getNamespaceObject(), false);
+            cleanCacheAndCloseReader(namespaceObject, false);
             return;
         }
         reader.readNextAsync()
                 .thenAccept(msg -> {
                     try {
                         refreshTopicPoliciesCache(msg);
+                        getMessageHandlerTracker(namespaceObject).handleMessageId((MessageIdAdv) msg.getMessageId());
```
BewareMyPower: If an exception is thrown here, `notifyListener` will be skipped.

The exception could be thrown by `future.complete(null)`, which triggers a user-provided callback that is out of the control. In this case, we should not skip `notifyListener`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
BewareMyPower: I've run the `TopicPoliciesUpdateTest` and found the contention happened between `broker-client-shared-internal-executor` (client I/O thread) and `pulsar-io` (broker I/O thread). i.e. when the future is completed in `handleMessageId`, the `pulsar-io` threads might be blocked by the callbacks.

The callbacks can only be registered by `updateTopicPoliciesAsync` and `deleteTopicPoliciesAsync`, which are only. No thread switching is okay, because all these method calls are only called in `PersistentTopicsBase`, where the callback just logs a message and call `asyncResponse.resume`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -178,92 +190,248 @@ public CompletableFuture<Void> deleteTopicPoliciesAsync(TopicName topicName,
...
+                    PulsarEvent event = getPulsarEvent(topicName, actionType, policies, isGlobalPolicy);
+                    String eventKey = getEventKey(event, isGlobalPolicy);
+
+                    if (actionType == ActionType.DELETE) {
+                        return writer.deleteAsync(eventKey, event)
+                                .exceptionally(ex -> {
+                                    log.error("Failed to delete {} topic policy [{}] error.",
+                                            isGlobalPolicy ? "global" : "local", topicName, ex);
+                                    return null;
+                                });
```
BewareMyPower: It seems that here it should not return the future returned by `exceptionally`. If you only want to log for exception, it should be:

```java
                        final var future = writer.deleteAsync(eventKey, event);
                        future.exceptionally(ex -> {
                            log.error("Failed to delete {} topic policy [{}] error.",
                                    isGlobalPolicy ? "global" : "local", topicName, ex);
                            return null;
                        });
```

I see null message ids are handled like:

```java
                                            return sendTopicPolicyEventInternal(topicName, actionType, policiesToUpdate,
                                                    isGlobalPolicy);
                                        })
                                        .thenCompose(messageId -> {
                                            if (messageId == null) {
```

If it's intentional to return a future that completes a null message id when `deleteAsync` failed, why not do the same for `writer.writeAsync`?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -5417,29 +5242,18 @@ protected CompletableFuture<EntryFilters> internalGetEntryFilters(boolean applie
...
-        return getTopicPoliciesAsyncWithRetry(topicName, isGlobal)
-                        .thenCompose(op -> {
-                            if (!op.isPresent()) {
-                                return CompletableFuture.completedFuture(null);
-                            }
-                            op.get().setEntryFilters(null);
-                            op.get().setIsGlobal(isGlobal);
-                            return pulsar().getTopicPoliciesService().updateTopicPoliciesAsync(topicName, op.get());
-                        });
+        return pulsar().getTopicPoliciesService().updateTopicPoliciesAsync(topicName, isGlobal, false, policies -> {
```
BewareMyPower:
```suggestion
        return pulsar().getTopicPoliciesService().updateTopicPoliciesAsync(topicName, isGlobal, true, policies -> {
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -3534,66 +3496,53 @@ protected CompletableFuture<RetentionPolicies> internalGetRetention(boolean appl
...
+                                if (!checkBacklogQuota(backlogQuota, retentionToSet)) {
+                                    log.warn("[{}] Failed to update retention quota configuration for topic {}: "
+                                            + "conflicts with retention quota", clientAppId(), topicName);
+                                    return FutureUtil.failedFuture(new RestException(Status.PRECONDITION_FAILED,
+                                            "Retention Quota must exceed configured backlog quota for topic. "
+                                                    + "Please increase retention quota and retry"));
+                                }
+                            }
+                            return pulsar().getTopicPoliciesService()
+                                    .updateTopicPoliciesAsync(topicName, isGlobal, retentionToSet == null, policies -> {
```
BewareMyPower: `retentionToSet == null` is always true.

The previous logic here is a bit confusing that when `retentionToSet` is null, this method will do nothing rather than deleting the retention policy. This method could only be called by `PersistentTopics#setRetention`, whose `retention` argument should never be null.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
lhotari: Addressed this by both catching exceptions for `future.complete(null)` and putting `notifyListener` in a finally block.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
lhotari: Fixed this to just log it like it was in the original logic.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
lhotari: fixed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
lhotari: fixed.

lhotari:
Thanks for the very helpful review, @BewareMyPower. PTAL

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -107,8 +111,16 @@ protected NamespaceEventsSystemTopicFactory initialize() {
...
     final Map<TopicName, List<TopicPolicyListener>> listeners = new ConcurrentHashMap<>();
 
+    private final Map<NamespaceName, TopicPolicyMessageHandlerTracker> topicPolicyMessageHandlerTrackers =
+            new ConcurrentHashMap<>();
+
     private final AsyncLoadingCache<NamespaceName, SystemTopicClient.Writer<PulsarEvent>> writerCaches;
 
+    // Sequencer for policy updates per topic and per policy type (global/local)
+    // Key: Pair<TopicName, Boolean (isGlobal)>, Value: CompletableFuture<Void> representing the last update in sequence
+    private final ConcurrentHashMap<Pair<TopicName, Boolean>, CompletableFuture<Void>> topicPolicyUpdateSequencer =
```
BewareMyPower, line 121: You can use `BooleanObjectPair<TopicName>` from fastutil instead of `Pair<TopicName, Boolean>`

BewareMyPower (approved):
LGTM, just leave a minor comment

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
lhotari: I guess there wouldn't be such a benefit in using `BooleanObjectPair<TopicName>` (or `ObjectBooleanPair<TopicName>`) since the `java.lang.Boolean` doesn't have the same problem as the number wrappers which would have to create a wrapper for each primitive number.

BewareMyPower: Ah yes, it's right.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -178,92 +190,257 @@ public CompletableFuture<Void> deleteTopicPoliciesAsync(TopicName topicName,
...
-                return writer.deleteAsync(getEventKey(event, false), event);
-            });
-        deletePolicies.exceptionally(ex -> {
-            log.error("Failed to delete topic policy [{}] error.", topicName, ex);
-            return null;
-        });
-        return deletePolicies;
+                                                                      @Nullable TopicPolicies policies,
+                                                                      boolean isGlobalPolicy) {
+        return writerCaches.get(topicName.getNamespaceObject())
```
Technoboy-, line 423: Should check if the writer could be got here and `invalidate` if has exception

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -178,92 +190,257 @@ public CompletableFuture<Void> deleteTopicPoliciesAsync(TopicName topicName,
...
+            // future
+            topicPolicyUpdateSequencer.compute(sequencerKey, (key, existingFuture) -> {
+                if (existingFuture != null && existingFuture.isDone()) {
+                    // Remove the completed future from the sequencer map
+                    return null;
+                }
+                return existingFuture;
+            });
+            if (ex != null) {
+                writerCaches.synchronous().invalidate(topicName.getNamespaceObject());
```
Technoboy-: Remove the writer cache isn't right here, I think it should be https://github.com/apache/pulsar/pull/24427/files#r2297933245

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
lhotari: We should be able to rely on AsyncLoadingCache providing a valid instance

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
lhotari: oh yes, checking again

lhotari: resolved

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
lhotari: Resolved this and added a comment about this.

Technoboy- (approved):

lhotari:
dev mailing list thread to cherry-pick to branch-4.0: https://lists.apache.org/thread/ljnhcm4pgl933kc75ftgz9yzmy03rdqg

nodece:
@lhotari This PR fixes `NullPointerException`:
```
java.lang.NullPointerException: Cannot invoke "java.util.concurrent.atomic.AtomicInteger.incrementAndGet()" because the return value of "java.util.Map.get(Object)" is null

  at org.apache.pulsar.broker.service.SystemTopicBasedTopicPoliciesService.addOwnedNamespaceBundleAsync(SystemTopicBasedTopicPoliciesService.java:405) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at org.apache.pulsar.broker.service.SystemTopicBasedTopicPoliciesService$2.onLoad(SystemTopicBasedTopicPoliciesService.java:509) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at org.apache.pulsar.broker.namespace.NamespaceService.notifyNamespaceBundleOwnershipListener(NamespaceService.java:1322) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at org.apache.pulsar.broker.namespace.NamespaceService.onNamespaceBundleOwned(NamespaceService.java:1285) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at org.apache.pulsar.broker.namespace.OwnershipCache.lambda$tryAcquiringOwnership$3(OwnershipCache.java:218) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:684) ~[?:?]

  at java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:662) ~[?:?]

  at java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:2168) ~[?:?]

  at org.apache.pulsar.broker.namespace.OwnershipCache.tryAcquiringOwnership(OwnershipCache.java:216) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at org.apache.pulsar.broker.namespace.NamespaceService.searchForCandidateBroker(NamespaceService.java:631) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at org.apache.pulsar.broker.namespace.NamespaceService.lambda$findBrokerServiceUrl$10(NamespaceService.java:466) ~[pulsar-broker.jar:3.0.14-SNAPSHOT]

  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) ~[?:?]

  at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]

  at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) ~[?:?]

  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]

  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]

  at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[netty-common-4.1.124.Final.jar:4.1.124.Final]

  at java.lang.Thread.run(Thread.java:840) ~[?:?]
```

Do you consider fixing this on the branch-3.0?

lhotari:
> Do you consider fixing this on the branch-3.0?

@nodece please propose on the mailing list

nodece:
@lhotari, It seems that fixing the `NullPointerException` may be sufficient.

nodece:
I submitted https://github.com/apache/pulsar/pull/24758 to the branch-3.0.

PR #24430, opened 2025-06-19
[improve][broker] Reduce cpu usage of InMemoryDelayedDeliveryTracker.

thetumbled:
### Motivation
As https://lists.apache.org/thread/3dc1c3n7d0tmwm8974qh9w771jzgy6f6 show, method `getNumberOfDelayedMessages` is called frequently and the current implementation of this method is time cosuming, we need to implement this method with a much more efficient approach.

### Modifications

Introduce an `AtomicLong delayedMessagesCount`.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/74

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -31,7 +31,6 @@ import java.util.NavigableSet;
 import java.util.TreeSet;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicLong;
-
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.Position;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -66,7 +66,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     // The bit count to trim to reduce memory occupation.
     private final int timestampPrecisionBitCnt;
 
-    // 添加一个原子类型的计数器
+    // Count of delayed messages in the tracker.
     private final AtomicLong delayedMessagesCount = new AtomicLong(0);
 
     InMemoryDelayedDeliveryTracker(AbstractPersistentDispatcherMultipleConsumers dispatcher, Timer timer,
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -30,6 +30,8 @@ import java.time.Clock;
 import java.util.NavigableSet;
 import java.util.TreeSet;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicLong;
+
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.Position;
@@ -63,6 +65,9 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
 
     // The bit count to trim to reduce memory occupation.
     private final int timestampPrecisionBitCnt;
+
+    // 添加一个原子类型的计数器
+    private final AtomicLong delayedMessagesCount = new AtomicLong(0);
 
     InMemoryDelayedDeliveryTracker(AbstractPersistentDispatcherMultipleConsumers dispatcher, Timer timer,
                                    long tickTimeMillis,
@@ -125,6 +130,8 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
         delayedMessageMap.computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
                 .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap())
                 .add(entryId);
+        delayedMessagesCount.incrementAndGet();
+
         updateTimer();
 
         checkAndUpdateHighest(deliverAt);
@@ -183,6 +190,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
                         positions.add(PositionFactory.create(ledgerId, entryId));
                     });
                     n -= cardinality;
+                    delayedMessagesCount.addAndGet(-cardinality);
                     ledgerIdToDelete.add(ledgerId);
                 } else {
                     long[] entryIdsArray = entryIds.toArray();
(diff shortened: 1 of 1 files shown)


nodece (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -125,6 +130,8 @@ public boolean addMessage(long ledgerId, long entryId, long deliverAt) {
         delayedMessageMap.computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
                 .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap())
                 .add(entryId);
+        delayedMessagesCount.incrementAndGet();
```
lhotari, line 132: What if the same entryId is added multiple times with the same timestamp? Would the counter get out of sync in that case? Would it be possible to add test cases?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -221,14 +230,13 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
     @Override
     public CompletableFuture<Void> clear() {
         this.delayedMessageMap.clear();
+        this.delayedMessagesCount.set(0);
```
lhotari, line 232: On line 220 there's `if (delayedMessageMap.isEmpty()) {`. Would it make sense to add `this.delayedMessagesCount.set(0);` also to that code block?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -221,14 +230,13 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
...
         return CompletableFuture.completedFuture(null);
     }
 
     @Override
     public long getNumberOfDelayedMessages() {
-        return delayedMessageMap.values().stream().mapToLong(
-                ledgerMap -> ledgerMap.values().stream().mapToLong(
-                        Roaring64Bitmap::getLongCardinality).sum()).sum();
+        return delayedMessagesCount.get();
```
lhotari, line 238: Do we already have sufficient test coverage to ensure that the counter provides an accurate value in all cases?

dao-jun (approved):

codelipenghui:
@thetumbled, Thanks for the quick fix.

I will remove the label release/4.0.6 since we only ship bug fixes for the LTS branch. If you still think it's required to cherry-pick the memory improvement to branch-4.0, we can discuss under the mailing list. Sometimes we need to handle something case by case.

thetumbled:
> @thetumbled, Thanks for the quick fix.
> 
> I will remove the label release/4.0.6 since we only ship bug fixes for the LTS branch. If you still think it's required to cherry-pick the memory improvement to branch-4.0, we can discuss under the mailing list. Sometimes we need to handle something case by case.

I'm fine. I'm wondering which branch is the so-called 'feature branch'.

lhotari (commented):
I forgot to submit my review comments that I had made earlier.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: Is there any chance that the same entryId is added multiple times?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: There is no need to do that, these logic is used for feature `fixedDelayDetectionLookahead`.
```
if (delayedMessageMap.isEmpty()) {
            // Reset to initial state
            highestDeliveryTimeTracked = 0;
            messagesHaveFixedDelay = true;
        }
```

When the `delayedMessageMap` is empty, the value of `delayedMessagesCount` must be 0; otherwise, there are other issues that need to be resolved.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: There are some tests that directly call this method:
image

Each time we add a delay message or retrieve a delay message, this method will be called indirectly, which underscores the importance of an efficient approach.
Therefore, I believe there is sufficient test coverage for this method.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
lhotari: I guess it is designed in a way where it shouldn't be possible to be added multiple time.

there is code to call the clear method of the InMemoryDelayedDeliveryTracker:
https://github.com/apache/pulsar/blob/f2618c15bb5a9f3fcb577068584df5d0e2e4f335/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L186-L201

However, it's hard to see if all possible code execution paths are covered. In dispatcher code, `cursor.rewind()` is called in multiple locations. For example:

https://github.com/apache/pulsar/blob/f2618c15bb5a9f3fcb577068584df5d0e2e4f335/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L833

It does look like `.clear()` would be called either before or after `.rewind()`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
lhotari: > When the `delayedMessageMap` is empty, the value of `delayedMessagesCount` must be 0; otherwise, there are other issues that need to be resolved.

That's true. I think it would make sense to add an error log if delayedMessageCount isn't 0 at this point. That would help detect possible issues.

codelipenghui:
> I'm fine. I'm wondering which branch is the so-called 'feature branch'.

@thetumbled Here is the documentation for release policy https://pulsar.apache.org/contribute/release-policy/

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
thetumbled: That is fine. Should i create a new PR to add the debug log?

lhotari: @thetumbled I think it makes sense. It shouldn't be a debug log, but an error log if that ever happens. In addition, I think it should reset the count to 0, so that the state would be fixed.

thetumbled: Hi, Lari. I have created a pr to check this. But i think it should not be set to 0 because of the posibility of concurrent execution. 
Suppose there is another thread that adds a message to the map after the check `if (delayedMessageMap.isEmpty()) {;` in that case, the value should not be zero, correct?
I suggest we had better keep it unchanged and only leave a warning log.

PR #18877, opened 2022-12-11
[fix] [broker] getLastMessageId returns a wrong batch index of last message if enabled read compacted

poorbarcode:
### Motivation

The method `consumer.getLastMessageId` will return the latest message which can be received.
- If disabled `read compacted`, will return the last confirmed position of `ManagedLedger`.
- If enabled `read compacted`, will return the latest message id which can be read from the compacted topic.

If we send a batch message like this:

```java
producer.newMessage().key("k1").value("v0").sendAsync(); // message-id is [3:1,-1:0]
producer.newMessage().key("k1").value("v1").sendAsync(); // message-id is [3:1,-1:1]
producer.newMessage().key("k1").value("v2").sendAsync(); // message-id is [3:1,-1:2]
producer.newMessage().key("k2").value("v0").sendAsync(); // message-id is [3:1,-1:3]
producer.newMessage().key("k2").value("v1").sendAsync(); // message-id is [3:1,-1:4]
producer.newMessage().key("k2").value(null).sendAsync(); // message-id is [3:1,-1:5]
producer.flush();
```

After the compaction task is done, the messages with key `k2` will be deleted by the compaction task. Then the latest message that can be received will be `[3:1:-1:2]`.

--- 
When we call `consumer.getLastMessageId`, the expected result is:

```
[3:1,-1:2]
```

--- 
But the actual result is:

```
[3:1,-1:5]
```

### Modifications
If enabled `read compacted` and the latest entry of the compacted topic is a batched message, extract the entry and calculate all internal messages, then return the latest message which is not marked `compacted out`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: 
- https://github.com/poorbarcode/pulsar/pull/49

PR #24431, opened 2025-06-19
[feat] Support get retained message count from MessageMetadata

gaoran10:
### Motivation

Currently, if the broker wants to determine the valid batch index in a compacted entry, it can only calculate it by examining the compactedOut flag in SingleMessageMetadata. Pulsar supports different protocols, if other protocols doesn't add SingleMessageMetadata for each message in a batch, such as Kafka protocol handler may build the payload in Kafka format.

### Modifications

Provide a way to retrieve the retained message count from MessageMetadata.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


BewareMyPower:
I mean, this PR only adds a static method `calculateTheLastBatchIndexInBatch`, which does not depend on any private method.

I don't see any extra change in `ServerCnx`, did you miss something?

pulsar-broker/src/main/java/org/apache/pulsar/compaction/CompactedTopicUtils.java
```diff
@@ -104,4 +114,60 @@ public static void asyncReadCompactedEntries(TopicCompactionService topicCompact
...
+            return -1;
+        }
+        for (KeyValue kv : metadata.getPropertiesList()) {
+            if (RETAINED_MESSAGE_COUNT_PROPERTY.equals(kv.getKey())) {
+                try {
+                    return Integer.parseInt(kv.getValue());
+                } catch (NumberFormatException e) {
+                    // Ignore and return default value
+                    return -1;
+                }
```
BewareMyPower, line 167: IMO, this hacky behavior should be avoided. It gives a new meaning that if the value of "RetainedMessageCount" property is a valid integer, the value will represent the number of messages in the batch.

BewareMyPower (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2453,33 +2450,6 @@ private void handleLastMessageIdFromCompactionService(PersistentTopic persistent
...
-        SingleMessageMetadata singleMessageMetadata = new SingleMessageMetadata();
-        int lastBatchIndexInBatch = -1;
-        for (int i = 0; i < batchSize; i++){
-            ByteBuf singleMessagePayload =
-                    Commands.deSerializeSingleMessageInBatch(payload, singleMessageMetadata, i, batchSize);
-            singleMessagePayload.release();
-            if (singleMessageMetadata.isCompactedOut()){
-                continue;
-            }
-            lastBatchIndexInBatch = i;
```
BewareMyPower, line 2478: The key point is here assumes the compaction batch should have `SingleMessageMetadata` for each single message. The behavior was introduced from https://github.com/apache/pulsar/pull/18877. However, it requires the message payload can always have a serialized `SingleMessageMetadata` if the batch size is greater than 1, which could be too strict.

This comment also makes sense: https://github.com/apache/pulsar/pull/18877/files#r1165000498

pulsar-broker/src/main/java/org/apache/pulsar/compaction/Compactor.java
```diff
@@ -35,6 +35,7 @@ public abstract class Compactor {
     private static final Logger log = LoggerFactory.getLogger(Compactor.class);
     public static final String COMPACTION_SUBSCRIPTION = "__compaction";
     public static final String COMPACTED_TOPIC_LEDGER_PROPERTY = "CompactedTopicLedger";
+    public static final String RETAINED_MESSAGE_COUNT_PROPERTY = "RetainedMessageCount";
```
codelipenghui, line 38: Is this a concept of Pulsar?
Will the client side set the value for this property?

BewareMyPower: No. The background is complicated. I'm writing a proposal to fix the root cause.

This draft implementation might help you understand: https://github.com/BewareMyPower/pulsar/pull/42/files#diff-f2b074eb97e870ec15676e349d61de41fc94067e2b01e86335d05b4bc4019407

i.e. the customized compaction service could use a property to determine which messages are compacted out rather than checking the `compactedOut` field. The root cause is that each entry is guaranteed to have a `MessageMetadata`, even the message was produced via the protocol handler, but it's not guaranteed to have a `SingleMessageMetadata`.

You can also wait for my proposal (PIP-429) to be available

coderzc:
If we purpose are to ensure that the compaction task is successful, we only need to check `lastCompactedPosition` and `lastDispatchablePosition` beforen trigger compaction task. If `lastDispatchablePosition <= lastCompactedPosition`, then I just need to skip the compact task because there is no new data need to be compacted.

If we need make the `Pulsar reader` to read Kafka format data, then we need this change or Introduce a new client API(PIP-429).

PR #24432, opened 2025-06-19
[fix][ml]Still got BK ledger, even though it has been deleted after offloaded

poorbarcode:
### Motivation

**The reproduction steps of the issue:**
- There are `4` ledgers in the ManagedLedger
- `3` ledgers are offloaded.
- The BK ledgers are deleted because they were offloaded.
- Issue:
  - Actual behavior: `ManagedLedger.getLedgerHandle(long ledgerId)` returns BK ledger handles, which have been deleted
  - Expected behavior: `ManagedLedger.getLedgerHandle(long ledgerId)` returns Offloader ledger handles.

**Root cause**
Trimming ledgers, BK ledgers are deleted, and `LedgerCache` of `ManagedLedger` does not remove the cached ledger handle in memory.

### Modifications

- In scope: Remove cached ledger handles when trimming offloaded ledgers.
- Not in scope: This PR will not solve the issue: Ledger is being deleted, but a cursor still uses the ledger handle to read entries. This issue will be resolved automatically because ManagedLedger will recreate the ledger handle when it receives a failed read entry callback 

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -2980,7 +2980,15 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
                     for (LedgerInfo ls : offloadedLedgersToDelete) {
                         log.info("[{}] Deleting offloaded ledger {} from bookkeeper - size: {}", name, ls.getLedgerId(),
                                 ls.getSize());
-                        asyncDeleteLedgerFromBookKeeper(ls.getLedgerId());
+                        asyncDeleteLedgerFromBookKeeper(ls.getLedgerId()).thenAccept(__ -> {
+                            log.info("[{}] Deleted and invalidated offloaded ledger {} from bookkeeper - size: {}",
+                                    name, ls.getLedgerId(), ls.getSize());
+                            invalidateReadHandle(ls.getLedgerId());
+                        }).exceptionally(ex -> {
+                            log.error("[{}] Failed to delete offloaded ledger {} from bookkeeper - size: {}",
+                                    name, ls.getLedgerId(), ls.getSize(), ex);
+                            return null;
+                        });
                     }
                     promise.complete(null);
                 }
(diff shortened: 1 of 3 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2980,7 +2980,15 @@ public void operationComplete(Void result, Stat stat) {
                     for (LedgerInfo ls : offloadedLedgersToDelete) {
                         log.info("[{}] Deleting offloaded ledger {} from bookkeeper - size: {}", name, ls.getLedgerId(),
                                 ls.getSize());
-                        asyncDeleteLedgerFromBookKeeper(ls.getLedgerId());
+                        asyncDeleteLedgerFromBookKeeper(ls.getLedgerId()).thenAccept(__ -> {
+                            log.info("[{}] Deleted and invalidated offloaded ledger {} from bookkeeper - size: {}",
+                                    name, ls.getLedgerId(), ls.getSize());
+                            invalidateReadHandle(ls.getLedgerId());
```
lhotari: Just wondering if invalidation should happen before deleting ledgers from bookkeeper. Could there be a race condition if the invalidation happens after deletion?

poorbarcode: Ah, I think you are right. Invalidating the ledger handle first is better. I have changed the logic

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -3219,22 +3227,31 @@ private void asyncDeleteLedger(long ledgerId, LedgerInfo info) {
...
             if (isNoSuchLedgerExistsException(rc)) {
                 log.warn("[{}] Ledger was already deleted {}", name, ledgerId);
+                future.complete(null);
             } else if (rc != BKException.Code.OK) {
                 log.error("[{}] Error deleting ledger {} : {}", name, ledgerId, BKException.getMessage(rc));
+                if (retry <= 1) {
+                    // The latest once of retry has failed
+                    log.warn("[{}] Failed to delete ledger after retries {}, code: {}", name, ledgerId, rc);
+                    future.completeExceptionally(BKException.create(rc));
+                }
```
Technoboy-, line 3248: Do we need to add `return` here ?

poorbarcode: Fixed, Thanks ❤️

Technoboy- (approved):

wangchao316 (approved):

PR #24434, opened 2025-06-20
[improve][broker] Improve the log when namespace bundle is not available

BewareMyPower:
### Motivation

The `Namespace bundle for topic (%s) not served by this instance` log does not include the actual bundle info. We can check logs of `OwnershipCache` to see a bundle's ownership. But it's hard to see if the current topic's bundle is owned by the current broker when the error happens. We have to manually compute the hash code and see which bundle this topic belong to.

```
long hashCode = Hashing.crc32().hashString(topicName.toString(), StandardCharsets.UTF_8).padToLong();
System.out.println(String.format("0x%x", hashCode));
```

### Modifications

Return the bundle info in `checkTopicOwnership` as well.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentDispatcherFailoverConsumerTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentDispatcherFailoverConsumerTest.java
@@ -163,7 +163,8 @@ public class PersistentDispatcherFailoverConsumerTest {
         NamespaceService nsSvc = pulsarTestContext.getPulsarService().getNamespaceService();
         doReturn(true).when(nsSvc).isServiceUnitOwned(any(NamespaceBundle.class));
         doReturn(true).when(nsSvc).isServiceUnitActive(any(TopicName.class));
-        doReturn(CompletableFuture.completedFuture(true)).when(nsSvc).checkTopicOwnership(any(TopicName.class));
+        doReturn(CompletableFuture.completedFuture(mock(NamespaceBundle.class))).when(nsSvc).getBundleAsync(any());
+        doReturn(CompletableFuture.completedFuture(true)).when(nsSvc).checkBundleOwnership(any(), any());
 
         setupMLAsyncCallbackMocks();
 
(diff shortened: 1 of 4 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2364,19 +2364,19 @@ public CompletableFuture<Boolean> isTopicNsOwnedByBrokerAsync(TopicName topicNam
...
+        return namespaceService.getBundleAsync(topicName).thenCompose(bundle ->
+                namespaceService.checkBundleOwnership(topicName, bundle).thenCompose(ownedByThisInstance -> {
                     if (ownedByThisInstance) {
                         return CompletableFuture.completedFuture(null);
                     } else {
-                        String msg = String.format("Namespace bundle for topic (%s) not served by this instance:%s. "
-                                        + "Please redo the lookup. Request is denied: namespace=%s",
-                                topic, pulsar.getBrokerId(), topicName.getNamespace());
+                        String msg = String.format("Namespace bundle (%s) for topic (%s) not served by this instance:"
+                                + "%s", bundle, topic, pulsar.getBrokerId());
```
Technoboy-: Do we need to keep the `namespace` here ?

Technoboy- (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
BewareMyPower: No, because namespace is already included in the topic name

poorbarcode (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2364,19 +2364,19 @@ public CompletableFuture<Boolean> isTopicNsOwnedByBrokerAsync(TopicName topicNam
...
 
-        return pulsar.getNamespaceService().checkTopicOwnership(topicName)
-                .thenCompose(ownedByThisInstance -> {
+        return namespaceService.getBundleAsync(topicName).thenCompose(bundle ->
+                namespaceService.checkBundleOwnership(topicName, bundle).thenCompose(ownedByThisInstance -> {
                     if (ownedByThisInstance) {
                         return CompletableFuture.completedFuture(null);
                     } else {
-                        String msg = String.format("Namespace bundle for topic (%s) not served by this instance:%s. "
-                                        + "Please redo the lookup. Request is denied: namespace=%s",
```
lhotari, line 2374: ServiceUnitStateChannelTest is failing since this was removed. 
It might be good to preserve some advice in the warning message. I'm not saying that the previous one is optimal, but it could be useful to have something in the message that it's a transient error and that a new lookup should fix the problem.

PR #24439, opened 2025-06-21
[improve][pip] PIP-429: Optimize Handling of Compacted Last Entry by Skipping Payload Buffer Parsing

BewareMyPower:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-429.md
+++ b/pip/pip-429.md
@@ -111,6 +111,42 @@ The improvements are:
 
 ## Public-facing Changes
 
+### Public API
+
+Remove `readLastCompactedEntry` and `findEntryByEntryIndex` methods from `TopicCompactionService` interface.
+- `findEntryByEntryIndex`: it's never used other in tests
+- `readLastCompactedEntry`: it exposes the `Entry` to the caller, while the caller should only care about the position of the last message.
+
+Change the return value of `findEntryByPublishTime` from `CompletableFuture<Entry>` to `CompletableFuture<Position>`. This change is made because the caller only needs the position of the last message, not the entire entry.
+
+Add the following new methods:
+
+```java
+/**
+ * Retrieve the position of the last message before compaction.
+ *
+ * @return A future that completes with the position of the last message before compaction, or
+ *         {@link MessagePosition#EARLIEST} if no such message exists.
+ */
+CompletableFuture<MessagePosition> getLastMessagePosition();
+
+/**
+ * Represents the position of a message.
+ * <p>
+ * The `ledgerId` and `entryId` together specify the exact entry to which the message belongs. For batched messages,
+ * the `batchIndex` field indicates the index of the message within the batch. If the message is not part of a
+ * batch, the `batchIndex` field is set to -1. The `publishTime` field corresponds to the publishing time of the
+ * entry's metadata, providing a timestamp for when the entry was published.
+ * </p>
+ */
+record MessagePosition(long ledgerId, long entryId, int batchIndex, long publishTime) {
+
+    public static final MessagePosition EARLIEST = new MessagePosition(-1L, -1L, 0, 0);
+}
+```
+
+After this change, the processing of the entry buffer will be handled within the `TopicCompactionService` instead of being managed on the caller side (`ServerCnx`). This adjustment enhances the flexibility of the implementation, enabling more advanced use cases. For example, it allows the position to be stored in an external metadata service, providing greater scalability and modularity.
+
 ### Binary protocol
 
 A new field will be added to `MessageMetadata`:
--- a/pip/pip-429.md
+++ b/pip/pip-429.md
@@ -66,7 +66,7 @@ The `compacted_out` field was first introduced in the early stages of developmen
 
 Decompressing the payload buffer solely to check whether individual messages have the `compacted_out` field set is both inefficient and restrictive, as it imposes constraints on the payload buffer format. Furthermore, when using a custom topic compaction service, the entry buffer in the compacted ledger may not include a `SingleMessageMetadata` for every single message, adding further complexity to the process.
 
-This challenge is exacerbated when messages are encrypted, as decompression is not possible without the public key required for decryption. This limitation also impacts the current compaction service, as encrypted messages cannot be compacted. Consequently, operations such as the `GetLastMessageId` RPC will fail, resulting in an error similar to the following:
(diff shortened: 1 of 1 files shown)


gaoran10 (dismissed):
I prefer to make changes on the server side for several reasons.
- Broker can handle this issue well and provide a compatibility solution.
- Add a flag for the valid compacted position can avoid deserializing and uncompressing the entire batch messages metadata header.
- If making this change at the client-side, it increases the complexity for the client, and all language clients need to change.
- It's weird to return EntryBuffer when getting the last message ID.
- Actually, Pulsar only needs to retain valid message data in a compacted entry, but it retains all compacted messages with  "empty header" and "empty payload".

BewareMyPower:
> If making this change at the client-side, it increases the complexity for the client, and all language clients need to change.

Please check the `Backward & Forward Compatibility` section

> When an older client sends a `GetLastMessageId` request to a newer broker, the broker computes the last message ID using its standard process. To ensure backward compatibility, the broker excludes the payload buffer from the response, aligning with the expectations of older clients.

BewareMyPower:
> Actually, Pulsar only needs to retain valid message data in a compacted entry, but it retains all compacted messages with "empty header" and "empty payload".

Yes. Unfortunately, it's limited by client side's logic that the default logic to parse the payload buffer looks like:

```java
int batchSize = msgMetadata.getNumMessagesInBatch();
for (int i = 0; i < batchSize; i++) {
    int batchIndex = i;
    final var singleMessageMetadata = parse(payload);
    if (singleMessageMetadata.isCompactedOut()) {
        break;
    }
    // Create a message, whose batch index is i, from the payload buffer
}
```

> Add a flag for the valid compacted position can avoid deserializing and uncompressing the entire batch messages metadata header.

This makes sense to me. Adding a new field to `MessageMetadata` might be more simple. But we need to take care of the compatibility issue. We can simplify the payload format when this field is set. Let me try a demo first.

coderzc:
> I prefer to make changes on the server side for several reasons.
> 
> * Broker can handle this issue well and provide a compatibility solution.
> * Add a flag for the valid compacted position can avoid deserializing and uncompressing the entire batch messages metadata header.
> * If making this change at the client-side, it increases the complexity for the client, and all language clients need to change.
> * It's weird to return EntryBuffer when getting the last message ID.
> * Actually, Pulsar only needs to retain valid message data in a compacted entry, but it retains all compacted messages with  "empty header" and "empty payload".

If we purpose are to ensure that the compaction task is successful, we only need to check `lastCompactedPosition and lastDispatchablePosition` before trigger compaction task. If `lastDispatchablePosition <= lastCompactedPosition`, then I just need to skip the compact task because there is no new data need to be compacted.

If we need make the Pulsar reader to read Kafka format data, then we need this change.

BewareMyPower:
> If we need make the Pulsar reader to read Kafka format data, then we need this change.

No. We need to ensure the `GetLastMessageId` RPC works even if the last entry is compacted with a non-default payload format.

The consumer is able to configure a `MessagePayloadProcessor` to parse the message of different format. The reader does not have the configuration and I will use another PR to add it.

pip/pip-429.md
```diff
@@ -0,0 +1,161 @@
...
+    // In this case, this field will be set to `[0, 2]`, and the payload buffer will only include the retained messages.
+    //
+    // Note: Batches compacted by older versions of the compaction service do not include this field. For such batches, 
+    // the `compacted_out` field in `SingleMessageMetadata` must be checked to identify and filter out compacted messages 
+    // (e.g., `k1 => v1` and `k1 => null` in the example above).
+    repeated int32 compacted_batch_indexes = 31;
+```
+
+# Backward & Forward Compatibility
+
```
RobertIndie, line 167: This section is empty. I think we could move the section line 124-126 here.

BewareMyPower: Added

RobertIndie (approved):

pip/pip-429.md
```diff
@@ -0,0 +1,165 @@
...
+# Motivation
+
+Decompressing the payload buffer solely to check whether individual messages have the `compacted_out` field set is both inefficient and restrictive, as it imposes constraints on the payload buffer format. Furthermore, when using a custom topic compaction service, the entry buffer in the compacted ledger may not include a `SingleMessageMetadata` for every single message, adding further complexity to the process.
+
+This challenge is exacerbated when messages are encrypted, as decompression is not possible without the public key required for decryption. This limitation also impacts the current compaction service, as encrypted messages cannot be compacted. Consequently, operations such as the `GetLastMessageId` RPC will fail, resulting in an error similar to the following:
```
codelipenghui:
```suggestion
This challenge is exacerbated when messages are encrypted, as decryption is not possible without the public key required for decryption. This limitation also impacts the current compaction service, as encrypted messages cannot be compacted. Consequently, operations such as the `GetLastMessageId` RPC will fail, resulting in an error similar to the following:
```

pip/pip-429.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+Instead, the expected behavior is to return the last message ID (e.g., `k1 => null` in the previous example).
+
+Another issue arises from the assumption made by the `GetLastMessageId` RPC that the compacted entry's payload buffer must always contain a `SingleMessageMetadata` list. However, this is not always the case. For instance, a custom topic compaction service might write a payload buffer that omits the `SingleMessageMetadata`. In such cases, the `compactedOut` information could instead be stored in the properties of the `MessageMetadata`, but the `GetLastMessageId` RPC will always fail.
+
+The custom topic compaction service has the flexibility to serialize and deserialize the payload buffer in a different format. However, it still depends on the `GetLastMessageId` RPC in the `hasMessageAvailableAsync` and `getLastMessageIdAsync` methods of the `RawReader` to compute the last message ID. This reliance creates a compatibility issue, as the `GetLastMessageId` RPC will fail when working with a payload buffer in a non-standard format, breaking the functionality of these methods.
```
codelipenghui, line 80: I'm thinking, how about introducing a new method `getLastMessageID()` to the TopicCompactionService for this case?

The MessageMetadata change is still reasonable for pulsar format encrypted data.

BewareMyPower: Accidentally I have this idea last year when we're implementing a dummy topic compaction service: https://github.com/BewareMyPower/pulsar/pull/38

It makes sense to me. Let me update it

BewareMyPower: I've updated the changes on `TopicCompactionService`, PTAL again

dao-jun (approved):
LGTM, good change!

coderzc (approved):

codelipenghui (approved):

PR #24441, opened 2025-06-23
[fix][ml] Enhance OpFindNewest to support skip non-recoverable data

coderzc:
### Motivation

When some ledgers are not offloaded successfully, they will be removed from the metadata store so that seek by timestamp might fail.

```
Error opening ledger for reading at position 4098289:438 - org.apache.bookkeeper.mledger.ManagedLedgerException$LedgerNotExistException: No such ledger exists on Metadata Server

```
skipNonRecoverableData is true, but it does not handle the seek case. The workaround is to see by message id.

### Modifications

When `skipNonRecoverableData` is true, OpFindNewest can auto skip non-recoverable data.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
@@ -196,9 +196,8 @@ class OpFindNewest implements ReadEntryCallback {
                 callback.findEntryComplete(lastMatchedPosition, OpFindNewest.this.ctx);
                 return;
             } catch (Exception e) {
-                log.error("[{}] Failed to skip non-recoverable data at position {}: {}", ledger.getName(),
-                    searchPosition, e.getMessage(), e);
-                callback.findEntryFailed(new ManagedLedgerException("Failed to skip non-recoverable data", e),
+                callback.findEntryFailed(
+                    new ManagedLedgerException("Failed to skip non-recoverable data during search position", e),
                     Optional.ofNullable(searchPosition), OpFindNewest.this.ctx);
                 return;
             }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
@@ -147,53 +147,61 @@ class OpFindNewest implements ReadEntryCallback {
     public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
         if (exception instanceof ManagedLedgerException.NonRecoverableLedgerException
             && ledger.getConfig().isAutoSkipNonRecoverableData()) {
-            log.info("[{}] Ledger {} is not recoverable, skip non-recoverable data, state:{}", ledger.getName(),
-                searchPosition, state);
-            checkArgument(state == State.checkFirst || state == State.checkLast || state == State.searching);
-            if (state == State.checkFirst) {
-                // If we failed to read the first entry, try next valid position
-                Position nextPosition = findNextValidPosition(searchPosition, exception);
-                if (nextPosition != null && nextPosition.getEntryId() != -1) {
-                    long numberOfEntries =
-                        ledger.getNumberOfEntries(Range.closedOpen(searchPosition, nextPosition));
-                    searchPosition = nextPosition;
-                    min += numberOfEntries;
-                    find();
-                    return;
-                }
-            } else if (state == State.checkLast) {
-                Position prevPosition = findPreviousValidPosition(searchPosition, exception);
-                if (prevPosition != null && prevPosition.getEntryId() != -1) {
-                    long numberOfEntries =
-                        ledger.getNumberOfEntries(Range.openClosed(prevPosition, searchPosition));
-                    searchPosition = prevPosition;
-                    max -= numberOfEntries;
-                    find();
-                    return;
-                }
-            } else if (state == State.searching) {
-                // In searching state, if we failed to read the mid entry, try next valid position
-                Position nextPosition = findNextValidPosition(searchPosition, exception);
-                if (nextPosition != null && nextPosition.getEntryId() != -1) {
-                    searchPosition = nextPosition;
-                    find();
-                    return;
-                } else {
-                    // If we can't find next valid position, try previous valid position
-                    Position prevPosition = findPreviousValidPosition(searchPosition, exception);
-                    if (prevPosition != null && prevPosition.getEntryId() != -1) {
-                        searchPosition = prevPosition;
+            try {
+                log.info("[{}] Ledger {} is not recoverable, skip non-recoverable data, state:{}", ledger.getName(),
+                    searchPosition, state);
+                checkArgument(state == State.checkFirst || state == State.checkLast || state == State.searching);
+                if (state == State.checkFirst) {
+                    // If we failed to read the first entry, try next valid position
+                    Position nextPosition = findNextValidPosition(searchPosition, exception);
+                    if (nextPosition != null && nextPosition.getEntryId() != -1) {
+                        long numberOfEntries =
+                            ledger.getNumberOfEntries(Range.closedOpen(searchPosition, nextPosition));
+                        searchPosition = nextPosition;
+                        min += numberOfEntries;
                         find();
                         return;
                     }
+                } else if (state == State.checkLast) {
+                    Position prevPosition = findPreviousValidPosition(searchPosition, exception);
+                    if (prevPosition != null && prevPosition.getEntryId() != -1) {
+                        long numberOfEntries =
+                            ledger.getNumberOfEntries(Range.openClosed(prevPosition, searchPosition));
+                        searchPosition = prevPosition;
+                        max -= numberOfEntries;
+                        find();
+                        return;
+                    }
+                } else if (state == State.searching) {
+                    // In searching state, if we failed to read the mid entry, try next valid position
+                    Position nextPosition = findNextValidPosition(searchPosition, exception);
+                    if (nextPosition != null && nextPosition.getEntryId() != -1) {
+                        searchPosition = nextPosition;
+                        find();
+                        return;
+                    } else {
+                        // If we can't find next valid position, try previous valid position
+                        Position prevPosition = findPreviousValidPosition(searchPosition, exception);
+                        if (prevPosition != null && prevPosition.getEntryId() != -1) {
+                            searchPosition = prevPosition;
+                            find();
+                            return;
+                        }
+                    }
                 }
-            }
 
-            // If don't find any entry, return the last matched position
-            log.warn("[{}] Failed to find next valid entry. Returning last matched position: {}", ledger.getName(),
-                lastMatchedPosition);
-            callback.findEntryComplete(lastMatchedPosition, OpFindNewest.this.ctx);
-            return;
+                // If don't find any entry, return the last matched position
+                log.warn("[{}] Failed to find next valid entry. Returning last matched position: {}", ledger.getName(),
+                    lastMatchedPosition);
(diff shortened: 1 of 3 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
```diff
@@ -115,31 +119,93 @@ public void readEntryComplete(Entry entry, Object ctx) {
...
+                        searchPosition = prevPosition;
+                        max -= numberOfEntries;
+                        find();
+                        return;
+                    }
+                    break;
+                case searching:
+                    // In searching state, if we failed to read the mid entry, try next valid ledger
+                    if (nextPosition != null && nextPosition.getEntryId() != -1) {
+                        searchPosition = nextPosition;
```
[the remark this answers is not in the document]
coderzc: I don't think we should adjust min, because we still don't know if the message of searchPosition meets the conditions.

codelipenghui:
> When some ledgers are not offloaded successfully, they will be removed from the metadata store so that seek by timestamp might fail.

@coderzc Could you please elaborate above description? It's not clear about why the ledger will be removed from the metadata store when they are not offloaded successfully.

As I understand, it should be an expected behavior for the ledger removed from the metadata storage. It might cause by BUGs, or users deleted the ledger manually without deleting the topic, right?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
```diff
@@ -115,31 +119,93 @@ public void readEntryComplete(Entry entry, Object ctx) {
...
     public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
+        if (exception instanceof ManagedLedgerException.NonRecoverableLedgerException
+            && ledger.getConfig().isAutoSkipNonRecoverableData()) {
+            log.info("[{}] Ledger {} is not recoverable, skip non-recoverable data, state:{}", ledger.getName(),
+                searchPosition, state);
+            Position nextPosition = null;
+            if (exception instanceof ManagedLedgerException.LedgerNotExistException) {
+                Long nextLedgerId = ledger.getNextValidLedger(searchPosition.getLedgerId());
+                if (nextLedgerId != null) {
+                    nextPosition = PositionFactory.create(nextLedgerId, 0);
```
poorbarcode: Can the next ledger be empty?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
```diff
@@ -115,31 +119,93 @@ public void readEntryComplete(Entry entry, Object ctx) {
...
 
     @Override
     public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
+        if (exception instanceof ManagedLedgerException.NonRecoverableLedgerException
+            && ledger.getConfig().isAutoSkipNonRecoverableData()) {
+            log.info("[{}] Ledger {} is not recoverable, skip non-recoverable data, state:{}", ledger.getName(),
+                searchPosition, state);
+            Position nextPosition = null;
+            if (exception instanceof ManagedLedgerException.LedgerNotExistException) {
+                Long nextLedgerId = ledger.getNextValidLedger(searchPosition.getLedgerId());
```
poorbarcode: Two concerns

### 1. Will the next ledger be out of the range that users want to search?

---

### 2. Will this progress go to an infinite loop as follows?
0. Entry ranges: [min...unrecoverable...max]
0. ...(find first, find latest...)
0. Calculate `mid()`, the result is an `unrecoverable` entry, we mark it as `Position-U`, then try the next ledger, we mark this position as `Position-Next`
0. Read `Position-Next` successfully, and calculate next `mid()` by `[min, Position-Next]`, get `Position-U` again.
0. Loop to step 3

coderzc:
> As I understand, it should be an expected behavior for the ledger removed from the metadata storage. It might cause by BUGs, or users deleted the ledger manually without deleting the topic, right?

Yes, this is an expected behavior, this fix only to unblock seeking by timestamp for topic with non-recoverable ledger. for why this expected behavior occurs, we need to continue investigating.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
coderzc: > 2. Will this progress go to an infinite loop as follows?

3. If read Position-Next successfully, and calculate next mid() by [min, Position-U), since last mid don't been changed.

coderzc: > 1. Will the next ledger be out of the range that users want to search?

@poorbarcode I've added boundary checks and refactored the code, please take a look.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
coderzc: Add check logic

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SubscriptionSeekTest.java
```diff
@@ -563,6 +566,91 @@ public void testSeekByTimestamp() throws Exception {
...
+        @Cleanup
+        Producer<String> producer =
+            pulsarClient.newProducer(Schema.STRING).topic(topicName).enableBatching(false).create();
+        for (int i = 0; i < 55; i++) {
+            producer.send(("message-" + i));
+            Thread.sleep(10);
+        }
+
+        PersistentTopic topic = (PersistentTopic) pulsar.getBrokerService().getTopic(topicName, false).get().get();
+
```
BewareMyPower:
```suggestion
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SubscriptionSeekTest.java
```diff
@@ -563,6 +566,91 @@ public void testSeekByTimestamp() throws Exception {
...
+        deletedLedgerIds.add(ledgerIds.get(ledgerIds.size() - 1));
+        int mid = ledgerIds.size() / 2;
+        deletedLedgerIds.add(ledgerIds.get(mid));
+
+        for (Long deletedLedgerId : deletedLedgerIds) {
+            pulsar.getBookKeeperClient().deleteLedger(deletedLedgerId);
+            log.info("delete ledger: {}", deletedLedgerId);
+        }
+
+        restartBroker();
```
BewareMyPower: I think we don't need to restart the broker here. If the goal is only to reload the managed ledger, you can just unload the namespace.

```suggestion
        admin.namespaces().unload(TopicName.get(topicName).getNamespace());
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentMessageFinderTest.java
```diff
@@ -439,6 +449,153 @@ void testMessageExpiryWithTimestampNonRecoverableException() throws Exception {
...
+        // {0、1、2、3、4} (0) 3 x
+        // {5、6、7、8、9} (1) 4
+        // {10、11、12、13、14} (2) 5
+        // {15、16、17、18、19} (3) 6
+        // {20、21、22、23、24} (4) 7 x
+        // {25、26、27、28、29} (5) 8 x
+        // {30、31、32、33、34} (6) 9
+        // {35、36、37、38、39} (7) 10
+        // {40、41、42、43、44} (8) 11
+        // {45、46、47、48、49} (9) 12 x
```
BewareMyPower: `、` is not a ASCII character, could you replace it with `, `?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SubscriptionSeekTest.java
```diff
@@ -563,6 +566,91 @@ public void testSeekByTimestamp() throws Exception {
...
+            .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest).subscribe();
+        long[] timestamps = timestampToMessageId.keySet().stream().mapToLong(Long::longValue).toArray();
+        ArraySorter.sort(timestamps);
+
+        timestampToMessageId.values().removeIf(messageId -> deletedLedgerIds.contains(messageId.getLedgerId()));
+
+        final int lastNonRecoverableEntryNums = 5;
+
+        for (int i = 0; i < timestamps.length - lastNonRecoverableEntryNums; i++) {
+            MessageIdImpl nextValidMessageId = (MessageIdImpl) timestampToMessageId.get(timestamps[i]);
```
BewareMyPower:
```suggestion
            MessageIdImpl nextValidMessageId = timestampToMessageId.get(timestamps[i]);
```

coderzc:
@codelipenghui @BewareMyPower @poorbarcode PTAL

BewareMyPower (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
```diff
@@ -115,34 +120,131 @@ public void readEntryComplete(Entry entry, Object ctx) {
             } else {
                 // start binary search
                 state = State.searching;
-                searchPosition = ledger.getPositionAfterN(startPosition, mid(), PositionBound.startExcluded);
+                this.mid = mid();
+                searchPosition = ledger.getPositionAfterN(startPosition, this.mid, PositionBound.startExcluded);
```
codelipenghui, line 122: @coderzc could you please also help check line:104 which was added by https://github.com/apache/pulsar/pull/22891/files#diff-7d6cd459f9134d4aeda2418630110e5abefdf6d367f6e83747c65f8637a6f851R99-R100? It should be duplicated with line:102?

coderzc: Yes, they are duplicated, I remove line:104

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
BewareMyPower: @poorbarcode Do you have a chance to take a look again?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpFindNewest.java
```diff
@@ -115,34 +118,131 @@ public void readEntryComplete(Entry entry, Object ctx) {
...
+            searchPosition = ledger.getPositionAfterN(startPosition, this.mid, PositionBound.startExcluded);
             find();
         }
     }
 
     @Override
     public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
+        if (exception instanceof ManagedLedgerException.NonRecoverableLedgerException
+            && ledger.getConfig().isAutoSkipNonRecoverableData()) {
+            log.info("[{}] Ledger {} is not recoverable, skip non-recoverable data, state:{}", ledger.getName(),
```
Technoboy-: Do we need to add `try/catch` block to this new added logic in case of exceptions causing the callback failing to execute?

coderzc: Add catch

Technoboy- (approved):

PR #24444, opened 2025-06-23
[pip] PIP-430: Pulsar Broker cache improvements: refactoring eviction and adding a new cache strategy based on expected read count

lhotari:
### Motivation

I'd like to propose PIP-430, which addresses performance and efficiency issues in Pulsar broker's entry cache eviction mechanisms and introduces a more efficient caching strategy.

The current broker entry cache implementation has several production-impacting issues. The size-based eviction doesn't guarantee removal of globally oldest entries, leading to suboptimal cache utilization. More critically, the timestamp-based eviction iterates through all ManagedLedgers every 10ms by default, causing high CPU utilization in brokers with many topics. Mixed read patterns like tailing, catch-up, and Key_Shared replays break eviction assumptions, resulting in unnecessary BookKeeper and S3 reads that increase operational costs.

PIP-430 introduces two main improvements. First, a centralized eviction mechanism using a global RangeCacheRemovalQueue that tracks all cached entries in insertion order. This replaces the expensive per-ledger iteration with a single periodic task and ensures true oldest-first eviction globally. Second, a new "expected read count" cache strategy where entries track how many active cursors are anticipated to read them. This allows the cache to intelligently retain entries that have higher utility, especially in fan-out scenarios and Key_Shared subscriptions.

The benefits include reduced CPU overhead, improved cache hit rates through better eviction decisions, and proper handling of diverse read patterns. The new strategy is configurable via cacheEvictionByExpectedReadCount (default: true) and maintains full backward compatibility with no client-facing API changes.

This addresses long-standing performance issues that particularly affect production deployments with high topic counts or diverse consumption patterns. The refactored architecture also provides a solid foundation for future cache optimizations.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-430.md
+++ b/pip/pip-430.md
@@ -1,0 +1,335 @@
+# PIP-430: Pulsar Broker cache improvements: refactoring eviction and adding a new cache strategy based on expected read count
+
+# Background knowledge
+
+Apache Pulsar brokers maintain an in-memory entry cache to reduce latency and load on BookKeeper and tiered storage (S3) by serving frequently accessed message data directly from memory.
+
+Key concepts:
+- **`ManagedLedgerImpl`**: Manages the storage for a single topic partition. Each `ManagedLedgerImpl` instance has its own `EntryCache` (typically `RangeEntryCacheImpl`) instance.
+- **`EntryCache`**: Stores message entries (payloads) in memory. The default implementation is `RangeEntryCacheImpl`, which uses a `RangeCache` internally.
+- **`RangeCache`**: A specialized cache storing entries mapped by their `Position` (ledgerId, entryId). It supports range-based operations for retrieval and expiration.
+- **`EntryCacheManager` (`RangeEntryCacheManagerImpl`)**: A global component that limits the total size of all entry caches in a broker. When the total size exceeds a threshold, it triggers eviction.
+- **Cache Eviction Policies**: Mechanisms to remove entries from the cache to make space or remove old data.
+    - **Timestamp-based eviction**: Removes entries older than a configured threshold (e.g., `managedLedgerCacheEvictionTimeThresholdMillis`). This is currently handled periodically by `ManagedLedgerFactoryImpl` iterating over all managed ledgers.
+    - **Size-based eviction**: Removes entries when the total cache size exceeds a configured limit (`managedLedgerCacheSizeMB`). The current implementation resides in `EntryCacheDefaultEvictionPolicy`, which selects a subset of larger caches and proportionally evicts entries from each entry cache to keep the total cache size under the limit.
+    - **Cursor-based eviction**: Invalidates cache entries up to the slowest consumer's read position or mark-delete position, depending on `cacheEvictionByMarkDeletedPosition`.
+
+The broker cache serves various read patterns:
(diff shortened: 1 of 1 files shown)


lhotari:
Implementation related:
* First part is implemented in https://github.com/apache/pulsar/pull/24363
* There's a WIP PoC of the second part in https://github.com/lhotari/pulsar/pull/209
  * This doesn't implement `managedLedgerCacheEvictionTimeThresholdMillisMax` yet.

lhotari:
Vote thread https://lists.apache.org/thread/p6jroxhtjpk7hbgqt3hzjtjd2233zmxy

pip/pip-430.md
```diff
@@ -0,0 +1,335 @@
...
+
+This centralized approach replaces the distributed, per-cache iteration for timestamp eviction and the less precise `EntryCacheDefaultEvictionPolicy` for size eviction, leading to more globally optimal and efficient eviction decisions.
+
+## 2. "Expected Read Count" Cache Strategy
+
+This new strategy aims to improve cache hit rates by retaining entries that are likely to be read by multiple active consumers/cursors.
+-   **`EntryReadCountHandler`**: Each cached entry (`CachedEntryImpl`) will be associated with an `EntryReadCountHandlerImpl`. This handler maintains an `expectedReadCount` (an atomic integer).
+-   **Initialization**:
+    -   When a new entry is added to the ledger (`OpAddEntry`), its `expectedReadCount` is initialized to the number of *active* cursors.
+    -   When entries are read from BookKeeper or tiered storage and inserted into the cache (`RangeEntryCacheImpl.readFromStorage`), the `expectedReadCount` is initialized based on the current state of active cursors currently positioned before or at the entry being added. This information is sourced from `ManagedCursorContainer.getNumberOfCursorsAtSamePositionOrBefore(ManagedCursor)`.
```
codelipenghui, line 86: Is it better to also check the read position of the active cursors? If they are far away from this entry, we might don't need consider it in  expectedReadCount?

lhotari: @codelipenghui That's possible. However, calculating the distance is potentially a costly operation. 

In the current design, there are 2 separate configuration options for the TTL. If there's a far-away active cursor, the only impact would be that the second TTL option would be used for expiration and the entry would be unnecessarily kept in the cache for `managedLedgerCacheEvictionTimeThresholdMillisMax` instead of getting evicted after `managedLedgerCacheEvictionTimeThresholdMillis`.

In future changes, it would be possible to improve the way `expectedReadCount` is calculated. There would need to be a fast way to perform the calculation, and this might not be an easy problem to solve. I have some ideas, but I didn't want to complicate the design of this PIP. Ideally, it would be possible to quickly calculate which cursors coming "after" the cached entry position could potentially read the entry before it reaches TTL. The current throughput of the cursors could be useful information in performing this calculation; however, it would be very costly to take all of this into account. Therefore, there would need to be a quick way to estimate.

lhotari:
Vote passed: https://lists.apache.org/thread/4bb3lsztr9j4qvgcdzv6fszdnf6bq7t8

PR #24449, opened 2025-06-24
[improve][broker] Enhance compaction triggering with message check to avoid unnecessary operations

coderzc:
### Motivation

If no more message need to compact, then we should skip compaction task

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -4041,9 +4041,12 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
     public synchronized CompletableFuture<Void> triggerCompactionWithCheckHasMoreMessages() {
         return getLastDispatchablePosition().thenCombine(topicCompactionService.getLastCompactedPosition(),
             (lastDispatchablePosition, lastCompactedPosition) -> {
+                if (lastDispatchablePosition == null) {
+                    lastDispatchablePosition = PositionFactory.EARLIEST;
+                }
                 return lastCompactedPosition == null || lastDispatchablePosition.compareTo(lastCompactedPosition) > 0;
-            }).thenAccept(hasMoreMessages -> {
-            if (!hasMoreMessages) {
+            }).thenAccept(hasMoreMessagesToBeCompacted -> {
+            if (!hasMoreMessagesToBeCompacted) {
                 log.info("[{}] No more messages to compact, skip triggering compaction", topic);
                 return;
             }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -2137,11 +2137,11 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                 if (backlogEstimate > compactionThreshold) {
                     if (log.isDebugEnabled()) {
                         log.debug(
-                                "topic:{} backlogEstimate:{} is bigger than compactionThreshold:{}. Triggering "
-                                        + "compaction", topic, backlogEstimate, compactionThreshold);
+                            "topic:{} backlogEstimate:{} is bigger than compactionThreshold:{}. Triggering "
+                                + "compaction", topic, backlogEstimate, compactionThreshold);
                     }
 
-                triggerCompactionWithCheckHasMoreMessages();
+                    triggerCompactionWithCheckHasMoreMessages();
 
                 }
             }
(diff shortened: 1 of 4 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4014,6 +4010,28 @@ public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
         return completableFuture;
     }
 
+    public synchronized CompletableFuture<Void> triggerCompactionWithCheckHasMoreMessages() {
+        return getLastDispatchablePosition().thenCombine(topicCompactionService.getLastCompactedPosition(),
+            (lastDispatchablePosition, lastCompactedPosition) -> {
+                return lastCompactedPosition == null || lastDispatchablePosition.compareTo(lastCompactedPosition) > 0;
```
3pacccccc, line 4047: I think lastDispatchablePosition might be null, so need to add null check to prevent NPE

3pacccccc:
I'm wondering does ```lastDispatchablePosition.compareTo(lastCompactedPosition) > 0``` really means need to triggering compaction? For example, I publish a message with no key present after last compaction, there's actually no need to triggering compaction as well.

coderzc:
> I'm wondering does `lastDispatchablePosition.compareTo(lastCompactedPosition) > 0` really means need to triggering compaction? For example, I publish a message with no key present after last compaction, there's actually no need to triggering compaction as well.

if `topicCompactionRetainNullKey=false`, then a message with no key may be deleted during compaction, so trigger compaction task is meaningful

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
coderzc: fix

3pacccccc:
> > I'm wondering does `lastDispatchablePosition.compareTo(lastCompactedPosition) > 0` really means need to triggering compaction? For example, I publish a message with no key present after last compaction, there's actually no need to triggering compaction as well.
> 
> if `topicCompactionRetainNullKey=false`, then a message with no key may be deleted during compaction, so trigger compaction task is meaningful

OK, I get it, thank you, now LGTM

codelipenghui (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4014,6 +4010,31 @@ public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
         return completableFuture;
     }
 
+    public synchronized CompletableFuture<Void> triggerCompactionWithCheckHasMoreMessages() {
+        return getLastDispatchablePosition().thenCombine(topicCompactionService.getLastCompactedPosition(),
```
Technoboy-, line 4042: `checkBeforeTriggerCompaction` is better ?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4042,6 +4038,31 @@ public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
...
+                    lastDispatchablePosition = PositionFactory.EARLIEST;
+                }
+                return lastCompactedPosition == null || lastDispatchablePosition.compareTo(lastCompactedPosition) > 0;
+            }).thenAccept(hasMoreMessagesToBeCompacted -> {
+            if (!hasMoreMessagesToBeCompacted) {
+                log.info("[{}] No more messages to compact, skip triggering compaction", topic);
+                return;
+            }
+            try {
+                triggerCompaction();
```
poorbarcode, line 4054: Should we also check whether the topic has enabled compaction?

poorbarcode (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
coderzc: `compactionThreshold > 0`? It seems that this is just the threshold for a compaction check task.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
coderzc: It seems that it didn't specify what to check.

PR #24453, opened 2025-06-24
[fix][broker] replication does not work due to the mixed and repetitive sending of user messages and replication markers

poorbarcode:
### Motivation

**Background of deduplication**: Always regard `markers` as non-repetitive
**Background of replicated subscription**: The internal producer of the replicator will send markers to replicate the subscription ack state to the remote side.

**Reproduce steps of the issue**

0. Replicator replicate a user message(`user-msg-1`) to the remote cluster
0. Replicator sends a replication marker(`marker-1`) to the remote cluster
0. Replicator replicate a user message(`user-msg-2`) to the remote cluster
0. The messages `[marker, user-msg-2]` have not received the send receipt yet.
0. A disconnection of the internal producer occurs.
0. Resend the `2` messages: `[user-msg-1, marker, user-msg-1]`.
0. The remote broker is handling the write of `marker-1`, because it regards all markers as non-repetitive.
  a. Has not replied to the response to the source cluster yet. 
0. Received the send receipt of `user-msg-1`, the producer gets a message ID whose value is `-1:-1` because it is repeated. It is faster than the message `marker-1` because it does not write BK.
  a. Attempts to remove the first pending send. But fails because the first pending send has not received a send receipt yet.
  b. Reconnects.
  c. Loop to `Step-4`

The send receipt of `marker-2` will never be received due to infinite reconnection.

### Scenarios that are affected

**Background**: There are 2` types of Maker: TXN and Replicated Subscription
- TXN internal component only sends markers, never sends user messages
- Replicator sends both markers and user messages, which may encounter the issue.

### Solutions

1. Let all receipts of sent in order:
  a. It can hardly be implemented, because the broker should maintain all pending sends for all producers
  b. It will break the behavior defined in the previous version.
  c. It only works for a few scenarios that are used by Replication.
2. Let the replicator handle the receipts who have out of order(what the current PR does)

### Modifications

Fix the issue with solution 2

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
@@ -27,6 +27,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.pulsar.client.api.Schema;
 import org.apache.pulsar.client.impl.conf.ProducerConfigurationData;
 import org.apache.pulsar.common.api.proto.KeyValue;
+import org.apache.pulsar.common.api.proto.MarkerType;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.protocol.Markers;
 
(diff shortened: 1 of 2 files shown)


Technoboy- (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/GeoReplicationProducerImpl.java
```diff
@@ -159,9 +160,27 @@ private void ackReceivedReplicatedMsg(ClientCnx cnx, OpSendMsg op, long sourceLI
             return;
         }
 
-        // Case-4: Unexpected
-        //   4-1: got null source cluster's entry position, which is unexpected.
-        //   4-2: unknown error, which is unexpected.
+        // Case-4: Received an out-of-order msg send receipt, and the first item in pending queue is a marker.
+        if (op.msg.getMessageBuilder().hasMarkerType()
+                && Markers.isReplicationMarker(op.msg.getMessageBuilder().getMarkerType())) {
+            log.warn("[{}] [{}] Received an out-of-order msg send receipt because enabled replicated subscription,"
```
gaoran10, line 166: Could we add a description and shorten the log? BTW, is the warning log used for future investigation?

poorbarcode: > BTW, is the warning log used for future investigation?

Yes, since the issue related to Geo Replication is hard to investigate, a long log is better.

gaoran10 (approved):

Issue #24445, opened 2025-06-23
[Bug] Memory leak in Pulsar Proxy with TopicName


frankjkelly:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

**Broker Version**: 3.3.2
**Broker Operating System**: Linux platform-pulsar-broker-0 5.10.237-230.949.amzn2.x86_64 #1 SMP Thu Jun 5 23:30:10 UTC 2025 x86_64 Linux
**Java Version**: openjdk version "21.0.4" 2024-07-16 LTS
OpenJDK Runtime Environment Corretto-21.0.4.7.1 (build 21.0.4+7-LTS)
OpenJDK 64-Bit Server VM Corretto-21.0.4.7.1 (build 21.0.4+7-LTS, mixed mode)

### Issue Description

Starting in Pulsar 3.3.2 (after upgrading from Pulsar 2.11.3) we noticed an increase in proxy heap memory usage that eventually leads to Proxy getting OOMKilled

I analyzed a recent case with Eclipse MAT 
![Image](https://github.com/user-attachments/assets/d6b2f6ec-048f-4420-a16c-b9845d57bd93)
![Image](https://github.com/user-attachments/assets/73123b18-0053-4fa2-beec-74b99cab47d9)

In communication with @lhotari he said

```
That definitely seems to be a memory leak in the Proxy. Well it's a memory leak in a form of a cache that will never expire.
It looks like this leak was introduced by https://github.com/apache/pulsar/pull/23052, added in Pulsar 3.0.6 / 3.3.1
One possible workaround would be to pass JVM option -XX:+UseStringDeduplication so that duplicate java.lang.String instances referred by the cache would be de-duplicated. That doesn't fix the issue, but would reduce the amount of heap memory spent by duplicate domain, tenant and namespace String instances in memory.
[...]

PR #24457, opened 2025-06-24
[fix] Fix memory leak in Pulsar Proxy and Pulsar client caused by TopicName cache and reduce heap memory usage

lhotari:
Fixes 24445

### Motivation

There's currently a memory leak in Pulsar Proxy and Pulsar client caused by TopicName cache.
The TopicName cache will grow without bounds and eventually fill up heap memory when there's high cardinality of topic names that are resolved during the runtime of a single JVM.

Before #23052, Guava's LoadingCache was used to limit the cache size to 100000 entries. That was replaced in that PR with usage of ConcurrentHashMap due to CPU overhead of using Guava compared to plain ConcurrentHashMap.
However, #23052 didn't address cache expiration for any other components than the broker. Since TopicName cache is included in the Pulsar client, it should be self contained without requiring to start any other maintenance tasks to keep the cache size bounded.

### Modifications

- revert #23052 changes
- Add a new a abstract class NameCache that uses ConcurrentHashMap with a SoftReference value to allow memory pressure to garbage collect unreferenced entries in TopicName cache when the JVM runs low on memory
  - Implement TopicNameCache extending NameCache
  - Implement NamespaceNameCache extending NameCache
- Use Guava's Interner class to deduplicate the String instances of different components of the TopicName (tenant, namespace, localName, ...) and NamespaceName so that unique instances of these classes are shared and less heap memory is used as a result.
- Add a JMH benchmark to validate performance
  - Soft references and Interning adds overhead compared to previous implementation, but this is a reasonable tradeoff since the solution in this PR also minimizes the heap memory usage by deduplicating instances.
  - end-to-end performance for TopicName.get on Mac M3 Max is about 28M ops/second so performance shouldn't be a concern.
  - the custom NameCache implementation is about 3x more performant than Caffeine cache with similar logic.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


poorbarcode:
@lhotari 

The benefits of the solution introduced by https://github.com/apache/pulsar/pull/23052, I have pushed a PR to fix the issue #24458

- Allow temporary cache overloading, which reduces the overhead of cache evicting/adding elements and creating new TopicName
- This will not cause memory waste because if a user uses more than 10,000 topics, the memory overhead of other components of this topic, such as consumer, producer, and consumerSelector, will far exceed that of TopicName
- Configurable cache size
- "Self-contained without requiring additional maintenance tasks to keep the cache size bounded" is easy to implement, see https://github.com/apache/pulsar/pull/24458/files#diff-93feae220f18ea80cb01ec7f2cefeab410aa0f3f181eed1206da4a294db0f701R94-R95

---

> Implement memory-efficient caching: Use soft references to ensure cache entries are automatically cleared when the JVM runs low on memory

- It also increases the GC pressure, especially when the number of Topics exceeds the limit. When the limit is reached, it will lead to more GC pressure and CPU pressure (queue management and topic creation).
- The management of the queue that is defined `Reference.queue` also leads to more CPU pressure. BTW, the queue uses a `synchronized` lock to guarantee thread-safety, which leads to the situation bader.

> Add string deduplication: Use Guava's Interner class to deduplicate String instances of TopicName and NamespaceName components (tenant, namespace, localName, etc.) to reduce heap memory usage

Users should afford the memory usage of TopicName if they use more than `10,000` topics,  the memory overhead of other components of this topic, such as consumer, producer, and consumerSelector, will far exceed that of TopicName.
We shouldn't increase the burden on the CPU for this little memory.

lhotari:
@poorbarcode 
> @lhotari
> 
> The benefits of the solution introduced by #23052, I have pushed a PR to fix the issue #24458
> * Allow temporary cache overloading, which reduces the overhead of cache evicting/adding elements and creating new TopicName
> * This will not cause memory waste because if a user uses more than 10,000 topics, the memory overhead of other components of this topic, such as consumer, producer, and consumerSelector, will far exceed that of TopicName
> * Configurable cache size
> * "Self-contained without requiring additional maintenance tasks to keep the cache size bounded" is easy to implement, see https://github.com/apache/pulsar/pull/24458/files#diff-93feae220f18ea80cb01ec7f2cefeab410aa0f3f181eed1206da4a294db0f701R94-R95

It can be considered a security issue when a broker or proxy can be made to exceed the memory limits by performing a lot of dummy lookups. I'm aware that Pulsar doesn't defend against DoS and security isn't the main concern, but nevertheless, it's a concern when there's a cache that isn't bounded.

In the solution, the check will run every 7200 seconds. That's more or less useless for many use cases.


> 
> > Implement memory-efficient caching: Use soft references to ensure cache entries are automatically cleared when the JVM runs low on memory
> 
> * It also increases the GC pressure, especially when the number of Topics exceeds the limit. When the limit is reached, it will lead to more GC pressure and CPU pressure (queue management and topic creation).

This argument doesn't have much backing. Obviously there's a tradeoff in soft reference tracking, but it's worth doing this tradeoff to achieve a cache that will automatically push out entries when the JVM is out of memory.

> * The management of the queue that is defined `Reference.queue` also leads to more CPU pressure. BTW, the queue uses a `synchronized` lock to guarantee thread-safety, which leads to the situation bader.

This is not relevant for performance, as you can see in the benchmark. On M3 Max, TopicName.get can be called across 100000 topics with about 28M operations per second. With 100 threads, the throughput was reasonable, about 21M ops/s.
I know that plain ConcurrentHashMap is many times faster, but it's reasonable to do the tradeoff.

> 
> > Add string deduplication: Use Guava's Interner class to deduplicate String instances of TopicName and NamespaceName components (tenant, namespace, localName, etc.) to reduce heap memory usage
> 
> Users should afford the memory usage of TopicName if they use more than `10,000` topics, the memory overhead of other components of this topic, such as consumer, producer, and consumerSelector, will far exceed that of TopicName. We shouldn't increase the burden on the CPU for this little memory.

That's a very opinionated argument. It's a waste to have topic name components multiplied in the heap multiple times. When the name components are deduplicated, this reduces memory pressure. Without deduplicating components, each TopicName will have it's own instance of tenant and namespacePortion. SImilarly each NamespaceName instance will have it's own instance of tenant and namespacePortion. In addition, the key in the cache will be a different instance than the completeTopicName instance in the TopicName, unless it's deduplicated. This duplication will add up in the overall heap memory consumption.

dao-jun (approved):
+1

poorbarcode:
> It can be considered a security issue when a broker or proxy can be made to exceed the memory limits by performing a lot of dummy lookups. I'm aware that Pulsar doesn't defend against DoS and security isn't the main concern, but nevertheless, it's a concern when there's a cache that isn't bounded.

This is a problem related to permissions that needs to be considered. If the user can break through your auth authentication, then almost any defense measures will be useless, such as calling the API related to write, such as the unload topic

lhotari:
> This is a competing PR to fix the issue. I've already submitted #24458 and provided my argumentation in [#24457 (comment)](https://github.com/apache/pulsar/pull/24457#issuecomment-3001418568).

@poorbarcode You are free to compete. I was assigned to fix #24445 and submitted this PR before your PR. There's a clear argument why this PR will serve Pulsar users better than the solution in #24458. Feel free to disagree.

lhotari:
@poorbarcode Here's a flamegraph to prove that TopicName.get in this PR is solely limited by ConcurrentHashMap performance.
* [rendered html](https://htmlpreview.github.io/?https://gist.githubusercontent.com/lhotari/a99b5c29f104a2c201b5fec8c98a564a/raw/87db1a011be1668153004f23ae916bc49813ee32/flame-cpu-forward.html)

image

I hope this helps get confidence that this PR makes a lot of sense for all Pulsar users without the need to worry about configuring the TopicName or NamespaceName cache sizes.


To reproduce on Linux x64_64:

Install async-profiler 4.0:
```shell
mkdir ~/tools
cd ~/tools
curl -L -O https://github.com/async-profiler/async-profiler/releases/download/v4.0/async-profiler-4.0-linux-x64.tar.gz
tar zxvf async-profiler-4.0-linux-x64.tar.gz
rm async-profiler-4.0-linux-x64.tar.gz
ln -s async-profiler-4.0-linux-x64 async-profiler
```

I used these Linux kernel settings (includes also other common settings)
```
# Tune Linux kernel settings in a persistent way
cat <<EOF | sudo tee /etc/sysctl.d/99-vm-tuning.conf
# set swappiness to 1 to use swapping as a last resort
vm.swappiness=1
# set max_map_count to allow large memory-mapped files
vm.max_map_count=262144
# set aio-max-nr to allow large asynchronous I/O, required by some docker containers
fs.aio-max-nr=1048576
# set inotify limits to allow large number of files to be watched
fs.inotify.max_user_instances=1024
fs.inotify.max_user_watches=1048576
# allow async-profiler to profile non-root processes
# https://github.com/jvm-profiling-tools/async-profiler#basic-usage
# non-root process requires setting two runtime variables
kernel.perf_event_paranoid=1
kernel.kptr_restrict=0
# https://github.com/jvm-profiling-tools/async-profiler#restrictionslimitations
kernel.perf_event_max_stack=1024
# Profiler allocates 8kB perf_event buffer for each thread of the target process.
# Make sure value is large enough (more than 8 * threads)
kernel.perf_event_mlock_kb=2048
EOF
sudo sysctl -p /etc/sysctl.d/99-vm-tuning.conf
```

Run JMH benchmark with async-profiler
```shell
mvn -Pcore-modules,microbench,-main -T 1C clean package
LIBASYNCPROFILER_PATH=$HOME/tools/async-profiler/lib/libasyncProfiler.so
java -jar microbench/target/microbenchmarks.jar -prof async:libPath=$LIBASYNCPROFILER_PATH\;output=flamegraph\;dir=profile-results ".*TopicNameBenchmark.*"
# capture flamegraph for the 10 threads invalidateCache=false run
gh gist create profile-results/org.apache.pulsar.common.naming.TopicNameBenchmark.topicLookup010-Throughput-invalidateCache-false/flame-cpu-forward.html
```

BewareMyPower:
image

We should not dismiss the requested change so quickly before reaching the consensus. Such change that affects a widely used common class needs a more careful review.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -152,18 +139,18 @@ private TopicName(String completeTopicName) {
...
             } else if (parts.size() == 4) {
                 // Legacy topic name that includes cluster name
-                this.tenant = parts.get(0);
-                this.cluster = parts.get(1);
-                this.namespacePortion = parts.get(2);
-                this.localName = parts.get(3);
+                this.tenant = StringInterner.intern(parts.get(0));
+                this.cluster = StringInterner.intern(parts.get(1));
+                this.namespacePortion = StringInterner.intern(parts.get(2));
+                this.localName = StringInterner.intern(parts.get(3));
```
codelipenghui, line 153: Will it actually improve the performance for applying to local name? It only help with local topic is same but namespace is different?

And actually, intern() is also not free. If the repetition rate wasn't that high, this could be detrimental.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
```diff
@@ -0,0 +1,156 @@
...
+            keyParam = StringInterner.intern(keyParam);
+            // add new entry or replace the possible stale entry
+            valueInstance = cache.compute(keyParam, (key, existingRef) -> {
+                if (existingRef == null || existingRef.get() == null) {
+                    return createSoftReferenceValue(key);
+                }
+                return existingRef;
+            }).get();
+            if (cache.size() > getCacheMaxSize()) {
+                cacheShrinkNeeded.compareAndSet(false, true);
```
BewareMyPower:
```suggestion
                cacheShrinkNeeded.set(true);
```

No need for CAS here.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
```diff
@@ -0,0 +1,156 @@
...
+ * It uses soft references to allow garbage collection of unused instances under heavy memory pressure.
+ * This cache uses ConcurrentHashMap for lookups for performance over Guava Cache and Caffeine Cache
+ * since there was a concern in https://github.com/apache/pulsar/pull/23052 about high CPU usage for TopicName lookups.
+ */
+abstract class NameCache<V> {
+    // Cache instances using ConcurrentHashMap and SoftReference to allow garbage collection to clear unreferenced
+    // entries when heap memory is running low.
+    private final ConcurrentMap<String, SoftReferenceValue> cache = new ConcurrentHashMap<>();
+    // Reference queue to hold cleared soft references, which will be used to remove entries from the cache.
+    private final ReferenceQueue<? super V> referenceQueue = new ReferenceQueue<>();
```
codelipenghui, line 42: I have concern about using SoftReference and another queue for invalidate the cache. It seems a little over engineering to me.

How many topics we want to support on a single broker and how many topics we want to support to create in 2 hours. The original issue was a BUG which caused millions of topic names get cached but never expired. We can just add the missed expiration logic? IMO, we are not build a KV cache system. I would like suggest to keep the topic name cache simple and `stupid`, and more simpler, easy to maintain.

And the new solution will use more memory to handle the topic name cache. The SoftReference might also cause more pressure to the JVM if the JVM under high GC pressure which caused by other part, the CPU get throttled, the SoftReferences get removed which need more CPU resources to reload the cache.

pulsar-common/src/test/java/org/apache/pulsar/common/naming/TopicNameCacheTest.java
```diff
@@ -0,0 +1,70 @@
...
+        long defaultPurgeIntervalNanos = TopicNameCache.referenceQueuePurgeIntervalNanos;
+        try {
+            TopicNameCache.cacheMaxSize = Integer.MAX_VALUE;
+            TopicNameCache.referenceQueuePurgeIntervalNanos = TimeUnit.MILLISECONDS.toNanos(10);
+
+            TopicNameCache cache = TopicNameCache.INSTANCE;
+            for (int i = 0; i < 2_000_000; i++) {
+                cache.get("persistent://tenant/namespace/topic" + RandomStringUtils.randomAlphabetic(100));
+                if (i % 100_000 == 0) {
+                    System.out.println(i + " topics added to cache. Current size: " + cache.size());
```
BewareMyPower, line 60: This test asserts nothing. We should add an assertion like

```java
            assertTrue(cache.size() < 2_000_000);
```

 BTW, this test seems to rely on GC on the soft reference, could it be flaky? Do we need a `System.gc()` here?

dao-jun: System.gc cannot enforce JVM do GC, it just notify JVM can do GC at the time

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicNameCache.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * A cache for TopicName instances that allows deduplication and efficient memory usage.
+ * It uses soft references to allow garbage collection of unused TopicName instances under heavy memory pressure.
+ * This cache uses ConcurrentHashMap for lookups for performance over Guava Cache and Caffeine Cache
+ * since there was a concern in https://github.com/apache/pulsar/pull/23052 about high CPU usage for cache lookups.
+ */
+class TopicNameCache extends NameCache<TopicName> {
+    static final TopicNameCache INSTANCE = new TopicNameCache();
+    // Configuration for the cache. These settings aren't currently exposed to end users.
+    static int cacheMaxSize = 100000;
```
codelipenghui, line 32: Any reason to remove the configuration topicNameCacheMaxCapacity and hardcode the max capacity here?

codelipenghui (changes requested):
The major goal of https://github.com/apache/pulsar/pull/23052 is not supporting millions of

codelipenghui:
And we should not cherry-pick this PR to branch-3.0 and branch-4.0. The most of the changes from this PR are improvements, not bug fix. We can have an easier way to fix the issue for LTS branch such as https://github.com/apache/pulsar/pull/24458 which provided a configurable way for users to control the max capacity and cache expiration. And I believe it can work for 99% use cases.

BewareMyPower:
I ran a test to compare the `get` performance with the old (existing) implementation. The `LegacyTopicName` is migrated from current `TopicName` in master.

[benchmark.diff.txt](https://github.com/user-attachments/files/20896048/benchmark.diff.txt)

All keys have been inserted into the cache. The new implementation's performance gap seems acceptable (only 1/2 of the current implementation)

```
Benchmark      Mode  Cnt     Score     Error  Units
Main.testNew  thrpt    5  2013.904 ± 214.360  ops/s
Main.testOld  thrpt    5  4278.375 ±  82.866  ops/s
```

But when I increased the loop count from 10000 to 200000, the gap became large.

```
Benchmark      Mode  Cnt   Score   Error  Units
Main.testNew  thrpt    5   4.659 ± 0.565  ops/s
Main.testOld  thrpt    5  63.299 ± 3.645  ops/s
```

loop count: 1000000

```
Benchmark      Mode  Cnt   Score   Error  Units
Main.testNew  thrpt    5   0.506 ± 0.200  ops/s
Main.testOld  thrpt    5  11.127 ± 1.189  ops/s
```

BewareMyPower:
I changed `TopicName`'s constructor to public and add a separated test:

```java
    @Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
    @Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
    @Benchmark
    public void testNoCache() {
        for (int i = 0; i < 1000000; i++) {
            new TopicName("test-" + i);
        }
    }
```

loop count 1000000:

```
Main.testNew      thrpt    5   0.513 ± 0.204  ops/s
Main.testNoCache  thrpt    5   0.713 ± 0.361  ops/s
Main.testOld      thrpt    5  11.027 ± 0.648  ops/s
```

loop count 200000:

```
Main.testNew      thrpt    5   4.922 ± 0.609  ops/s
Main.testNoCache  thrpt    5   4.649 ± 0.188  ops/s
Main.testOld      thrpt    5  65.304 ± 1.194  ops/s
```

loop count 50000:

```
Benchmark          Mode  Cnt    Score     Error  Units
Main.testNew      thrpt    5  278.312 ± 150.342  ops/s
Main.testNoCache  thrpt    5   21.216 ±   1.151  ops/s
Main.testOld      thrpt    5  463.201 ±  17.506  ops/s
```

lhotari:
> I ran a test to compare the `get` performance with the old (existing) implementation. The `LegacyTopicName` is migrated from current `TopicName` in master.
> 
> [benchmark.diff.txt](https://github.com/user-attachments/files/20896048/benchmark.diff.txt)
> 
> All keys have been inserted into the cache. The new implementation's performance gap seems acceptable (only 1/2 of the current implementation)
> 
> ```
> Benchmark      Mode  Cnt     Score     Error  Units
> Main.testNew  thrpt    5  2013.904 ± 214.360  ops/s
> Main.testOld  thrpt    5  4278.375 ±  82.866  ops/s
> ```
> 
> But when I increased the loop count from 10000 to 200000, the gap became large.
> 
> ```
> Benchmark      Mode  Cnt   Score   Error  Units
> Main.testNew  thrpt    5   4.659 ± 0.565  ops/s
> Main.testOld  thrpt    5  63.299 ± 3.645  ops/s
> ```
> 
> loop count: 1000000
> 
> ```
> Benchmark      Mode  Cnt   Score   Error  Units
> Main.testNew  thrpt    5   0.506 ± 0.200  ops/s
> Main.testOld  thrpt    5  11.127 ± 1.189  ops/s
> ```

This has an explanation. The soft  cache will push out entries that aren't used (no strong reference in other locations).

lhotari:
@BewareMyPower please check the JMH flame graph and repro instructions in https://github.com/apache/pulsar/pull/24457#issuecomment-3001990703 . 

The flamegraph also shows that it delegates directly to ConcurrenHashMap. That's why it's performance is similar.

One missing detail in the test setup is about setting heap parameters for the JMH forked JVM since the results can vary without that due to the soft reference usage. I also noticed that in the test, it would be needed to keep strong refs in another data structure to simulate instances which are in use (like in a real broker or pulsar client) since otherwise soft refs could be cleared unless initial heap size is reasonable. I believe you came across this with larger amounts of entries.

JMH produces often skewed results on Mac OS. Another reason is the lack of using JMH "black hole" (or returning a computed value). What's your test setup like?

BewareMyPower:
You can check [benchmark.diff.txt](https://github.com/user-attachments/files/20896048/benchmark.diff.txt) for how I set up the tests.

BTW, @codelipenghui  mentioned "over engineering" [here](https://github.com/apache/pulsar/pull/24457#discussion_r2165507421). However, IMO, the cache itself is "over engineering" to me. I checked many places that `TopicName` is used. Mostly it's just used as a temporary variable to do some conversion.

Both the existing solution and this solution have issues when a lot of different topic names are used. The existing solution relies on manual call of `clearIfReachedMaxCapacity`. Introducing a scheduled task also brings unnecessary complexity. The solution in this PR relies on GC to clean up entries from cache. If any issue happens, it could be harder to diagnose.

Actually, I think we should remove the cache, which brings many issues that should never happen. From my perspective, introducing the cache is just to get around the poor implementation of `TopicName`'s constructor and it's a "GC PTSD" coding style of Pulsar.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
lhotari: It's helpful in reducing duplicate `java.lang.String` instances. For consistency, it makes sense to apply to all components. Guava's `Interner` is very performant and since the TopicName is cached, this is not a bottleneck. JMH benchmark can be easily profiled with async-profiler and there's no actual bottlenecks caused by this interning.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
lhotari: The `ReferenceQueue` is a standard part of using SoftReferences. That's how GC notifies the application layer that a reference has been collected. In this case, the reference queue will be checked every 10 seconds and it is needed to remove the empty soft reference instances from the cache. It's a common approach to use SoftReferences for cache implementations that will release strong references when available heap memory is low.

> And the new solution will use more memory to handle the topic name cache. The SoftReference might also cause more pressure to the JVM if the JVM under high GC pressure which caused by other part, the CPU get throttled, the SoftReferences get removed which need more CPU resources to reload the cache.

Only unused TopicNames will get collected. If the broker or the application holds a strong reference, it won't get removed from the cache. What is the concern?

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicNameCache.java
lhotari: > Any reason to remove the configuration topicNameCacheMaxCapacity and hardcode the max capacity here?

I happened to start changes by reverting the previous change in this area. I could restore the topicNameCacheMaxCapacity configuration. When the cache automatically adapts to entries that are really used using soft references, there isn't a common need to tune the configuration. The only reason for having the setting is to avoid the memory overhead that keeping the strong key reference (full topic name) and empty soft reference causes.

pulsar-common/src/test/java/org/apache/pulsar/common/naming/TopicNameCacheTest.java
lhotari: > This test asserts nothing. We should add an assertion like

@BewareMyPower It's true that there's no assertion. The assertion is that no crashes happen when the soft reference processing code path gets executed. That depends on the JVM arguments. When the JVM has a lot of heap, the soft reference collection won't get triggered. Testing this would require running the test separately with smaller heap size. GC gets triggered and all references get processed (asynchronously) with `System.gc();     System.runFinalization();` as long as `-XX:+UnlockExperimentalVMOptions -XX:ReferencesPerThread=0 -XX:+ParallelRefProcEnabled` are in JVM parameter. We already have them in `pom.xml` for Pulsar's tests.

Yes, the test setup for this one could be improved to really test what it's intended to do.

lhotari:
> We should not dismiss the requested change so quickly before reaching the consensus. Such change that affects a widely used common class needs a more careful review.

@BewareMyPower There was no review and there's rationale that this PR makes sense and improves Pulsar without causing performance regressions. I agree that in this case, it might take longer to reach consensus.

lhotari:
> You can check [benchmark.diff.txt](https://github.com/user-attachments/files/20896048/benchmark.diff.txt) for how I set up the tests.

@BewareMyPower How did you setup JMH to run your benchmarks that you shared? JMH results could be skewed for multiple reasons in this case as explained in my previous comment, https://github.com/apache/pulsar/pull/24457#issuecomment-3003542993

lhotari:
> I changed `TopicName`'s constructor to public and add a separated test:
> 
> ```java
>     @Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
>     @Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
>     @Benchmark
>     public void testNoCache() {
>         for (int i = 0; i < 1000000; i++) {
>             new TopicName("test-" + i);
>         }
>     }
> ```

@BewareMyPower This is an invalid JMH test case. In JMH, you need to use the "black hole" to consume computed values.
For example:

```java
    @Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
    @Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
    @Benchmark
    public void testNoCache(Blackhole blackhole) {
        for (int i = 0; i < 1000000; i++) {
            blackhole.consume(new TopicName("test-" + i));
        }
    }
```

lhotari:
> ```java
>     @Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
>     @Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
> ```

Another detail is the short benchmark duration. A better choice could be to run a single longer iteration:
```
    @Warmup(iterations = 1, time = 10, timeUnit = TimeUnit.SECONDS)
    @Measurement(iterations = 1, time = 10, timeUnit = TimeUnit.SECONDS)
```

lhotari:
I improved the JMH benchmark and fixed some flaws in it.

MacOS results (will soon run on Linux):
```
Benchmark                                (testMode)   Mode  Cnt         Score          Error  Units
TopicNameBenchmark.topicLookup001             basic  thrpt    3  26378883.001 ±  3852072.154  ops/s
TopicNameBenchmark.topicLookup001   invalidateCache  thrpt    3  24498489.317 ±  6733474.934  ops/s
TopicNameBenchmark.topicLookup001  strongReferences  thrpt    3  27146611.212 ± 20090217.310  ops/s
TopicNameBenchmark.topicLookup010             basic  thrpt    3  22199659.357 ±  5861674.879  ops/s
TopicNameBenchmark.topicLookup010   invalidateCache  thrpt    3  21913068.107 ±  3784974.042  ops/s
TopicNameBenchmark.topicLookup010  strongReferences  thrpt    3  22537707.826 ±  4154289.677  ops/s
TopicNameBenchmark.topicLookup100             basic  thrpt    3  16356441.900 ± 12729573.003  ops/s
TopicNameBenchmark.topicLookup100   invalidateCache  thrpt    3  15799393.963 ±  5250569.159  ops/s
TopicNameBenchmark.topicLookup100  strongReferences  thrpt    3  16439089.083 ± 11551885.443  ops/s
```

lhotari:
Linux results
```
Benchmark                                      (testMode)   Mode  Cnt         Score           Error  Units
TopicNameBenchmark.topicLookup001                   basic  thrpt    3   8296768.710 ±   3847917.674  ops/s
TopicNameBenchmark.topicLookup001:async             basic  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup001         invalidateCache  thrpt    3   7719755.753 ±   1550321.165  ops/s
TopicNameBenchmark.topicLookup001:async   invalidateCache  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup001        strongReferences  thrpt    3   8340214.037 ±   1532080.072  ops/s
TopicNameBenchmark.topicLookup001:async  strongReferences  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup010                   basic  thrpt    3  33723499.610 ±  77685435.770  ops/s
TopicNameBenchmark.topicLookup010:async             basic  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup010         invalidateCache  thrpt    3  29339606.271 ±  43286582.744  ops/s
TopicNameBenchmark.topicLookup010:async   invalidateCache  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup010        strongReferences  thrpt    3  37706919.154 ±  33147758.020  ops/s
TopicNameBenchmark.topicLookup010:async  strongReferences  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup100                   basic  thrpt    3  37475011.818 ± 159568954.829  ops/s
TopicNameBenchmark.topicLookup100:async             basic  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup100         invalidateCache  thrpt    3  33964067.415 ±  28649039.174  ops/s
TopicNameBenchmark.topicLookup100:async   invalidateCache  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup100        strongReferences  thrpt    3  41307757.659 ±  48036888.867  ops/s
TopicNameBenchmark.topicLookup100:async  strongReferences  thrpt                NaN                    ---
```

[flamegraph](https://htmlpreview.github.io/?https://gist.githubusercontent.com/lhotari/03fb329e7776f00eb72a266cc3a54022/raw/4669fd6aa62796ea717962b7530f0b2de124f00a/flame-cpu-forward.html)

BewareMyPower:
Well, based on my improvement in https://github.com/apache/pulsar/pull/24463, I migrated the cache implementation of this PR.  The benchmark ran on a ubuntu 22.04 runner in GitHub Actions, here is the result from See https://github.com/BewareMyPower/pulsar/actions/runs/15872123416/job/44750991782?pr=45:

```
TopicNameBenchmark.testConstruct      thrpt       254.510          ops/s
TopicNameBenchmark.testReadFromCache  thrpt       294.822          ops/s
``` 

The result is similar to my benchmark on macOS:

```
TopicNameBenchmark.testConstruct      thrpt       345.498          ops/s
TopicNameBenchmark.testReadFromCache  thrpt       375.209          ops/s
```

Both shows getting the topic name from cache is not significantly better than my latest optimization.

Please leave comments directly in https://github.com/BewareMyPower/pulsar/pull/45 if you think there is something wrong with my benchmark code since you said

> I wouldn't trust JMH results from benchmarking on Mac OS.

lhotari:
> Both shows getting the topic name from cache is not significantly better than my latest optimization.

Great work @BewareMyPower . The main aspect why I wouldn't remove the cache is due to deduplication functionality of the cache. If TopicName instances are strongly referenced, the more there are, the worse the problem is.

> Please leave comments directly in [BewareMyPower#45](https://github.com/BewareMyPower/pulsar/pull/45) if you think there is something wrong with my benchmark code since you said
> 
> > I wouldn't trust JMH results from benchmarking on Mac OS.

The benchmark code looks fine now. On Mac OS, the results just could be different than on Linux x86_64 in many cases. Especially if there's any code that accesses `System.nanoTime()`. It's many times slower on Mac OS than on Linux.

lhotari:
Latest benchmark results on Linux after removing System.nanoTime and making cache maintenance handled during the single threaded maintenance task:

```
Benchmark                                      (testMode)   Mode  Cnt         Score           Error  Units
TopicNameBenchmark.topicLookup001                   basic  thrpt    3   8772255.043 ±   2202646.758  ops/s
TopicNameBenchmark.topicLookup001:async             basic  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup001         invalidateCache  thrpt    3   7787641.717 ±   4826933.972  ops/s
TopicNameBenchmark.topicLookup001:async   invalidateCache  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup001        strongReferences  thrpt    3   8697105.056 ±   2378331.526  ops/s
TopicNameBenchmark.topicLookup001:async  strongReferences  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup010                   basic  thrpt    3  59980650.077 ±  18606653.977  ops/s
TopicNameBenchmark.topicLookup010:async             basic  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup010         invalidateCache  thrpt    3  54319848.644 ± 104656951.523  ops/s
TopicNameBenchmark.topicLookup010:async   invalidateCache  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup010        strongReferences  thrpt    3  61755152.726 ±  28077697.228  ops/s
TopicNameBenchmark.topicLookup010:async  strongReferences  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup100                   basic  thrpt    3  93982448.999 ±  36720417.649  ops/s
TopicNameBenchmark.topicLookup100:async             basic  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup100         invalidateCache  thrpt    3  81061925.069 ±  55132917.420  ops/s
TopicNameBenchmark.topicLookup100:async   invalidateCache  thrpt                NaN                    ---
TopicNameBenchmark.topicLookup100        strongReferences  thrpt    3  73160471.462 ±  47659030.231  ops/s
TopicNameBenchmark.topicLookup100:async  strongReferences  thrpt                NaN                    ---
```

[flamegraph](https://htmlpreview.github.io/?https://gist.githubusercontent.com/lhotari/3ce7dac107999cd9443aa894b37794fd/raw/fc0631acc41696c3d88fbc6f150f3daf0c80bb9b/flame-cpu-forward.html)

recent detail:
image
It would require a longer run to get more accurate results, but it looks like the performance is now very close to ConcurrentHashMap.get. Obviously running the benchmark with the current master branch would be a way to get an accurate comparison. I'll handle that later.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
poorbarcode: ### Summary

Little benefits for memory, and paid CPU costs.

If you have ten Topics in one broker, the tenants and namespaces of the ten topics will be merged into one copy, and the remaining nine copies will be GC away. However, we have also paid a price. All the String segments will be placed in a huge map, and each intern will bring CPU overhead  


### Details

- The theory of Guava's `instances` is holding a duplicate String in memory; it relies on a Map object.
- If we have two topics, Broker will use the same `tenant/namespace` String for the different TopicName
- It enables active objects to share a String object, but it does not relieve the pressure on the GC. The second String that is not in the Map still needs to be GC
- The internal Map stores a weak reference object to wrap the String object, which also costs memory.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
poorbarcode: This is indeed a reverse optimization
- **Background**: When memory is very tight, the objects referenced by SoftReference will be cleared.
- TopicName is a core concept in Pulsar. If the Topic does not exist, the TopicName object will be cleared during Young GC.
- On the contrary, if the TopicName is still in memory when memory usage is tight, it indicates that this topic is an active topic. After being cleared, it will be immediately rebuilt, causing an infinite loop and seriously affecting CPU performance

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicNameCache.java
poorbarcode: Different users have different usage, a configuration `topicNameCacheMaxCapacity` is very useful to adjust their system.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
lhotari: > This is indeed a reverse optimization
> - **Background**: When memory is very tight, the objects referenced by SoftReference will be cleared.
> - TopicName is a core concept in Pulsar. If the Topic does not exist, the TopicName object will be cleared during Young GC.
> - On the contrary, if the TopicName is still in memory when memory usage is tight, it indicates that this topic is an active topic. After being cleared, it will be immediately rebuilt, causing an infinite loop and seriously affecting CPU performance

@poorbarcode A soft reference won't get cleared if there's a strong reference elsewhere.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
lhotari: The duplicate tenant and namespace parts add up. Within a cluster, lookups could happen on any broker. The topic pattern listing use case will load up more topic names into a single broker than it has assigned to it. For partitioned topics this is multiplied by the number of partitions. There's a need to optimize memory usage of topic pattern listing use cases and String instance deduplication is a key solution in that. We can come back to that later.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicNameCache.java
lhotari: The configuration setting could be restored.

lhotari:
> https://github.com/apache/pulsar/pull/24457#discussion_r2166665826
> https://github.com/apache/pulsar/pull/24457#discussion_r2166661706
> https://github.com/apache/pulsar/pull/24457#discussion_r2166640724

@poorbarcode we can proceed with your PR and come back to the usefulness of a soft reference cache and instance deduplication in optimizing memory usage of topic pattern listing / watching use cases.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
poorbarcode: @lhotari 
> A soft reference won't get cleared if there's a strong reference elsewhere.

When passed one or several full GCs and the effect is not ideal, the soft reference object will be cleared, which, as I mentioned above. 
However, if the topicName object can survive until that time, it must be an active topic and will definitely be recreated

lhotari: > @lhotari 
> > A soft reference won't get cleared if there's a strong reference elsewhere.
> 
> When passed one or several full GCs and the effect is not ideal, the soft reference object will be cleared, which, as I mentioned above. 
> However, if the topicName object can survive until that time, it must be an active topic and will definitely be recreated
> 

Computing the new instance will be cheap as @BewareMyPower has demonstrated with his optimizations.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
poorbarcode: > The duplicate tenant and namespace parts add up. Within a cluster, lookups could happen on any broker. The topic pattern listing use case will load up more topic names into a single broker than it has assigned to it. For partitioned topics this is multiplied by the number of partitions. There's a need to optimize memory usage of topic pattern listing use cases and String instance deduplication is a key solution in that. We can come back to that later.

Let's review the lookup process, and you will understand where the problem lies
1. Received a `lookup` request
2. If the Topic Name in the cache
  a. Broker gets the cached TopicName 
  b. Release the object `lookup`
3. If the Topic Name is not in the cache 
  a. after a split, Broker gets a string array `[tenant, namespace, topic]`
  b. The Broker calls `intern(tenant), intern(namespace), intern(topic)` here, and get a new TopicName with shared `tenant & namespace`
4. After the broker responds to the client
  a. JVM will release the string array
  b. JVM will release the TopicName object, since it is not useful anymore.

Whether there is a step 3-b or not, the objects that the JVM needs to release (ultimately) are the same:
- a string array `[tenant, namespace, topic]`
- a TopicName.

But the step 3-b costs CPU resource.

codelipenghui: Is there any expected performance improvement from line 153? I see the benefit of deduplicating tenant and namespace, but I'm missing the context for the topic name. I'm wondering what cases would lead to a large number of duplicate topic names.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/NameCache.java
codelipenghui: > Only unused TopicNames will get collected. If the broker or the application holds a strong reference, it won't get removed from the cache. What is the concern?

My concern is unpredictability since the decision made by JVM is based on heuristics and is influenced by multiple factors, different JVM implemetation, different GC algorithms, different versions.

We don't want to see the topic names cache get removed and reloaded frequently, it's not data. Especially when you have a tiny broker (200MB heap memory and 100m CPU assigned). For example, the topic names cache might take 20MB in this case, if the topic names cache gets removed by the JVM GC frequently, the topic name cache clear and reload might exhaust all the CPU resources.

ben-manes:
A few observations if helpful,

- It looks like Pulsar is using an old version of Caffeine (2.9.1; May 2021) in order to run on Java 8. Is that still necessary or can you use the current version (3.2.1; June 2025)?
- You were using expireAfterAccess which incurs an expensive system time call on every read, while in the replacement it has no time threshold and an optional space limit. Can you use a size bounding for a fairer comparison?
- The maintenance is O(n) on the caller thread, which is likely fine but is a caller-facing penalty. The scan is of the underlying array, which at the previous setting of 100K entries would be up to 175-250K slots to traverse. That can have surprisingly bad performance when not handled properly. e.g. the clustering effect breaks [ehcache's sampled lru](https://gist.github.com/ben-manes/6312727adfa2235cb7c5e25cae523ad0)) assumptions.
  - Caffeine uses strictly amortized O(1) algorithms that are inexpensive. However by default it will defer to ForkJoinPool.commonPool() since we don't know the usage costs of foreign code (e.g. CHM blocks on a hashbin collision so a long compute could stall eviction). You can use `executor(Runnable::run)` in most cases. 
- If soft references are over used then they have traditionally been costly by increasing GC pauses for a slower system, making it look good in benchmarks and very problematic in real workloads. They can cause old gc thrashing, so newer GCs will often evict aggressively so there is less content captured on a real system. The eviction will likely discard useful content due to the global LRU where a polluter of inexpensive data forces the eviction of valuable expensive content. This spooky action at a distance due to the eviction policy not being localized can be harder to debug and causes overall slower system performance as projects evolve and the tradeoffs are forgotten.

codelipenghui:
Hi @ben-manes 

Thanks for your help.
I have created a PR https://github.com/apache/pulsar/pull/24527 to upgrade Caffeine Cache first.

lhotari:
Closing since this is covered in other PRs.

PR #24458, opened 2025-06-24
[fix][prox]Pulsar Proxy OOM because forget to clear topic name cache

### LINKED ISSUE #24445: [Bug] Memory leak in Pulsar Proxy with TopicName

poorbarcode:
Fixes #24445

Main Issue: #24445

### Motivation

PR https://github.com/apache/pulsar/pull/23052 fixed the issue caused by creating TopicName and adding items into the queue of TopicName Cache.

Issue: PR added a clarification for Pulsar Broker, but forgot to add the clarification for Pulsar Proxy, Pulsar WebSocket Service, and Pulsar Client.

### Modifications

- Pulsar Proxy and Pulsar WebSocket Service: to add a scheduled task to clear the cache
- Pulsar Client: clear cache when its capacity reaches `10_000`, which is the same value as the original behavior before https://github.com/apache/pulsar/pull/23052  

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
@@ -24,6 +24,8 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.pulsar.common.util.Codec;
 
@@ -36,6 +38,8 @@ public class TopicName implements ServiceUnitId {
     public static final String DEFAULT_NAMESPACE = "default";
 
     public static final String PARTITIONED_TOPIC_SUFFIX = "-partition-";
+
+    public static final int CLIENT_MAX_TOPIC_NAME_CACHE = 10_000;
 
     private final String completeTopicName;
 
@@ -50,6 +54,10 @@ public class TopicName implements ServiceUnitId {
     private final int partitionIndex;
 
     private static final ConcurrentHashMap<String, TopicName> cache = new ConcurrentHashMap<>();
+
+    @Getter
+    @Setter
+    private static boolean evictCacheByScheduledTask = false;
 
     public static void clearIfReachedMaxCapacity(int maxCapacity) {
         if (maxCapacity < 0) {
@@ -82,7 +90,11 @@ public class TopicName implements ServiceUnitId {
         if (tp != null) {
             return tp;
         }
-        return cache.computeIfAbsent(topic, k -> new TopicName(k));
+        tp = cache.computeIfAbsent(topic, k -> new TopicName(k));
+        if (!evictCacheByScheduledTask) {
+            clearIfReachedMaxCapacity(CLIENT_MAX_TOPIC_NAME_CACHE);
+        }
+        return tp;
     }
 
     public static TopicName getPartitionedTopicName(String topic) {
--- a/conf/proxy.conf
+++ b/conf/proxy.conf
@@ -39,6 +39,14 @@ brokerServiceURLTLS=
 brokerWebServiceURL=
 brokerWebServiceURLTLS=
 
+# Max capacity of the topic name cache. -1 means unlimited cache; 0 means broker will clear all cache
+# per "maxSecondsToClearTopicNameCache", it does not mean broker will not cache TopicName.
+topicNameCacheMaxCapacity=100000
+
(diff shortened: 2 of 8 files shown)


lhotari (dismissed):
This is a competing PR to fix the issue. I've already submitted #24457 and provided my argumentation in https://github.com/apache/pulsar/pull/24457#issuecomment-3001458771 .

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -51,6 +55,10 @@ public class TopicName implements ServiceUnitId {
 
     private static final ConcurrentHashMap<String, TopicName> cache = new ConcurrentHashMap<>();
 
+    @Getter
+    @Setter
+    private static boolean evictCacheByScheduledTask = false;
```
coderzc, line 60:
```suggestion
    private static boolean autoEvictCacheOnGet = true;
```
It seems that this is easier to understand?

poorbarcode: Thanks for the suggestion. But the TopicName will clear the cache itself when `autoEvictCacheOnGet` is false: https://github.com/apache/pulsar/pull/24458/files#diff-93feae220f18ea80cb01ec7f2cefeab410aa0f3f181eed1206da4a294db0f701R95

poorbarcode:
@lhotari 

> Unblocking this PR since we can come back to the https://github.com/apache/pulsar/pull/24457 while optimizing topic pattern related memory usage. String deduplication is mostly relevant for those use cases.

Yes, let's first solve the problem with the original plan, and then discuss the rationality of the optimization


Thanks ❤️

lhotari (approved):
LGTM

BewareMyPower:
@lhotari @poorbarcode @coderzc I pushed a new solution that:
- has better performance with a channel based topic cache
- does not introduce any new configurations or scheduled task

PTAL https://github.com/apache/pulsar/pull/24465

3pacccccc:
I have a question, under what scenario will evictCacheByScheduledTask be false? I see you set it to true in BrokerService、WebSocketService and ProxyService

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
coderzc: I meant to also need to changes the logic of clear the cache itself.
```java
if (autoEvictCacheOnGet) {
    clearIfReachedMaxCapacity(CLIENT_MAX_TOPIC_NAME_CACHE);
}
```

poorbarcode:
> I pushed a new solution that:

> has better performance with a channel based topic cache
> does not introduce any new configurations or scheduled task
> PTAL https://github.com/apache/pulsar/pull/24465

Close the PR

PR #24463, opened 2025-06-25
[improve][broker] Improve the performance of TopicName constructor

BewareMyPower:
### Motivation

The `TopicName`'s constructor has poor performance:
- ~~`NamespaceName#get` is very slow~~
- `Splitter.on("/").splitToList(rest)` is slow
- `String.format` is slower than the `+` operation on strings and `completeTopicName` is unnecessarily created again

### Modifications

- ~~Initialize `NamespaceName` in a lazy way~~ (don't do that because it assumes the constructor is called more frequently than `getNamespace` or `getNamespaceObject`)
- Replace `Splitter` with a manually written `splitBySlash` introduced from https://github.com/apache/pulsar/pull/24465. Actually, `StringUtils#split` has good performance as well. But it will split `"//xxx/yyy/zzz"` to `["xxx", "yyy", "zzz"]` without reporting any error.
- Initialize `completeTopicName` from the argument directly without any concentrate operation
- ~~Replace `String.format` with `+` in `fromPersistenceNamingEncoding`~~

`NamespaceName` and `fromPersistenceNamingEncoding` will not be handled in this PR.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/44

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
@@ -115,41 +115,46 @@ public class TopicName implements ServiceUnitId {
         try {
             // The topic name can be in two different forms, one is fully qualified topic name,
             // the other one is short topic name
-            if (!completeTopicName.contains("://")) {
+            int index = completeTopicName.indexOf("://");
+            if (index < 0) {
                 // The short topic name can be:
                 // - <topic>
                 // - <tenant>/<namespace>/<topic>
-                String[] parts = StringUtils.split(completeTopicName, '/');
-                if (parts.length == 3) {
-                    completeTopicName = TopicDomain.persistent.name() + "://" + completeTopicName;
-                } else if (parts.length == 1) {
-                    completeTopicName = TopicDomain.persistent.name() + "://"
-                        + PUBLIC_TENANT + "/" + DEFAULT_NAMESPACE + "/" + parts[0];
+                List<String> parts = splitBySlash(completeTopicName, 0);
+                this.domain = TopicDomain.persistent;
+                if (parts.size() == 3) {
+                    this.tenant = parts.get(0);
+                    this.namespacePortion = parts.get(1);
+                    this.localName = parts.get(2);
+                } else if (parts.size() == 1) {
+                    this.tenant = PUBLIC_TENANT;
+                    this.namespacePortion = DEFAULT_NAMESPACE;
+                    this.localName = parts.get(0);
                 } else {
                     throw new IllegalArgumentException(
                         "Invalid short topic name '" + completeTopicName + "', it should be in the format of "
                         + "<tenant>/<namespace>/<topic> or <topic>");
                 }
-            }
-
-            // Expected format: persistent://tenant/namespace/topic
-            List<String> parts = Splitter.on("://").limit(2).splitToList(completeTopicName);
-            this.domain = TopicDomain.getEnum(parts.get(0));
-
-            String rest = parts.get(1);
-
-            // Scalable topic domains (topic://, segment://) only support the new format
-            // and local names may contain '/', so use limit(3) to keep the rest as localName.
-            boolean isScalableDomain = this.domain == TopicDomain.topic
-                    || this.domain == TopicDomain.segment;
-            int splitLimit = isScalableDomain ? 3 : 4;
-            parts = Splitter.on("/").limit(splitLimit).splitToList(rest);
-            if (parts.size() == 4) {
-                throw new IllegalArgumentException(
-                        "V1 topic names (with cluster component) are no longer supported. "
-                        + "Please use the V2 format: '<domain>://tenant/namespace/topic'. Got: "
-                        + completeTopicName);
-            } else if (parts.size() == 3) {
+                this.segmentRange = null;
+                this.segmentId = -1;
+                this.completeTopicName = domain.name() + "://" + tenant + "/" + namespacePortion + "/" + localName;
+            } else {
+                this.domain = TopicDomain.getEnum(completeTopicName.substring(0, index));
+                // Scalable topic domains (topic://, segment://) only support the new format
+                // and local names may contain '/', so use limit(3) to keep the rest as localName.
+                boolean isScalableDomain = this.domain == TopicDomain.topic
+                        || this.domain == TopicDomain.segment;
+                int splitLimit = isScalableDomain ? 3 : 4;
+                List<String> parts = splitBySlash(completeTopicName.substring(index + "://".length()),
+                        splitLimit);
+                if (parts.size() == 4) {
+                    throw new IllegalArgumentException(
+                            "V1 topic names (with cluster component) are no longer supported. "
+                                    + "Please use the V2 format: '<domain>://tenant/namespace/topic'. Got: "
+                                    + completeTopicName);
+                } else if (parts.size() != 3) {
+                    throw new IllegalArgumentException("Invalid topic name " + completeTopicName);
+                }
                 this.tenant = parts.get(0);
                 this.namespacePortion = parts.get(1);
                 String rawLocalName = parts.get(2);
@@ -160,8 +165,8 @@ public class TopicName implements ServiceUnitId {
                     if (lastSlash <= 0) {
                         throw new IllegalArgumentException(
                                 "Invalid segment topic name: local name must contain"
-                                + " '<parent-topic>/<hashStart>-<hashEnd>-<segmentId>'. Got: "
-                                + completeTopicName);
+                                        + " '<parent-topic>/<hashStart>-<hashEnd>-<segmentId>'. Got: "
+                                        + completeTopicName);
                     }
                     this.localName = rawLocalName.substring(0, lastSlash);
                     String descriptor = rawLocalName.substring(lastSlash + 1);
@@ -169,7 +174,7 @@ public class TopicName implements ServiceUnitId {
                     if (descParts.length != 3) {
                         throw new IllegalArgumentException(
                                 "Invalid segment descriptor: expected '<hexStart>-<hexEnd>-<segmentId>',"
-                                + " got: '" + descriptor + "'");
+                                        + " got: '" + descriptor + "'");
                     }
                     this.segmentRange = HashRange.of(
                             Integer.parseInt(descParts[0], 16),
@@ -181,27 +186,23 @@ public class TopicName implements ServiceUnitId {
                     this.segmentId = -1;
                 }
 
-                this.partitionIndex = getPartitionIndex(completeTopicName);
-                this.namespaceName = NamespaceName.get(tenant, namespacePortion);
-            } else {
-                throw new IllegalArgumentException("Invalid topic name: " + completeTopicName);
+                if (this.domain == TopicDomain.segment) {
+                    this.completeTopicName = String.format("%s://%s/%s/%s/%s",
+                            domain, tenant, namespacePortion, localName,
+                            String.format("%04x-%04x-%d", segmentRange.start(), segmentRange.end(), segmentId));
+                } else {
+                    this.completeTopicName = completeTopicName;
+                }
             }
 
             if (StringUtils.isBlank(localName)) {
                 throw new IllegalArgumentException(String.format("Invalid topic name: %s. Topic local name must not"
                         + " be blank.", completeTopicName));
             }
-
+            this.partitionIndex = getPartitionIndex(localName);
+            this.namespaceName = NamespaceName.get(tenant, namespacePortion);
         } catch (NullPointerException e) {
             throw new IllegalArgumentException("Invalid topic name: " + completeTopicName, e);
-        }
-        if (this.domain == TopicDomain.segment) {
-            this.completeTopicName = String.format("%s://%s/%s/%s/%s",
-                    domain, tenant, namespacePortion, localName,
-                    String.format("%04x-%04x-%d", segmentRange.start(), segmentRange.end(), segmentId));
-        } else {
-            this.completeTopicName = String.format("%s://%s/%s/%s",
-                    domain, tenant, namespacePortion, localName);
         }
     }
 


BewareMyPower:
Before 48ffb7abfe1b5d5ad71cc3a07627326fb84c7a5a, there is a constructor parameter that determines whether to initialize the `NamespaceName`. I also compared with the existing `TopicName` constructor, which can be found here: https://gist.github.com/BewareMyPower/6b83c897552c110f336e51965cc91c24

The benchmark result is:

```
TopicNameBenchmark.testConstruct                          thrpt       312.983          ops/s
TopicNameBenchmark.testConstructWithNamespaceInitialized  thrpt       101.973          ops/s
TopicNameBenchmark.testLegacyTopicName                    thrpt        53.944          ops/s
TopicNameBenchmark.testReadFromCache                      thrpt       721.392          ops/s
```

As you can see
- Accessing it from the cache is only 2.3x faster than the latest implementation, which is not significant. However, to avoid debate on whether to remove cache in this PR, I only exposed the public constructor.
- Skipping initialization of `NamespaceName` is 3x faster than initializing it immediately
- It's 6x faster than the original implementation. Even if initializing `NamespaceName` immediately, it's still about 2x faster.

lhotari:
> Before [48ffb7a](https://github.com/apache/pulsar/commit/48ffb7abfe1b5d5ad71cc3a07627326fb84c7a5a), there is a constructor parameter that determines whether to initialize the `NamespaceName`. I also compared with the existing `TopicName` constructor, which can be found here: https://gist.github.com/BewareMyPower/6b83c897552c110f336e51965cc91c24
> 
> The benchmark result is:
> 
> ```
> TopicNameBenchmark.testConstruct                          thrpt       312.983          ops/s
> TopicNameBenchmark.testConstructWithNamespaceInitialized  thrpt       101.973          ops/s
> TopicNameBenchmark.testLegacyTopicName                    thrpt        53.944          ops/s
> TopicNameBenchmark.testReadFromCache                      thrpt       721.392          ops/s
> ```
> 
> As you can see
> 
> * Accessing it from the cache is only 2.3x faster than the latest implementation, which is not significant. However, to avoid debate on whether to remove cache in this PR, I only exposed the public constructor.
> * Skipping initialization of `NamespaceName` is 3x faster than initializing it immediately
> * It's 6x faster than the original implementation. Even if initializing `NamespaceName` immediately, it's still about 2x faster.

I wouldn't trust JMH results from benchmarking on Mac OS. In the case of PR #24457, the results are very different when benchmarking on Linux x86_64 with Intel i9 processor (Dell XPS 2019 laptop). 
If we'd completely remove the cache, there would be more duplicate TopicName and NamespaceName instances and more duplicate string instance. For very long tenant, namespace and topic names, that could add up a significant amount of wasted heap memory.

BewareMyPower:
@lhotari So could you help run benchmark in your Linux environment to see the difference? I can also create a workflow via GitHub Actions to see the result.

BewareMyPower:
> If we'd completely remove the cache, there would be more duplicate TopicName and NamespaceName instances and more duplicate string instance. For very long tenant, namespace and topic names, that could add up a significant amount of wasted heap memory.

I don't want to debate about it in this PR. If you have a chance to look at how `TopicName` is used in Pulsar, you will find a lot of temporary `TopicName` instances just for conversion. I know near all Pulsar constributors are very sensitive to the topic about GC, so I don't change the `TopicName#get` implementation.

The ultimate solution is to create some utils methods instead of leveraging the `TopicName` class.

BewareMyPower:
Updated test results to compare it with the legacy implementation in https://github.com/BewareMyPower/pulsar/actions/runs/15872434081/job/44752001433?pr=45

```
TopicNameBenchmark.testConstruct                             thrpt       258.932          ops/s
TopicNameBenchmark.testConstructWithNamespaceInitialization  thrpt       158.876          ops/s
TopicNameBenchmark.testLegacyConstruct                       thrpt        44.290          ops/s
TopicNameBenchmark.testReadFromCache                         thrpt       340.613          ops/s
```

- The constructor is 5.8x faster than the existing one
- The constructor with `NamespaceName` initialization is 3.6x faster than the existing one

BewareMyPower:
@lhotari @codelipenghui @nodece @dao-jun @poorbarcode @coderzc This PR is now ready to review, PTAL

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -211,6 +208,16 @@ public String getNamespace() {
      */
     @Override
     public NamespaceName getNamespaceObject() {
+        if (namespaceName != null) {
+            return namespaceName;
```
merlimat: We shouldn't need to have `namespaceName` volatile, we can just check it again in the synchronized block.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -111,64 +110,70 @@ public static String getPattern(String topic) {
...
+     * The constructor from a topic name string. The difference from {@link TopicName#get(String)} is that the `get`
+     * method can leverage the built-in cache mechanism, which can be slightly faster, but at the cost of higher JVM
+     * memory usage that could lead to unnecessary GC, as well as potentially OOM in extreme cases.
+     * You can benchmark `TopicName.get(topic)` and `new TopicName(topic)` to see the actual performance gap.
+     *
+     * @param completeTopicName the topic name
+     */
     @SuppressFBWarnings("DCN_NULLPOINTER_EXCEPTION")
-    private TopicName(String completeTopicName) {
+    public TopicName(String completeTopicName) {
```
merlimat: If it's to be used in the benchmark, let's leave the constructor package protected.

microbench/src/main/java/org/apache/pulsar/broker/qos/TopicNameBenchmark.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String[] topicBases = {"test",
+            "tenant/ns/test",
+            "persistent://another-tenant/another-ns/test"
+    };
+
+    @Threads(1)
+    @Benchmark
+    @Warmup(time = 5, iterations = 1)
+    @Measurement(time = 5, iterations = 1)
+    public void testReadFromCache(Blackhole blackhole) {
```
merlimat: Since this contains loops, it's better to add annotation to have more understandable number: `@OperationsPerInvocation(10_000 * 3)`

Also, `@OutputTimeUnit(TimeUnit.MICROSECONDS)` might be good

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
BewareMyPower: No. It's not just used in the benchmark. The downstream (e.g. the protocol handler) application can maintain its own cache mechanism for `TopicName` once the constructor is public.

lhotari:
@BewareMyPower My point from an earlier comment isn't addressed:

> If we'd completely remove the cache, there would be more duplicate TopicName and NamespaceName instances and more duplicate string instance. For very long tenant, namespace and topic names, that could add up a significant amount of wasted heap memory.

In this case, I think it's irrelevant to just benchmark the performance of creating/looking up TopicName instances.

Duplicate `java.lang.String` instances could increase significantly after introducing a non-caching solution. Heapdump analysis tools have such checks for duplicate instances so that's one way how to compare the difference.

I do agree that a lot of the TopicName handling code is a mess. For example in Topic listing, the topic name is converted from/to String multiple times. That's adding up a lot of pressure for a fast lookup solution when instances are cached. So I'm not against changing the current caching, it's just that there's a need to consider duplicate instances as well. It's likely that the caching is more relevant for NamespaceName than TopicName regarding duplicate instances. We might not be keeping a reference to TopicName instance in that many places when broker is running.

BewareMyPower:
> If we'd completely remove the cache,

I've read more code and changed my idea a bit. Caching could be helpful in many cases. But how to establish the cache might depend on specific use cases. Writing a common cache is hard. I don't like the solution in https://github.com/apache/pulsar/pull/23052, but it's anyway good to resolve the issue encountered at that time.

As I've mentioned [here](https://github.com/apache/pulsar/pull/24463#discussion_r2230192986), exposing the public method is helpful for downstream to construct its own cache. In addition, the `TopicName` can be cached as well as some other classes, e.g. `Map<String, Pair<TopicName, PersistentTopic>>` would be better than maintaining two maps:
- A built-in `Map<String, TopicName>`
- An external `Map<String, PersistentTopic>`

It's just an example, `PersistentTopic` can be replaced by another topic abstraction class.

BewareMyPower:
Regarding this PR, I'm going to revert other changes and only leaving the improvement on `TopicName`'s constructor itself.

Exposing a public method will be easier for the downstream to maintain its custom cache, but it will also be confusing for the core Pulsar developers to make the decision on `TopicName.get` and `new TopicName`.

Anyway, we should improve the use of `TopicName` case by case. Take `BrokerService` for example, a `TopicName` is passed to `getTopic`, but the string returned by `toString()` is passed to
- `loadOrCreatePersistentTopic`
- `TopicLoadingContext#topic`

The `TopicName` instance is constructed again by them to  `checkOwnershipAndCreatePersistentTopic`.

In this case, `TopicName#get` makes sense than `new TopicName`. But the code is still inefficient, we should not perform the unnecessary `topicName.toString() -> TopicName.get(topic)` conversions.

lhotari:
This PR will be an improvement. Regarding the argument I made in a previous comment about duplicate tenant and namespace `java.lang.String` instances, that problem is already present in the current code base. That could be solved in a different PR.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
BewareMyPower: But could it have the issue with [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking)?

> The code generated by the compiler is allowed to update the shared variable to point to a partially constructed object before A has finished performing the initialization.

lhotari: > But could it have the issue with [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking)?

That's true, there's a good explanation of the problem in https://shipilev.net/blog/2014/safe-public-construction/ 
The double-checked locking pattern isn't recommended in general since it could lead to a situation where the shared instance seems to be partially instantiated, unless it's completely immutable (this applies to possibly created contained instances in the constructor). In Java "benign data races" don't lead to crashes, but to potential consistency issues where the state of fields are inconsistent.

Perhaps in this case, double checked locking would be possible assuming that NamespaceName is completely immutable in regards to fields set in the constructor?

lhotari: > Perhaps in this case, double checked locking would be possible assuming that NamespaceName is completely immutable in regards to fields set in the constructor?

According to https://shipilev.net/blog/2014/safe-public-construction/, it wouldn't be safe. There isn't really a way around it.

lhotari: > Safe publication makes all the values written before the publication visible to all readers that observed the published object. It is a great simplification over the JMM rules of engagement with regards to actions, orders and such.
> There are a few trivial ways to achieve safe publication:
> 
> * Exchange the reference through a properly locked field ([JLS 17.4.5](http://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.4.5))
> 
> * Use static initializer to do the initializing stores ([JLS 12.4](http://docs.oracle.com/javase/specs/jls/se7/html/jls-12.html#jls-12.4))
> 
> * Exchange the reference via a volatile field ([JLS 17.4.5](http://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.4.5)), or [as the consequence of this rule](http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/atomic/package-summary.html), via the AtomicX classes
> 
> * Initialize the value into a final field ([JLS 17.5](http://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.5)).

lhotari: Well in this case, since we aren't constructing NamespaceName but instead calling NamespaceName.get (which handles thread safety with the cache implementation), it would be possible to avoid using a volatile field and just use DCL.

BewareMyPower: But the thread safety depends on the implementation of `NamespaceName.get`. It would also be dangerous in case someone else did copy-and-paste in future.

I decided to remove the lazy initialization for now. The prerequisite of the performance improvement is that the `TopicName`'s constructor is called more frequently than the `getNamespace` or `getNamespaceObject` methods.

Let's use a conservative solution for now

lhotari (approved):
LGTM

lhotari (commented):
For deduplicating the tenant and namespace String instances, it would be useful to assign the `tenant` and `namespacePortion` fields from the `NamespaceName` instance. This wouldn't add much overhead, but benefit in reducing the amount of heap memory since there would be less duplication of `java.lang.String` instances at runtime.

dao-jun (approved):

BewareMyPower:
@lhotari Do you have a chance to review this PR again according to https://github.com/apache/pulsar/pull/25367#issuecomment-4280093496?

I've minimize the change of this PR to only optimize the constructor of the `TopicName`. It would save the CPU time when the cache is missed. `NamespaceName` and `TopicName#fromPersistenceNamingEncoding` won't be handled in this PR.

BewareMyPower:
Updated benchmark can be found here: https://github.com/BewareMyPower/JavaBenchmark/actions/runs/24712104796/job/72279087737

```
TopicNameBenchmark.testConstruct                                  thrpt             101.457                 ops/s
TopicNameBenchmark.testLegacyConstruct                            thrpt              36.061                 ops/s
```

my local run:

```
TopicNameBenchmark.testConstruct                                  thrpt             133.459                ops/s
TopicNameBenchmark.testLegacyConstruct                            thrpt              53.429                ops/s
```

It's 2.5x ~ 2.8x faster.

Actually, the performance is also affected by the slow `NamespaceName#get` call. I tried removing the `namespace` field from both `LegacyTopicName` and `TopicName` and ran the benchmark again, the result is:

```
Benchmark                                Mode  Cnt    Score   Error  Units
TopicNameBenchmark.testConstruct        thrpt       526.592          ops/s
TopicNameBenchmark.testLegacyConstruct  thrpt        73.782          ops/s
```

It's 7x faster

lhotari (approved):
LGTM, great work @BewareMyPower

lhotari:
I backported this to maintenance branches too.

lhotari:
> Actually, the performance is also affected by the slow `NamespaceName#get` call. I tried removing the `namespace` field from both `LegacyTopicName` and `TopicName` and ran the benchmark again, the result is:

One additional detail to check is to see if optimizing org.apache.pulsar.common.naming.TopicDomain#getEnum would help. A general performance advice has been in the past to avoid calling Enum's `values()` in hot paths.

something like this:
```java
    public static TopicDomain getEnum(String value) {
        if (persistent.value.equalsIgnoreCase(value)) {
           return persistent;
        }
        if (non_persistent.value.equalsIgnoreCase(value)) {
           return non_persistent;
        }
        if (topic.value.equalsIgnoreCase(value)) {
           return topic;
        }
        if (segment.value.equalsIgnoreCase(value)) {
           return segment;
        }
        throw new IllegalArgumentException("Invalid topic domain: '" + value + "'");
    }
```

I'm not sure if this type of optimization applies to modern JVMs.

PR #24465, opened 2025-06-25
[fix][proxy] Fix proxy OOM by replacing TopicName with a simple conversion method

### LINKED ISSUE #24445: [Bug] Memory leak in Pulsar Proxy with TopicName

BewareMyPower:
Fixes #24445

### Motivation

See https://github.com/apache/pulsar/pull/24458 for details. However
- The solution of #24458 introduces new configurations and an extra scheduled task.
- https://github.com/apache/pulsar/pull/24457 is an alternative solution that aims at providing a better designed cache that can shrink automatically. It might be able to fix this issue as well, but the impact might be wide and needs more careful review and discussion.

After reviewing the usages of `TopicName` in `pulsar-proxy` module, I found the only case is to convert a user-provided topic to the full topic name. Using `TopicName` has many unnecessary overhead like parsing a `NamespaceName` (see the cost from the benchmark [here](https://github.com/apache/pulsar/pull/24463#issuecomment-3004068829)), or maintaining the `TopicName` object in a global cache that requires a thread-safe access via `ConcurrentHashMap`.

### Modifications

Introduce a new method `TopicName#toFullTopicName` for the purpose of `TopicName` in `pulsar-proxy` module. Instead of using a global cache, just call this method directly for the conversion.

It does not introduce any new configuration. From the benchmark, it's only 2.2x slower than the `TopicName#get`, when there are only 200k keys. But it's faster than `TopicName#get`.

However, it saves memory and the performance should be acceptable that it can be executed 8.5 millions in 1 second.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/46


pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
```diff
@@ -85,6 +87,9 @@ public class LookupProxyHandler {
             .create().register();
     private final Semaphore lookupRequestSemaphore;
 
+    // Maps the topic name from the request to the full topic name
+    private final Map<String, String> topicNameCache = new HashMap<>();
```
codelipenghui: When multiple connections access the same topic, will it cause more heap memory occupation?

BewareMyPower: Yes. However, once a connection is disconnected, the cache will be garbage collected. In extreme cases, if many consumers subscribe 

How about removing the cache? This conversion is actually very fast. It's only about 2.2x slower than reading from a `TopicName` cache.

```
TopicNameBenchmark.testConvert        thrpt       284.114          ops/s
TopicNameBenchmark.testReadFromCache  thrpt       644.718          ops/s
```

and it can execute 30000*284 operations in 1 second.

image

BewareMyPower: i.e. each operation only spends only 0.12 **MICROSECONDS**. The "space-for-time tradeoff" might not worth here.

BewareMyPower: I removed the cache and updated the PR description

BewareMyPower: image

Well, if there are 500k keys, `TopicName#get` is slower.

```
TopicNameBenchmark.testConvert        thrpt       5.857          ops/s
TopicNameBenchmark.testReadFromCache  thrpt       5.326          ops/s
```

BewareMyPower: 200k keys:

```
TopicNameBenchmark.testConvert        thrpt       14.706          ops/s
TopicNameBenchmark.testReadFromCache  thrpt       16.639          ops/s
```

nodece (approved):
LGTM, I think we can move `toFullTopicName` to the `TopicName` class.

BewareMyPower:
> I think we can move toFullTopicName to the TopicName class.

Yes. But let's keep it in `TopicNameUtils` for now in case there are some conflicts.

BewareMyPower:
I changed my idea and moved the method to `TopicName` because it has a duplicated method with https://github.com/apache/pulsar/pull/24463

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -442,4 +443,78 @@ public boolean includes(TopicName otherTopicName) {
...
+     * "<domain>://<tenant>/<cluster>/<namespace>/<local-topic>" (v1). However, for convenient, it's allowed for clients
+     * to pass a short topic name with v2 format:
+     * - "<local-topic>", which represents "persistent://public/default/<local-topic>"
+     * - "<tenant>/<namespace>/<local-topic>, which represents "persistent://<tenant>/<namespace>/<local-topic>"
+     *
+     * @param topic the topic name from client
+     * @return the full topic name.
+     */
+    public static String toFullTopicName(String topic) {
+        final int index = topic.indexOf("://");
```
3pacccccc, line 459: considering make the string '://' as a unified constant variable? I noticed that Pulsar has so many instances where the string '://' is hardcoded

3pacccccc (approved):

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
BewareMyPower:
```
% find . -name "*.java" | grep main | xargs grep -n "\"://\""        
./pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java:59:        String pkgName = type + "://" + tenant + "/" + namespace + "/" + name + "@" + version;
./pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java:64:        String pkgName = type + "://" + name + "@" + version;
./pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java:77:        if (!packageName.contains("://")) {
./pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java:81:        List<String> parts = Splitter.on("://").limit(2).splitToList(packageName);
./pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SnapshotSegmentAbortedTxnProcessorImpl.java:346:        if (!topics.contains(TopicDomain.persistent + "://" + topicName.getNamespace() + "/"
./pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LeaderBroker.java:46:        int uriSeparatorPos = serviceUrl.indexOf("://");
./pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java:235:        String newUri = (originalUri.split(":")[0] + "://")
./pulsar-common/src/main/java/org/apache/pulsar/common/net/ServiceURI.java:253:            sb.append("://");
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:66:        String name = domain + "://" + namespaceName.toString() + '/' + topic;
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:71:        String name = domain + "://" + tenant + '/' + namespace + '/' + topic;
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:77:        String name = domain + "://" + tenant + '/' + cluster + '/' + namespace + '/' + topic;
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:119:            if (!completeTopicName.contains("://")) {
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:125:                    completeTopicName = TopicDomain.persistent.name() + "://" + completeTopicName;
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:127:                    completeTopicName = TopicDomain.persistent.name() + "://"
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:140:            List<String> parts = Splitter.on("://").limit(2).splitToList(completeTopicName);
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:459:        final int index = topic.indexOf("://");
./pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java:462:            final List<String> parts = splitBySlash(topic.substring(index + "://".length()), 4);
./pulsar-common/src/main/java/org/apache/pulsar/common/functions/Utils.java:46:                && destPkgUrl.contains("://"));
./pulsar-common/src/main/java/org/apache/pulsar/common/topics/TopicList.java:39:    private static final String SCHEME_SEPARATOR = "://";
./pulsar-io/kafka-connect-adaptor/src/main/java/org/apache/pulsar/io/kafka/connect/AbstractKafkaConnectSource.java:248:            if (topic.contains("://")) {
./pulsar-io/azure-data-explorer/src/main/java/org/apache/pulsar/io/azuredataexplorer/ADXSinkUtils.java:24:    static final String PROTOCOL_SUFFIX = "://";
./pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/MetadataStoreFactoryImpl.java:118:        if (!metadataURL.contains("://")) {
./pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/oxia/OxiaMetadataStoreProvider.java:59:        if (metadataURL == null || !metadataURL.startsWith(urlScheme() + "://")) {
./pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java:63:        String hostUrl = serviceUri.getServiceScheme() + "://" + host.getHostString() + ":" + host.getPort();
./pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarServiceNameResolver.java:90:            String hostUrl = uri.getServiceScheme() + "://" + host;
```

It's true that this string is used in many places. But `://` is a standard separator between "scheme" and "path" in [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). Even if I made it a constant string like `TopicList#SCHEME_SEPARATOR` or `ADXSinkUtils#PROTOCOL_SUFFIX`, other people might not consider reusing it.

JVM could store it in a constant string pool no matter if you declare it as a static string, because `"://"` itself is a string literal.

Additionally, IMO,

```java
String name = domain + "://" + namespaceName.toString() + '/' + topic;
```

would have better readability than

```java
String name = domain + URI_SEPARATOR + namespaceName.toString() + SLASH + topic;
```

3pacccccc: make sense

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
BewareMyPower: @codelipenghui Could you take a look again?

dao-jun (approved):

poorbarcode (approved):

PR #24470, opened 2025-06-28
[improve][broker] PIP427:Align pulsar-admin Default for Mark-Delete Rate with Broker Configuration

codelipenghui:
### Motivation

Implementation of pip-427 #24425

### Modifications

- Adjusted the logic in BrokerService to set the throttle mark delete rate based on persistence policies, allowing for a fallback to the broker's default rate when necessary.
- Added comprehensive unit tests to validate the behavior of namespace and topic-level persistence policies, ensuring correct application of the managed ledger max mark delete rate.
- Updated CLI commands to accept a default value of -1 for the mark delete rate, indicating that it should use the broker's default configuration.
- Enhanced test coverage for CLI commands to verify the handling of various mark delete rate scenarios, including negative and zero values.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [x] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [x] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -2056,7 +2056,9 @@ public class BrokerService implements Closeable {
                 }
             }
 
-            managedLedgerConfig.setThrottleMarkDelete(persistencePolicies.getManagedLedgerMaxMarkDeleteRate());
+            managedLedgerConfig.setThrottleMarkDelete(persistencePolicies.getManagedLedgerMaxMarkDeleteRate() >= 0
+                    ? persistencePolicies.getManagedLedgerMaxMarkDeleteRate()
+                    : serviceConfig.getManagedLedgerDefaultMarkDeleteRateLimit());
             managedLedgerConfig.setDigestType(serviceConfig.getManagedLedgerDigestType());
             managedLedgerConfig.setPassword(serviceConfig.getManagedLedgerPassword());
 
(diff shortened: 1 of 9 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2056,7 +2056,9 @@ public CompletableFuture<ManagedLedgerConfig> getManagedLedgerConfig(@NonNull To
                 }
             }
 
-            managedLedgerConfig.setThrottleMarkDelete(persistencePolicies.getManagedLedgerMaxMarkDeleteRate());
+            managedLedgerConfig.setThrottleMarkDelete(persistencePolicies.getManagedLedgerMaxMarkDeleteRate() >= 0
+                    ? persistencePolicies.getManagedLedgerMaxMarkDeleteRate()
+                    : serviceConfig.getManagedLedgerDefaultMarkDeleteRateLimit());
```
3pacccccc, line 2061: I think  it might  be more reasonable to retain the last valid value for ```ThrottleMarkDelete``` rather than resetting it to the default 1.0? 
for example, last ```ThrottleMarkDelete``` I set is 75.0, then I use command ```set-persistence persistent://public/default/topic -e 2 -w 2 -a 2``` to reset the policy, I just want to change other fields and keep the ```ThrottleMarkDelete``` still be 75.

codelipenghui: Yes, that's reasonable, but it's out of the scope of this proposal.
Now, Pulsar doesn't support partial updates for a policy.

For this case, user might also want to use -1 to remove the setting of ThrottleMarkDelete on the namespace level.

PR #24478, opened 2025-07-02
[fix][broker]Data lost due to conflit loaded up a topic for two brokers, when enabled ServiceUnitStateMetadataStoreTableViewImpl

poorbarcode:
### Motivation

We encountered an issue that the ledger of a topic was lost after restarting ZKs

image

After checking brokers' logs, we found that both brokers loaded up the same topic.

image

And we found the following logs that printed by `broker-0`

```
2025-06-24T14:57:27,492+0000 [metadata-store-10-1] INFO  org.apache.bookkeeper.mledger.impl.ManagedLedgerFactoryImpl - Received MetadataStore session event: SessionLost
2025-06-24T14:58:20,273+0000 [pulsar-load-manager-1-1] WARN  org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImpl - Found this broker:logging-broker-0:8080 has not registered yet. Trying to register it
2025-06-24T14:58:35,067+0000 [CompletableFutureDelayScheduler] WARN  org.apache.pulsar.broker.loadbalance.extensions.channel.ServiceUnitStateChannelImpl - logging-broker-0:8080 failed to wait for owner for serviceUnit:xxx/xxx/0xca6d5a6c_0xd448fc50; Trying to return the current owner:Optional[logging-broker-0:8080]
java.util.concurrent.TimeoutException: null
	at java.base/java.util.concurrent.CompletableFuture$Timeout.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) ~[?:?]
```

### Root cause
- `broker-0` lost the ZK session
- Pulsar cluster switches the topic owner from `broker-0` to `broker-2`
- `broker-0` loses the events of bundle owner switching because it has lost the session
- Issue: it still maintains the old value of the bundle owner, which indicates `broker-0` is the topic's owner.

### Modifications

- Add new events when brokers lose the ZK session: itemOutdatedListeners
  - Broker will unload all bundles that received the event
  - Clear the cached, invalidated data of `MetadataStoreTableViewImpl` 
- Shutdown the broker if it can not recover the bundle states after the ZK session is reestablished.
- Keep the behavior the same as before if users do not implement the new event listener `itemOutdatedListeners`, which guarantees compatibility

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


codelipenghui:
@poorbarcode Do you know why Ledger 1049720 gets deleted? Or is it expected to be deleted?

poorbarcode:
@codelipenghui 

> @poorbarcode Do you know why Ledger 1049720 gets deleted? Or is it expected to be deleted?

I am not sure of the exact details, I guess there is a possibility:

Since two brokers assumed it is the owner, they overwrite the cursor's metadata, then we get a wrong cursor `md-position`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
```diff
@@ -69,12 +81,21 @@ public void start(PulsarService pulsar,
                 this::validateServiceUnitPath,
                 List.of(this::updateOwnedServiceUnits, tailItemListener),
                 List.of(this::updateOwnedServiceUnits, existingItemListener),
-                TimeUnit.SECONDS.toMillis(pulsar.getConfiguration().getMetadataStoreOperationTimeoutSeconds())
+                List.of(this::invalidateOwnedServiceUnits, itemOutdatedListeners),
```
Technoboy-: `itemOutdatedListeners` -> `outdatedItemListeners`

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
```diff
@@ -69,12 +81,21 @@ public void start(PulsarService pulsar,
...
+                t -> handleTableViewShutDownEvent(t)
         );
         tableview.start();
 
     }
 
+    protected void handleTableViewShutDownEvent(Throwable throwable) {
+        log.error("The component of load-balance, which named metadata store table view has shutdown. This Broker can"
+                    + " not work anymore, start tp shutdow,");
+        pulsar.shutdownNow();
```
codelipenghui, line 96: Is there any case that we will shutdown the table view without shutdown the broker?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableView.java
```diff
@@ -110,4 +118,9 @@ void start(PulsarService pulsar,
      * @throws TimeoutException
      */
     void flush(long waitDurationInMillis) throws ExecutionException, InterruptedException, TimeoutException;
+
+    /**
+     * Whether it depends on the local metadata store.
+     */
+    boolean isMetadataStoreBased();
```
codelipenghui, line 125: Something like expose the internal implementation to the API level. Do we have any other way to handle this case?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
```diff
@@ -54,13 +56,23 @@ public class ServiceUnitStateMetadataStoreTableViewImpl extends ServiceUnitState
...
         conflictResolver = new ServiceUnitStateDataConflictResolver();
         conflictResolver.setStorageType(MetadataStore);
+        if (!(pulsar.getLocalMetadataStore() instanceof AbstractMetadataStore)
+            && !MetadataSessionExpiredPolicy.shutdown.equals(pulsar.getConfig().getZookeeperSessionExpiredPolicy())) {
+            String errorMsg = "Your current metadata store does not support the registration of session event"
+                    + " listeners. Please set \"zookeeperSessionExpiredPolicy\" to \"SHUTDOWN\"; otherwise, you will"
+                    + " encounter the issue that messages lost because of conflicted topic loading";
+            log.error(errorMsg);
+            throw new IllegalStateException(errorMsg);
+        }
```
codelipenghui: Is it difficult to support retry policy? This cause will cause a behavior change which they will not able to start broker with retry policy?

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
poorbarcode: > Is there any case that we will shutdown the table view without shutting down the broker?

If other components use `TableViewImpl.java` instead of `system topic`, and the expired data can be tolerated, the broker does not need to shut down.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
poorbarcode: > Is it difficult to support retry policy?

Yes, we do not know what the implementation of `MetadataStore` is, so we do not know how we could determine whether the stats are outdated.

> This cause will cause a behavior change which they will not able to start broker with retry policy?

Good point, let me modify the log level to `warning` and remove the error throwing.

poorbarcode: >> This cause will cause a behavior change which they will not able to start broker with retry policy?

> Good point, let me modify the log level to warning and remove the error throwing.

Done

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableView.java
poorbarcode: It is `ServiceUnitStateTableView.java`, not `MetadataStoreTableView`, the interface will only be used by the broker load manager, so I think it is safe

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateMetadataStoreTableViewImpl.java
poorbarcode: Renamed

Technoboy- (approved):
LGTM

Demogorgon314 (approved):

heesung-sohn:
Hi guys!

I think this issue probably might have revealed a bigger problem.

In fact, by design, I think pulsar brokers can see temporary inconsistent ownership views because of network delay or zk session issues.

In this case, brokers could see conflict ownership views, but by the "pulsar fencing logic", only one broker should serve(write to) the topic, and the other brokers should close the topic/ledger from the fenced topic / metadata conflict exception. 

Can we confirm if pulsar fencing logic works fine?

heesung-sn:
Btw, when the metadata session is re-established, dont we already invalidate the cache at the MetadataCache layer? So, the ownership tableview is eventually synced/invalidated? It should periodically refresh its values, too?

heesung-sn:
And the original design was that pulsar brokers run as-is(best effort) when metadata connections/sessions are lost(worst case scenario). 

After this change, if all brokers lose metadata connections, will all bundles be unloaded(no topics will be served)?

heesung-sohn:
Regarding this warn

> 2025-06-24T14:58:35,067+0000 [CompletableFutureDelayScheduler] WARN  org.apache.pulsar.broker.loadbalance.extensions.channel.ServiceUnitStateChannelImpl - logging-broker-0:8080 failed to wait for owner for serviceUnit:xxx/xxx/0xca6d5a6c_0xd448fc50; Trying to return the current owner:Optional[logging-broker-0:8080]

I think we can consider failing new lookups/ownership assignments in this situation, when the broker lost metadata connections.

Returning the current candidate "owner" from the local info was added to improve availability, but it appears that we better optimize for consistency.

heesung-sn:
> Btw, when the metadata session is re-established, dont we already invalidate the cache at the MetadataCache layer? So, the ownership tableview is eventually synced/invalidated? It should periodically refresh its values, too?

Maybe we can improve the code to sync the tableview with metadata cache better, especially when the metadata cache is invalidated all. 

Or, if this additional sync logic makes the code too complicated, we could consider removing this shadow hashmap but directly using the metadata cache for the metadata store tableview. (This requires handling CompletableFuture more gracefully)

heesung-sn:
> Btw, when the metadata session is re-established, dont we already invalidate the cache at the MetadataCache layer? So, the ownership tableview is eventually synced/invalidated? It should periodically refresh its values, too?

Now I think this metadatastore, metadatacache, metadatatableview sync logic makes the code too complicated and fragile. We could consider removing this shadow hashmap(the hashmap in the tableview) and directly using the metadata cache for the metadata store tableview. (This requires handling CompletableFuture more gracefully)

heesung-sohn:
Now I think this metadatastore, metadatacache, metadatatableview sync logic makes the code too complicated and fragile. We could consider removing this shadow hashmap(the hashmap in the tableview) and directly using the metadata cache for the metadata store tableview. (This requires handling CompletableFuture more gracefully)

poorbarcode:
Thanks @heesung-sn 

> In this case, brokers could see conflict ownership views, but by the "pulsar fencing logic", only one broker should serve(write to) the topic, and the other brokers should close the topic/ledger from the fenced topic / metadata conflict exception.
> Can we confirm if pulsar fencing logic works fine?

It does not work fine, because the first owner has loss notifications because of a session lost.

> I think we can consider failing new lookups/ownership assignments in this situation, when the broker lost metadata connections.
> Returning the current candidate "owner" from the local info was added to improve availability, but it appears that we better optimize for consistency.

Good pointer, the current PR has fixed it. See also https://github.com/apache/pulsar/pull/24478/files#diff-191ed1df4f5804c8fd6cdaf909cca01d071a110c3e701bdc98352f99e7a92e8eR469

> Now I think this metadatastore, metadatacache, metadatatableview sync logic makes the code too complicated and fragile. We could consider removing this shadow hashmap(the hashmap in the tableview) and directly using the metadata cache for the metadata store tableview. (This requires handling CompletableFuture more gracefully)

The map has another useful case: we need to unload owned topics once the broker state is not fine, the map maintains which bundle is owned by the current broker.

heesung-sohn:
> It does not work fine, because the first owner has loss notifications because of a session lost.

Interesting. I thought the fencing logic is guarded by the quorum of bk, regardless of zk.

heesung-sohn:
> The map has another useful case: we need to unload owned topics once the broker state is not fine, 

In fact, the original design was when the metadata store is unstable, pulsar should behave as-is at the best effort basis(minimal impact on existing clients) This "unloading all topics when lost zk connection" seems to be a new behavior.

> the map maintains which bundle is owned by the current broker.

Ok. One brute-force way to sync cache and tv in this case is to re-init the tableview(refresh the tableview again when connection is reestablished.) I think you already added this logic in this PR.

heesung-sohn:
@poorbarcode I raised this PR to print warn log when they are out of sync, https://github.com/apache/pulsar/pull/24513

PR #24481, opened 2025-07-03
[improve][pip] PIP-432: Add isEncrypted field to EncryptionContext

codelipenghui:
### Motivation

Applications using Pulsar's encryption feature with `ConsumerCryptoFailureAction.CONSUME` need to determine whether received messages were successfully decrypted or if decryption failed. This is essential for:

1. **Error handling**: Applications need to know when they receive encrypted (undecrypted) data to handle it appropriately
2. **Monitoring**: Applications want to track decryption success/failure rates for monitoring and alerting
3. **Manual decryption**: When automatic decryption fails, applications may want to attempt manual decryption using the EncryptionContext
4. **Security compliance**: Applications need to ensure they're not inadvertently processing encrypted data as plain text

**Current situation:**
- Consumers with `CONSUME` action receive messages regardless of decryption success
- No programmatic way to distinguish between successfully decrypted and failed decryption messages
- Applications must implement workarounds to detect encrypted vs. decrypted content

**Use cases this solves:**
1. Consumer without private key configured → should know decryption failed
2. Consumer with mismatched private key → should know decryption failed  
3. Consumer with correct private key → should know decryption succeeded

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


pip/pip-432 Add isEncrypted field to EncryptionContext.md
```diff
@@ -0,0 +1,120 @@
...
+
+- Changes to encryption/decryption algorithms or protocols
+- Modifications to `ConsumerCryptoFailureAction` behavior
+- Performance improvements to encryption/decryption operations
+- New encryption features or capabilities
+- Changes to producer-side encryption logic
+
+# High Level Design
+
+The solution adds a simple boolean field `isEncrypted` to the existing `EncryptionContext` class. This field is set during message processing in the consumer:
```
RobertIndie, line 61: The current field name `isEncrypted` creates semantic ambiguity. Users cannot distinguish whether this represents:
1. The message is an encrypted message (vs. plain text)
2. The message failed to decrypt (vs. successfully decrypted)

This ambiguity becomes problematic for use cases like counting encrypted messages consumed by a consumer - `isEncrypted=true` could mean either "this is an encrypted message" or "decryption failed", leading to incorrect statistics.

While users can check if `EncryptionContext` is empty to determine if it's an encrypted message, this is not intuitive from an API design perspective.

**Recommendation**: Use `isDecrypted` or `isFailedDecrypted` instead. This clearly indicates whether the current message was successfully decrypted or failed to decrypt, eliminating any ambiguity. The field name should reflect the actual state the user cares about: "Can I use this message payload directly, or do I need to handle it as encrypted data?"

codelipenghui: @RobertIndie Initially, I used `isDecryptionFailed`, but I realized that users only need to know if decryption needs to be applied again. And adding `isFailedDecrypted` to `EncryptionContext` is also confusing.

codelipenghui: > The current field name isEncrypted creates semantic ambiguity. Users cannot distinguish whether this represents:

> The message is an encrypted message (vs. plain text)
> The message failed to decrypt (vs. successfully decrypted)

And actually, they can.

- Plain test means there is no EncryptionContext
- Failed to decryped (EncryptionContext.isEncryped is true)
- Decrypted data (EncryptionContext.isEncryped is false)

RobertIndie: Thanks for your explaination. But seems both are confusing to me. I'm fine with the current design. We can add the doc to avoid confusion.

pulsar-client-api/src/main/java/org/apache/pulsar/common/api/EncryptionContext.java
```diff
@@ -39,6 +39,7 @@ public class EncryptionContext {
     private CompressionType compressionType;
     private int uncompressedMessageSize;
     private Optional<Integer> batchSize;
+    private boolean isEncrypted;
```
RobertIndie, line 43: It's better to add a doc to explain this field.

3pacccccc:
LGTM. Additionally, could we introduce a new decryptFailListener to handle decryption failures, instead of using if-else checks? This would provide a more elegant solution. The code might look like this:
```
pulsarClient.newConsumer()
    .topic(topicName)
    .subscriptionName(subName)
    .messageListener((consumer, msg) -> {
        // Handle successfully decrypted messages
        handleDecryptedMessage(msg);
    })
    .decryptFailListener((consumer, msg) -> {
        // Handle decryption failures
        handleEncryptedMessage(msg);
    });
```

codelipenghui:
@3pacccccc Yes, that's a good idea to have a listener for the decryption failures. Do you want to create a proposal for it? With the decryption failure listener, we can suggest users to use listener instead of ConsumerCryptoFailureAction. They can decide in the listener to discard, or decrypt, or anything they want.

pulsar-client-api/src/main/java/org/apache/pulsar/common/api/EncryptionContext.java
codelipenghui: Add @RobertIndie

pip/pip-432 Add isEncrypted field to EncryptionContext.md
codelipenghui: Added @RobertIndie

3pacccccc:
> @3pacccccc Yes, that's a good idea to have a listener for the decryption failures. Do you want to create a proposal for it? With the decryption failure listener, we can suggest users to use listener instead of ConsumerCryptoFailureAction. They can decide in the listener to discard, or decrypt, or anything they want.

yes，I‘m sooo glad if I can do anything to help

pip/pip-432 Add isEncrypted field to EncryptionContext.md
BewareMyPower: This field name is very confusing, I'd like to use `isUndecryptedPayload` instead

image

PR #24484, opened 2025-07-03
[fix][io] Ensure infinite retention for Debezium history topic

shibd:
### Motivation

The Debezium DatabaseHistory mechanism requires a complete and immutable log of all DDL changes to correctly reconstruct table schemas upon connector restart. The current PulsarDatabaseHistory implementation stores this history in a Pulsar topic but does not enforce a retention policy.

### Modifications
Create a durable subscription and never receive/ack message to make sure history topic data is retained.  The data is actually very small. Additionally, we can also flexibly control this subscription's cursor to fix or skip some history data.

### Verifying this change
- You can reproduce this issue by testing: `PulsarDatabaseHistoryTest`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-io/debezium/core/src/main/java/org/apache/pulsar/io/debezium/PulsarDatabaseHistory.java
+++ b/pulsar-io/debezium/core/src/main/java/org/apache/pulsar/io/debezium/PulsarDatabaseHistory.java
@@ -100,6 +100,8 @@ public final class PulsarDatabaseHistory extends AbstractDatabaseHistory {
         DatabaseHistory.NAME,
         READER_CONFIG);
 
+    protected static final String DURABLE_SUB_NAME = "durable-sub";
+
     private final ObjectMapper mapper = new ObjectMapper();
     private final DocumentReader reader = DocumentReader.defaultReader();
     private String topicName;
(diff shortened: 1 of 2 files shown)


pulsar-io/debezium/core/src/main/java/org/apache/pulsar/io/debezium/PulsarDatabaseHistory.java
```diff
@@ -165,6 +167,15 @@ public void initializeStorage() {
             log.error("Failed to initialize storage", pce);
             throw new RuntimeException("Failed to initialize storage", pce);
         }
+        try (org.apache.pulsar.client.api.Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING)
+                .subscriptionName(DURABLE_SUB_NAME).topic(topicName).subscribe()) {
```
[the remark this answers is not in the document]
shibd: No need to set initial position

pulsar-io/debezium/core/src/main/java/org/apache/pulsar/io/debezium/PulsarDatabaseHistory.java
```diff
@@ -100,6 +100,8 @@ public final class PulsarDatabaseHistory extends AbstractDatabaseHistory {
         DatabaseHistory.NAME,
         READER_CONFIG);
 
+    protected static final String DURABLE_SUB_NAME = "durable-sub";
```
codelipenghui, line 103: It might also confuse user if they set a specific data retention.
Is it better to use retention policy.

For example, if the topic is not created, the connector can create the topic with infinite data retention by default.

shibd: Under the current functionality, this topic must have infinite retention. Allowing users to set a custom data retention policy could lead to incomplete data and recovery failures.

Furthermore, if we were to manage this using data retention policies, it would introduce a dependency on pulsar-admin and likely change the permission requirements for the user when creating a connector.

Therefore, using a durable subscription is a simple and effective solution.

nodece: @shibd If message TTL and retention are already configured on the namespace level, your change is meaningless. To achieve infinite retention, it's recommended to use pulsar-admin and set the retention value to 0, which disables retention.

shibd: hi, @nodece thanks for the reminder, you're right.

Unfortunately, for the connector to use pulsar-admin, it would require more interface support. Furthermore, we typically only need to grant `produce` and `consume` permissions to create a connector, but this approach would also require users to grant `admin` permissions.

I think a more direct and simple solution is to modify the documentation and tell users to manually set the retention policy before they create the Debezium source connector.

Issue #24417, opened 2025-06-16
[Bug] Topic does not exist in the remote cluster for non-paritioned topics


saxena-gaurav:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

Broker version:  4.0.4

### Issue Description

Non-Paritioned Topics are not getting created in the remote cluster if topic auto-creation is disabled.  But partition topics work fine. 

### Error messages

```text

```

### Reproducing the issue

1. Create geo-replicating cluster
2. Disable topic auto-creation on the remote cluster.
3. Create non-parttion topic on the local cluster.
4. Send message to the topic on the local cluster.
5. Message will not replicate to the remote cluster.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

zihenzzz:
i want to fix it

poorbarcode:
@zihenzzz Could you show your solution first?

zihenzzz:
> [@zihenzzz](https://github.com/zihenzzz) Could you show your solution first?

I think this is because Pulsar handles partitioned and non-partitioned topics differently: Partitioned topics: When replication is enabled, Pulsar actively creates the topic in the remote cluster without relying on the auto-creation feature. Non-partitioned topics: The replicators work like ordinary producers and rely on the auto-creation feature of the remote cluster. If the remote cluster has disabled auto-creation, the topic cannot be created, leading to replication failure. 
The simplest solution - enable automatic creation on the remote cluster. The safest solution - manually create on both sides.

poorbarcode:
@zihenzzz 

> The simplest solution - enable automatic creation on the remote cluster. The safest solution - manually create on both sides.

Good point, but the solution we should implement is to let replication work for the scenario where users are switching the policy of topic autop-creation

zihenzzz:
> [@zihenzzz](https://github.com/zihenzzz)
> 
> > The simplest solution - enable automatic creation on the remote cluster. The safest solution - manually create on both sides.
> 
> Good point, but the solution we should implement is to let replication work for the scenario where users are switching the policy of topic autop-creation

I have a general understanding. If possible, I can try to fix it

poorbarcode:
[...]

PR #21679, opened 2023-12-06
[fix][broker] Automatically create topic for remote cluster in geo

nodece:
Fixes https://github.com/apache/pulsar/issues/21153

### Motivation

Enable the geo-replication only on the topic level to replicate the partitioned topic, if the topic was not created in the remote cluster, it becomes the non-partition topic by the `allowAutoTopicCreation=true` and `allowAutoTopicCreationType=non-partitioned` configurations, which is not in line with the expected behavior. Therefore, I suggest that geo-replication automatically creates the topic for the remote cluster.

How to reproduce this issue:

1. Create two clusters(cluster-a and cluster-b) and use the different configuration stores.
2. Add cluster data to each cluster
3. Create a tenant in each cluster:
```
bin/pulsar-admin tenants create my-tenant --allowed-clusters cluster-a,cluster-b
```
5. Create a namespace in the cluster-a: 
```
bin/pulsar-admin namespaces set-clusters my-tenant/my-namespace --clusters cluster-a
```
7. Create a namespace in the cluster-b
```
bin/pulsar-admin namespaces set-clusters my-tenant/my-namespace --clusters cluster-b
```
9. Create a topic and enable the geo-replication in the cluster-a:
```
bin/pulsar-admin topics create-partitioned-topic -p 5 my-tenant/my-namespace/geo-p-5
bin/pulsar-admin topics set-replication-clusters --clusters cluster-a,cluster-b my-tenant/my-namespace/geo-p-5
```
10. List the partitioned topics in the cluster-b:
```
bin/pulsar-admin topics list-partitioned-topics my-tenant/my-namespace
empty
```

### Modifications

- For `NonPersistentTopic` and `PersistentTopic`, the geo automatically creates the topic for the remote cluster.
- Use the different configuration stores in the geo-replication test

### Verifying this change

- Add a `GeoReplicationOnTopicLevelTest` to cover these changes.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24485, opened 2025-07-03
[improve][pip] PIP-433: Optimize the conflicts of the replication and automatic creation mechanisms, including the automatic creation of topics and schemas

poorbarcode:

- [Discussion](https://lists.apache.org/thread/p16gwhfx6rkxdp8dm9pckn43o5875o1s)
- [Vote](https://lists.apache.org/thread/3nwsbqlkgorswr1oynjwmcz6blkkl5vm)

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pip/pip-433.md
+++ b/pip/pip-433.md
@@ -32,7 +32,7 @@ The internal producer of the geo replicator starts with an `auto-produce` schema
 **Steps to reproduce the issue**
 - Configurations of both the source cluster and the remote cluster 
   - `allowAutoTopicCreation`: `true`
-  - `allowAutoTopicCreationType`: `non-partitioned`
+  - `allowAutoTopicCreationType`: `partitioned`
   - `defaultNumPartitions`: `2`
 - The namespace `public/default` exists, but you have not enabled Geo-Replication for the namespace yet.
 - Start a producer for a topic `persistent://public/default/topic` on the source cluster.


poorbarcode:
The PIP also solves the issue https://github.com/apache/pulsar/issues/24417

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+- `Client of source cluster`: start a consumer/producer for a topic
+- `Client of source cluster`: Get partitions for topic
+    - At this step, the client will try to get the partitioned metadata of the topic. If the topic does not exist, the broker will create the partitioned metadata automatically.
+- `Client of source cluster`: Start internal consumers/producers for each sub-topic
+    - At this step, the client will try to connect to the sub-topic. If the sub-topic does not exist, the broker will create the partitioned metadata automatically.
+- `Source broker`: starts the geo replicator when a sub-topic is loading up. 
+  - The geo replicator maintains an internal producer for the topic under the remote cluster.
+  - The internal producer is a single sub-topic producer; it will not trigger a partitioned metadata auto-creation.
+  - When starting the internal producer, it confirms that the target topic under the remote cluster should be a non-partitioned topic or a sub-topic.
+  - **(Highlight)** Otherwise, prints error logs and stops to start.
```
codelipenghui: If the topic creation is triggered by the clients, the default partitions of topic auto-creation will be used in this case?

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+- The source broker checks whether the partitioned topic exists under the remote cluster
+- If yes, the source broker compares both partition counts, enabling the topic-level replication if both clusters have the same number of partitions; otherwise, you will get a bad request error.
+- If not, it will create the partitioned topic with the same partitions count automatically under the remote cluster.
+
+#### Topic auto-creation by clients if you have enabled Geo Replication.
+- `Client of source cluster`: start a consumer/producer for a topic
+- `Client of source cluster`: Get partitions for topic
+    - At this step, the client will try to get the partitioned metadata of the topic. If the topic does not exist, the broker will create the partitioned metadata automatically.
+- `Client of source cluster`: Start internal consumers/producers for each sub-topic
+    - At this step, the client will try to connect to the sub-topic. If the sub-topic does not exist, the broker will create the partitioned metadata automatically.
```
codelipenghui: What does sub topic means here? internal partition of a partitioned topic?

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+
+**The original design of `pulsar-admin topics set-is-allow-auto-update-schema`**
+
+```shell
+pulsar-admin namespaces set-is-allow-auto-update-schema
+  -d, --disable         Disable schema validation enforced
+  -e, --enable          Enable schema validation enforced
+```
+
+To add a new param `--always-allow-for-replicator-if-compatible`, which means that always allow the replicator to register a new schema if compatible. The default value is `true`.
```
codelipenghui: Do we need this option in broker.conf?

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+  -n, --num-partitions=<defaultNumPartitions>
+                        The default number of partitions of a topic to be
+                          auto-created, applicable to partitioned topics only
+  -t, --type=<type>     Type of topic to be auto-created. Possible values:
+                          (partitioned, non-partitioned). Default value:
+                          non-partitioned
+                          Default: non-partitioned
+```
+
+To add a new param `--always-allow-for-replicator`, which means that always allow the replicator to trigger a topic auto-creation, including partitioned topics. The default value is `true`.
```
codelipenghui: Do we need this option in broker.conf?

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+
+# Goals
+
+1. Add an optional choice: always allow the replicator to register schemas if compatible, even if users set `set-is-allow-auto-update-schema` to false.
+2. Add an optional choice: always allow the replicator to trigger a topic auto-creation, including partitioned topics, even if users have disabled topic auto-creation.
+  a. It only triggers auto-creation if the remote side has the same auto-creation policy of partition-type and partition-num.
+3. Checks compatibility between two clusters when enabling namespace-level replication, which includes the following
+  a. All topics’ partitions that have been created should be the same, including `__change_events`
+  b. Auto-creation policies should be the same
+  c. Support for adjusting policies automatically when enabling namespace-level policies, and provide an API to review the changes that will be applied. By the way, this feature will be an option param of enabling namespace-level geo replication
```
codelipenghui: I think the geo-replication should get rid of the topic auto-creation. So we don't need to sync the policies.

poorbarcode: > I think the geo-replication should get rid of the topic auto-creation. So we don't need to sync the policies.

This change is used to support the topics that already exist before enabling a namespace-level Geo-Replication. For example:

- Both the source cluster and the remote cluster have a namespace `public/default`
- Source cluster has a system topic `public/default/__change_events`, which is a non-partitioned topic
- Remote cluster has a system topic `public/default/__change_events`, which is a partitioned topic with `2` partitions.
- At the moment, user wants to enable a namespace-level Geo-Replication for `public/defaut`

pip/pip-433.md
poorbarcode: > What does sub topic means here? internal partition of a partitioned topic?

Yes, a internal partition of a partitioned topic

poorbarcode: Thanks for mentioning this, I added a description such as `sub-topic(an internal partition of a partitioned topic)`

pip/pip-433.md
poorbarcode: > If the topic creation is triggered by the clients, the default partitions of topic auto-creation will be used in this case? 

Yes. The goal 3-c is in order to solve this issue

pip/pip-433.md
poorbarcode: After this PIP, the broker-level configuration `isAllowAutoUpdateSchemaEnabled ` will never affect the behavior of the replicator anymore. See also: https://github.com/apache/pulsar/pull/24485/files#diff-46571a0d56c1897159ba02ef4830473ce8de4a6d0902e3678db1eead5beaed36R143-R144

> ### Configuration
> #### Regarding the goal-1 and goal-2
> - The configuration `broker.conf -> allowAutoTopicCreation` will never affect the behavior of the replicator anymore.
> - The configuration `broker.conf -> isAllowAutoUpdateSchemaEnabled` will never affect the behavior of the replicator anymore.

pip/pip-433.md
poorbarcode: After this PIP, the broker-level configuration `allowAutoTopicCreation` will never affect the behavior of the replicator anymore. See also: https://github.com/apache/pulsar/pull/24485/files#diff-46571a0d56c1897159ba02ef4830473ce8de4a6d0902e3678db1eead5beaed36R143-R144

> ### Configuration
> #### Regarding the goal-1 and goal-2
> - The configuration `broker.conf -> allowAutoTopicCreation` will never affect the behavior of the replicator anymore.
> - The configuration `broker.conf -> isAllowAutoUpdateSchemaEnabled` will never affect the behavior of the replicator anymore.

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+
+The source broker will do the following things when you try to enable a topic-level Geo-Replication, and you did not enable a namespace-level Geo-Replication.
+- The source broker checks whether the partitioned topic exists under the remote cluster
+- If yes, the source broker compares both partition counts, enabling the topic-level replication if both clusters have the same number of partitions; otherwise, you will get a bad request error.
+- If not, it will create the partitioned topic with the same partitions count automatically under the remote cluster.
+
+#### Topic auto-creation by clients if you have enabled Geo Replication.
+- `Client of source cluster`: start a consumer/producer for a topic
+- `Client of source cluster`: Get partitions for topic
+    - At this step, the client will try to get the partitioned metadata of the topic. If the topic does not exist, the broker will create the partitioned metadata automatically.
```
nodece, line 17: The broker didn't create the partitioned metadata automatically when `org.apache.pulsar.client.impl.conf.ProducerConfigurationData#isNonPartitionedTopicExpected` is `true`.

Please check https://github.com/apache/pulsar/blob/v4.0.5/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java#L448

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+- `Client of source cluster`: Start internal consumers/producers for each sub-topic(an internal partition of a partitioned topic)
+    - At this step, the client will try to connect to the sub-topic. If the sub-topic does not exist, the broker will create the partitioned metadata automatically.
```
nodece: If the sub-topic does not exist, the broker will create the sub-topic automatically, and this requires that the partitioned metadata exists.

Please see https://github.com/apache/pulsar/pull/24213
﻿

pip/pip-433.md
nodece: Right now, we have two ways to create the partitioned metadata.

1. When a partitioned topic is created on the source cluster, it will be synced to the remote cluster.
2. Create the same partitioned topic on the remote cluster by pulsar-admin.

Your description:
> The source broker will copy the REST requests that is a partitioned topic creation to the remote cluster if you have already enabled a namespace-level Geo Replication.

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+- `Source broker`: starts the geo replicator when a sub-topic is loading up. 
+  - The geo replicator maintains an internal producer for the topic under the remote cluster.
+  - The internal producer is a single sub-topic producer; it will not trigger a partitioned metadata auto-creation.
+  - When starting the internal producer, it confirms that the target topic under the remote cluster should be a non-partitioned topic or a sub-topic.
+  - **(Highlight)** Otherwise, prints error logs and stops.
+
+#### Schemas replication if you have enabled Geo Replication.
+The internal producer of the geo replicator starts with an `auto-produce` schema, copies a new schema it reads from the source cluster to the remote cluster, and it will be stuck once a schema is incompatible to the remote cluster.
```
nodece, line 27: The GEO replicator can copy the message with the schema, but I didn't check how it works.
```
msg.setSchemaInfoForReplicator(schemaFuture.get());
```

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+  - `allowAutoTopicCreationType`: `non-partitioned`
+  - `defaultNumPartitions`: `2`
+- The namespace `public/default` exists, but you have not enabled Geo-Replication for the namespace yet.
+- Start a producer for a topic `persistent://public/default/topic` on the source cluster.
+  - It triggers a partitioned topic with `2` partitions created.
+- Enable namespace-level Geo-Replication for the namespace `public/default`.
+- Start geo replicator for the existing sub-topic `public/default/topic-partition-0`
+  - Without [PIP-414: Enforce topic consistency check](https://github.com/apache/pulsar/pull/24213), the geo replicator will trigger a non-partitioned topic creation, which is named `public/default/topic-partition-0`
+  - With [PIP-414: Enforce topic consistency check](https://github.com/apache/pulsar/pull/24213), the geo replicator get a denied error.
+- However, the user wants to allow the replicator to copy messages to the remote cluster.
```
nodece, line 44: The partitioned topic requires partitioned metadata on the remote cluster, so you got a denied error.

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+**Steps to reproduce the issue**
+- The topic`public/default/topic` has enabled a geo replication.
+- Users controls the topic schema manually, do not allow auto update schema by consumers/producers for both clusters.
+- The internal producer is stuck after the user sets a new schema on the source cluster.
+- However, the user wants to allowthe  replicator to copy the schema to the remote cluster.
+
+# Goals
+
+1. Add an optional choice: always allow the replicator to register schemas if compatible, even if users set `set-is-allow-auto-update-schema` to false.
+2. Add an optional choice: always allow the replicator to trigger a topic auto-creation, including partitioned topics, even if users have disabled topic auto-creation.
```
nodece: You're already introducing `org.apache.pulsar.broker.ServiceConfiguration#createTopicToRemoteClusterForReplication`.

Alternative: https://github.com/apache/pulsar/pull/21679

We should also introduce dedicated metrics or statistics to surface internal replicator errors, which are often the root cause when geo-replication fails.

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+  a. It only triggers auto-creation if the remote side has the same auto-creation policy of partition-type and partition-num.
+3. Checks compatibility between two clusters when enabling namespace-level replication, which includes the following
+  a. All topics’ partitions that have been created should be the same, including `__change_events`
```
nodece: It looks like the users want to use global policies. When the system topic type is different between the source and remote clusters, we will get an error.

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+
+```shell
+pulsar-admin namespaces set-is-allow-auto-update-schema
+  -d, --disable         Disable schema validation enforced
+  -e, --enable          Enable schema validation enforced
+```
+
+To add a new param `--always-allow-for-replicator-if-compatible`, which means that always allow the replicator to register a new schema if compatible. The default value is `true`.
+
+#### Regarding the goal-2: "always allow the replicator to trigger a topic auto-creation"
```
nodece: Please see https://github.com/apache/pulsar/pull/24485/files#r2189536354

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+- However, the user wants to allow the replicator to copy messages to the remote cluster.
+
+#### Issue 2: Replication is stuck because the remote side does not allow schema updates
+**Steps to reproduce the issue**
+- The topic`public/default/topic` has enabled a geo replication.
+- Users controls the topic schema manually, do not allow auto update schema by consumers/producers for both clusters.
+- The internal producer is stuck after the user sets a new schema on the source cluster.
+- However, the user wants to allowthe  replicator to copy the schema to the remote cluster.
+
+# Goals
```
nodece, line 53: In my opinion, we should not rely on the remote cluster's topic auto-creation policies to determine whether replication can proceed. Usually, the configurations of all clusters involved in geo-replication should be consistent, including topic type, auto-creation settings, and default partition numbers.

When the source cluster wants to replicate messages to the remote cluster, it must ensure the topic already exists on the remote side and matches the expected type (e.g., partitioned vs non-partitioned). If the topic does not exist or has a mismatched type, the source cluster should disconnect the replicator and expose a clear error, rather than silently trying to recover or infer remote settings.

> Introducing additional configuration or logic to handle inconsistencies between clusters can add unnecessary complexity, increase coupling between clusters, and make the system harder to operate and debug. Instead, we should prioritize:

> * **Fail-fast behavior** when topic types are inconsistent.
> * **Better observability and error reporting**, so users can identify and fix misconfigurations early.

Overall, I believe operational safety and transparency are more important than automation in this case. Making replication logic depend on remote cluster policies could hide real problems and lead to hard-to-diagnose issues in production.

pip/pip-433.md
poorbarcode: > The broker didn't create the partitioned metadata automatically when org.apache.pulsar.client.impl.conf.ProducerConfigurationData#isNonPartitionedTopicExpected is true.
> Please check https://github.com/apache/pulsar/blob/v4.0.5/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java#L448

The behavior I explained this line is the behavior of source cluster, please see the prefix of this line `Client of source cluster`

pip/pip-433.md
poorbarcode: The behavior I explained this line is the behavior of source cluster, please see the prefix `Client of source cluster`

pip/pip-433.md
poorbarcode: Please read the document carefully: `once a schema is incompatible to the remote cluster.`

pip/pip-433.md
poorbarcode: Please read the document carefully: I explained two behaviors that contain PIP-414 or not

> - Start geo replicator for the existing sub-topic `public/default/topic-partition-0`
>  - Without [PIP-414: Enforce topic consistency check](https://github.com/apache/pulsar/pull/24213), the geo replicator will trigger a non-partitioned topic creation, which is named `public/default/topic-partition-0`
>  - With [PIP-414: Enforce topic consistency check](https://github.com/apache/pulsar/pull/24213), the geo replicator get a denied error.

pip/pip-433.md
poorbarcode: Please read the document carefully:  the improvement that you said has been explained in the section " Regarding the goal-3: "checks compatibility between two clusters when enabling namespace-level replication""

pip/pip-433.md
poorbarcode: > You're already introducing org.apache.pulsar.broker.ServiceConfiguration#createTopicToRemoteClusterForReplication.

If you want to discuss PIP-370, please leave comments under that mailing list. They are separate features

> We should also introduce dedicated metrics or statistics to surface internal replicator errors, which are often the root cause when geo-replication fails.

This PIP does not involve this feature

pip/pip-433.md
poorbarcode: > It looks like the users want to use global policies. When the system topic type is different between the source and remote clusters, we will get an error.

It does not relate to global policies. Once you enable a namespace-level geo-replication, the system topic will be replicated to the remote side

pip/pip-433.md
poorbarcode: > Please see https://github.com/apache/pulsar/pull/24485/files#r2189536354

Answered that comment

pip/pip-433.md
nodece: Ok, got. 

>     - At this step, the client will try to get the partitioned metadata of the topic. If the topic does not exist, the broker will create the partitioned metadata automatically.

This behavior depends on the following configuration settings:
```
allowAutoTopicCreation=true
allowAutoTopicCreationType=partitioned
defaultNumPartitions=1
```

If these settings are configured as shown above, and the topic does not exist, the broker will automatically create it as a partitioned metadata with one partition.

Please note that the broker creates the non-partitioned topic by default.

pip/pip-433.md
nodece: We can disable replication for the system topic.

pip/pip-433.md
nodece: > If you want to discuss PIP-370, please leave comments under that mailing list. They are separate features

We already have the `createTopicToRemoteClusterForReplication` config, which controls whether to auto-create topics on the remote cluster when creating a partitioned topic. This config should be leveraged instead of adding a new one.

pip/pip-433.md
codelipenghui: @poorbarcode We could use partition and partitioned topic directly instead of introducing new concept.

pip/pip-433.md
codelipenghui: I mean this is a new configuration added to the namespace policy, do we need the same configuration to control on the broker level?

pip/pip-433.md
poorbarcode: > We can disable replication for the system topic.

Seems you want to remove [PIP-92](https://github.com/apache/pulsar/issues/13444), please start a new proposal for it, the current PIP does not includes this change

pip/pip-433.md
poorbarcode: > We already have the createTopicToRemoteClusterForReplication config, which controls whether to auto-create topics on the remote cluster when creating a partitioned topic. This config should be leveraged instead of adding a new one.

It seems that you have misunderstood some concepts
- `createTopicToRemoteClusterForReplication`: whether to start replicator in the current cluster if enabled replication that replicates messages to the remote side
- `the current PIP`: whether it always allows the topic auto-creation from an outside producer of a replicator

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+
+# Alternatives
+Nothing.
+
+# Links
+
+<!--
+Updated afterwards
+-->
+* Mailing List discussion thread: https://lists.apache.org/thread/p16gwhfx6rkxdp8dm9pckn43o5875o1s
```
poorbarcode, line 150: @nodece 

Start a channel to discuss the following comment, which you left [here](https://github.com/apache/pulsar/pull/24485#pullrequestreview-2995985140)

> This PIP attempts to work around user misconfigurations rather than enforce proper system boundaries. As a result, it introduces hidden complexity and encourages fragile, non-transparent behavior. The main issues are:

> 1. Replicator relies on implicit topic auto-creation on the remote cluster, assuming the auto-creation policies (like number of partitions and topic type) match. But this does not solve the root problem. Remote cluster policies may change at any time, and relying on them does not eliminate the risk of topic-type mismatches.

Once the user changed the policy, the behavior before and after this proposal is the same. Could you explain the details of your concern?

> 2. Bypassing schema auto-update restrictions for replication is a workaround, not a proper solution.
>   If a remote cluster disables schema auto-updates, it reflects an explicit user decision to reject schema changes. Replicator should respect this setting. Injecting schemas automatically breaks the data contract and risks corrupting consumers.
>   A safer and more transparent approach would be to allow users to manually reset the cursor to a compatible schema point if they wish to resume replication.

It is an optional choice; the user can disable it.

> 3. Automatically modifying remote cluster policies is a dangerous design.
>   It violates cluster isolation and can introduce unintended side effects. In large-scale, multi-tenant deployments, such implicit adjustments are risky and potentially disruptive.
> 4. Instead of using automatic adjustments or bypass logic, replication should require explicit provisioning of topics and schemas on both clusters.
>   This keeps behavior predictable, under user control, and consistent with system boundaries.

It is an optional choice, and the default value is `false`. BTW, there is a command to preview the changes that will be applied.

pip/pip-433.md
nodece: > `createTopicToRemoteClusterForReplication`: whether to start replicator in the current cluster if enabled replication that replicates messages to the remote side

This description is a mismatch with the following:
```
# Whether the internal replication of the local cluster will trigger topic auto-creation on the remote cluster.
# 1. After enabling namespace-level Geo-Replication: whether the local broker will create topics on the remote
#   cluster automatically when calling `pulsar-admin topics create-partitioned-topic`.
# 2. When enabling topic-level Geo-Replication on a partitioned topic: whether the local broker will create topics on
#   the remote cluster.
# 3. Whether the internal Geo-Replicator in the local cluster will trigger non-persistent topic auto-creation for
#   remote clusters.
# It is not a dynamic config, the default value is "true" to preserve backward-compatible behavior.
createTopicToRemoteClusterForReplication=true
```

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+  - Without [PIP-414: Enforce topic consistency check](https://github.com/apache/pulsar/pull/24213), the geo replicator will trigger a non-partitioned topic creation, which is named `public/default/topic-partition-0`
+  - With [PIP-414: Enforce topic consistency check](https://github.com/apache/pulsar/pull/24213), the geo replicator get a denied error.
+- However, the user wants to allow the replicator to copy messages to the remote cluster.
+
+#### Issue 2: Replication is stuck because the remote side does not allow schema updates
+**Steps to reproduce the issue**
+- The topic`public/default/topic` has enabled a geo replication.
+- Users controls the topic schema manually, do not allow auto update schema by consumers/producers for both clusters.
+- The internal producer is stuck after the user sets a new schema on the source cluster.
+- However, the user wants to allowthe  replicator to copy the schema to the remote cluster.
```
Technoboy-: `allowthe` -> `allow the`

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+# Detailed Design
+
+### Public API
+
+#### Regarding Goal 1: "always allow the replicator to register schemas if compatible"
+
+**The original design of `pulsar-admin topics set-is-allow-auto-update-schema`**
+
+```shell
+pulsar-admin namespaces set-is-allow-auto-update-schema
```
Technoboy-: `set-is-allow-auto-update-schema` -> `auto-update-schema`

pip/pip-433.md
nodece: > Once the user changed the policy, the behavior before and after this proposal is the same. Could you explain the details of your concern?

This proposal still relies on implicit behavior: specifically, auto-creating topics during replication based on remote cluster policies. Even if the outcome appears the same after a policy change, the core issue remains: the replicator is depending on an external policy that may change without coordination. This is not a reliable or predictable design, especially in production environments.

> It is an optional choice; the user can disable it.

Making something optional does not make it safe. Features like `--adjust-namespace-level-policies-automatically` silently modify remote cluster settings, which is a serious violation of cluster boundaries and isolation. This kind of cross-cluster side effect is difficult to observe, audit, or control — and in multi-tenant or regulated environments, it may even be unacceptable.

nodece: This isn't just a convenience toggle, and you introduce implicit and potentially destructive behavior that's hard to debug and easy to misuse.

poorbarcode: @nodece 

> This proposal still relies on implicit behavior: specifically, auto-creating topics during replication based on remote cluster policies. Even if the outcome appears the same after a policy change, the core issue remains: the replicator is depending on an external policy that may change without coordination. This is not a reliable or predictable design, especially in production environments.

What I want to explain is that the proposal never changes the behavior, you should not argue with this point 😂 . 

BTW, the proposal is not in order to solve the issue that the replication does not work because users manually change the policies.

> Making something optional does not make it safe. Features like --adjust-namespace-level-policies-automatically silently modify remote cluster settings, which is a serious violation of cluster boundaries and isolation. This kind of cross-cluster side effect is difficult to observe, audit, or control — and in multi-tenant or regulated environments, it may even be unacceptable.

The user can preview what the command will change, and it can only be applied when the user manually calls the command with `--adjust-namespace-level-policies-automatically`. Why is it not safe?

> This isn't just a convenience toggle, and you introduce implicit and potentially destructive behavior that's hard to debug and easy to misuse.

Seems nothing is implicit.

pip/pip-433.md
poorbarcode: Renamed `sub topic` to `partition`

Haha, I learned this concept from other PIP 😂

pip/pip-433.md
poorbarcode: Fixed

pip/pip-433.md
poorbarcode: This command is correct, see also https://github.com/apache/pulsar/blob/master/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdNamespaces.java#L2822

**CmdNamespaces.java**
```java
addCommand("get-is-allow-auto-update-schema", new GetIsAllowAutoUpdateSchema());
addCommand("set-is-allow-auto-update-schema", new SetIsAllowAutoUpdateSchema());
```

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+#### Schemas replication if you have enabled Geo Replication.
+The internal producer of the geo replicator starts with an `auto-produce` schema, copies a new schema it reads from the source cluster to the remote cluster, and it will be stuck once a schema is incompatible to the remote cluster.
+
+# Motivation
+
+#### Issue 1: conflict topic creation if enabled Geo-Replication
+**Steps to reproduce the issue**
+- Configurations of both the source cluster and the remote cluster 
+  - `allowAutoTopicCreation`: `true`
+  - `allowAutoTopicCreationType`: `non-partitioned`
```
codelipenghui:
```suggestion
  - `allowAutoTopicCreationType`: `non-partitioned`
```

Do you mean partitioned?

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+
+**The original design of `pulsar-admin topics set-is-allow-auto-update-schema`**
+
+```shell
+pulsar-admin namespaces set-is-allow-auto-update-schema
+  -d, --disable         Disable schema validation enforced
+  -e, --enable          Enable schema validation enforced
+```
+
+To add a new param `--always-allow-for-replicator-if-compatible`, which means that always allow the replicator to register a new schema if compatible. The default value is `true`.
```
codelipenghui:
```suggestion
To add a new param `--enable-for-replicator`, which means that always allow the replicator to register a new schema if compatible. The default value is `true`.
```

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+  -n, --num-partitions=<defaultNumPartitions>
+                        The default number of partitions of a topic to be
+                          auto-created, applicable to partitioned topics only
+  -t, --type=<type>     Type of topic to be auto-created. Possible values:
+                          (partitioned, non-partitioned). Default value:
+                          non-partitioned
+                          Default: non-partitioned
+```
+
+To add a new param `--always-allow-for-replicator`, which means that always allow the replicator to trigger a topic auto-creation, including partitioned topics. The default value is `true`.
```
codelipenghui: Can we just let replicator to use admin api to create partitioned topic? It can check the topic first and decide if it's need to create the partitioned topic. We can only let the owner broker of partition-0 to do the check and create partitioned topic. Other partition owners can just keep retrying.

pip/pip-433.md
```diff
@@ -0,0 +1,172 @@
...
+--adjust-namespace-level-policies-automatically broker uses the policies that applied on the source cluster as a standard
+                                                value, to adjust remote cluster policies, including auto-creation
+                                                policies, topic partitions of existing topics, and schema update policies.
+```
+
+Add a new parameter for the exists pulsar-admin command `pulsar-admin namespaces set-clusters` to preview what will be modified for the new feature `--adjust-namespace-level-policies-automatically`
+
+```shell
+--preview preview the changes that will be applied when enabling "--adjust-namespace-level-policies-automatically"
+```
```
codelipenghui: If we use admin API, we don't need these options?

Technoboy-: does it good adding this `--preview` to a set command?

Technoboy-: +1

BewareMyPower:
Could you rebase to master so the checkstyle for tests will be applied?

pip/pip-433.md
nodece: > Can we just let replicator to use admin api to create partitioned topic? It can check the topic first and decide if it's need to create the partitioned topic. 

I agreed with you.

@codelipenghui Please see https://github.com/apache/pulsar/pull/24136.

The replicator requires improving observability (e.g., metrics, failure visibility) and helping users fix configuration issues explicitly and safely.

pip/pip-433.md
poorbarcode: Corrected it

pip/pip-433.md
poorbarcode: Changed the solution of the PIP, please take a look again

poorbarcode: Changed the solution of the PIP, please take a look again

poorbarcode:
@BewareMyPower 

> Could you rebase to master so the checkstyle for tests will be applied?

Seems this comment is in order to leave under other PRs?

pip/pip-433.md
poorbarcode: Changed the solution to use Admin Api Client to trigger auto creation

pip/pip-433.md
```diff
@@ -0,0 +1,151 @@
...
+#### Issue 2: Replication is stuck because the remote side does not allow schema updates
+**Steps to reproduce the issue**
+- The topic`public/default/topic` has enabled a geo replication.
+- Users controls the topic schema manually, do not allow auto update schema by consumers/producers for both clusters.
+- The internal producer is stuck after the user sets a new schema on the source cluster.
+- However, the user wants to allow the  replicator to copy the schema to the remote cluster.
+
+# Goals
+1. Add an optional choice: always allow the replicator to register schemas if compatible, even if users set `set-is-allow-auto-update-schema` to false.
+2. Regarding auto-creation that is triggered by replication, rather than the original solution that triggers auto-creation when the internal producer of the replicator, we add the admin api client into the replicator to call the admin api to create topics.
```
nodece, line 55: https://github.com/apache/pulsar/pull/24136 includes this feature. Is it correct?

pip/pip-433.md
```diff
@@ -0,0 +1,151 @@
...
+  - All topics’ partitions that have been created should be the same, including `__change_events`
+  - Auto-creation policies between both clusters should be the same, including broker-level and namespace-level.
+
+# Detailed Design
+
+### Implementation overview
+
+**Regarding goal 2: uses admin API client to trigger topic creation on the remote side**
+- Replicators will involve an admin api client, which previously only involved a Pulsar client.
+  - Do not trigger a topic auto-creation if the value of the configuration `createTopicToRemoteClusterForReplication` is `false`, which keeps the previous behavior. See more details  [PIP-370: configurable remote topic creation in geo-replication ](https://github.com/apache/pulsar/pull/23124).
```
nodece, line 66: https://github.com/apache/pulsar/pull/24485#discussion_r2196352691

This is inconsistent with what you said before. 

> `createTopicToRemoteClusterForReplication`: whether to start replicator in the current cluster if enabled replication that replicates messages to the remote side


I mean that this config should decide whether the request to create the topic is sent to the remote cluster.

pip/pip-433.md
```diff
@@ -0,0 +1,151 @@
...
+```
+pulsar-admin namespaces set-schema-autoupdate-strategy
+  -c, --compatibility=<strategyParam>
+                        Compatibility level required for new schemas created
+                          via a Producer. Possible values (Full, Backward,
+                          Forward).
+  -d, --disabled        Disable automatic schema updates
+```
+
+To add a new param `--enable-for-replicator`, which means that always allow the replicator to register a new schema if compatible. The default value is `true`.
```
nodece, line 91: -1, I don't suggest we ignore the replicator check. Once the schema is incompatible, we should disconnect the replicator.

pip/pip-433.md
```diff
@@ -0,0 +1,151 @@
...
+**`pulsar-admin topics set-replication-clusters -c <clusters> <topic name>`**
+- Previous behavior:
+  - It confirms that both partitioned topics between the two clusters have the same partitions.
+  - It copies the creation request to the remote cluster if the topic does not exist on the remote cluster
+  - It sets the policy.
+- The behaviors with the proposal
+  - It confirms that both partitioned topics between the two clusters have the same partitions.
+  - It sets the policy.
+
+#### Regarding the goal-3: "checks compatibility between two clusters when enabling namespace-level replication"
```
nodece, line 112: The replicator doesn't depend on the auto-creation feature after PIP-433, so we don't need to check this.

nodece, line 112: If you want to replicate `__change_events`, please make sure the topic type is the same between the source and remote clusters.

Technoboy- (approved):

pip/pip-433.md
poorbarcode: What I wrote in the proposal is "To add a new param `--enable-for-replicator`, which means that always allow the replicator to register a new schema if compatible."

Highlights "to register a new schema if compatible", could you review the proposal carefully?

pip/pip-433.md
poorbarcode: Have you found that the current comments argue the other comment you left [here](https://github.com/apache/pulsar/pull/24485#discussion_r2215392689)

pip/pip-433.md
poorbarcode: Regarding https://github.com/apache/pulsar/pull/24136, we had discussed on WeChat on `2025-04-09` to solve the conflict of topic creation when using Geo Relocation, which #24136 attempts to solve
After the discussion on WeChat, I submitted the current PIP and invited you to join the discussion on `2025-07-03`.

There are two solutions we had discussed:
1. Improve the previous topic auto-creation mechanisms if they are not suitable, which I suggested [here](https://github.com/apache/pulsar/pull/24136#discussion_r2018700515)
2. If we let the replicator involve Admin Api, we'd better remove the old mechanism, which makes the auto creation behavior clearer

The proposal has been changed twice:
- The first version: I consider the solution 1, but you and @codelipenghui suggested the solution 2 [here](https://github.com/apache/pulsar/pull/24485#discussion_r2199542223)
- The second version is the current one, which is the same as you and @codelipenghui suggested

The same things between #24136 and the current proposal:
- Let replicators involve the Admin API, to create remote side topics, instead of the previous implementation, which relied on topic auto-creation by the internal producer of the replicator

The difference between #24136 and the current proposal:
1. Add an optional choice: always allow the replicator to register schemas if compatible, even if users set `set-is-allow-auto-update-schema` to false.
2. Checks compatibility between two clusters when enabling namespace-level replication, which includes the following
  - All topics’ partitions that have been created should be the same, including `__change_events`
  - Auto-creation policies between both clusters should be the same, including broker-level and namespace-level.
3. To remove the previous mechanisms, which trigger remote-side topic auto-creation when creating a partitioned topic and enabling topic-level geo replication

pip/pip-433.md
poorbarcode: I have described [here](https://github.com/apache/pulsar/pull/24485#discussion_r2196349268)

> Seems you want to remove https://github.com/apache/pulsar/issues/13444, please start a new proposal for it, the current PIP does not includes this change

pip/pip-433.md
poorbarcode: Please do not mix the two versions of the solutions.

Two solutions: https://github.com/apache/pulsar/pull/24485#discussion_r2215467670

- https://github.com/apache/pulsar/pull/24485#discussion_r2196352691 argues the Goal 2 of the previous version of the pip

> 2. Add an optional choice: always allow the replicator to trigger a topic auto-creation, including partitioned topics, even if users have disabled topic auto-creation.

- The current implementation of PIP has changed, which you suggested [here](https://github.com/apache/pulsar/pull/24485#discussion_r2199542223)

----

BTW, what's the difference between the two expressions you mentioned?

>> createTopicToRemoteClusterForReplication: whether to start replicator in the current cluster if enabled replication that replicates messages to the remote side

> I mean that this config should decide whether the request to create the topic is sent to the remote cluster.

PR #24488, opened 2025-07-06
[improve][client] PIP-420: Supports users implement external schemas

gaoran10:
### Motivation

This is the implementation for [PIP-420](https://github.com/apache/pulsar/pull/24328).

### Modifications

Add a new SchemaType `EXTERNAL` and modify the corresponding protobuf messages.
Add a new method `Map<String, String> getConfigs()` for interface `SchemaInfoProvider`.
Add a new configuration `schemaProperties` for the Pulsar client builder.
Add `ExternalSchemaCompatibilityCheck` for the broker configuration files `broker.conf` and `standalone.conf`.
Close the schema while closing the Producer and Consumer.

### Verifying this change

Add a test to verify the Schema interface changes.

- [x] The public API
- [x] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheck.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void checkCompatible(Iterable<SchemaData> from, SchemaData to, SchemaCompatibilityStrategy strategy)
+            throws IncompatibleSchemaException {
+        if (strategy == SchemaCompatibilityStrategy.ALWAYS_COMPATIBLE) {
+            return;
+        }
+        while (from.iterator().hasNext()) {
+            SchemaData fromSchema = from.iterator().next();
+            checkCompatible(fromSchema, to, strategy);
+        }
```
Denovo1998: Sorry, it is here.
```java
for (SchemaData fromSchema : from) {
    checkCompatible(fromSchema, to, strategy);
}
```

gaoran10: fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/schema/validator/StructSchemaDataValidator.java
```diff
@@ -71,6 +71,7 @@ public void validate(SchemaData schemaData) throws InvalidSchemaDataException {
         } catch (InvalidSchemaDataException invalidSchemaDataException) {
             throw invalidSchemaDataException;
         } catch (Exception e) {
+            e.printStackTrace();
```
Demogorgon314:
```suggestion
```

gaoran10: fixed

BewareMyPower:
@Denovo1998 Any chance to take a look again?

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Message.java
```diff
@@ -208,6 +208,15 @@ public interface Message<T> {
      */
     byte[] getSchemaVersion();
 
+    /**
+     * Get schema ID of the message.
+     * PIP-420 provides a way to produce messages with external schema,
+     * and the schema ID will be set to the message metadata.
+     *
+     * @return schema ID of the message if the message is produced with external schema.
```
BewareMyPower: Please note `null` value can be returned if it does not have a schema id.

I also noticed an empty array also represents "no schema id". Please make the behavior clear in documents.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheck.java
```diff
@@ -0,0 +1,50 @@
...
+        return SchemaType.EXTERNAL;
+    }
+
+    @Override
+    public void checkCompatible(SchemaData from, SchemaData to, SchemaCompatibilityStrategy strategy)
+            throws IncompatibleSchemaException {
+        if ((SchemaType.EXTERNAL.equals(from.getType()) || SchemaType.EXTERNAL.equals(to.getType()))
+                && !from.getType().equals(to.getType())) {
+            throw new IncompatibleSchemaException("External schema is not compatible with the other schema types.");
+        }
```
Technoboy-, line 39: Does here forget to call `checkCompatible(fromSchema, to, strategy);`

BewareMyPower: The schema compatibility check between two `EXTERNAL` schemas should be done in the external schema registry, so it just do nothing if `from` and `to` are both `EXTERNAL`

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheckTest.java
```diff
@@ -0,0 +1,68 @@
...
+                        .type(SchemaType.PROTOBUF)
+                        .build(),
+                SchemaData.builder()
+                        .type(SchemaType.PROTOBUF_NATIVE)
+                        .build()
+        };
+    }
+
+    @Test(dataProvider = "otherSchemasProvider")
+    public void testExternalSchemaCompatibilityCheck(SchemaData schemaData) {
```
BewareMyPower, line 58: Could you test `checkCompatible(externalSchemaData, schemaData` as well?

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/EncodeData.java
```diff
@@ -0,0 +1,37 @@
...
+public class EncodeData {
+
+    private byte[] data;
+
+    private byte[] schemaId;
+
+    public EncodeData(byte[] data) {
+        this.data = data;
+        this.schemaId = new byte[0];
+    }
```
BewareMyPower: It's better to remove this constructor. I see `new EncodeData(data, null)` is called in the default `Schema#encode(String topic, T message)` implementation.

Mixing the use cases of `null` and `new byte[0]` makes code hard to read. Actually the immutable `record` should be best here.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Schema.java
```diff
@@ -184,6 +197,10 @@ default void configureSchemaInfo(String topic, String componentName,
      */
     Schema<T> clone();
 
+    default CompletableFuture<Void> closeAsync() {
+        return CompletableFuture.completedFuture(null);
+    }
```
Technoboy-, line 202: Seems this method only has an default implementation.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
```diff
@@ -501,16 +518,16 @@ private T decode(byte[] schemaVersion) {
         }
     }
 
-    private T decodeBySchema(byte[] schemaVersion) {
-        T value = poolMessage ? schema.decode(payload.nioBuffer(), schemaVersion) : null;
+    private T decodeBySchema(byte[] schemaFlag) {
```
BewareMyPower: Should it be renamed with `schemaId`? The `schemaFlag` concept in code is confusing. I see the result of `getSchemaFlag()` is still passed to `getSchemaInfo` method, which accepts a schema version

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
```diff
@@ -424,6 +424,23 @@ public byte[] getSchemaVersion() {
...
+            return (schemaId.length == 0) ? null : schemaId;
+        }
+        return null;
+    }
+
+    private byte[] getSchemaFlag() {
+        if (msgMetadata.hasSchemaId()) {
+            byte[] schemaId = msgMetadata.getSchemaId();
+            return (schemaId.length == 0) ? null : schemaId;
+        }
```
BewareMyPower: Now if the message has schema id, the schema id will be passed to all places that need a schema version. This should be a wrong behavior because these components only retrieve the schema from **built-in** schema registry. Passing the schema id from external schema registry makes no sense.

BewareMyPower (changes requested):

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Schema.java
BewareMyPower: Because built-in `Schema` implementations don't need to implement `closeAsync`. This method is added to support closing the connection to an external schema registry service for an external `Schema` implementation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheck.java
gaoran10: In any strategy mode, the external schema is not compatible with other Pulsar schemas; the external schema manager will manage the external schema compatibility check.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheckTest.java
gaoran10: added

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/EncodeData.java
gaoran10: changed to `record`, and removed all `new EncodeData(data, null)`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
gaoran10: Yes, the schema ID should only be used to decode the message payload.

- Changed get schemaInfo to use the schema version.
- Changed get schema reader to use the schema version.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Message.java
gaoran10: I changed to use the `Optional` object as the return type.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
gaoran10: Changed get schemaInfo to use the schema version, changed the parameter name to `schemaIdOrVersion`.

pulsar-common/src/main/proto/PulsarApi.proto
```diff
@@ -173,6 +174,7 @@ message MessageMetadata {
     // the `compacted_out` field in `SingleMessageMetadata` must be checked to identify and filter out compacted
     // messages (e.g., `k1 => v1` and `k1 => null` in the example above).
     repeated int32 compacted_batch_indexes = 31;
+    optional bytes schema_id = 32;
```
Denovo1998, line 177: It looks like it says `31` in the PIP documentation, should we update the PIP documentation in this PR?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/KeyValueSchemaImpl.java
```diff
@@ -187,8 +223,12 @@ public KeyValue<K, V> decode(byte[] keyBytes, byte[] valueBytes, byte[] schemaVe
...
             v = null;
         } else {
-            if (valueSchema.supportSchemaVersioning() && schemaVersion != null) {
-                v = valueSchema.decode(valueBytes, schemaVersion);
+            if (valueSchema.supportSchemaVersioning() && schemaIdOrVersion != null) {
+                byte[] valueSchemaIdOrVersion = schemaIdOrVersion;
+                if (SchemaType.EXTERNAL.equals(valueSchema.getSchemaInfo().getType())) {
+                    valueSchemaIdOrVersion = KeyValue.getSchemaId(schemaIdOrVersion, false);
+                }
+                v = valueSchema.decode(topic, valueBytes, valueSchemaIdOrVersion);
```
Denovo1998: Summarize
```java
V v = null;
if (valueBytes != null) {
    if (valueSchema.supportSchemaVersioning() && schemaIdOrVersion != null) {
        byte[] valueSchemaIdOrVersion = schemaIdOrVersion;
        if (SchemaType.EXTERNAL.equals(valueSchema.getSchemaInfo().getType())) {
            valueSchemaIdOrVersion = KeyValue.getSchemaId(schemaIdOrVersion, false);
        }
        v = valueSchema.decode(topic, valueBytes, valueSchemaIdOrVersion);
    } else {
        v = valueSchema.decode(valueBytes);
    }
}
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientBuilderImpl.java
```diff
@@ -505,4 +505,5 @@ public ClientBuilder lookupProperties(Map<String, String> properties) {
         conf.setLookupProperties(properties);
         return this;
     }
+
```
Denovo1998: Unnecessary modifications.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/KeyValueSchemaImpl.java
```diff
@@ -158,26 +180,40 @@ public KeyValue<K, V> decode(byte[] bytes, byte[] schemaVersion) {
...
             k = null;
         } else {
-            if (keySchema.supportSchemaVersioning() && schemaVersion != null) {
-                k = keySchema.decode(keyBytes, schemaVersion);
+            if (keySchema.supportSchemaVersioning() && schemaIdOrVersion != null) {
+                byte[] keySchemaIdOrVersion = schemaIdOrVersion;
+                if (SchemaType.EXTERNAL.equals(keySchema.getSchemaInfo().getType())) {
+                    keySchemaIdOrVersion = KeyValue.getSchemaId(schemaIdOrVersion, true);
+                }
+                k = keySchema.decode(topic, keyBytes, keySchemaIdOrVersion);
```
Denovo1998: Summarize
```java
K k = null;
if (keyBytes != null) {
    if (keySchema.supportSchemaVersioning() && schemaIdOrVersion != null) {
        byte[] keySchemaIdOrVersion = schemaIdOrVersion;
        if (SchemaType.EXTERNAL.equals(keySchema.getSchemaInfo().getType())) {
            keySchemaIdOrVersion = KeyValue.getSchemaId(schemaIdOrVersion, true);
        }
        k = keySchema.decode(topic, keyBytes, keySchemaIdOrVersion);
    } else {
        k = keySchema.decode(keyBytes);
    }
}
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/generic/MultiVersionSchemaInfoProvider.java
```diff
@@ -100,4 +100,5 @@ private CompletableFuture<SchemaInfo> loadSchema(byte[] schemaVersion) {
     public PulsarClientImpl getPulsarClient() {
         return pulsarClient;
     }
+
```
Denovo1998: Unnecessary modifications.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java
```diff
@@ -508,4 +508,5 @@ public void setLookupProperties(Map<String, String> lookupProperties) {
     public Map<String, String> getLookupProperties() {
         return (lookupProperties == null) ? Collections.emptyMap() : Collections.unmodifiableMap(lookupProperties);
     }
+
```
Denovo1998: Unnecessary modifications.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
```diff
@@ -717,4 +717,5 @@ ClientBuilder authentication(String authPluginClassName, Map<String, String> aut
      * - The `loadManagerClassName` config in broker is a class that implements the `ExtensibleLoadManager` interface
      */
     ClientBuilder lookupProperties(Map<String, String> properties);
+
```
Denovo1998: Unnecessary modifications.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/schema/KeyValueSchemaTest.java
```diff
@@ -375,7 +375,7 @@ public void testSeparatedKeyValueEncodingTypeSchemaEncodeAndDecode() {
         } catch (SchemaSerializationException e) {
             Assert.assertTrue(e.getMessage().contains("This method cannot be used under this SEPARATED encoding type"));
         }
-        KeyValue<Foo, Bar>  keyValue = ((KeyValueSchemaImpl) keyValueSchema).decode(fooSchema.encode(foo),
+        KeyValue<Foo, Bar>  keyValue = ((KeyValueSchemaImpl) keyValueSchema).decode(null, fooSchema.encode(foo),
```
Denovo1998: Here are two spaces
```java
KeyValue<Foo, Bar> keyValue
```

Denovo1998: It seems there are also two spaces in the `testInlineKeyValueEncodingTypeSchemaEncodeAndDecode` above.

Denovo1998 (changes requested):
LGTM!

pulsar-client-api/src/main/java/org/apache/pulsar/common/schema/KeyValue.java
```diff
@@ -100,28 +101,82 @@ public interface KeyValueDecoder<K, V> {
...
+                return keySchemaId;
+            }
+        }
+        int valueSchemaLength = buffer.getInt();
+        if (valueSchemaLength > 0) {
+            byte[] valueSchemaId = new byte[valueSchemaLength];
+            buffer.get(valueSchemaId);
+            return valueSchemaId;
+        }
+        throw new IllegalArgumentException("Invalid schemaId length: " + schemaId.length);
```
coderzc: Is valueSchema not allowed to be null?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheckTest.java
BewareMyPower: I mean this test only tests the case that `from` is a built-in schema and `to` is an external schema. It's better to test the case that `from` is an external schema and `to` is a built-in schema.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
BewareMyPower: +1

pulsar-client-api/src/main/java/org/apache/pulsar/common/schema/KeyValue.java
```diff
@@ -100,28 +101,82 @@ public interface KeyValueDecoder<K, V> {
...
+                4 + keySchemaId.length + 4 + valueSchemaId.length);
+        byteBuffer
+                .putInt(keySchemaId.length)
+                .put(keySchemaId)
+                .putInt(valueSchemaId.length)
+                .put(valueSchemaId);
         return byteBuffer.array();
     }
 
+    public static byte[] getSchemaId(byte[] schemaId, boolean isKey) {
```
BewareMyPower: Please add unit tests for this function. The current implementation is wrong.

```java
        final var schemaId = generateKVSchemaId(null, new byte[]{0x11, 0x22});
        final var keySchemaId = getSchemaId(schemaId, true);
```

In this case, `keySchemaId` should be an empty schema id array. However, it's the value schema.

BTW, I found allowing a null schema id could make key value schema very confusing. With the same code example, it's impossible to distinguish if the key schema is null or an empty array.

One solution is that never allow a null schema id. Then all APIs will become much simpler, e.g. `getSchemaId` won't need to return an `Optional`.

The other solution is to have special encoding for null schema id, e.g. use a negative length to represent null.

pulsar-common/src/main/proto/PulsarApi.proto
BewareMyPower: +1

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
BewareMyPower: I still think we should not mix schema id and version even in decoding. For example, you changed

```java
            return decode(schema.supportSchemaVersioning() ? getSchemaIVersion() : null);
```

to

```java
            return decode(schema.supportSchemaVersioning() ? getSchemaIdOrVersion() : null);
```

It changes the semantics of the `supportSchemaVersioning` method that if it's true, schema id could be used to decode the payload, which is unexpected.

Why not just keep the original logic not changed and add a separated method to decode value by schema id?

```java
    @Override
    public T getValue() {
        if (msgMetadata.hasSchemaId()) {
            return decodeBySchemaId(msgMetadata.getSchemaId());
        }
        // ...
```

pulsar-client-api/src/main/java/org/apache/pulsar/common/schema/KeyValue.java
gaoran10: Improved, and I added a test to cover different cases. https://github.com/apache/pulsar/pull/24488/files#diff-2c0d48104dc0a3b48606a351a0af057d337274afd61e96ebc8bd0aabee3c8a18R152

pulsar-client-api/src/main/java/org/apache/pulsar/common/schema/KeyValue.java
gaoran10: Improved, and I added a test to cover different cases. https://github.com/apache/pulsar/pull/24488/files#diff-2c0d48104dc0a3b48606a351a0af057d337274afd61e96ebc8bd0aabee3c8a18R152

pulsar-client/src/test/java/org/apache/pulsar/client/impl/schema/KeyValueSchemaTest.java
gaoran10: fixed

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
gaoran10: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java
gaoran10: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/generic/MultiVersionSchemaInfoProvider.java
gaoran10: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/KeyValueSchemaImpl.java
gaoran10: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientBuilderImpl.java
gaoran10: fixed

pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/KeyValueSchemaImpl.java
gaoran10: fixed

pulsar-common/src/main/proto/PulsarApi.proto
gaoran10: updated PIP-420

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java
gaoran10: Sure, I can add a new method for decoding with schemaId. We added two decode methods in the `Schema`, they should be only used for decoding with the schema ID

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/schema/ExternalSchemaCompatibilityCheckTest.java
gaoran10: added

pulsar-client-api/src/main/java/org/apache/pulsar/common/schema/KeyValue.java
```diff
@@ -100,26 +102,82 @@ public interface KeyValueDecoder<K, V> {
...
+
+    public static byte[] getSchemaId(byte[] schemaId, boolean isKey) {
+        ByteBuffer buffer = ByteBuffer.wrap(schemaId);
+        int keySchemaLength = buffer.getInt();
+        byte[] keySchemaId = new byte[0];
+        if (keySchemaLength > 0) {
+            keySchemaId = new byte[keySchemaLength];
+            buffer.get(keySchemaId);
+        }
+        if (isKey) {
```
BewareMyPower: If `isKey` is true, we can just skip the key bytes like:

```java
        if (isKey) {
            // ...
        } else {
            int keySchemaLength = buffer.getInt();
            buffer.position(buffer.position() + keySchemaLength);
            // TODO:
        }
```

Denovo1998 (approved):

PR #24498, opened 2025-07-10
[fix][ml] Fix the CAS operation in initiate() failure

dao-jun:
### Motivation

In our pulsar cluster, we have encountered an extremely strange problem.
A topic has been running for a long time, but at a time, it starts `fenced` for a long time(more than 3 hours, until we unload it manually).

After we analysis the heap dump and broker log, we find the root cause:

For some reason the CAS operation in `OpAddEntry.initiate()` failed, but it logs `initiate with unexpected state OPEN, expect OPEN state.`
image
Which means the `state` is `OPEN`, but the CAS operation still fails.
Because of the CAS operation failed, the request did not send to bookies.

The next request passed the CAS operation and sent to bookies and received the response, when doing callback, it will check the `ml.pendingAddEntries`.
image
It polls the 1st entry, but the 1st entry did not send to bookies, so it will `firstInQueue.failed(...)` the 1st entry, and the topic will be fenced.
image

But the 2nd entry will always hang there, **because it only fails the 1st entry but do nothing to itself**. So the topic keeps fencing and never be unfenced.

The PR change maybe could causes message out of ordering, so do not merge.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


dao-jun:
More context:
1. No new ledger created at the time for the ML, so `OpAddEntry.initiate()` only called in `ManagedLedger.internalAsyncAddEntry`
2. The topic is fenced and `PersistentTopic.pendingWriteOps` is 1
image
3. The `OpAddEntry`'s `state = COMPLETED` and `entryId` was set, it means it was sent to bookies and received response, but it still in the queue
image

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -172,11 +172,19 @@ public void initiate() {
             ledger.asyncAddEntry(duplicateBuffer, this, addOpCount);
         } else {
             State state = STATE_UPDATER.get(OpAddEntry.this);
+            if (state == State.INITIATED || state == State.COMPLETED || state == State.CLOSED) {
+                log.warn("[{}] OpAddEntry is already initiated or completed or closed, state={}", ml.getName(), state);
+                return;
+            }
+
```
nodece, line 179: It looks unnecessary.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -172,11 +172,19 @@ public void initiate() {
...
             log.warn("[{}] initiate with unexpected state {}, expect OPEN state.", ml.getName(), state);
-            if (state != State.INITIATED && state != State.COMPLETED && state != State.CLOSED) {
-                ml.pendingAddEntries.remove(OpAddEntry.this);
-                this.failed(new ManagedLedgerException("Unexpected state " + state));
+            ManagedLedgerException ex = ml.interceptorException;
+            if (ex == null) {
```
nodece, line 182: You can directly set the ManagedLedgerException.

zhaizhibo:
We encountered a similar issue, where all bundles on this broker were unloaded following Zookeeper session expiration, after that, some futures leaks.
pendingWriteOps

PR #23184, opened 2024-08-16
[fix] [broker] Topic can never be loaded up due to broker maintains a failed topic creation future

poorbarcode:
### Motivation

There are two scenarios that may leave a failed topic creation future in `brokerService.topics` because forgot to remove the failed future from the map, which leads to the topic never being loaded up successfully.
- Topic load and delete execute at the same time: see the test `testTopicLoadAndDeleteAtTheSameTime`
- Check ownership fails: see the test `testCheckOwnerShipFails `

### Modifications

Remove the failed future if load topic fails

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #23004, opened 2024-07-04
[fix][broker] Fix MessageDeduplication replay timeout cause topic loading stuck

TakaHiR07:
Fixes https://github.com/apache/pulsar/issues/23003

### Motivation

As shown in the issue. fix the risk of topic loading stuck. 

### Modifications

1. make takeSnapshot() async , and if topic loading timeout, topic close should wait until takeSnapshot finish. This would not influence the other area where using takeSnapshot()
3. add test for this issue scene

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24500, opened 2025-07-10
[fix][broker] Fix failed topic future not removed before the deduplication recovery is done

### LINKED ISSUE #24497: [Bug] Topic cannot be accessed for "Failed to load topic within timeout" unless unloading the namespace
### Search before reporting

### Read release policy

### User environment

StreamNative Pulsar 4.0.5.2, which is based on https://github.com/apache/pulsar/commit/6b5fdbf2282efcfed30badab2c79bab3a6f92393

### Issue Description

This issue happens when we restarted brokers for an upgrade. There were two partitions (10 and 23 of a partitioned topic) owned by a broker stuck at initializing the `PersistentTopic`. Here are the info from the heap dump:

```yaml
topic:
  - name: "<topic>-partition-23"
    messageDeduplication:
      snapshotCounter: 21
      managedCursor:
        entriesReadCount: 42
        lastMarkDeleteEntry:
          newPosition: "860608:298"
        pendingReadOps: 0
        readPosition: "861217:3"
      managedLedger:
        config:
          readEntryTimeoutSeconds: 0
        executor:
          runner: "BookKeeperClientWorker-OrderedExecutor-0-0"
          tasksFailed: 1
        lastConfirmedEntry: "861217:2"
        ledgers:
          - key: 860608
            value: {"entries": 338}
          - key: 861217
            value: {"entries": 3}
          - key: 861315
            value: {"entries": 0}
  - name: "<topic>-partition-10"
    messageDeduplication:
      snapshotCounter: 0
      managedCursor:
        entriesReadCount: 84
        lastMarkDeleteEntry:
          newPosition: "860650:231"
        pendingReadOps: 1
        readPosition: "861214:0"
      managedLedger:
        currentLedger:
          ledgerId: 861295
        lastConfirmedEntry: "861214:1"
        ledgers:
          - key: 860650
            value: {"entries": 314}
          - key: 861214
            value: {"entries": 2}
          - key: 861295
            value: {"entries": 0}
        state: LedgerOpened
```

The key is the `MessageDeduplication` field:
- `status` is `Initialized`
- `managedCursor` is not null
- `snapshotCounter` is non-zero

From https://github.com/apache/pulsar/blob/ce102dac32978740b8dd7ddb1598ce3263ef6202/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java#L292-L298

we can get a conclusion that the future of `recoverSequenceIdsMap` never completed.

The difference between partition 10 and 23 is:
- Partition 10's cursor read position didn't reach the last confirmed entry, and there was a pending read operation. It seems that a read operation was somehow stuck forever.
- Partition 23's cursor read position exceeds the last confirmed entry, the future should have been completed. There is another suspicious point that `snapshotCounter` is 21 while `entriesReadCount` is 42.

The most possible reason is that exceptions were thrown in a callback like https://github.com/apache/pulsar/blob/ce102dac32978740b8dd7ddb1598ce3263ef6202/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java#L178

There was something wrong with our log collection, the key logs of `SingleThreadExecutor#run` have been lost. But we can check the `taskFailed` field is 1 from the heap dump.

### Error messages

```text

```

### Reproducing the issue

Currently I cannot reproduce it.

### Additional information

Currently Pulsar's topic caching mechanism is bad. There is a timeout configured by `topicLoadTimeoutSeconds` (default: 60). 1 minute is usually long enough in most cases. However, when the timeout happens, the topic cache won't be removed from `BrokerService#topics` until the pending `PersistentTopic` initialization completes. See https://github.com/apache/pulsar/blob/ce102dac32978740b8dd7ddb1598ce3263ef6202/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1849-L1855

`initialize()`:
1. Create a compaction service and set `topicCompactionService` with it.
2. Create subscriptions for all cursors in the managed ledger.
3. Get namespace policies for this topic from the metadata store.
4. Initialize some fields according to the namespace policies.
5. Register itself to the topic policies service.
6. Get topic policies and update some fields.
7. Remove orphan replication cursors.

`preCreateSubscriptionForCompactionIfNeeded()`: Create `__compaction` subscription, which might open the durable cursor.

`checkReplication()`:
1. Get allowed clusters from the metadata store.
2. Start replicators if necessary.

`checkDeduplicationStatus()`: Perform the topic replay.

**If there is a bug that makes `PersistentTopic` stuck at initialization, e.g. `checkDeduplicationStatus()`, the topic will be unavailable until manual intervention.**

BewareMyPower:
Fixes #24497

### Background about the initialization of `PersistentTopic`

The initialization includes the chained asynchronous calls of `initialize`, `preCreateSubscriptionForCompactionIfNeeded`, `checkReplication` and `checkDeduplicationStatus` methods. Calling all of them in `BrokerService` is unnecessarily complicated. Ideally, they should be wrapped in a single `initialize()` method for `BrokerService` to call.

The recovery of transaction buffer (`TopicTransactionBuffer`) should belong to the part of initialization as well. But it adopts an asynchronous way that if it failed to recover, it would close the `PersistentTopic`. This makes code complicated, which requires a more complicated fix (https://github.com/apache/pulsar/pull/23884) for the bug that could never have a chance to happen if the `PersistentTopic` cannot be accessed before the transaction buffer recovery is done.

### Motivation

If a `PersistentTopic` is stuck at initialization due to some bugs, the topic cache (`BrokerService#topics`) will never remove it. There is a topic initialization timeout configured by `topicLoadTimeoutSeconds` (60 by default). However, if the topic loading does not finish within the timeout, **it won't be removed from the cache**.

The dirty topic cache issue has been standing for long time, fixes like https://github.com/apache/pulsar/pull/22860 and https://github.com/apache/pulsar/pull/23184 only aim at calling `topics.remove(topic, topicFuture)` in many code paths. But there are still some cases not considered, for example, in https://github.com/apache/pulsar/issues/24497, we found two topics were not available until unloading them.

### Modifications

This PR adopts a simple and effective solution by removing the topic from the cache when the topic loading times out. The problem is that there could be two pending topic initializations. However, the old `PersistentTopic` never has a chance to be accessed because it does not exist in the topic cache.

If the topic loading already times out when a managed ledger is opened, just skip creating the `PersistentTopic`. Then next time the same topic is loaded, the managed ledger will already exist in the cache.

In addition, to address the possible stuck issue in `MessageDeduplication`, this PR:
- Adopts a future-based managed ledger util methods instead of callback-based managed ledger APIs.
- When the topic times out, failing the future returned by `MessageDeduplication#checkStatus`
- To keep the behavior of https://github.com/apache/pulsar/pull/23004 not changed, when the case above happens, take snapshot for the latest replayed snapshot
- To avoid being affected by the closing of outdated topic, skip closing the managed ledger in `PersistentTopic#close`

The following tests are added to `OrphanPersistentTopicTest`
- `testDeduplicationReplayStuck`: inject high latency for `asyncReadEntries`, after the topic loading timeout, a new managed cursor will be created for replay, which does not have high read latency.
- `testOrphanManagedLedgerRemovedAfterUnload`: since an orphan managed ledger will be created for loading timeout, verify it will be removed if the topic is unloaded.
- `testDuplicatedTopics`: verify closing the old `PersistentTopic` won't close the managed ledger so that the current `PersistentTopic` can still work

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/48


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -1657,16 +1658,20 @@ public PulsarAdmin getClusterPulsarAdmin(String cluster, Optional<ClusterData> c
...
+    @VisibleForTesting
+    public CompletableFuture<Optional<Topic>> loadOrCreatePersistentTopic(final String topic,
             boolean createIfMissing, Map<String, String> properties) {
         final CompletableFuture<Optional<Topic>> topicFuture = FutureUtil.createFutureWithTimeout(
                 Duration.ofSeconds(pulsar.getConfiguration().getTopicLoadTimeoutSeconds()), executor(),
                 () -> FAILED_TO_LOAD_TOPIC_TIMEOUT_EXCEPTION);
 
-        topicFuture.exceptionally(t -> {
+        topicFuture.exceptionallyAsync(e -> {
             pulsarStats.recordTopicLoadFailed();
```
poorbarcode, line 1669: The primary change of https://github.com/apache/pulsar/pull/22860 is to remove the failed Future case by case instead of removing a failed topic creation future here.

@shibd @codelipenghui  Could you also review the current PR?

poorbarcode (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
BewareMyPower: Yes, I explained all details in the PR description.

BewareMyPower: Key point:

> But there are still some cases not considered

https://github.com/apache/pulsar/issues/24497 is the best example. https://github.com/apache/pulsar/pull/23184 is another example before.

The main concern is that if an orphan pending `PersistentTopic` could affect the new pending `PersistentTopic`, mainly due to:
- Managed ledger, which was handled in this PR
- Replicators, especially the Pulsar producer

Even if it could make the new `PersistentTopic` fail to load, it will be the same with the existing case, with better exception message (e.g. `replicator create failure`) rather than the misleading timeout exception (`Failed to load topic within timeout`).

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -1842,6 +1846,12 @@ public void createPersistentTopic0(final String topic, boolean createIfMissing,
...
                             try {
+                                if (topicFuture.isCompletedExceptionally()) {
+                                    // Don't close the managed ledger because next time the topic is accessed, the
+                                    // managed ledger will be created again. The managed ledger will be removed if the
+                                    // ownership has been changed.
+                                    return;
+                                }
                                 PersistentTopic persistentTopic = isSystemTopic(topic)
                                         ? new SystemTopic(topic, ledger, BrokerService.this)
                                         : newTopic(topic, ledger, BrokerService.this, PersistentTopic.class);
```
Demogorgon314, line 1857: Maybe we can `checkDeduplicationStatus` first before `checkReplication` to avoid creating replicator.

PR #21143, opened 2023-09-07
[improve][io] PIP-297: Support terminating Function & Connector with the fatal exception

RobertIndie:
PIP: #21079 

### Motivation

Please see PIP-297:
https://github.com/apache/pulsar/blob/3cb79264c4476a2cd34b0807133afc29750c047d/pip/pip-297.md?plain=1#L71-L91

### Modifications

https://github.com/apache/pulsar/blob/3cb79264c4476a2cd34b0807133afc29750c047d/pip/pip-297.md?plain=1#L108-L112

### Verifying this change

This change added tests.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24501, opened 2025-07-10
[fix][io] Fix data loss issue in Kinesis source connector

shibd:
### Motivation
The current Kinesis source connector violates the `at-least-once` delivery guarantee, which can lead to data loss during failures or restarts. The root cause is that the Kinesis checkpointing is based on a time interval, completely decoupled from Pulsar's acknowledgment mechanism.

This creates a race condition that leads to data loss, as illustrated below:

- Time 1: KinesisRecordProcessor fetches record 0 from a Kinesis shard and puts it into the internal queue.
- Time 2: The KinesisSource thread reads record 0 from the queue and attempts to send it to a Pulsar topic, but the send operation is delayed or failing.
- Time 3: The KinesisRecordProcessor's timer triggers. It checkpoints its progress, marking that it has successfully processed up to record 0, even though this record has not yet been acknowledged by Pulsar.
- Time 4: The connector crashes before record 0 can be successfully sent and acked. Upon restart, the KCL worker reads the last checkpoint and starts consuming from the position after record 0, causing record 0 to be permanently lost.

This PR fixes this critical issue by ensuring that checkpoints are only committed after a record has been successfully acknowledged by the Pulsar IO runtime.

### Modifications

1. **Refactored Checkpointing Logic**: The checkpointing mechanism in KinesisRecordProcessor was changed from being time-based to being driven by Pulsar acks. It now checkpoints a specific sequenceNumber provided by an acknowledged record.

2. **Implemented ack() and fail()**: In KinesisRecord.java, the ack() method was implemented to call back to the KinesisRecordProcessor and update the latest sequence number that is safe to checkpoint. The fail() method now calls sourceContext.fatal() to ensure the connector restarts upon failure.

3. **Passed SourceContext**: The SourceContext is now passed down to the KinesisRecordProcessor to allow it to trigger a fatal connector restart upon unrecoverable errors (e.g., failed checkpoint attempts).

### Verifying this change
-  Added Unit Tests: Created KinesisRecordProcessorTest.java to verify the new ack-driven checkpointing logic.

### Integration test
1. Create a pulsar topic and a subscription.
2. Create a Kinesis source connector.
3. Send 100 messages to Kinesis with 1s internal.
```java
    public static void main(String[] args) throws Exception {

        AWSCredentialsProvider credentialsProvider =
                new AWSStaticCredentialsProvider(new BasicAWSCredentials("", ""));

        KinesisProducerConfiguration kinesisConfig = new KinesisProducerConfiguration();
        kinesisConfig.setRegion("ap-northeast-1");
        kinesisConfig.setCredentialsProvider(credentialsProvider);
        KinesisProducer kinesis = new KinesisProducer(kinesisConfig);
        // Put some records 
        for (int i = 0; i < 100; ++i) {
            ByteBuffer data = ByteBuffer.wrap("test-kinesis-data".getBytes("UTF-8"));
            // doesn't block       
            kinesis.addUserRecord("connector-source", "myPartitionKey", data);
            kinesis.flush();
            Thread.sleep(1000);
        }
        kinesis.flush();
        Thread.sleep(10000);
        kinesis.destroy();
    }
```
4. Topic unloads and connector restarts will occur at random intervals during this time.

Result:

Before this PR:  **will receive fewer than 100 messages**
```
Receive message test-kinesis-data with count 96
Receive message test-kinesis-data with count 97
Received 97 messages
```

After this PR:  **Verify that the number of received messages is greater than or equal to 100, which fulfills the at-least-once delivery guarantee.*
```
Receive message test-kinesis-data with count 101
Receive message test-kinesis-data with count 102
Received 102 messages
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
+++ b/pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
@@ -21,6 +21,8 @@ package org.apache.pulsar.io.kinesis;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.client.api.PulsarClientException;
+import org.apache.pulsar.io.core.SourceContext;
 import software.amazon.kinesis.exceptions.InvalidStateException;
 import software.amazon.kinesis.exceptions.KinesisClientLibDependencyException;
 import software.amazon.kinesis.exceptions.ShutdownException;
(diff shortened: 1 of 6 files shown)


pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
```diff
@@ -82,49 +93,66 @@ private void checkpoint(RecordProcessorCheckpointer checkpointer) {
...
             }
         }
 
         // Checkpoint once every checkpoint interval.
         if (System.nanoTime() > nextCheckpointTimeInNanos) {
-            checkpoint(processRecordsInput.checkpointer());
+            if (sequenceNumberToCheckpoint != null
+                    && !sequenceNumberToCheckpoint.equals(lastCheckpointedSequenceNumber)) {
+                checkpoint(processRecordsInput.checkpointer(), sequenceNumberToCheckpoint);
+            }
```
RobertIndie, line 132: As we discussed privately, it's better to add a timer to invoke the checkpoint peridocally.

RobertIndie: It's a little hard to do this since we need to get the checkpointer from each processRecordsInput. I'm +1 for the current implementation.

BewareMyPower:
Could you rebase to master so the checkstyle for tests will be applied?

lhotari:
This change depends on #21143, @shibd could you please handle backporting to branch-3.0 or remove the release label for 3.0.x ?

shibd:
Thanks for the reminder, I think we can remove the label for 3.0 and 3.3

PR #24510, opened 2025-07-15
[improve][pip] PIP-434: Expose Netty channel configuration WRITE_BUFFER_WATER_MARK to pulsar conf and pause receive requests when channel is unwritable

poorbarcode:

- [Discussion](https://lists.apache.org/thread/hnbm9q3yvyf2wcbdggxmjzhr9boorqkn)
- [Vote](https://lists.apache.org/thread/vpvtf4jnbbrhsy9y5fg00mpz9qhb0cp5)

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


pip/pip-434.md
```diff
@@ -0,0 +1,76 @@
...
+
+> Solution we got:
+> - Expose Netty channel configuration WRITE_BUFFER_WATER_MARK to pulsar conf
+> - Stops receive requests continuously once the Netty channel is unwritable, users can use the new config to control the threshold that limits the max bytes that are pending write.
+
+# Goals
+
+## In Scope
+- Expose Netty channel configuration WRITE_BUFFER_WATER_MARK to pulsar conf
+- Stops receive requests continuously once the Netty channel is unwritable, users can use the new config to control the threshold that limits the max bytes that are pending write.
```
lhotari, line 23: besides configuring the high water mark and the low water mark, there should be a way to disable the action (`ServerCnxThrottleTracker`'s
`incrementThrottleCount`/`decrementThrottleCount`) that is executed in the `channelWritabilityChanged` callback method.

pip/pip-434.md
```diff
@@ -0,0 +1,76 @@
...
+
+As we discussed along the discussion: https://lists.apache.org/thread/6jfs02ovt13mnhn441txqy5m6knw6rr8
+
+> Problem Statement:
+> We've encountered a critical issue in our Apache Pulsar clusters where brokers experience Out-Of-Memory (OOM) errors and continuous restarts under specific load patterns. This occurs when Netty channel write buffers become full, leading to a buildup of unacknowledged responses in the broker's memory.
+
+> Background:
+> Our clusters are configured with numerous namespaces, each containing approximately 8,000 to 10,000 topics. Our consumer applications are quite large, with each consumer using a regular expression (regex) pattern to subscribe to all topics within a namespace.
+
+> The problem manifests particularly during consumer application restarts. When a consumer restarts, it issues a getTopicsOfNamespace request. Due to the sheer number of topics, the response size is extremely large. This massive response overwhelms the socket output buffer, causing it to fill up rapidly. Consequently, the broker's responses get backlogged in memory, eventually leading to the broker's OOM and subsequent restart loop.
```
lhotari, line 13: Throttling incoming requests for individual channels (connections) isn't a very effective way to solve this problem. However, I do support adding a way to configure WRITE_BUFFER_WATER_MARK parameters for the Netty channel.

There will be a negative impact to performance in cases where the write buffer size exceeds due to other reasons. This is due to the fact that there will be less pipelining (parallel processing) of incoming requests when incoming requests are throttled whenever the channel isn't writable. For example, this will negatively impact the latency of publish (send) requests while consumers are dispatching messages (and the messages in the write buffer cause the size of the write buffer to exceed the high water mark).

I have explained this [in the mailing list thread](https://lists.apache.org/thread/6jfs02ovt13mnhn441txqy5m6knw6rr8), but the arguments seem to have been ignored there, so let's handle the argument here.

pip/pip-434.md
poorbarcode: Added a new param `pulsarChannelPauseReceivingRequestsIfUnwritable=false`

pip/pip-434.md
poorbarcode: > I have explained this [in the mailing list thread](https://lists.apache.org/thread/6jfs02ovt13mnhn441txqy5m6knw6rr8), but the arguments seem to have been ignored there, so let's handle the argument here.

It was not ignored. I have answered it [in the mailing list thread](https://lists.apache.org/thread/6jfs02ovt13mnhn441txqy5m6knw6rr8)
> Once the channel is not writable, all requests that are sent to this
channel can not receive a reply anymore because the response can not be
written out. The results are the same; the clients receive replies delayed.
To improve performance, users should consider using more channels; in other
words, they can set a bigger `connectionsPerBroker` or separate clients.

---

> For example, this will negatively impact the latency of publish (send) requests while consumers are dispatching messages (and the messages in the write buffer cause the size of the write buffer to exceed the high water mark).

Yes, it will. However, once the channel is not writable, the producer cannot receive a send response, which also delays message publishing. And we have exposed the 3 configurations to users, they can adjust them if needed

lhotari: > Yes, it will. However, once the channel is not writable, the producer cannot receive a send response, which also delays message publishing. 

Yes, I agree that from the client's perspective, it would be delayed since it wouldn't receive a response immediately. However, there's an important difference here. When incoming send requests are continued to be processed, but the messages will get persisted and processed by the broker and therefore pipelining is happening. Allowing pipelining is important for Pulsar's performance, where it is common that the same connection is shared for multiple purposes (consumers, producers, etc.).

> And we have exposed the 3 configurations to users, they can adjust them if needed

Sure, users can tune the parameters and that makes sense. However, the default values and configuration examples should contain useful examples. The default high watermark value should be relatively high so that it wouldn't impact use cases where there's consumers and producers sharing the same connection.

pip/pip-434.md
lhotari: The new parameter resolves this. It would be also be useful to describe the implementation detail of handling the action in  `channelWritabilityChanged` and the usage of `ServerCnxThrottleTracker` in this action (somewhere in the PIP document).

pip/pip-434.md
poorbarcode: > The default high watermark value should be relatively high so that it wouldn't impact use cases where there's consumers and producers sharing the same connection.

Since the default value of `pulsarChannelPauseReceivingRequestsIfUnwritable` is `false`, we'd better set the watermark threshold to the default value of netty.

Another case, if the default value of `pulsarChannelPauseReceivingRequestsIfUnwritable` is `true`, the default value of the watermark threshold should be a higher value.

lhotari: > Since the default value of `pulsarChannelPauseReceivingRequestsIfUnwritable` is `false`, we'd better set the watermark threshold to the default value of netty.

In the case of Pulsar, enabling throttling based on writability would result in bad performance in most cases since it would reduce the amount of pipelining.  That's why the default Netty values aren't a sensible example for default values. The channel writability is completely ignored if the writability status of the channel and the event `channelWritabilityChanged` is ignored. That's currently the case in Pulsar broker code (excluding the Pulsar Proxy, where writability change is used for backpressure in proxied connections).

I think this could be resolved by providing sensible defaults (finding out the values would require benchmarking) and providing an explanation.

pip/pip-434.md
```diff
@@ -0,0 +1,78 @@
...
+# Once the writer buffer is full, the channel stops dealing with new requests until it changes to writable
+pulsarChannelPauseReceivingRequestsIfUnwritable=false
+```
+
+### CLI
+
+### Metrics
+| Name                                                 | Description                                                                                 | Attributes   | Units|
+|------------------------------------------------------|---------------------------------------------------------------------------------------------|--------------| --- |
+| `pulsar_server_channel_write_buf_memory_used_bytes` | Counter. The number of replicators.                      | cluster | - |
```
codelipenghui: It seems the description is wrong.

And it's for single channel, or the total memory usage?

pip/pip-434.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Configuration
+
+```shell
+# It relates to configuration "WriteBufferHighWaterMark" of Netty Channel Config. If the number of bytes queued in the write buffer exceeds this value, channel writable state will start to return "false".
+pulsarChannelWriteBufferHighWaterMark=64k
+# It relates to configuration "WriteBufferLowWaterMark" of Netty Channel Config. If the number of bytes queued in the write buffer is smaller than this value, channel writable state will start to return "true".
+pulsarChannelWriteBufferLowWaterMark=32k
+# Once the writer buffer is full, the channel stops dealing with new requests until it changes to writable
+pulsarChannelPauseReceivingRequestsIfUnwritable=false
```
codelipenghui, line 39: Is that mean the broker will also stop receiving messages from the producer once it reaches the high water mark?

codelipenghui (dismissed):

pip/pip-434.md
poorbarcode: > Is that mean the broker will also stop receiving messages from the producer once it reaches the high water mark?

Yes, it will. But it affects little because the command `CommandSendReceipt` can not be sent out once the socket is not writable, which also leads to the publishing being delayed

pip/pip-434.md
poorbarcode: Thanks for mentioning this mistake, I have corrected it

pip/pip-434.md
lhotari: > Yes, it will. However, once the channel is not writable, the producer cannot receive a send response, which also delays message publishing. And we have exposed the 3 configurations to users, they can adjust them if needed

Accurately speaking, Netty channel's writability itself doesn't impact whether something can be sent out.

The concept of "writability" means that when the number of buffered bytes exceeds the high water mark setting, `Channel.isWritable()` returns `false`. Once the buffered bytes drop below the low water mark,
the writability state changes back to `true`. This is explained in  https://netty.io/4.1/api/io/netty/channel/WriteBufferWaterMark.html

It's true that responses would be delayed when the writability status is `false` and remains in that status for a longer period of time. However, it's a normal situation that writability status is changing constantly when bursts of data goes out and until it gets transferred to OS network buffers.

Pulsar's current approach to backpressure is currently based on different settings. For publishing there's `maxMessagePublishBufferSizeInMB` and for dispatching there's `managedLedgerMaxReadsInFlightSizeInMB`.
The topic listing limit should be global for the broker in a similar way to really be helpful.

The channel writability level settings could be useful as the last resort, but it will require setting the values a lot higher than the Netty defaults since Pulsar's performance requires pipelining. That's why it won't be possible to use this solution to solve the problems with topic listings. There will need to be another setting that is suitable as a broker level limit so that broker OOM can be avoided.

pip/pip-434.md
lhotari: > The new parameter resolves this. It would be also be useful to describe the implementation detail of handling the action in `channelWritabilityChanged` and the usage of `ServerCnxThrottleTracker` in this action (somewhere in the PIP document).

@poorbarcode This hasn't been described in the detailed design. Please do so.

pip/pip-434.md
poorbarcode: @lhotari 

> I think this could be resolved by providing sensible defaults (finding out the values would require benchmarking) and providing an explanation.
> The channel writability level settings could be useful as the last resort, but it will require setting the values a lot higher than the Netty defaults since Pulsar's performance requires pipelining. That's why it won't be possible to use this solution to solve the problems with topic listings. There will need to be another setting that is suitable as a broker level limit so that broker OOM can be avoided.

Since we added a metric named `pulsar_server_channel_write_buf_memory_used_bytes` by this proposal, we can get a suitable value after real practice. Then we can start a new proposal to make the mechanisms better

pip/pip-434.md
poorbarcode: > @poorbarcode This hasn't been described in the detailed design. Please do so.

Added [here](https://github.com/apache/pulsar/pull/24510/files#diff-b2fc1ccd29f3f3e1429b9b39c13808f7a85e67d27f2c332189ecbc4c3356e49aR46-R56)

lhotari: @poorbarcode you seemed to ignore the advice about using `ServerCnxThrottleTracker` for throttling. Just clarifying that channel's auto read property shouldn't be directly changed. It should be left to `ServerCnxThrottleTracker` to handle that. You should be using `incrementThrottleCount` when writability changes from `true` to `false` and `decrementThrottleCount` when writability changes from `false` to `true`. Please add this to the design document.

poorbarcode: > It should be left to ServerCnxThrottleTracker to handle that. You should be using incrementThrottleCount when writability changes from true to false and decrementThrottleCount when writability changes from false to true. Please add this to the design document.

It is not a big change, I think it is not needed to write into the proposal

lhotari: It is essential to mention it in the PIP. Pulsar rate limiting and backpressure based on `maxMessagePublishBufferSizeInMB` breaks with this solution unless `ServerCnxThrottleTracker` is used.

Technoboy- (approved):

coderzc (approved):

pip/pip-434.md
```diff
@@ -0,0 +1,94 @@
...
+
+### How it works
+With the settings `pulsarChannelPauseReceivingRequestsIfUnwritable=false`, the behaviour is exactly the same as the previous.
+
+After setting `pulsarChannelPauseReceivingRequestsIfUnwritable` to `true`, the channel state will be changed as follows.
+- Netty sets `channel.writable` to `false` when there is too much data that is waiting to be sent out(the size of the data cached in `ChannelOutboundBuffer` is larger than `{pulsarChannelWriteBufferHighWaterMark}`)
+  - Netty will trigger an event `channelWritabilityChanged`
+- Stops receiving requests that come into the channel, which relies on the attribute `autoRead` of the `Netty channel`.
+- Netty sets `channel.writable` to `true` once the size of the data that is waiting to be sent out is less than `{pulsarChannelWriteBufferLowWaterMark}`
+- Starts to receive requests(sets `channel.autoRead` to `true`).
```
lhotari, line 54: As mentioned in comment https://github.com/apache/pulsar/pull/24510/files#r2262580121, this definitely has to be modified before this PIP can be accepted.

pip/pip-434.md
```diff
@@ -0,0 +1,94 @@
...
+  - To avoid avalanches, Pulsar will start a timed rate-limiter, which limits the rate of handling the request backlog("pulsarChannelRateLimitingRateAfterResumeFromUnreadable" requests per second).
+  - After "{pulsarChannelRateLimitingSecondsAfterResumeFromUnreadable}" seconds, the rate-limiter will be removed automatically. Once the bytes that are waiting to be sent out reach the pulsarChannelWriteBufferHighWaterMark\"， the timer will be reset.
```
lhotari: The detailed design should describe the way how this rate limiting is being implemented. 

In Netty, blocking IO threads is not allowed and therefore rate limiting should be implemented in a non-blocking way with throttling, using `ServerCnxThrottleTracker` (which coordinates toggling of autoread) and AsyncTokenBucket. When the limit is exceeded, the delay of the throttling could be calculated using AsyncTokenBucket's `calculateThrottlingDuration` to release the throttling (decreaseThrottleCount) using a scheduled task. 
The solution should be described in this PIP.

If this PIP gets backported to Pulsar 3.0.x, it will require a different solution since in Pulsar 3.0.x, the solution to handle throttling in a coordinated way is missing. That's something that could be described in a separate PIP for Pulsar 3.0.x. if not in this PIP.

pip/pip-434.md
poorbarcode: Added , please take a look again

pip/pip-434.md
poorbarcode: Mentioned it, please take a look

lhotari: This part looks ok now.

pip/pip-434.md
```diff
@@ -0,0 +1,95 @@
...
+- Stops receiving requests that come into the channel, which relies on the attribute `autoRead` of the `Netty channel`.
+- Netty sets `channel.writable` to `true` once the size of the data that is waiting to be sent out is less than `{pulsarChannelWriteBufferLowWaterMark}`
+- Starts to receive requests(sets `channel.autoRead` to `true`).
+  - Note: relies on `ServerCnxThrottleTracker`, which will track the "throttle count". When a throttling condition is present, the throttle count is increased and when it's no more present, the count is decreased. The autoread should be switched to false when the counter value goes from 0 to 1 and only when it goes back from 1 to 0 should it set to true again. The autoread flag is no more controlled directly from the rate limiters. Rate limiters are only responsible for their part and it's ServerCnxThrottleTracker that decides when autoread flag is toggled.  See more details [pip-322](https://github.com/apache/pulsar/blob/master/pip/pip-322.md)
+  - To avoid avalanches, Pulsar will start a timed rate-limiter, which limits the rate of handling the request backlog("pulsarChannelRateLimitingRateAfterResumeFromUnreadable" requests per second).
```
lhotari: "avalanches" -> "a thundering herd problem" (assuming you mean https://en.wikipedia.org/wiki/Thundering_herd_problem)

pip/pip-434.md
lhotari: This part is ok

pip/pip-434.md
poorbarcode: Modified the explaination

pip/pip-434.md
poorbarcode: I will use two separate PRs to implement the change
- The first PR does not involve `ServerCnxThrottleTracker`
- The second PR will reconstruct the code with `ServerCnxThrottleTracker`

PR #24511, opened 2025-07-15
[fix][broker] Fix deduplication replay might never complete for exceptions

BewareMyPower:
Main Issue: #24497

> P.S. I've found some other issues of #24497, so I will open more PRs for that.

### Motivation

If any exception is thrown by managed cursor's API or the callback, `MessageDeduplication#checkStatus` will never have a chance to complete, which will make the `PersistentTopic` never available unless unloading the namespace or topic.

### Modifications

- Add `ManagedLedgerUtils`, which includes a set of future-based managed ledger APIs, which don't need explicit exception catching for callback-based APIs.
- Add a `ManagedLedgerReplayTask` to perform replay on a managed ledger from a given position. It can handle any unexpected exception from the managed ledger APIs or callbacks.
- Leverage `ManagedLedgerReplayTask` in `MessageDeduplication` for recovery and `ManagedLedgerUtils` for safe snapshot taking. These two classes can be applied in some other places for safety as well, but it's beyond the scope of this PR.

### Verifying this change

- Add `ManagedLedgerReplayTaskTest` to verify `ManagedLedgerReplayTask` works in many cases.
- Add `checkStatusFail` to verify `MessageDeduplication#checkStatus` will complete exceptionally for unexpected exception.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/49


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java
```diff
@@ -0,0 +1,91 @@
...
+            @Override
+            public void markDeleteComplete(Object ctx) {
+                future.complete(null);
+            }
+
+            @Override
+            public void markDeleteFailed(ManagedLedgerException exception, Object ctx) {
+                future.completeExceptionally(exception);
+            }
+        }, new Object());
```
dao-jun: maybe pass null is OK

pulsar-broker/src/test/java/org/apache/pulsar/broker/BrokerMessageDeduplicationTest.java
```diff
@@ -84,5 +109,24 @@ public void markerMessageNotRecordPersistent() {
         }
     }
 
-
+    @Test
+    public void checkStatusFail() throws Exception {
```
dao-jun, line 113: LGTM

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java
BewareMyPower: Yes, I intended to use `null`, `new Object()` should be written by the auto completion.

PR #24515, opened 2025-07-15
[fix][ml] Fix asyncReadEntries might never complete if empty entries are read from BK

BewareMyPower:
Main Issue: #24497

### Motivation

When the `OpReadEntry#readEntriesComplete` is triggered by `RangeEntryCacheImpl#readFromStorage` with an empty list of entries, NPE will happen here: https://github.com/apache/pulsar/blob/bc5b933c0015beec8bb1907273c630730cffc853/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L95

Then `updateReadPosition` and `checkReadCompletion` will be skipped, where the result is very similar to https://github.com/apache/pulsar/issues/24497#issuecomment-3073740788
- The callback of `asyncReadEntries` will never complete
- The `entriesReadCount` will be updated but the `readPosition` won't
- `pendingReadOps` will not be reduced to 0

Unfortunately, from the existing analysis, the issue of 24497 might not be caused by this bug, but this NPE is still possible.

### Modifications

Remove the confusing `lastPosition` argument in `internalReadEntriesComplete`. It's only set when all entries have been filtered out and it needs to reset the `readPosition` in `ManagedLedgerImpl#internalReadFromLedger`. It can be replaced by a simple combination of `updateReadPosition` and `checkReadCompletion`.

In `internalReadEntriesComplete`, just call `checkReadCompletion` when no entries are read to trigger another round of read.

In addition, catch exceptions when triggering user-provided callback and print logs with better information. Without this change, the exception will be caught by `SingleThreadTask#safeRunTask` and the log will be `Error while running task:` that does not have description for a specific cursor or read position.

### Verifying this change

Make `EntryCache` configurable in tests add add `testReadNoEntries` to verify that the `asyncReadEntries` will eventually succeed even if one read operation returns an empty entry list.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java
@@ -21,6 +21,7 @@ package org.apache.bookkeeper.mledger.impl;
 import static org.apache.bookkeeper.mledger.ManagedLedgerException.getManagedLedgerException;
 import static org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.NULL_OFFLOAD_PROMISE;
 import static org.apache.pulsar.common.util.Runnables.catchingAndLoggingThrowables;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicates;
 import com.google.common.collect.BoundType;
 import com.google.common.collect.Maps;
@@ -47,6 +48,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import lombok.Getter;
@@ -170,7 +172,8 @@ public class ManagedLedgerFactoryImpl implements ManagedLedgerFactory {
                                     ManagedLedgerFactoryConfig config)
             throws Exception {
         this(metadataStore, new DefaultBkFactory(bkClientConfiguration),
-                true /* isBookkeeperManaged */, config, NullStatsLogger.INSTANCE, OpenTelemetry.noop());
+                true /* isBookkeeperManaged */, config, NullStatsLogger.INSTANCE, OpenTelemetry.noop(),
+                factory -> new RangeEntryCacheManagerImpl(factory, factory.scheduledExecutor, OpenTelemetry.noop()));
     }
 
     public ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore, BookKeeper bookKeeper)
@@ -184,12 +187,20 @@ public class ManagedLedgerFactoryImpl implements ManagedLedgerFactory {
         this(metadataStore, (policyConfig) -> CompletableFuture.completedFuture(bookKeeper), config);
     }
 
+    @VisibleForTesting
+    public ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore, BookKeeper bookKeeper,
+                                    Function<ManagedLedgerFactoryImpl, EntryCacheManager> entryCacheManagerCreator)
+            throws Exception {
+        this(metadataStore, __ -> CompletableFuture.completedFuture(bookKeeper), false,
+                new ManagedLedgerFactoryConfig(), NullStatsLogger.INSTANCE, OpenTelemetry.noop(),
+                entryCacheManagerCreator);
+    }
+
     public ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore,
                                     BookkeeperFactoryForCustomEnsemblePlacementPolicy bookKeeperGroupFactory,
                                     ManagedLedgerFactoryConfig config)
             throws Exception {
-        this(metadataStore, bookKeeperGroupFactory, false /* isBookkeeperManaged */,
-                config, NullStatsLogger.INSTANCE, OpenTelemetry.noop());
+        this(metadataStore, bookKeeperGroupFactory, config, NullStatsLogger.INSTANCE, OpenTelemetry.noop());
     }
 
     public ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore,
@@ -198,7 +209,8 @@ public class ManagedLedgerFactoryImpl implements ManagedLedgerFactory {
                                     OpenTelemetry openTelemetry)
             throws Exception {
         this(metadataStore, bookKeeperGroupFactory, false /* isBookkeeperManaged */,
-                config, statsLogger, openTelemetry);
+                config, statsLogger, openTelemetry, factory -> new RangeEntryCacheManagerImpl(factory,
+                        factory.scheduledExecutor, openTelemetry));
     }
 
     private ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore,
@@ -206,7 +218,9 @@ public class ManagedLedgerFactoryImpl implements ManagedLedgerFactory {
                                      boolean isBookkeeperManaged,
                                      ManagedLedgerFactoryConfig config,
                                      StatsLogger statsLogger,
-                                     OpenTelemetry openTelemetry) throws Exception {
+                                     OpenTelemetry openTelemetry,
+                                     Function<ManagedLedgerFactoryImpl, EntryCacheManager> entryCacheManagerCreator)
+            throws Exception {
         MetadataCompressionConfig compressionConfigForManagedLedgerInfo =
                 config.getCompressionConfigForManagedLedgerInfo();
         MetadataCompressionConfig compressionConfigForManagedCursorInfo =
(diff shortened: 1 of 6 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -67,28 +66,27 @@ public static OpReadEntry create(ManagedCursorImpl cursor, Position readPosition
...
+            return;
+        }
         // Filter the returned entries for individual deleted messages
         int entriesCount = returnedEntries.size();
         long entriesSize = 0;
-        for (int i = 0; i < entriesCount; i++) {
-            entriesSize += returnedEntries.get(i).getLength();
+        for (final var returnedEntry : returnedEntries) {
+            entriesSize += returnedEntry.getLength();
         }
```
lhotari: It's a common pattern in Pulsar source code to use a for loop in this type of cases. The reason for this is that this avoids creating an Iterator instance. I would avoid making such changes, although it would be "cleaner" to change this.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -217,5 +218,47 @@ public void recycle() {
...
 
+    private void complete(Object ctx) {
+        cursor.ledger.getExecutor().execute(() -> {
+            try {
+                callback.readEntriesComplete(entries, ctx);
+            } catch (Throwable throwable) {
+                log.error("[{}] readEntriesComplete failed (last position: {})", this, lastEntryPosition(), throwable);
+            } finally {
+                recycle();
+            }
```
lhotari: When an exception is thrown, it's better to not recycle the instance. Recycling isn't mandatory.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -217,5 +218,47 @@ public void recycle() {
...
+    }
+
+    private void fail(ManagedLedgerException e, Object ctx) {
+        try {
+            callback.readEntriesFailed(e, ctx);
+            cursor.ledger.mbean.recordReadEntriesError();
+        } catch (Throwable throwable) {
+            log.error("[{}] readEntriesFailed failed (exception: {})", this, e.getMessage(), throwable);
+        } finally {
+            recycle();
```
lhotari: When an exception is thrown, it's better to not recycle the instance. Recycling isn't mandatory.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
BewareMyPower: I can revert this change since it's not a necessary part of this part and keep the original commit id for `git blame`.

But I don't agree that it's a common pattern in Pulsar.

```
% find pulsar-broker/src/main -name "*.java" | xargs grep -n "for (.*:" | wc -l 
     343
```

Creating an Iterator (in range-for loop or other cases) is widely accepted in **Java world**, it's hardly a real concern.

BewareMyPower:
`testScan` failed after this change, mark it as drafted first

BewareMyPower:
@lhotari Could you take a look again?

lhotari (commented):
I forgot to submit these minor comments earlier.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
BewareMyPower: Agreed.

BewareMyPower: Addressed. BTW, I still retain the try-catch for better logging (see the updated PR description).

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java
```diff
@@ -198,15 +209,18 @@ public ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore,
...
     private ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore,
                                      BookkeeperFactoryForCustomEnsemblePlacementPolicy bookKeeperGroupFactory,
                                      boolean isBookkeeperManaged,
                                      ManagedLedgerFactoryConfig config,
                                      StatsLogger statsLogger,
-                                     OpenTelemetry openTelemetry) throws Exception {
+                                     OpenTelemetry openTelemetry,
+                                     Function<ManagedLedgerFactoryImpl, EntryCacheManager> entryCacheManagerCreator)
+            throws Exception {
         MetadataCompressionConfig compressionConfigForManagedLedgerInfo =
```
poorbarcode, line 225: Considering the following modification, which is simpler and clearer

- Setting the param `entryCacheManagerCreator` to be `nullable`
- In this constructor: calling `new RangeEntryCacheManagerImpl(this, scheduledExecutor, openTelemetry)` if the param is `null`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -67,7 +66,11 @@ public static OpReadEntry create(ManagedCursorImpl cursor, Position readPosition
         return op;
     }
 
-    void internalReadEntriesComplete(List<Entry> returnedEntries, Object ctx, Position lastPosition) {
+    private void internalReadEntriesComplete(List<Entry> returnedEntries) {
+        if (returnedEntries.isEmpty()) {
+            checkReadCompletion();
```
poorbarcode, line 72: We'd better add a warning log here, since it is not a normal scenario

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -217,5 +219,45 @@ public void recycle() {
         recyclerHandle.recycle(this);
     }
 
+    private void complete(Object ctx) {
+        cursor.ledger.getExecutor().execute(() -> {
+            try {
+                callback.readEntriesComplete(entries, ctx);
+                recycle();
```
poorbarcode, line 227: Should we call `recycle` in the finally block?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -67,7 +66,11 @@ public static OpReadEntry create(ManagedCursorImpl cursor, Position readPosition
         return op;
     }
 
-    void internalReadEntriesComplete(List<Entry> returnedEntries, Object ctx, Position lastPosition) {
+    private void internalReadEntriesComplete(List<Entry> returnedEntries) {
+        if (returnedEntries.isEmpty()) {
```
poorbarcode, line 70: Just add a marker: this line is the primary change

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -217,5 +219,45 @@ public void recycle() {
...
+        } catch (Throwable throwable) {
+            log.error("[{}] readEntriesFailed failed (exception: {})", this, e.getMessage(), throwable);
+        }
+    }
+
+    @Override
+    public String toString() {
+        final var cursor = this.cursor;
+        if (cursor != null) {
+            return cursor.ledger.getName() + ":" + cursor.getName();
```
poorbarcode: Can we also add `readPosition`, `entry size`, and `count` into the log, which is helpful for troubleshooting?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
BewareMyPower: See https://github.com/apache/pulsar/pull/24515#discussion_r2208418567. /cc @lhotari

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
BewareMyPower: Agreed

lhotari:
@BewareMyPower The "EntryCacheCreator" solution added in this PR is obsolete. There's already a way to intercept reads in Pulsar tests with PulsarMockReadHandleInterceptor.

example usage:
https://github.com/apache/pulsar/blob/86b036bf38d2fc49c34829406b1721e7cc715ca8/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java#L3148-L3156

Since the changes to add the "EntryCacheCreator" collide with #24363 changes, I'd suggest that we remove the "EntryCacheCreator" and update the tests added in this PR to use PulsarMockReadHandleInterceptor. I can do a PR to handle that.

lhotari:
the replacement using ReadHandleInterceptor is

```java
    bkc.setReadHandleInterceptor(new PulsarMockReadHandleInterceptor() {
            @Override
            public CompletableFuture<LedgerEntries> interceptReadAsync(long ledgerId, long firstEntry, long lastEntry,
                                                                       LedgerEntries entries) {
                if (firstRead.compareAndSet(true, false)) {
                    // LedgerEntriesImpl doesn't allow empty entries list.
                    // Implementing a dummy LedgerEntries that returns an empty list.
                    return CompletableFuture.completedFuture(new LedgerEntries() {
                        @Override
                        public org.apache.bookkeeper.client.api.LedgerEntry getEntry(long entryId) {
                            return null;
                        }

                        @Override
                        public Iterator<org.apache.bookkeeper.client.api.LedgerEntry> iterator() {
                            return EmptyIterator.INSTANCE;
                        }

                        @Override
                        public void close() {

                        }
                    });
                } else {
                    return CompletableFuture.completedFuture(entries);
                }
            }
        });
```

BewareMyPower:
Let me try this approach tomorrow

lhotari:
> Let me try this approach tomorrow

I already submitted the PR in #24552 since I'll use that to be able to rebase and test #24363.

PR #23817, opened 2025-01-06
[fix][broker] Remove failed OpAddEntry from pendingAddEntries

gaoran10:
### Motivation

The [PR](https://github.com/apache/pulsar/pull/23683) cached the exceptions for entry payload interceptor processor, but the failed OpAddEntry wasn't removed from pendingAddEntries, this will cause the next write to fail.

### Modifications

Remove failed OpAddEntry from pendingAddEntries.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24522, opened 2025-07-16
[fix][ml] Fix the possibility of message loss or disorder when ML PayloadProcessor processing fails

dao-jun:
### Motivation

In https://github.com/apache/pulsar/pull/23683 and https://github.com/apache/pulsar/pull/23817, we handled the failure of ML PayloadProcessor.

But it maybe lead to message lost or out-of-order.

You can reproduce the issue via `PublishWithMLPayloadProcessorTest`:
1. Publish `message-0` ~ `message-9` to the topic and consume.
2. When enable message deduplication,  we can consume the following messages:
```txt
Received message: message-0
Received message: message-1
Received message: message-2
Received message: message-4
Received message: message-5
Received message: message-6
Received message: message-7
Received message: message-8
Received message: message-9
```
**The `message-3` is lost**
3. When disable message deduplication,  we can consume the following messages:
```txt
Received message: message-0
Received message: message-1
Received message: message-2
Received message: message-4
Received message: message-5
Received message: message-6
Received message: message-7
Received message: message-8
Received message: message-9
Received message: message-0
Received message: message-1
Received message: message-2
Received message: message-3
Received message: message-4
Received message: message-5
Received message: message-6
Received message: message-7
Received message: message-8
Received message: message-9
```
** The `message-3` is after `message-4`, message is out-of-order.

The key-point is we didn't fail all the incoming messages until `Topic` un-fenced.

### Modifications

1. Fail all the incoming write request until topic un-fenced.
2. Add tests

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
@@ -433,7 +433,7 @@ public interface ManagedLedger {
      * **Implementation Note:** Downstream projects that support the ML PayloadProcessor should implement
      * this method. Otherwise, do not implement it.
      */
-    default void resetInterceptorException() {
+    default void unfenceForInterceptorException() {
         // Default implementation does nothing
     }
 
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
@@ -152,7 +152,7 @@ public class OpAddEntry implements AddCallback, CloseCallback, Runnable, Managed
                             .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
                 } catch (Exception e) {
                     ManagedLedgerException mle = new ManagedLedgerException.ManagedLedgerInterceptException(e);
-                    ml.interceptorException = mle;
+                    ml.fenceForInterceptorException(mle);
                     ml.pendingAddEntries.remove(this);
                     ReferenceCountUtil.safeRelease(duplicateBuffer);
                     log.error("[{}] Error processing payload before ledger write", ml.getName(), e);
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -4415,8 +4415,12 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
     }
 
     @Override
-    public void resetInterceptorException() {
+    public void unfenceForInterceptorException() {
         this.interceptorException = null;
+    }
+
+    protected void fenceForInterceptorException(ManagedLedgerException e) {
+        this.interceptorException = e;
     }
 
     @Override
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -4277,7 +4277,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
 
     private synchronized void unfence() {
         isFenced = false;
-        ledger.resetInterceptorException();
+        ledger.unfenceForInterceptorException();
         cancelFencedTopicMonitoringTask();
     }
 
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/intercept/ManagedLedgerInterceptorImplTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/intercept/ManagedLedgerInterceptorImplTest.java
@@ -502,7 +502,7 @@ public class ManagedLedgerInterceptorImplTest  extends MockedBookKeeperTestCase 
                     expectedException.add(t);
                     countDownLatch.countDown();
                 } finally {
-                    ledger.resetInterceptorException();
+                    ledger.unfenceForInterceptorException();
                 }
             } else {
                 try {
@@ -513,7 +513,7 @@ public class ManagedLedgerInterceptorImplTest  extends MockedBookKeeperTestCase 
                     expectedException.add(t);
                     countDownLatch.countDown();
                 } finally {
-                    ledger.resetInterceptorException();
+                    ledger.unfenceForInterceptorException();
                 }
             }
         }


pulsar-broker/src/test/java/org/apache/pulsar/broker/PublishWithMLPayloadProcessorTest.java
```diff
@@ -0,0 +1,120 @@
...
+        conf.setBrokerEntryPayloadProcessors(
+                Collections.singleton("org.apache.pulsar.broker.ManagedLedgerPayloadProcessor0"));
+    }
+
+    @AfterMethod(alwaysRun = true)
+    @Override
+    protected void cleanup() throws Exception {
+        super.internalCleanup();
+    }
+
```
BewareMyPower: Calling `internalSetup` and `internalCleanup` for each test case cost much time unnecessarily. You can set the namespace level policy of `brokerDeduplication` config like:

```java
admin.namespaces().setDeduplicationStatus("public/default", true);
```

pulsar-broker/src/test/java/org/apache/pulsar/broker/PublishWithMLPayloadProcessorTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Test(singleThreaded = true)
+    public void testPublishWithoutDeduplication() throws Exception {
+        String topic = "persistent://public/default/testPublishWithoutDeduplication";
+        conf.setBrokerDeduplicationEnabled(false);
+        super.internalSetup();
+        super.producerBaseSetup();
+        pulsarTestContext.getMockBookKeeper().addEntryDelay(1, TimeUnit.SECONDS);
```
BewareMyPower: Could you reduce the delay (e.g. 100ms)? 1 second for each message is too long.

BewareMyPower (commented):
We should take care of the test time when adding tests.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -133,6 +133,15 @@ public void setCloseWhenDone(boolean closeWhenDone) {
     public void initiate() {
         if (STATE_UPDATER.compareAndSet(OpAddEntry.this, State.OPEN, State.INITIATED)) {
             ByteBuf duplicateBuffer = data.retainedDuplicate();
+            // Fail the add operation if the managed ledger is in a state that prevents adding entries.
+            ManagedLedgerException exbw = ml.exceptionBeforeWrite;
+            if (exbw != null) {
+                ml.pendingAddEntries.remove(this);
+                this.failed(exbw);
+                ReferenceCountUtil.safeRelease(data);
+                recycle();
```
lhotari: It's better to avoid recycling in exceptional paths. This can help avoid hard-to-reproduce issues.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -143,10 +152,13 @@ public void initiate() {
...
                             .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
                 } catch (Exception e) {
+                    ManagedLedgerException mle = new ManagedLedgerException.ManagedLedgerInterceptException(e);
+                    ml.exceptionBeforeWrite = mle;
                     ml.pendingAddEntries.remove(this);
                     ReferenceCountUtil.safeRelease(duplicateBuffer);
                     log.error("[{}] Error processing payload before ledger write", ml.getName(), e);
-                    this.failed(new ManagedLedgerException.ManagedLedgerInterceptException(e));
+                    this.failed(mle);
+                    recycle();
```
lhotari: Don't recycle in exceptional execution path

pulsar-broker/src/test/java/org/apache/pulsar/broker/PublishWithMLPayloadProcessorTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.testng.annotations.Test;
+
+@Slf4j
+public class PublishWithMLPayloadProcessorTest extends ProducerConsumerBase {
+
+    @BeforeMethod(alwaysRun = true)
+    @Override
+    protected void setup() throws Exception {
+        conf.setBrokerEntryPayloadProcessors(
+                Collections.singleton("org.apache.pulsar.broker.ManagedLedgerPayloadProcessor0"));
```
BewareMyPower:
```suggestion
        conf.setBrokerEntryPayloadProcessors(Set.of(ManagedLedgerPayloadProcessor0.class.getName()));
```

Be IDE friendly that IDE can search the reference of `ManagedLedgerPayloadProcessor0`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
```diff
@@ -420,6 +420,13 @@ void asyncOpenCursor(String name, InitialPosition initialPosition, Map<String, L
      */
     long getOffloadedSize();
 
+    /**
+     * Reset Exception before write to null.
+     */
+    default void resetExceptionBeforeWrite() {
```
BewareMyPower: This method is confusing that how should it be implemented for a custom managed ledger implementation. Could it be better to introduce `fence` and `unfence` mechanisms to managed ledger and call them in the same method of `PersistentTopic`?

Besides, I think there might be some other solutions to avoid introducing a confusing default method to the managed ledger:
1. Maintain the exception in `PersistentTopic` in `addFailed` or `fence()` and fail the `asyncAddEntry` if the exception is not null. The challenge is that though `internalAsyncAddEntry` is synchronized, it's executed in the internal executor, so `asyncAddEntry` cannot synchronize the fence operation.
2. Maintain the exception in managed ledger interceptor.

Just some thoughts, I didn't try these solutions at the moment.

BewareMyPower: I see the 1st solution is not possible because when the topic is fenced, there have been `OpAddEntry` objects queued in the managed ledger.

image

I think introducing fence mechanism to ML might work

dao-jun: Yes, I add a delay here is for reproduce this case https://github.com/apache/pulsar/pull/24522/files#diff-dc0d94cf41b0e8f6f7a343c2da4faf18fb02632e67c84c2bb831a2550ccc4dc6R38

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
dao-jun: But the object needs recycle, when should I recycle it? @lhotari

pulsar-broker/src/test/java/org/apache/pulsar/broker/PublishWithMLPayloadProcessorTest.java
dao-jun: Add here delay is for testing an OpAddEntry can be finished before the previous one, I adjust it to 500ms

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
dao-jun: @BewareMyPower 
I understand add `resetExceptionBeforeWrite` maybe could lead to confusion, but `ManagedLedgerImpl.state` is very complex, and it's the core of pulsar, making hasty modifications may lead to serious problems.
The current impl is the easiest way to solve the issue, I think. Compromised towards complexity.

BewareMyPower: Anyway, the document of this method is very ambiguous.

Assuming I'm going to use managed ledger directly in the downstream rather than using the `PersistentTopic`, when should I call `resetExceptionBeforeWrite`? The comment only describes what this method does, but it does not describe when should it be called and the side effect after it's called.

I agree it should be done within the `managed-ledger` module. Is it possible to move it to the interceptor?

BewareMyPower: Because the exception could only be set when the method of interceptor fails.

BewareMyPower: If it's hard to use a different solution, please at least improve the document of this method.

dao-jun: @BewareMyPower add `resetExceptionBeforeWrite` to `ManagedLedger` is necessary, because we need to call the method after `PersistentTopic` un-fenced. 
But I need to improve the method name or doc.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
```diff
@@ -420,6 +420,20 @@ void asyncOpenCursor(String name, InitialPosition initialPosition, Map<String, L
...
+    /**
+     * Reset the exception that was thrown by the PayloadProcessor during the add entry operation to null.
+     * If one add entry operation failed due to the interceptor, it will fail all the incoming add entry operations,
+     * it's for keep the message ordering and consistency.
+     * <p>
+     * The method MUST be called after all the pendingAddOperations are completed (such as Topic un-fenced),
+     * otherwise it will cause the ML unable to write forever.
+     * <p>
+     * For downstream projects, if supports ML PayloadProcessor, it should impl the method. Otherwise, don't.
+     */
```
BewareMyPower:
```suggestion
    /**
     * Resets the exception thrown by the PayloadProcessor during an add entry operation to null.
     * <p>
     * **Context:** When an add entry operation fails due to an interceptor, all subsequent incoming add entry
     * operations will also fail. This behavior ensures message ordering and consistency.
     * <p>
     * **Important:** This method MUST only be called after all pending add operations are fully completed
     * (e.g., after a Topic is unfenced). Calling it prematurely will prevent the Managed Ledger (ML)
     * from being able to write indefinitely.
     * <p>
     * **Implementation Note:** Downstream projects that support the ML PayloadProcessor should implement
     * this method. Otherwise, do not implement it.
     */
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
gaoran10: Hi, @lhotari , could you explain the potential issue?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
```diff
@@ -420,6 +420,23 @@ void asyncOpenCursor(String name, InitialPosition initialPosition, Map<String, L
...
+     * Resets the exception thrown by the PayloadProcessor during an add entry operation to null.
+     * <p>
+     * **Context:** When an add entry operation fails due to an interceptor, all subsequent incoming add entry
+     * operations will also fail. This behavior ensures message ordering and consistency.
+     * <p>
+     * **Important:** This method MUST only be called after all pending add operations are fully completed
+     * (e.g., after a Topic is unfenced). Calling it prematurely will prevent the Managed Ledger (ML)
+     * from being able to write indefinitely.
+     * <p>
+     * **Implementation Note:** Downstream projects that support the ML PayloadProcessor should implement
```
gaoran10, line 433: Sorry, why should the ML PayloadProcessor implement this method? Do you mean the PayloadProcessor can perform some operation in this method? Such as resource clean-up.

gaoran10 (commented):
Good catch!

BewareMyPower:
@dao-jun @gaoran10 @poorbarcode @lhotari I opened a mail list to discuss about the topic for whether to recycle objects in exceptional paths. https://lists.apache.org/thread/po08w0tkhc7q8gc5khpdft6stxnr1v2y

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
BewareMyPower: > why should the ML PayloadProcessor implement this method?

It's the **Downstream projects** that support the ML PayloadProcessor should implement this method, not the **ML PayloadProcessor**. Because if there is no ML PayloadProcessor, no exception will be set for failure.

gaoran10: Sorry, making a misunderstanding. I want to ask why the Downstream projects need to implement this method, this method belongs to the interface of the ManagedLedger.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -133,6 +133,16 @@ public void setCloseWhenDone(boolean closeWhenDone) {
...
         if (STATE_UPDATER.compareAndSet(OpAddEntry.this, State.OPEN, State.INITIATED)) {
             ByteBuf duplicateBuffer = data.retainedDuplicate();
+            // Fail the add operation if the managed ledger is in a state that prevents adding entries.
+            ManagedLedgerException exbw = ml.interceptorException;
+            if (exbw != null) {
+                ml.pendingAddEntries.remove(this);
+                this.failed(exbw);
+                ReferenceCountUtil.safeRelease(data);
+                // Don't recycle object here, see: https://lists.apache.org/thread/po08w0tkhc7q8gc5khpdft6stxnr1v2y
+                // recycle();
```
gaoran10: Maybe we can remove this line.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -143,10 +153,14 @@ public void initiate() {
...
                 } catch (Exception e) {
+                    ManagedLedgerException mle = new ManagedLedgerException.ManagedLedgerInterceptException(e);
+                    ml.interceptorException = mle;
                     ml.pendingAddEntries.remove(this);
                     ReferenceCountUtil.safeRelease(duplicateBuffer);
                     log.error("[{}] Error processing payload before ledger write", ml.getName(), e);
-                    this.failed(new ManagedLedgerException.ManagedLedgerInterceptException(e));
+                    this.failed(mle);
+                    // Don't recycle object here, see: https://lists.apache.org/thread/po08w0tkhc7q8gc5khpdft6stxnr1v2y
+                    // recycle();
```
gaoran10: Same as the above case.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
dao-jun: It indicates that if there is a downstream project which is  implements the ManagedLedger interface directly and supports the PayloadProcessor, it also need to impl the method.

pulsar-broker/src/test/java/org/apache/pulsar/broker/PublishWithMLPayloadProcessorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    protected void setup() throws Exception {
+        conf.setBrokerEntryPayloadProcessors(
+                Collections.singleton(ManagedLedgerPayloadProcessor0.class.getName()));
+        super.internalSetup();
+        super.producerBaseSetup();
+        admin.tenants().createTenant("my-test-tenant", createDefaultTenantInfo());
+        admin.namespaces().createNamespace("my-test-tenant/my-test-ns");
+    }
+
+    @BeforeClass(alwaysRun = true)
```
lhotari: this should be `@AfterClass`

pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/MetricsAuthenticationTest.java
```diff
@@ -61,6 +61,7 @@ void testGetMetricsByAuthenticate() throws Exception {
 
     @Test
     void testGetMetricsByDefault() throws Exception {
+        conf.setAuthenticateMetricsEndpoint(false);
```
lhotari: no need to change this in this PR

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -132,8 +132,16 @@ public void setCloseWhenDone(boolean closeWhenDone) {
...
         if (STATE_UPDATER.compareAndSet(OpAddEntry.this, State.OPEN, State.INITIATED)) {
-            ByteBuf duplicateBuffer = data.retainedDuplicate();
+            // Fail the add operation if the managed ledger is in a state that prevents adding entries.
+            ManagedLedgerException exbw = ml.interceptorException;
+            if (exbw != null) {
+                ml.pendingAddEntries.remove(this);
+                this.failed(exbw);
+                // Don't recycle the object here, see: https://lists.apache.org/thread/po08w0tkhc7q8gc5khpdft6stxnr1v2y
+                return;
+            }
```
codelipenghui, line 142: We have `State` for the managed ledger which already have Fenced and `FencedForDeletion`. Is it better to  have a new one called `FencedForInterceptorException`? So that we can add a API for ManageLedger could called `unfenceForInterceptorException()`, is will be more clear than the newly added `resetInterceptorException()`?

codelipenghui: And the method `ManagedLedgerImpl#internalAsyncAddEntry` should already have the check for the for the `state`. Once we changed the state to `FencedForInterceptorException`, the new add operation will not be added to the pendingAddEntries?

But from this change, it allows the managed ledger to add the `OpAddEntry` to the pendingAddEntries first and then remove it from here.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -143,10 +151,14 @@ public void initiate() {
                     payloadProcessorHandle = ml.getManagedLedgerInterceptor()
                             .processPayloadBeforeLedgerWrite(this.getCtx(), duplicateBuffer);
                 } catch (Exception e) {
+                    ManagedLedgerException mle = new ManagedLedgerException.ManagedLedgerInterceptException(e);
+                    ml.interceptorException = mle;
```
codelipenghui: If we go with the `State` management solution mentioned above, it seems like we can just have a method to fence to the ManagedLedger, e.g. `fenceForInterceptorException()`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
dao-jun: The ML's state is very complex and it's core of the messaging and there is no precedent of the case.
FencedForDeletion, Fenced is the final state of the ML, WriteFailed is for create ledger failed, and since it's the failure before write to BK, we don't have a strong reason to close the current ledger and create a new one.

how about we just we just  rename rename  `resetInterceptorException()` to  `unfenceForInterceptorException()` ?

dao-jun: > And the method ManagedLedgerImpl#internalAsyncAddEntry should already have the check for the for the state. > Once we changed the state to FencedForInterceptorException, the new add operation will not be added to the 
> pendingAddEntries?

> But from this change, it allows the managed ledger to add the OpAddEntry to the pendingAddEntries first and 
> then remove it from here.

because there is more than one place to call `initiate()`, so I check the interceptor exception here.

codelipenghui: > because there is more than one place to call initiate(), so I check the interceptor exception here.

But the OpAddEntry is first added to pendingAddEntries and only removed later, correct? If we mark the managed ledger as not ready to accept new messages, shouldn’t we also prevent adding new items to pendingAddEntries?

Also, the new exception effectively introduces a new state for the managed ledger—rejecting all new entries when an interceptor exception is present. We’re now using a different mechanism to represent this state, which adds complexity (managing two conditions instead of one).

Note that the current implementation already checks isFenced() before adding entries to pendingAddEntries.

Using a dedicated state to represent this behavior would make the code easier to reason about. If the managed ledger enters a fenced state, it rejects all new requests. With the current state + exception approach, the ledger rejects new requests either when it’s fenced or when the interceptor exception is not null—effectively two ways of modeling one concept.

codelipenghui (approved):

PR #24523, opened 2025-07-16
[improve][broker] PIP-429: Optimize Handling of Compacted Last Entry by Skipping Payload Buffer Parsing 

BewareMyPower:
PIP: #24439 

### Verifying this change

The functionality is covered by existing tests. `testGetLastMessageIdForEncryptedMessage` is added to verify that `getLastMessageId` still works when encryption is enabled. It does not work without introducing the `compacted_batch_indexes` field to `MessageMetadata` because the broker side does not know how to deserialize the `SingleMessageMetadata`.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [x] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/43


pulsar-broker/src/main/java/org/apache/pulsar/compaction/TopicCompactionService.java
```diff
@@ -67,14 +60,27 @@ public interface TopicCompactionService extends AutoCloseable {
     * @param publishTime  the publish time of entry.
     * @return the first entry metadata that greater or equal to target publishTime, this entry can be null.
     */
-    CompletableFuture<Entry> findEntryByPublishTime(long publishTime);
+    CompletableFuture<Position> findEntryByPublishTime(long publishTime);
```
coderzc, line 63: Why do we need to change this?

BewareMyPower: See https://github.com/apache/pulsar/blob/master/pip/pip-429.md#public-api

> Change the return value of findEntryByPublishTime from CompletableFuture<Entry> to CompletableFuture<Position>. This change is made because the caller only needs the position of the last message, not the entire entry.

coderzc (approved):
LGTM

PR #24525, opened 2025-07-16
[improve][misc] Optimize topic list hashing so that potentially large String allocation is avoided

lhotari:
### Motivation

Pulsar's regular expression pattern-based consumers rely on hashing topic listings for a namespace. The hash for all topics in the namespace is calculated in multiple locations throughout the solution. For example, whenever a new topic is added or deleted, the hash will be recalculated for all topics in the namespace.

The current hash calculation implementation concatenates all topics into a single String instance, which is then passed to the hash calculation. This approach is suboptimal since the Guava Hashing API provides a way to calculate hashes of individual items. The current implementation causes unnecessary large java.lang.String and byte[] allocations. In the optimized version, no new java.lang.String instances are allocated, and the byte[] allocation consists of small allocations for each individual String instance (the topic name).

### Modifications

- Optimize topic list hashing by using the Guava Hashing API's Hasher, where individual items can be passed to hash calculation
- Add tests that ensure the modified hash calculation produces the same result as the previous implementation

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/topics/TopicList.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/topics/TopicList.java
@@ -18,9 +18,11 @@
  */
 package org.apache.pulsar.common.topics;
 
+import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
 import com.google.re2j.Pattern;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


3pacccccc:
Just simple tested the old vs. new method on my local machine. looks like the  new one uses less memory and runs faster.
old:
image
new:
image

3pacccccc (approved):

PR #24534, opened 2025-07-18
[fix][io] Fix Kinesis checkpoint mechanism to prevent data duplication

shibd:
### Motivation

The Kinesis Source connector had several critical issues in its checkpointing mechanism that could lead to data loss, data duplication, or stalled processing.

1.  **Incorrect `shardEnded` Handling**: The connector did not correctly implement the `ShardRecordProcessor#shardEnded` method. It checkpointed at a specific sequence number instead of the required `SHARD_END` marker, which can cause the entire stream processing to stall during Kinesis resharding events. Reference: [Developing KCL 2.x Consumers in Java - shardEnded](https://docs.aws.amazon.com/streams/latest/dev/develop-kcl-consumers-java.html#kcl-java-interface-shardrecordprocessor)
2.  **Unreliable Periodic Checkpoints**: The periodic checkpoint logic was tied to the `processRecords` method. This is unreliable in Kinesis's Enhanced Fan-Out (push) mode, where this method is not called on idle streams, preventing timely checkpoints.
3.  **Inaccurate Checkpoints for Aggregated Records**: The logic did not account for KPL aggregation, ignoring the `subSequenceNumber`. This caused the reprocessing of entire aggregated batches upon failure, leading to significant data duplication.
4.  **Unstable Worker ID**: The `workerId` was generated with a random UUID, causing KCL to treat every restart as a new worker and ignore existing checkpoints, leading to consistent data reprocessing.

### Modifications

This PR overhauls the Kinesis Source connector's checkpointing and record processing logic for correctness and robustness.

1.  **Stable Worker ID**:
    - The `workerId` is now generated from the stable `sourceContext.getInstanceId()` instead of a random UUID. This is critical for ensuring KCL can recognize a restarted processor and correctly resume from its last checkpoint.

2.  **Independent Checkpoint Scheduler**:
    - Introduced a `ScheduledExecutorService` in `KinesisSource` to drive periodic checkpoints, completely decoupling it from the `processRecords` message flow. This ensures timely checkpoints in both Polling and Enhanced Fan-Out modes.
    - The new scheduler uses a non-blocking, asynchronous retry mechanism with backoff for `ThrottlingException`.

3.  **Precise Checkpointing with `subSequenceNumber`**:
    - The connector now correctly handles de-aggregated records from the Kinesis Producer Library (KPL) by checkpointing both the `sequenceNumber` and `subSequenceNumber`.
    - An immutable `CheckpointSequenceNumber` record was introduced to atomically and safely manage the checkpoint state across threads.

4.  **Robust `shardEnded` and `shutdownRequested` Handling**:
    - Implemented a blocking `waitForInFlightRecords` helper with a timeout to ensure all queued records are processed before a final checkpoint.
    - `shardEnded` now correctly performs a `SHARD_END` checkpoint (with retries) on success and a best-effort checkpoint on timeout, fully complying with the KCL contract.

5.  **Concurrency and Safety Fixes**:
    - Fixed a race condition in `processRecords` by ensuring the in-flight record counter is incremented **before** a record is enqueued.
    - Replaced the shared `static CharsetDecoder` in `KinesisRecord` with a thread-safe call to prevent an `IllegalStateException`.

### Verifying this change

The entire `KinesisRecordProcessorTest` suite was overhauled to support testing the new asynchronous, scheduled checkpointing logic. The new tests verify:
- The asynchronous scheduler correctly triggers checkpoints after records are acknowledged.
- The checkpoint logic correctly advances only to a newer `CheckpointSequenceNumber`.
- The `shardEnded` method correctly performs a `SHARD_END` checkpoint on the success path.
- The `shardEnded` and `shutdownRequested` methods correctly time out and perform best-effort checkpoints on the failure path.

- [ ] The public API
- [ ] The schema
- [x] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

A dedicated `ScheduledExecutorService` has been introduced in `KinesisSource` to manage checkpointing asynchronously. Additionally, the `shardEnded` and `shutdownRequested` methods now block the KCL thread for a short period to wait for in-flight messages.

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR in forked repository: ```

### DIFF
--- a/pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
+++ b/pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
@@ -132,7 +132,7 @@ public class KinesisRecordProcessor implements ShardRecordProcessor {
                     checkpointExecutor.schedule(this::triggerCheckpoint, checkpointInterval, TimeUnit.MILLISECONDS);
                 }
             }
-        } catch (Exception e) {
+        } catch (Throwable e) {
             log.error("Error while triggering checkpoint for shard {}. Terminating.", kinesisShardId, e);
             sourceContext.fatal(e);
         }
--- a/pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
+++ b/pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
@@ -28,9 +28,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.io.core.SourceContext;
-import software.amazon.kinesis.exceptions.InvalidStateException;
 import software.amazon.kinesis.exceptions.KinesisClientLibDependencyException;
-import software.amazon.kinesis.exceptions.ShutdownException;
 import software.amazon.kinesis.exceptions.ThrottlingException;
 import software.amazon.kinesis.lifecycle.events.InitializationInput;
 import software.amazon.kinesis.lifecycle.events.LeaseLostInput;
@@ -82,10 +80,6 @@ public class KinesisRecordProcessor implements ShardRecordProcessor {
             log.info("Successfully checkpointed shard {} at {}", kinesisShardId, checkpoint);
             isCheckpointing.set(false);
             checkpointExecutor.schedule(this::triggerCheckpoint, checkpointInterval, TimeUnit.MILLISECONDS);
-        } catch (ShutdownException | InvalidStateException e) {
-            log.error("Caught a non-retryable exception for shard {} during checkpoint at {}. Terminating.",
-                    kinesisShardId, checkpoint, e);
-            sourceContext.fatal(e);
         } catch (ThrottlingException | KinesisClientLibDependencyException e) {
             if (attempt >= numRetries) {
                 log.error("Checkpoint for shard {} failed after {} attempts at {}. Terminating.",
@@ -97,6 +91,10 @@ public class KinesisRecordProcessor implements ShardRecordProcessor {
                 checkpointExecutor.schedule(() -> tryCheckpointWithRetry(checkpointer, checkpoint, attempt + 1),
                         backoffTime, TimeUnit.MILLISECONDS);
             }
+        } catch (Exception e) {
+            log.error("Caught a non-retryable exception for shard {} during checkpoint at {}. Terminating.",
+                    kinesisShardId, checkpoint, e);
+            sourceContext.fatal(e);
         }
     }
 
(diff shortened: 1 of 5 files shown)


pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
```diff
@@ -39,120 +44,178 @@
...
+            final CheckpointSequenceNumber currentCheckpoint = this.sequenceNumberNeedToCheckpoint;
+
+            if (checkpointer != null && currentCheckpoint != null && !currentCheckpoint.equals(
+                    lastCheckpointSequenceNumber)) {
+                tryCheckpointWithRetry(checkpointer, currentCheckpoint, 1);
+            } else {
+                isCheckpointing.set(false);
+                checkpointExecutor.schedule(this::triggerCheckpoint, checkpointInterval, TimeUnit.MILLISECONDS);
+            }
+        }
```
RobertIndie: We should add a try-catch for this method. Or inside the tryCheckpointWithRetry to catch any runtime exceptions/throwable like NPE. Because if there is an exception thrown in the `ScheduledExecutorService` thread, it won't print any logs and it's hard to debug.

pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
```diff
@@ -39,120 +44,178 @@
...
     public void processRecords(ProcessRecordsInput processRecordsInput) {
+        this.checkpointerRef.set(processRecordsInput.checkpointer());
         log.info("Processing {} records from {}", processRecordsInput.records().size(), kinesisShardId);
         long millisBehindLatest = processRecordsInput.millisBehindLatest();
 
         for (KinesisClientRecord record : processRecordsInput.records()) {
             try {
+                log.debug("Add record with sequence number {}:{} to queue for shard {}.",
+                        record.sequenceNumber(), record.subSequenceNumber(), kinesisShardId);
+                numRecordsInFlight.incrementAndGet();
```
RobertIndie: Is it better to move this line after the following catch block?

pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/KinesisRecordProcessor.java
```diff
@@ -39,120 +42,181 @@
...
+                final CheckpointSequenceNumber currentCheckpoint = this.sequenceNumberNeedToCheckpoint;
+                if (checkpointer != null && currentCheckpoint != null && !currentCheckpoint.equals(
+                        lastCheckpointSequenceNumber)) {
+                    tryCheckpointWithRetry(checkpointer, currentCheckpoint, 1);
+                } else {
+                    isCheckpointing.set(false);
+                    checkpointExecutor.schedule(this::triggerCheckpoint, checkpointInterval, TimeUnit.MILLISECONDS);
+                }
+            }
+        } catch (Exception e) {
```
RobertIndie: It's better to use Throwable

RobertIndie (approved):

PR #22829, opened 2024-06-03
[improve] Use Google re2/j library for user provided regexes

merlimat:
### Motivation

Google RE2/J library provide regex library that has worst case linear time processing of pattern, as opposed to standard JDK regexes which could have recursive processing. It's better to avoid these corner cases by just swapping the `Pattern` class implementation. It's a drop-in replacement for the basic methods we're using, and RE2/J is already a transitive dependency.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [X] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24537, opened 2025-07-18
[fix][misc] Fix topics pattern consumer backwards compatibility

lhotari:
## Motivation

PR #22829 introduced a breaking change by switching the regular expression library from JDK's `java.util.regex.Pattern` to RE2/J in the topics pattern consumer implementation. This change breaks backwards compatibility with advanced regular expressions that rely on features not supported by RE2/J.

**Problem**: RE2/J does not support certain regex features by design, such as negative lookaheads. A common use case is subscribing to all topics except those with specific prefixes. For example, the pattern `persistent://public/default/(?!(test2|test3|test4)-).+` should subscribe to all topics in the `public/default` namespace that don't start with `test2-`, `test3-`, or `test4-`.

**Current Behavior**: In Pulsar 4.0.x, this pattern fails with:
```
org.apache.pulsar.shade.com.google.re2j.PatternSyntaxException: error parsing regexp: invalid or unsupported Perl syntax: `(?!`
```

## Modifications

This PR introduces a configurable solution that maintains backwards compatibility while allowing users to choose their preferred regex implementation:

### 1. New TopicsPattern Interface
- Encapsulates topics pattern usage into a specific interface `TopicsPattern`
- Provides methods used by both client and broker for topics pattern consumers
- Abstracts the underlying regex implementation

### 2. Broker Configuration Option
Added `topicsPatternRegexImplementation` configuration with three options:

- **`RE2J_WITH_JDK_FALLBACK`** (default): Uses RE2J implementation with automatic fallback to JDK when pattern compilation fails with RE2/J. Provides backwards compatibility while maintaining performance for simple patterns.

- **`RE2J`**: Uses only the RE2J implementation. More performant but doesn't support all regex features (e.g., negative lookaheads). Recommended for security-conscious deployments to prevent regex DoS attacks.

- **`JDK`**: Uses the standard Java regex implementation. Supports all regex features but can be slower and potentially vulnerable to regex DoS attacks with malicious patterns.

### 3. Implementation Classes
- `RE2JTopicsPattern`: High-performance RE2J implementation
- `JDKTopicsPattern`: Full-featured JDK implementation  
- `MatchAllTopicsPattern`: Optimized implementation for `.*` patterns
- `TopicsPatternFactory`: Creates appropriate implementation based on configuration

### 4. Client Compatibility
- Maintains full backwards compatibility for Java client users
- Existing `Pattern.compile()` usage continues to work seamlessly
- The public Java client API hasn't changed although technically public methods have been modified

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/topics/JDKTopicsPattern.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/topics/JDKTopicsPattern.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.common.topics;
+
+import java.util.regex.Pattern;
+
+class JDKTopicsPattern implements TopicsPattern {
+    private final Pattern pattern;
+
+    public JDKTopicsPattern(String regex) {
+        this.pattern = Pattern.compile(regex);
+    }
+
+    public JDKTopicsPattern(Pattern pattern) {
+        this.pattern = pattern;
+    }
+
+    @Override
+    public boolean matches(String topicName) {
+        return pattern.matcher(topicName).matches();
+    }
+
+    @Override
+    public String pattern() {
+        return pattern.pattern();
+    }
+}
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/topics/JDKTopicsPattern.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/topics/JDKTopicsPattern.java
@@ -21,14 +21,17 @@ package org.apache.pulsar.common.topics;
 import java.util.regex.Pattern;
 
 class JDKTopicsPattern implements TopicsPattern {
+    private final String inputPattern;
     private final Pattern pattern;
 
-    public JDKTopicsPattern(String regex) {
-        this.pattern = Pattern.compile(regex);
+    public JDKTopicsPattern(String inputPattern, String regexWithoutTopicDomainScheme) {
+        this.inputPattern = inputPattern;
+        this.pattern = Pattern.compile(regexWithoutTopicDomainScheme);
     }
 
     public JDKTopicsPattern(Pattern pattern) {
         this.pattern = pattern;
+        this.inputPattern = pattern.pattern();
     }
 
     @Override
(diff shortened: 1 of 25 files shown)


pulsar-proxy/src/test/java/org/apache/pulsar/proxy/server/ProxyParserTest.java
```diff
@@ -110,7 +117,7 @@ public void testProducer() throws Exception {
         PulsarClient client = PulsarClient.builder().serviceUrl(proxyService.getServiceUrl())
                 .build();
         Producer<byte[]> producer =
-                client.newProducer(Schema.BYTES).topic("persistent://sample/test/local/producer-topic")
+                client.newProducer(Schema.BYTES).topic("persistent://sample/local/producer-topic")
```
BewareMyPower: Why did you modify many namespaces from V1 to V2 in tests? (`sample/test/local` to `sample/local`) It seems irrelevant with the core changes.

BewareMyPower: I see `Topic name format V1 is explicitly prevented`. But will these tests fail without these changes?

lhotari: I'll check later. Testing for V1 wouldn't make much sense since it's been deprecated since Pulsar 2.0.0. However I agree that the change should be made separately to switch from V1 to V2 topic names in tests. We should also have an option to globally block V1 topics in a broker.

BewareMyPower: Yes, I agree.

lhotari: @BewareMyPower I found the issue in supporting V1 topic names. I'll restore the support in this PR. 
We really need to remove V1 topics from Pulsar at some point since it's a mess to keep on supporting them, but that would be taken care on separately.

lhotari: I'll push a force push where these topic name V1 related changes are reverted.

pulsar-common/src/main/java/org/apache/pulsar/common/topics/JDKTopicsPattern.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.apache.pulsar.common.topics;
+
+import java.util.regex.Pattern;
+
+class JDKTopicsPattern implements TopicsPattern {
+    private final Pattern pattern;
+
+    public JDKTopicsPattern(String regex) {
+        this.pattern = Pattern.compile(regex);
```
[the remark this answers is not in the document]
lhotari: It's possible to configure the broker to use RE2J when a user is concerned about ReDoS vulnerabilities.

pulsar-common/src/main/java/org/apache/pulsar/common/topics/JDKTopicsPattern.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.regex.Pattern;
+
+class JDKTopicsPattern implements TopicsPattern {
+    private final String inputPattern;
+    private final Pattern pattern;
+
+    public JDKTopicsPattern(String inputPattern, String regexWithoutTopicDomainScheme) {
+        this.inputPattern = inputPattern;
+        this.pattern = Pattern.compile(regexWithoutTopicDomainScheme);
```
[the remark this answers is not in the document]
lhotari: It's possible to configure the broker to use RE2J when a user is concerned about ReDoS vulnerabilities.

PR #24539, opened 2025-07-21
[fix][client] Close orphan producer or consumer when the creation is interrupted

BewareMyPower:
### Motivation

The producer or reader creation has the same issue like what https://github.com/apache/pulsar/pull/24100 fixes for consumer: when the thread is interrupted, the returned future will complete exceptionally with the wrapped `InterruptedException`, but the asynchronous creation will continue in the background.

### Modifications

- Add a common `FutureUtils.wait` method to call a callback on the completed result.
- Apply the method above to the creation of producer, consumer and reader.
- Add `ClientInterruptTest` to cover the fixes above and remove duplicated `ConsumerCloseTest#testInterruptedWhenCreateConsumer`.

It should be noted that the complicated fix approach of https://github.com/apache/pulsar/pull/24100 is replaced by the solution in this PR.

This PR also reverts the change that the thread interrupt state is set, which should be the caller's responsibility to do that. This is actually a wrong behavior for many existing synchronous APIs.

Without this change, the 2nd creation of the consumer in the code snippet will fail when the 1st creation is interrupted:

```java
try {
    @Cleanup final var consumer = client.newConsumer().topic("tp").subscriptionName("sub")
            .subscribe();
    fail();
} catch (PulsarClientException e) {
    assertTrue(e.getCause() instanceof InterruptedException);
}
try {
    @Cleanup final var consumer = client.newConsumer().topic("another-tp").subscriptionName("sub")
            .subscribe();
} catch (PulsarClientException e) {
    fail();
}
```

If we want to make the 2nd call fail, the caller should handle the `InterruptException` like:

```java
} catch (PulsarClientException e) {
    if (e.getCause() instanceof InterruptedException) {
    }
}
```

In short,
- Before https://github.com/apache/pulsar/pull/24100, after the creation of producer, consumer or reader is interrupted, the thread interrupt flag won't be set.
- After https://github.com/apache/pulsar/pull/24100, after the creation of consumer is interrupted, the thread interrupt flag will be set.
- After this PR, after the creation of producer, consumer or reader is interrupted, the thread interrupt flag won't be set.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
@@ -35,6 +35,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -380,5 +381,20 @@ public class FutureUtil {
                     completableFuture.completeExceptionally(throwable);
                     return null;
                 });
+    }
+
+    /**
+     * When calling {@link CompletableFuture#get()} on a future, it could throw a {@link InterruptedException} when
+     * the thread is interrupted. However, the asynchronous task might not be interrupted and the result could be
+     * created successfully eventually. In this case, we might need to close the result.
+     */
+    public static <T> T wait(CompletableFuture<T> future, Consumer<T> closeCallback)
+            throws InterruptedException, ExecutionException {
+        try {
+            return future.get();
+        } catch (InterruptedException e) {
+            future.thenAccept(closeCallback);
+            throw e;
+        }
     }
 }
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientInterruptTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientInterruptTest.java
@@ -1,0 +1,188 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.client.impl;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.broker.intercept.MockBrokerInterceptor;
+import org.apache.pulsar.broker.service.BrokerService;
+import org.apache.pulsar.broker.service.Consumer;
+import org.apache.pulsar.broker.service.Producer;
+import org.apache.pulsar.broker.service.ServerCnx;
+import org.apache.pulsar.broker.service.persistent.PersistentTopic;
+import org.apache.pulsar.client.api.MessageId;
+import org.apache.pulsar.client.api.ProducerConsumerBase;
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.PulsarClientException;
+import org.apache.pulsar.common.naming.TopicName;
+import org.awaitility.Awaitility;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+@Slf4j
+@Test(groups = "broker-impl")
+public class ClientInterruptTest extends ProducerConsumerBase {
+
+    private final CreationInterceptor interceptor = new CreationInterceptor();
+    private int index = 0;
+    private PulsarClientImpl client;
+    private String topic;
+    private ExecutorService executor;
+    private CompletableFuture<Void> delayTriggered;
+
+    @BeforeClass
+    @Override
+    protected void setup() throws Exception {
+        super.internalSetup();
+        super.producerBaseSetup();
+        client = (PulsarClientImpl) PulsarClient.builder().serviceUrl(pulsar.getBrokerServiceUrl()).build();
+        pulsar.getBrokerService().setInterceptor(interceptor);
+    }
+
+    @AfterClass(alwaysRun = true)
+    @Override
+    protected void cleanup() throws Exception {
+        super.internalCleanup();
+    }
+
+    @BeforeMethod
+    public void setupTopic() {
+        interceptor.numConsumerCreated.set(0);
+        interceptor.numProducerCreated.set(0);
+        executor = Executors.newCachedThreadPool();
+        TopicName topicName = TopicName.get("test-topic-" + index++);
+        topic = topicName.toString();
+
+        final var mlPath = BrokerService.MANAGED_LEDGER_PATH_ZNODE + "/" + topicName.getPersistenceNamingEncoding();
+        delayTriggered = new CompletableFuture<>();
+        mockZooKeeper.delay(1000L, (op, path) -> {
+            final var result = path.equals(mlPath);
+            if (result) {
+                log.info("Injected delay for {} {}", op, path);
+                delayTriggered.complete(null);
+            }
+            return result;
+        });
+    }
+
+    @AfterMethod(alwaysRun = true, timeOut = 10000)
+    public void cleanupTopic() {
+        executor.shutdown();
+    }
+
+    @Test(timeOut = 10000)
+    public void testCreateProducer() throws Exception {
+        testCreateInterrupt("producer", () -> client.newProducer().topic(topic).create());
+    }
+
+    @Test(timeOut = 10000)
+    public void testSubscribe() throws Exception {
+        testCreateInterrupt("consumer", () -> client.newConsumer().topic(topic).subscriptionName("sub")
+                .subscribe());
+    }
+
+    @Test(timeOut = 10000)
+    public void testCreateReader() throws Exception {
+        testCreateInterrupt("reader", () -> client.newReader().topic(topic).startMessageId(MessageId.earliest)
+                .create());
+    }
+
+    private void testCreateInterrupt(String name, PulsarClientSyncTask task) throws Exception {
+        final var exception = new AtomicReference<PulsarClientException>();
+        final var threadInterrupted = new CompletableFuture<Boolean>();
+        final var future = executor.submit(() -> {
+            try {
+                task.run();
+                exception.set(new PulsarClientException("Task " + name + " succeeded"));
+            } catch (PulsarClientException e) {
+                exception.set(e);
+            }
+
+            try {
+                Thread.sleep(1);
+                threadInterrupted.complete(false);
+            } catch (InterruptedException __) {
+                threadInterrupted.complete(true);
+            }
+        });
+        delayTriggered.get();
+        future.cancel(true);
+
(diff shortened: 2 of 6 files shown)


BewareMyPower:
Actually @nodece raised the opinion about thread interruption here, but it's not addressed, maybe due to the misunderstanding. https://github.com/apache/pulsar/pull/24100#pullrequestreview-2704697654

IMO, the key point should be:

> When using the Pulsar Client/Admin, the users should not call the thread interrupt.

The client SDK should guarantee when the thread is interrupted, there will be no resource leak. But the SDK should not set the thread interruption flag, which might lead to unexpected behavior.

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -381,4 +382,19 @@ public static void safeRunAsync(Runnable runnable,
...
                 });
     }
+
+    /**
+     * When calling {@link CompletableFuture#get()} on a future, it could throw a {@link InterruptedException} when
+     * the thread is interrupted. However, the asynchronous task might not be interrupted and the result could be
+     * created successfully eventually. In this case, we might need to close the result.
+     */
+    public static <T> T wait(CompletableFuture<T> future, Consumer<T> closeCallback)
+            throws InterruptedException, ExecutionException {
```
Demogorgon314:
```suggestion
    /**
     * Blocks to get the result of a CompletableFuture, while ensuring resources are cleaned up
     * if the wait is interrupted.
     * <p>
     * If the current thread is interrupted while waiting, this method registers a cleanup action
     * to be executed when the future eventually completes. This prevents resource leaks that
     * could otherwise occur when an interruption happens but the underlying asynchronous task
     * finishes successfully later. After registering the action, it re-throws the
     * {@link InterruptedException}.
     *
     * @param future         The CompletableFuture to wait for.
     * @param cleanupAction  A consumer that performs a cleanup action (e.g., closing a resource)
     * on the result if the wait is interrupted.
     * @param <T>            The type of the future's result.
     * @return The computed result from the future.
     * @throws InterruptedException if the current thread was interrupted while waiting.
     * @throws ExecutionException   if the future completed exceptionally.
     */
    public static <T> T getAndCleanupOnInterrupt(CompletableFuture<T> future, Consumer<T> cleanupAction)
            throws InterruptedException, ExecutionException {
```
Change the method name to make it more clearer.

lhotari (commented):
> This PR also reverts the change that the thread interrupt flag is set, which should be the caller's responsibility to do that. This is actually a wrong behavior for many existing synchronous APIs.

I'm not sure if I agree on this. Whenever code in some method calls another method that throws `InterruptedException` and it is not propagated to the caller (the method doesn't throw InterruptedException in the method signature), it should set the interrupted flag with `Thread.currentThread().interrupt()` (possibly first handling the exception and then setting the interrupted flag). This is the way how InterruptedException should be handled in Java according to "Java Concurrency in Practice". More details in https://stackoverflow.com/questions/3976344/handling-interruptedexception-in-java, also with the reference to "Java Concurrency in Practice". 

For cancelling asynchronous tasks where a CompletableFuture is the result, `.cancel(true)` should be called and there should be an action that is done when the future is cancelled. There's a helper `FutureUtils.whenCancelledOrTimedOut` in Pulsar's code base for this purpose. This will only work if the `cancel` method is called on the instance where the `Runnable` has been "attached" with `FutureUtils.whenCancelledOrTimedOut`. (CompletableFuture doesn't have a way to propagate signals "upstream" like there is in ReactiveStreams)

lhotari:
> The client SDK should guarantee when the thread is interrupted, there will be no resource leak. But the SDK should not set the thread interruption flag, which might lead to unexpected behavior.

As mention in my previous comment, I disagree on this point.

BewareMyPower:
> Whenever code in some method calls another method that throws InterruptedException and it is not propagated to the caller (the method doesn't throw InterruptedException in the method signature)

The key point is that from my perspective, the `InterruptedException` is propagated to the caller from the underlying of `PulsarClientException`. It's an issue with the poor JavaDocs for client APIs. The synchronous API should note how should the caller handle the `PulsarClientException` except for simply logging it.

The real use case is that I'm using a single thread executor to perform two types of tasks:
- LOAD: create a reader and read a topic from a given position to load key-values of this topic
- UNLOAD: remove the key-values from this topic

They're synchronous tasks executed in a single thread, executed by `ExecutorService#submit`. Before scheduling each task, I will cancel the previous task (typically LOAD because it's blocking) by calling `.cancel(true)` and handling the possible `InterruptedException` after unwrapping `PulsarClientException`. If the interruption flag is set internally by the client SDK, there is no graceful way to cancel this flag because this thread could schedule new tasks in future.

You can see https://gist.github.com/BewareMyPower/fbd1e67fa947967c71b4722289ba7cea for reference.

Since I used `Reader` rather than `Consumer`, I didn't suffer the breaking change brought by https://github.com/apache/pulsar/pull/24100.

Since this PR uses a common method to handle interruption for creation of producer, reader and consumer. I have to choose whether to set the interruption flag. I chose NOT from the perspective of the downstream application developer.

BewareMyPower:
> But the SDK should not set the thread interruption flag, which might lead to unexpected behavior.

I'm correcting the description to "in certain cases".

For example, setting the interruption flag in `Consumer#receive` makes sense. Because once it's interrupted, it means the internal blocking queue has been interrupted. There is no way from the caller side to continue calling `receive` safely, so the client SDK must set the interruption flag to fail all subsequent `receive` calls in the same thread.

However, for this case, take consumer for example, if the creation failed:

```java
    @Cleanup final var consumer = client.newConsumer().topic("tp").subscriptionName("sub")
            .subscribe();
```

The `consumer` variable will never be touched by the caller. Setting the interruption flag could fail all subsequent blocking calls that might throw an `InterruptedException` (e.g. `Thread.sleep`), this would be really confusing.

BTW, @shibd I'm wondering that if you know the impact when you added `Thread.currentThread().interrupt();` in https://github.com/apache/pulsar/pull/24100?

shibd:
hi, @BewareMyPower @lhotari. Thank you all for discussing this.

> BTW, @shibd I'm wondering that if you know the impact when you added Thread.currentThread().interrupt(); in https://github.com/apache/pulsar/pull/24100?

Initially, I didn't think through the use cases carefully and just followed the best practice, as mentioned [here](https://github.com/apache/pulsar/pull/24539#pullrequestreview-3037017249)

I've reconsidered this case, and I can think of a scenario like this:

Suppose a user tries to create a consumer `twice` within a single thread. If the creation of the first consumer fails, the code then attempts to create the second one and subsequently receive messages.

If we don't restore the interrupt flag, the creation of the second consumer will succeed. This is not the expected behavior, because the user's intent was to stop the entire thread.


``` java
    @Test
    public void testInterruptedWhenCreateConsumer() throws InterruptedException {

        String tpName = BrokerTestUtil.newUniqueName("persistent://public/default/tp");
        String subName = "test-sub";
        String mlCursorPath = BrokerService.MANAGED_LEDGER_PATH_ZNODE + "/"
                + TopicName.get(tpName).getPersistenceNamingEncoding() + "/" + subName;

        // Make create cursor delay 1s
        CountDownLatch topicLoadLatch = new CountDownLatch(1);
        for (int i = 0; i < 5; i++) {
            mockZooKeeper.delay(1000, (op, path) -> {
                if (mlCursorPath.equals(path)) {
                    topicLoadLatch.countDown();
                    return true;
                }
                return false;
            });
        }

        Thread startConsumer = new Thread(() -> {
            // 1. First, try to create consumer
            Consumer<byte[]> consumer = null;
            try {
                 consumer = pulsarClient.newConsumer()
                        .topic(tpName)
                        .subscriptionName(subName)
                        .subscribe();
            } catch (PulsarClientException e) {
                log.error("failed to create consumer");
            }

            log.info("Thread stats {}", Thread.currentThread().isInterrupted());
            
            // 2. if creating pulsar consumer failed, try to create consumer again
            try {
                if (consumer == null) {
                    consumer = pulsarClient.newConsumer()
                            .topic(tpName)
                            .subscriptionName(subName)
                            .subscribe();
                }
                // receive message and do somthing
            } catch (PulsarClientException e) {
            }
            
            while (!Thread.currentThread().isInterrupted()) {
                try {
                    Message<byte[]> receive = consumer.receive();
                    log.info("Receive message {}", receive);
                } catch (PulsarClientException e) {
                }
            }
        });
        
        startConsumer.start();
        topicLoadLatch.await();
        // user try to interrupt the thread
        startConsumer.interrupt();

        PulsarClientImpl clientImpl = (PulsarClientImpl) pulsarClient;
        Awaitility.await().ignoreExceptions().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> {
            Assert.assertEquals(clientImpl.consumersCount(), 0);
        });
        log.info("Thread is alive {}", startConsumer.isAlive());
    }
```

BewareMyPower:
> the user's intent was to stop the entire thread.

You said, it's the user's intent. Then user can manually set the thread's interruption flag to fail the subsequent `receive` calls:

```java
            // 1. First, try to create consumer
            Consumer<byte[]> consumer = null;
            try {
                /* ... */
            } catch (PulsarClientException e) {
    // HERE
    if (e.getCause() instanceof InterruptedException) {
    }
            }
```

**Why does it have to be the client SDK that does that?** If the user don't want to stop the thread before, there is no way unless a very ugly solution like:

```java
            // 1. First, try to create consumer
            Consumer<byte[]> consumer = null;
            try {
                /* ... */
            } catch (PulsarClientException e) {
        // HERE
        if (e.getCause() instanceof InterruptedException) {
            try {
                Thread.sleep(1);
            } catch (InterruptedException ignored) {
            }
        }
            }
```

The code above is also anti-intuitive because when an `InterruptedException` is thrown, it usually means the thread interruption flag is cleared. See docs of `Thread.sleep`:

```
     * @throws  InterruptedException
     *          if any thread has interrupted the current thread. The
     *          <i>interrupted status</i> of the current thread is
     *          cleared when this exception is thrown.
```

My [previous comment](https://github.com/apache/pulsar/pull/24539#issuecomment-3095802016) already mentions this example. This is a real world case.

The key point is, JDK provides the ability to set the flag by `Thread#interrupt` but does not provide the ability to clear the flag directly (as my example shows, users can unset the flag by manually clearing the `InterruptedException`). When the creation of a consumer fails, the `PulsarClient`'s state is still healthy. There is no reason to fail the subsequent creation.

shibd (approved):
I've reconsidered your example, and I now understand that in a real-world use case, it is indeed difficult for the caller to handle the situation if we set the interrupt flag internally.

Maybe in the future, we should consider explicitly declaring that our methods throw InterruptedException instead of wrapping it in a PulsarClientException.

For now, I agree with the current approach of not setting the interrupt flag.

lhotari:
> The key point is, JDK provides the ability to set the flag by `Thread#interrupt` but does not provide the ability to clear the flag directly (as my example shows, users can unset the flag by manually clearing the `InterruptedException`). 

@BewareMyPower The interrupt flag can be cleared with the `interrupted` method in `java.lang.Thread`. It's not very intuitively named method.

> When the creation of a consumer fails, the `PulsarClient`'s state is still healthy. There is no reason to fail the subsequent creation.

When the current thread is interrupted, there's some sort of shutdown in progress. I agree that it's very context specific whether subsequent calls should fail or not. It depends a lot on which thread is in question. In most cases, when a thread's interrupted flag is set, that thread should finish up work after cleanup. Since when a thread is interrupted, there's some sort of shutdown in progress. Typical cases are that a task running in a ExecutorService is cancelled, or the complete JVM is shutdown and it's part of the shutdown sequence that threads get interrupted.
If there's a need to keep on calling blocking methods to cleanup, there are ways to handle that. 
For example, Guava contains `com.google.common.util.concurrent.Uninterruptibles` utilities which will run the work in the current thread after resetting the possible interrupted status and then restore the interrupted status.

lhotari:
> > Whenever code in some method calls another method that throws InterruptedException and it is not propagated to the caller (the method doesn't throw InterruptedException in the method signature)
> 
> The key point is that from my perspective, the `InterruptedException` is propagated to the caller from the underlying of `PulsarClientException`. It's an issue with the poor JavaDocs for client APIs. The synchronous API should note how should the caller handle the `PulsarClientException` except for simply logging it.
> 
> The real use case is that I'm using a single thread executor to perform two types of tasks:
> 
> * LOAD: create a reader and read a topic from a given position to load key-values of this topic
> * UNLOAD: remove the key-values from this topic
> 
> They're synchronous tasks executed in a single thread, executed by `ExecutorService#submit`. Before scheduling each task, I will cancel the previous task (typically LOAD because it's blocking) by calling `.cancel(true)` and handling the possible `InterruptedException` after unwrapping `PulsarClientException`. If the interruption flag is set internally by the client SDK, there is no graceful way to cancel this flag because this thread could schedule new tasks in future.
> 
> You can see https://gist.github.com/BewareMyPower/fbd1e67fa947967c71b4722289ba7cea for reference.
> 
> Since I used `Reader` rather than `Consumer`, I didn't suffer the breaking change brought by #24100.
> 
> Since this PR uses a common method to handle interruption for creation of producer, reader and consumer. I have to choose whether to set the interruption flag. I chose NOT from the perspective of the downstream application developer.

Java Concurrency in Practice is pretty clear about the correct handling of InterruptedException:
image

BewareMyPower:
Yes, the key point is that `InterruptedException` has already been propagated as the cause of the `PulsarClientException`. The issue is that it's not documented to be a formal behavior, but all methods, including the method that has set the interruption flag, all have the same behavior.

BewareMyPower:
Let me merge this PR first, since the key point of this PR is to close orphan producer or consumer.

But how to handle the interruption will be another interesting topic. This is a part that previous PRs hardly touched and discussed in details. I'm preparing a PIP at the moment. We can discuss more after it's out.

PR #24550, opened 2025-07-23
[improve][client] Terminate consumer.receive() when consumer is closed

3pacccccc:
### Motivation
Currently, when `consumer.receive()` is called, it blocks indefinitely even when the consumer is being closed. This can lead to threads hanging indefinitely in applications that need to gracefully shutdown consumers. The issue can be reproduced with:
``` java
Thread thread = new Thread(() -> {
    try {
        // sleep 0.1 second to close consumer to ensure consumer.receive() is triggerd
        Thread.sleep(1000);
        consumer.close();
    } catch (Exception e) {
    }
});
thread.start();
// hangs forever
consumer.receive();
```
in fact, this could happen in many scenario as long as ```consumer.receive()``` called before ```consumer.close()```

**so, this PR aims to ensure that `consumer.receive()` will immediately throw an `IllegalStateException` when the consumer is being closed, providing clear feedback and preventing thread hangs**
### Modifications
1.Added termination handling in GrowableArrayBlockingQueue to:

- Check termination flag in take() and poll() methods

- Throw IllegalStateException with "Queue is terminated" message when terminated

2.Modified ```ConsumerImpl.closeConsumerTasks()``` to terminate the incoming message queue

3.Added tests covering:

- Regular consumer close during receive()

- Zero queue size consumer receives a batchMessage and another thread close consuer

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/15

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1243,19 +1243,22 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         }
         negativeAcksTracker.close();
         stats.getStatTimeout().ifPresent(Timeout::cancel);
-        if (poolMessages) {
-            releasePooledMessagesAndStopAcceptNew();
-        }
-        incomingMessages.terminate();
+        processIncomingMessagesWhenClose();
     }
 
     /**
-     * If enabled pooled messages, we should release the messages after closing consumer and stop accept the new
+     * when consumer is closed.we should
+     * 1. terminate incomingMessages queue
+     * 2. If enabled pooled messages, we should release the messages after closing consumer and stop accept the new
      * messages.
      */
-    private void releasePooledMessagesAndStopAcceptNew() {
-        incomingMessages.terminate(message -> message.release());
-        clearIncomingMessages();
+    private void processIncomingMessagesWhenClose() {
+        if (poolMessages) {
+            incomingMessages.terminate(message -> message.release());
+            clearIncomingMessages();
+        } else {
+            incomingMessages.terminate(null);
+        }
     }
 
     void activeConsumerChanged(boolean isActive) {
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1246,6 +1246,7 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         if (poolMessages) {
             releasePooledMessagesAndStopAcceptNew();
         }
+        incomingMessages.terminate();
     }
 
     /**
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1243,22 +1243,9 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         }
         negativeAcksTracker.close();
         stats.getStatTimeout().ifPresent(Timeout::cancel);
-        processIncomingMessagesWhenClose();
-    }
-
-    /**
-     * when consumer is closed.we should
-     * 1. terminate incomingMessages queue
-     * 2. If enabled pooled messages, we should release the messages after closing consumer and stop accept the new
-     * messages.
-     */
-    private void processIncomingMessagesWhenClose() {
-        if (poolMessages) {
-            incomingMessages.terminate(message -> message.release());
-            clearIncomingMessages();
-        } else {
-            incomingMessages.terminate(null);
-        }
+        //terminate incomingMessages queue, stop accept the new messages and waking up blocked thread.
+        incomingMessages.terminate(Message::release);
+        clearIncomingMessages();
     }
 
     void activeConsumerChanged(boolean isActive) {
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
@@ -223,6 +223,9 @@ public class GrowableArrayBlockingQueue<T> extends AbstractQueue<T> implements B
                 if (timeoutNanos <= 0) {
                     return null;
                 }
+                if (terminated) {
+                    throw new IllegalStateException("Queue is terminated");
+                }
 
                 timeoutNanos = isNotEmpty.awaitNanos(timeoutNanos);
             }
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
@@ -447,7 +447,7 @@ public class GrowableArrayBlockingQueue<T> extends AbstractQueue<T> implements B
         // Signal waiting consumer threads to prevent indefinite blocking after termination
         headLock.lock();
         try {
-            isNotEmpty.signal();
+            isNotEmpty.signalAll();
         } finally {
             headLock.unlock();
         }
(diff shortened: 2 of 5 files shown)


pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
```diff
@@ -438,6 +444,20 @@ public void terminate(@Nullable Consumer<T> itemAfterTerminatedHandler) {
...
+        terminate(itemAfterTerminatedHandler);
+
+        // 2. Wake up waiting threads with proper handling
+        headLock.lock();
+        try {
+            isNotEmpty.signal();
+        } finally {
+            headLock.unlock();
+        }
+    }
```
lhotari: instead of adding a new terminate method, It's recommend adding the logic to signal the wake up directly in the existing terminate method.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
```diff
@@ -194,6 +194,9 @@ public T take() throws InterruptedException {
 
         try {
             while (SIZE_UPDATER.get(this) == 0) {
+                if (terminated) {
+                    throw new IllegalStateException("Queue is terminated");
+                }
```
lhotari: instead of throwing an exception, I'd recommend returning null. GrowableArrayBlockingQueue implements a BlockingQueue and using IllegalStateException would be violating the contract.
One possible solution would be to return null and the caller could check the consumer state when a null is returned and throw AlreadyClosedException in that case to indicate that the consumer has been closed.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
```diff
@@ -194,6 +194,9 @@ public T take() throws InterruptedException {
...
                 isNotEmpty.await();
             }
```
lhotari, line 202: there would need to be another check for "terminated" here since a thread might be waiting when the queue gets terminated.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1246,6 +1246,7 @@ private void closeConsumerTasks() {
         if (poolMessages) {
             releasePooledMessagesAndStopAcceptNew();
         }
+        incomingMessages.terminate();
```
lhotari: releasePooledMessagesAndStopAcceptNew already calls the other terminate method. Refactor this code since there should be a single terminate method in GroableArrayBlockingQueue that handles termination. I'd suggest renaming releasePooledMessagesAndStopAcceptNew and call it in all cases with the consumer to release new messages only in the case of pooled messages.

lhotari (commented):
Good catch! left some review comments

3pacccccc:
> Good catch! left some review comments

thank you so such for pointing these out, I've pushed a new commit, please take a look again.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
3pacccccc: > instead of throwing an exception, I'd recommend returning null. GrowableArrayBlockingQueue implements a BlockingQueue and using IllegalStateException would be violating the contract. One possible solution would be to return null and the caller could check the consumer state when a null is returned and throw AlreadyClosedException in that case to indicate that the consumer has been closed.

Actually return null also violating the contract, because the method ```BlockingQueue#take``` declared as ```@NotNull```, so I throw a InterruptedException, I think it might be more properly

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
3pacccccc: > there would need to be another check for "terminated" here since a thread might be waiting when the queue gets terminated.

I think no need to, because if a thread can access here, it means there is data exist in queue, so just return data no matter ```terminated``` or not. or maybe I don't get your point, we can dive deeper into this discussion

lhotari: oh yes, I guess it's fine since in terminate, putting to the queue will be blocked first and then the signaling will happen.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1243,18 +1243,22 @@ private void closeConsumerTasks() {
...
-    private void releasePooledMessagesAndStopAcceptNew() {
-        incomingMessages.terminate(message -> message.release());
-        clearIncomingMessages();
+    private void processIncomingMessagesWhenClose() {
+        if (poolMessages) {
+            incomingMessages.terminate(message -> message.release());
+            clearIncomingMessages();
+        } else {
+            incomingMessages.terminate(null);
+        }
```
lhotari: btw. it's odd that "clearIncomingMessages" is only called in the case of poolMessage. Since message.release() is a no-op when the message isn't a pooled message, it would be fine to remove the condition for poolMessages.
The implication of not calling "clearIncomingMessages" in all cases is that it won't release memory back to the memory controller for messages that are in the queue. That might also be a problem with the terminate callback.

lhotari: This is a different matter, but it would be great to check it while we are on this code.

3pacccccc: > btw. it's odd that "clearIncomingMessages" is only called in the case of poolMessage. Since message.release() is a no-op when the message isn't a pooled message, it would be fine to remove the condition for poolMessages. The implication of not calling "clearIncomingMessages" in all cases is that it won't release memory back to the memory controller for messages that are in the queue. That might also be a problem with the terminate callback.
@lhotari good point, I've pushed a new commit, LTAL

lhotari (approved):
LGTM, good work @3pacccccc

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
```diff
@@ -436,6 +443,14 @@ public void terminate(@Nullable Consumer<T> itemAfterTerminatedHandler) {
         } finally {
             tailLock.unlockWrite(stamp);
         }
+
+        // Signal waiting consumer threads to prevent indefinite blocking after termination
+        headLock.lock();
+        try {
+            isNotEmpty.signal();
```
lhotari: We should wake up all waiting threads.
```suggestion
            isNotEmpty.signalAll();
```
Please add a test case for more than one thread so that this case is also covered.

lhotari (changes requested):
There's a need to handle the case where more than one thread is waiting.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java
lhotari: I guess it would be extraordinary to use the Consumer API this way, but that's possible and the API doesn't prevent multiple threads to call `.receive()` at the same time.

3pacccccc:
> There's a need to handle the case where more than one thread is waiting.

IMO, because of the  ```headLock``` ensures that there is at most one thread can be waiting on the condition. so, I think the scenario ```handle the case where more than one thread is waiting.``` might not be existed.
@lhotari

lhotari:
> IMO, because of the `headLock` ensures that there is at most one thread can be waiting on the condition. so, I think the scenario `handle the case where more than one thread is waiting.` might not be existed.
> @lhotari

`isNotEmpty` has been created with `headLock.newCondition()`. When `.await()` is called, the lock is released. When `.signal` is called, the thread that was awaiting will acquire the lock again. The difference between `.signal` and `.signalAll` is that the first one will only wake up one thread. `.signalAll` will wake up one thread at a time.
There is a need to handle the multi-thread case.

3pacccccc:
> > IMO, because of the `headLock` ensures that there is at most one thread can be waiting on the condition. so, I think the scenario `handle the case where more than one thread is waiting.` might not be existed.
> > @lhotari
> 
> `isNotEmpty` has been created with `headLock.newCondition()`. When `.await()` is called, the lock is released. When `.signal` is called, the thread that was awaiting will acquire the lock again. The difference between `.signal` and `.signalAll` is that the first one will only wake up one thread. `.signalAll` will wake up one thread at a time. There is a need to handle the multi-thread case.

@lhotari yes! you're right. Thank you so much for pointing this out, I've pushed a new commit, PTAL

Issue #22157, opened 2024-02-29
[Bug] Broker memory leak 


graysonzeng:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Version

v3.1.1

### Minimal reproduce step

After running for a period of time, the broker memory will gradually increase and eventually lead to a restart.
![企业微信截图_6fc91d9b-5071-41bd-a698-d7c7c8a5b975](https://github.com/apache/pulsar/assets/21187269/74c6be15-bc83-448c-bce3-6f948e0884e6)

After the heap dump, it was found that many ManagedLedgerImpl instances were retained in the memory, and these instances occupied most of the memory.

![企业微信截图_a61faea0-4ef0-4488-905f-c91d5275cf37](https://github.com/apache/pulsar/assets/21187269/cfc71bbc-8593-4565-8b97-bd84be15db1f)

 Common Path To the Accumulation Point：
![image](https://github.com/apache/pulsar/assets/21187269/cae8ad90-e218-430b-b8f0-d57f2a981484)

![image](https://github.com/apache/pulsar/assets/21187269/6cd9c36d-8bba-48dd-89c6-46db03fcb0f2)

- [ ] 

### What did you expect to see?

Normal memory GC

### What did you see instead?

broker restart

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

dao-jun:
could you please upload the heap dump file？

graysonzeng:
The heap dump has reached 13GB and cannot be uploaded

dao-jun:
> The heap dump has reached 13GB and cannot be uploaded

compress it and upload to cloud driver, such as 百度云? it is very important to locate the root cause

lhotari:
> > The heap dump has reached 13GB and cannot be uploaded
> 
> compress it and upload to cloud driver, such as 百度云? it is very important to locate the root cause

@graysonzeng @dao-jun please notice that the heap dump could contain sensitive data. it should be never shared without encryption because of this.

lhotari:
I'd recommend adding https://github.com/vlsi/mat-calcite-plugin plugin to Eclipse MAT so that you can do SQL queries to the heap dump. Eclipse MAT has OQL support, but that's not so handy as the SQL queries where you can do anything that Calcite supports with SQL.

not useful for this case, but an example of a Calcite query using Eclipse MAT + Calcite plugin:
```
[...]

PR #22454, opened 2024-04-07
[fix][broker] Fix consumer stops receiving messages when with large backlogs processing

Technoboy-:
Fixes #22435

### Motivation

When the backlog size > `managedLedgerCursorBackloggedThreshold` (default 1000), the cursor will be set inactive first. so this will cause the non-durable cursor not to add to the `waiting cursor.` This leads to the consumer stopping to receive messages.
https://github.com/apache/pulsar/blob/2469b97b7e4de10fec64cc7ff1f4f46a410ad125/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java#L3082-L3090
https://github.com/apache/pulsar/blob/2469b97b7e4de10fec64cc7ff1f4f46a410ad125/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3816-L3824

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24551, opened 2025-07-23
[fix][broker] Fix Broker OOM due to too many waiting cursors and reuse a recycled OpReadEntry incorrectly

poorbarcode:
#### Background

**1. Pending read if no entries to read.**
- Cursor sets itself into `managedledger.waitingCursors` if there are no entries to read.
- Managed ledger triggers `cursor.notifyEntriesAvailable` once new entries were added.

**2, Double-check that it has more entries.**
- The cursor does an additional check if there are no entries to read<sup>[code-1]</sup>.
- The cursor delays the second checking if the configuration `newEntriesCheckDelayInMillis` is larger than `0`<sup>[code-1]</sup>.

**3. The variable waitingReadOp of cursor**
- It used to maintain an `OpReadEntry` to trigger a new read after getting notified that new entries were added.
- It prevents an issue from occurring, where more than one pending read is in progress

**4. The variable waitingCursors of Managed Cursor**
- It maintains the list of cursors that are waiting for new entries.

---

**Issue 1: The same cursor was added to `managedledger.waitingCursors` thousands of times.**
- Configurations
  - subscription type: `Failover`
- Reproduce flow
  - Add consumer(`C1`); it becomes an active consumer.
  - Add consumer(`C2`); C2 becomes the active consumer after a choice by the dispatcher
    - Triggers a `scheduleReadOnActiveConsumer`
    - **(Highlight)** The method `scheduleReadOnActiveConsumer` resets the variable `cursor.waitingReadOp`, but it forgets to remove the cursor from `managedledger.waitingCursors`
    - The issue occurs, which is the same as #22157. You can reproduce the issue with the new test `testWaitingCursorsCountAfterSwitchingActiveConsumers`

Screenshot 2025-07-24 at 00 01 32

---

**Issue 2: Reused a recycled OpReadEntry, which causes repeated delivery and other unexpected issues**
| time | cursor -> read entries | cursor -> cancel pending read | Other tasks |
| --- | --- | --- | --- |
| 1 | has no entries to read |
| 2 | create an `OpReadEntry` and set to `cursor.waitingReadOp` |
| 3 | Schedule a delay task to do the double confirm of whether Managed Ledger has entries to read |
| 4 | The delayed task is waiting to get run |
| 5 | | start to cancel pending read |
| 6 | | clear `cursor.waitingReadOp` |
| 7 | | recycle stored `OpReadEntry`, which was created by the task `read entries` |
| 8 | Run delayed task |
| 9 | | | reused the recycled `OpReadEntry` |
| 10 | Uses the `OpReadEntry` to read entries, but it has been recycled and reused by other tasks |

---

**Issue 3: The method `cursor.checkForNewEntries` forgets to remove cursor from `managedledger.waitingCursors`.**
- `cursor.checkForNewEntries` implements the double-check that checks whether there are entries to read.
- Clear `cursor.waitingReadOp` and trigger a reading if there are entries to read after the first check.
- The method forgot to remove the cursor from `managedledger.waitingCursors` after clearing the variable `cursor.waitingReadOp`

---

- **[code-1]**: https://github.com/apache/pulsar/blob/v4.0.5/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L1033-L1052

```java
            if (!WAITING_READ_OP_UPDATER.compareAndSet(this, null, op)) {
                op.recycle();
                callback.readEntriesFailed(new ManagedLedgerException.ConcurrentWaitCallbackException(), ctx);
                return;
            }
            if (getConfig().getNewEntriesCheckDelayInMillis() > 0) {
                ledger.getScheduledExecutor()
                        .schedule(() -> checkForNewEntries(op, callback, ctx),
                                getConfig().getNewEntriesCheckDelayInMillis(), TimeUnit.MILLISECONDS);
            } else {
                // If there's no delay, check directly from the same thread
                checkForNewEntries(op, callback, ctx);
            }
```

- **[code-2]**: https://github.com/apache/pulsar/blob/v4.0.5/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L1083-L1099

```java
            if (hasMoreEntries()) {
                if (WAITING_READ_OP_UPDATER.compareAndSet(this, op, null)) {
                    if (log.isDebugEnabled()) {
                        log.debug("[{}] [{}] Cancelled notification and scheduled read at {}", ledger.getName(),
                                name, op.readPosition);
                    }
                    PENDING_READ_OPS_UPDATER.incrementAndGet(this);
                    ledger.asyncReadEntries(op);
                } 
            }
```

### Modifications

- Fix the `3` issues
- Instead of modifying `managedledger.waitingCursors` and `cursor.waitingReadOp` everywhere, Managed Cursor handles the variables itself

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -1053,8 +1053,8 @@ public class ManagedCursorImpl implements ManagedCursor {
                             op.readPosition);
                 }
 
-                // Check again for new entries after the configured time, then if still no entries are available register
-                // to be notified
+                // Check again for new entries after the configured time, then if still no entries are available
+                // register to be notified.
                 if (getConfig().getNewEntriesCheckDelayInMillis() > 0) {
                     delayCheckForNewEntriesTask = new DelayCheckForNewEntriesTask(op, callback, ctx);
                 } else {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -168,7 +168,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         AtomicReferenceFieldUpdater.newUpdater(ManagedCursorImpl.class, OpReadEntry.class, "waitingReadOp");
     @SuppressWarnings("unused")
     private volatile OpReadEntry waitingReadOp = null;
-    private volatile DelayCheckForNewEntriesTask delayCheckForNewEntriesTask;
+    private DelayCheckForNewEntriesTask delayCheckForNewEntriesTask;
 
     public static final int FALSE = 0;
     public static final int TRUE = 1;
@@ -1065,57 +1065,55 @@ public class ManagedCursorImpl implements ManagedCursor {
         }
     }
 
+    private enum DelayCheckForNewEntriesTaskState {
+        INIT, RUNNING, CANCELLED, DONE
+    }
+
     private class DelayCheckForNewEntriesTask implements Runnable {
 
         private final OpReadEntry op;
         private final ReadEntriesCallback callback;
         private final Object ctx;
-        private ScheduledFuture<?> scheduledFuture;
-        /**
-         * 0: init
-         * 1: running
-         * 2: cancelled
-         * 3: done
-         */
-        private final AtomicInteger state = new AtomicInteger();
+        private final ScheduledFuture<?> scheduledFuture;
+        private DelayCheckForNewEntriesTaskState state = DelayCheckForNewEntriesTaskState.INIT;
 
         public DelayCheckForNewEntriesTask(OpReadEntry op, ReadEntriesCallback callback, Object ctx) {
             this.op = op;
             this.callback = callback;
             this.ctx = ctx;
-            start();
-        }
-
-        private void start() {
             scheduledFuture = ledger.getScheduledExecutor().schedule(this,
                     getConfig().getNewEntriesCheckDelayInMillis(), TimeUnit.MILLISECONDS);
         }
 
         @Override
         public void run() {
-            if (!state.compareAndSet(0, 1)) {
-                return;
-            }
             synchronized (pendingReadOpMutex) {
+                if (state != DelayCheckForNewEntriesTaskState.INIT) {
+                    return;
+                }
+                state = DelayCheckForNewEntriesTaskState.RUNNING;
                 checkForNewEntries(op, callback, ctx);
-                state.compareAndSet(1, 3);
+                state = DelayCheckForNewEntriesTaskState.DONE;
             }
         }
 
         public boolean isDone() {
-            return state.get() == 3;
+            return state == DelayCheckForNewEntriesTaskState.DONE;
         }
 
         public boolean cancel() {
-            // Not all implementations of Executor guarantee that the Runnable will be no long be executed after a
-            // successful cancel, such as Guava MoreExecutors, see also https://github.com/google/guava/blob
-            // /v32.1.2/guava/src/com/google/common/util/concurrent/MoreExecutors.java#L709.
-            // The current task gurantees.
-            if (state.compareAndSet(0, 2)) {
+            synchronized (pendingReadOpMutex) {
+                // Not all implementations of Executor guarantee that the Runnable will be no long be executed after a
+                // successful cancel, such as Guava MoreExecutors, see also https://github.com/google/guava/blob
+                // /v32.1.2/guava/src/com/google/common/util/concurrent/MoreExecutors.java#L709.
+                // The current task guarantees.
+                if (state != DelayCheckForNewEntriesTaskState.INIT) {
+                    return false;
+                }
+                state = DelayCheckForNewEntriesTaskState.CANCELLED;
                 scheduledFuture.cancel(false);
                 return true;
             }
-            return false;
         }
     }
 
@@ -1160,8 +1158,8 @@ public class ManagedCursorImpl implements ManagedCursor {
                     PENDING_READ_OPS_UPDATER.incrementAndGet(this);
                     ledger.asyncReadEntries(op);
                 } else {
-                    log.info("[{}] [{}] notification that new entries added was already be called, skipped the curren"
-                        + "t new entry checking", ledger.getName(), name);
+                    log.info("[{}] [{}] notification that new entries added was already be called, skipped the current"
+                        + " new entry checking", ledger.getName(), name);
                 }
             } else if (ledger.isTerminated()) {
                 // At this point we registered for notification and still there were no more available
@@ -1191,7 +1189,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                 return false;
             }
             Function<Boolean, Boolean> clearWaitingReadOp = removeWaitingCursor -> {
-                if (WAITING_READ_OP_UPDATER.compareAndSet(this, op,null)) {
+                if (WAITING_READ_OP_UPDATER.compareAndSet(this, op, null)) {
                     if (removeWaitingCursor) {
                         ledger.removeWaitingCursor(this);
                     }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -49,6 +49,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -157,10 +158,17 @@ public class ManagedCursorImpl implements ManagedCursor {
             MarkDeleteEntry.class, "lastMarkDeleteEntry");
     protected volatile MarkDeleteEntry lastMarkDeleteEntry;
 
+    /** Protects the method "asyncReadEntriesWithSkipOrWait" and "cancelPendingReadRequest" runs concurrently. **/
+    private final Object pendingReadOpMutex = new Object();
+    /**
+     *  'ManagedLedger.notifyCursors' relies on this CAS to avoid using "pendingReadOpMutex" to guarantee thread-safety,
+     *  which improved the performance of publishing messages.
+     */
     protected static final AtomicReferenceFieldUpdater<ManagedCursorImpl, OpReadEntry> WAITING_READ_OP_UPDATER =
         AtomicReferenceFieldUpdater.newUpdater(ManagedCursorImpl.class, OpReadEntry.class, "waitingReadOp");
     @SuppressWarnings("unused")
     private volatile OpReadEntry waitingReadOp = null;
+    private volatile DelayCheckForNewEntriesTask delayCheckForNewEntriesTask;
 
     public static final int FALSE = 0;
     public static final int TRUE = 1;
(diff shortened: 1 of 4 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1033,26 +1041,81 @@ public void asyncReadEntriesWithSkipOrWait(int maxEntries, long maxSizeBytes, Re
...
+        private final OpReadEntry op;
+        private final ReadEntriesCallback callback;
+        private final Object ctx;
+        private ScheduledFuture<?> scheduledFuture;
+        /**
+         * 0: init
+         * 1: running
+         * 2: cancelled
+         * 3: done
+         */
```
lhotari: Please don't use magic values in code.
Replace with constants or enums. (In Java, all Enums have integer values with `.ordinal()`)

BewareMyPower: The readability is bad even if there is a comment. For example, when I read `state.compareAndSet(0, 1)`, I need to go back here to see the corresponding meaning.

If you don't want to wrap an enum with `AtomicReference`, you can use an `AtomicInteger` like:

```java
    enum State {
        INIT,
        RUNNING,
        CANCELLED,
        DONE
    }

    private final AtomicInteger state = new AtomicInteger(State.INIT.ordinal());

    public void run() {
        if (!state.compareAndSet(State.INIT.ordinal(), State.RUNNING.ordinal())) {
            return;
        }
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -157,10 +158,17 @@ public class ManagedCursorImpl implements ManagedCursor {
...
+    private final Object pendingReadOpMutex = new Object();
+    /**
+     *  'ManagedLedger.notifyCursors' relies on this CAS to avoid using "pendingReadOpMutex" to guarantee thread-safety,
+     *  which improved the performance of publishing messages.
+     */
     protected static final AtomicReferenceFieldUpdater<ManagedCursorImpl, OpReadEntry> WAITING_READ_OP_UPDATER =
         AtomicReferenceFieldUpdater.newUpdater(ManagedCursorImpl.class, OpReadEntry.class, "waitingReadOp");
     @SuppressWarnings("unused")
     private volatile OpReadEntry waitingReadOp = null;
+    private volatile DelayCheckForNewEntriesTask delayCheckForNewEntriesTask;
```
lhotari: you don't need `volatile` if this field in accessed (read or written) always under the synchronized block for `pendingReadOpMutex`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1033,26 +1041,81 @@ public void asyncReadEntriesWithSkipOrWait(int maxEntries, long maxSizeBytes, Re
...
-                        .schedule(() -> checkForNewEntries(op, callback, ctx),
-                                getConfig().getNewEntriesCheckDelayInMillis(), TimeUnit.MILLISECONDS);
-            } else {
-                // If there's no delay, check directly from the same thread
+    private class DelayCheckForNewEntriesTask implements Runnable {
+
+        private final OpReadEntry op;
+        private final ReadEntriesCallback callback;
+        private final Object ctx;
+        private ScheduledFuture<?> scheduledFuture;
```
lhotari: `scheduledFuture` could be final too if the start method is inlined.

pulsar-broker/src/test/java/org/apache/pulsar/client/api/FailoverSubscriptionTest.java
```diff
@@ -0,0 +1,90 @@
...
+            consumerMap.put(consumerLoop.getConsumerName(), consumerLoop);
+            if (dispatcher != null && dispatcher.getActiveConsumer() != null) {
+                String activeConsumer = dispatcher.getActiveConsumer().consumerName();
+                consumerMap.remove(activeConsumer).close();
+            }
+        }
+
+        ManagedLedgerImpl managedLedger = (ManagedLedgerImpl) topic.getManagedLedger();
+        ConcurrentLinkedQueue<ManagedCursorImpl> waitingCursors =
+                WhiteboxImpl.getInternalState(managedLedger, "waitingCursors");
```
BewareMyPower: Please avoid using reflection in tests whenever possible. It makes code hard to refactor. Just add a getter with `@VisibleForTesting`.

BewareMyPower (changes requested):

pulsar-broker/src/test/java/org/apache/pulsar/client/api/FailoverSubscriptionTest.java
BewareMyPower: Actually there is a `getWaitingCursorsCount()` method. The code can be replaced with

```java
        Awaitility.await().untilAsserted(() -> assertEquals(managedLedger.getWaitingCursorsCount(), 1));
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1089,16 +1152,16 @@ private void checkForNewEntries(OpReadEntry op, ReadEntriesCallback callback, Ob
...
+                            name, op.readPosition);
                     }
                     PENDING_READ_OPS_UPDATER.incrementAndGet(this);
                     ledger.asyncReadEntries(op);
                 } else {
-                    if (log.isDebugEnabled()) {
-                        log.debug("[{}] [{}] notification was already cancelled", ledger.getName(), name);
-                    }
+                    log.info("[{}] [{}] notification that new entries added was already be called, skipped the curren"
+                        + "t new entry checking", ledger.getName(), name);
```
BewareMyPower: Could you avoid splitting a word  (`"current"`) into two string literals?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1121,11 +1184,49 @@ public boolean cancelPendingReadRequest() {
...
-        if (op != null) {
-            op.recycle();
+        synchronized (pendingReadOpMutex) {
+            final OpReadEntry op = WAITING_READ_OP_UPDATER.get(this);
+            // Case 1: the pending read has executed, or there is no pending read.
+            if (op == null) {
+                return false;
+            }
+            Function<Boolean, Boolean> clearWaitingReadOp = removeWaitingCursor -> {
+                if (WAITING_READ_OP_UPDATER.compareAndSet(this, op,null)) {
```
BewareMyPower:
```suggestion
                if (WAITING_READ_OP_UPDATER.compareAndSet(this, op, null)) {
```

I think checkstyle should fail. Have you run the checkstyle check locally?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1033,26 +1041,81 @@ public void asyncReadEntriesWithSkipOrWait(int maxEntries, long maxSizeBytes, Re
...
 
-            if (!WAITING_READ_OP_UPDATER.compareAndSet(this, null, op)) {
-                op.recycle();
-                callback.readEntriesFailed(new ManagedLedgerException.ConcurrentWaitCallbackException(), ctx);
-                return;
-            }
+            synchronized (pendingReadOpMutex) {
+                if (!WAITING_READ_OP_UPDATER.compareAndSet(this, null, op)) {
+                    op.recycle();
+                    callback.readEntriesFailed(new ManagedLedgerException.ConcurrentWaitCallbackException(), ctx);
```
BewareMyPower, line 1047: You can call `OpReadEntry#fail` (make it public) instead

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1033,26 +1041,81 @@ public void asyncReadEntriesWithSkipOrWait(int maxEntries, long maxSizeBytes, Re
...
+            scheduledFuture = ledger.getScheduledExecutor().schedule(this,
+                    getConfig().getNewEntriesCheckDelayInMillis(), TimeUnit.MILLISECONDS);
+        }
+
+        @Override
+        public void run() {
+            if (!state.compareAndSet(0, 1)) {
+                return;
+            }
+            synchronized (pendingReadOpMutex) {
```
BewareMyPower, line 1090: Actually I think making `state` atomic is unnecessarily complicated.

All modifications on `state` are:
- CAS(0, 1) in `run` without `synchronized (pendingReadOpMutex)`
- CAS(1, 3) in `run` with `synchronized (pendingReadOpMutex)`
- CAS(0, 2) in `cancel`, which is always called in `cancelPendingReadRequest`'s `synchronized (pendingReadOpMutex)` code block

The only read operation of `state` is `isDone()`, which is always called in `cancelPendingReadRequest`'s `synchronized (pendingReadOpMutex)` code block.

i.e. only CAS(0, 1) in `run()` is not protected by `pendingReadOpMutex`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
BewareMyPower: +1

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -383,15 +383,12 @@ public synchronized void removeConsumer(Consumer consumer, boolean isResetCursor
...
-                                topic.getManagedLedger().removeWaitingCursor(cursor);
                             } catch (InterruptedException | ManagedLedgerException e) {
                                 log.warn("[{}] [{}] Failed to remove non durable cursor", topic.getName(), subName, e);
                             }
                         }
                     }, topic.getBrokerService().pulsar().getExecutor());
                 });
-            } else {
-                topic.getManagedLedger().removeWaitingCursor(cursor);
             }
```
Technoboy-, line 392: I remember this was fixing a non-durable sub issue, need to see the CI result

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

pulsar-broker/src/test/java/org/apache/pulsar/client/api/FailoverSubscriptionTest.java
poorbarcode: Addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: This PR just keeps the previous behavior: https://github.com/apache/pulsar/pull/24551/files#diff-fad355f91bd15cc041161f9a46fce62b7fee87fbfb8f0ff8a8b724a1bd1f29eeL1038

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Addressed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
poorbarcode: Yes, the test [testAddWaitingCursorsForNonDurable2](https://github.com/apache/pulsar/pull/22454/files#diff-2ec08bef8a943b1a830d9a1eef04f2a4eec2bf0d80692005c59a993137c74035R719) passed

BewareMyPower (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1033,25 +1041,78 @@ public void asyncReadEntriesWithSkipOrWait(int maxEntries, long maxSizeBytes, Re
...
+                            op.readPosition);
+                }
+
+                // Check again for new entries after the configured time, then if still no entries are available
+                // register to be notified.
+                if (getConfig().getNewEntriesCheckDelayInMillis() > 0) {
+                    delayCheckForNewEntriesTask = new DelayCheckForNewEntriesTask(op, callback, ctx);
+                } else {
+                    // If there's no delay, check directly from the same thread
+                    checkForNewEntries(op, callback, ctx);
```
lhotari, line 1062: is it possible that this results in a deadlock since this is called while holding `pendingReadOpMutex`?

PR #22206, opened 2024-03-06
[fix] [client] PIP-344 Do not create partitioned metadata when calling pulsarClient.getPartitionsForTopic(topicName)

poorbarcode:
### Motivation

see also [PIP-344 Correct the behavior of the public API pulsarClient.getPartitionsForTopic(topicName)](https://github.com/apache/pulsar/pull/22182)

#### Background
- When calling `pulsarClient.getPartitionsForTopic(topicName)`, Pulsar will automatically create the partitioned topic metadata if it does not exist, either using `HttpLookupService` or `BinaryProtoLookupService`.
- The partitioned topic auto-creation is dependent on `pulsarClient.getPartitionsForTopic`
- It triggers partitioned metadata creation by `pulsarClient.getPartitionsForTopic`
- And triggers the topic partition creation by producers' registration and consumers' registration.

#### Issue
Since the method `pulsarClient.getPartitionsForTopic` is a public API, and it is named `getxxx,` it should edit nothing.

### Modifications

- Do not create partitioned metadata when calling pulsarClient.getPartitionsForTopic(topicName)

### Next PRs
- Section [Backward & Forward Compatibility](https://github.com/apache/pulsar/blob/master/pip/pip-344.md#backward--forward-compatibility)
- Section 2 of [Goals](https://github.com/apache/pulsar/blob/master/pip/pip-344.md#goals): Instead of returning a 0 partitioned metadata, respond to a not found error when calling pulsarClient.getPartitionsForTopic(String) if the topic does not exist.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #10708, opened 2021-05-26
On multi-topic consumer, we shouldn't keep checking the partitioned metadata

merlimat:
### Motivation

When using a multi-topic consumer, either through the regex or by passing a list of topics, we are ending up checking the partitioned topic metadata every 1 min, even if the topics are not partitioned. 

We don't need to re-check that information if the topics are not partitioned. Also, this can be causing a lot of read traffic on ZK since we keep checking that "partitioned-metadata" that these topics don't have and, in 2.7, this information is not cached in brokers when the metadata is not found. 

### Modifications

When adding non-partitioned topics to the multi-topic consumer, use 0 for partitions when the topic is not partitioned in order to skip the periodic re-check of the partitions count.

PR #24554, opened 2025-07-24
[fix][client] Partitioned topics are unexpectedly created by client after deletion

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/22206 misses handling the auto partitions update feature. By default, a producer or consumer schedules a periodical task to query the partition metadata via the old `PulsarClient#getPartitionsForTopic` API, which will trigger the partitioned topic auto creation.

This issue does not exist when `allowAutoTopicCreationType` is `non-partitioned` (default) because the number of partitions is 0 for non-partitioned topics, while
- Producer does not support partitions shrinking
- Consumer will ignore querying partition metadata of non-partitioned topics after https://github.com/apache/pulsar/pull/10708

### Modifications

- Use the new `getPartitionsForTopic` overload introduced in PIP-344 to disable topic auto creation in the auto partitions update task
- Add `testPartitionsNotCreatedAfterDeletion` to verify the change
- Ignore the exception from `ManagedCursor#asyncClearBacklog` when deleting the topic, otherwise the test will fail for the 3rd `verifier.accept`. We might need a better solution for this issue in future because the exception is `CursorAlreadyClosedException`.

NOTE: Since it relies on PIP-344, it can only be cherry-picked to branch-4.0.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/TopicAutoCreationTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/TopicAutoCreationTest.java
@@ -24,6 +24,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import static org.testng.Assert.assertThrows;
 import static org.testng.Assert.assertTrue;
+import java.io.Closeable;
 import java.net.InetSocketAddress;
 import java.time.Duration;
 import java.util.List;
@@ -37,6 +38,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.client.admin.PulsarAdminException;
 import org.apache.pulsar.client.api.ClientBuilder;
 import org.apache.pulsar.client.api.Consumer;
+import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.Producer;
 import org.apache.pulsar.client.api.ProducerConsumerBase;
 import org.apache.pulsar.client.api.PulsarClient;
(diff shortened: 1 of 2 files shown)


lhotari:
> NOTE: Since it relies on PIP-344, it can only be cherry-picked to branch-4.0.

@BewareMyPower PIP-344 has been backported to branch-3.0 and branch-3.3 too. https://lists.apache.org/thread/23ymkfdn5mhfsk1wjym8m1h0sgj6l937

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/TopicAutoCreationTest.java
```diff
@@ -248,4 +254,39 @@ public void testClientWithAutoCreationGotNotFoundException() throws PulsarAdminE
...
+        };
+        verifier.accept(() -> client.newProducer().topic(topic)
+                .autoUpdatePartitionsInterval(interval.toSecondsPart(), TimeUnit.SECONDS).create());
+        verifier.accept(() -> client.newConsumer().topic(topic).subscriptionName("sub")
+                .autoUpdatePartitionsInterval(interval.toSecondsPart(), TimeUnit.SECONDS).subscribe());
+        verifier.accept(() -> client.newReader().topic(topic).startMessageId(MessageId.earliest)
+                .autoUpdatePartitionsInterval(interval.toSecondsPart(), TimeUnit.SECONDS).create());
+    }
+
+    private interface ThrowableConsumer<T> {
```
lhotari, line 283: btw. for future test cases, it's possible to use AssertJ's `org.assertj.core.api.ThrowingConsumer` together with AssertJ's "satisfies". I'd recommend switching to use AssertJ since the failure messages are better and there's a lot of flexibility with it's fluent API. We should migrate towards AssertJ assertions in our code base.

lhotari:
I just noticed that ["PIP-414: Enforce topic consistency check"](https://github.com/apache/pulsar/blob/master/pip/pip-414.md) changes in #24118 are slightly related. /cc @nodece

lhotari:
@BewareMyPower The test fails in branch-4.0. I'll postpone this for 4.0.7 . Would you be able to backport this to branch-4.0 in a separate PR where we could first verify that tests pass before merging?

BewareMyPower:
Yes, let me fix the tests

lhotari:
backport for maintenance branches in #24571

Shawyeok:
@BewareMyPower @poorbarcode 
This patch has been backported to the 3.0.x branch and released in `3.0.13`.

However, pulsar-client `3.0.13` is incompatible with older brokers (versions without PIP-344), even if the topic already exists on the broker or `allowAutoTopicCreation` is true.  version `3.0.12` is ok.

Is this the expected behavior?

```
20:28:25.218 [pulsar-client-lookup-27-1] ERROR org.apache.pulsar.client.impl.PartitionedProducerImpl - [persistent://public/default/test-250926] Auto getting partitions failed
java.util.concurrent.CompletionException: org.apache.pulsar.client.api.PulsarClientException$FeatureNotSupportedException: The feature of getting partitions without auto-creation is not supported by the broker. Please upgrade the broker to version that supports PIP-344 to resolve this issue.
```

BewareMyPower:
@Shawyeok Could you open discussion in mail list so that more people will see it? I think it's an issue with the compatibility with the semantic of `getPartitionedTopicMetadata` after PIP-344, not only this PR. Client 4.x should have the same issue.

IMO, the code here https://github.com/apache/pulsar/blob/1050f48035c97a14094b812467f046f6d37ff6f4/pulsar-client/src/main/java/org/apache/pulsar/client/impl/BinaryProtoLookupService.java#L301-L308 should be changed to something like:

```java
            ByteBuf request;
            if (clientCnx.isSupportsGetPartitionedMetadataWithoutAutoCreation()) {
                request = Commands.newPartitionMetadataRequest(topicName.toString(), requestId,
                        metadataAutoCreationEnabled);
            } else {
                request = Commands.newPartitionMetadataRequest(topicName.toString(), requestId);
            }
```

```java
    public static ByteBuf newPartitionMetadataRequest(String topic, long requestId) {
        BaseCommand cmd = localCmd(Type.PARTITIONED_METADATA);
        cmd.setPartitionMetadata()
                .setTopic(topic)
                .setRequestId(requestId);
        return serializeWithSize(cmd);
    }
```

Shawyeok:
Mail thread: https://lists.apache.org/thread/tpd2kgl4v3x8bkw0skv0nj38rzrk56t7

lhotari:
When the fix is made, it's worth adding a backwards compatibility integration test to `tests/integration` to ensure that such regressions don't happen in the future.

ZachChuba:
Just want to add a +1 that this reported PIP-344 backwards compatibility is an issue that my team and other users are facing too, currently blocking our pulsar-client upgrade.

PR #24557, opened 2025-07-25
[fix][broker][branch-3.3] Disable broken ExtensibleLoadManager tests and add closeInternalTopics in follower monitor

heesung-sohn:
…

### Motivation

Disabled the broken tests from ExtensibleLoadManager to unblock 3.3 release. We could fix the broken tests later(but I don't think it is worth to fix them, as they should work better in the mainline branch). I believe we can disable them to unblock the release.

### Modifications

- added this logic closeInternalTopics  in the monitor thread to clean internal topics from the follower in case they are loaded 

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


lhotari (approved):
LGTM, Thank you @heesung-sn!

PR #24560, opened 2025-07-25
[fix][broker] Fix maxTopicsPerNamespace might report a false failure

BewareMyPower:
### Motivation

The `createIfMissing` argument only represents the topic can be created automatically when it does not exist, but it might not trigger the topic creation in this `getTopic` call. Hence, we cannot simply pass 1 as the 2nd argument here: https://github.com/apache/pulsar/blob/d95cf3f34f34e8de2d3605374374a028c3e38853/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1790

Assuming `maxTopicsPerNamespace` is 10, then the following two cases will both fail.

Case 1:

```bash
> admin topics create-partitioned-topic my-topic -p 10
> client produce -m msg my-topic
```

Case 2:

```bash
> admin topics create-partitioned-topic dummy -p 9
> admin topics create my-topic
> admin topics unload my-topic
> client produce -m msg my-topic
```

### Modifications

Add an overload of `checkMaxTopicsPerNamespace`, which verifies if the topic exists first. Only call the other overload when the topic does not exist. In `createPersistentTopic0`, call this new overload. Improve `testMaxTopicsPerNamespace` to verify this change.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -1162,8 +1162,7 @@ public class BrokerService implements Closeable {
                     return FutureUtil.failedFuture(new NotAllowedException(
                             "Broker is unable to load persistent topic"));
                 }
-                return pulsar.getPulsarResources().getTopicResources().persistentTopicExists(topicName)
-                        .thenCompose(exists -> {
+                return checkNonPartitionedTopicExists(topicName).thenCompose(exists -> {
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM, good work @BewareMyPower

BewareMyPower:
I see 4.0.6 has not started the release, change the release label to 4.0.6

PR #13939, opened 2022-01-25
[Broker] waitingCursors potential  heap memory leak 

gaozhangmin:
### Motivation

`org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl#waitingCursors` has memory leak problem, reproduced steps: 
1. Create a consumer
2. Consumers  is closed due to some error, without consuming any message.
3. Day by day, it will cause broker heap memory leak problem.

Reproduce test: I don't know why getWaitingCursorsCount aways 1 less than for loop count.
```
@Test
    public void testSub() throws Exception {
        final String topicName = "persistent://prop/ns-abc/stuckSubscriptionTopic";
        for (int i=0;i<10;i++) {
            Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING).topic(topicName)
                    .subscriptionType(SubscriptionType.Failover).subscriptionName("test" + i).subscribe();
            consumer.close();
        }
        PersistentTopic topicRef = (PersistentTopic) pulsar.getBrokerService().getTopicReference(topicName).get();
        ManagedLedgerImpl ml = (ManagedLedgerImpl)(topicRef.ledger);
        assertEquals(9, ml.getWaitingCursorsCount());
    }
```

Detail: 

1.  When create consumer and  connection opens.
https://github.com/apache/pulsar/blob/a18f8a150b2d5c1af1725f6bf0dcb6582024b0d5/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L807-L809
2. Cursor was added to waitingCursors by `org.apache.bookkeeper.mledger.impl.ManagedCursorImpl#checkForNewEntries`
https://github.com/apache/pulsar/blob/a18f8a150b2d5c1af1725f6bf0dcb6582024b0d5/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L788-L806
3. But cursor were only removed when there were new messages.  if consumer close without consuming messages, The cursor was not removed permantly.
`org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl#notifyCursors`
https://github.com/apache/pulsar/blob/a18f8a150b2d5c1af1725f6bf0dcb6582024b0d5/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2219-L2228 

### Modifications

1、remove cursor when consumer closed.

### Documentation

Check the box below or label this PR directly (if you have committer privilege).

[...]

PR #24569, opened 2025-07-26
[fix][broker] Fix ManagedCursor state management race conditions and lifecycle issues

### LINKED ISSUE #24566: Flaky-test: ReplicatorTest.testCloseReplicatorStartProducer
### Search before reporting

### Example failure

https://github.com/apache/pulsar/actions/runs/16524077290/job/46733436629?pr=24564#step:11:1441

### Exception stacktrace

```text
Error:  Tests run: 216, Failures: 1, Errors: 0, Skipped: 211, Time elapsed: 71.584 s <<< FAILURE! - in org.apache.pulsar.broker.service.ReplicatorTest
  Error:  org.apache.pulsar.broker.service.ReplicatorTest.testCloseReplicatorStartProducer  Time elapsed: 6.594 s  <<< FAILURE!
  java.lang.AssertionError: expected [null] but found [ProducerBase{topic='persistent://pulsar/ns1/closeCursor-62f37ab9-9343-4bb8-b2b4-fbd8ee936786'}]
  	at org.testng.Assert.failNotSame(Assert.java:1573)
  	at org.testng.Assert.assertNull(Assert.java:1506)
  	at org.testng.Assert.assertNull(Assert.java:1494)
  	at org.apache.pulsar.broker.service.ReplicatorTest.testCloseReplicatorStartProducer(ReplicatorTest.java:1092)
```

lhotari:
Fixes #24566

### Motivation

The test `ReplicatorTest.testCloseReplicatorStartProducer` was consistently failing due to critical issues introduced in PR #24551. Investigation (shared in https://github.com/apache/pulsar/issues/24566#issuecomment-3119993989 by @3pacccccc ) revealed that the changes led to infinite recursive calls in `ManagedCursorImpl#cancelPendingReadRequest`, causing broker instability and test failures.

The root cause was improper state management for pending read operations, particularly when:
- Multiple threads attempt to cancel pending read requests concurrently
- Cursors are closed while read operations are pending
- OpReadEntry instances are recycled and reused incorrectly
- Cursor state transitions occur without proper synchronization
- Cursor deletion and closure operations race with active read operations

### Modifications

This PR addresses the state management issues through several key changes:

#### Core Synchronization Fixes

1. **Reverted problematic synchronization approach**: Removed the complex `pendingReadOpMutex` and `DelayCheckForNewEntriesTask` implementation added in #24551 that caused infinite recursion, returning to a more robust compare-and-swap based solution.

2. **Added operation ID tracking**: Introduced an `id` field to `OpReadEntry` instances to prevent recycled operations from being processed incorrectly. This ensures that only the correct operation instance is handled when checking for new entries.

3. **Improved atomic state transitions**: Enhanced the CAS (Compare-And-Swap) operations using `getAndUpdate()` with lambda functions that verify both the operation instance and its ID before making state changes:
   ```java
                OpReadEntry waitingReadOpItem = WAITING_READ_OP_UPDATER.getAndUpdate(this,
                        current -> {
                            if (current == op && current.id == opReadId) {
                                // update the value to null to cancel the waiting read op
                                return null;
                            } else {
                                // keep the current waiting read op value
                                return current;
                            }
                        });
                // If the waiting read op was the same as the one we are trying to cancel, it means that it was now
                // cleared from the waitingReadOp field and therefore "cancelled"
                if (waitingReadOpItem == op && waitingReadOpItem.id == opReadId) {
                    if (log.isDebugEnabled()) {
                        log.debug("[{}] [{}] Cancelled notification and scheduled read at {}", ledger.getName(),
                                name, op.readPosition);
                    }
                    PENDING_READ_OPS_UPDATER.incrementAndGet(this);
                    ledger.asyncReadEntries(op);
   ```

#### Enhanced State Management System

4. **Comprehensive state model enhancement**: Significantly expanded the cursor `State` enum with:
   - New states: `Deleting`, `Deleted`, `DeletingFailed` for better deletion lifecycle tracking
   - Added `closedState` boolean field and `isClosed()` method for more robust state checking
   - Added `isDeletingOrDeleted()` method for deletion state validation
   - Enhanced state constructor to properly categorize closed vs. open states

5. **Thread-safe state transition methods**: Implemented atomic state transition utilities:
   - `changeStateIfNotClosed()` - prevents state changes on closed cursors
   - `changeStateIfNotDeletingOrDeleted()` - prevents state changes on deleting/deleted cursors
   - Enhanced `trySetStateToClosing()` with atomic operations

#### Cursor Lifecycle Management

6. **Proper cursor closure handling**: Added comprehensive cursor closure logic:
   - `closeWaitingCursor()` - handles normal cursor closure scenarios
   - `cancelWaitingCursorsWhenDeactivated()` - handles cursor deactivation scenarios
   - `internalCloseWaitingCursor()` - unified internal implementation with exception supplier pattern
   - Added sentinel value (`WAITING_READ_OP_FOR_CLOSED_CURSOR`) to handle closed cursor scenarios

7. **Enhanced cursor deactivation**: Updated `setInactive()` to properly cancel waiting cursors when deactivated, preventing orphaned read operations.

8. **Robust deletion lifecycle**: Improved `asyncDeleteCursorLedger()` with:
   - Proper state transitions through deletion lifecycle
   - Enhanced retry logic with backoff for failed deletions
   - Better error handling and state recovery for deletion failures
   - Prevention of duplicate deletion attempts

#### Exception Handling Improvements

9. **New exception types**: Added `CursorDeactivatedWaitCallbackException` for better error propagation when cursors are deactivated without properly cancelling pending requests.

10. **Enhanced error propagation**: Improved error handling throughout the cursor lifecycle to ensure proper cleanup and user notification.

#### Managed Ledger Integration

11. **Enhanced cursor deletion in ManagedLedgerImpl**: Improved `asyncDeleteCursor()` with:
    - Proper state checking before deletion attempts
    - Separate handling for durable vs. non-durable cursors
    - Better error recovery and state management
    - Prevention of concurrent deletion attempts

12. **Enhanced waiting cursor registration**: Implemented thread-safe registration/deregistration of waiting cursors to prevent race conditions when adding or removing cursors from the waiting list, with improved debug logging.

#### Specialized Cursor Implementation Updates

13. **Non-durable cursor improvements**: Updated `NonDurableCursorImpl.asyncClose()` to properly close waiting cursors and set inactive state.

14. **Read-only cursor improvements**: Updated `ReadOnlyCursorImpl.asyncClose()` with the same proper cleanup logic.

#### Misc refactoring in ManagedCursorImpl

15. In the previous code, the `state` field was sometimes read directly and sometimes with `STATE_UPDATER.get`. Similarly, it was sometimes set directly and mostly with `STATE_UPDATER.set`. There's no need to use `STATE_UPDATER.get` and `STATE_UPDATER.set`. The volatile `state` field can be read and set directly without `STATE_UPDATER`. 
`STATE_UPDATER` is only needed for CAS operations on the `state` field.

### Key Technical Improvements

- **Eliminated infinite recursion**: The previous implementation could enter infinite recursive calls in `cancelPendingReadRequest()`, which is now resolved
- **Thread-safe operation lifecycle**: OpReadEntry instances are properly tracked through their lifecycle to prevent use-after-recycle scenarios
- **Comprehensive state management**: New state model provides better visibility and control over cursor lifecycle
- **Enhanced error handling**: Cursor closure and deletion events now properly propagate appropriate exceptions to pending operations
- **Performance optimization**: Reduced lock contention by using atomic operations instead of synchronized blocks where possible
- **Improved debugging**: Added comprehensive debug logging for cursor registration/deregistration operations
- **Robust deletion handling**: Deletion operations now properly handle failures, retries, and state recovery

### Verifying this change

This change is already covered by existing tests, specifically:
- `ReplicatorTest.testCloseReplicatorStartProducer` - verifies proper cleanup when replicator cursors are closed
- `FailoverSubscriptionTest.testWaitingCursorsCountAfterSwitchingActiveConsumers` - verifies that each cursor doesn't get added more than once into ManagedLedgerImpl's waiting cursors list
- Existing managed ledger tests that exercise concurrent read operations
- Integration tests that stress-test cursor state management under load
- Tests covering cursor deletion and state transitions

The modifications also include timeout adjustments for the test to account for the improved error handling paths.

### Impact Assessment

This change affects:
- **Threading model**: Improved thread safety in cursor state management with atomic state transitions
- **Error handling**: Better propagation of cursor closure and deactivation exceptions
- **Performance**: Reduced lock contention in high-throughput scenarios through atomic operations
- **Lifecycle management**: More robust cursor deletion and closure handling
- **State consistency**: Enhanced state model prevents invalid state transitions
- **Resource cleanup**: Better cleanup of waiting cursors and pending operations

The changes are backward compatible and do not affect public APIs or configuration defaults.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -827,22 +827,6 @@ public class ManagedCursorImpl implements ManagedCursor {
         });
     }
 
-    /**
-     * Change the state of the cursor if it is not already Deleting or Deleted.
-     * This is to prevent invalid state transitions when the cursor is already being deleted or has been deleted.
-     *
-     * @param newState The new state to set
-     * @return The previous state of the cursor
-     */
-    State changeStateIfNotDeletingOrDeleted(State newState) {
-        return STATE_UPDATER.getAndUpdate(this, current -> {
-            if (current.isDeletingOrDeleted()) {
-                return current;
-            }
-            return newState;
-        });
-    }
-
     void initialize(Position position, Map<String, Long> properties, Map<String, String> cursorProperties,
                     final VoidCallback callback) {
         recoveredCursor(position, properties, cursorProperties, null);
@@ -3638,9 +3622,16 @@ public class ManagedCursorImpl implements ManagedCursor {
     }
 
     private void asyncDeleteCursorLedger(int retry) {
-        State previousState = changeStateIfNotDeletingOrDeleted(State.Deleting);
-        if (previousState == State.Deleted) {
-            log.warn("[{}-{}] Cursor ledger is already deleting or deleted. state={}", ledger.getName(), name, state);
+        State beforeChangingState = STATE_UPDATER.getAndUpdate(this, current -> {
+            // don't change the state if it's already deleting or deleted
+            if (current.isDeletingOrDeleted()) {
+                return current;
+            }
+            return State.Deleting;
+        });
+        if (beforeChangingState.isDeletingOrDeleted()) {
+            log.warn("[{}-{}] Cursor ledger is already deleting or deleted. state={}", ledger.getName(), name,
+                    beforeChangingState);
             return;
         }
 
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -296,7 +296,8 @@ public class ManagedCursorImpl implements ManagedCursor {
         Closing(true), // The managed cursor is closing
         Closed(true), // The managed cursor has been closed
         Deleting(true), // The managed cursor is being deleted
-        Deleted(true);  // The managed cursor has been deleted
+        Deleted(true), // The managed cursor has been deleted
+        DeletingFailed(true); // The managed cursor deletion failed, state allows retrying deletion.
 
         // Indicate if the cursor is in a state that is considered closed
         private final boolean closedState;
@@ -312,7 +313,7 @@ public class ManagedCursorImpl implements ManagedCursor {
             return closedState;
         }
 
-        public boolean isDeletion() {
+        public boolean isDeletingOrDeleted() {
             return this == Deleting || this == Deleted;
         }
     }
@@ -821,7 +822,7 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     State changeStateIfNotDeletingOrDeleted(State newState) {
         return STATE_UPDATER.getAndUpdate(this, current -> {
-            if (current.isDeletion()) {
+            if (current.isDeletingOrDeleted()) {
                 return current;
             }
             return newState;
@@ -3633,9 +3634,9 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     private void asyncDeleteCursorLedger(int retry) {
         State previousState = changeStateIfNotDeletingOrDeleted(State.Deleting);
-        if (previousState == State.Deleted) {
-            log.warn("[{}-{}] Cursor ledger {} is already deleted.", ledger.getName(), name,
-                    cursorLedger.getId());
+        if (previousState.isDeletingOrDeleted()) {
+            log.warn("[{}-{}] Cursor ledger {} is already deleting or deleted. state=", ledger.getName(), name,
+                    cursorLedger.getId(), state);
             return;
         }
         closeWaitingCursor();
@@ -3658,6 +3659,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                 log.warn("[{}][{}] Failed to delete ledger {}: {}", ledger.getName(), name, cursorLedger.getId(),
                         BKException.getMessage(rc));
                 if (!isNoSuchLedgerExistsException(rc)) {
+                    state = State.DeletingFailed;
                     ledger.getScheduledExecutor().schedule(() -> asyncDeleteCursorLedger(retry - 1),
                             DEFAULT_LEDGER_DELETE_BACKOFF_TIME_SEC, TimeUnit.SECONDS);
                 } else {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -227,7 +227,9 @@ public class ManagedCursorImpl implements ManagedCursor {
     // active state cache in ManagedCursor. It should be in sync with the state in activeCursors in ManagedLedger.
     private volatile boolean isActive = false;
 
+    // This is a lock used to update the registration state of the cursor in the managed ledger.
     private Object waitingRegistrationLock = new Object();
+    // This is used to track if the cursor is waiting for registration in the managed ledger.
     boolean waitingRegistered = false;
 
     class MarkDeleteEntry {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -292,12 +292,16 @@ public class ManagedCursorImpl implements ManagedCursor {
         Open, // Metadata ledger is ready
         SwitchingLedger, // The metadata ledger is being switched
         Closing, // The managed cursor is closing
-        Closed // The managed cursor has been closed
+        Closed; // The managed cursor has been closed
+
+        public boolean isClosed() {
+            return this == Closing || this == Closed;
+        }
     }
 
     protected static final AtomicReferenceFieldUpdater<ManagedCursorImpl, State> STATE_UPDATER =
         AtomicReferenceFieldUpdater.newUpdater(ManagedCursorImpl.class, State.class, "state");
-    protected volatile State state = null;
+    protected volatile State state = State.Uninitialized;
 
     protected final ManagedCursorMXBean mbean;
 
@@ -326,7 +330,6 @@ public class ManagedCursorImpl implements ManagedCursor {
             this.batchDeletedIndexes = null;
         }
         this.digestType = BookKeeper.DigestType.fromApiDigestType(getConfig().getDigestType());
-        STATE_UPDATER.set(this, State.Uninitialized);
         PENDING_MARK_DELETED_SUBMITTED_COUNT_UPDATER.set(this, 0);
         PENDING_READ_OPS_UPDATER.set(this, 0);
         RESET_CURSOR_IN_PROGRESS_UPDATER.set(this, FALSE);
@@ -786,7 +789,16 @@ public class ManagedCursorImpl implements ManagedCursor {
         // assign cursor-ledger so, it can be deleted when new ledger will be switched
         this.cursorLedger = recoveredFromCursorLedger;
         this.isCursorLedgerReadOnly = true;
-        STATE_UPDATER.set(this, State.NoLedger);
+        changeStateIfNotClosed(State.NoLedger);
+    }
+
+    private State changeStateIfNotClosed(State newState) {
+        return STATE_UPDATER.getAndUpdate(this, current -> {
+            if (current.isClosed()) {
+                return current;
+            }
+            return newState;
+        });
     }
 
     void initialize(Position position, Map<String, Long> properties, Map<String, String> cursorProperties,
@@ -800,7 +812,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                 new MetaStoreCallback<>() {
                     @Override
                     public void operationComplete(Void result, Stat stat) {
-                        STATE_UPDATER.set(ManagedCursorImpl.this, State.NoLedger);
+                        changeStateIfNotClosed(State.NoLedger);
                         callback.operationComplete();
                     }
                     @Override
@@ -1134,7 +1146,7 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     @Override
     public boolean isClosed() {
-        return state == State.Closed || state == State.Closing;
+        return state.isClosed();
     }
 
     @Override
@@ -2174,7 +2186,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         // We cannot write to the ledger during the switch, need to wait until the new metadata ledger is available
         synchronized (pendingMarkDeleteOps) {
             // The state might have changed while we were waiting on the queue mutex
-            switch (STATE_UPDATER.get(this)) {
+            switch (state) {
             case Closed:
                 callback.markDeleteFailed(new ManagedLedgerException
                         .CursorAlreadyClosedException("Cursor was already closed"), ctx);
@@ -2305,7 +2317,7 @@ public class ManagedCursorImpl implements ManagedCursor {
             }
         };
 
-        if (State.NoLedger.equals(STATE_UPDATER.get(this))) {
+        if (state == State.NoLedger) {
             if (ledger.isNoMessagesAfterPos(mdEntry.newPosition)) {
                 log.error("[{}][{}] Metadata ledger creation failed, try to persist the position in the metadata"
                         + " store.", ledger.getName(), name);
@@ -2892,7 +2904,11 @@ public class ManagedCursorImpl implements ManagedCursor {
 
                     @Override
                     public void closeComplete(Object ctx) {
-                        STATE_UPDATER.set(ManagedCursorImpl.this, State.Closed);
+                        if (!STATE_UPDATER.compareAndSet(ManagedCursorImpl.this, State.Closing, State.Closed)) {
+                            log.warn("[{}] [{}] State was modified from closing to {} while closing", ledger.getName(),
+                                    name, state);
+                            state = State.Closed;
+                        }
                         callback.closeComplete(ctx);
                     }
 
@@ -3017,8 +3033,8 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     void startCreatingNewMetadataLedger() {
         // Change the state so that new mark-delete ops will be queued and not immediately submitted
-        State oldState = STATE_UPDATER.getAndSet(this, State.SwitchingLedger);
-        if (oldState == State.SwitchingLedger) {
+        State oldState = changeStateIfNotClosed(State.SwitchingLedger);
+        if (oldState == State.SwitchingLedger || oldState.isClosed()) {
             // Ignore double request
             return;
         }
@@ -3038,7 +3054,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                     flushPendingMarkDeletes();
 
                     // Resume normal mark-delete operations
-                    STATE_UPDATER.set(ManagedCursorImpl.this, State.Open);
+                    changeStateIfNotClosed(State.Open);
                 }
             }
 
@@ -3047,7 +3063,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                 log.error("[{}][{}] Metadata ledger creation failed {}", ledger.getName(), name, exception);
                 synchronized (pendingMarkDeleteOps) {
                     // At this point we don't have a ledger ready
-                    STATE_UPDATER.set(ManagedCursorImpl.this, State.NoLedger);
+                    changeStateIfNotClosed(State.NoLedger);
                     // There are two case may cause switch ledger fails.
                     // 1. No enough BKs; BKs are in read-only mode...
                     // 2. Write ZK fails.
@@ -3071,19 +3087,8 @@ public class ManagedCursorImpl implements ManagedCursor {
      * @return false if the {@link #state} already is {@link State#Closing} or {@link State#Closed}.
      */
     private boolean trySetStateToClosing() {
-        State previousState = STATE_UPDATER.getAndUpdate(this, current -> {
-            if (current != State.Closing && current != State.Closed) {
-                return State.Closing;
-            }
-            return current;
-        });
-        if (previousState == State.Closing || previousState == State.Closed) {
-            // Already closing or closed
-            return false;
-        } else {
-            // Successfully set to closing
-            return true;
-        }
+        State previousState = changeStateIfNotClosed(State.Closing);
+        return !previousState.isClosed();
     }
 
     private void flushPendingMarkDeletes() {
@@ -3339,7 +3344,7 @@ public class ManagedCursorImpl implements ManagedCursor {
                 mbean.addWriteCursorLedgerSize(data.length);
                 callback.operationComplete();
             } else {
-                if (state == State.Closed) {
+                if (state.isClosed()) {
                     // After closed the cursor, the in-progress persistence task will get a
                     // BKException.Code.LedgerClosedException.
                     callback.operationFailed(new CursorAlreadyClosedException(String.format("%s %s skipped this"
@@ -3360,8 +3365,7 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     public boolean periodicRollover() {
         LedgerHandle lh = cursorLedger;
-        if (State.Open.equals(STATE_UPDATER.get(this))
-                && lh != null && lh.getLength() > 0) {
+        if (state == State.Open && lh != null && lh.getLength() > 0) {
             boolean triggered = rolloverLedgerIfNeeded(lh);
             if (triggered) {
                 log.info("[{}] Periodic rollover triggered for cursor {} (length={} bytes)",
@@ -3419,7 +3423,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         if (ledger.getFactory().isMetadataServiceAvailable()
                 && (lh.getLastAddConfirmed() >= getConfig().getMetadataMaxEntriesPerLedger()
                 || lastLedgerSwitchTimestamp < (now - getConfig().getLedgerRolloverTimeout() * 1000))
-                && (STATE_UPDATER.get(this) != State.Closed && STATE_UPDATER.get(this) != State.Closing)) {
+                && !state.isClosed()) {
             // It's safe to modify the timestamp since this method will be only called from a callback, implying that
             // calls will be serialized on one single thread
             lastLedgerSwitchTimestamp = now;
@@ -3533,7 +3537,6 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     void decrementPendingMarkDeleteCount() {
         if (PENDING_MARK_DELETED_SUBMITTED_COUNT_UPDATER.decrementAndGet(this) == 0) {
-            final State state = STATE_UPDATER.get(this);
             if (state == State.SwitchingLedger) {
                 // A metadata ledger switch was pending and now we can do it since we don't have any more
                 // outstanding mark-delete requests
@@ -3545,7 +3548,7 @@ public class ManagedCursorImpl implements ManagedCursor {
     void readOperationCompleted() {
         if (PENDING_READ_OPS_UPDATER.decrementAndGet(this) == 0) {
             synchronized (pendingMarkDeleteOps) {
-                if (STATE_UPDATER.get(this) == State.Open) {
+                if (state == State.Open) {
                     // Flush the pending writes only if the state is open.
                     flushPendingMarkDeletes();
                 } else if (PENDING_MARK_DELETED_SUBMITTED_COUNT_UPDATER.get(this) != 0) {
@@ -3592,6 +3595,7 @@ public class ManagedCursorImpl implements ManagedCursor {
     }
 
     private void asyncDeleteCursorLedger(int retry) {
+        // TODO: State transition would need to be handled here
         STATE_UPDATER.set(this, State.Closed);
 
         if (cursorLedger == null || retry <= 0) {
@@ -3775,7 +3779,7 @@ public class ManagedCursorImpl implements ManagedCursor {
     }
 
     public String getState() {
-        return STATE_UPDATER.get(this).toString();
+        return state.toString();
     }
 
     @Override
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -2886,12 +2886,7 @@ public class ManagedCursorImpl implements ManagedCursor {
             callback.closeComplete(ctx);
             return;
         }
-        OpReadEntry opReadEntry = WAITING_READ_OP_UPDATER.getAndSet(ManagedCursorImpl.this,
-                OpReadEntry.WAITING_READ_OP_FOR_CLOSED_CURSOR);
-        if (opReadEntry != null) {
-            opReadEntry.readEntriesFailed(new ManagedLedgerException.CursorAlreadyClosedException(
-                    "Cursor is closing"), opReadEntry.ctx);
-        }
+        closeWaitingCursor();
         persistPositionWhenClosing(lastMarkDeleteEntry.newPosition, lastMarkDeleteEntry.properties,
                 new AsyncCallbacks.CloseCallback(){
 
@@ -2908,6 +2903,20 @@ public class ManagedCursorImpl implements ManagedCursor {
                         callback.closeFailed(exception, ctx);
                     }
                 }, ctx);
+    }
+
+    protected void closeWaitingCursor() {
+        synchronized (waitingRegistrationLock) {
+            if (waitingRegistered) {
+                ledger.removeWaitingCursor(this);
+            }
+        }
+        OpReadEntry opReadEntry = WAITING_READ_OP_UPDATER.getAndSet(ManagedCursorImpl.this,
+                OpReadEntry.WAITING_READ_OP_FOR_CLOSED_CURSOR);
+        if (opReadEntry != null) {
+            opReadEntry.readEntriesFailed(new CursorAlreadyClosedException(
+                    "Cursor is closing"), opReadEntry.ctx);
+        }
     }
 
     /**
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -1205,7 +1205,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         if (op != null) {
             op.recycle();
         }
-        return op != null;
+        return op != null && op != OpReadEntry.WAITING_READ_OP_FOR_CLOSED_CURSOR;
     }
 
     public boolean hasPendingReadRequest() {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3640,7 +3640,7 @@ public class ManagedCursorImpl implements ManagedCursor {
     private void asyncDeleteCursorLedger(int retry) {
         State previousState = changeStateIfNotDeletingOrDeleted(State.Deleting);
         if (previousState == State.Deleted) {
-            log.warn("[{}-{}] Cursor ledger is already deleting or deleted. state=", ledger.getName(), name, state);
+            log.warn("[{}-{}] Cursor ledger is already deleting or deleted. state={}", ledger.getName(), name, state);
             return;
         }
 
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3634,18 +3634,21 @@ public class ManagedCursorImpl implements ManagedCursor {
 
     private void asyncDeleteCursorLedger(int retry) {
         State previousState = changeStateIfNotDeletingOrDeleted(State.Deleting);
-        if (previousState.isDeletingOrDeleted()) {
-            log.warn("[{}-{}] Cursor ledger {} is already deleting or deleted. state=", ledger.getName(), name,
-                    cursorLedger.getId(), state);
+        if (previousState == State.Deleted) {
+            log.warn("[{}-{}] Cursor ledger is already deleting or deleted. state=", ledger.getName(), name, state);
             return;
         }
+
         closeWaitingCursor();
 
-        if (cursorLedger == null || retry <= 0) {
-            if (cursorLedger != null) {
-                log.warn("[{}-{}] Failed to delete ledger after retries {}", ledger.getName(), name,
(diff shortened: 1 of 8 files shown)


lhotari:
I learned through some failing tests that there are some really hairy details of the `havePendingRead` state in dispatcher and the related state in ManagedCursorImpl (`waitingReadOp`) and ManagedLedgerImpl (`waitingCursors`). It's like there's a circular runtime dependency of state.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2958,12 +2940,18 @@ public void asyncClose(final AsyncCallbacks.CloseCallback callback, final Object
...
         persistPositionWhenClosing(lastMarkDeleteEntry.newPosition, lastMarkDeleteEntry.properties,
                 new AsyncCallbacks.CloseCallback(){
 
                     @Override
                     public void closeComplete(Object ctx) {
-                        STATE_UPDATER.set(ManagedCursorImpl.this, State.Closed);
+                        if (!STATE_UPDATER.compareAndSet(ManagedCursorImpl.this, State.Closing, State.Closed)) {
+                            log.warn("[{}] [{}] State was modified from closing to {} while closing", ledger.getName(),
+                                    name, state);
+                            state = State.Closed;
```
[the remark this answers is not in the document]
lhotari: It's safe here

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3629,27 +3638,40 @@ void asyncDeleteCursorLedger() {
...
                 log.info("[{}][{}] Deleted cursor ledger {}", ledger.getName(), name, cursorLedger.getId());
             } else {
                 log.warn("[{}][{}] Failed to delete ledger {}: {}", ledger.getName(), name, cursorLedger.getId(),
                         BKException.getMessage(rc));
                 if (!isNoSuchLedgerExistsException(rc)) {
+                    state = State.DeletingFailed;
                     ledger.getScheduledExecutor().schedule(() -> asyncDeleteCursorLedger(retry - 1),
                             DEFAULT_LEDGER_DELETE_BACKOFF_TIME_SEC, TimeUnit.SECONDS);
+                } else {
+                    state = State.Deleted;
```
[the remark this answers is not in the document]
lhotari: sure, valid point, but STATE_UPDATE.get and get aren't providing any value over direct referencing and assignment of a volatile field.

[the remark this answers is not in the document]
lhotari: considered

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3629,27 +3638,40 @@ void asyncDeleteCursorLedger() {
...
+                    cursorLedger.getId());
+            state = State.DeletingFailed;
             return;
         }
 
         ledger.mbean.startCursorLedgerDeleteOp();
         bookkeeper.asyncDeleteLedger(cursorLedger.getId(), (rc, ctx) -> {
             ledger.mbean.endCursorLedgerDeleteOp();
             if (rc == BKException.Code.OK) {
+                state = State.Deleted;
```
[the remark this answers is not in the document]
lhotari: considered

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3629,27 +3638,40 @@ void asyncDeleteCursorLedger() {
...
+        closeWaitingCursor();
+
+        if (cursorLedger == null) {
+            return;
+        }
+
+        if (retry <= 0) {
+            log.warn("[{}-{}] Failed to delete ledger after retries {}", ledger.getName(), name,
+                    cursorLedger.getId());
+            state = State.DeletingFailed;
```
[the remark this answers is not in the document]
lhotari: considered

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2246,7 +2228,7 @@ protected void internalAsyncMarkDelete(final Position newPosition, Map<String, L
         // We cannot write to the ledger during the switch, need to wait until the new metadata ledger is available
         synchronized (pendingMarkDeleteOps) {
             // The state might have changed while we were waiting on the queue mutex
-            switch (STATE_UPDATER.get(this)) {
+            switch (state) {
```
3pacccccc, line 2215: why changing this?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1040,87 +1089,36 @@ public void asyncReadEntriesWithSkipOrWait(int maxEntries, long maxSizeBytes, Re
...
-
-                // Check again for new entries after the configured time, then if still no entries are available
-                // register to be notified.
-                if (getConfig().getNewEntriesCheckDelayInMillis() > 0) {
-                    delayCheckForNewEntriesTask = new DelayCheckForNewEntriesTask(op, callback, ctx);
-                } else {
-                    // If there's no delay, check directly from the same thread
-                    checkForNewEntries(op, callback, ctx);
-                }
+            int opReadId = op.id;
```
3pacccccc, line 1076: could we check like this?
```java
            OpReadEntry op = OpReadEntry.EMPTY_PLACE_HOLDER;
            if (!WAITING_READ_OP_UPDATER.compareAndSet(this, null, op)) {
                op.recycle();
                callback.readEntriesFailed(new ManagedLedgerException.ConcurrentWaitCallbackException(), ctx);
                return;
            }
            op = OpReadEntry.create(this, readPosition, numberOfEntriesToRead, callback,
                    ctx, maxPosition, skipCondition);
            int opReadId = op.id;
            WAITING_READ_OP_UPDATER.set(this, op);
```
create a new EMPTY_PLACE_HOLDER just like ```OpReadEntry#WAITING_READ_OP_FOR_CLOSED_CURSOR``` which can reduce a lot of useless OpReadEntry create and recycle while thread racing

3pacccccc:
LGTM

3pacccccc (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: Not really, since the point of the CAS operation is to do it atomically. OpReadEntries are recycled, so there's no useless creation of OpReadEntries. It's only about assigning fields to the instance that is from the recycling pool. If there's a race, it would be recycled immediately. The role of `WAITING_READ_OP_FOR_CLOSED_CURSOR` is different. It's a sentinel value that prevents any races to change the value when the cursor has already been closed.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: `STATE_UPDATER.get` and `STATE_UPDATER.set` are not necessary. The volatile field can be read and set directly without `STATE_UPDATER`. The intention of the change is to use `STATE_UPDATER` only for CAS changes of the `state` field.

poorbarcode:
@lhotari 

I pushed a PR to fix the root cause 
- https://github.com/apache/pulsar/pull/24573

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
3pacccccc: I mean , We can optimize the code by avoiding unnecessary OpReadEntry.create() and OpReadEntry.recycle() operations when WAITING_READ_OP_UPDATER is not null. This improvement offers two key benefits:
- Resource Efficiency - Eliminates wasteful object creation and recycling when the operation would immediately fail
- ID Management - Since ```OpReadEntry.id``` auto-increments in this case, this may prevents ```OpReadEntry.id``` grow
rapidly and only few ```OpReadEntry.id``` are actually used for valid operations

lhotari:
> > Multiple threads attempt to cancel pending read requests concurrently
> > Cursors are closed while read operations are pending
> > OpReadEntry instances are recycled and reused incorrectly
> 
> Could you describe the detauls of how the `OpReadEntry` will be reused

@poorbarcode  Since OpReadEntry instances are recycled, they are reused. It's not sufficient do use `compareAndSet`. There's a need to have some unique identifier on the instance to do the CAS operation comparison. It's possible to find code paths where this could happen and new problems could be introduced unless we have a proper solution to for CAS operations.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
```diff
@@ -1046,7 +1045,7 @@ public void testResumptionAfterBacklogRelaxed() throws Exception {
      *
      * @throws Exception
      */
-    @Test(timeOut = 15000)
+    @Test(timeOut = 30000)
     public void testCloseReplicatorStartProducer() throws Exception {
```
poorbarcode, line 1049: Start a channel to discuss the following topic

@poorbarcode left a comment

> Multiple threads attempt to cancel pending read requests concurrently
> Cursors are closed while read operations are pending
> OpReadEntry instances are recycled and reused incorrectly

Could you describe the details of how the `OpReadEntry` will be reused

---

@lhotari  left a comment

Since OpReadEntry instances are recycled, they are reused. It's not sufficient do use compareAndSet. There's a need to have some unique identifier on the instance to do the CAS operation comparison. It's possible to find code paths where this could happen and new problems could be introduced unless we have a proper solution to for CAS operations.

poorbarcode: Hi @lhotari 

> Multiple threads attempt to cancel pending read requests concurrently
> Cursors are closed while read operations are pending
> OpReadEntry instances are recycled and reused incorrectly

`cursor.cancelPendingReadRequest` will compare and set `WAITING_READ_OP_UPDATER` to `null`, then recycle ` OpReadEntry`.

Since `WAITING_READ_OP_UPDATER` can only be compared and set to `null` at most once, how can it be incorrectly reused?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1149,17 +1147,33 @@ private void checkForNewEntries(OpReadEntry op, ReadEntriesCallback callback, Ob
...
+                        });
+                // If the waiting read op was the same as the one we are trying to cancel, it means that it was now
+                // cleared from the waitingReadOp field and therefore "cancelled"
+                if (waitingReadOpItem == op && waitingReadOpItem.id == opReadId) {
                     if (log.isDebugEnabled()) {
                         log.debug("[{}] [{}] Cancelled notification and scheduled read at {}", ledger.getName(),
-                            name, op.readPosition);
+                                name, op.readPosition);
                     }
                     PENDING_READ_OPS_UPDATER.incrementAndGet(this);
```
poorbarcode, line 1154: We need to call `removeWaitingCursor` here

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
lhotari: > Since `WAITING_READ_OP_UPDATER` can only be compared and set to `null` at most once, how can it be incorrectly reused?

`cursor.cancelPendingReadRequest` or might be called in the main thread before the scheduled `checkForNewEntries` completes. `cursor.notifyEntriesAvailable` might also be called to to complete the `waitingReadOp` before `checkForNewEntries` completes. The main thread could then call `asyncReadEntriesWithSkipOrWait` and the get the same `OpReadEntry` instance (this is possible, but rare) from the recycler (since Netty's recycler is thread local). 
When the `checkForNewEntries` completes, the `compareAndSet` check would succeed since the same instance continues to be in the `waitingReadOp` field. I guess this would result in duplicate read entries in most cases. 
This instance mix-up could happen across different `managedCursorImpl` instances. Without instance id checks for handling `waitingReadOp` CAS operations, it leads to non-deterministic results as you can see from the few examples above. That's why there's a need to have a `id` for `OpReadEntry`.

poorbarcode: The issue you described has been fixed by https://github.com/apache/pulsar/pull/24551, right?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1149,17 +1147,33 @@ private void checkForNewEntries(OpReadEntry op, ReadEntriesCallback callback, Ob
...
                     ledger.asyncReadEntries(op);
                 } else {
-                    log.info("[{}] [{}] notification that new entries added was already be called, skipped the current"
```
poorbarcode, line 1161: The method `cancelPendingRead` can not cancel this task, right? which causes a orphan reading more entries

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: That's true that it could be done, however, it doesn't seem to be necessary in this solution. Calling `removeWaitingCursor` is potentially an expensive operation since `waitingCursors` is a `ConcurrentLinkedQueue` in `ManagedLedgerImpl` and when there would be a large amount of waiting cursors. That's why I'd like to avoid calling `removeWaitingCursor` here.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: In this case, `opReadEntry` in `waitingReadOp` was null or some other instance so it was either cancelled or notified before `checkNewEntries` was called.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
lhotari: @poorbarcode Not in general. When using recycled objects, it's error prone to use CAS (`compareAndSet`) method since there might be cases where the object is recycled in between and it's a different usage of the same object instance. It's possible that your changes in PR 24551 prevent that from happening by using synchronization. Since ManagedCursorImpl operations are on the performance hot path, we should avoid synchronization as much as possible and continue to use the CAS based solution with proper identity check by using `getAndUpdate` where the existing value could be checked before making the update.

lhotari:
> > Adding `Runnable` as the parameter of `addWaitingCursorRequested` and `removeWaitingCursor` makes code a bit hard to read. When I read the code, I was wondering how many places could it be called and found only one place. IMO, it's a over-abstraction.

@BewareMyPower Yes, you are right about that it's hard to read. The reason why I didn't directly do what you are suggesting is that it's a bad idea to mutate the internal state of another instance. That breaks object encapsulation principles.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
```diff
@@ -34,7 +35,14 @@
 import org.slf4j.LoggerFactory;
 
 class OpReadEntry implements ReadEntriesCallback {
-
+    static final OpReadEntry WAITING_READ_OP_FOR_CLOSED_CURSOR = new OpReadEntry();
```
BewareMyPower, line 38: It can be a private static final field of `ManagedCursorImpl`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -4012,4 +4034,55 @@ public ManagedLedgerInternalStats.CursorStats getCursorStats() {
...
         return cs;
     }
+
+    /**
+     * Called by ManagedLedgerImpl to execute the Runnable inside the lock to remove the cursor from it's
+     * waiting cursors list.
+     * The cursor state is set to unregistered, and it can be registered again for waiting in ManagedLedgerImpl.
+     */
+    void removeWaitingCursorRequested(Runnable removeWaitingCursorRunnable) {
+        synchronized (waitingRegistrationLock) {
```
poorbarcode: This method will be called on the publishing messages path, which is a really hot path, right?

lhotari: to optimize, we could remove `ledger.removeWaitingCursor(this)` in `ManagedCursorImpl` completely for the read cycle since the waiting cursor will get removed when the entries are notified or when ledger.removeWaitingCursor is called from other parts of the code (`PersistentSubscription`). As mentioned in https://github.com/apache/pulsar/pull/24569#discussion_r2235714944, the removal from the `ConcurrentLinkedQueue` is potentially costly since it has to iterate through the complete queue to remove a cursor. That's why it would be better to avoid that operation.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1174,61 +1188,29 @@ private void checkForNewEntries(OpReadEntry op, ReadEntriesCallback callback, Ob
...
-            // Case 5: delayCheckForNewEntriesTask is running, but not done, which only occurs at a corner case. It
-            // only happens when the task is starting. Calling "cancelPendingReadRequest" here will release the lock
-            // "pendingReadOpMutex" and let the task go ahead.
-            return cancelPendingReadRequest();
+            return null;
+        });
+        if (op != null) {
+            op.recycle();
         }
+        return op != null;
```
BewareMyPower: It also returns true even when `op` is `WAITING_READ_OP_FOR_CLOSED_CURSOR`. However, it means `closeWaitingCursor` has been called that the cursor has been removed from `waitingCursors`. Returning true here does not make sense.

poorbarcode (approved):
After a full review, I think this PR has no bugs, and it has solved the deadlock that I explained [here](https://github.com/apache/pulsar/pull/24573#issue-3268713658)

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -791,7 +808,39 @@ private void recoveredCursor(Position position, Map<String, Long> properties,
...
+    }
+
+    /**
+     * Change the state of the cursor if it is not already Deleting or Deleted.
+     * This is to prevent invalid state transitions when the cursor is already being deleted or has been deleted.
+     *
+     * @param newState The new state to set
+     * @return The previous state of the cursor
+     */
+    State changeStateIfNotDeletingOrDeleted(State newState) {
```
BewareMyPower: `newState` is always `Deleting`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Okay, `removeWaitingCursor` is expensive because it uses the `synchronized` lock

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Okay, it will not trigger a repeated reading, just left an orphan cursor in `ml.waitingCursors`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3629,27 +3638,40 @@ void asyncDeleteCursorLedger() {
     }
 
     private void asyncDeleteCursorLedger(int retry) {
-        STATE_UPDATER.set(this, State.Closed);
+        State previousState = changeStateIfNotDeletingOrDeleted(State.Deleting);
+        if (previousState == State.Deleted) {
+            log.warn("[{}-{}] Cursor ledger is already deleting or deleted. state={}", ledger.getName(), name, state);
```
BewareMyPower: The log seems wrong. In `changeStateIfNotDeletingOrDeleted`, there are only three cases:
| previousState | state |
| :- | :- |
| Deleting | Deleting |
| Deleted | Deleted |
| Other | Deleting |

So `state` should always be `Deleted` here. The log `is already deleting or deleted` should be `is already deleted`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java
BewareMyPower: Sorry, ignore this comment

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -354,6 +354,13 @@ public synchronized void removeConsumer(Consumer consumer, boolean isResetCursor
 
         if (dispatcher != null && dispatcher.getConsumers().isEmpty()) {
             deactivateCursor();
+            // Remove the cursor from the waiting cursors list.
+            // For durable cursors, we should *not* cancel the pending read. This is because internally,
+            // in the dispatcher implementations, there is a "havePendingRead" flag that is not reset. If the pending
+            // read is cancelled, the dispatcher will not continue reading from the managed ledger when a new
+            // consumer is added to the dispatcher since based on the "havePendingRead" state, it will continue to
+            // expect that a read is pending and will not submit a new read.
+            topic.getManagedLedger().removeWaitingCursor(cursor);
```
BewareMyPower, line 366: It's still hard to understand the motivation here even with the comments. The comments also make me confused.

> For durable cursors

The code here does not check if the cursor is durable.

> cancel the pending read

But it does not call `cursor.cancelPendingReadRequest()`.

Is it possible to add a test?

BewareMyPower (changes requested):
I've gone through the whole PR, overall LGTM except for some coding style issues and some questions I've left. Please address these comments before merging.

P.S. The state machine is hard to read. It would be helpful to have a graph to show all possible state changes, but it's not required.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: Correct. That would get removed when the `ml.notifyCursors` method has been called.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: I'll change to `op != null && op != 
OpReadEntry.WAITING_READ_OP_FOR_CLOSED_CURSOR`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: There's a possibility that deletion fails. In that case, the current state would be `DeletingFailed` and that that would allow this check to proceed so that deletion could be retried.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
lhotari: > The code here does not check if the cursor is durable.

I'll improve the comment. For non-durable cursors, it would be possible to call `cancelPendingReadRequest`. For durable cursors, that shouldn't be called since it would cause the state of the havePendingRead flag in dispatcher to get out of sync.
There's existing tests that fail if `cancelPendingReadRequest` would be called.

The `removeWaitingCursor` calls were removed in #24551 changes. Before that, the removeWaitingCursor was called in different location for the non durable and durable cursors. Since the call should be made for both, it's better to have it before the conditional logic.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: True. I'll inline this method since it's only used in one location. It just adds confusion to have this method around.

lhotari: Actually there's 2 locations where it's used. In ManagedLedgerImpl.asyncDeleteCursor it's also used. The "Deleting" state cover all stages of deletion. If any stage fails, the state will be set to DeletingFailed, so that a retry could happen.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: actually, the reason why it's relatively expensive is the scanning of the complete ConcurrentLinkedQueue in ManagedLedgerImpl to remove the instance.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: @BewareMyPower Thanks, I've improved this in the most recent changes.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: There's no calls to `ledger.removeWaitingCursor(this)` in the hot path, so this should be fine.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
lhotari: Comment improved.

lhotari:
> I've gone through the whole PR, overall LGTM except for some coding style issues and some questions I've left. Please address these comments before merging.
> 
> P.S. The state machine is hard to read. It would be helpful to have a graph to show all possible state changes, but it's not required.

Thank you @BewareMyPower. I've addressed your review comments. PTAL.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -354,6 +354,16 @@ public synchronized void removeConsumer(Consumer consumer, boolean isResetCursor
...
+            // Remove the cursor from the waiting cursors list.
+            // For durable cursors, we should *not* cancel the pending read with cursor.cancelPendingReadRequest.
+            // This is because internally, in the dispatcher implementations, there is a "havePendingRead" flag
+            // that is not reset. If the pending read is cancelled, the dispatcher will not continue reading from
+            // the managed ledger when a new consumer is added to the dispatcher since based on the "havePendingRead"
+            // state, it will continue to expect that a read is pending and will not submit a new read.
+            // For non-durable cursors, there's no difference since the cursor is not expected to be used again.
+
+            // remove waiting cursor from the managed ledger, this applies to both durable and non-durable cursors.
+            topic.getManagedLedger().removeWaitingCursor(cursor);
```
codelipenghui, line 366: If we look back at the first issue that was fixed by  https://github.com/apache/pulsar/pull/24551,

```
(Highlight) The method scheduleReadOnActiveConsumer resets the variable cursor.waitingReadOp, but it forgets to remove the cursor from managedledger.waitingCursors
```

https://github.com/apache/pulsar/pull/24551/files#diff-1d721da3aaff7689077d33bf3e55fd8d6a5c9184549857a6120d8050c5d67798L385-L386

Here, try to add another one.

codelipenghui: Can we optimize the method of the ManagedCursor to remove from the waiting cursors if the cursor has been registered to the managed ledger? So that we can avoid call `topic.getManagedLedger().removeWaitingCursor(cursor);` here.

```
    public void setInactive() {
        if (isActive) {
            ledger.deactivateCursor(this);
            isActive = false;
        }
    }
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -235,6 +227,11 @@ public class ManagedCursorImpl implements ManagedCursor {
     // active state cache in ManagedCursor. It should be in sync with the state in activeCursors in ManagedLedger.
     private volatile boolean isActive = false;
 
+    // This is a lock used to update the registration state of the cursor in the managed ledger.
+    private final Object waitingRegistrationLock = new Object();
+    // This is used to track if the cursor is waiting for registration in the managed ledger.
+    boolean waitingRegistered = false;
```
codelipenghui:
```suggestion
    boolean registeredToWaitingCursors = false;
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -235,6 +227,11 @@ public class ManagedCursorImpl implements ManagedCursor {
     // active state cache in ManagedCursor. It should be in sync with the state in activeCursors in ManagedLedger.
     private volatile boolean isActive = false;
 
+    // This is a lock used to update the registration state of the cursor in the managed ledger.
+    private final Object waitingRegistrationLock = new Object();
```
codelipenghui:
```suggestion
    private final Object registerToWaitingCursorsLock = new Object();
```

codelipenghui (dismissed):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3629,32 +3622,62 @@ void asyncDeleteCursorLedger() {
     }
 
     private void asyncDeleteCursorLedger(int retry) {
-        STATE_UPDATER.set(this, State.Closed);
+        State beforeChangingState = changeStateToDeletingIfNotDeleted();
+        if (beforeChangingState == State.Deleted) {
+            log.warn("[{}-{}] Cursor ledger is already deleted. state={}", ledger.getName(), name,
+                    beforeChangingState);
```
BewareMyPower, line 3628: Logging the state seems meaningless because `beforeChangingState ` is always `Deleted`.

BewareMyPower (approved):
LGTM, left a comment about logging but it's not an important issue

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: Correct. Code might change in the future, so having the logging in place is useful from that perspective.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
lhotari: > Can we optimize the method of the ManagedCursor to remove from the waiting cursors if the cursor has been registered to the managed ledger? So that we can avoid call `topic.getManagedLedger().removeWaitingCursor(cursor);` here.
> 
> ```
>     public void setInactive() {
>         if (isActive) {
>             ledger.deactivateCursor(this);
>             isActive = false;
>         }
>     }
> ```

deactivateCursor doesn't remove the cursor from waiting cursors and it shouldn't. I experimented with that and found out that it would also cause a state inconsistency with dispatcher's "havePendingRead" and make dispatchers to stop receiving messages.

lhotari: > If we look back at the first issue that was fixed by #24551,
> 
> ```
> (Highlight) The method scheduleReadOnActiveConsumer resets the variable cursor.waitingReadOp, but it forgets to remove the cursor from managedledger.waitingCursors
> ```
> 
> https://github.com/apache/pulsar/pull/24551/files#diff-1d721da3aaff7689077d33bf3e55fd8d6a5c9184549857a6120d8050c5d67798L385-L386
> 
> Here, try to add another one.

Removing from the waitingCursors is a relatively costly operation due to the data structure which is a ConcurrentLinkedQueue in ManagedLedgerImpl. That is most likely why it hasn't been done in the original code. Instead of removing it, it's better to rely on ManagedLedgerImpl.notifyCursors to unqueue the waiting cursor. Therefore, the more optimal solution is to track whether there's a waiting cursor already in the queue when it has been requested.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: Thanks, I'll rename this.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: Thanks, I'll rename this.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
codelipenghui: My concern is that we used some internal concepts from the ManagedLedger in PersistentSubscription. We should find a way to avoid that. The state of `waitingCursors` should be self-managed by ManagedLedger.

lhotari: > My concern is that we used some internal concepts from the ManagedLedger in PersistentSubscription. We should find a way to avoid that. The state of `waitingCursors` should be self-managed by ManagedLedger.

I completely agree that it should be avoided. There should have been some sort of event for handling this.
btw. The call `topic.getManagedLedger().removeWaitingCursor(cursor)` wasn't added in this PR. Before #24551, there was a similar call. The call `topic.getManagedLedger().removeWaitingCursor(cursor)` was added to `PersistentSubscription`  in #13939. Another PR in this area was #22454.

PR #24571, opened 2025-07-28
[fix][client][branch-4.0] Partitioned topics are unexpectedly created by client after deletion (#24554)

BewareMyPower:
(cherry picked from commit 969920794840d5be884d039d9626acf64cb362a8)

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


BewareMyPower:
For release branches, partitions automatically created by reconnection should be avoided. PIP-414 prevents this case by checking if the partition metadata exists when `getTopic` is called on a topic with the partition suffix.

lhotari:
@BewareMyPower I assume we could target 4.0.6 for this change? How about 3.3.8 and 3.0.13? Does the same backport apply to those branches?

BewareMyPower:
Yes, branch-3.0 and 3.3 should cherry-pick this PR rather than the commit in master branch.

PR #24572, opened 2025-07-28
[improve][pip]PIP-436: Add decryptFailListener to Consumer

3pacccccc:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-436.md
+++ b/pip/pip-436.md
@@ -1,0 +1,141 @@
+# PIP-436: Add decryptFailListener to Consumer
+
+# Background knowledge
+Apache Pulsar supports client-side encryption where messages can be encrypted by producers and decrypted by consumers. The current mechanism for handling decryption failures uses ConsumerCryptoFailureAction which provides three options:
+
+- FAIL: Fail message consumption (default)
+
+- DISCARD: Silently discard the message
+
+- CONSUME: Deliver the encrypted message to the application
+
+With the introduction of ```isEncrypted``` field in ```EncryptionContext``` ([PIP-432](https://github.com/apache/pulsar/pull/24481)), applications can now determine whether a message was successfully decrypted by checking this flag. However, this requires manual inspection of each message.
+# Motivation
+**This PIP builds upon and extends [PIP-432]((https://github.com/apache/pulsar/pull/24481)) by providing a more elegant way to handle decryption failures. Which was introduced in https://github.com/apache/pulsar/pull/24481#issuecomment-3048640812**
+
+The proposed decryptFailListener would:
+
+- Provide a cleaner, more elegant solution for handling decryption failures
+- Allow more flexible handling of decrypted vs encrypted messages
+- With the ```decryptionFailListener```, users can use listener instead of ```ConsumerCryptoFailureAction```. They can decide in the listener to discard, or decrypt, or anything they want
+- Maintain backward compatibility with existing ConsumerCryptoFailureAction behavior
+
+# Goals
+
+## In Scope
+- Add new DecryptFailListener and ReaderDecryptFailListener interfaces
+
+- Add corresponding builder methods to ConsumerBuilder and ReaderBuilder
+
+- Implement the listener callback mechanism in consumer implementations
+
+- Ensure proper interaction with existing ConsumerCryptoFailureAction settings
+
+- Maintain backward compatibility with existing applications
+
+- Enforce mutual exclusivity between decryptFailListener and ConsumerCryptoFailureAction
+
+## Out of Scope
+
+- Changes to encryption/decryption algorithms or protocols
+- Modifications to the core encryption/decryption process
+- Performance optimizations specific to the new listener
+- Changes to producer-side encryption logic
+# High Level Design
+
+The solution adds new listener interfaces specifically for decryption failures:
+```java
+public interface DecryptFailListener<T> {
+    void received(Consumer<T> consumer, Message<T> message);
+}
+
+public interface ReaderDecryptFailListener<T> {
+    void received(Reader<T> reader, Message<T> message);
+}
+```
+### Configuration Validation:
+1.Mutual Exclusivity Check:
+```java
+if (conf.getDecryptFailListener() != null && conf.getCryptoFailureAction() != null) {
+    throw new InvalidConfigurationException(
+        "decryptFailListener cannot be used with cryptoFailureAction - choose one approach");
+}
+```
+2.Listener Dependency Check:
+```java
+if (conf.getDecryptFailListener() != null && conf.getMessageListener() == null) {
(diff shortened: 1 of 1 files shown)


Technoboy-:
@3pacccccc Please raise a vote for this PIP

3pacccccc:
> @3pacccccc Please raise a vote for this PIP

@Technoboy- OK, thank you for pointing this out, I'll raise a vote right away

poorbarcode (approved):

3pacccccc:
@codelipenghui @poorbarcode @Technoboy- @lhotari @BewareMyPower @dao-jun 
Hi, Sorry to bother you, but could you please take a moment to take a look at this pip and vote on the mailing list when you’re free? I’d really appreciate your help.

pip/pip-436.md
```diff
@@ -0,0 +1,145 @@
...
+}
+
+public interface ReaderDecryptFailListener<T> {
+    void received(Reader<T> reader, Message<T> message);
+}
+```
+### Configuration Validation:
+1.Mutual Exclusivity Check:
+```java
+if (conf.getDecryptFailListener() != null && conf.getCryptoFailureAction() != null) {
```
RobertIndie, line 59: The default value for `CryptoFailureAction` is `FAIL`. This means that users should explicitly set `CryptoFailureAction` to `null` if they are using `DecryptFailListener`. Would it be a good idea to automatically disable the `CryptoFailureAction` feature when `DecryptFailListener` is set?

pip/pip-436.md
```diff
@@ -0,0 +1,145 @@
...
+    throw new InvalidConfigurationException(
+        "decryptFailListener cannot be used with cryptoFailureAction - choose one approach");
+}
+```
+2.Listener Dependency Check:
+```java
+if (conf.getDecryptFailListener() != null && conf.getMessageListener() == null) {
```
RobertIndie, line 66: Why do we need this check?

pip/pip-436.md
```diff
@@ -0,0 +1,145 @@
...
+}
+```
+### Runtime Behavior:
+1. For successful decryption:
+    - Always routes to `messageListener`
+
+2. For failed decryption:
+    - If `decryptFailListener` is configured:
+        - Routes to `decryptFailListener`
+        - Skips `ConsumerCryptoFailureAction` behavior
```
RobertIndie, line 84: We already skip the `ConsumerCryptoFailureAction` behavior here. So seems like we don't need the check `if (conf.getDecryptFailListener() != null && conf.getCryptoFailureAction() != null)`

pip/pip-436.md
```diff
@@ -0,0 +1,145 @@
...
+    void received(Reader<T> reader, Message<T> message);
+}
+```
+#### New ConsumerBuilder methods:
+```java
+ConsumerBuilder<T> decryptFailListener(DecryptFailListener<T> decryptFailListener);
+```
+#### New ReaderBuilder methods:
+```java
+ReaderBuilder<T> readerDecryptFailListener(ReaderDecryptFailListener<T> readerDecryptFailListener);
```
RobertIndie, line 107:
```suggestion
ReaderBuilder<T> decryptFailListener(ReaderDecryptFailListener<T> readerDecryptFailListener);
```

Why don't we use the same config name with the consumer?

pip/pip-436.md
3pacccccc: yes, agree with this point, and that's what I'm planing to do in the implementation

pip/pip-436.md
3pacccccc: Because currently consumer implementation has numerous behaviors tightly coupled with the presence of a MessageListener, such as:
- Determines whether to use push (listener) or pull (receive()) mode
- Affects threading model and message dispatch logic
- Governs automatic vs manual acknowledgement behavior
and so on...

so if Allowing standalone DecryptFailListener would maybe significant refactoring of core consumer logic and makes consumer API complicated.

pip/pip-436.md
3pacccccc: using ```(conf.getDecryptFailListener() != null && conf.getCryptoFailureAction() != null)``` is in order to prevent user set both ```DecryptFailListener``` and ```CryptoFailureAction```

pip/pip-436.md
3pacccccc: I followed the existing pattern where the main listener is called ```readerListener```, so for consistency I named it ```readerDecryptFailListener``` to match.

3pacccccc:
@RobertIndie thank you so much for your reviewing, I left some comments, PTAL

pip/pip-436.md
RobertIndie: Why don't we just skip the checking for `conf.getCryptoFailureAction()` and document it? Otherwise, the user should explicitly set `CryptoFailureAction` to `null` if they want to use `DecryptFailListener`. Also, setting a null value to an enum config seems confusing to me.

3pacccccc: Hi, @RobertIndie  I'm sorry I don't describe this clearly. Let me clarify this more clearly

>  the user should explicitly set CryptoFailureAction to null if they want to use DecryptFailListener

Actually, users don't need to do this. Here's why:
1. my plan is set default value of ```cryptoFailureAction``` to null. source code (https://github.com/apache/pulsar/blob/0a9c439cd3bc88f93baaefb1e2616c33fa06c567/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ConsumerConfigurationData.java#L287)
2. when executing ```consumer.subscribe()``` or ```consumer.subscribeAsync()```, will check if both ```cryptoFailureAction``` and ```DecryptFailListener``` are null, automatically set ```CryptoFailureAction``` as ```ConsumerCryptoFailureAction.FAIL```


**this change is fully backward compatible - it won't break any existing behavior.**
**Example 1**: Normal usage (same as before)
```java
Consumer<byte[]> consumer = pulsarClient.newConsumer()
    .topic("persistent://my-property/my-ns/test-flush-enabled")
    .subscriptionName("my-subscriber-name")
    .subscribe();
// cryptoFailureAction = FAIL (auto-set)
```
**Example 2**: Using DecryptFailListener
```java
ProducerBuilder<byte[]> producerBuilder = pulsarClient.newProducer()
    .topic("persistent://my-property/my-ns/test-flush-enabled")
    .decryptFailListener((consumer, msg) -> {
        // process msg
    })
    .batchingMaxPublishDelay(1, TimeUnit.HOURS)
    .batchingMaxMessages(10000);
// cryptoFailureAction remains null
```

RobertIndie: Thanks for the explaination. It's clear to me now.

PR #24580, opened 2025-07-29
[fix][broker]Fix never recovered metadata store bad version issue if received a large response from ZK

poorbarcode:
### Motivation

**Background: how ZK metadata store handles the node exists error when creating a new node**
ZK metadata store throws a `BadVersionException` error if it receives a `NODEEXISTS` error when creating a new node.

https://github.com/apache/pulsar/blob/v4.0.5/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java#L269-L272

```java
if (code == Code.NODEEXISTS) {
    // We're emulating a request to create a node, so the version is invalid
    op.getFuture().completeExceptionally(getException(Code.BADVERSION, op.getPath()));
}
```

**Background: how ZK metadata store handles disconnect error when executing operations**
ZK metadata store will retry the operation if it receives a `CONNECTIONLOSS` error when executing operations.

https://github.com/apache/pulsar/blob/v4.0.5/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java#L200-L227

```java
if (code == Code.CONNECTIONLOSS) {
    // Retry with the individual operations
    executor.schedule(() -> {
        ops.forEach(o -> batchOperation(Collections.singletonList(o)));
    }, 100, TimeUnit.MILLISECONDS);
}
```

**Background: how ZK clients/servers limit the max length of packet**
- ZK clients/servers limit the max length of a packet to 10MB by default, which was set in [bin/pulsar.sh](https://github.com/apache/pulsar/blob/master/bin/pulsar#L276)
- Users can use a Java environment variable `jute.maxbuffer` to change the maximum packet length, which controls the maximum packet length for both requests and receives.
- Once the maximum packet length is reached, the ZK client will throw an error and reconnect.

**Issue: ZK client throws an error and reconnects when the max length of the packet is exceeded by the server**
- As the background "How ZK clients/servers limit the max length of packet" said, not only is the request packet is in limiting, but the response packet is not limited.
- The requests that type `Get` or `GET_CHILDREN` will receive a response from the server, and the response packet may be larger than the max length of packet length.
- When the response packet is larger than the maximum packet length, the ZK client will throw an error and reconnect.
- The reconnecting will cause the `BadVersionException` error mentioned in the background "How ZK metadata store handles the node exists error when creating a new node".
- Then the operation of creating a node can never be completed.

You can reproduce the issue by running the new test `testReceivedHugeResponse`

```
2025-07-30T02:00:53,768 - WARN  - [bookkeeper-ml-scheduler-OrderedScheduler-6-0:ManagedCursorImpl] - [public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e] Failed to update cursor metadata for s1 due to version conflict org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
2025-07-30T02:00:53,782 - WARN  - [bookkeeper-ml-scheduler-OrderedScheduler-6-0:ManagedLedgerImpl] - [public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e] Failed to open cursor: ManagedCursorImpl{ledger=public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e, name=s1, ackPos=2:-1, readPos=2:0}
2025-07-30T02:00:53,782 - WARN  - [bookkeeper-ml-scheduler-OrderedScheduler-6-0:PersistentTopic] - [persistent://public/default/tp-9d89a816-a363-4607-bf0d-1deea4fe566e] Failed to create subscription for s1: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
2025-07-30T02:00:53,782 - ERROR - [bookkeeper-ml-scheduler-OrderedScheduler-6-0:PersistentTopic] - [persistent://public/default/tp-9d89a816-a363-4607-bf0d-1deea4fe566e] Failed to create subscription: s1
java.util.concurrent.CompletionException: org.apache.pulsar.broker.service.BrokerServiceException$PersistenceException: org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1141) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162) ~[?:?]
	at org.apache.pulsar.broker.service.persistent.PersistentTopic$3.openCursorFailed(PersistentTopic.java:1153) ~[classes/:?]
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl$6.operationFailed(ManagedLedgerImpl.java:1048) ~[classes/:?]
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl$3.operationFailed(ManagedCursorImpl.java:813) ~[classes/:?]
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl$25$1.operationComplete(ManagedCursorImpl.java:2923) ~[classes/:?]
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl$25$1.operationComplete(ManagedCursorImpl.java:2918) ~[classes/:?]
	at org.apache.bookkeeper.mledger.impl.MetaStoreImpl.lambda$asyncGetCursorInfo$13(MetaStoreImpl.java:226) ~[classes/:?]
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire$$$capture(CompletableFuture.java:718) [?:?]
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java) [?:?]
	at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:482) [?:?]
	at org.apache.bookkeeper.common.util.OrderedExecutor$TimedRunnable.run(OrderedExecutor.java:203) [bookkeeper-common-4.17.2.jar:4.17.2]
	at org.apache.bookkeeper.common.util.SingleThreadSafeScheduledExecutorService$SafeRunnable.run(SingleThreadSafeScheduledExecutorService.java:46) [bookkeeper-common-4.17.2.jar:4.17.2]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) [?:?]
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:264) [?:?]
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java) [?:?]
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) [?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.122.Final.jar:4.1.122.Final]
	at java.base/java.lang.Thread.run(Thread.java:833) [?:?]
Caused by: org.apache.pulsar.broker.service.BrokerServiceException$PersistenceException: org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
	... 19 more
Caused by: org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
Caused by: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.getException(ZKMetadataStore.java:516) ~[classes/:?]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.handlePutResult(ZKMetadataStore.java:292) ~[classes/:?]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.lambda$batchOperation$9(ZKMetadataStore.java:258) ~[classes/:?]
	... 8 more
Caused by: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for /managed-ledgers/public/default/persistent/tp-9d89a816-a363-4607-bf0d-1deea4fe566e/s1
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:121) ~[zookeeper-3.9.3.jar:3.9.3]
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:53) ~[zookeeper-3.9.3.jar:3.9.3]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.getException(ZKMetadataStore.java:512) ~[classes/:?]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.handlePutResult(ZKMetadataStore.java:292) ~[classes/:?]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.lambda$batchOperation$9(ZKMetadataStore.java:258) ~[classes/:?]
	... 8 more
```

### Modifications

Add a mechanism to limit `get/getChildren` operations of the ZK metadata store. This mechanism merely avoids problems with as few resources as possible, but it cannot prevent all problems. To solve all the problems, there will be a PIP later to make the Settings configurable, see also the section "Next things to do"

### Next things to do

I will submit a PIP to do the following things
- Make `MetadataNodePayloadLenEstimator` configurable
- Add metrics of the node path that has the max packet length of each type
- Fix the issue of out-of-order command execution in the metadata store. I will describe this problem in detail in PIP
- Solve the following issue

**The Metadata store can not confirm whether the bad version error is expected or not after a reconnection.**
- ZK client sent a `Put` request.
  - Server has received the request and is handling it.
  - Server is responding.
- The channel reconnected before the `Put` request was responded.
- The `Put` request is sent again, but the node already exists(or already modified), so Pulsar get a `BadVersionException` error.
  - **(Highlight)** The previous change is caused by the first try, but the second try is stuck due to the ZK node has been changed by the first try.
- Once this situation happens, we can only restart the broker or reload topics to fix it, since we do not know whether the bad version error is expected or not.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
@@ -220,14 +220,20 @@ public class ZKMetadataStore extends AbstractBatchedMetadataStore
                                         Collectors.groupingBy(MetadataOp::getType, Collectors.summingInt(op -> 1)))
                                 .entrySet().stream().map(e -> e.getValue() + " " + e.getKey().name())
                                 .collect(Collectors.joining(", "));
+                        boolean shouldLimitLogLen = ops.size() > 16;
                         List<Triple<String, String, Integer>> opsForLog = ops.stream()
-                                .filter(item -> switch (item.getType()) {
-                                    case PUT -> item.asPut().getData().length > logThresholdPut;
-                                    case GET -> nodeSizeStats
-                                            .getMaxSizeOfSameResourceType(item.getPath()) > logThresholdGet;
-                                    case GET_CHILDREN -> nodeSizeStats
-                                            .getMaxChildrenCountOfSameResourceType(item.getPath()) > 512;
-                                    default -> false;
+                                .filter(item -> {
+                                    if (!shouldLimitLogLen) {
+                                        return true;
+                                    }
+                                    return switch (item.getType()) {
+                                        case PUT -> item.asPut().getData().length > logThresholdPut;
+                                        case GET -> nodeSizeStats
+                                                .getMaxSizeOfSameResourceType(item.getPath()) > logThresholdGet;
+                                        case GET_CHILDREN -> nodeSizeStats
+                                                .getMaxChildrenCountOfSameResourceType(item.getPath()) > 512;
+                                        default -> false;
+                                    };
                                 })
                                 .map(op -> Triple.of(op.getPath(), op.getType().toString(), op.size()))
                                 .collect(Collectors.toList());
@@ -235,8 +241,8 @@ public class ZKMetadataStore extends AbstractBatchedMetadataStore
                         log.warn("Connection loss while executing batch operation of {} "
                                 + "of total requested data size of {}. "
                                 + "Retrying individual operations one-by-one."
-                                + " ops whose req size > {} or resp size > {} or children count > 512: {}",
-                                countsByType, totalSize, logThresholdPut, logThresholdGet, opsForLog);
+                                + " ops that maybe large: {}",
+                                countsByType, totalSize, opsForLog);
 
                         // Retry with the individual operations
                         executor.schedule(() -> {


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,34 @@
...
+    void recordPut(String path, byte[] data);
+
+    void recordGetRes(String path, GetResult getResult);
+
+    void recordGetChildrenRes(String path, List<String> list);
+
+    int estimateGetResPayloadLen(String path);
+
+    int estimateGetChildrenResPayloadLen(String path);
+}
```
lhotari: It would be useful to add javadocs to the interface level and all methods.

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.HashMap;
+import java.util.List;
+import org.apache.pulsar.common.util.StringSplitter;
+import org.apache.pulsar.metadata.api.GetResult;
+import org.apache.pulsar.metadata.api.MetadataNodePayloadLenEstimator;
+
+public class MaxValueMetadataNodePayloadLenEstimator implements MetadataNodePayloadLenEstimator {
+
+    private final HashMap<PathType, Integer> maxLenPerPathType = new HashMap<>();
+    private final HashMap<PathType, Integer> maxLenPerListPathType = new HashMap<>();
```
lhotari: HashMap isn't thread safe. It's better to use ConcurrentHashMap to avoid problems.

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,132 @@
...
+            return Math.max(v, getResult.getValue().length);
+        });
+    }
+
+    @Override
+    public void recordGetChildrenRes(String path, List<String> list) {
+        PathType pathType = getPathType(path);
+        maxLenPerListPathType.compute(pathType, (k, v) -> {
+            if (v == null) {
+                return list.stream().map(String::length).reduce(0, Integer::sum);
```
lhotari:
```suggestion
                return list.stream().mapToInt(String::length).sum();
```

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,132 @@
...
+            }
+            return Math.max(v, list.stream().map(String::length).reduce(0, Integer::sum));
```
lhotari:
```suggestion
            return Math.max(v, list.stream().mapToInt(String::length).sum());
```

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private PathType getMlPathType(String[] parts) {
+        return switch (parts.length) {
+            case 4 -> PathType.ML_NAMESPACE;
+            case 5 -> PathType.TOPIC;
+            case 6 -> PathType.SUBSCRIPTION;
+            default -> PathType.OTHERS;
+        };
+    }
```
lhotari: not sure if it matters, but V1 topics use a different persistent path than V2 topics. 
https://github.com/apache/pulsar/blob/ae050b9c44ef0be474cedcf78c4dd6ba31cf6cc8/pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java#L333-L349

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,132 @@
...
+    public int estimateGetResPayloadLen(String path) {
+        PathType pathType = getPathType(path);
+        return maxLenPerPathType.getOrDefault(pathType, DEFAULT_LEN);
+    }
+
+    @Override
+    public int estimateGetChildrenResPayloadLen(String path) {
+        PathType pathType = getPathType(path);
+        int maxValueCached = maxLenPerListPathType.getOrDefault(pathType, DEFAULT_LEN);
+        return maxValueCached + (maxValueCached >> 3);
```
lhotari: what's the reasoning behind this estimation formula?

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: Addressed

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: Addressed

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: V1 topic will be calculated as subscription, it is okay.
I added a new case handling for the v1 subscription.

```java
// v2 subscription and v1 topic.
case 6 -> PathType.SUBSCRIPTION;
// v1 subscription.
case 7 -> PathType.SUBSCRIPTION;
```

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: Removed it and added a more precise method to cover the length of ZK system props in the response

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: I have used an array instead of the map, which will not be expanded or shrunk anymore. It is not important enough to ensure thread safety for an accurate value, so skipped to guarantee the thread safety

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataNodePayloadLenEstimator.java
poorbarcode: Added javadocs

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,210 @@
...
+
+    {
+        int pathTypeCount = PathType.values().length;
+        maxLenOfGetMapping = new int[pathTypeCount];
+        maxLenOfListMapping = new int[pathTypeCount];
+        for (int i = 0; i < pathTypeCount; i++) {
+            maxLenOfGetMapping[i] = UNSET;
+            maxLenOfListMapping[i] = UNSET;
+        }
+    }
```
lhotari: could we use ordinary constructor syntax here?

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
lhotari: > It is not important enough to ensure thread safety for an accurate value, so skipped to guarantee the thread safety

In Java, that's not a proper way to handle thread safety. Another thread might never "see" the changes made by another thread. In practice this rarely happens, but it can happen due to compiler optimizations. That's why such patterns should be avoided in general since they lead to non-deterministic and inconsistent behavior.

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: Could you show an example?

lhotari: replace the first `{` with `MaxValueMetadataNodePayloadLenEstimator() {` so that it becomes a constructor.

poorbarcode: Addressed

lhotari:
> I think that there's another way with ZooKeeper which this fix is targeting.
> 
> In current implementation the async version of "multi" is used:
> 
> https://github.com/apache/pulsar/blob/4bfdcd85ff5c9f21ba4fd976adf7785465053151/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java#L195-L202
> 
> The ZooKeeper client also contains a synchronous version: https://javadoc.io/doc/org.apache.zookeeper/zookeeper/latest/org/apache/zookeeper/ZooKeeper.html#multi(java.lang.Iterable) Using the async multi API causes the problem in the first place. It would be possible to retry the reads by splitting the batch etc until it succeeds.

I checked this and it won't be useful since it would require that pipelining wouldn't be used, which would impact performance a lot. It seems that stats/estimation based approach is the way to go to mitigate the problem.

@poorbarcode I noticed that ZooKeeper has `getNumChildren()` in the stats for ZNodes. Perhaps it would be useful to use that before a namespace listing is performed? That would add latency to the first time the namespace listing is performed, but the number of children could be cached after that. The number of topics in a namespace could vary significantly so that's probably the main source of the problems.

Another source of problems in Pulsar is the ZNode data size for ml cursors (ManagedCursorInfo). I guess that stats/estimation based approach would work very well for that. It would be possible to set the estimated size high so that such operations would be executed in small batches.
Perhaps there could be tunables for configuring the initial estimates for different znode categories?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Estimate the payload length of get result.
+     */
+    int estimateGetResPayloadLen(String path);
+
+    /**
+     * Estimate the payload length of list result.
+     */
+    int estimateGetChildrenResPayloadLen(String path);
```
lhotari: For estimating the size of get children calls, I think that it would be very useful if there would be a way to first retrieve the number of children (ZK has numChildren in stats) to be able to do this calculation. It doesn't make sense to do this estimation for all calls, but for listing topics in a namespace, this would be very useful.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,52 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.metadata.api;
+
+import java.util.List;
+
+public interface MetadataNodePayloadLenEstimator {
```
lhotari: Please add the javadoc for the interface itself so that it becomes clear what this is about

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ZKMetadataStoreBatchIOperationTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.testng.annotations.Test;
+
+@Slf4j
+@Test(groups = "broker")
+public class ZKMetadataStoreBatchIOperationTest extends CanReconnectZKClientPulsarServiceBaseTest {
+
+    @Override
+    @BeforeClass(alwaysRun = true, timeOut = 300000)
+    public void setup() throws Exception {
+        System.setProperty("jute.maxbuffer", "16384");
```
lhotari, line 41: Since this is modified, it should be resetted in cleanup so that other tests aren't impacted.

It might also be necessary to configure the CI run to run this test class in isolation. That would be needed if the ZK client caches the jute.maxbuffer setting and we cannot find a way to reset the cached value with reflection.

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: Changed its type to `AtomicReferenceArray`

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ZKMetadataStoreBatchIOperationTest.java
poorbarcode: Fixed

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataNodePayloadLenEstimator.java
poorbarcode: Added

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataNodePayloadLenEstimator.java
poorbarcode: Good idea, but it leads to the metadata store taking care of a certain use case, which makes the metadata store not clear enough.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
```diff
@@ -91,7 +93,12 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
 
     protected abstract CompletableFuture<Boolean> existsFromStore(String path);
 
-    protected AbstractMetadataStore(String metadataStoreName, OpenTelemetry openTelemetry) {
+    protected MetadataNodePayloadLenEstimator payloadLenRegistrar;
```
lhotari: why is the field name called `payloadLenRegistrar`?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStoreConfig.java
```diff
@@ -99,4 +99,9 @@ public class MetadataStoreConfig {
      */
     @Builder.Default
     private OpenTelemetry openTelemetry = OpenTelemetry.noop();
+
+    /**
+     * The estimator to estimate the payload length of metadata node, which used to limit the batch size requested.
```
lhotari, line 104: Fix grammar and improvement this javadoc. It's useful to also expand the explanation to mention the ZooKeeper use case so that it's more understandable.

lhotari: LLM can be used to check and fix the grammar.

lhotari: You can also reference the javadoc of MetadataNodePayloadLenEstimator with `{@link MetadataNodePayloadLenEstimator}`

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,211 @@
...
+public class MaxValueMetadataNodePayloadLenEstimator implements MetadataNodePayloadLenEstimator {
+
+    // Default to max int value, let the first query command do not execute with batch.
+    public static final int DEFAULT_LEN = Integer.MAX_VALUE;
+    public static final int UNSET = -1;
+    // xid: int, zxid: long, err: int, len: int -> 20
+    // Node stat:
+    //   czxid: long, mzxid: long, ctime: long, mtime: long, version: int, cversion: int, aversion: int, -> 44
+    //   ephemeralOwner: long, dataLength: int, numChildren: int, pzxid: long -> 24
+    // total: 88, let's double it to cover different serialization versions.
```
lhotari: please add some explanation to this comment. it's currently not very clear.

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,211 @@
...
+    public void recordGetChildrenRes(String path, List<String> list) {
+        PathType pathType = getPathType(path);
+        // ZK serialize each string with 4 bytes length prefix, so we add 4 bytes for each string.
+        int totalLen = list.stream().mapToInt(String::length).sum() + list.size() * 4;
+        maxLenOfListMapping.set(pathType.ordinal(), Math.max(maxLenOfListMapping.get(pathType.ordinal()), totalLen));
+    }
+
+    @Override
+    public int estimateGetResPayloadLen(String path) {
+        return internalEstimateGetResPayloadLen(path) + ZK_PACKET_SYSTEM_PROPS_LEN;
```
lhotari: why is `ZK_PACKET_SYSTEM_PROPS_LEN` added even in cases where ZK wouldn't be used? Adding a comment could be useful. Is this implementation only for ZooKeeper?

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
```diff
@@ -0,0 +1,211 @@
...
+        SplitPathRes splitPathRes = splitPath(path);
+        if (splitPathRes.partCount < 2) {
+            return PathType.OTHERS;
+        }
+        return switch (splitPathRes.parts[0]) {
+            case "admin" -> getAdminPathType(splitPathRes);
+            case "managed-ledgers" -> getMlPathType(splitPathRes);
+            default -> PathType.OTHERS;
+        };
+    }
```
lhotari: Does this also work for cases where the ZK prefix (root path) would contain multiple levels?

codelipenghui:
> ZK clients/servers limit the max length of a packet to 1MB by default.

In Pulsar, we have added `-Djute.maxbuffer=10485760"` which change the max length to 10MB. Please help update the PR description to avoid the confusion

poorbarcode:
@lhotari @codelipenghui 

I have rewritten with a new solution, could you please review again?

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: I have redefined the interfaces

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: I have redefined the interfaces

pulsar-broker/src/main/java/org/apache/pulsar/zookeeper/MaxValueMetadataNodePayloadLenEstimator.java
poorbarcode: I have redefined the interfaces

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStoreConfig.java
poorbarcode: I have redefined the interfaces

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
poorbarcode: I have redefined the interfaces

coderzc (approved):

PR #24581, opened 2025-07-30
alternative to LeastLongTermMessageRate rather than select broker ran…

gaozhangmin:
# Background
In the current implementation, AvgShedder relies on bundleBrokerMap to make broker selection decisions.
However, this map is empty before load shedding occurs, causing selectBroker() to fall back to random broker assignment.

When users disable load shedding, this results in all bundles being assigned randomly, leading to unbalanced load distribution across brokers, which is undesirable and inconsistent with production expectations.

 # Enhancement
Prior to the introduction of AvgShedder, the default broker selection strategy was LeastLongTermMessageRate.
To address the issue of randomness and ensure a more reasonable broker assignment:

AvgShedder is now modified to fallback to LeastLongTermMessageRate
when bundleBrokerMap is empty during broker selection.

 # Benefit
Ensures stable and deterministic broker assignment, even if load shedding is disabled

Prevents random distribution of bundles, improving cluster balance and consistency

Maintains backward compatibility with the previously used strategy

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/AvgShedder.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/AvgShedder.java
@@ -56,6 +56,9 @@ public class AvgShedder implements LoadSheddingStrategy, ModularLoadManagerStrat
     private final Map<String, MutableInt> brokerHitCountForLow = new HashMap<>();
     private static final double MB = 1024 * 1024;
 
+    // placement strategy for selecting broker, only if bundleBrokerMap is empty, for first use.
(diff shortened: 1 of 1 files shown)


lhotari:
@gaozhangmin Please set a proper title for this PR

thetumbled:
Have you performed enough performance test on this PR to verify that it offers improvements?

thetumbled (changes requested):
I don't agree with removing the old logic. `AvgShedder` should not interfere with other load balancing algorithms, as the existing algorithms have complex relationships that would create chaos.
By the way, it isn't random selection logic; it's hash-based logic, which distributes bundles evenly across all brokers, as opposed to randomness.

PR #24596, opened 2025-08-01
[improve][broker]Call scheduleAtFixedRateNonConcurrently for scheduled tasks, instead of scheduleAtFixedRate

poorbarcode:
### Motivation

**Background of `scheduleWithFixedDelay` and `scheduleWithFixedRate`**
- `scheduleAtFixedRate`: Executes tasks at fixed intervals (period) regardless of task execution time, 
  - If task execution time ≤ period: Runs normally at the specified interval (e.g., every 1 second).
  - If task execution time > period: New tasks will trigger immediately (or queue up), causing backlog (may lead to consecutive executions). 
- `scheduleWithFixedDelay`: Trigger Mechanism: Waits for the previous task to complete, then delays by the fixed interval (delay) before the next execution.
  - Regardless of task duration, the interval is always measured from the end of the last task.

**Issue: broker ran checkMessageExpiry many times in 1s**

When system resources are tight, it leads to a large backlog of tasks `checkMessageExpiry`, which are executed multiple times at the same time, causing even greater pressure

image

### Modifications

Use `scheduleAtFixedRateNonConcurrently` to override `scheduleWithFixedRate`, which executes tasks at fixed intervals (period) regardless of task execution time, and never causes backlogs

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -362,8 +362,8 @@ public class BrokerService implements Closeable {
         this.inactivityMonitor = new SingleThreadNonConcurrentFixedRateScheduler("pulsar-inactivity-monitor");
         this.messageExpiryMonitor = new SingleThreadNonConcurrentFixedRateScheduler("pulsar-msg-expiry-monitor");
         this.compactionMonitor = new SingleThreadNonConcurrentFixedRateScheduler("pulsar-compaction-monitor");
-        this.consumedLedgersMonitor
-                = new SingleThreadNonConcurrentFixedRateScheduler("pulsar-consumed-ledgers-monitor");
+        this.consumedLedgersMonitor =
+                new SingleThreadNonConcurrentFixedRateScheduler("pulsar-consumed-ledgers-monitor");
         this.backlogQuotaManager = new BacklogQuotaManager(pulsar);
         this.backlogQuotaChecker = new SingleThreadNonConcurrentFixedRateScheduler("pulsar-backlog-quota-checker");
         this.authenticationService = new AuthenticationService(pulsar.getConfiguration(),


Technoboy-:
Could we just fix the `checkMessageExpiry ` from `scheduleWithFixedRate ` to `scheduleWithFixedDelay `?

lhotari (commented):
To avoid making a major behavioral change that could cause breaking changes, we should only make `scheduleAtFixedRate` -> `scheduleWithFixedDelay` changes to the tasks where the execution interval is relatively short.

For scheduled  tasks that continue use `scheduleAtFixedRate`, it would be useful to implement a wrapper that keeps state of in progress execution and skips that task and logs a warning if the previous task is still in execution. We already have the `org.apache.pulsar.common.util.Runnables` class in pulsar-common and we could add this logic there. The existing `catchingAndLoggingThrowables` method could take another argument `boolean skipConcurrentExecution` that would handle skipping a concurrent call to the `Runnable` instance while the previous execution is happening.

WDYT?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -1404,13 +1404,13 @@ protected void handleSubscribe(final CommandSubscribe subscribe) {
 
                             if (consumerFuture.complete(consumer)) {
                                 log.info("[{}] Created subscription on topic {} / {}",
-                                        remoteAddress, topicName, subscriptionName);
+                                        remoteAddress, topicName, subscriptionName); // TODO consumer id.
```
lhotari: Please omit adding unrelated TODOs in this PR. You can keep the TODOs to yourself and handle these in a separate PR.

dao-jun:
I think maybe we don't need to replace all `scheduleWithFixedDelay` to `scheduleWithFixedRate`, for these tasks which costs long time, replace to `scheduleWithFixedRate` I afraid it will cost high broker work load?

lhotari:
> I think maybe we don't need to replace all `scheduleWithFixedDelay` to `scheduleWithFixedRate`, for these tasks which costs long time, replace to `scheduleWithFixedRate` I afraid it will cost high broker work load?

@dao-jun  I don't think that anyone is suggesting to replace `scheduleWithFixedDelay` with `scheduleAtFixedRate`. It's the other way around that is in this PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Thanks for mentioning this, I have removed these changes

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java
```diff
@@ -231,7 +231,7 @@ private ManagedLedgerFactoryImpl(MetadataStoreExtended metadataStore,
         this.entryCacheManager = new RangeEntryCacheManagerImpl(this, scheduledExecutor, openTelemetry);
         this.statsTask = scheduledExecutor.scheduleWithFixedDelay(catchingAndLoggingThrowables(this::refreshStats),
                 0, config.getStatsPeriodSeconds(), TimeUnit.SECONDS);
-        this.flushCursorsTask = scheduledExecutor.scheduleAtFixedRate(catchingAndLoggingThrowables(this::flushCursors),
+        this.flushCursorsTask = scheduledExecutor.scheduleWithFixedDelay(catchingAndLoggingThrowables(this::flushCursors),
                 config.getCursorPositionFlushSeconds(), config.getCursorPositionFlushSeconds(), TimeUnit.SECONDS);
```
lhotari: Changing from `scheduleAtFixedRate` to `scheduleWithFixedDelay` could have a significant impact to behavior. Let's say if cursorPositionFlushSeconds is set to 1 and flushing takes 300ms. After the change, cursors would get flushed every 1300 ms instead of every 1000 ms.
This is why I think that we should consider each case closely where we are going to make a change from `scheduleAtFixedRate` to `scheduleWithFixedDelay`.
One possible alternative solution would be to keep on using `scheduleAtFixedRate` and implement the wrapper in `Runnables` that I explained in a separate comment.

lhotari:
@poorbarcode The description refers `scheduleWithFixedRate`. The correct method name is `scheduleAtFixedRate`

BewareMyPower (dismissed):
This PR involves many changes.

Share the comparison of these two methods from Gemini AI:

> `scheduleAtFixedRate` is best for:
> Time-sensitive tasks that need a fixed schedule (e.g., cron jobs, hourly chimes).	Tasks where a fixed pause between executions is important, especially if task duration is unpredictable.
>
> `scheduleWithFixedDelay` is best for
> Tasks where a fixed pause between executions is important, especially if task duration is unpredictable.
>
> The choice between scheduleAtFixedRate and scheduleWithFixedDelay depends entirely on your specific use case. 

Since the issue is exactly with `checkMessageExpiry`, please only change its related logic.

BewareMyPower:
Even with this change, it's nearly impossible to prevent new code from calling `scheduleAtFixedRate`. It's unfair to say one is definitely better than the other.

lhotari:
> This PR involves many changes.
> 
> Share the comparison of these two methods from Gemini AI:
> 
> > `scheduleAtFixedRate` is best for:
> > Time-sensitive tasks that need a fixed schedule (e.g., cron jobs, hourly chimes).	Tasks where a fixed pause between executions is important, especially if task duration is unpredictable.
> > `scheduleWithFixedDelay` is best for
> > Tasks where a fixed pause between executions is important, especially if task duration is unpredictable.
> > The choice between scheduleAtFixedRate and scheduleWithFixedDelay depends entirely on your specific use case.
> 
> Since the issue is exactly with `checkMessageExpiry`, please only change its related logic.

@BewareMyPower I think that analysis misses the detail that `scheduleWithFixedDelay` will prevent concurrent execution. My suggestion would be to continue to use `scheduleAtFixedRate`, but add the solution to `Runnables` class to prevent concurrent execution which could cause many issues. (explained in my previous comment)

BewareMyPower:
Actually the original analysis includes that point, but it's too long that I didn't include it.

However, back to the issue, users should not configure a short interval like 1s and it's impossible because it can only be configured by `messageExpiryCheckIntervalInMinutes`, whose time unit is minute. Even with the smallest config (1 minute), could `scheduleAtFixedRate` be an actual issue? If so, it means `BrokerService#checkMessageExpiry` could take more than 1 minute, which is crazy. Instead of switching to `scheduleWithFixedDelay`, we should investigate why it could take so long. If it cannot be optimized, it should not be fixed at Pulsar side. In this case, it should be an operation issue that should be fixed by:
- Increase the CPU (since it means there are a lot of topics to be served)
- Increase the number of brokers (reduce the number of topics processed by each broker)
- Increase the `messageExpiryCheckIntervalInMinutes` config (to avoid frequent time consuming operation)

P.S. From the screenshot, message expiry check is scheduled every 5 minutes, which is a default value. It also shows no queued tasks being executed immediately.

`scheduleWithFixedDelay` might be able to avoid queuing scheduled tasks in extreme cases, but from all existing info I can see, it does not solve an actual issue.

lhotari:
> `scheduleWithFixedDelay` might be able to avoid queuing scheduled tasks in extreme cases, but from all existing info I can see, it does not solve an actual issue.

@BewareMyPower I agree that it doesn't fix the actual issue. That's why I'd recommend using `scheduleAtFixedRate` with the wrapper that optionally logs a warning when the previous execution is still running. In that case, it could be caused by a deadlock, for example. Logging would help in investigating and fixing the root cause. It would also be possible to track the thread that is in execution and log the stack trace of the thread that is stuck in executing the task.

Without preventing concurrent job execution, the problem would get worse since another task would start executing the same task concurrently. That could already happen in cases where we have very short delays and lets say use synchronization to run the task one-by-one. All later coming tasks would get back logged in the executor which is a problem that should be solved. I have understood that it's the problem that @poorbarcode would like to address.

BewareMyPower:
Yes, adding a wrapper of `scheduleAtFixedRate` makes sense

poorbarcode:
> Yes, adding a wrapper of scheduleAtFixedRate makes sense
> That's why I'd recommend using scheduleAtFixedRate with the wrapper that optionally logs a warning when the previous execution is still running.

@BewareMyPower @lhotari Thanks for your suggestion, I will improve the implementation

poorbarcode:
>> Yes, adding a wrapper of scheduleAtFixedRate makes sense
>> That's why I'd recommend using scheduleAtFixedRate with the wrapper that optionally logs a warning when the previous execution is still running.

> @BewareMyPower @lhotari Thanks for your suggestion, I will improve the implementation 

Finished, please review again

pulsar-common/src/main/java/org/apache/pulsar/common/util/SingleThreadSafeScheduledExecutorService.java
```diff
@@ -0,0 +1,382 @@
...
+import java.util.concurrent.RunnableScheduledFuture;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class SingleThreadSafeScheduledExecutorService extends ScheduledThreadPoolExecutor
```
lhotari: The name `SingleThreadSafeScheduledExecutorService` is misleading. This is "single threaded", but the purpose is to prevent concurrent execution of tasks that are scheduled with fixed rate. Therefore a better name would be SingleThreadNonConcurrentFixedRateScheduler`. There's no need to extend `ScheduledThreadPoolExecutor`. This class could simply contain the reference to the `OrderedScheduler` and contain a single method to schedule tasks that are wrapped with a `Runnable` implementation that prevents concurrent execution.

```
OrderedScheduler.newSchedulerBuilder()
                .name(name)
                .numThreads(1)
                .build();
```

There could be a single method `scheduleAtFixedRateNonConcurrently`.

lhotari: I'd recommend implementing the wrapper for preventing concurrent execution of the `Runnable` task in the existing `org.apache.pulsar.common.util.Runnables` class, since in that case, it could be reused even without using this class.

lhotari (changes requested):

pulsar-common/src/main/java/org/apache/pulsar/common/util/SingleThreadSafeScheduledExecutorService.java
poorbarcode: Renamed the class and the method
- `SingleThreadSafeScheduledExecutorService` -> `SingleThreadNonConcurrentFixedRateScheduler`
- `scheduleAtFixedRateAndDropOutdatedTask` -> `scheduleAtFixedRateNonConcurrently`

poorbarcode: > I'd recommend implementing the wrapper for preventing concurrent execution of the Runnable task in the existing org.apache.pulsar.common.util.Runnables class, since in that case, it could be reused even without using this class.

I think the new class can benefit clearer, simpler and more efficient. BTW, we can hardly know whether the Runnble is convectively executing.

lhotari: > I think the new class can benefit clearer, simpler and more efficient. 

I don't think that implementing a complete ScheduledExecutorService is necessary at all. The implementation looks very complex, which isn't needed to address the problem.

> BTW, we can hardly know whether the Runnble is convectively executing.

Oh yes, now I see the problem. With a single threaded executor, there wouldn't be a way to skip scheduling the task if the previous task is overdue. 

For a multi-threaded task executor, this type of wrapper would be useful:

```java
    public static Runnable preventConcurrentExecution(Runnable runnable, String taskName) {
        return new NonConcurrentRunnable(runnable, taskName);
    }

    private static final class NonConcurrentRunnable implements Runnable {
        private final Runnable runnable;
        private final String taskName;
        private final AtomicLong taskStarted = new AtomicLong(0);

        private NonConcurrentRunnable(Runnable runnable, String taskName) {
            this.runnable = runnable;
            this.taskName = taskName;
        }

        @Override
        public void run() {
            if (taskStarted.compareAndSet(0L, System.nanoTime())) {
                try {
                    runnable.run();
                } catch (Throwable t) {
                    LOGGER.error("Unexpected throwable caught when executing task '{}'", taskName, t);
                } finally {
                    taskStarted.set(0L);
                }
            } else {
                long elapsedTimeNanos = System.nanoTime() - taskStarted.get();
                LOGGER.warn(
                        "Runnable for task '{}' is already running, skipping execution. "
                        + "Elapsed time since current task started: {} ms", taskName,
                        TimeUnit.NANOSECONDS.toMillis(elapsedTimeNanos));
            }
        }
    }
```

In this case, since we are executing in a single dedicated thread, I think that we shouldn't implement the ScheduledExecutorService at all and instead make it specialized for the use case where a single task is executed with an ordinary java.lang.Thread (well it makes sense to use Netty's io.netty.util.concurrent.FastThreadLocalThread thread implementation) and with logic which calculates the Thread.sleep between task executions so that it executes at fixed rate, preventing the backlogging of the single task that it can execute.

poorbarcode: > I don't think that implementing a complete ScheduledExecutorService is necessary at all. The implementation looks very complex, which isn't needed to address the problem.

If you ever read the implementation of `ScheduledThreadPoolExecutor` and the internal class `ScheduledFutureTask`, you will know that the new class is very simple, which just changed the implementation of get next runtime of the scheduled task: [setNextRunTime](https://github.com/apache/pulsar/pull/24596/files#diff-cc1fdb222ce13e343daa9172f667282c9fa0a96ade66830b87ff4066fc2d646aR345-R353)

- the previous: `nextRunTime = latestRunTime + period`
- the PR did: `nextRunTime = latestRunTime + ((System.nanoTime() - latestRunTime) / period + 1) * period`

>  I think that we shouldn't implement the ScheduledExecutorService at all and instead make it specialized for the use case where a single task is executed with an ordinary java.lang.Thread (well it makes sense to use Netty's io.netty.util.concurrent.FastThreadLocalThread thread implementation) and with logic which calculates the Thread.sleep between task executions so that it executes at fixed rate, preventing the backlogging of the single task that it can execute.

The current implementation does this

coderzc:
> Modifications
> Call scheduleWithFixedDelay instead of scheduleWithFixedRate

Please update the change description.

poorbarcode:
@coderzc 

> Please update the change description.

Thanks for mentioning this, please review again

coderzc (approved):

coderzc:
@lhotari please review this pr again

lhotari (approved):

Issue #24585, opened 2025-07-30
[Enhancement] Add dependency file name information to error message when .nar file validation fails with ZipException


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

While investigating a CI failure, the only error from the end user level was "java.util.zip.ZipException: zip END header not found" when trying to create a Pulsar IO Connector instance.

In this type of case, it would be useful to return the file name of the dependency in META-INF/bundled-dependencies which fails to validate

here's an example exception in function worker logs:

```
196-2025-07-29T22:19:27,036+0000 [function-web-36-4] ERROR org.apache.pulsar.functions.worker.rest.api.SourcesImpl - Invalid register Source request @ /public/default/test-source-debezium-oracle-PROCESS-ikvvswvz
197:java.io.UncheckedIOException: java.util.zip.ZipException: zip END header not found
198-    at org.apache.pulsar.functions.utils.FunctionFilePackage.<init>(FunctionFilePackage.java:82) ~[org.apache.pulsar-pulsar-functions-utils-4.0.6-SNAPSHOT.jar:4.0.6-SNAPSHOT]
199-    at org.apache.pulsar.functions.utils.io.Connector.getConnectorFunctionPackage(Connector.java:55) ~[org.apache.pulsar-pulsar-functions-utils-4.0.6-SNAPSHOT.jar:4.0.6-SNAPSHOT]
200-    at org.apache.pulsar.functions.worker.rest.api.SourcesImpl.validateUpdateRequestParams(SourcesImpl.java:681) ~[org.apache.pulsar-pulsar-functions-worker-4.0.6-SNAPSHOT.jar:4.0.6-SNAPSHOT]
201-    at org.apache.pulsar.functions.worker.rest.api.SourcesImpl.registerSource(SourcesImpl.java:153) ~[org.apache.pulsar-pulsar-functions-worker-4.0.6-SNAPSHOT.jar:4.0.6-SNAPSHOT]
202-    at org.apache.pulsar.functions.worker.rest.api.v3.SourcesApiV3Resource.registerSource(SourcesApiV3Resource.java:72) ~[org.apache.pulsar-pulsar-functions-worker-4.0.6-SNAPSHOT.jar:4.0.6-SNAPSHOT]
--
[...]

PR #24598, opened 2025-08-04
[Enhancement] Add dependency file name information to error message when .nar file validation fails with ZipException #

falser101:
Main Issue: #24585


pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/NamespacesTest.java
lhotari, line 1: This doesn't seem to be related to the issue that is being resolved.

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/PersistentTopicsTest.java
lhotari, line 1: This doesn't seem to be related to the issue that is being resolved.

lhotari (changes requested):
There are unrelated changes in this PR.
One of the problems is that you are using your `master` branch for this PR. It's recommended to create a unique PR branch for each PR so that you can have multiple outstanding PRs. You won't be able to use ["Personal CI"](https://pulsar.apache.org/contribute/personal-ci/) if you are using the `master` branch in your own fork. When contributing to `apache/pulsar`, you can run CI in your own fork with the "Personal CI" instructions, as long as you have a unique branch name. 

Another useful change would be to add some tests to ensure that this solution works. It's not absolutely necessary since we don't require 100% test coverage, but it's useful.

PR #23499, opened 2024-10-22
[fix][client] Fix producer/consumer stop to reconnect or Pub/Sub due to IO thread race-condition 

poorbarcode:
### Motivation

**Issue 1**: see the test `testUnstableNetWorkWhenCreatingProducer`

The producer fails to start if the network is unstable.

| time | socket closed | start producer |
| --- | --- | --- |
| 1 | | get a pooled connection from the pool |
| 2 | | before registering the producer to the connection |
| 3 | socket closed |
| 4 | reconnect all producers, which is empty now |
| 5 | | register the producer to a closed connection |
| 6 | | failed to send `Command.newProducer` to brokers |
| 7 | | users receive an error `io.netty.channel.StacklessClosedChannelException: null` |

---

<del>
**Issue 2**: see the test: `testSendAfterCreateProducerAsync`
- Start a producer/consumer async
- Call a sync method in the callback method of method `createAsync`, such as `producer.send`, and `consumer.receive`.
- The IO thread encounters a deadlock
</del>

### Modifications

- fix the issues above

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24599, opened 2025-08-04
[fix][client] Retry for unknown exceptions when creating a producer or consumer

BewareMyPower:
### Motivation

There are several methods that get a connection from the `ConnectionPool`.
1. `ConnectionPool#getConnection(ServiceNameResolver)`
    It's only used in `BinaryProtoLookupService`. The callbacks are all executed in `PulsarClientImpl#lookupExecutorProvider`: a single thread executor whose thread name starts with `pulsar-client-lookup`.
2. `ConnectionPool#getConnection(InetSocketAddress)`
    It's called by the 1st method directly. Besides, it's only called by `BinaryProtoLookupService#findBroker`, which also uses `PulsarClientImpl#lookupExecutorProvider` to execute the callback.
3. `ConnectionPool#getConnection(InetSocketAddress logicalAddress, InetSocketAddress physicalAddress, int randomKey)`
    It's called by the 2nd method directly. Besides, it's only called by the 4th method.
4. `PulsarClientImpl#getConnection(InetSocketAddress, InetSocketAddress, int)`
    It's called in `ConnectionHandler#grabCnx` to establish a connection between broker and client (producer, consumer or reader). The callback calls `connectionOpened` or `handleConnectionError` without switching to another executor.
5. Other methods in `PulsarClientImpl`
    Including:
    - `getConnectionToServiceUrl`
    - `getConnection(String, int)`
    - `getConnection(String, String)`
    They all call the 4th method in the callback of `LookupService#getBroker` and only used in `grabCnx`.

To solve a race condition caused by the fact that socket is closed in Netty's I/O thread while `connectionOpened` that sends the command is executed in another thread, #23499 completes the future of `ConnectionPool#getConnection` in Netty's' I/O thread as well. However, this adds an additional thread switching for all usages in method 1 above, which is not necessary.

I found this issue when I found a producer creation was blocked forever due to a deadlock in `sendAsync`'s callback, which is executed in Netty's I/O thread. When checking the heap dump, I found `ClientCnx#pendingRequests` was empty, which means `client.getCnxPool().getConnection(socketAddress)` never complete, see https://github.com/apache/pulsar/blob/1e57827b33395a2124593d95fa7641ee5e125d9b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/BinaryProtoLookupService.java#L218

Though even if without the change, other response processing will still be blocked because the I/O thread is blocked. It's very confusing when reviewing the heap dump:
- `BinaryProtoLookupService#partitionedMetadataInProgress` is not empty
- `ClientCnx#pendingRequests` is empty (the only connection in the pool)

image

image

Actually, the root cause of the issue described in #23499 is that the `StacklessClosedChannelException` is treated as an exception cannot be retried. However, all network exceptions should be retried. Hence, this PR proposed a different solution to retry for such errors. Technically, only a few known exceptions should be treated as not retriable, e.g. `AuthorizationException`. Other known or unknown exceptions should be retried.

It's not guaranteed that `writeAndFlush` will always succeed. For example, if the code reaches here: https://github.com/apache/pulsar/blob/d2728253c666f7d4bd4f111356f3b97663603b6a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L1062

The future of `ClientCnx#sendRequestWithId` could fail with an exception that is not `PulsarClientException`.

### Modifications

Revert the change in #23499 and retry for retriable exception even if it's not `PulsarClientException`. Improve `SimpleProduceConsumeIoTest` to cover consumer creation as well. Since this test only covers a very limited case, add `testUnknownRpcExceptionFor*` tests that inject failure on the 1st `writeAndFlush` in `connectionOpened`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/SimpleProduceConsumeIoTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/SimpleProduceConsumeIoTest.java
@@ -18,30 +18,42 @@
  */
 package org.apache.pulsar.client.impl;
 
-import static org.testng.Assert.assertFalse;
-import static org.testng.Assert.assertNotNull;
-import static org.testng.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
+import static org.testng.Assert.assertEquals;
+import io.netty.channel.ChannelFuture;
+import io.netty.util.concurrent.GenericFutureListener;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.broker.BrokerTestUtil;
+import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.Producer;
-import org.apache.pulsar.client.api.ProducerBuilder;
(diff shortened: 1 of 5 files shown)


lhotari (approved):
LGTM

Technoboy- (approved):

PR #10352, opened 2021-04-24
Make Consumer thread safe and lock-free

315157973:
### Motivation
Lock-free solution for https://github.com/apache/pulsar/pull/10240

PR #24601, opened 2025-08-04
[improve][doc] Improve the JavaDocs of sendAsync to avoid improper use

BewareMyPower:
### Motivation

It's never documented that the future of `sendAsync` is called in a single I/O thread. Without knowing it, it would be very easy to use `sendAsync` improperly, which might affect other producers or consumers in the same client, or even lead to deadlocks.

### Modifications

Improve the JavaDocs of `Producer#sendAsync`.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/Producer.java
@@ -20,6 +20,8 @@ package org.apache.pulsar.client.api;
 
 import java.io.Closeable;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.function.Consumer;
 import org.apache.pulsar.client.api.transaction.Transaction;
 import org.apache.pulsar.common.classification.InterfaceAudience;
 import org.apache.pulsar.common.classification.InterfaceStability;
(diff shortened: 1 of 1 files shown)


Technoboy- (approved):

BewareMyPower:
Yes, `sendAsync` should be the most frequently used async API and it's a critical path. I don't have a chance to check all other async methods, but I have checked `receiveAsync` as well. Unlike `sendAsync`, `receiveAsync` switches to a different thread pool (`PulsarClientImpl#internalExecutorProvider`) from https://github.com/apache/pulsar/pull/10352, so it does not have the similar issue, but it could still affect other processing. Here are my notes when I read the code.

----

Currently a `PulsarClient` will create the following thread pools:

| Thread Pool Name | Number of Threads |
| :- | :- |
| `internalExecutorProvider` | `numIoThreads` |
| `externalExecutorProvider` | `numListenerThreads` |
| `lookupExecutorProvider` | 1 |

`internalExecutorProvider` is used for:
- Execute message listener, callbacks of `receiveAsync()` and `batchReceiveAsync()` for consumers
- Many callbacks in `TransactionMetaStoreHandler`

`externalExecutorProvider` is used for:
- Trigger message listener
- Trigger consumer event listener

----

As you can see, transaction metadata processing could be affected. But I think it's because executors are used without any careful consideration. Before adding documents, we might need to think if the existing behavior should be documented.

PR #24606, opened 2025-08-05
[improve][broker]Remove block calling that named cursor.asyncGetNth when expiring messages

poorbarcode:
### Motivation

Broker checks TTL policies in the background per 5 minutes in a single thread. It calls a block method named `topic.isOldestMessageExpired`, which causes a prolonged thread blockage.

### Modifications

Instead of calling a block method, call a non-blocked method

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
@@ -87,8 +87,7 @@ public class PersistentMessageExpiryMonitor implements FindEntryCallback, Messag
 
     @Override
     public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
-        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds),
-                cursor.getManagedLedger().asyncMigrate().defaultExecutor());
+        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds), topic.getOrderedExecutor());
     }
 
     @Override
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
@@ -87,7 +87,7 @@ public class PersistentMessageExpiryMonitor implements FindEntryCallback, Messag
 
     @Override
     public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
-        return CompletableFuture.completedFuture(expireMessages(messageTTLInSeconds));
+        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds));
     }
 
     @Override
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
@@ -87,7 +87,8 @@ public class PersistentMessageExpiryMonitor implements FindEntryCallback, Messag
 
     @Override
     public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
-        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds));
+        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds),
+                cursor.getManagedLedger().asyncMigrate().defaultExecutor());
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


poorbarcode:
> I‘m +1 with the change, but do we need to add tests for the new methods?

I have added the same test cases that call `expireMessages` directly. The other case that does not call `expireMessages` directly should be covered by the old tests, such as `testDefaultBacklogTTL`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
```diff
@@ -84,6 +85,11 @@ public boolean isAutoSkipNonRecoverableData() {
                 && this.cursor.getManagedLedger().getConfig().isAutoSkipNonRecoverableData();
     }
 
+    @Override
+    public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
+        return CompletableFuture.completedFuture(expireMessages(messageTTLInSeconds));
```
lhotari: `expireMessages` is a blocking implementation.
This could be made to run asynchronously in a separate thread (Executor) with `CompletableFuture.supplyAsync` without having to rewrite the existing blocking implementation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -1236,6 +1236,25 @@ public boolean expireMessages(int messageTTLInSeconds) {
...
 
+    @Override
+    public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
+        long backlog = getNumberOfEntriesInBacklog(false);
+        if (backlog == 0) {
+            return CompletableFuture.completedFuture(false);
+        }
+        if (dispatcher != null && dispatcher.isConsumerConnected() && backlog < MINIMUM_BACKLOG_FOR_EXPIRY_CHECK) {
+            return topic.isOldestMessageExpiredAsync(cursor, messageTTLInSeconds).thenCompose(bool -> {
+                if (bool) {
```
lhotari: use a proper variable name instead of `bool`. For example `oldestMsgExpired`.

poorbarcode: Renamed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
poorbarcode: Changed to `CompletableFuture.supplyAsync`.

BTW, the other child classes need to add a new method, such as PersistentSubscription

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
```diff
@@ -84,6 +85,11 @@ public boolean isAutoSkipNonRecoverableData() {
                 && this.cursor.getManagedLedger().getConfig().isAutoSkipNonRecoverableData();
     }
 
+    @Override
+    public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
+        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds));
```
lhotari: Plain `supplyAsync` would use the default ForkJoinPool for execution. I guess that would be fine since ForkJoinPool can handle blocking methods by spanning more threads when needed (and shrink the number of threads later). However, we might want to choose the thread pool which is used for execution.

poorbarcode: Added the managed ledger executor

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
```diff
@@ -84,6 +85,12 @@ public boolean isAutoSkipNonRecoverableData() {
                 && this.cursor.getManagedLedger().getConfig().isAutoSkipNonRecoverableData();
     }
 
+    @Override
+    public CompletableFuture<Boolean> expireMessagesAsync(int messageTTLInSeconds) {
+        return CompletableFuture.supplyAsync(() -> expireMessages(messageTTLInSeconds),
+                cursor.getManagedLedger().asyncMigrate().defaultExecutor());
```
lhotari: `asyncMigrate().defaultExecutor()` doesn't seem to make sense.

poorbarcode: Could you explain the reason?

lhotari: In https://github.com/apache/pulsar/pull/24606#discussion_r2256253527 you mentioned using managed ledger executor. calling the `asyncMigration()` method doesn't return an executor. It runs the `asyncMigration()` method and returns a `CompletableFuture`. The `CompletableFuture` has a `defaultExecutor` which returns the default ForkJoinPool by default. This is why this doesn't make sense to me.

poorbarcode: Sorry, the IDE mistakenly completed auto-filling. The question has been modified to `topic.getOrderedExecutor()`

Issue #24608, opened 2025-08-05
Flaky-test: EntryCacheManagerTest.simple


3pacccccc:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/3pacccccc/pulsar/actions/runs/16750061782/job/47418435612?pr=20#step:11:2369

### Exception stacktrace

```text
2025-08-05T12:42:39.3658587Z [ERROR] org.apache.bookkeeper.mledger.impl.EntryCacheManagerTest.simple  Time elapsed: 0.005 s  <<< FAILURE!
2025-08-05T12:42:39.3659452Z java.lang.AssertionError: expected [10] but found [6]
2025-08-05T12:42:39.3660441Z 	at org.testng.Assert.fail(Assert.java:110)
2025-08-05T12:42:39.3661165Z 	at org.testng.Assert.failNotEquals(Assert.java:1577)
2025-08-05T12:42:39.3662431Z 	at org.testng.Assert.assertEqualsImpl(Assert.java:149)
2025-08-05T12:42:39.3663191Z 	at org.testng.Assert.assertEquals(Assert.java:131)
2025-08-05T12:42:39.3663868Z 	at org.testng.Assert.assertEquals(Assert.java:979)
2025-08-05T12:42:39.3664456Z 	at org.testng.Assert.assertEquals(Assert.java:955)
2025-08-05T12:42:39.3665175Z 	at org.testng.Assert.assertEquals(Assert.java:989)
2025-08-05T12:42:39.3666335Z 	at org.apache.bookkeeper.mledger.impl.EntryCacheManagerTest.simple(EntryCacheManagerTest.java:111)
2025-08-05T12:42:39.3667932Z 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
2025-08-05T12:42:39.3668963Z 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
2025-08-05T12:42:39.3670228Z 	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
2025-08-05T12:42:39.3671579Z 	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
[...]

PR #24609, opened 2025-08-05
[fix][test] Fix flaky EntryCacheManagerTest.simple

3pacccccc:
Fixes https://github.com/apache/pulsar/issues/24608
### motivation
this test was flaky due to
**line108** in ```EntryCacheManagerTest.java``` of current master branch
already triggered cache eviction, so sometimes the eviction is over, sometimes not, cause this flaky.

### Modifications
1. Removed one entry insertion (`EntryImpl.create(2, 2, new byte[1]`) to:
   - Ensure predictable cache eviction timing
   - Maintain test's original purpose of verifying eviction logic
2. Updated all affected assertions to match the new expected cache sizes:
   - Reduced expected sizes by 1 across all assertions
   - Maintained consistent state validation throughout test flow
### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/21

### DIFF
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/EntryCacheManagerTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/EntryCacheManagerTest.java
@@ -105,10 +105,9 @@ public class EntryCacheManagerTest extends MockedBookKeeperTestCase {
 
         cache2.insert(EntryImpl.create(2, 0, new byte[1]));
         cache2.insert(EntryImpl.create(2, 1, new byte[1]));
-        cache2.insert(EntryImpl.create(2, 2, new byte[1]));
 
-        assertEquals(cache2.getSize(), 3);
-        assertEquals(cacheManager.getSize(), 10);
+        assertEquals(cache2.getSize(), 2);
+        assertEquals(cacheManager.getSize(), 9);
 
         // Next insert should trigger a cache eviction to force the size to 8
         // The algorithm should evict entries from cache1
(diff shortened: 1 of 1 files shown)


managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/EntryCacheManagerTest.java
lhotari, line 1: The way how eviction happens was changed in #24363.
To address the test flakiness, the test should control the evictions. Here's an example from another test case:
https://github.com/apache/pulsar/blob/8ce67b97b3e428eac02dcfe0bb41d698deb78fd4/pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/PrometheusMetricsTest.java#L694-L696

lhotari (changes requested):

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/EntryCacheManagerTest.java
3pacccccc: thank you for your suggestions @lhotari 
But I think The root cause of this test flakiness stems from incorrect assumptions about eviction timing. Currently, eviction occurs immediately after inserting ```EntryImpl.create(2, 2, new byte[1])```, but the test incorrectly assumes eviction happens during the next insertion.

The test's original intention appears to be verifying when eviction gets triggered, but it mispredicts the exact trigger point. I think we don't need to  manually blocking evictions.

Or Maybe I don't get your point, Would you like to  show me how to properly deal with it?

lhotari: Oh yes, that's correct. There's a hard coded value that starts eviction when the total cache size is more than 98% consumed, https://github.com/apache/pulsar/blob/dedba1a9bf0becb1b3f80f1d89a779451734fbe6/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheManagerImpl.java#L55-L66

Issue #23263, opened 2024-09-06
[Bug] Permit calculations are mixing individual message and batch message counts, similar issue in documentation


lhotari:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [X] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

master

### Minimal reproduce step

Fixing some issues as part of #23231 such as
https://github.com/apache/pulsar/blob/14395295b4996dcfb7eac288d92baf1104c9c576/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L790-L791

The problem is that permits use a batch message count and unack messages uses individual message counts. These shouldn't be mixed. It's possible to convert the batch message count to an estimated number of messages with the average number of messages in batch value that is kept in the consumer.

There are several locations where this is mixed up.

This also applies to documentation. The javadoc of receiverQueueSize is very confusing and doesn't tell that it's counted in batch messages: https://github.com/apache/pulsar/blob/10f4e0248f0f985b1dc7ad38970c906b7fe629be/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java#L356-L387

### What did you expect to see?

The permit calculations in dispatchers shouldn't mix individual message counts and batch message counts. 
The documentation needs updates too so that it's clearly defined whether a "message" refers to an individual message or a batch message (a batch of individual messages).

### What did you see instead?

- calculations mixed in several locations
- documentation doesn't clearly define whether a message is an individual message or a batch of messages. example is the receiverQueueSize javadoc.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
I recently found out that the unit for permits in dispatching depends on the `preciseDispatcherFlowControl` setting:

https://github.com/apache/pulsar/blob/0a2ffe4743799dc253814d49fa3fd29b933444ac/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L546-L552

[...]

Issue #11850, opened 2021-08-31
Java Client not allow newConsumer().receiverQueueSize(0)


graceon:
My Code
```
import org.apache.pulsar.client.api.*;
import org.apache.pulsar.client.impl.PulsarClientImpl;
import org.apache.pulsar.client.impl.conf.ClientConfigurationData;

import java.nio.charset.StandardCharsets;

class ConsumerThread extends Thread{
    private final long processTime;
    private final String name;
    //    private final ConsumerListener consumerListener;
    Consumer<byte[]> consumer;
    public ConsumerThread(String name,long processTime) throws PulsarClientException {
        this.processTime = processTime;
        this.name = name;
//        consumerListener = new ConsumerListener(name,processTime);
        PulsarClient pulsarClient = PulsarClient.builder()
                .serviceUrl("pulsar://192.168.184.131:6650")
                .build();
        consumer = pulsarClient.newConsumer()
                .receiverQueueSize(1)
//                .batchReceivePolicy()
                .topic("my_topic")
                .subscriptionName("my_subscription")
                .subscriptionType(SubscriptionType.Shared)
                .subscribe();
    }
    @Override
    public void run(){
        while (true){
            try {
                Message<byte []>message = consumer.receive();
                Thread.sleep(processTime);
                System.out.println(name + ":" + new String(message.getData()));
                consumer.acknowledge(message);
            } catch (PulsarClientException | InterruptedException e) {
                e.printStackTrace();
            }
        }
    }
}

public class StartManager {
    public static void main(String[] args) throws PulsarClientException, InterruptedException {
        PulsarClient pulsarClient = PulsarClient.builder()
                .serviceUrl("pulsar://192.168.184.131:6650")
                .build();
        new ConsumerThread("a",100).start();
        new ConsumerThread("      b",1000).start();

        Producer<byte[]> producer = pulsarClient.newProducer()
                .topic("my_topic").create();
        for (int i = 0;i < 50;i++){
            producer.send(String.valueOf(i).getBytes(StandardCharsets.UTF_8));
        }
        while (true){


            Thread.sleep(1000);
        }
    }
}
```
debug it run into
```
# ZeroQueueConsumerImpl.class
    void receiveIndividualMessagesFromBatch(MessageMetadata msgMetadata, int redeliveryCount, List<Long> ackSet, ByteBuf uncompressedPayload, MessageIdData messageId, ClientCnx cnx) {
[...]

PR #24610, opened 2025-08-05
[fix][client]Prevent ZeroQueueConsumer from receiving batch messages when using MessagePayloadProcessor

3pacccccc:
### Motivation
Currently, `ZeroQueueConsumerImpl` with `MessagePayloadProcessor` enabled can incorrectly receive batch messages when it shouldn't. This occurs because:

1. The message payload processor processes batch messages before checking if the consumer is a zero-queue consumer
2. This behavior contradicts the expected behavior of zero-queue consumers which should not buffer any messages

The issue can be reproduced when:
- Using a `MessagePayloadProcessor` that processes batch messages

### Modifications
1. Added early check in `processPayloadByProcessor` to detect zero-queue consumers receiving batch messages
2. For zero-queue consumers with batch messages, bypass the payload processor and directly use `receiveIndividualMessagesFromBatch`
3. Added new test case `testZeroQueueSizeConsumerWithPayloadProcessorReceiveBatchMessage` to verify the fix

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/20

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1396,7 +1396,7 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         final AtomicInteger skippedMessages = new AtomicInteger(0);
         if (this instanceof ZeroQueueConsumerImpl<T> && entryContext.isBatch()) {
             this.receiveIndividualMessagesFromBatch(brokerEntryMetadata,
-                messageMetadata,redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
+                messageMetadata, redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
             return;
         }
         try {
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1394,6 +1394,11 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         final MessagePayloadContextImpl entryContext = MessagePayloadContextImpl.get(
                 brokerEntryMetadata, messageMetadata, messageId, this, redeliveryCount, ackSet, consumerEpoch);
         final AtomicInteger skippedMessages = new AtomicInteger(0);
+        if (this instanceof ZeroQueueConsumerImpl<T> && entryContext.isBatch()) {
+            this.receiveIndividualMessagesFromBatch(brokerEntryMetadata,
+                    messageMetadata,redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
+            return;
+        }
         try {
             conf.getPayloadProcessor().process(payload, entryContext, schema, message -> {
                 if (message != null) {
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1396,7 +1396,7 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         final AtomicInteger skippedMessages = new AtomicInteger(0);
         if (this instanceof ZeroQueueConsumerImpl<T> && entryContext.isBatch()) {
             this.receiveIndividualMessagesFromBatch(brokerEntryMetadata,
-                    messageMetadata,redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
+                messageMetadata,redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
             return;
         }
         try {
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ZeroQueueSizeTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ZeroQueueSizeTest.java
@@ -605,12 +605,15 @@ public class ZeroQueueSizeTest extends BrokerTestBase {
                 .messagePayloadProcessor(new MessagePayloadProcessor() {
                                              @Override
                                              public <T> void process(MessagePayload payload,
-                                                                     MessagePayloadContext context, Schema<T> schema,
-                                                                     java.util.function.Consumer<Message<T>> messageConsumer) throws Exception {
+                                                             MessagePayloadContext context, Schema<T> schema,
+                                                             java.util.function.Consumer<Message<T>> messageConsumer) {
                                                  if (context.isBatch()) {
                                                      final int numMessages = context.getNumMessages();
                                                      for (int i = 0; i < numMessages; i++) {
-                                                         messageConsumer.accept(context.getMessageAt(i, numMessages, payload, true, schema));
+                                                         messageConsumer.
+                                                                 accept(context.getMessageAt(i, numMessages,
+                                                                         payload, true, schema)
+                                                                 );
                                                      }
                                                  } else {
                                                      messageConsumer.accept(context.asSingleMessage(payload, schema));
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ZeroQueueSizeTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ZeroQueueSizeTest.java
@@ -38,7 +38,11 @@ import org.apache.pulsar.broker.service.BrokerTestBase;
 import org.apache.pulsar.client.admin.PulsarAdminException;
 import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.Message;
+import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageListener;
+import org.apache.pulsar.client.api.MessagePayload;
+import org.apache.pulsar.client.api.MessagePayloadContext;
+import org.apache.pulsar.client.api.MessagePayloadProcessor;
 import org.apache.pulsar.client.api.MessageRoutingMode;
 import org.apache.pulsar.client.api.Producer;
 import org.apache.pulsar.client.api.ProducerBuilder;
@@ -363,6 +367,7 @@ public class ZeroQueueSizeTest extends BrokerTestBase {
             .ackTimeout(1, TimeUnit.SECONDS)
             .messageListener((MessageListener<Integer>) (c, msg) -> {
                 try {
+                    System.out.println("----receive data : " + msg.getData());
                     receivedMessages.add(msg.getValue());
                 } finally {
                     latch.countDown();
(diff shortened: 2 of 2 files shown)


lhotari (changes requested):
The motivation isn't very clear yet.

> Currently, `ZeroQueueConsumerImpl` with `MessagePayloadProcessor` enabled can incorrectly receive batch messages when it shouldn't. This occurs because:
> 
> 1. The message payload processor processes batch messages before checking if the consumer is a zero-queue consumer
> 2. This behavior contradicts the expected behavior of zero-queue consumers which should not buffer any messages

Answering these questions would help:

1. Please elaborate more on why you think that this contradicts expected behavior of zero-queue consumers?
2. What do you consider as "buffering"?
3. What is your use case?

3pacccccc:
Thanks for your review @lhotari . Here's my answer to your questions:
> 1.Please elaborate more on why you think that this contradicts expected behavior of zero-queue consumers?

The current behavior with MessagePayloadProcessor creates an unpredictable interaction pattern. Here's a detailed breakdown of what happens:
```java
Producer<byte[]> producer = pulsarClient.newProducer().topic(topicName)
        .batchingMaxMessages(5)
        .enableBatching(true)
        .create();

ConsumerImpl<byte[]> consumer = (ConsumerImpl<byte[]>) pulsarClient.newConsumer()
        .topic(topicName)
        .messagePayloadProcessor(new CustomProcessor())
        .subscriptionName(subscriptionName)
        .receiverQueueSize(0)
        .subscribe();

for (int i = 0; i < totalMessages; i++) {
    String message = messagePredicate + i;
    producer.sendAsync(message.getBytes());
}
producer.flush();
for (int i = 0; i < 10; i++) {
    Message<byte[]> receive = consumer.receive();
}
```

In this case, here's the step of what will happen:

| consumer                                                                                                                                                                                                                                                                               | broker                                                                              |
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------|
| receive()  called                                                                                                                                                                                                                                                                      |                                                                                     |
| send 1 flow-permits to broker<sup>[1]</sup>                                                                                                                                                                                                                                            |                                                                                     |
| Blocks waiting for messages                                                                                                                                                                                                                                                            |                                                                                     |
|                                                                                                                                                                                                                                                                                        | receive flow-permits, and send 5 messages to broker, set ```messagePermits``` as -4 |
| receive 5 messages in a batch                                                                                                                                                                                                                                                          |                                                                                     |
| processed batch via ```processPayloadByProcessor```<sup>[2]</sup>                                                                                                                                                                                                                      |                                                                                     |
| put messages into ```incomingMessages``` queue                                                                                                                                                                                                                                         |                                                                                     |
| consume 1 message.                                                                                                                                                                                                                                                                     |                                                                                     |
| trigger one more receive()                                                                                                                                                                                                                                                             |                                                                                     |
| find  messages in the incomingMessages, maybe 1 message or more, maybe no message, totally depends on the process speed of user's customize ```processPayloadByProcessor```<sup>[2]</sup>, let's assume there's 4 message's in the ```incomingQueue```, release them all<sup>[1]</sup> |                                                                                     |
| send 1 flow-permits to broker                                                                                                                                                                                                                                                          |                                                                                     |
| block and waiting for new messages coming                                                                                                                                                                                                                                              |                                                                                     |
|                                                                                                                                                                                                                                                                                        | receive flow-permits, don't send message, set ```messagePermits``` as -3            |

[1]:https://github.com/apache/pulsar/blob/994b1bea110ad1fd284c87d6fee3a3ecfa4ea010/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java#L96
[2]:https://github.com/apache/pulsar/blob/885bf8707e76d1f7d0a25bb5e5f899ed57e1ff32/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1387
so, as you can see, this ```zeroQueueConsumer``` will be blocked indefinitely in this case after consume 1 message. and maybe after consume 2 messages or more, and even won't get blocked, this totally depends on the process speed of user's customize ```MessagePayloadProcessor```.

**Key Problems**:
- Permit Mismatch: The broker sends more messages (batches) than requested

- Hidden Buffering: Messages accumulate in incomingMessages despite queueSize=0

- Unpredictable Blocking: Behavior depends on processor speed, violating zero-queue guarantees

> 2.What do you consider as "buffering"?

In my understanding, the fundamental design principle of ```ZeroQueueConsumer``` (receiverQueueSize=0) is to be a **strictly non-buffering consumer** that:

- Operates in pure request-response mode:

  - Each receive() call should trigger exactly one network request to the broker

  -  No messages should be retained client-side between receive() calls
> 3.What is your use case?

We need ZeroQueueConsumer to:

1.**Maintain Strict Pull Semantics**: Each receive() should trigger exactly one broker interaction

2.**Handle Batches Predictably**: Either:Reject batches entirely, or Process them atomically without intermediate buffering

lhotari:
> In my understanding, the fundamental design principle of `ZeroQueueConsumer` (receiverQueueSize=0) is to be a **strictly non-buffering consumer** that:
> 
> * Operates in pure request-response mode:
>   
>   * Each receive() call should trigger exactly one network request to the broker
>   * No messages should be retained client-side between receive() calls

I don't think that this is a correct assumption. In Pulsar, the receive method returns individual messages in a batch.
When the broker sends a batch message to the client, it should be buffered on the client side, also with ZeroQueueConsumer. 

It's very much possible that you have faced bugs in the ZeroQueueConsumerImpl implementation. When checking at the implementation, it doesn't seem to make a distinction between batch messages. In Pulsar code, it's bit of a mess that in some cases the permits are about batch messages and some times there are about individual messages. I have created an issue about that in https://github.com/apache/pulsar/issues/23263 . It's possible that the information in the issue isn't accurate.

3pacccccc:
> > In my understanding, the fundamental design principle of `ZeroQueueConsumer` (receiverQueueSize=0) is to be a **strictly non-buffering consumer** that:
> > 
> > * Operates in pure request-response mode:
> >   
> >   * Each receive() call should trigger exactly one network request to the broker
> >   * No messages should be retained client-side between receive() calls
> 
> I don't think that this is a correct assumption. In Pulsar, the receive method returns individual messages in a batch. When the broker sends a batch message to the client, it should be buffered on the client side, also with ZeroQueueConsumer.
> 
> It's very much possible that you have faced bugs in the ZeroQueueConsumerImpl implementation. When checking at the implementation, it doesn't seem to make a distinction between batch messages. In Pulsar code, it's bit of a mess that in some cases the permits are about batch messages and some times there are about individual messages. I have created an issue about that in #23263 . It's possible that the information in the issue isn't accurate.

you can reproduce this issue with the example I mentioned above. so, I'm wondering is it ok that consumer gets blocked indefinitely?

lhotari:
I searched for ZeroQueueConsumerImpl related issues.

Now I finally understand the motivation. It appears that ZeroQueueConsumerImpl isn't intended to be used with batch messages at all, which is a bit confusing. The issue #11850 hints that. It refers this method:
https://github.com/apache/pulsar/blob/dedba1a9bf0becb1b3f80f1d89a779451734fbe6/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java#L194-L211
Which is the same method that you are calling.

I wasn't aware of this. I'll review the changes again.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1394,6 +1394,11 @@ private void processPayloadByProcessor(final BrokerEntryMetadata brokerEntryMeta
         final MessagePayloadContextImpl entryContext = MessagePayloadContextImpl.get(
                 brokerEntryMetadata, messageMetadata, messageId, this, redeliveryCount, ackSet, consumerEpoch);
         final AtomicInteger skippedMessages = new AtomicInteger(0);
+        if (this instanceof ZeroQueueConsumerImpl<T> && entryContext.isBatch()) {
+            this.receiveIndividualMessagesFromBatch(brokerEntryMetadata,
+                messageMetadata, redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
+            return;
+        }
```
lhotari: instead of modifying `ConsumerImpl`, modify the `processPayloadByProcessor` method to be a protected method so that it can be overridden in `ZeroQueueConsumerImpl`. 

in the overridden method in `ZeroQueueConsumerImpl`, you can perform the check `if (!entryContext.isBatch()) {` and call the super method in the `else` block you can do the error handling. Instead of calling the `receiveIndividualMessagesFromBatch` method directly, extract the method for error handling and call it.

I think that this way circular dependencies between ZeroQueueConsumerImpl and ConsumerImpl would be avoided.

3pacccccc: Thank you so much, this is a nich suggestion. I'll update soon

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java
```diff
@@ -215,4 +215,18 @@ protected void setCurrentReceiverQueueSize(int newSize) {
...
     }
+
+    @Override
+    protected void processPayloadByProcessor(BrokerEntryMetadata brokerEntryMetadata,
+                                             MessageMetadata messageMetadata, ByteBuf byteBuf,
+                                             MessageIdImpl messageId, Schema<T> schema,
+                                             int redeliveryCount, List<Long> ackSet, long consumerEpoch) {
+        if (this.isBatch(messageMetadata)) {
+            this.receiveIndividualMessagesFromBatch(brokerEntryMetadata,
+                    messageMetadata, redeliveryCount, ackSet, byteBuf, null, null, consumerEpoch, false);
```
lhotari: refactor the receiveIndividualMessagesFromBatch method and extract an error handling method instead of calling receiveIndividualMessagesFromBatch here. The call to receiveIndividualMessagesFromBatch was causing the confusion in the initial version of the code and this problem would be resolved by extracting the error handling method. It could be called `rejectBatchMessageByClosingConsumer` etc.

lhotari:
please add javadoc to ConsumerBuilder.messagePayloadProcessor about the behavior with receiverQueueSize=0 (don't mention ZeroQueueConsumerImpl in ConsumerBuilder javadoc since it is an implementation detail):
https://github.com/apache/pulsar/blob/dedba1a9bf0becb1b3f80f1d89a779451734fbe6/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java#L821-L827

This is the receiverQueueSize javadoc:
https://github.com/apache/pulsar/blob/dedba1a9bf0becb1b3f80f1d89a779451734fbe6/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java#L377-L408

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java
3pacccccc: thanks for the suggestion, much better handling! @lhotari

lhotari (approved):
LGTM

3pacccccc:
@lhotari I've added javadoc, please take a look again.

lhotari:
Btw. Not directly related to this PR, but could you check if there's any use for the PulsarClientException.InvalidMessageException that is thrown? Does it get used at all? I think that we have made changes to closing of the consumer, that it will fail all pending operations with PulsarClientException.AlreadyClosedException.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java
```diff
@@ -196,6 +196,29 @@ void receiveIndividualMessagesFromBatch(BrokerEntryMetadata brokerEntryMetadata,
...
+                                             int redeliveryCount, List<Long> ackSet, long consumerEpoch) {
+        if (this.isBatch(messageMetadata)) {
+            rejectBatchMessageByClosingConsumer();
+        } else {
+            super.processPayloadByProcessor(brokerEntryMetadata, messageMetadata, byteBuf, messageId, schema,
+                    redeliveryCount, ackSet, consumerEpoch);
+        }
+    }
+
+    private void rejectBatchMessageByClosingConsumer() {
```
lhotari: Please pass the message id to this method so that it could be logged. improve the warning log message so that the message id is logged. This would help investigate the issue if someone happens to encounter this problem.

3pacccccc: added

3pacccccc:
> Btw. Not directly related to this PR, but could you check if there's any use for the PulsarClientException.InvalidMessageException that is thrown? Does it get used at all? I think that we have made changes to closing of the consumer, that it will fail all pending operations with PulsarClientException.AlreadyClosedException.

@lhotari OK, I'm glad to do that

lhotari:
> > Btw. Not directly related to this PR, but could you check if there's any use for the PulsarClientException.InvalidMessageException that is thrown? Does it get used at all? I think that we have made changes to closing of the consumer, that it will fail all pending operations with PulsarClientException.AlreadyClosedException.
> 
> @lhotari OK, I'm glad to do that

I'll merge this PR so that could be done separately

lhotari (approved):

3pacccccc:
> Btw. Not directly related to this PR, but could you check if there's any use for the PulsarClientException.InvalidMessageException that is thrown? Does it get used at all? I think that we have made changes to closing of the consumer, that it will fail all pending operations with PulsarClientException.AlreadyClosedException.

Hi, lari, @lhotari 
I've checked all the InvalidMessageException usages:

- mostly on the producer side (which isn't related to our current changes)

- few on the consumer side (mainly for message validation purposes)

The current usage looks correct and doesn't need modification. However, I did notice and fix some inaccurate javadocs - please take a look when you have time.

3pacccccc:
@Technoboy- @lhotari 
Hi Jiwei, Lari. I think this PR might needs to be cherry-picked to branch3.x and 4.x because:
1.The issue fixed by this PR also exists in branch3.x and 4.x
2.The javadoc behavior described in this PR is also present in branch3.x and 4.x"

PR #24618, opened 2025-08-11
[fix][broker] fix bk configuration bookkeeperMetadataServiceUri 

gaozhangmin:
### Motivation
The issue occurs when configuring 
bookkeeperMetadataServiceUri=metadata-store:zk:zk1:2181/ledgers along with enabling:

text
-Dbookkeeper.metadata.bookie.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataBookieDriver  
-Dbookkeeper.metadata.client.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataClientDriver  
Root Cause Analysis
The problem arises in AbstractMetadataDriver when creating MetadataStoreExtended. Specifically:

The resolveLedgersRootPath() method has a bug in path parsing.

When using ZooKeeper (ZK), the connection string typically includes a double slash (//) (e.g., zk://metadata-store:2181/ledgers), which ensures URI.getPath() correctly returns /ledgers.

However, Pulsar’s metadata service connection string (metadata-store:zk:metadata-store:2181/ledgers) lacks the // prefix, causing URI.getPath() to incorrectly parse the ledger root path, defaulting to /ledgers regardless of the configured path.

```
URI metadataServiceUri = URI.create(metadataServiceUriStr);
String path = metadataServiceUri.getPath();
return path == null ? BookKeeperConstants.DEFAULT_ZK_LEDGERS_ROOT_PATH : path;
```

### Modifications
The fix should ensure proper path resolution regardless of the connection string format. Possible approaches:

Normalize the connection string before parsing (e.g., ensure // is present for consistency).

Improve resolveLedgersRootPath() to handle both formats (with or without //).

Example: Extract the path segment after the last : or / if the URI parsing fails.

Explicitly enforce a default /ledgers fallback only when path extraction fails.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:  https://github.com/gaozhangmin/pulsar/tree/fix_resolveLedgersRootPath

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1031,6 +1031,8 @@ maxConcurrentHttpRequests=1024
 # For example: zk+hierarchical://localhost:2181/ledgers
 # The metadata service uri list can also be semicolon separated values like below:
 # zk+hierarchical://zk1:2181;zk2:2181;zk3:2181/ledgers
+# or
+# metadata-store:zk:localhost:2181/ledgers If we set -Dbookkeeper.metadata.bookie.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataBookieDriver -Dbookkeeper.metadata.client.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataClientDriver
 bookkeeperMetadataServiceUri=
 
 # Authentication plugin to use when connecting to bookies
--- a/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java
@@ -306,7 +306,8 @@ public class PulsarClusterMetadataSetup {
         if (arguments.existingBkMetadataServiceUri == null && arguments.bookieMetadataServiceUri == null) {
             ServerConfiguration bkConf = new ServerConfiguration();
             bkConf.setListDelimiterHandler(new DisabledListDelimiterHandler());
-            bkConf.setMetadataServiceUri("metadata-store:" + arguments.metadataStoreUrl);
+            bkConf.setMetadataServiceUri("metadata-store:" + arguments.metadataStoreUrl
+                    + BookKeeperConstants.DEFAULT_ZK_LEDGERS_ROOT_PATH);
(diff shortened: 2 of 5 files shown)


lhotari (commented):
Thanks, @gaozhangmin. Please add tests to cover the bug and the fix.

zhannao456:
To avoid the path resolution bug in AbstractMetadataDriver, we should remove /ledgers from the bookkeeperMetadataServiceUri configuration .  Since that, this pr wouldn't be needed any more.

But I found there is another problem, the default zktimeout of bk is 10s, but 
We caculate tickTimeMillis  =  zktimeout/15. 

https://github.com/apache/pulsar/blob/277835a65dd586de941b7d0ad448843b7b5589a4/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L63

In below future.get, always timeout.
https://github.com/apache/pulsar/blob/277835a65dd586de941b7d0ad448843b7b5589a4/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L108-L112

We should support set bk zktimeut by 
 bkConf.setZkTimeout((int) conf.getMetadataStoreSessionTimeoutMillis());

gaozhangmin:
I found there is another problem related to this, the default zktimeout of bk is 10s, but 
We caculate tickTimeMillis  =  zktimeout/15. 

https://github.com/apache/pulsar/blob/277835a65dd586de941b7d0ad448843b7b5589a4/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L63

In below future.get, always timeout.
https://github.com/apache/pulsar/blob/277835a65dd586de941b7d0ad448843b7b5589a4/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L108-L112

We should support set bk zktimeut by 
 bkConf.setZkTimeout((int) conf.getMetadataStoreSessionTimeoutMillis());

PR #24622, opened 2025-08-12
[improve][broker]Find the target position at most once, during expiring messages for a topic, even though there are many subscriptions

poorbarcode:
#### 1. Instead of finding the target position for each subscription, search the target position once, and expire messages for each subscription.
**Background**: The steps of expiry of a topic
- Call expiry for each subscription
  - The subscription finds the target position, then calls mark deleting for the cursor
- Call expiry for each replicator
  - The replicator finds the target position, then calls mark deleting for the cursor

So the task calls for finding the target position many times when a topic expires. But the finding position operation is really expensive

#### 2. Improve the method `expireMessages(Position)`

The previous steps of expiring messages by position:
- Search the target position with a binary search
- Expire messages by the target position.

It costs a lot, we can simplify the steps by `managedLedger.getPreviousPosition(x)`

### Modifications

Since all of the tasks that find the target position that relates to the same topic will get the same result, we can call the finding operation at most once, during expiring messages for a topic

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -2123,8 +2123,8 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             return;
         }
 
-        // If managed ledger is not instance of ManagedLedgerImpl, each subscription find position and handle expiring
-        // itself.
+        // Fallback to the slower solution if managed ledger is not an instance of ManagedLedgerImpl: each
+        // subscription find position and handle expiring itself.
         ManagedLedger managedLedger = getManagedLedger();
         if (!(managedLedger instanceof ManagedLedgerImpl ml)) {
             subscriptionsCheckMessageExpiryEachOther(messageTtlInSeconds);
@@ -2133,10 +2133,8 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
 
         // Find the target position at one time, then expire all subscriptions and replicators.
         ManagedCursor cursor = ml.getCursors().getCursorWithOldestPosition().getCursor();
-        // TODO reuse finder.
         PersistentMessageFinder finder = new PersistentMessageFinder(topic, cursor, brokerService.getPulsar()
                 .getConfig().getManagedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis());
-        // TODO check the time calculated is correct.
         // Find the target position.
         long expiredMessageTimestamp = System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(messageTtlInSeconds);
         CompletableFuture<Position> positionToMarkDelete = new CompletableFuture<>();
@@ -2149,7 +2147,10 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             @Override
             public void findEntryFailed(ManagedLedgerException exception, Optional<Position> failedReadPosition,
                                         Object ctx) {
-                positionToMarkDelete.completeExceptionally(exception);
+                log.error("[{}] Error finding expired position, failed reading position is {}", topic,
+                        failedReadPosition.orElse(null), exception);
+                // Since we have logged the error, we can skip to print error log at next step.
+                positionToMarkDelete.complete(null);
             }
         });
         positionToMarkDelete.thenAccept(position -> {
@@ -2169,6 +2170,9 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
                     -> ((PersistentReplicator) replicator).expireMessages(position));
             shadowReplicators.forEach((__, replicator)
                     -> ((PersistentReplicator) replicator).expireMessages(position));
+        }).exceptionally(ex -> {
+            log.error("[{}] Failed to expire messages by position", topic, ex);
+            return null;
         });
     }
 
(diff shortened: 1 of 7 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -2118,19 +2119,75 @@ private CompletableFuture<Void> checkShadowReplication() {
...
+                return;
+            }
+            // Expire messages by position, which is more efficient.
             subscriptions.forEach((__, sub) -> {
                 if (!isCompactionSubscription(sub.getName())
                         && (additionalSystemCursorNames.isEmpty()
-                            || !additionalSystemCursorNames.contains(sub.getName()))) {
-                   sub.expireMessagesAsync(messageTtlInSeconds);
+                        || !additionalSystemCursorNames.contains(sub.getName()))) {
+                    sub.expireMessages(position);
```
codelipenghui, line 2187: I noticed expire message by position will try to get the previous position of `position`. It seems a behavior change with the current implementation?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -2118,19 +2119,75 @@ private CompletableFuture<Void> checkShadowReplication() {
...
-                    -> ((PersistentReplicator) replicator).expireMessagesAsync(messageTtlInSeconds));
-        }
+                    -> ((PersistentReplicator) replicator).expireMessages(position));
+        }).exceptionally(ex -> {
+            log.error("[{}] Failed to expire messages by position", topic, ex);
+            return null;
+        });
+    }
+
+    private void subscriptionsCheckMessageExpiryEachOther(int messageTtlInSeconds) {
```
codelipenghui: I would like suggest to have 2 methods

- checkMessageExpiryWithoutSharedPosition()
- checkMessageExpiryWithSharedPosition()

So that checkMessageExpiry() method can call above 2 methods, which should be better for code organization?

pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/PersistentMessageExpiryMonitorTest.java
```diff
@@ -138,4 +147,52 @@ void testConcurrentlyExpireMessages() throws Exception {
...
+            }
+            return invocationOnMock.callRealMethod();
+        }).when(spyReadHandle).readAsync(anyLong(), anyLong());
+        ml.ledgerCache.put(firstLedger, CompletableFuture.completedFuture(spyReadHandle));
+        // Verify: the first ledger will be accessed only once after expiry for two subscriptions.
+        PersistentTopic persistentTopic =
+                (PersistentTopic) pulsar.getBrokerService().getTopic(topicName, false).join().get();
+        persistentTopic.checkMessageExpiry();
+        Thread.sleep(2000);
+        assertEquals(1, accessedCount.get());
```
codelipenghui, line 196: Is it better to use Awaitibility instead of Thread.sleep?

pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/PersistentMessageExpiryMonitorTest.java
```diff
@@ -138,4 +147,52 @@ void testConcurrentlyExpireMessages() throws Exception {
...
+
+        // cleanup.
+        producer.close();
+        admin.topics().delete(topicName);
```
codelipenghui, line 200: Interesting.. This operation should get failed since there are subscription still exist?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: The logic are the same, see: https://github.com/apache/pulsar/pull/24622/files#diff-813cddfda73ce8df0722a7c83ac1cb8d6d3db62ea13b44b5dc1bc438010fdd40R141-R181

poorbarcode: I also added the section `2. Improve the method expireMessages(Position)` into the motivation

pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/PersistentMessageExpiryMonitorTest.java
poorbarcode: No, the topic deletion only checks whether there are consumers connected. It does not care about how many subscriptions exist

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: Done

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -2118,19 +2119,78 @@ private CompletableFuture<Void> checkShadowReplication() {
...
+        }
+
+        ManagedLedger managedLedger = getManagedLedger();
+        if (managedLedger instanceof ManagedLedgerImpl ml) {
+            checkMessageExpiryWithSharedPosition(ml, messageTtlInSeconds);
+        } else {
+            // Fallback to the slower solution if managed ledger is not an instance of ManagedLedgerImpl: each
+            // subscription find position and handle expiring itself.
+            checkMessageExpiryWithoutSharedPosition(messageTtlInSeconds);
+        }
```
coderzc, line 2133: Besides ManagedLedgerImpl, do we have any other implementations?

poorbarcode: NO, we do not have

coderzc (approved):

PR #24624, opened 2025-08-13
[fix][broker] fix CheckConnectionStatus  disconnected 

gaozhangmin:
### Motivation

I found there is problem related to bk's zktimeout config, the default bk zktimeout of bk is 10s, but 
We caculate tickTimeMillis  =  zktimeout/15. 

https://github.com/apache/pulsar/blob/277835a65dd586de941b7d0ad448843b7b5589a4/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L63

In below future.get, always timeout.
https://github.com/apache/pulsar/blob/277835a65dd586de941b7d0ad448843b7b5589a4/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKSessionWatcher.java#L108-L112

We should support set bk zktimeut by 
 bkConf.setZkTimeout((int) conf.getMetadataStoreSessionTimeoutMillis());

### Modifications

bkConf.setZkTimeout((int) conf.getMetadataStoreSessionTimeoutMillis());

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1028,9 +1028,11 @@ maxConcurrentHttpRequests=1024
 # Metadata service uri that bookkeeper is used for loading corresponding metadata driver
 # and resolving its metadata service location.
 # This value can be fetched using `bookkeeper shell whatisinstanceid` command in BookKeeper cluster.
-# For example: zk+hierarchical://localhost:2181/ledgers
+# For example: zk+hierarchical://localhost:2181
 # The metadata service uri list can also be semicolon separated values like below:
-# zk+hierarchical://zk1:2181;zk2:2181;zk3:2181/ledgers
+# zk+hierarchical://zk1:2181;zk2:2181;zk3:2181
+# or
+# metadata-store:zk:localhost:2181  If we set -Dbookkeeper.metadata.bookie.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataBookieDriver -Dbookkeeper.metadata.client.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataClientDriver
 bookkeeperMetadataServiceUri=
 
 # Authentication plugin to use when connecting to bookies
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1028,9 +1028,9 @@ maxConcurrentHttpRequests=1024
 # Metadata service uri that bookkeeper is used for loading corresponding metadata driver
 # and resolving its metadata service location.
 # This value can be fetched using `bookkeeper shell whatisinstanceid` command in BookKeeper cluster.
-# For example: zk+hierarchical://localhost:2181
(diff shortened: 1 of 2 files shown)


conf/broker.conf
```diff
@@ -1031,6 +1031,8 @@ maxConcurrentHttpRequests=1024
 # For example: zk+hierarchical://localhost:2181/ledgers
 # The metadata service uri list can also be semicolon separated values like below:
 # zk+hierarchical://zk1:2181;zk2:2181;zk3:2181/ledgers
+# or
+# metadata-store:zk:localhost:2181  If we set -Dbookkeeper.metadata.bookie.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataBookieDriver -Dbookkeeper.metadata.client.drivers=org.apache.pulsar.metadata.bookkeeper.PulsarMetadataClientDriver
```
lhotari: Since `bin/pulsar` will set these system properties unless `BK_METADATA_OPTIONS` is set to `none`, is this comment about the system properties necessary? It might be better to update the complete documentation for `bookkeeperMetadataServiceUri` to match the default settings. 

https://github.com/apache/pulsar/blob/bbf56f651ebe357745b46654a1d9fb7dc4899fd7/bin/pulsar#L349-L355

lhotari: @gaozhangmin ping

lhotari:
@gaozhangmin During a test related PR #24171 I noticed this about the same code location:

> This will trigger a ConnectionLost session event. That's why a delay should be kept under 2000 ms.
> Currently MockZooKeeper's session timeout is hard coded to 30000ms and the check interval is 1/15 of this, therefore 2000ms.
> 
> When there's a test that blocks the ZooKeeper thread for more than 2000ms, it will result in a ConnectionLost event.
> This PR makes the session timeout configurable and it's possible to increase the session timeout for tests where it's necessary to use a longer session timeout to avoid ConnectionLost session events.

This applies likewise to production code. This is problematic since the effective timeout becomes 1/15 of the zk session timeout for Pulsar operations.

Together with this change made in #23018 it will add significant load to ZooKeepers:
https://github.com/apache/pulsar/blob/829df71901de726a11bb3c8ee934d14b67ee9694/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java#L513-L519

I'll file a new issue about resolving this problem.

lhotari:
Issue about ZKSessionWatcher is #24674

Technoboy- (approved):

PR #21798, opened 2023-12-24
[improve][broker] PIP-315: Configurable max delay limit for delayed delivery

KevinLiLu:
PIP-315: Configurable max delay limit for delayed delivery (https://github.com/apache/pulsar/blob/master/pip/pip-315.md)

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #17907, opened 2022-09-30
Allow to configure and disable the size of lookahead for detecting fixed delays in messages

merlimat:
### Motivation

Allow to configure and disable the size of lookahead for detecting fixed delays in messages

### Documentation

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)

- [X] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

### Matching PR in forked repository

PR in forked repository: [<!-- ENTER URL HERE ](https://github.com/merlimat/pulsar/pull/2)

PR #24625, opened 2025-08-13
[improve][pip] PIP-437: Granular and Fixed-Delay Policies for Message Delivery

chrisqwang3:
PIP: 437

### Motivation
The current system lacks administrative controls on message delivery delays, which introduces risks to cluster stability and data integrity. This proposal aims to provide granular control at the topic and namespace levels to add a fixed delay delivery configuration 

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-437.md
+++ b/pip/pip-437.md
@@ -138,5 +138,5 @@ If there are alternatives that were already considered by the authors or, after 
 <!--
 Updated afterwards
 -->
-* Mailing List discussion thread:https://lists.apache.org/thread/23t4zzbfjrm5r4pbzrofn3d85c0171yn
-* Mailing List voting thread:
+* Mailing List discussion thread: https://lists.apache.org/thread/23t4zzbfjrm5r4pbzrofn3d85c0171yn
+* Mailing List voting thread: https://lists.apache.org/thread/m8y5brc0o8fgcby1sop3hybjpm46xgrr
--- a/pip/pip-437.md
+++ b/pip/pip-437.md
@@ -1,0 +1,167 @@
+# PIP-437: Granular Control for Message Delivery Delays
+
+# Background knowledge
+Pulsar currently allows producers to specify deliverAt timestamps for delayed message delivery.
+However, there's no mechanism to constrain how far into the future a message can be scheduled. With each producer
+scheduling various delivery time it makes the system less predictable and bring challenges for support.
+
+# Motivation
+The current system lacks administrative controls on message delivery delays, which introduces risks to cluster stability and data integrity.
+
+## Risks in the current system:
+Unbounded Resource Pressure: Excessively delayed messages must be retained in storage for extended periods.
+More importantly, they create a long-lived state in the broker's memory that cannot be cleared,
+leading to unpredictable resource consumption and potential garbage collection pressure
+
+### The Limitation:
+To protect broker memory, this tracker is capped at a configurable fixed number of entries (e.g., 10,000).
+A large volume of messages with widely varying delivery times can easily exhaust this capacity.
+Once the tracker is full, it stops persisting the index for new delayed messages, retaining this state only in memory.
+If the broker restarts for any reason (e.g., crash, rolling upgrade), this volatile in-memory state is completely lost.
+
+### The Impact: Upon restart, the broker's only source of truth is the last persisted cursor position.
+This position does not account for the lost tracking information, forcing the broker to re-dispatch messages that were already processed.
+This results in significant and difficult-to-predict message duplication for downstream consumers.
+
+The lack of administrative controls on message delivery delays introduces critical risks to cluster stability and data integrity.
+This proposal aims to provide granular control at the topic and namespace levels to provide fixed delay delivery configuration and to add a maximum delay delivery restriction.
+
+# Goals
+## In Scope
+- Add a new config that limits the maximum allowed delivery delay for a topic
+- Add a fixed delivery delay config for a topic
+- Add a new config that limits the maximum allowed delivery delay for a namespace
+- Add a fixed delivery delay config for a namespace
+- Provide safe defaults and observability.
+
+# High Level Design
+- Add configurations `maxDelayedDeliveryTimeMs`  `fixedDelayedDeliveryTimeMs` for Topic-level and Namespace-level configurations
+- A three-tiered policy hierarchy (Topic > Namespace > Broker) for setting the maximum allowed delivery delay
+- A new, overriding policy (Topic > Namespace) for enforcing a fixed delivery delay
+- If both fixed-delivery-delay and max-delivery-delay policies are configured, the fixed-delivery-delay will always take precedence, overriding any client-side settings and ignoring the max-delivery-delay policy.
+- Default Behavior: If no policies are configured, the broker will honor the client's requested deliverAt time, constrained only by the broker-wide delayedDeliveryMaxDelayInMillis setting. If a policy is set at the namespace level but not the topic level, the topic will inherit the namespace's policy.
+
+# Detailed Design
+## Design & Implementation Details
+In the broker dispatch path (Producer.sendMessage), reject the messages that are not satisfied
+if a fixed value was set for future delivery, then ignore the client's configuration
+```
+if (messageMetadata.hasDeliverAtTime()) {
+long delay = messageMetadata.getDeliverAtTime() - System.currentTimeMillis();
+long maxDelay = getTopicMaxDeliverAtTime(topic);
+    if (delay > maxDelay) {
+        // Reject the message
+        ctx.writeAndFlush(new CommandSendError(...));
+        return;
+    }
+}
+```
+1.  **Data Model Changes**:
+    *   Add `maxDelayedDeliveryTimeInSeconds` and `fixedDelayedDeliveryTimeInSeconds` fields to `TopicPolicies.java` and `Policies.java` to store the configuration values at the topic and namespace levels, respectively.
+    *   Update `HierarchyTopicPolicies.java` to include the new fields in the effective policy resolution.
+2.  **Enforcement Logic**:
+    *   Modify the `publishMessage` method in `Producer.java` to:
+        *   Enforce the `maxDelayedDeliveryTimeInSeconds` policy by rejecting messages with a delivery delay exceeding the configured maximum.
+        *   Override the message's `deliverAtTime` with the `fixedDelayedDeliveryTimeInSeconds` if the fixed delay policy is enabled.
+3.  **Admin API**:
+    *   Add new subcommands to `CmdTopicPolicies.java` and `CmdNamespaces.java` to manage the new policies through the admin CLI.
+    *   Implement new REST API endpoints for setting, getting, and removing the policies.
+
+## Public-facing Changes
+
+### Public API
+#### Topic-Level
+| Method   | Endpoint                                                                      | Description                                                                          |
+| :------- | :---------------------------------------------------------------------------- | :----------------------------------------------------------------------------------- |
+| `PUT`    | `/admin/v2/topicPolicies/{tenant}/{namespace}/{topic}/maxDelayedDeliveryTime` | Sets the maximum allowed delivery delay for a topic.                                 |
+| `PUT`    | `/admin/v2/topicPolicies/{tenant}/{namespace}/{topic}/fixedDelayedDeliveryTime` | Sets the fixed delivery delay for a topic.                                           |
+| `GET`    | `/admin/v2/topicPolicies/{tenant}/{namespace}/{topic}/maxDelayedDeliveryTime` | Gets the maximum allowed delivery delay for a topic.                                 |
+| `GET`    | `/admin/v2/topicPolicies/{tenant}/{namespace}/{topic}/fixedDelayedDeliveryTime` | Gets the fixed delivery delay for a topic.                                           |
+| `DELETE` | `/admin/v2/topicPolicies/{tenant}/{namespace}/{topic}/maxDelayedDeliveryTime` | Removes the maximum allowed delivery delay for a topic, reverting to namespace default. |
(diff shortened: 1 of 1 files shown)


pip/pip-437.md
```diff
@@ -0,0 +1,167 @@
...
+*   `get-fixed-delivery-delay`
+*   `remove-fixed-delivery-delay`
+
+New subcommands added to `pulsar-admin namespace`:
+*   `set-max-delivery-delay`
+*   `get-max-delivery-delay`
+*   `remove-max-delivery-delay`
+*   `set-fixed-delivery-delay`
+*   `get-fixed-delivery-delay`
+*   `remove-fixed-delivery-delay`
```
lhotari: There's a related "PIP-315: Configurable max delay limit for delayed delivery", https://github.com/apache/pulsar/blob/master/pip/pip-315.md, implemented by https://github.com/apache/pulsar/pull/21798 . It would be great that the new proposal would follow a similar solution for the API.

pip/pip-437.md
```diff
@@ -0,0 +1,167 @@
...
+This results in significant and difficult-to-predict message duplication for downstream consumers.
+
+The lack of administrative controls on message delivery delays introduces critical risks to cluster stability and data integrity.
+This proposal aims to provide granular control at the topic and namespace levels to provide fixed delay delivery configuration and to add a maximum delay delivery restriction.
+
+# Goals
+## In Scope
+- Add a new config that limits the maximum allowed delivery delay for a topic
+- Add a fixed delivery delay config for a topic
+- Add a new config that limits the maximum allowed delivery delay for a namespace
```
lhotari: I think that the maximum allowed delivery delay is already covered by ["PIP-315: Configurable max delay limit for delayed delivery"](https://github.com/apache/pulsar/blob/master/pip/pip-315.md)

pip/pip-437.md
chrisqwang3: Just updated accordingly

pip/pip-437.md
```diff
@@ -0,0 +1,134 @@
...
+# Motivation
+The current system lacks administrative controls on message delivery delays, which introduces risks to cluster stability and data integrity.
+
+## Risks in the current system:
+Unbounded Resource Pressure: Excessively delayed messages must be retained in storage for extended periods.
+More importantly, they create a long-lived state in the broker's memory that cannot be cleared,
+leading to unpredictable resource consumption and potential garbage collection pressure
+
+### The Limitation:
+To protect broker memory, this tracker is capped at a configurable fixed number of entries (e.g., 10,000).
```
lhotari: As background, it would be useful to reference the configuration settings behind this limit. There are some details in discussion https://github.com/apache/pulsar/discussions/23990.

pip/pip-437.md
```diff
@@ -0,0 +1,134 @@
...
+A large volume of messages with widely varying delivery times can easily exhaust this capacity.
+Once the tracker is full, it stops persisting the index for new delayed messages, retaining this state only in memory.
```
lhotari: Which tracker is this referring to? The `InMemoryDelayedDeliveryTracker` or `BucketDelayedDeliveryTracker` itself doesn't have a limit. 

In the case of `InMemoryDelayedDeliveryTracker`, it will always discard the in-memory state when all consumers have disconnected and when the first consumer arrives while the consumer list is empty. 

`tracker.clear()` is called on line 198:
https://github.com/apache/pulsar/blob/f2618c15bb5a9f3fcb577068584df5d0e2e4f335/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java#L179-L201

The way `InMemoryDelayedDeliveryTracker` behaves is that it will always read all messages from the "mark delete position" (since `cursor.rewind()` moves the "read position" to "mark delete position") in this case.

I assume that the limit that you are referring to is the `managedLedgerMaxUnackedRangesToPersist` limit (also referred to in https://github.com/apache/pulsar/discussions/23990 discussion).

pip/pip-437.md
```diff
@@ -0,0 +1,134 @@
...
+If the broker restarts for any reason (e.g., crash, rolling upgrade), this volatile in-memory state is completely lost.
+
+### The Impact: Upon restart, the broker's only source of truth is the last persisted cursor position.
+This position does not account for the lost tracking information, forcing the broker to re-dispatch messages that were already processed.
```
lhotari: Please be more specific here about what is the "lost tracking information" (similar to previous comments).

pip/pip-437.md
```diff
@@ -0,0 +1,134 @@
...
+### Configuration
+
+### CLI
+*   The existing `set-delayed-delivery` command in `CmdTopicPolicies.java` and `CmdNamespaces.java` will be updated with a new optional parameter: `--fixed-delay` (or `-fd`).
+*   The REST endpoints under `/delayedDelivery` will be updated to accept and return the new `fixedDeliveryDelayInMillis` field in their JSON payload.
+
+### Metrics
+
+To provide visibility into the enforcement of the new policies, we will introduce a new counter metric.
+*   **Full Name**: `pulsar.broker.topic.messages.delayed.rejected`
```
lhotari, line 79: It would be useful to mention that this metric is for `maxDeliveryDelayInMillis ` added PIP-315.

lhotari:
@chrisqwang3 Your email message to the Pulsar dev mailing list had gone into the moderation queue. I released it from the queue today: https://lists.apache.org/thread/23t4zzbfjrm5r4pbzrofn3d85c0171yn . Please make sure to subscribe to the dev mailing list unless you have already done that. [Instructions](https://pulsar.apache.org/contact/#mailing-lists).

pip/pip-437.md
chrisqwang3: updated

lhotari (approved):
LGTM. Great work @chrisqwang3

pip/pip-437.md
```diff
@@ -0,0 +1,129 @@
+# PIP-437: Granular and Fixed-Delay Policies for Message Delivery
```
lhotari, line 1: please make the title of the PR and the title here match

lhotari:
@chrisqwang3 Thank you for driving this proposal.
Please go ahead and start the PIP vote thread since this PIP has been in discussion for some time.
Here's an example of how I've started vote threads: https://lists.apache.org/thread/ndh0dhfjqf1htcmodgyyqt6k4bgdfj3x

pip/pip-437.md
```diff
@@ -0,0 +1,129 @@
...
+-   Introduce a new `fixed-delivery-delay` policy, configurable at the namespace and topic levels, to enforce a mandatory delivery delay.
+-   Enhance the existing `/delayedDelivery` admin API endpoints and `pulsar-admin` commands to manage both the existing `maxDeliveryDelayInMillis` and the new `fixedDeliveryDelayInMillis` policies within the same policy group.
+-   Ensure the `fixed-delivery-delay` policy takes precedence over any client-specified delay and the existing `max-delivery-delay` policy.
+
+# High Level Design
+This proposal will enhance the existing `DelayedDeliveryPolicies` object to include a new `fixedDeliveryDelayInMillis` field.
+This new policy will follow Pulsar's standard hierarchical model, allowing it to be set at the namespace level and overridden at the topic level.
+The core logic will be implemented with the following precedence:
+
+1.  **If `fixed-delivery-delay` is set:** The broker will ignore any `deliverAt` time sent by the producer and will override it by calculating `publish_time + fixed_delay`. The `max-delivery-delay` policy is ignored.
```
coderzc, line 39: It doesn't seem like a good practice to directly override the client's `deliverAt` time, this can increase the risk of misuse

lhotari: @coderzc I think that it's the point of this PIP to override the client's `deliverAt` time. I think that it's valuable for certain type of designs where a specific topic should always delay all messages.

coderzc: I think it's better to have a clear flag or special value on the client side to indicate that the message will use the fixed delay on the server side, directly override the client's deliverAt time may cause confusion for the users.

lhotari: > I think it's better to have a clear flag or special value on the client side to indicate that the message will use the fixed delay on the server side, directly override the client's deliverAt time may cause confusion for the users.

@coderzc How would you handle that? The purpose of the feature is to enforce a specific fixed delay on the broker side without letting the client be in control of the delay.

chrisqwang3: If a client could choose not to set the flag or use the fixed delay value, they could still publish messages with varied delays, re-introducing the in stability problem this PIP aims to solve. We can add a server log warning whenever the client provide a different deliverAt time to inform that they are using a fixedDelay topic

coderzc: > We can add a server log warning whenever the client provide a different deliverAt time to inform that they are using a fixedDelay topic

I agree this way

lhotari: > > We can add a server log warning whenever the client provide a different deliverAt time to inform that they are using a fixedDelay topic
> 
> I agree this way

Logging a warning for each published message can quickly fill up logs. The vast amount of logging could become a problem. How often does an end user look into server side logs? This type of logging would be useful for very small amount of users.

lhotari: A possible way to resolve this would be to add a topic stat and possibly also a metric which has a counter for how many messages have their delay adjusted because of the fixedDelay policy. 
That would provide observability into this without requiring to scan logs. Topic stats are the usual way to debug any message delivery issues. @coderzc WDYT?

Denovo1998: @lhotari @coderzc @chrisqwang3  
Can we save a flag in the receipt of the producer sending the message, which is used to record that the delay time has been overwritten? On the producer side, there is no need to record the warning every time a message that exceeds the delay time limit is sent. Just record it when it is sent for the first time?

Denovo1998:
@lhotari @coderzc @chrisqwang3 
Why do we already have max-delivery-delay, and we still need a fixed-delivery-delay? I understand that this PIP is to fix the delay time of the message, but it feels strange when I see that fixed-delivery-delay can overwrite max-delivery-delay.

Why don‘t we continue to use max-delivery-delay, and then add a configuration similar to delivery-delay-type? It can be FIXED or VARIABLE. If it is FIXED, it is a fixed delay time.

lhotari:
> @lhotari @coderzc @chrisqwang3 Why do we already have max-delivery-delay, and we still need a fixed-delivery-delay? I understand that this PIP is to fix the delay time of the message, but it feels strange when I see that fixed-delivery-delay can overwrite max-delivery-delay.

@Denovo1998  There is a reason. When all messages in a topic have the same delay, there will be significantly less state to manage. A system can be designed in a way where topics are used as schedulers and this will also work at very high scale without issues. This would work well with InMemoryDelayedDeliveryTracker since it has the `delayedDeliveryFixedDelayDetectionLookahead` feature (#16609, #17907). There's no need to build a large index of the delayed messages.

https://github.com/apache/pulsar/blob/84205ebd849479edac2c6533ea9259091e2e5bed/conf/broker.conf#L659-L663

btw. It would be useful if there would be a way to specify `delayedDeliveryTrackerFactoryClassName` in a namespace level or topic level policy since the fixed delivery delay feature is mainly useful with InMemoryDelayedDeliveryTracker and the fixed delay detection. At the same time, there might be a requirement to use BucketDelayedDeliveryTrackerFactory for other delay delivery use cases.

> Why don‘t we continue to use max-delivery-delay, and then add a configuration similar to delivery-delay-type? It can be FIXED or VARIABLE. If it is FIXED, it is a fixed delay time.

Here's the current DelayedDeliveryPolicies class:
https://github.com/apache/pulsar/blob/master/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/policies/data/DelayedDeliveryPolicies.java#L23-L41

I guess the reason is that we would break backwards compatibility if we'd make changes in the way that you are suggesting.

Denovo1998:
Oh, it‘s very clear. I see. +1

pip/pip-437.md
```diff
@@ -0,0 +1,129 @@
...
+Pulsar's delayed delivery feature allows producers to schedule messages to be delivered at a future time. To provide administrative control and prevent abuse, **PIP-315** introduced a `maxDeliveryDelayInMillis` policy.
+While this provides an important safeguard by setting an *upper bound*, it does not address all administrative needs. There is currently no way to enforce a *specific, non-overridable* delay for all messages on a topic.
+
+# Motivation
+
+### The Limitation:
+Pulsar's delayed message delivery relies on tracking individual messages until their delivery time. This tracking state is persisted as part of the subscription's cursor metadata,
+which records gaps or "holes" for unacknowledged messages. The Pulsar Managed Ledger has a hard limit on the number of disjoint unacknowledged ranges it can persist for a cursor, 
+configured by `managedLedgerMaxUnackedRangesToPersist` (defaulting to 10,000).
+A large volume of messages with widely varying delivery times can easily exhaust this capacity with each delayed message creating a separate unacknowledged hole.
```
lhotari, line 13: It would also be useful to document the benefits of the fixed delivery delay. There's a comment https://github.com/apache/pulsar/pull/24625#issuecomment-3232734535 with some content that would be useful to include in the PIP document.

pip/pip-437.md
chrisqwang3: Since the fixed-delay policy is an admin control feature, server-side observability looks more appropriate and pragmatic solution to me

lhotari: This has been addressed with the `pulsar.broker.topic.messages.fixed.delay.overridden` counter for monitoring.

PR #24626, opened 2025-08-14
[fix][proxy] Fix TooLongFrameException with Pulsar Proxy

lhotari:
### Motivation

Using the Pulsar Proxy could result in a `TooLongFrameException`.

When the Pulsar Proxy switches to direct proxy mode, it removes the frame decoder:
https://github.com/apache/pulsar/blob/baa288a0e5acdb7dca94e5c4b066e4a41e06fc55/pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/DirectProxyHandler.java#L419-L426

However, this fails since the client changes the "frameDecoder" handler's name to "newFrameDecoder"
https://github.com/apache/pulsar/blob/baa288a0e5acdb7dca94e5c4b066e4a41e06fc55/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L396-L404

A possible workaround is to use `proxyZeroCopyModeEnabled=true` since in the zero copy mode, the channel handlers are ignored. The Zero copy mode cannot be enabled when TLS is enabled between the Pulsar Proxy and Pulsar Brokers.

### Modifications

- fix the issue in removing the frame decoder when switching to direct proxy mode
- refactor frame decoder logic to reduce code duplication and future errors

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/FrameDecoderUtil.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/FrameDecoderUtil.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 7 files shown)


merlimat (approved):

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/FrameDecoderUtil.java
```diff
@@ -0,0 +1,67 @@
...
+import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
+import lombok.experimental.UtilityClass;
+
+/**
+ * Utility class for managing Netty LenghtFieldBasedFrameDecoder instances in a Netty ChannelPipeline
+ * for the Pulsar binary protocol.
+ */
+@UtilityClass
+public class FrameDecoderUtil {
```
mattisonchao, line 30: Is it possible to add a unit test for the utils public api?

mattisonchao (approved):

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/FrameDecoderUtil.java
lhotari: These are single line utility methods. I don't think it makes sense to unit test these.

PR #24631, opened 2025-08-14
[feat] Pulsar authentication for Client authentication using private_key_jwt method

### LINKED ISSUE #22371: Enable Pulsar authentication for Client authentication using private_key_jwt method
### Search before asking

### Motivation

The motivation for this enhancement stems from the need to bolster security and efficiency in Pulsar’s client authentication. The private_key_jwt method offers a secure means for client authentication, utilizing a private key and JSON Web Token (JWT). By implementing this method, Pulsar can enhance its security framework.

### Solution

The proposed solution involves enabling Pulsar authentication for Client authentication using the private_key_jwt method. 

Related materials:

For further details on the private_key_jwt method, refer to the specification outlined in https://kb.authlete.com/en/s/oauth-and-openid-connect/a/client-auth-private-key-jwt

### Alternatives

While evaluating alternatives, the current OAuth2 flow with client credentials (client_id and client_secret) was noted. This in our case is not an option, since we use the flow with client_assertion_type and client_assertion. 

Implementing the private_key_jwt method offers a more secure and efficient alternative, reducing dependency on client_secret and providing enhanced security through private key and JWT-based authentication.

### Anything else?

_No response_

dlg99:
Fixes #22371

### Motivation

support private_key_jwt client authentication

### Modifications

Implemented authentication, added tests

### Verifying this change

This change added unit tests

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/dlg99/pulsar/pull/24

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -243,7 +243,7 @@ flexible messaging model and an intuitive client API.</description>
     <debezium.mysql.version>8.0.33</debezium.mysql.version>
     <!-- Override version that brings CVE-2022-3143 with debezium -->
     <wildfly-elytron.version>1.15.16.Final</wildfly-elytron.version>
-    <jsonwebtoken.version>0.11.1</jsonwebtoken.version>
+    <jsonwebtoken.version>0.11.5</jsonwebtoken.version>
     <opencensus.version>0.28.0</opencensus.version>
     <hadoop3.version>3.4.0</hadoop3.version>
     <dnsjava3.version>3.6.2</dnsjava3.version>
@@ -311,6 +311,7 @@ flexible messaging model and an intuitive client API.</description>
 
     <!-- test dependencies -->
     <testcontainers.version>1.20.4</testcontainers.version>
+    <mock-oauth2-server.version>2.1.11</mock-oauth2-server.version>
     <!-- Set docker-java.version to the version of docker-java used in org.testcontainers:testcontainers pom -->
     <docker-java.version>3.4.0</docker-java.version>
     <hamcrest.version>2.2</hamcrest.version>
@@ -1203,17 +1204,7 @@ flexible messaging model and an intuitive client API.</description>
 
       <dependency>
         <groupId>io.jsonwebtoken</groupId>
-        <artifactId>jjwt-api</artifactId>
-        <version>${jsonwebtoken.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.jsonwebtoken</groupId>
-        <artifactId>jjwt-impl</artifactId>
-        <version>${jsonwebtoken.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.jsonwebtoken</groupId>
-        <artifactId>jjwt-jackson</artifactId>
+        <artifactId>jjwt</artifactId>
         <version>${jsonwebtoken.version}</version>
       </dependency>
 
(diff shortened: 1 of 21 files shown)


pom.xml
```diff
@@ -1727,6 +1718,26 @@ flexible messaging model and an intuitive client API.</description>
...
+      <dependency>
+        <groupId>io.jsonwebtoken</groupId>
+        <artifactId>jjwt-api</artifactId>
+        <version>${jsonwebtoken.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.jsonwebtoken</groupId>
+        <artifactId>jjwt-impl</artifactId>
+        <version>${jsonwebtoken.version}</version>
+        <scope>runtime</scope>
```
lhotari, line 1732: scope shouldn't be added to dependencyManagement dependencies since dependencyManagement's purpose is to pin the version of a dependency, regardless of the scope of the actual dependency.

pom.xml
```diff
@@ -1727,6 +1718,26 @@ flexible messaging model and an intuitive client API.</description>
...
+      </dependency>
+      <dependency>
+        <groupId>io.jsonwebtoken</groupId>
+        <artifactId>jjwt-jackson</artifactId> <!-- or jjwt-gson if Gson is preferred -->
+        <version>${jsonwebtoken.version}</version>
+        <scope>runtime</scope>
```
lhotari, line 1738: remove scope

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/PrivateKeyReader.java
```diff
@@ -0,0 +1,486 @@
...
+ * <p>
+ * <p/>It can read PEM files with PKCS#8 or PKCS#1 encodings.
+ * It doesn't support encrypted PEM files.
+ * <p>
+ * "borrowed" from https://github.com/groovenauts/jmeter_oauth_plugin/blob/master/jmeter/
+ * src/main/java/org/apache/jmeter/protocol/oauth/sampler/PrivateKeyReader.java
+ * with some modifications:
+ * - not tied to key specified as a file path
+ * - minus extra dependencies from jmeter
+ * - minus key caching
```
lhotari, line 60: There's a copyright issue in "borrowing", although it's ASL 2.0 licensed. I think it would be better to use a similar approach as is used in Zookeeper, https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/util/PemReader.java . That is from another ASF project, https://github.com/airlift/airlift/blob/master/security/src/main/java/io/airlift/security/pem/PemReader.java . 
The benefit would also be that it would support also EC keys which are commonly preferred over RSA keys.

lhotari: Another approach would be to use Pulsar's existing PEM key support classes and improve them if there's a gap.

PR #24636, opened 2025-08-18
[improve][broker] Add guard to prevent concurrent snapshot creation with warning log by ReplicationSubscriptionController

### LINKED ISSUE #24635: [Enhancement] Concurrent Snapshots for Replicated Subscriptions controller could cause BadVersionException
### Search before reporting

### Motivation

Replicated subscription snapshot creation is scheduled at a fixed interval. There is a risk of initiating overlapping snapshot tasks in slow environments with small interval.

This could result in the following exception in the ManagedLedger.

> java.util.concurrent.CompletionException: org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException: org.apache.pulsar.metadata.api.MetadataStoreException$BadVersionException: org.apache.zookeeper.KeeperException$BadVersionException: KeeperErrorCode = BadVersion for 

### Solution

Added a concurrency guard using an AtomicBoolean to ensure only one startNewSnapshot() runs at a time. When a snapshot is already in progress, subsequent attempts are skipped with a warning log.

### Alternatives

_No response_

### Anything else?

_No response_

sandeep-mst:
Fixes #24635

### Motivation

Replicated subscription snapshot creation is scheduled at a fixed interval. There is a risk of initiating overlapping snapshot tasks in slow environments with small interval.

While `lastCompletedSnapshotStartTime` already prevents redundant snapshot creation when no new data is present, it does not safeguard against concurrent execution if a previous snapshot is still in progress.

### Modifications

- Introduced an AtomicBoolean `isRunning` to guard the `startNewSnapshot()` method and prevent concurrent executions.
- Added a warning log when a new snapshot is skipped due to one already in progress.

### Verifying this change

  - *Will be updated*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: 
https://github.com/cognitree/pulsar/pull/26

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
@@ -279,7 +279,8 @@ public class ReplicatedSubscriptionsController implements AutoCloseable, Topic.P
             pendingSnapshotsMetric.inc();
             stats.recordSnapshotStarted();
             ReplicatedSubscriptionsSnapshotBuilder builder = new ReplicatedSubscriptionsSnapshotBuilder(this,
-                    topic.getReplicators().keySet(), topic.getBrokerService().pulsar().getConfiguration(), Clock.systemUTC());
+                    topic.getReplicators().keySet(), topic.getBrokerService().pulsar().getConfiguration(),
+                    Clock.systemUTC());
             pendingSnapshots.put(builder.getSnapshotId(), builder);
             builder.start();
 


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
```diff
@@ -219,61 +222,71 @@ private void receiveSubscriptionUpdated(ReplicatedSubscriptionsUpdate update) {
...
-            //    the timeout mechanism of pendingSnapshots is used to ensure retries.
-            //
-            // In other words, when hit this case, The frequency of sending SNAPSHOT_REQUEST
-            // will use `replicatedSubscriptionsSnapshotTimeoutSeconds`.
-            if (log.isDebugEnabled()) {
-                log.debug("[{}] PendingSnapshot exists but has never succeeded. "
-                        + "Skipping snapshot creation until pending snapshot timeout.", topic.getName());
-            }
+        if (!isRunning.compareAndSet(false, true)) {
+            log.warn("[{}] Snapshot already in progress. Skipping new snapshot.", topic.getName());
```
lhotari, line 226: `warn` level is perhaps too strong since this isn't a real problem. I think `info` would be more appropriate.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
```diff
@@ -92,6 +93,8 @@ public class ReplicatedSubscriptionsController implements AutoCloseable, Topic.P
 
     private final OpenTelemetryReplicatedSubscriptionStats stats;
 
+    private final AtomicBoolean isRunning = new AtomicBoolean(false);
```
lhotari, line 96: `isRunning` is not a great name for this field. `startingNewSnapshot` would be better since this only applies to the `startNewSnapshot` method.

lhotari:
> Replicated subscription snapshot creation is scheduled at a fixed interval. There is a risk of initiating overlapping snapshot tasks in slow environments with small interval.

btw. a similar problem exists in multiple locations in Pulsar. There's #24596 to address such issues.

PR #22283, opened 2024-03-16
[fix] [broker] fix mismatch between dispatcher.consumerList and dispatcher.consumerSet

poorbarcode:
### Motivation

| time | client | broker |
| --- | --- | --- |
| 1 | | Unload topic |
| 2 | Subscribe(1st) after disconnect |
| 3 | | Initialize ManagedLedger. This operation is slowly |
| 4 | Subscribe(1st) timeout, try to close the consumer |
| 5 | | Received the command `CloseConsumer`, then remove `consumerFuture(1st)` from `ServerCnx`, then the mechanism to prevent registering a consumer with the same id is invalid now | 
| 6 | | Initialize ManagedLedger finish, add consumer(1st) to Dispatcher. <br> &nbsp;&nbsp;- `dispatcher.consumerSet.size`: `1` <br> &nbsp;&nbsp;- `dispatcher.consumerList.size`: `1` |
| 7 | Retry to subscribe(2nd) |
| 8 | | add consumer(2nd) to Dispatcher.  <br> &nbsp;&nbsp;- `dispatcher.consumerSet.size`: `1` <br> &nbsp;&nbsp;- `dispatcher.consumerList.size`: `2` |
| 9 | | Since the future of consumer(1st) is failed, remove it <br> &nbsp;&nbsp;- `dispatcher.consumerSet.size`: `0` <br> &nbsp;&nbsp;- `dispatcher.consumerList.size`: `1` |

This issue may cause a stuck of dispatchers closing/unsubscribing and topic closing/deleting. See more context: https://github.com/apache/pulsar/pull/22270

BTW, the scenario of producer registration has no issue because the producers with the same `ServerCnx + producerId` will override each other.

### Modifications

- The subsequent subscribing will always fail until the previous one is done even if the previous subscribing was timed out in the client view, which guarantees that there is no race condition about twice subscribing.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24638, opened 2025-08-18
[fix][broker] Remove consumer from cache if closed before creation

nodece:
### Motivation

When a consumer sends a subscribe command and then immediately sends a close command while the subscription is still in progress, the broker logs messages like:
```
2025-08-15T14:42:52,889+0800 [pulsar-io-4-40] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.184.74.6:44058] Closing consumer: consumerId=13296
2025-08-15T14:42:52,889+0800 [pulsar-io-4-40] INFO  org.apache.pulsar.broker.service.ServerCnx - [/10.184.74.6:44058] Closed consumer before its creation was completed. consumerId=13296
2025-08-15T14:42:53,258+0800 [pulsar-io-4-40] WARN  org.apache.pulsar.broker.service.ServerCnx - [/10.184.74.6:44058][persistent://public/10000/__change_events-partition-0][multiTopicsReader-f3f7fd855d] A failed consumer with id is already present on the connection, consumerId=13296
```

This happens because the consumer remains in the connection's cache even after it is closed before the subscription is fully established. As a result, any new subscription attempt with the same consumer ID fails with `A failed consumer with id is already present on the connection.`

This regression was introduced by [#22283](https://github.com/apache/pulsar/pull/22283).

### Modifications

- Remove consumer from cache if closed before creation in the `handleCloseConsumer`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -2236,6 +2236,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
             // create operation will complete, the new consumer will be discarded.
             log.info("[{}] Closed consumer before its creation was completed. consumerId={}",
                      remoteAddress, consumerId);
+            consumers.remove(consumerId, consumerFuture);
             commandSender.sendSuccessResponse(requestId);
             return;
         }
@@ -2243,6 +2244,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         if (consumerFuture.isCompletedExceptionally()) {
             log.info("[{}] Closed consumer that already failed to be created. consumerId={}",
                      remoteAddress, consumerId);
+            consumers.remove(consumerId, consumerFuture);
             commandSender.sendSuccessResponse(requestId);
             return;
         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -62,6 +62,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import javax.naming.AuthenticationException;
 import javax.net.ssl.SSLSession;
+import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import lombok.Getter;
@@ -89,9 +90,11 @@ import org.apache.pulsar.broker.loadbalance.extensions.data.BrokerLookupData;
 import org.apache.pulsar.broker.namespace.LookupOptions;
 import org.apache.pulsar.broker.namespace.NamespaceService;
 import org.apache.pulsar.broker.service.BrokerServiceException.ConsumerBusyException;
+import org.apache.pulsar.broker.service.BrokerServiceException.ConsumerClosedException;
 import org.apache.pulsar.broker.service.BrokerServiceException.ServerMetadataException;
 import org.apache.pulsar.broker.service.BrokerServiceException.ServiceUnitNotReadyException;
 import org.apache.pulsar.broker.service.BrokerServiceException.SubscriptionNotFoundException;
+import org.apache.pulsar.broker.service.BrokerServiceException.TopicMigratedException;
 import org.apache.pulsar.broker.service.BrokerServiceException.TopicNotFoundException;
 import org.apache.pulsar.broker.service.persistent.PersistentSubscription;
 import org.apache.pulsar.broker.service.persistent.PersistentTopic;
@@ -1258,8 +1261,11 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         // epoch using different consumer futures, and only remove the consumer future from the map
         // if subscribe failed .
         CompletableFuture<Consumer> consumerFuture = new CompletableFuture<>();
-        CompletableFuture<Consumer> existingConsumerFuture =
-                consumers.putIfAbsent(consumerId, consumerFuture);
+        consumerFuture.exceptionally((ex) -> {
+            consumers.remove(consumerId, consumerFuture);
+            return null;
+        });
+        CompletableFuture<Consumer> existingConsumerFuture = consumers.putIfAbsent(consumerId, consumerFuture);
         isAuthorizedFuture.thenApply(isAuthorized -> {
             if (isAuthorized) {
                 if (log.isDebugEnabled()) {
@@ -1274,7 +1280,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                             service.getPulsar().getConfiguration().getMaxConsumerMetadataSize());
                 } catch (IllegalArgumentException iae) {
                     final String msg = iae.getMessage();
-                    consumers.remove(consumerId, consumerFuture);
+                    consumerFuture.completeExceptionally(iae);
                     commandSender.sendErrorResponse(requestId, ServerError.MetadataError, msg);
                     return null;
                 }
@@ -1396,7 +1402,8 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                             if (consumer.checkAndApplyTopicMigration()) {
                                 log.info("[{}] Disconnecting consumer {} on migrated subscription on topic {} / {}",
                                         remoteAddress, consumerId, subscriptionName, topicName);
-                                consumers.remove(consumerId, consumerFuture);
+                                consumerFuture.completeExceptionally(
+                                        new TopicMigratedException("Topic has been migrated"));
                                 return;
                             }
 
@@ -1423,7 +1430,8 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                                                     + " after timeout on client side {}: {}",
                                             remoteAddress, consumer, e.getMessage());
                                 }
-                                consumers.remove(consumerId, consumerFuture);
+                                consumerFuture.completeExceptionally(new ConsumerClosedException(
+                                        "Cleared consumer created after timeout on client side"));
                             }
 
                         })
@@ -1450,7 +1458,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                                         log.info("consumer client doesn't support topic migration handling {}-{}-{}",
                                                 topicName, remoteAddress, consumerId);
                                     }
-                                    consumers.remove(consumerId, consumerFuture);
+                                    consumerFuture.completeExceptionally(exception);
                                     closeConsumer(consumerId, Optional.empty());
                                     return null;
                                 }
@@ -1471,7 +1479,6 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                                         BrokerServiceException.getClientErrorCode(exception.getCause()),
                                         exception.getCause().getMessage());
                             }
-                            consumers.remove(consumerId, consumerFuture);
 
                             return null;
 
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ServerCnxTest.java
```diff
@@ -2436,44 +2436,30 @@ public void testSubscribeBookieTimeout() throws Exception {
...
         // Close succeeds
         response = getResponse();
         assertEquals(response.getClass(), CommandSuccess.class);
         assertEquals(((CommandSuccess) response).getRequestId(), 2);
 
-        // Subscribe fails
-        response = getResponse();
-        assertEquals(response.getClass(), CommandError.class);
-        assertEquals(((CommandError) response).getRequestId(), 3);
+        // We should not receive response for 1st consumer, since it was cancelled by the close
```
nodece: It's better to respond to the client even if the subscribe request was cancelled by a close command, otherwise the client will be left hanging.

This is a TODO.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2229,13 +2229,15 @@ protected void handleCloseConsumer(CommandCloseConsumer closeConsumer) {
...
                      remoteAddress, consumerId);
+            consumers.remove(consumerId, consumerFuture);
             commandSender.sendSuccessResponse(requestId);
             return;
         }
 
         if (consumerFuture.isCompletedExceptionally()) {
             log.info("[{}] Closed consumer that already failed to be created. consumerId={}",
                      remoteAddress, consumerId);
+            consumers.remove(consumerId, consumerFuture);
```
BewareMyPower, line 2247: Instead of manually calling `remove`, which is easily to be missed or removed, a better pattern is to register `exceptionallyAsync` to remove the failed consumer.

In `handleSubscribe`:

```java
        CompletableFuture<Consumer> consumerFuture = new CompletableFuture<>();
        consumerFuture.exceptionallyAsync(e -> {
            if (consumers.remove(consumerId, consumerFuture)) {
                log.info("Removed failed consumer due to {}", e.getMessage());
            }
            return null;
        });
        CompletableFuture<Consumer> existingConsumerFuture =
                consumers.putIfAbsent(consumerId, consumerFuture);
```

nodece: `handleSubscribe` contains this code, and it's possible that the future was not completed, so the consumer remained in the cache. You can see the relevant code here:
https://github.com/apache/pulsar/blob/c726ab7b584dfe09c98050d84c15f69dd4497985/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L1474

Please note that my change is in the `handleCloseConsumer` method, not `handleSubscribe`. The modification ensures that if a consumer is closed before its creation completes, it is properly removed from the cache.

nodece: @BewareMyPower

nodece:
> > This happens because the consumer remains in the connection's cache even after it is closed before the subscription is fully established. As a result, any new subscription attempt with the same consumer ID fails with `A failed consumer with id is already present on the connection.`
> 
> I think this PR makes sense to me, but I'm just wondering why would a client perform a new subscription attempt with the same consumer ID if it already closed the consumer? Explaining the scenario in the PR description would be useful context.

@lhotari Yes, that's correct. I guess that the issue occurs when:
1. The client sends a subscribe request, but the broker is still processing it.
2. The client times out waiting for the response and sends a close subscribe request.
3. If the client then attempts to send a new subscribe request, the broker may still have a "failed" consumer cached from the previous request, resulting in the error: `A failed consumer with id is already present on the connection`.

The producer has same bugfix: https://github.com/apache/pulsar/blob/v3.0.12/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L2073

dao-jun:
I personally think we don't need to remove the Consumer Future unless it's done.
Maybe we should introduce Subscription creation timeout for the case, to avoid sub creation costs long time.

nodece:
> I personally think we don't need to remove the Consumer Future unless it's done. Maybe we should introduce Subscription creation timeout for the case, to avoid sub creation costs long time.

This is an async(improvement) issue and is not related to this PR.

When the consumer sends a close request, the broker will complete the future. We must remove the future at this point; Otherwise, the broker cache retains the failed consumer, and subsequent subscribe requests will hit the error:
`A failed consumer with id is already present on the connection.`

Please see https://github.com/apache/pulsar/pull/24638#issuecomment-3197127542.

lhotari:
> 3. If the client then attempts to send a new subscribe request, the broker may still have a "failed" consumer cached from the previous request, resulting in the error: `A failed consumer with id is already present on the connection`.

@nodece I think that we need @poorbarcode to explain why #22283 was needed since that seems to intentionally change the behavior and let the failed consumer future remain in the map.

lhotari:
@nodece @poorbarcode 
Based on this comment https://github.com/apache/pulsar/pull/22283#issuecomment-2012394478, I get the understanding that race conditions should be prevented while subscribing to the topic and that's why the solution was added.

Instead of leaving the failed future in the map to prevent race conditions, perhaps `org.apache.pulsar.common.util.FutureUtil.Sequencer` could be used to execute subscription requests sequentially so that race conditions could be avoided?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
BewareMyPower: This line is the `exceptionally` on the future of handling the subscribe request. However, what I meant is that when you inserted a new consumer future to `ServerCnx#consumers`, you'd better register a callback to remove it from `consumers` once it failed.

The current code looks like:

```java
        final var consumers = new ConcurrentHashMap<Integer, CompletableFuture<Void>>();
        final var future = new CompletableFuture<Void>();
        final var key = 1;
        // simulate handleSubscribe
        consumers.put(key, future);
        CompletableFuture.delayedExecutor(1, TimeUnit.HOURS).execute(() ->
                future.completeExceptionally(new RuntimeException("fail")));

        Thread.sleep(1000);

        // simulate handleClose
        final var existingFuture = consumers.get(key);
        if (existingFuture == null) {
            return;
        }
        if (!existingFuture.isDone() && existingFuture.completeExceptionally(new RuntimeException("timeout"))) {
            consumers.remove(key, existingFuture); // NOTE: it should not be missed
            return;
        }
        if (existingFuture.isCompletedExceptionally()) {
            consumers.remove(key, existingFuture); // NOTE: it should not be missed
            return;
        }
        System.out.println("close");
```

It can be improved to:

```java
        consumers.put(key, future);
        future.exceptionally(e -> {
            consumers.remove(key, future);
            return null;
        });
        /* ... */
        if (!existingFuture.isDone() && existingFuture.completeExceptionally(new RuntimeException("timeout"))) {
            return;
        }
```

1. You don't need to check `existingFuture.isCompletedExceptionally()` anymore
2. You don't need to manually remove the future after calling `existingFuture.completeExceptionally`

poorbarcode: @BewareMyPower 

@nodece wants to remove a consumer future that has not been done. Which is the main change of this PR.

---

> @dao-jun I personally think we don't need to remove the Consumer Future unless it's done.
Maybe we should introduce Subscription creation timeout for the case, to avoid sub creation costs long time.

Agree with @dao-jun 

---

@nodece 

> When a consumer sends a subscribe command and then immediately sends a close command while the subscription is still in progress, the broker logs messages like:

Could you explain the progress of reproducing the issue? seems it is a client-side bug, which should not call removing before the subscribing is done. If the consumer can not confirm the result of the command that registering the consumer, the client should close the socket

nodece: > @nodece wants to remove a consumer future that has not been done. Which is the main change of this PR.

I know the real subscription future has not been done. Once the consumer closes the subscription, I think the consumer can use the same consumer id to send a new subscription.

> Could you explain the progress of reproducing the issue?

You can see this: https://github.com/apache/pulsar/pull/24638#issuecomment-3197127542

> seems it is a client-side bug, which should not call removing before the subscribing is done. If the consumer can not confirm the result of the command that registering the consumer, the client should close the socket

Not a bug. Once the consumer checks the request timeout, it will send the close subscription request to the broker.

Please see https://github.com/apache/pulsar/blob/v3.0.12/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L907-L914

nodece: @poorbarcode

nodece:
> @nodece @poorbarcode Based on this comment [#22283 (comment)](https://github.com/apache/pulsar/pull/22283#issuecomment-2012394478), I get the understanding that race conditions should be prevented while subscribing to the topic and that's why the solution was added.

Once a subscription (including close) has completed, the consumer can reuse the same consumer ID (final value) to send a new subscription request.

The main goal here seems to be preventing a consumer from sending multiple concurrent subscription requests. However, the old code also blocks reuse of the same consumer ID even after the subscription has done, which looks more like a bug since at that point the consumer should be allowed to resume with the same ID.

I also noticed that we already handle similar concurrency cases in the ledger module, so we could apply the same approach here for `newNonDurableCursor`. This would keep concurrency control in the lower layer, and reduce unnecessary work in the upper layer:
https://github.com/apache/pulsar/blob/v3.0.12/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L978-L1005

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
nodece: @BewareMyPower Thank you, this has been addressed.

nodece:
A non-durable cursor bugfix: https://github.com/apache/pulsar/pull/24643

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: @nodece 

> Not a bug. Once the consumer checks the request timeout, it will send the close subscription request to the broker.


Could you write a test for this case?

poorbarcode:
@nodece 

> Not a bug. Once the consumer checks the request timeout, it will send the close subscription request to the broker.


Could you write a test for this case?

nodece:
> @nodece
> 
> > Not a bug. Once the consumer checks the request timeout, it will send the close subscription request to the broker.
> 
> Could you write a test for this case?

@poorbarcode We only need to verify the logic in `handleSubscribe` and `handleCloseConsumer` in ServerCnx.java.

The existing test `testSubscribeBookieTimeout` already covers this scenario and validates the changes introduced in this PR.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ServerCnxTest.java
```diff
@@ -2436,44 +2436,32 @@ public void testSubscribeBookieTimeout() throws Exception {
                 null /* assignedBrokerServiceUrl */, null /* assignedBrokerServiceUrlTls */);
         channel.writeInbound(closeConsumer);
```
poorbarcode, line 2437: @nodece 

> We only need to verify the logic in handleSubscribe and handleCloseConsumer in ServerCnx.java.

> The existing test testSubscribeBookieTimeout already covers this scenario and validates the changes introduced in this PR.

The test succeeded without the current PR

Screenshot 2025-08-20 at 15 58 42


Screenshot 2025-08-20 at 15 58 03

nodece: @poorbarcode 

Right, the test passes, but I don’t think it's correct. Previously, the simulated subscription behavior was flawed: sending consecutive subscription requests after closing a subscription could lead to inconsistent subscription state or behavior that doesn't match the author’s intent. This PR fixes both the test and its implementation, ensuring that the intended behavior is correctly verified.

poorbarcode: Could you write a test for this case?

nodece:
Thanks for the review and discussion. After re-evaluating the logic, I realized that the current behavior is correct: The consumer should only be removed from the cache once the subscription completes successfully.

Removing it earlier could lead to inconsistent states. I'll close this PR.

PR #24641, opened 2025-08-18
[fix][broker] Allow empty replication clusters for namespace policies

### LINKED ISSUE #24640: [Bug] Replication clusters cannot be cleared for a namespace
### Search before reporting

### Read release policy

### User environment

Pulsar 4.0.5

### Issue Description

Apache Pulsar allows replication clusters to be set at the namespace level. This means every topic in a namespace gets replicated as configured.

To date there is no way to disable replication at the namespace level. If an empty list of clusters is passed in, Pulsar returns an error:

```
 curl -XPOST  https://:CLUSTER_ENDPOINT/admin/v2/:TENANT/:NS/:TOPIC/replication -D '[]'
{"reason":"ClusterIds should not be null or empty"}
```

The expectation is for this to disable replication.

### Error messages

```text
{"reason":"ClusterIds should not be null or empty"}
```

### Reproducing the issue

Apache Pulsar allows replication clusters to be set at the namespace level. This means every topic in a namespace gets replicated as configured.

To date there is no way to disable replication at the namespace level. If an empty list of clusters is passed in, Pulsar returns an error:

```
 curl -XPOST  https://:CLUSTER_ENDPOINT/admin/v2/:TENANT/:NS/:TOPIC/replication -D '[]'
{"reason":"ClusterIds should not be null or empty"}
```

The expectation is for this to disable replication.

### Additional information

_No response_

omarkj:
This change removes validation that prevented setting empty replication clusters for namespace policies. Previously, the API would reject empty cluster lists, but this is a valid configuration for disabling inter-cluster replication while maintaining topic accessibility within individual clusters.

Changes:
- Remove CollectionUtils.isEmpty(clusterIds) validation in NamespacesBase.java
- Remove empty cluster validation in PersistentTopicsBase.java for topic-level policies
- Add comprehensive test coverage for empty replication clusters in V1 and V2 namespace APIs
- Verify bidirectional functionality (empty ↔ non-empty cluster configurations)
- Ensure no auto-population occurs when explicitly setting empty clusters

Fixes #24640

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -755,9 +755,6 @@ public abstract class NamespacesBase extends AdminResource {
         return validateNamespacePolicyOperationAsync(namespaceName, PolicyName.REPLICATION, PolicyOperation.WRITE)
                 .thenCompose(__ -> validatePoliciesReadOnlyAccessAsync())
                 .thenApply(__ -> {
-                    if (CollectionUtils.isEmpty(clusterIds)) {
-                        throw new RestException(Status.PRECONDITION_FAILED, "ClusterIds should not be null or empty");
-                    }
                     if (!namespaceName.isGlobal() && !(clusterIds.size() == 1
                             && clusterIds.get(0).equals(pulsar().getConfiguration().getClusterName()))) {
                             throw new RestException(Status.PRECONDITION_FAILED,
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -3346,10 +3346,6 @@ public class PersistentTopicsBase extends AdminResource {
     }
 
     protected CompletableFuture<Void> internalSetReplicationClusters(List<String> clusterIds, boolean isGlobal) {
-        if (CollectionUtils.isEmpty(clusterIds)) {
-            return CompletableFuture.failedFuture(new RestException(Status.PRECONDITION_FAILED,
-                    "ClusterIds should not be null or empty"));
-        }
         Set<String> replicationClusters = Sets.newHashSet(clusterIds);
         return validatePoliciesReadOnlyAccessAsync()
                 .thenAccept(__ -> {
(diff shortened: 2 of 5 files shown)


lhotari:
The changes so far LGTM. I'm just concerned how to remove replication clusters on the command line with `pulsar-admin topics set-replication-clusters -c "" topic_name` and whether the current command accepts an empty string for the `-c`/`--clusters=` parameter. Adding tests for that would be useful. However, I'm not sure if the current way of parsing the command line in the unit tests would behave in the same way as running the tool in a command shell and passing an empty argument such as `-c ""`.

Current tests:
https://github.com/apache/pulsar/blob/34f86576758715f881d755747a74929c189df5ce/pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java#L2146-L2147

btw. I now noticed that there's a way to remove the topic policy level configuration for replication clusters.

https://github.com/apache/pulsar/blob/34f86576758715f881d755747a74929c189df5ce/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/PersistentTopics.java#L2423-L2448

It maps to the `pulsar-admin topics` `remove-replication-clusters` command:

https://github.com/apache/pulsar/blob/34f86576758715f881d755747a74929c189df5ce/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java#L266

However, *this isn't the same as setting the list of replication clusters to an empty list*. When the topic level replication clusters is removed, it would use the namespace level setting. Being able to set it to an empty list would be necessary to override the namespace level setting so that the topic wouldn't get replicated when using a global configuration store.

omarkj:
I would think that since there is already precedence for `remove-replication-clusters` within `CmdTopics.java`, maintaining that pattern for `CmdNamespaces.java` would seem reasonable just to keep those similar APIs consistent.

Your point on the subtle difference between `remove-replication-clusters` (reverting to the namespace setting) or setting it to the empty list (disabling it) does highlight a nuance I had not considered. 

Another way to address this change would be to add an API similar to `remove-replication-clusters` to the Namespace API, and again disallow passing in the empty cluster set. It would then behave the same way as the topic APIs do today.

This would work for _my_ use case. On the other hand, I do not see why the API should not allow users to disable replication for a specific topic in a replicated namespace.

lhotari:
It seems that PIP-422 https://github.com/apache/pulsar/blob/master/pip/pip-422.md might be conflicting with the result that we'd like to achieve. PIP-422 contains behavior that once the replication clusters is updated for a global topic policy, it will delete the topics in cluster that are no longer part of the replication clusters. Automatically deleting topics in remote clusters could be considered as risky behavior and there should be a way to control this behavior since there might be a requirement where someone would simply want to later on disconnect the replication for a specific topic and continue with the topics independently. @poorbarcode Have you thought about this use case?

poorbarcode:
Seems this PR is not needed, I have answered [here](https://github.com/apache/pulsar/issues/24640#issuecomment-3222380316)

PR #24648, opened 2025-08-20
[fix][broker]User topic failed to delete after removed cluster because of failed delete data from transaction buffer topic

poorbarcode:
#### Background
- Users may start 2 clusters with the Geo-Replication feature with Global ZK
- Users want to hire one cluster, so remove the cluster from `namespace-level replicatedClusters`
  - Broker will remove the topics because the cluster was removed.
- Issue occurs:
  - User topics can not be deleted anymore, because the operation `delete data from transaction buffer` can not be executed successfully anymore, and will get an error "Namespace missing local cluster name in clusters list". Because the topic `__transaction_buffer_snapshot` can not be accessed anymore since the cluster was removed.
  - The task `topic.checkReplication` and other operations will retry again and again, as a result, the broker will crash of OOM  

### Modifications
- Do not create the transaction buffer component for transaction system topics.
- Remove the system topic `__transaction_buffer_snapshot` if it is not allowed to be accessed(removed current cluster from `allowed cluster`)
- Skip removing topic-level policies if the system topic `__transaction_buffer_snapshot` has been deleted.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/systopic/NamespaceEventsSystemTopicFactory.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/systopic/NamespaceEventsSystemTopicFactory.java
@@ -18,6 +18,8 @@
  */
 package org.apache.pulsar.broker.systopic;
 
+import java.util.concurrent.CompletableFuture;
+import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.service.SystemTopicTxnBufferSnapshotService;
 import org.apache.pulsar.client.api.PulsarClient;
 import org.apache.pulsar.common.events.EventType;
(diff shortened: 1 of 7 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/systopic/NamespaceEventsSystemTopicFactory.java
```diff
@@ -47,6 +49,28 @@ public static TopicName getEventsTopicName(NamespaceName namespaceName) {
...
+        // 3. The partitioned metadata was shared using by c1 and c2.
+        // 4. Pulsar only delete partitions when the topic is deleting from c1, because c2 is still using
+        //    partitioned metadata.
+        TopicName topicName = getSystemTopicName(namespaceName, eventType);
+        CompletableFuture<Boolean> nonPartitionedExists =
+                pulsar.getPulsarResources().getTopicResources().persistentTopicExists(topicName);
+        CompletableFuture<Boolean> partition0Exists =
+                pulsar.getPulsarResources().getTopicResources().persistentTopicExists(topicName.getPartition(0));
+        return nonPartitionedExists.thenCombine(partition0Exists, (a, b) -> {
+            System.out.println("===> a: " + a + ", b: " + b);
```
coderzc: Remove this print out

poorbarcode: Removed it. Thanks for mentioning @coderzc

poorbarcode:
The test `PartitionCreationTest.testCreateMissedPartitions` is flaky. I have pushed another PR to fix it. Please ignore it
- https://github.com/apache/pulsar/pull/24679

coderzc (approved):
LGTM

PR #24649, opened 2025-08-20
[fix][offload] Exclude unnecessary dependencies from tiered storage provider / offloader nar files

lhotari:
### Motivation

There are a lot of unnecessary dependencies packaged in tiered storage provider / offloader nar files.

### Modifications

- make `managed-ledger` dependency a provided scope dependency
- exclude Netty dependencies since they are available in the broker

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/tiered-storage/file-system/pom.xml
+++ b/tiered-storage/file-system/pom.xml
@@ -31,39 +31,44 @@
     <artifactId>tiered-storage-file-system</artifactId>
     <name>Apache Pulsar :: Tiered Storage :: File System</name>
     <dependencies>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>managed-ledger</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <version>${hdfs-offload-version3}</version>
-            <exclusions>
-                <exclusion>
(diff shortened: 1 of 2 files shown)


dao-jun (approved):

BewareMyPower (approved):

tiered-storage/file-system/pom.xml
```diff
@@ -31,39 +31,44 @@
...
+      <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>managed-ledger</artifactId>
+          <version>${project.version}</version>
+          <scope>provided</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-common</artifactId>
+        <version>${hdfs-offload-version3}</version>
```
yaalsn, line 43: Can we remove this and unify to use `hadoop3.version` which declared in root `pom.xml`? https://github.com/apache/pulsar/blob/master/pom.xml#L1376-L1379

lhotari: It maps to `hadoop3.version` here:
https://github.com/apache/pulsar/blob/64137f629fbf9d31842489e4ba89e84d48d97925/pom.xml#L250

I guess the reason why it's been separate is to be able to control the version separately since the hadoop dependency is included in the .nar file.

PR #24652, opened 2025-08-20
[improve][broker] PIP-433: Ensure topic creation before starting GEO

nodece:
PIP: #24485 

### Motivation

In a GEO replication scenario, if the remote cluster does not have the replicated topic and the auto-creation type differs between the local and remote clusters, message replication may fail. To ensure seamless replication, the topic metadata must be properly synchronized across clusters.

This is part of PIP-433.

https://github.com/apache/pulsar/pull/24136 is the same as this PR.

### Modifications

- When both the local and remote partitioned topic metadata indicate `partitions=0`, this means the topic is non-partitioned. In this case, the local cluster sends a non-partitioned topic creation request to the remote cluster.

- If the local partitioned topic metadata has `partitions>0`, this means the topic is partitioned:
  - If the remote partitioned topic metadata has `partitions=0`, the local cluster sends a partitioned topic creation request to the remote cluster.
  - If partitions differ between the local and remote cluster, please stop GEO.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentTopicTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentTopicTest.java
@@ -1724,6 +1724,8 @@ public class PersistentTopicTest extends MockedBookKeeperTestCase {
 
         @Cleanup
         PulsarAdmin admin = mockReplicationAdmin();
+        PulsarService pulsar = brokerService.getPulsar();
+        doReturn(admin).when(pulsar).getAdminClient();
         brokerService.getClusterAdmins().put(remoteCluster, admin);
         Optional<ClusterData> clusterData = brokerService.pulsar().getPulsarResources().getClusterResources()
                 .getCluster(remoteCluster);
@@ -1781,6 +1783,7 @@ public class PersistentTopicTest extends MockedBookKeeperTestCase {
 
         @Cleanup
         PulsarAdmin admin = mockReplicationAdmin();
+        doReturn(admin).when(pulsar).getAdminClient();
         brokerService.getClusterAdmins().put(remoteCluster, admin);
         Optional<ClusterData> clusterData = brokerService.pulsar().getPulsarResources().getClusterResources()
                 .getCluster(remoteCluster);


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/GeoPersistentReplicator.java
```diff
@@ -56,7 +63,115 @@ protected String getProducerName() {
...
+            TopicName completeTopicName = TopicName.get(localTopicName);
+            TopicName baseTopicName;
+            if (completeTopicName.isPartitioned()) {
+                baseTopicName = TopicName.get(completeTopicName.getPartitionedTopicName());
+            } else {
+                baseTopicName = completeTopicName;
+            }
+            // Set useFallbackForNonPIP344Brokers to true when mix of PIP-344 and non-PIP-344 brokers are used, it
+            // can still work.
+            return client.getLookup().getPartitionedTopicMetadata(baseTopicName, false, true)
```
poorbarcode: Since we have `pulsarAdmin` object, can we use `pulsar-admin.topics.getPartitionedTopicMetadataAsync` here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -2236,11 +2237,11 @@ protected CompletableFuture<Void> addReplicationCluster(String remoteCluster, Ma
...
                 .thenCompose(__ -> brokerService.pulsar().getPulsarResources().getClusterResources()
-                        .getClusterAsync(remoteCluster)
-                        .thenApply(clusterData ->
-                                brokerService.getReplicationClient(remoteCluster, clusterData)))
-                .thenAccept(replicationClient -> {
-                    if (replicationClient == null) {
+                        .getClusterAsync(remoteCluster))
+                .thenAccept((clusterData) -> {
+                    PulsarClient replicationClient = brokerService.getReplicationClient(remoteCluster, clusterData);
+                    PulsarAdmin replicationAdmin = brokerService.getClusterPulsarAdmin(remoteCluster, clusterData);
```
poorbarcode, line 2350: `brokerService.getClusterPulsarAdmin` may throw a runtime exception. Should we consider it?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
```diff
@@ -1858,4 +1861,162 @@ public void testReplicatorsInflightTaskListIsEmptyAfterReplicationFinished() thr
...
+            assertFalse(persistentTopic.getReplicators().isEmpty());
+        });
+
+        @Cleanup
+        Producer<String> p1 = client1.newProducer(Schema.STRING).topic(tp).create();
+        p1.send("msg-1");
+
+        Awaitility.await().untilAsserted(() -> {
+            if (isPartitioned){
+                assertThat(admin2.topics().getPartitionedTopicList(ns)).contains(tp);
```
poorbarcode: We'd better check that partitions are also created

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
```diff
@@ -1858,4 +1861,162 @@ public void testReplicatorsInflightTaskListIsEmptyAfterReplicationFinished() thr
...
+            PersistentTopic persistentTopic = (PersistentTopic) broker2.getTopic(tp, false).join().get();
+            persistentTopic.getReplicators().forEach((key, value) -> {
+                assertFalse(value.isConnected());
+            });
+        });
+        assertThat(admin1.topics().getList(ns)).doesNotContain(tp);
+    }
+
+    @Test
+    public void testReplicatorWhenPartitionCountsDiffer() throws Exception {
```
poorbarcode, line 2011: Good test

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
```diff
@@ -1858,4 +1861,162 @@ public void testReplicatorsInflightTaskListIsEmptyAfterReplicationFinished() thr
...
+                .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest).subscribe();
+
+        while (true) {
+            Message<String> receive = c2.receive(3, TimeUnit.SECONDS);
+            if (receive == null) {
+                break;
+            }
+            assertEquals(receive.getValue(), "msg-p2-1");
+        }
+    }
```
poorbarcode, line 2084: Could you also add another test?
- Both clusters have different topic-auto-creation rules
- The partitioned topic will by created by replicator at remote-side

poorbarcode (commented):
@nodece Thanks for driving this feature

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/GeoPersistentReplicator.java
nodece: Good point, I will refactor this logic.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
nodece: I will move this logic from PersistentTopic to the replicator.

> `brokerService.getClusterPulsarAdmin` may throw a runtime exception. Should we consider it?

If an exception is thrown, this future will be complete with an exception.

poorbarcode:
Rebase master

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
nodece: ~@poorbarcode This is an asynchronous method, but the keyword `async` is missing from its name.~

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
nodece: This case has been added to `testReplicatorWhenPartitionCountsDiffer`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: The mehod is noty a asynchronous method, see https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1716-L1720

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/GeoPersistentReplicator.java
nodece: @poorbarcode This is an asynchronous method, but the keyword `async` is missing from its name.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
nodece: Do you want to use the asynchronous method?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/GeoPersistentReplicator.java
poorbarcode: > Since we have pulsarAdmin object, can we use pulsar-admin.topics.getPartitionedTopicMetadataAsync here?

I have modified the code, please review it. https://github.com/apache/pulsar/pull/24652/commits/f7d38d74f4ba8806882e9eba43b3c4dc3b626e8b

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: > Do you want to use the asynchronous method?

No, let's focus on the main change, I will try to improve the code with a seperate PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/GeoPersistentReplicator.java
nodece: @poorbarcode Good, I should use pulsarAdmin instead of pulsarClient, this is better way.

nodece:
The tests `testReplicatorWithSpecialNonPartitionedTopic`, `testFailureReplicatorWithSpecialNonPartitionedTopic`, and `testFailureReplicatorWithSpecialNonPartitionedTopic2` were removed in commit 4238cbd8f56f43198b5285713b6f1a9616d8ff97. This behavior is no longer applicable because the corresponding scenario introduced by PIP-414 has been disabled, and PIP-433 was not cherry-picked into the older branch.

The new code has been refactored by the AI.

nodece:
ping @poorbarcode

poorbarcode:
@nodece I will review the PR again by this weekend

PR #24658, opened 2025-08-22
[improve][broker] Optimize Reader creation in TopicPoliciesService

dao-jun:
### Motivation

In our cluster, we encountered an exception on a `change_events` partition:
image
Thousands of `get last message id failed` was thrown due to `No such ledger` exception.

In the `SystemTopicBasedTopicPoliciesService`, it will create thousands of `Reader` instances.

For some reason, all the `Reader`s didn't closed successfully(maybe we were unload the ns at the time), after the `change_events` partition transferred to a new broker, all the `Reader`s was reconnected to the new broker, and never be closed.
image

There was more than 20k consumers of the partition of `change_events`:
image
  

### Modifications

Don't clean the `readerCache` and close the `Reader` unless:
1. The consumer was already closed
2. All ns bundles were removed from the broker
3. The reader creation was failed
4. TopicPoliciesService was closed

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
@@ -741,9 +741,17 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
             return;
         }
 
+        TopicPolicyMessageHandlerTracker topicPolicyMessageHandlerTracker =
+                topicPolicyMessageHandlerTrackers.remove(namespace);
+        if (topicPolicyMessageHandlerTracker != null) {
+            topicPolicyMessageHandlerTracker.close();
+        }
+
         CompletableFuture<SystemTopicClient.Reader<PulsarEvent>> readerFuture = readerCaches.remove(namespace);
         policyCacheInitMap.compute(namespace, (k, v) -> {
             policiesCache.entrySet().removeIf(entry -> Objects.equals(entry.getKey().getNamespaceObject(), namespace));
+            globalPoliciesCache.entrySet()
+                    .removeIf(entry -> Objects.equals(entry.getKey().getNamespaceObject(), namespace));
             return null;
         });
         if (readerFuture != null && !readerFuture.isCompletedExceptionally()) {
@@ -764,10 +772,6 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
         ownedBundlesCountPerNamespace.remove(namespace);
     }
 
-
-    private void cleanCacheAndCloseReader(@NonNull NamespaceName namespace, boolean cleanOwnedBundlesCount) {
-        cleanCacheAndCloseReader(namespace, cleanOwnedBundlesCount, false);
-    }
 
     private void cleanCacheAndCloseReader(@NonNull NamespaceName namespace, boolean cleanOwnedBundlesCount,
                                           boolean cleanWriterCache) {


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -526,27 +526,30 @@ private void initPolicesCache(SystemTopicClient.Reader<PulsarEvent> reader, Comp
...
         }
-        if (readerFuture != null && !readerFuture.isCompletedExceptionally()) {
-            readerFuture.thenCompose(SystemTopicClient.Reader::closeAsync)
-                    .exceptionally(ex -> {
-                        log.warn("[{}] Close change_event reader fail.", namespace, ex);
-                        return null;
-                    });
-        }
 
         policyCacheInitMap.compute(namespace, (k, v) -> {
```
nodece: @dao-jun Please move this logic inside the `cleanReaderCache` condition. Otherwise, there's a risk of partial policy loss: You're always cleaning up the `policiesCache`, and when the reader is not closed, the cache state becomes inconsistent.

nodece (approved):

codelipenghui:
> For some reason, all the Readers didn't closed successfully(maybe we were unload the ns at the time), after the change_events partition transferred to a new broker, all the Readers was reconnected to the new broker, and never be closed.

@dao-jun Do you see the logs from https://github.com/apache/pulsar/pull/24658/files#diff-9d2948d863c111e4be6d508a1c573667a1326b98c4314e917ba9e344bb61dc27L546 ? or what is the reason for the reader close failure? The behavior you mentioned seems not expected. If user triggered the reader.close(), it should not reconnect again for any reason.

dao-jun:
> > For some reason, all the Readers didn't closed successfully(maybe we were unload the ns at the time), after the change_events partition transferred to a new broker, all the Readers was reconnected to the new broker, and never be closed.
> 
> @dao-jun Do you see the logs from https://github.com/apache/pulsar/pull/24658/files#diff-9d2948d863c111e4be6d508a1c573667a1326b98c4314e917ba9e344bb61dc27L546 ? or what is the reason for the reader close failure? The behavior you mentioned seems not expected. If user triggered the reader.close(), it should not reconnect again for any reason.

I don't have a chance to find out the pulsar client's bug, @nodece will handle it.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -526,32 +533,41 @@ private void initPolicesCache(SystemTopicClient.Reader<PulsarEvent> reader, Comp
...
 
+        CompletableFuture<SystemTopicClient.Reader<PulsarEvent>> readerFuture = readerCaches.remove(namespace);
         policyCacheInitMap.compute(namespace, (k, v) -> {
             policiesCache.entrySet().removeIf(entry -> Objects.equals(entry.getKey().getNamespaceObject(), namespace));
             return null;
         });
+        if (readerFuture != null && !readerFuture.isCompletedExceptionally()) {
+            readerFuture
+                    .thenCompose(r -> {
+                        log.info("[{}] Closing the topic policies reader.", namespace);
```
BewareMyPower: This extra log might be redundant.  There is a `Closed consumer` log here when closing: https://github.com/apache/pulsar/blob/f2709bc823c7d048065dfb3d2845e47bcb0e3e0f/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1221

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -526,32 +533,41 @@ private void initPolicesCache(SystemTopicClient.Reader<PulsarEvent> reader, Comp
...
         }
-        if (readerFuture != null && !readerFuture.isCompletedExceptionally()) {
-            readerFuture.thenCompose(SystemTopicClient.Reader::closeAsync)
-                    .exceptionally(ex -> {
-                        log.warn("[{}] Close change_event reader fail.", namespace, ex);
-                        return null;
-                    });
+
+        if (!cleanReaderCache) {
+            return;
```
BewareMyPower: It will skip the cleanup on `policyCacheInitMap` as well. Is it desired?

BewareMyPower (commented):
Adding a new parameter and keeping two overloads of `cleanCacheAndCloseReader` (now `cleanCache`) makes code hard to read.

Method invocations like `clean(ns, true, false)` or `clean(ns, true, true)` really harms the readability. Since you're already touching this part, could you do a further refactoring by splitting the original `cleanCacheAndCloseReader` method into three methods? For example:
- `cleanWriterCache`
- `cleanReaderCache`
- `cleanPolicyInitCache`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -473,15 +478,16 @@ public boolean test(NamespaceBundle namespaceBundle) {
...
             return;
         }
         reader.hasMoreEventsAsync().whenComplete((hasMore, ex) -> {
             if (ex != null) {
                 log.error("[{}] Failed to check the move events for the system topic",
                         reader.getSystemTopic().getTopicName(), ex);
                 future.completeExceptionally(ex);
-                cleanCacheAndCloseReader(reader.getSystemTopic().getTopicName().getNamespaceObject(), false);
+                cleanCache(reader.getSystemTopic().getTopicName().getNamespaceObject(), false,
+                        isAlreadyClosedException(ex));
```
BewareMyPower: > Thousands of get last message id failed was thrown due to No such ledger exception.

This should fail the `hasMoreEventsAsync` method in `initPolicesCache`. After the change here, the reader won't be removed from the cache. Could you explain how it fixes the issue you encountered?

If the failed reader was kept in cache after that, next time `hasMoreEventsAsync` should fail again because `No such ledger` should not be a temporary error. If this issue is not resolved, keeping the failed reader in cache could only save a reader creation but would still face the `hasMoreEventsAsync` error and GetLastMessageId RPC failure.

dao-jun: Yes, it not solved the root cause. The root cause was introduced by topic compaction. I'm working on it.

But this modification can avoid making the situation worse. I mean, it will not try to create thousands of readers anymore.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
BewareMyPower: I realized they are coupled:

```java
                            return policyCacheInitMap.computeIfAbsent(namespace, (k) -> {
                                final CompletableFuture<SystemTopicClient.Reader<PulsarEvent>> readerCompletableFuture =
                                        createSystemTopicClient(namespace);
                                readerCaches.put(namespace, readerCompletableFuture);
```

It's really not a good design.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
BewareMyPower: Yes, it makes sense to avoid creating the reader if the previous creation failed and no ownership switching happened after that. But could you add a test to mock this case? I mean, we should guarantee `getTopicPoliciesAsync` will still return a value that might be outdated rather than fail in this case.

BewareMyPower: BTW, could you address the comment here https://github.com/apache/pulsar/pull/24658?notification_referrer_id=NT_kwDOARXIg7QxODM5MzEyMjI2ODoxODIwNDgwMw&notifications_query=is%3Aunread#pullrequestreview-3149250672 as well? `cleanCache(namespace, true, true, true);` is really hard to read, just like the `close` method of `PersistentTopic`.

I admit IDE could show the parameter names, but it's still not clear to know what the boolean argument represent. For example, in my previous comment https://github.com/apache/pulsar/pull/24658?notification_referrer_id=NT_kwDOARXIg7QxODM5MzEyMjI2ODoxODIwNDgwMw&notifications_query=is%3Aunread#discussion_r2296574279, it surprised me that it's reasonable to skip clearing `policyCacheInitMap` if we're going to skip clearing `readerCache`. It would be more readable to add a separated method to do that rather than using a boolean flag to control the behavior.

The existing code looks like:

```java
void cleanup(String key, boolean closeA, boolean closeB) {
    if (closeA()) {
        doCloseA();
    }
    if (!closeB) {
        closeC();
        return; // it's really confusing here
    }
    closeB();
    closeCInAlternativeApproach();
}
```

And it can be called like `cleanup(key, true, true)`, `cleanup(key, true, false)`, `cleanup(key, false, true)`, `cleanup(key, false, false)`. Actually, after this PR, it will have 2^3 = 8 possible combinations.

Separating `cleanup` to two methods `closeA()` and `closeBAndC()` will be more clear.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -526,32 +533,39 @@ private void initPolicesCache(SystemTopicClient.Reader<PulsarEvent> reader, Comp
...
         }
-        if (readerFuture != null && !readerFuture.isCompletedExceptionally()) {
-            readerFuture.thenCompose(SystemTopicClient.Reader::closeAsync)
-                    .exceptionally(ex -> {
-                        log.warn("[{}] Close change_event reader fail.", namespace, ex);
-                        return null;
-                    });
+
+        if (!cleanReaderCache) {
+            policyCacheInitMap.remove(namespace);
```
BewareMyPower, line 740: Removing the cache from `policyCacheInitMap` could trigger the reader creation again.

```java
                            return policyCacheInitMap.computeIfAbsent(namespace, (k) -> {
                                final CompletableFuture<SystemTopicClient.Reader<PulsarEvent>> readerCompletableFuture =
                                        createSystemTopicClient(namespace);
                                readerCaches.put(namespace, readerCompletableFuture);
```

nodece: We can remove cache from `policyCacheInitMap`, and improve the reader creation in the `prepareInitPoliciesCacheAsync`:
```
return policyCacheInitMap.computeIfAbsent(namespace, (k) -> {
    var readerCompletableFuture = readerCaches.compute(namespace, (__, v) -> {
        if (v == null || v.isCompletedExceptionally()) {
            CompletableFuture<SystemTopicClient.Reader<PulsarEvent>>
                    future = createSystemTopicClient(namespace);
            future.exceptionally(ex -> {
                log.error("[{}] Failed to create reader on __change_events topic",
                        namespace, ex);
                return null;
            });
            return future;
        }
        return v;
    });
    // todo
```

BewareMyPower: Yes, this should work. But should you also check if the exception is "already closed" when `v.isCompletedExceptionally()` returns true?

nodece: @BewareMyPower Good catch! This is just a quick check.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
dao-jun: It is makes sense to refactor the method, may I improve it with a split PR?
just make this PR well scoped.
@BewareMyPower

BewareMyPower: You've already touched this existing method by adding a new parameter, which makes the already hard-to-read code harder to read, so I think it's okay to apply the refactoring in this PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -419,6 +418,22 @@ public void addOwnedNamespaceBundleAsync(NamespaceBundle namespaceBundle) {
...
     }
 
+    private CompletableFuture<SystemTopicClient.Reader<PulsarEvent>> newReader(NamespaceName ns) {
+        return readerCaches.compute(ns, (__, existingFuture) -> {
+            if (existingFuture == null) {
+                return createSystemTopicClient(ns);
+            }
+
+            if (existingFuture.isCompletedExceptionally()) {
+                if (isAlreadyClosedException(existingFuture)) {
```
BewareMyPower: This should be applied to line 406 as well

```
                                        cleanCache(namespace, false,
                                                readerCompletableFuture.isCompletedExceptionally());
```

Let me submit a patch directly

BewareMyPower: Sorry please ignore this comment

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
BewareMyPower: I've drafted a change, could you take a look?

```diff
diff --git a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
index e8f6f5c0c4..231c9ed93e 100644
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
@@ -403,8 +403,11 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
                                         if (closed.get()) {
                                             return null;
                                         }
-                                        cleanCache(namespace, false,
-                                                readerCompletableFuture.isCompletedExceptionally());
+                                        if (readerCompletableFuture.isCompletedExceptionally()) {
+                                            cleanReaderCache(namespace);
+                                        } else {
+                                            policyCacheInitMap.remove(namespace);
+                                        }
                                     } catch (Throwable cleanupEx) {
                                         // Adding this catch to avoid break callback chain
                                         log.error("[{}] Failed to cleanup reader on __change_events topic",
@@ -458,7 +461,9 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
         }
         AtomicInteger bundlesCount = ownedBundlesCountPerNamespace.get(namespace);
         if (bundlesCount == null || bundlesCount.decrementAndGet() <= 0) {
-            cleanCache(namespace, true, true, true);
+            writerCaches.synchronous().invalidate(namespace);
+            ownedBundlesCountPerNamespace.remove(namespace);
+            cleanReaderCache(namespace);
         }
     }

@@ -488,7 +493,7 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
     private void initPolicesCache(SystemTopicClient.Reader<PulsarEvent> reader, CompletableFuture<Void> future) {
         if (closed.get()) {
             future.completeExceptionally(new BrokerServiceException(getClass().getName() + " is closed."));
-            cleanCache(reader.getSystemTopic().getTopicName().getNamespaceObject(), false, true);
+            cleanReaderCache(reader.getSystemTopic().getTopicName().getNamespaceObject());
             return;
         }
         reader.hasMoreEventsAsync().whenComplete((hasMore, ex) -> {
@@ -496,8 +501,12 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
                 log.error("[{}] Failed to check the move events for the system topic",
                         reader.getSystemTopic().getTopicName(), ex);
                 future.completeExceptionally(ex);
-                cleanCache(reader.getSystemTopic().getTopicName().getNamespaceObject(), false,
-                        isAlreadyClosedException(ex));
+                final var namespace = reader.getSystemTopic().getTopicName().getNamespaceObject();
+                if (isAlreadyClosedException(ex)) {
+                    cleanReaderCache(namespace);
+                } else {
+                    policyCacheInitMap.remove(namespace);
+                }
                 return;
             }
             if (hasMore) {
@@ -516,8 +525,9 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
                     log.error("[{}] Failed to read event from the system topic.",
                             reader.getSystemTopic().getTopicName(), e);
                     future.completeExceptionally(e);
-                    cleanCache(reader.getSystemTopic().getTopicName().getNamespaceObject(), false,
-                            isAlreadyClosedException(e));
+                    if (isAlreadyClosedException(e)) {
+                        cleanReaderCache(reader.getSystemTopic().getTopicName().getNamespaceObject());
+                    }
                     return null;
                 });
             } else {
@@ -543,26 +553,7 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
         });
     }

-    private void cleanCache(@NonNull NamespaceName namespace, boolean cleanOwnedBundlesCount,
-                            boolean closeReaderCache) {
-        cleanCache(namespace, cleanOwnedBundlesCount, false, closeReaderCache);
-    }
-
-    private void cleanCache(@NonNull NamespaceName namespace, boolean cleanOwnedBundlesCount,
-                            boolean cleanWriterCache, boolean cleanReaderCache) {
-        if (cleanWriterCache) {
-            writerCaches.synchronous().invalidate(namespace);
-        }
-
-        if (cleanOwnedBundlesCount) {
-            ownedBundlesCountPerNamespace.remove(namespace);
-        }
-
-        if (!cleanReaderCache) {
-            policyCacheInitMap.remove(namespace);
-            return;
-        }
-
+    private void cleanReaderCache(@NonNull NamespaceName namespace) {
         CompletableFuture<SystemTopicClient.Reader<PulsarEvent>> readerFuture = readerCaches.remove(namespace);
         policyCacheInitMap.compute(namespace, (k, v) -> {
             policiesCache.entrySet().removeIf(entry -> Objects.equals(entry.getKey().getNamespaceObject(), namespace));
@@ -586,7 +577,7 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
      */
     private void readMorePoliciesAsync(SystemTopicClient.Reader<PulsarEvent> reader) {
         if (closed.get()) {
-            cleanCache(reader.getSystemTopic().getTopicName().getNamespaceObject(), false, true);
+            cleanReaderCache(reader.getSystemTopic().getTopicName().getNamespaceObject());
             return;
         }
         reader.readNextAsync()
@@ -605,8 +596,7 @@ public class SystemTopicBasedTopicPoliciesService implements TopicPoliciesServic
                         if (isAlreadyClosedException(ex)) {
                             log.info("Closing the topic policies reader for {}",
                                     reader.getSystemTopic().getTopicName());
-                            cleanCache(
-                                    reader.getSystemTopic().getTopicName().getNamespaceObject(), false, true);
+                            cleanReaderCache(reader.getSystemTopic().getTopicName().getNamespaceObject());
                         } else {
                             log.warn("Read more topic polices exception, read again.", ex);
                             readMorePoliciesAsync(reader);
```

As you can see, there is actually three cases:
1. Clear the reader cache
2. Clear the writer cache, reader cache and owner bundle cache
3. Clear the reader cache if the exception is "already closed", otherwise, just clear `policyCacheInitMap`

With the new `cleanCache` method you've added in this PR, there could be technically 8 possible combinations, which makes code even harder to read than existing code.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -419,6 +418,22 @@ public void addOwnedNamespaceBundleAsync(NamespaceBundle namespaceBundle) {
...
+                    return existingFuture;
+                }
+                return createSystemTopicClient(ns);
+            }
```
BewareMyPower: A minor suggestion but not by force. When handling a failed future, you don't have to call `CompletableFuture#get` and handle the `InterruptedException` or `ExecutionException`.

```java
            if (existingFuture.isCompletedExceptionally()) {
                return existingFuture.exceptionallyCompose(e -> {
                    if (isAlreadyClosedException(e)) {
                        return existingFuture;
                    } else {
                        return createSystemTopicClient(ns);
                    }
                });
            }
```

Then you won't need a complicated overload of `isAlreadyClosedException`

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/TopicPoliciesTest.java
```diff
@@ -3577,9 +3579,13 @@ private void clearTopicPoliciesCache() {
         Map<TopicName, TopicPolicies> globalPoliciesCache =
                 WhiteboxImpl.getInternalState(topicPoliciesService, "globalPoliciesCache");
 
+        Map<NamespaceName, CompletableFuture<SystemTopicClient.Reader<PulsarEvent>>> readerCaches = WhiteboxImpl
+                .getInternalState(topicPoliciesService, "readerCaches");
```
BewareMyPower: Don't use reflection, just expose the field's visibility for tests.

I know existing code has used reflection as well, it's definitely a result of bad quality reviews, but please don't go far in a wrong way.

BewareMyPower (changes requested):

Technoboy- (changes requested):
We need to find the root cause of the issue, this fix is just a refactor

nodece:
@Technoboy- Makes sense. I agree this is an improvement to avoid creating the subscription multiple times, but we still need to find the root cause of the issue.

dao-jun:
@Technoboy- 

The root cause is `No such ledger` exception happened while reading compacted ledger.

At broker 1, the ledger ensemble is:
image
After a few minutes, `change_events-partition-0` transferred to another broker, at the time, the ensemble is :
image

lhotari:
There seems to be a remaining race condition issue: a namespace might be getting removed while it is being initialized again. One possible solution to this would be to break down the logic into another class. In the map for namespaces, the value could be an instance of this class that is in different states. This way it would be possible to properly handle different cases and avoid race conditions.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicBasedTopicPoliciesService.java
```diff
@@ -754,6 +805,9 @@ private void cleanCacheAndCloseReader(@NonNull NamespaceName namespace, boolean
         });
     }
 
+
```
nodece, line 808: Please remove the unnecessary blank.

nodece (approved):

dao-jun:
> There seems to be a remaining race condition issue: a namespace might be getting removed while it is being initialized again. One possible solution to this would be to break down the logic into another class. In the map for namespaces, the value could be an instance of this class that is in different states. This way it would be possible to properly handle different cases and avoid race conditions.

The concurrency control is based on `policyCacheInitMap`, and I remove the init future in the last step. I believe it's OK

lhotari:
Cherry-picking to branch-4.0 and branch-4.1 since #24980 depends on this PR.

Issue #24541, opened 2025-07-21
[Bug] Consumer listener gets stuck in an endless loop when sending DLQ message with mismatched schema


tonisojandu:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

* Broker version: 3.3.7
* Client library type: Java
* Client library version: 3.3.7
* Client Java version: 17

### Issue Description

This is an issue spun off from this original comment from another issue: https://github.com/apache/pulsar/issues/20635#issuecomment-1709616321

Even though that issue got resolved, we are still facing a problem where if
* we are using ALWAYS_COMPATIBLE schema compatibility on topic
* and an Avro message is sent with an incompatible message that fails to deserialize to a Specific Avro Record
then the Pulsar client listener tries to send the message to DLQ topic, but schema verification fails on the produce. This causes a reset of the message handling and another attempt and failure to deserialize and another attempt and failure to send to the DLQ topic. This will loop endlessly, which means this listener thread is effectively blocked from accepting further traffic.

### Error messages

```text
Exception from the loop produced with the example project linked below:

14:12:23.152 [pulsar-client-internal-14-1] WARN  o.a.pulsar.client.impl.ConsumerImpl - [persistent://public/default/test-topic] [test-subscription] [Lr0Ki] Failed to send DLQ message to dlq-topic for message id 3:0:-1
org.apache.pulsar.client.api.SchemaSerializationException: org.apache.avro.AvroTypeException: Found string, expecting long
        at org.apache.pulsar.client.impl.schema.reader.AbstractMultiVersionReader.read(AbstractMultiVersionReader.java:92)
        at org.apache.pulsar.client.impl.schema.AbstractStructSchema.decode(AbstractStructSchema.java:71)
        at org.apache.pulsar.client.impl.schema.AbstractStructSchema$WrappedVersionedSchema.decode(AbstractStructSchema.java:157)
        at org.apache.pulsar.client.api.Schema.validate(Schema.java:63)
        at org.apache.pulsar.client.impl.schema.AutoProduceBytesSchema.encode(AutoProduceBytesSchema.java:80)
[...]

Issue #20635, opened 2023-06-23
[Bug] pulsar keep creating dead letter queue producer and exceed the maximum limit


mingmcb:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Version
2.11

### Minimal reproduce step
Use pulsar java client library to create a consumer with dlq prodcer. 
1. after application starts, stop bookie
2. produce some message to the queue and trigger the consumer
3.  check the log. 
It seems keep creating dead letter queue producer, and eventually hit the maximum limit

### What did you expect to see?

extra producer should not be created if there is an issue on pulsar

### What did you see instead?

created over 10000 producers and eventually exceed the limits

### Anything else?
see logs

```
--
Starting Pulsar producer perf with config: {"topicName":"persistent://public/default/my-topic","producerName":null,"sendTimeoutMs":30000,"blockIfQueueFull":false,"maxPendingMessages":0,"maxPendingMessagesAcrossPartitions":0,"messageRoutingMode":"RoundRobinPartition","hashingScheme":"JavaStringHash","cryptoFailureAction":"FAIL","batchingMaxPublishDelayMicros":1000,"batchingPartitionSwitchFrequencyByPublishDelay":10,"batchingMaxMessages":1000,"batchingMaxBytes":131072,"batchingEnabled":true,"chunkingEnabled":false,"chunkMaxMessageSize":-1,"compressionType":"NONE","initialSequenceId":null,"autoUpdatePartitions":true,"autoUpdatePartitionsIntervalSeconds":60,"multiSchema":true,"accessMode":"Shared","lazyStartPartitionedProducers":false,"properties":{},"initialSubscriptionName":null}
--
[...]

PR #24663, opened 2025-08-23
[fix][client] Skip schema validation when sending messages to DLQ to avoid infinite loop when schema validation fails on an incoming message

3pacccccc:
Fixes https://github.com/apache/pulsar/issues/24541

### Motivation
When a consumer with a dead-letter policy (DLQ) encounters a message with a schema mismatch (e.g., producer sends Long but consumer expects String), and the message exceeds the maximum redelivery count, the following infinite loop occurs:
   1.  The consumer tries to send the message to the DLQ.
   2.  The DLQ send fails with a SchemaSerializationException due to the schema mismatch.
   3.  The consumer returns false from processPossibleToDLQ, causing the message to be redelivered.
   4.  The message is redelivered, exceeds the redelivery count again, and the process repeats indefinitely.
 
here's the example how to reproduce this issue:
```java
    public void sendDeadLetterTopicWithMismatchSchemaProducer() throws Exception {
        String namespace = BrokerTestUtil.newUniqueName("my-property/my-ns");
        admin.namespaces().createNamespace(namespace);
        // don't enforce schema validation
        admin.namespaces().setSchemaValidationEnforced(namespace, false);
        // set schema compatibility strategy to always compatible
        admin.namespaces().setSchemaCompatibilityStrategy(namespace, SchemaCompatibilityStrategy.ALWAYS_COMPATIBLE);
        String topic = BrokerTestUtil.newUniqueName("persistent://" + namespace
                + "/sendDeadLetterTopicWithMismatchSchemaProducer");

        // create topics
        admin.topics().createNonPartitionedTopic(topic);
        admin.topics().createNonPartitionedTopic(topic + "-DLQ");
        admin.topics().createNonPartitionedTopic(topic + "-RETRY");

        final int maxRedeliverCount = 1;
        final String subscriptionName = "my-subscription";
        Consumer<String> consumer = pulsarClient.newConsumer(Schema.AVRO(String.class))
                .topic(topic)
                .subscriptionName(subscriptionName)
                .deadLetterPolicy(DeadLetterPolicy.builder()
                        .deadLetterTopic(topic + "-DLQ")
                        .retryLetterTopic(topic + "-RETRY")
                        .maxRedeliverCount(maxRedeliverCount)
                        .build())
                .negativeAckRedeliveryDelay(1, TimeUnit.SECONDS)
                .messageListener((Consumer::negativeAcknowledge))
                .subscribe();

        Producer<Long> producer = pulsarClient.newProducer(Schema.AVRO(Long.class)).topic(topic).create();
        producer.send(1234567890L);
    }
```
and here's what's going on behind this scenario:

| producer                             | broker                          | consumer                                                                                                                                                                                            |
|--------------------------------------|---------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| sends message with AVRO(Long) schema |                                 |                                                                                                                                                                                                     |
|                                      | delivery to consumer            |                                                                                                                                                                                                     |
|                                      |                                 | received message and nack this message                                                                                                                                                              |
|                                      | delivery to consumer            |                                                                                                                                                                                                     |
|                                      |                                 | received message and nack this message                                                                                                                                                              |
|                                      | note: ***infinity loop start*** |
|                                      | delivery to consumer            |                                                                                                                                                                                                     |
|                                      |                                 | exceed ```maxRedeliverCount``` in ```deadLetterPolicy```, send to ```deadLetterTopic```                                                                                                             |
|                                      |                                 | send to ```deadLetterTopic``` failed due to mismatch schema<sup>[1]</sup>, and result will be false in ```ConsumerImpl.processPossibleToDLQ```<sup>[2]</sup>                                        |
|                                      |                                 | message will redelivery to broker<sup>[3]</sup>                                                                                                                                                     |
|                                      | delivery to consumer            |                                                                                                                                                                                                     |
|                                      |                                 | exceed ```maxRedeliverCount``` in ```deadLetterPolicy```, send to ```deadLetterTopic```                                                                                                             |
|                                      |                                 | send to ```deadLetterTopic``` failed with ```SchemaSerializationException```due to mismatch schema<sup>[1]</sup>, and result will be false in ```ConsumerImpl.processPossibleToDLQ```<sup>[2]</sup> |
|                                      |                                 | message will redelivery to broker<sup>[3]</sup>                                                                                                                                                     |
|                                      | delivery to consumer            |                                                                                                                                                                                                     |
|                                      |                                 | exceed ```maxRedeliverCount``` in ```deadLetterPolicy```, send to ```deadLetterTopic```                                                                                                             |
|                                      |                                 | send to ```deadLetterTopic``` failed due to mismatch schema<sup>[1]</sup>, and result will be false in ```ConsumerImpl.processPossibleToDLQ```<sup>[2]</sup>                                        |
|                                      |                                 | message will redelivery to broker<sup>[3]</sup>                                                                                                                                                     |
|                                      | ...                             |                                                                                                                                                                                                     |

[1]:https://github.com/apache/pulsar/blob/e606aeeebb6cc02f154758a9ae56fb85b621f8e0/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L2335
[2]:https://github.com/apache/pulsar/blob/e606aeeebb6cc02f154758a9ae56fb85b621f8e0/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L2379
[3]:https://github.com/apache/pulsar/blob/e606aeeebb6cc02f154758a9ae56fb85b621f8e0/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L2264
### Modifications
- Catch SchemaSerializationException in processPossibleToDLQ: When this exception occurs during DLQ send, mark the message as "processed" (i.e., complete with true) to prevent redelivery and break the loop.
- Improve logging:

   - Fix ambiguous debug log format from {}/{} to {}:{} for ledger and entry ID.

   - Unify and clarify error logs in processPossibleToDLQ to better distinguish between ProducerQueueIsFullError and other exceptions.
### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/24


pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2365,9 +2367,15 @@ private CompletableFuture<Boolean> processPossibleToDLQ(MessageIdAdv messageId)
...
                     } catch (Exception e) {
-                        log.warn("[{}] [{}] [{}] Failed to send DLQ message to {} for message id {}",
+                        log.warn("[{}] [{}] [{}] Failed to process DLQ message to {} for message id {}",
                                 topicName, subscription, consumerName, deadLetterPolicy.getDeadLetterTopic(), messageId,
                                 e);
+                        if (e instanceof SchemaSerializationException) {
+                            // If a SchemaSerializationException occurs, the message should be ignored to redelivery.
+                            // Otherwise, it will be redelivered and still can't send to the DLQ
+                            // and cause an infinite loop
+                            result.complete(true);
```
lhotari: Wouldn't this cause message loss?

3pacccccc: > Wouldn't this cause message loss?

@lhotari  Yes, it's true that this message might get lost. However, in this specific scenario, I couldn't find a better solution other than stopping the redelivery to the broker to break the infinite loop. Do you have any suggestions?

lhotari: @3pacccccc In https://github.com/apache/pulsar/issues/20635#issuecomment-2577625071 I have shared a suggestion: 
> Most likely there would need to be a separate mode for AUTO_PRODUCE_BYTES where it can be chosen on the client side to skip schema validation. That feature is also useful in other use cases since it can reduce CPU usage on the client side when schema validation is skipped at producing time.

I'm referring to skipping `schema.validate(message)` here:
https://github.com/apache/pulsar/blob/da0d11644f097ae657e79b2eb7835478d79ccd0e/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/AutoProduceBytesSchema.java#L70-L85

3pacccccc: @lhotari Thank you for the suggestion. This is indeed a better solution.

shall I simply remove the schema validation in ```AutoProduceBytesSchema``` directly? My concern is that this might impact the current behavior for existing users.

lhotari: > shall I simply remove the schema validation in `AutoProduceBytesSchema` directly? My concern is that this might impact the current behavior for existing users.

No, we shouldn't change the current behavior. We need another way to be able to configure it.
A simple solution would be to add a setter for `requireSchemaValidation` in `AutoProduceBytesSchema` class. 

Then it would be possible to disable validation this way in `org.apache.pulsar.client.impl.ConsumerImpl.initDeadLetterProducerIfNeeded` method:
```
                        AutoProduceBytesSchema<byte[]> autoProduceBytes =
                                (AutoProduceBytesSchema) Schema.AUTO_PRODUCE_BYTES(schema);
                        autoProduceBytes.setRequireSchemaValidation(false);
                        ProducerBuilder<byte[]> builder =
                                ((ProducerBuilderImpl<byte[]>) client.newProducer(autoProduceBytes))
```

3pacccccc: @lhotari Got it, thank you so much for your help!

pulsar-broker/src/test/java/org/apache/pulsar/client/api/DeadLetterTopicTest.java
```diff
@@ -1548,4 +1548,50 @@ public void testDeadLetterTopicWithMaxUnackedMessagesBlocking() throws Exception
...
+                .topic(topic)
+                .subscriptionType(SubscriptionType.Shared)
+                .subscriptionName(subscriptionName)
+                .deadLetterPolicy(DeadLetterPolicy.builder()
+                        .deadLetterTopic(topic + "-DLQ")
+                        .retryLetterTopic(topic + "-RETRY")
+                        .maxRedeliverCount(maxRedeliverCount)
+                        .build())
+                .negativeAckRedeliveryDelay(1, TimeUnit.SECONDS)
+                .messageListener((Consumer::negativeAcknowledge))
```
lhotari, line 1583: code style: there are unnecessary double parenthesis here?

pulsar-broker/src/test/java/org/apache/pulsar/client/api/DeadLetterTopicTest.java
```diff
@@ -1548,4 +1548,50 @@ public void testDeadLetterTopicWithMaxUnackedMessagesBlocking() throws Exception
...
+
+        Thread.sleep(3000L);
+
+        assertThat(pulsar.getBrokerService().getTopicReference(topic).get()
+                .getSubscription(subscriptionName).getConsumers().get(0).getMessageRedeliverCounter())
+                .describedAs("redeliver count of topic %s should be less than or equal to 2 because of mismatch schema",
+                        topic)
+                .isLessThanOrEqualTo(maxRedeliverCount + 1);
+        producer.close();
+        consumer.close();
```
lhotari, line 1608: At the end, it would be useful to test that it's possible to read the sent message from the DLQ using `Schema.AUTO_CONSUME` since it would be the way to handle such DLQs where there could be messages from different schemas in the same topic.

pulsar-broker/src/test/java/org/apache/pulsar/client/api/DeadLetterTopicTest.java
3pacccccc: fixed

pulsar-broker/src/test/java/org/apache/pulsar/client/api/DeadLetterTopicTest.java
3pacccccc: > it would be useful to test that it's possible to read the sent message from the DLQ using `Schema.AUTO_CONSUME` since it would be the way to handle such DLQs where there could be messages from different schemas in the same topic.

OK, I'll add test

3pacccccc:
@lhotari Hi, lari. I've updated the test, PTAL

lhotari (approved):

lhotari:
I ran some manual verification and things seem to work in an expected way. The message will get produced to the DLQ topic with the original schema version. It seems to be a property of the `AUTO_PRODUCE_BYTES` schema or `message.getReaderSchema()` that it's referring to the wrong schema when it attempts to run the validate. The solution in this PR seems to be a reasonable workaround to address this issue until we find a better solution.

One possible reason might be that getReaderSchema() might only be accurate in AUTO_CONSUME mode:
https://github.com/apache/pulsar/blob/1220951ac74fb4742abbbd331d6e751234c47015/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L436-L445

lhotari:
I think I found an answer to the wrong schema instance when using `AUTO_PRODUCE_BYTES` with `message.getReaderSchema().get()` as the schema.

https://github.com/apache/pulsar/blob/1220951ac74fb4742abbbd331d6e751234c47015/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L395-L413

For example, for Avro messages, this maps finally here:
https://github.com/apache/pulsar/blob/3b7bef1ecaf4004f36e071faaf31ddb49895115c/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/reader/MultiVersionAvroReader.java#L53-L54

It will always use the original Schema instance for reading and never refresh the "readerSchema" from the schema registry. This is why the mandatory validation before serializing the message to DLQ fails. Considering this fact, I think that that `AutoProduceBytesSchema.setRequireSchemaValidation(false)` is a very reasonable way to workaround this issue for the DLQ use case.

3pacccccc:
> I think I found an answer to the wrong schema instance when using `AUTO_PRODUCE_BYTES` with `message.getReaderSchema().get()` as the schema.
> 
> https://github.com/apache/pulsar/blob/1220951ac74fb4742abbbd331d6e751234c47015/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L395-L413
> 
> For example, for Avro messages, this maps finally here:
> 
> https://github.com/apache/pulsar/blob/3b7bef1ecaf4004f36e071faaf31ddb49895115c/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/reader/MultiVersionAvroReader.java#L53-L54
> 
> It will always use the original Schema instance for reading and never refresh the "readerSchema" from the schema registry. This is why the mandatory validation before serializing the message to DLQ fails. Considering this fact, I think that that `AutoProduceBytesSchema.setRequireSchemaValidation(false)` is a very reasonable way to workaround this issue for the DLQ use case.

@lhotari thank you so much for sharing this. this is very helpful and it's clear to me now. BTW, this solution is more reasonable than my initial attempt.

lhotari:
There's one remaining detail here. This solution will only work if SchemaCompatibilityStrategy is configured as `ALWAYS_COMPATIBLE` for the DLQ topic in the topic policy or in the namespace level. While testing this in action, it appeared that updating the SchemaCompatibilityStrategy changes at the namespace level aren't effective after a topic has been loaded. (example: this commit https://github.com/lhotari/pulsar-dlq-schema-issue-example/commit/5799194ca0c215b66997f730d1f079f0cb4436cd "fixed" the test app that was shared by @tonisojandu)

PR #17834, opened 2022-09-26
[improve][common] Make Bookkeeper metadata options configurable

lhotari:
### Motivation

PIP-45 related PRs #12770 and #13296 introduced changes in Pulsar and Bookkeeper configuration for Pulsar so that Pulsar Metadata driver is used for metadata operations in the bookkeeper client (in broker) and in the bookkeeper server (the bookies).
It would be good to have a way to opt-out of this change and configure Pulsar and Bookies without Pulsar Metadata store and use Zookeeper directly. 

This PR will allow to use Bookkeeper defaults by setting `export BK_METADATA_OPTIONS=none`.

If one would want to configure Bookies without Pulsar Metadata, it is necessary to configure
`metadataServiceUri` in bookkeeper.conf and `bookkeeperMetadataServiceUri` in broker.conf.

For example `bookkeeper.conf`
```properties
metadataServiceUri=zk://<zk_node_1>:2181/ledgers,zk://<zk_node_2>:2181/ledgers,zk://<zk_node_3>:2181/ledgers
```

The main purpose of this PR is to make it easier to compare the behavior of Pulsar when using Pulsar Metadata driver for Bookkeeper compared to using Zookeeper directly. It is not necessary make BK_METADATA_OPTIONS configurable to achieve this. However it's better to have an option to control it since that could prevent misconfiguration in cases where it is desired to not use Pulsar metadata driver. 

The additional benefit of being able to override BK_METADATA_OPTIONS is that the Pulsar Metadata driver classes wouldn't get loaded at all when they aren't used.

### Modifications

Make `BK_METADATA_OPTIONS` configurable.

### Documentation

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

### Matching PR in forked repository

PR in forked repository: https://github.com/lhotari/pulsar/pull/91

PR #24665, opened 2025-08-26
[fix][metadata] Fix missing call sync method causing data lost 

Technoboy-:
### Motivation

We're facing data loss during an cluster upgrade. The issue manifests as three distinct error types post-upgrade: BadVersion errors, "No Such Ledger" errors, and ledger authentication failures. The root cause appears to be a metadata synchronization issue where the Pulsar metadata store implementation lacks proper sync calls before listing ledgers.

As bk side, there is a `sync` call before `getChildren`
https://github.com/apache/bookkeeper/blob/2789316c18e12cbb6d17fa4a023410dbad6593a0/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ZkUtils.java#L284-L311

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/AbstractHierarchicalLedgerManager.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/AbstractHierarchicalLedgerManager.java
@@ -68,7 +68,8 @@ abstract class AbstractHierarchicalLedgerManager {
             final AsyncCallback.VoidCallback finalCb, final Object context,
             final int successRc, final int failureRc) {
 
-        store.getChildren(path)
+        store.sync(path)
+                .thenCompose(__ -> store.getChildren(path))
                 .thenAccept(levelNodes -> {
                     if (levelNodes.isEmpty()) {
                         finalCb.processResult(successRc, null, context);
@@ -162,7 +163,7 @@ abstract class AbstractHierarchicalLedgerManager {
      * Process ledgers in a single zk node.
      *
      * <p>
-     * for each ledger found in this zk node, processor#process(ledgerId) will be triggerred
+     * for each ledger found in this zk node, processor#process(ledgerId) will be triggered
      * to process a specific ledger. after all ledgers has been processed, the finalCb will
      * be called with provided context object. The RC passed to finalCb is decided by :
      * <ul>
@@ -188,7 +189,8 @@ abstract class AbstractHierarchicalLedgerManager {
             final String path, final BookkeeperInternalCallbacks.Processor<Long> processor,
             final AsyncCallback.VoidCallback finalCb, final Object ctx,
             final int successRc, final int failureRc) {
-        store.getChildren(path)
+        store.sync(path)
+                .thenCompose(__ -> store.getChildren(path))
                 .thenAccept(ledgerNodes -> {
                     Set<Long> activeLedgers = HierarchicalLedgerUtils.ledgerListToSet(ledgerNodes,
                             ledgerRootPath, path);
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/LegacyHierarchicalLedgerRangeIterator.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/LegacyHierarchicalLedgerRangeIterator.java
@@ -69,7 +69,7 @@ public class LegacyHierarchicalLedgerRangeIterator implements LedgerManager.Ledg
      * Iterate next level1 znode.
      *
      * @return false if have visited all level1 nodes
-     * @throws InterruptedException/KeeperException if error occurs reading zookeeper children
+     * @throws InterruptedException/ExecutionException/TimeoutException if error occurs reading zookeeper children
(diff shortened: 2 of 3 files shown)


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/AbstractHierarchicalLedgerManager.java
```diff
@@ -68,7 +68,8 @@ void asyncProcessLevelNodes(
             final AsyncCallback.VoidCallback finalCb, final Object context,
             final int successRc, final int failureRc) {
 
-        store.getChildren(path)
+        store.sync(path)
+                .thenCompose(__ -> store.getChildren(path))
```
lhotari: I don't think that this solves the issue. The reason is that `store.getChildren` will use the cache. The cache is updated asynchronously based on notifications so there's still a chance that the value isn't consistent.

lhotari:
It's great that we are finally resolving this issue. 

**Workaround for ZooKeeper, by-passing the Pulsar Metadata driver for BookKeeper client and BookKeeper server-side components**:
I faced this a few years ago. As a workaround at that time I introduced a way to by-pass the use of Pulsar Metadata Drivers in BookKeepers with this change: #17834. That prevents loading the drivers. In addition to that, it's necessary to configure BookKeeper to use direct ZK access without Pulsar Metadata Drivers. On the Broker (BookKeeper client), this requires setting `bookkeeperMetadataServiceUri: "zk+hierarchical://.."` and on BookKeeper components it's possible to set `metadataServiceUri: "zk+hierarchical://...` (and set the legacy values `zkServers` and `zkLedgersRootPath` to empty values since `metadataServiceUri` is the recommended way after [BP-29](https://bookkeeper.apache.org/bps/BP-29-metadata-store-api-module)).

In pulsar-helm-chart this is configurable and implemented this way:
broker side:
https://github.com/apache/pulsar-helm-chart/blob/ca43a3f1d1c160d0bb0127f3bafefcdffa7f422c/charts/pulsar/templates/broker-configmap.yaml#L40-L44
bookkeeper side:
https://github.com/apache/pulsar-helm-chart/blob/ca43a3f1d1c160d0bb0127f3bafefcdffa7f422c/charts/pulsar/templates/_bookkeeper.tpl#L134-L141

lhotari (changes requested):
More changes are needed to resolve the issue since store.getChildren(path) will use a cache that is updated asynchronously.

Another problem could be the potential performance implications. We might need to make changes on BookKeeper side so that caching would be safe under certain conditions if good performance relies on caching. Many environments have become reliant on the high level of caching, at least this is my understanding of the matter.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/AbstractHierarchicalLedgerManager.java
merlimat: Correct. We should be using `store.getChildrenFromStore()` (though it's currently protected), to bypass the cache completely.

Technoboy-: Updated, could you help review again ? @merlimat @lhotari

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
```diff
@@ -482,6 +482,31 @@ public void close() throws Exception {
         }
     }
 
+    @Override
+    public CompletableFuture<List<String>> getChildrenFromStore(String path) {
+        CompletableFuture<List<String>> cb = new CompletableFuture<>();
+        zkc.sync(path, (rc1, path1, ctx1) -> {
```
lhotari: Just wondering if we'd like to do `sync` before each and every `getChildrenFromStore` call? That would cause significant performance implications to the current behavior. Keeping the operations separate would let to choose when it's needed.

Technoboy-: ok, got it

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
```diff
@@ -482,6 +482,27 @@ public void close() throws Exception {
...
+                            .create(code, path1));
+                    cb.completeExceptionally(getException(code, path1));
+                    return;
+                }
+            }
+            cb.complete(nodes);
+        }, null);
+        return cb;
+    }
+
```
lhotari: I think that overriding this method isn't necessary. This would also have a performance impact since batching wouldn't be used if this is overridden.

Technoboy-: yes, right. removed. Thanks @lhotari

lhotari (approved):
LGTM

PR #24666, opened 2025-08-26
[improve][build] Increase maven resolver's sync context timeout

nodece:
### Motivation

During CI builds, especially when building Docker images or specific modules, Maven occasionally fails with:
```
Error:  Could not acquire lock(s)
Error:  java.lang.IllegalStateException: Could not acquire lock(s)
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error: Process completed with exit code 1.
```

This occurs even in concurrent(`-T 1C`) builds because Maven's default lock wait time (30 seconds) can be insufficient in some CI environments. Increasing the lock wait time prevents these intermittent failures.

### Modifications

- Added the JVM option `-Daether.syncContext.named.time=300` to increase the maximum lock wait time to 5 minutes.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,1 +1,1 @@
--Xss1500k
+-Xss1500k -Daether.syncContext.named.time=300


lhotari:
References:
* https://github.com/apache/maven/issues/9049
* https://maven.apache.org/resolver/maven-resolver-impl/synccontextfactory.html
* https://maven.apache.org/resolver/maven-resolver-named-locks/

lhotari (approved):
LGTM

lhotari:
@nodece In a local build I noticed a slow down and it seems to be in NAR file `[INFO] Generating documentation for NiFi extensions in the NAR...` step. 
For the master branch, we should make Java 21 the minimum JDK for building Pulsar so that we could start using `org.apache.nifi:nifi-nar-maven-plugin:2.0.0` which requires Java 21. Another option would be to fork https://github.com/apache/nifi-maven for Pulsar so that we could finally get a use of https://github.com/apache/nifi-maven/pull/35 which would speed up NAR file generation a lot.

lhotari:
I guess it would be possible to use maven profiles to use `org.apache.nifi:nifi-nar-maven-plugin:2.0.0` when Java 21 is used. I'll give it a try.

lhotari:
> I guess it would be possible to use maven profiles to use `org.apache.nifi:nifi-nar-maven-plugin:2.0.0` when Java 21 is used. I'll give it a try.

This is handled in #24668

PR #21759, opened 2023-12-19
[improve][broker] PIP-327: Support force topic loading for unrecoverable errors

rdhabalia:
It fixes https://github.com/apache/pulsar/issues/21751

PIP: #21752 

### Motivation

We have introduced a configuration called `autoSkipNonRecoverableData` before open-sourcing Pulsar as we have come across with various situations when it was not possible to recover ledgers belonging to managed-ledger or managed-cursors and the broker was not able to load the topics. In such situations,`autoSkipNonRecoverableData` flag helps to skip non-recoverable leger-recovery errors such as ledger_not_found and allows the broker to load topics by skipping such ledgers in disaster recovery.

Brokers can recognize such non-recoverable errors using bookkeeper error codes but in some cases, it’s very tricky and not possible to conclude non-recoverable errors. For example, the broker can not differentiate between all the ensemble bookies of the ledgers that are temporarily unavailable or are permanently removed from the cluster without graceful recovery, and because of that broker doesn’t consider all the bookies deleted as a non-recoverable error though we can not recover ledgers in such situations where all the bookies are removed due to various reasons such as Dev cluster clean up or system faced data disaster with multiple bookie loss. In such situations, the system admin has to manually identify such non-recoverable topics and update those topics’ managed-ledger and managed-cursor’s metadata and reload topics again which requires a lot of manual effort and sometimes it might not be feasible to handle such situations with a large number of topics that require this manual procedure to fix those topics.

```
2023-11-02T00:35:35,582+0000 [BookKeeperClientWorker-OrderedExecutor-5-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedCursorImpl - [prop/us-west2/ns1/per
sistent/t1-partition-220] Opened ledger 94267 for consumer pulsar.repl.dev-eastb. rc=0
2023-11-02T00:35:35,582+0000 [BookKeeperClientWorker-OrderedExecutor-5-0] INFO  org.apache.bookkeeper.client.DefaultBookieAddressResolver - Cannot resolve euw1bk--prod-booki
e-0.euw1bk--prod-bookie.bk-.svc.cluster.local:3181, bookie is unknown org.apache.bookkeeper.client.BKException$BKBookieHandleNotAvailableException: Bookie handle i
s not available
2023-11-02T00:35:35,582+0000 [BookKeeperClientWorker-OrderedExecutor-5-0] ERROR org.apache.bookkeeper.proto.PerChannelBookieClient - Cannot connect to euw1bk--prod-bookie-0.
[...]

PR #21752, opened 2023-12-19
[improve][pip] PIP-327: Support force topic loading for unrecoverable errors

rdhabalia:
### Motivation

It fixes https://github.com/apache/pulsar/issues/21751

PR #24672, opened 2025-08-26
fix: add missing managedLedgerForceRecovery broker configuration field

thomas-bousquet:
Main Issue: #21759

PIP: #21752 

### Motivation

`managedLedgerForceRecovery` is a broker configuration field, that can be set dynamically but it is missing when we want it to be set statically through broker configuration and applied when the application/container starts.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1360,6 +1360,10 @@ managedLedgerInfoCompressionThresholdInBytes=16384
 # corrupted at bookkeeper and managed-cursor is stuck at that ledger.
 autoSkipNonRecoverableData=false
 
+# Whether to allow brokers to forcefully load topics by skipping ledger failures to avoid topic unavailability and 
+# perform auto repairs of the topics.
+managedLedgerForceRecovery=false
+
 # Whether to recover cursors lazily when trying to recover a managed ledger backing a persistent topic.
 # It can improve write availability of topics.
 # The caveat is now when recovered ledger is ready to write we're not sure if all old consumers last mark


conf/broker.conf
```diff
@@ -1360,6 +1360,10 @@ managedLedgerInfoCompressionThresholdInBytes=16384
 # corrupted at bookkeeper and managed-cursor is stuck at that ledger.
 autoSkipNonRecoverableData=false
 
+# Whether to allow brokers to forcefully load topics by skipping ledger failures to avoid topic unavailability and 
+# perform auto repairs of the topics.
+managedLedgerForceRecovery=false
+
```
lhotari: please also add this to `standalone.conf`. That is the default config file used for `pulsar standalone`.

lhotari:
```
# Whether to allow brokers to forcefully load topics by skipping ledger failures to avoid topic unavailability and 
# perform auto repairs of the topics.
```

btw. The original description of this property isn't that great. It's missing that detail that the tradeoff of using this setting is that data is lost permanently.

When there are ledger failures, there's some other underlying issue that is causing them. One potential issue that is being addressed is https://github.com/apache/pulsar/pull/24665 . My assumption is that some ledger failures, could be resolved by restarting the BookKeeper cluster or restarting Pulsar cluster when the failures are caused by invalid in-memory state.
For 24665, there's a workaround to configure Pulsar cluster's BookKeeper client to by-pass the metadata store layer and use ZooKeeper directly. An the BookKeeper side to by-pass the Pulsar metadata store layer and use ZooKeeper directly. There's isn't yet a workaround when Oxia is used.

There are also other issues that might cause ledger failures. For example, BookKeeper 4.17.2 (included in Pulsar 4.0.6 release) contained a fix for a data loss issue, https://github.com/apache/bookkeeper/pull/4607.

conf/broker.conf
thomas-bousquet: 👍  I added it there and a mention in regards to the tradeoffs of enabling the option.

PR #24675, opened 2025-08-27
[improve][pip] PIP-438: add namespace topics limit metric

alexandrebrg:
### Motivation

We recently came with an issue where namespaces were reaching their limits, errors are lifted to producers and cannot be fixed by them (need either to delete topics or increase namespace limit). This PR creates a new metric to give topics limit for each namespace loaded in a broker. It will allow to correlate the current topic count (already existing metric) and the limit to determine the saturation of a namespace.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository

### DIFF
--- a/pip/pip-438.md
+++ b/pip/pip-438.md
@@ -67,5 +67,5 @@ No impact on geo-replication. Clusters can be upgraded independently.
 
 # Links
 
-* Mailing List discussion thread: [To be updated]
+* Mailing List discussion thread: https://lists.apache.org/thread/c3r2v2bhmtf82b0h1ff2pqc8wnosp7b8
 * Mailing List voting thread: [To be updated]
--- a/pip/pip-438.md
+++ b/pip/pip-438.md
@@ -1,0 +1,71 @@
+# PIP-437: Add namespace topics limit metric
+
+# Background knowledge
+
+Pulsar namespaces can be configured with a maximum number of topics. When this limit is reached, producers fail to create new topics. Currently, operators lack visibility into namespace topic limits through metrics.
+
+# Motivation
+
+Operators cannot proactively monitor namespace capacity because there's no metric exposing the configured topic limit. This leads to reactive troubleshooting when producers start failing.
+
+# Goals
+
+## In Scope
+
+- Add `pulsar_namespace_max_topics_limit` metric exposing the configured maximum topics limit for each namespace
+- Include cluster and namespace labels for correlation with existing topic count metrics
+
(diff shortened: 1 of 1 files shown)


pip/pip-438.md
```diff
@@ -0,0 +1,71 @@
...
+each export of metrics.
+
+# Detailed Design
+
+## Design & Implementation Details
+
+- Create method `getNamespaceTopicLimit` method in `NamespaceStatsAggregator`, fetching from namespace policies their 
+  max topics limit, falling back to the default value set in the broker configuration
+- Create method `printNamespaceTopicLimitStats` in `NamespaceStatsAggregator` to print the configured limit for each
+  namespace
```
lhotari, line 32: This might require some more changes since topic stats contains only the topics that are currently loaded in the broker.

Since this is a cluster wide limit, it would be unnecessary work for each broker in the cluster to calculate the current amount of topics that are actually created in the namespace. One possible solution would be to handle this on the "leader broker" so that it's the only broker that creates these statistics.

The current solution for finding out the number of topics in the namespace might not be optimal. One possible solution would be to extend the MetadataStore interface to contain a method "getNumberOfChildren" which would use an optimal method for the particular MetadataStore implementation to find the number of child nodes. For example, on Zookeeper, there's a field on the parent ZNode that tells how many child znodes there are. This would be an efficient way at least on Zookeeper to find out how many topics there are in a namespace.

update: Since there's MetadataCache layer to list the topics in the namespace, this might not be a significant bottleneck. However, it would lead to the situation where all namespaces would get listed on the leader broker and kept in memory. That could be a problem in some cases and therefore having a way to get the number of children directly would be better. For MetadataStore implementation that don't support getting the number of child nodes, the cache could be used.

alexandrebrg: I believe this PIP might have been misunderstood, we are looking to implement a metric for each namespace where it would give insight on the maximum number of topics that can be created in a namespace (which is either a namespace policy related value or broker default configuration value).

I understand that a namespace can be bundled and the namespace can be scattered between multiple brokers, isn't there a way to determine a "leader" broker for a namespace based on the topics loaded ? Or any metadata related to it ? (e.g. bundleId if it exists?)

PR #24692, opened 2025-09-01
[improve][build] Support building docker images that use Wolfi base image which is glibc based

lhotari:
### Motivation

Mailing list thread: https://lists.apache.org/thread/281kwr6txg6zgtqx5lry0fcox66gky6n

One of the drawbacks of the Alpine based image is that it's based on the "musl" libc library which causes incompatibilities for binaries that are compiled for glibc library. In Pulsar, we have already resolved this for binaries distributed with Pulsar and Pulsar IO connectors.

However, there's another disadvantage. The memory allocation library (malloc) in Alpine doesn't contain any tuneables and in Alpine, there's no documentation how to tune malloc operations.

It's possible to replace the default malloc library with other allocation libraries, but this could cause other instability. 

For benchmarking purposes, it would be useful to be able to build a glibc based Docker image for Pulsar and specify [glibc memory allocation tuneables](https://sourceware.org/glibc/manual/latest/html_node/Memory-Allocation-Tunables.html).

Various glibc tuneables can be viewed with this command:
```
docker run --rm -it cgr.dev/chainguard/wolfi-base sh -c "/lib64/ld-linux-*.so.1 --list-tunables"
```
The reference: https://sourceware.org/glibc/manual/latest/html_node/Memory-Allocation-Tunables.html

For example, to enable Transparent Huge Pages (THP) for malloc allocations:
```
export GLIBC_TUNABLES=glibc.malloc.hugetlb=1
```

Great resources about Transparent Huge Pages in Linux:
* https://www.oxla.com/blog/using-huge-pages-in-linux-applications
* https://www.oxla.com/blog/using-huge-pages-in-linux-applications-part-2-transparent-hugepage
* https://kernel.org/doc/html/latest/admin-guide/mm/transhuge.html

Alpine musl malloc criticism:
* https://nickb.dev/blog/default-musl-allocator-considered-harmful-to-performance/
  * [HN discussion](https://news.ycombinator.com/item?id=45143347)

For performance testing, it might be useful to also set `glibc.malloc.mmap_threshold` to the bytes value of 2MB:
```
export GLIBC_TUNABLES=glibc.malloc.hugetlb=1:glibc.malloc.mmap_threshold=2097152
```

For k8s or Docker configurations `GLIBC_TUNABLES` would be passed with the value `glibc.malloc.hugetlb=1:glibc.malloc.mmap_threshold=2097152`.

The reason why this matters is that most of BookKeeper's and Pulsar's memory is allocated by Netty. Under the covers the allocations are made using `os::malloc` which maps to the glibc `malloc`. To get the performance benefits of Transparent Huge Pages, it's necessary that the allocations to the operating system are made using mmap with madvise for `MADV_HUGEPAGE`. The Java VM will handle this for the Java heap memory with `-XX:+TransparentHugePages`, but that's not sufficient, since it doesn't cover the allocations made by Netty and RocksDB. When the malloc library has support for THP, there are better ways to control the behavior. When using glibc, it will mark the large allocations with `MADV_HUGEPAGE` when `glibc.malloc.hugetlb=1` is used and closes that gap for Pulsar.

THP is also supported in many other allocators and for example Microsoft's mimalloc (https://github.com/microsoft/mimalloc) could be another option. However, I faced some stability issues with it on Alpine. It's possibly an issue when `LD_PRELOAD` contains both `gcompat` and `mimalloc` libraries. In the Alpine images, `gcompat` is required for supporting Netty native libraries which have been built for glibc. This compatibility issue is also solved by using the Wolfi image.

### Additional Information

See #24688 regarding profiling.

### Modifications

- add additional Dockerfiles `Dockerfile.wolfi` for `pulsar` and `pulsar-all` images.
  - Wolfi (cgr.dev/chainguard/wolfi-base) is used as the base image 
  - More information about Wolfi OS: https://github.com/wolfi-dev/
    - Wolfi uses `apk` for package management like Alpine and package names are similar
    - Wolfi is backed by [Chainguard](https://www.chainguard.dev/) and they release security updates to CVEs nightly
- make changes to the maven build that Wolfi based images will be built when `-Pdocker-wolfi` is passed to the command line
- add GitHub Actions workflow changes to build the Wolfi images when Dockerfile.wolfi changes
- add GitHub Actions workflow changes to support building Wolfi images and running integration and system tests with the Wolfi images when the workflow gets triggered manually ([example build](https://github.com/lhotari/pulsar/actions/runs/17402166887))

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docker/pulsar/Dockerfile.wolfi
+++ b/docker/pulsar/Dockerfile.wolfi
@@ -100,11 +100,11 @@ RUN apk upgrade --no-cache
 ARG PULSAR_CLIENT_PYTHON_VERSION
 RUN pip3 install --break-system-packages --no-cache-dir \
     --only-binary \
-    pyyaml \
     grpcio==1.73.1 \
     protobuf==6.31.1 \
     pulsar-client[all]==${PULSAR_CLIENT_PYTHON_VERSION} \
-    kazoo
+    kazoo \
+    pyyaml
 
 COPY --from=jvm /opt/jvm /opt/jvm
 ENV JAVA_HOME=/opt/jvm


docker/pulsar/build-scripts/remove-unnecessary-native-binaries.sh
```diff
@@ -41,9 +41,13 @@ unzip $ROCKSDB_JAR > /dev/null
...
 then
-  ROCKSDB_TO_KEEP=linux64-musl
+  ROCKSDB_TO_KEEP=linux64
 else
-  ROCKSDB_TO_KEEP=linux-$ARCH-musl
+  ROCKSDB_TO_KEEP=linux-$ARCH
+fi
+
+if [ "$1" != "glibc" ]; then
+  ROCKSDB_TO_KEEP="${ROCKSDB_TO_KEEP}-musl"
```
nodece, line 50: Do you miss arch check?

nodece (commented):
Overall, it looks good to me. Could you add a job to build the Wolfi base image in the GHA? It would be great to verify that as well.

docker/pulsar/build-scripts/remove-unnecessary-native-binaries.sh
lhotari: There's no need for the arch check here. This appends `-musl` to `ROCKSDB_TO_KEEP` unless `glibc` is passed as the first argument to the script. This is to retain the previous behavior while supporting build of `glibc` based image.
In `Dockerfile.wolfi`, it will call this script with `glibc` parameter.

lhotari:
> Overall, it looks good to me. Could you add a job to build the Wolfi base image in the GHA? It would be great to verify that as well.

@nodece I pushed changes to build the Wolfi images when there are file changes directly related to Dockerfile.wolfi files so that our CI doesn't become slower. There's also the choice to manually trigger a build that uses the Wolfi base images. PTAL

image

nodece (approved):

PR #24700, opened 2025-09-04
[fix][broker] Fix race condition between readMoreEntries and sendInProgress which could skip or delay reading more entries in Key_Shared

### LINKED ISSUE #23845: [Bug] Key_Shared subscription doesn't deliver messages in the replay queue when no new messages are produced

### LINKED ISSUE #24490: [Bug] Perf regression: ~15% latency increase in e2e tests with shared subscriptions in Pulsar 4.0.x
### Search before reporting

### Read release policy

### User environment

Pulsar 4.0.x

### Issue Description

After running series of perf tests to compare 3.0.x and 4.0.x we noticed ~15% e2e (from the time message produced to the time message consumed) avg latency increase in the tests that use shared subscription.

The regression disappeared after we set 
```
subscriptionKeySharedUseClassicPersistentImplementation=true 
subscriptionSharedUseClassicPersistentImplementation=true
```

This makes recent-ish refactoring (PIP-379 ) of shared/keyshared subscription the primary suspect.
@lhotari FYI

### Error messages

```text

```

### Reproducing the issue

Test names give rough description of what's being tested:

- Pulsar 1m 10 topics 3 partitions 100b shared subs
- Pulsar 400k 10 topics 3 partitions 1k shared subs

@sandeep-mst can help with repro if needed

### Additional information

_No response_

lhotari:
Fixes #23845
Fixes #24490

### Motivation

There a race condition in `PersistentDispatcherMultipleConsumers` where reading more entries might get skipped or delayed when there's a send in progress and new permits get added.  The race condition is explained in https://github.com/apache/pulsar/issues/23845#issuecomment-2599470326.

Besides causing consuming to get stuck, the race condition could also cause latency delays and could be contributing to #24490.

### Modifications

- add a flag `moreEntriesRequestedWhileSendInProgress` for detecting and resolving the race condition
  - the flag is set when `readMoreEntries` gets blocked by `sendInProgress`
  - if the flag is set, `handleSendingMessagesAndReadingMore` will always result in calling `readMoreEntries`

Tests: TBD

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -139,6 +139,8 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
     protected int lastNumberOfEntriesProcessed;
     protected boolean skipNextBackoff;
     private final Backoff retryBackoff;
+    private boolean moreEntriesRequestedWhileSendInProgress;
+
     protected enum ReadType {
         Normal, Replay
     }
@@ -341,8 +343,10 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
                 log.debug("[{}] [{}] Skipping read for the topic, Due to sending in-progress.",
                         topic.getName(), getSubscriptionName());
             }
+            moreEntriesRequestedWhileSendInProgress = true;
             return;
         }
+        moreEntriesRequestedWhileSendInProgress = false;
         if (shouldPauseDeliveryForDelayTracker()) {
             if (log.isDebugEnabled()) {
                 log.debug("[{}] [{}] Skipping read for the topic, Due to pause delivery for delay tracker.",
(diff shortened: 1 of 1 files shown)


lhotari:
Closing this since I couldn't reproduce the issue.

PR #24704, opened 2025-09-04
[improve][pip] PIP-439: Adding Transaction Support to Pulsar Functions Through Managed Transaction Wrapping

wzhramc:
# Motivation

This PIP proposes extending Pulsar Functions with configuration-based automatic/managed transaction support to enable exactly-once processing semantics. Currently, Pulsar Functions cannot publish to multiple topics transactionally, which is a significant limitation for use cases requiring atomic multi-topic operations.
  
- [X] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [X] The metrics

### Documentation

- [X] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-439.md
+++ b/pip/pip-439.md
@@ -1,4 +1,4 @@
-# PIP-439: Adding Transaction Support to Pulsar Functions Through Auto-Transaction Wrapping
+# PIP-439: Adding Transaction Support to Pulsar Functions Through Managed Transaction Wrapping
 
 # Background knowledge
 
@@ -70,6 +70,8 @@ Adding transaction support to Pulsar Functions would finally ensure message proc
 3. Support transactional acknowledgment of input messages
 4. Ensure transactions are committed only if message processing completes successfully
 5. Provide transaction timeout configuration for Functions
+6. Add transaction support for async functions
+7. Handling multiple transactions in batches to improve performance, added in a later phase of implementation
 
 ## Out of Scope
 
@@ -110,6 +112,8 @@ public enum TransactionMode {
 public class TransactionConfig {
   private TransactionMode transactionMode = TransactionMode.OFF;
   private Long transactionTimeoutMs = 60000L;
+  private Integer transactionBatchingMaxEntries = 1;
+  private Long transactionBatchingQuietPeriodMs = 100L;
 
   // Getters and setters...
 }
@@ -133,6 +137,8 @@ message TransactionSpec {
   }
   TransactionMode transactionMode = 1;
   int64 transactionTimeoutMs = 2;
+  int64 transactionBatchingMaxEntries = 3;
+  int64 transactionBatchingQuietPeriodMs = 4;
 }
 
 message FunctionDetails {
@@ -141,46 +147,18 @@ message FunctionDetails {
 }
 ```
 
-### Modifications to Context Interface
+### Modifications to ContextImpl
 
-We will update the Context interface to expose the current transaction:
-
-```java
-public interface Context {
-      // Existing methods...
-
-    /**
-     * Returns the current transaction if function is running in managed transaction mode.
-     *
-     * <p>IMPORTANT: This method is not async-safe. When writing asynchronous functions that
-     * return CompletableFuture and need to use the transaction inside callbacks or chained
-     * operations, you must store a reference to the transaction locally before returning the future:
-     *
-     * <pre>{@code
-     * public CompletableFuture<String> process(String input, Context context) {
-     *     // Store transaction reference locally before async operations
-     *     Transaction txn = context.getCurrentTransaction();
-     *
-     *     return someAsyncOperation()
-     *         .thenApply(result -> {
-     *             // Use the locally stored transaction reference
-     *             // instead of calling context.getCurrentTransaction() here
-     *               return "processed";
-     *         });
-     * }
-     * }</pre>
-     *
-     * @return the current transaction, or null if transactions are disabled
-     */
-    Transaction getCurrentTransaction();
-}
-```
 
 ```java
 class ContextImpl implements Context, SinkContext, SourceContext, AutoCloseable {
     // Existing fields...
 
-    private Transaction currentTransaction;
+    // Finds the proper transaction to tie to current function execution (sync/async)
+    private Transaction getManagedTransaction() {
+
+        // implementation...
+    }
 
     // Existing methods...
 
@@ -189,23 +167,19 @@ class ContextImpl implements Context, SinkContext, SourceContext, AutoCloseable 
     }
 
     @Override
-    public Transaction getCurrentTransaction() {
-        return currentTransaction;
-    }
-
-    @Override
     public <T> TypedMessageBuilder<T> newOutputMessage(String topicName, Schema<T> schema)
           throws PulsarClientException {
       MessageBuilderImpl<T> messageBuilder = new MessageBuilderImpl<>();
       TypedMessageBuilder<T> typedMessageBuilder;
       Producer<T> producer = getProducer(topicName, schema);
+      Transaction managedTransaction = getManagedTransaction(); 
     
       if (currentTransaction != null) {
           if (schema != null) {
               // Uses the new API that supports both schema and transaction
-              typedMessageBuilder = producer.newMessage(schema, currentTransaction);
+              typedMessageBuilder = producer.newMessage(schema, managedTransaction);
           } else {
-              typedMessageBuilder = producer.newMessage(currentTransaction);
+              typedMessageBuilder = producer.newMessage(managedTransaction);
           }
       } else if (schema != null) {
           typedMessageBuilder = producer.newMessage(schema);
@@ -226,8 +200,44 @@ It's important to note that Pulsar Functions supports asynchronous processing, w
 For asynchronous functions:
 1. The transaction is created at the beginning of message processing, just like for synchronous functions
 2. When the function returns a `CompletableFuture`, the transaction is maintained until the future completes
+ - Any Context-related operations inside of the returned 'CompletableFuture' objects are tied to the correct transaction
 3. When the future completes successfully, the transaction is committed
 4. If the future completes exceptionally, the transaction is aborted
+
+## Batch Processing of Transactions
+
+To optimize performance and reduce the overhead on the Transaction Coordinator, this proposal introduces transaction batching.
+Transaction batching allows multiple incoming messages to be processed within the same transaction, reducing the total number of
+transactions created.
+
+## Transaction Batching Concept
+
+Transaction batching is distinct from Pulsar's message batching. While message batching combines multiple messages into a single "batch
+message" for efficient network transfer, transaction batching processes multiple incoming messages (or batch messages) within the scope
+of a single transaction.
+
+Key benefits of transaction batching include:
+1. **Reduced Load on Transaction Coordinator**: Fewer transactions means less coordination overhead
+2. **Improved Throughput**: Higher message processing capacity with lower per-message overhead
+3. **Optimized Resource Usage**: Better utilization of transaction resources
+4. **Consistent Performance at Scale**: Maintains performance characteristics under high load
+
+### Transaction Batching Parameters
+
+Transaction batching is controlled by two main parameters:
+
+1. **`transactionBatchingMaxEntries`**: The maximum number of entries (incoming messages or batch messages) to process within a single
+transaction before committing it and starting a new one.
+ - An "entry" refers to an incoming batch message which could itself contain multiple individual messages
+ - Default: 1 (recommended minimum to ensure batch index acknowledgment state doesn't span transactions)
+ - Setting this to higher values increases throughput but may impact latency
+
+2. **`transactionBatchingQuietPeriodMs`**: The maximum amount of time to wait for additional messages before committing a transaction if
+`transactionBatchingMaxEntries` is not reached.
+ - Default: 100ms
+ - This parameter handles scenarios where message flow isn't continuous
+ - When the quiet period elapses without new messages arriving, the current transaction is committed
+
 
 # Public-facing Changes
 
@@ -239,11 +249,26 @@ Transaction support will be configured using the new TransactionConfig class:
 TransactionConfig txnConfig = new TransactionConfig();
 txnConfig.setTransactionMode(TransactionMode.MANAGED);
 txnConfig.setTransactionTimeoutMs(30000L); // 30 seconds
+txnConfig.setTransactionBatchingMaxEntries(10); // Process up to 10 entries per transaction
+txnConfig.setTransactionBatchingQuietPeriodMs(50L); // Commit after 50ms of inactivity
 
 FunctionConfig functionConfig = new FunctionConfig();
 functionConfig.setTransaction(txnConfig);
 // Other configuration...
 ```
+
+The Functions worker configuration will include a setting to en-/disable `pulsar_function_txn_latency`:
+
+```yaml
+# Transaction metrics configuration
+transactionMetrics:
+  # Enable/disable specific transaction metrics individually
+  enabledMetrics:
+    txnLatency: false          # Transaction duration histograms (high cardinality)
+    # ... disable any future metric deemed as too expensive to enable per default
+```
+
+
 
 ## CLI
 
@@ -253,6 +278,8 @@ The Pulsar Admin CLI will be updated to support these new configuration options:
 $ pulsar-admin functions create \
   --auto-transactions-enabled true \
   --transaction-timeout-ms 30000 \
+  --transaction-batching-max-entries 10 \
+  --transaction-batching-quiet-period-ms 50 \
   ... other options ...
 ```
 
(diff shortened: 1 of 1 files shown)


lhotari:
@wzhramc Please start a new thread on the dev mailing list about this proposal. You can reference the previous mailing list thread https://lists.apache.org/thread/rll8qyovpd7t9v5yxth25qo44zksbgkn in the new thread. 
One reason to start the new thread is to reserve PIP-439 for this proposal since the first proposal to start a discussion with the PIP number reserves it.

pip/pip-439.md
```diff
@@ -0,0 +1,573 @@
...
+        if (schema != null) {
+            typedMessageBuilder = producer.newMessage(schema);
+        } else {
+            typedMessageBuilder = producer.newMessage();
+        }
+
+        // If there's an active transaction, associate the message with it
+        if (currentTransaction != null) {
+            typedMessageBuilder = typedMessageBuilder.sendTransaction(currentTransaction);
+        }
```
lhotari: this isn't correct. the transaction needs to be passed as an argument to `producer.newMessage` there's a pending [PIP-407](https://github.com/apache/pulsar/blob/master/pip/pip-407.md) to specifying both schema and transaction. That would be needed.

lhotari: I'll get the implementation PR #23942 for PIP-407 merged so that you can adapt this.

pip/pip-439.md
```diff
@@ -0,0 +1,573 @@
...
+        // register end time
+        stats.processTimeEnd();
+
+        if (result != null) {
+            // process the synchronous results
+            handleResult(currentRecord, result);
+        }
+    }
+
+    private void processWithTransaction() {
```
lhotari: This doesn't look right. In Pulsar, you can asynchronously handle messages so that you have more than one message in flight at a time. It's important for performance at scale since it enables "pipelining".
That's why it's better to handle use the existing solution and refactor it to support transactions.

pip/pip-439.md
lhotari: PIP-407 implementation is now merged.

pip/pip-439.md
```diff
@@ -0,0 +1,573 @@
...
+3. Adding transaction support to Pulsar IO connectors
+4. Changes to the Function interface itself
+
+# High Level Design
+
+The proposed solution introduces automatic transaction wrapping for Pulsar Functions through configuration settings. When enabled, each function execution will be
+automatically wrapped in a transaction without requiring code changes to the function implementation.
+
+The general flow will be:
+1. Function is configured with `autoTransactionsEnabled: true`
```
lhotari: I'd suggest to replace the "auto transactions" concept with something that is more descriptive.
One possibility would be to call it "container managed transactions" or just "managed transactions". 

Since there's a plan to possibly support "explicit transactions" later, it would be better to use an enum instead of a boolean.


```java
    public enum TransactionMode {
        OFF,
        MANAGED
    }
```

There could be a TransactionConfig object to follow a similar pattern as there is for ProducerConfig:

```java
public class TransactionConfig {
  TransactionMode transactionMode;
  Long transactionTimeoutMs;
}
```

This would allow extending the configuration later if there becomes a need.

pip/pip-439.md
```diff
@@ -0,0 +1,573 @@
...
+
+```java
+message FunctionDetails {
+    // Other existing fields...
+
+    // Whether to automatically wrap function execution in a transaction
+    bool autoTransactionsEnabled = X;
+
+    // Default transaction timeout in milliseconds
+    int64 transactionTimeoutMs = Y;
```
lhotari: In protobuf, the above `TransactionConfig` would map to `TransactionSpec` message included in `FunctionDetails`.
```protobuf
message FunctionDetails {
    // Other existing fields...
    TransactionSpec transaction = 24;
}

message TransactionSpec {
    enum TransactionMode {
        OFF = 0;
        MANAGED = 1;
    }
    TransactionMode transactionMode = 1;
    int64 transactionTimeoutMs = 2;
}

File: pip/pip-439.md (hunk 1 above)
Comment (wzhramc) [reply pip/pip-439.md]: Thank you! I've adopted it in the newest update

File: pip/pip-439.md (hunk 2 above)
Comment (wzhramc) [reply pip/pip-439.md]: I've removed that section from the proposal since I wasn't aware of async support when I wrote the initial draft. The proposal now includes adding transactional support for async functions as well.

File: pip/pip-439.md (hunk 3 above)
Comment (wzhramc) [reply pip/pip-439.md]: Perfect suggestion! Updated the proposal accordingly.

File: pip/pip-439.md (hunk 4 above)
Comment (wzhramc) [reply pip/pip-439.md]: Updated the proposal accordingly here as well.

File: pip/pip-439.md [hunk 5]
```diff
@@ -0,0 +1,370 @@
...
+ - Labels: `tenant`, `namespace`, `name` (function name), `instance_id`, `cluster`
+ - Unit: Count
+
+4. `pulsar_function_txn_timeout_total`: Counter tracking transactions that timed out
+ - Labels: `tenant`, `namespace`, `name` (function name), `instance_id`, `cluster`
+ - Unit: Count
+
+5. `pulsar_function_txn_latency`: Histogram of transaction duration from creation to commit/abort
+ - Labels: `tenant`, `namespace`, `name` (function name), `instance_id`, `cluster`
+ - Unit: Milliseconds
```
lhotari: one concern about histograms is that they add a lot of volume to the metrics. it would be useful if this metric would be configurable in the functions worker and turned off by default.

lhotari (dismissed):
LGTM, just a minor comment about pulsar_function_txn_latency metric, so that it would be disabled by default to reduce the volume and cardinality of the metrics.

lhotari:
One potential future improvement could be to add support for transaction batching. In transaction batching, multiple incoming messages would be processed within the same transaction due to performance reasons so the number of transactions can be reduced. There would need to be `transactionBatchingMaxEntries` ("entry" would refer to an incoming batch message which could contain multiple messages) and `transactionBatchingQuietPeriodMs` configuration parameters to control this type of batching. In most cases, it would be useful to configure `transactionBatchingMaxEntries` to at least 1 so that batch index acknowledgement state wouldn't have to be held across transactions. The `transactionBatchingQuietPeriodMs` would control how long to wait until "closing" the transaction batch in case the incoming messages don't flow in continuously and `transactionBatchingMaxEntries` isn't hit.

pip/pip-439.md
```diff
@@ -0,0 +1,370 @@
...
+# General Notes
+
+The implementation of transaction support in Pulsar Functions should be considered a stepping stone toward a more comprehensive exactly-once processing model for
+Pulsar's stream processing capabilities. Future work may include Transaction support in Pulsar IO connectors.
+
+# Links
+
+- https://github.com/apache/pulsar/issues/24588
+- https://pulsar.apache.org/docs/txn-why/
+- https://lists.apache.org/thread/rll8qyovpd7t9v5yxth25qo44zksbgkn
```
lhotari: Add a label to this link "Mailing List discussion thread: "
(after the voting thread is created there would also be "Mailing List vote thread:")

lhotari:
> One potential future improvement could be to add support for transaction batching. In transaction batching, multiple incoming messages would be processed within the same transaction due to performance reasons so the number of transactions can be reduced. There would need to be `transactionBatchingMaxEntries` ("entry" would refer to an incoming batch message which could contain multiple messages) and `transactionBatchingQuietPeriodMs` configuration parameters to control this type of batching. In most cases, it would be useful to configure `transactionBatchingMaxEntries` to at least 1 so that batch index acknowledgement state wouldn't have to be held across transactions. The `transactionBatchingQuietPeriodMs` would control how long to wait until "closing" the transaction batch in case the incoming messages don't flow in continuously and `transactionBatchingMaxEntries` isn't hit.

@wzhramc For high-scale use cases, it might be necessary to implement "transaction batching" so that unnecessary load isn't added to the broker (including the transaction coordinator) when enabling transactions. It might be useful to cover the "transaction batching" aspect directly in PIP-439 if you'd be ready to do that before proceeding to the vote. It's also possible to handle that later in another PIP.

jiangpengcheng (approved):
LGTM, just have one question:

for an async function, and it will call the `context.newOutputMessage` method, how to make the `newOutputMessage` method to use the right transaction? shall we create a new method `newOutputMessage` and make it accept a transaction as an argument?

lhotari:
> LGTM, just have one question:
> 
> for an async function, and it will call the `context.newOutputMessage` method, how to make the `newOutputMessage` method to use the right transaction? shall we create a new method `newOutputMessage` and make it accept a transaction as an argument?

@jiangpengcheng The async function doesn't have access to the Transaction in managed transactions. I think that the context must have a reference so that any new messages would get added to the active transaction for that asynchronous call chain.

pip/pip-439.md
```diff
@@ -0,0 +1,370 @@
...
+     * Returns the current transaction if function is running in managed transaction mode.
+     *
+     * <p>IMPORTANT: This method is not async-safe. When writing asynchronous functions that
+     * return CompletableFuture and need to use the transaction inside callbacks or chained
+     * operations, you must store a reference to the transaction locally before returning the future:
+     *
+     * <pre>{@code
+     * public CompletableFuture<String> process(String input, Context context) {
+     *     // Store transaction reference locally before async operations
+     *     Transaction txn = context.getCurrentTransaction();
```
lhotari: This shouldn't be exposed at all. It should be possible to handle this also for asynchronous functions.

pip/pip-439.md
```diff
@@ -0,0 +1,370 @@
...
+     *             // Use the locally stored transaction reference
+     *             // instead of calling context.getCurrentTransaction() here
+     *               return "processed";
+     *         });
+     * }
+     * }</pre>
+     *
+     * @return the current transaction, or null if transactions are disabled
+     */
+    Transaction getCurrentTransaction();
```
lhotari: this shouldn't be exposed at all.

pip/pip-439.md
lhotari: It's possible to have a `Context` implementation that delegates to an underlying context, but holds a reference to the transaction that the asynchronous function invocation belongs to.

wzhramc:
> > One potential future improvement could be to add support for transaction batching. In transaction batching, multiple incoming messages would be processed within the same transaction due to performance reasons so the number of transactions can be reduced. There would need to be `transactionBatchingMaxEntries` ("entry" would refer to an incoming batch message which could contain multiple messages) and `transactionBatchingQuietPeriodMs` configuration parameters to control this type of batching. In most cases, it would be useful to configure `transactionBatchingMaxEntries` to at least 1 so that batch index acknowledgement state wouldn't have to be held across transactions. The `transactionBatchingQuietPeriodMs` would control how long to wait until "closing" the transaction batch in case the incoming messages don't flow in continuously and `transactionBatchingMaxEntries` isn't hit.
> 
> @wzhramc For high-scale use cases, it might be necessary to implement "transaction batching" so that unnecessary load isn't added to the broker (including the transaction coordinator) when enabling transactions. It might be useful to cover the "transaction batching" aspect directly in PIP-439 if you'd be ready to do that before proceeding to the vote. It's also possible to handle that later in another PIP.

I'm not really familiar with batch processing in Pulsar functions in general, so it would take a bit more time for me to figure it out... I would eventually need help with that as well. It would be nice if someone could help me out with this ^^'

lhotari:
> I'm not really familiar with batch processing in Pulsar functions in general, so it would take a bit more time for me to figure it out... I would eventually need help with that as well. It would be nice if someone could help me out with this ^^'

Transaction batching wouldn't be directly related to Pulsar batching. It's simply about using the same transaction to process multiple incoming messages. This would be necessary to achieve high performance without putting too much pressure on the transaction coordinator and adding too much latency to processing. We could proceed without having this in PIP-439 and come back to this after testing the performance.

pip/pip-439.md
lhotari: @jiangpengcheng I added some more review comments here that the transaction shouldn't be exposed at all. I missed that in my previous review. Thanks for the good question!

wzhramc:
> > I'm not really familiar with batch processing in Pulsar functions in general, so it would take a bit more time for me to figure it out... I would eventually need help with that as well. It would be nice if someone could help me out with this ^^'
> 
> Transaction batching wouldn't be directly related to Pulsar batching. It's simply about using the same transaction to process multiple incoming messages. This would be necessary to achieve high performance without putting too much pressure on the transaction coordinator and adding too much latency to processing. We could proceed without having this in PIP-439 and come back to this after testing the performance.

Ah! That makes a lot of sense! Sorry, I mixed it up with Pulsar batching. I can include it in the PIP, it would be pretty neat to have this right away.

pip/pip-439.md
wzhramc: My reason for adding this was that it made it easier for me to think about how to tie a transaction to an async function: by just letting the user have a handle to the transaction inside the async operation... But yeah, I've realized that this totally breaks the idea of having managed transactions in the first place when responsibility is shifted to the user.

lhotari:
> Ah! That makes a lot of sense! Sorry, I mixed it up with Pulsar batching. I can include it in the PIP, it would be pretty neat to have this right away.

It might be tricky to implement directly, so a phased implementation could be useful in any case. One benefit of having it in this PIP is that we could find a way how to make the function configuration "future proof". The transaction batching feature makes sense only in the MANAGED mode.

pip/pip-439.md
wzhramc: removed it in the latest update

wzhramc:
> > Ah! That makes a lot of sense! Sorry, I mixed it up with Pulsar batching. I can include it in the PIP, it would be pretty neat to have this right away.
> 
> It might be tricky to implement directly, so a phased implementation could be useful in any case. One benefit of having it in this PIP is that we could find a way how to make the function configuration "future proof". The transaction batching feature makes sense only in the MANAGED mode.

@lhotari I've added a section and updated the configuration classes/definitions + metrics for Transaction Batching. I would appreciate it if you could review it when you have a chance

pip/pip-439.md
```diff
@@ -0,0 +1,410 @@
...
+1. **Reduced Load on Transaction Coordinator**: Fewer transactions means less coordination overhead
+2. **Improved Throughput**: Higher message processing capacity with lower per-message overhead
+3. **Optimized Resource Usage**: Better utilization of transaction resources
+4. **Consistent Performance at Scale**: Maintains performance characteristics under high load
+
+### Transaction Batching Parameters
+
+Transaction batching is controlled by two main parameters:
+
+1. **`transactionBatchingMaxEntries`**: The maximum number of entries (incoming messages or batch messages) to process within a single
```
lhotari, line 229: It should be mentioned that transaction batching can be disabled by setting this value to `0`.

pip/pip-439.md
```diff
@@ -0,0 +1,410 @@
...
+transaction before committing it and starting a new one.
+ - An "entry" refers to an incoming batch message which could itself contain multiple individual messages
+ - Default: 1 (recommended minimum to ensure batch index acknowledgment state doesn't span transactions)
+ - Setting this to higher values increases throughput but may impact latency
+
+2. **`transactionBatchingQuietPeriodMs`**: The maximum amount of time to wait for additional messages before committing a transaction if
+`transactionBatchingMaxEntries` is not reached.
+ - Default: 100ms
```
lhotari:
```suggestion
 - Default: 1ms
```
It's better to have a shorter default. For Pulsar Client producer, the default `batchingMaxPublishDelay` value is 1ms so that's why I'd go with 1ms.

lhotari:
> @lhotari I've added a section and updated the configuration classes/definitions + metrics for Transaction Batching. I would appreciate it if you could review it when you have a chance

@wzhramc Looks good, just minor details about the default values.

wzhramc:
> > @lhotari I've added a section and updated the configuration classes/definitions + metrics for Transaction Batching. I would appreciate it if you could review it when you have a chance
> 
> @wzhramc Looks good, just minor details about the default values.

I fixed the minor details in the latest commit. I'll start the PIP Voting now

lhotari (approved):
LGTM. The vote has passed on the mailing list.

lhotari:
@wzhramc Do you have a chance to continue working on the implementation for PIP-439?

wzhramc:
Hi @lhotari, I truly apologize for the super long silence! The project that originally needed this feature was shut down, so I stopped working on it at the time. I’d like to pick up PIP-439 again and start implementing it soon.

Before I begin, I’ll review the changes to Pulsar Functions and transactions since the PIP was written. Please let me know if anyone else has started working on it or if there are recent changes I should account for.

lhotari:
> Hi @lhotari, I truly apologize for the super long silence! The project that originally needed this feature was shut down, so I stopped working on it at the time. I’d like to pick up PIP-439 again and start implementing it soon.
> 
> Before I begin, I’ll review the changes to Pulsar Functions and transactions since the PIP was written. Please let me know if anyone else has started working on it or if there are recent changes I should account for.

Thank you. I don't think that there has been major changes. Please go ahead asap since we are closing the window for Pulsar 5.0.0 very soon.

lhotari:
@wzhramc While implementing, it makes sense to update the PIP document if there's a reason. In most cases many details are only noticed after the implementation feedback. In this way the PIP document isn't fixed.

PR #24706, opened 2025-09-05
[fix][broker] Fix NPE and annotate nullable return values for ManagedCursorContainer

BewareMyPower:
### Motivation

I observed the following NPE when I ran some tests.

```
2025-09-05 10:49:38.826 [pulsar-msg-expiry-monitor-OrderedScheduler-0-0] WARN  org.apache.bookkeeper.common.util.SingleThreadSafeScheduledExecutorService - Unexpected throwable from task class com.google.common.util.concurrent.MoreExecutors$ScheduledListeningDecorator$NeverSuccessfulListenableFutureTask: Cannot invoke "org.apache.bookkeeper.mledger.impl.ManagedCursorContainer$CursorInfo.getCursor()" because the return value of "org.apache.bookkeeper.mledger.impl.ManagedCursorContainer.getCursorWithOldestPosition()" is null
java.lang.NullPointerException: Cannot invoke "org.apache.bookkeeper.mledger.impl.ManagedCursorContainer$CursorInfo.getCursor()" because the return value of "org.apache.bookkeeper.mledger.impl.ManagedCursorContainer.getCursorWithOldestPosition()" is null
	at org.apache.pulsar.broker.service.persistent.PersistentTopic.checkMessageExpiryWithSharedPosition(PersistentTopic.java:2169)
```

It's because `getCursorWithOldestPosition` could return null but `checkMessageExpiryWithSharedPosition` does not apply null validation.

### Modifications

Add `@Nullable` annotations to all `ManagedCursorContainer`'s methods of that could return null. Then check all non-test code and add null validation if missed:
1. The `checkMessageExpiryWithSharedPosition` added in https://github.com/apache/pulsar/pull/24622
2. The missed null validation for `getSlowedConsumer` since a very early commit, but not sure why the NPE was never reported

P.S. Using `Optional` might be a better alternative to force users to check null because many developers still ignore warnings. To avoid unnecessary heap allocation and API changes, I didn't change the method signature.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -2166,7 +2166,11 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
 
     private void checkMessageExpiryWithSharedPosition(ManagedLedgerImpl ml, int messageTtlInSeconds) {
         // Find the target position at one time, then expire all subscriptions and replicators.
-        ManagedCursor cursor = ml.getCursors().getCursorWithOldestPosition().getCursor();
+        final var cursorWithOldestPosition = ml.getCursors().getCursorWithOldestPosition();
+        if (cursorWithOldestPosition == null) {
+            return;
+        }
+        ManagedCursor cursor = cursorWithOldestPosition.getCursor();
         PersistentMessageFinder finder = new PersistentMessageFinder(topic, cursor, brokerService.getPulsar()
                 .getConfig().getManagedLedgerCursorResetLedgerCloseTimestampMaxClockSkewMillis());
         // Find the target position.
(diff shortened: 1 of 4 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -2166,7 +2166,11 @@ private void checkMessageExpiryWithoutSharedPosition(int messageTtlInSeconds) {
 
     private void checkMessageExpiryWithSharedPosition(ManagedLedgerImpl ml, int messageTtlInSeconds) {
         // Find the target position at one time, then expire all subscriptions and replicators.
-        ManagedCursor cursor = ml.getCursors().getCursorWithOldestPosition().getCursor();
+        final var cursorWithOldestPosition = ml.getCursors().getCursorWithOldestPosition();
+        if (cursorWithOldestPosition == null) {
+            return;
```
poorbarcode, line 2172: We can add a code comment here: "skip because no subscription registered"

BewareMyPower: Added

lhotari (approved):

PR #24708, opened 2025-09-05
[improve][build] docker: use up to date wolfi-base location

xnox:
Chainguard containers on dockerhub are no longer updated directly (as in mirrored slower), instead the primary publication location is hosted publicly (without quota limits or tokens!) on cgr.dev/chainguard/wolfi-base.

They are also continuously rebuilt, and thus are actually always up to date. This is why you likely did upgrade call, as the dockerhub image might be seen as stale from time to time.

Note that apk add always pulls latest, and the cgr.dev/chainguard/wolfi-base is always up to date. Note due to minimalism, there are no pre/post/maintainer scripts to support package upgrades universally. Thus in general we recommend to always at most apk add to a base image, without running upgrade. We recommend to upgrade the image digest (rebuild docker file with --no-cache).

- [x] `doc-not-needed`

### DIFF
--- a/docker/pulsar-all/Dockerfile.wolfi
+++ b/docker/pulsar-all/Dockerfile.wolfi
@@ -18,7 +18,7 @@
 # under the License.
 #
 
-# Dockerfile for pulsar-all image which uses Wolfi (chainguard/wolfi-base) as it's base image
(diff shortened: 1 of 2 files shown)


xnox:
- [X] `doc-not-needed`

lhotari:
Thanks for the contribution!

lhotari (approved):
LGTM

PR #3383, opened 2019-01-17
 refactor AuthorizationProvider interface to allow custom logic for determining super user

jerrypeng:
### Motivation

Currently the logic for determining if a role is a super user is not pluggable.  Thus, users have to specify who is a super using the the configuration of a broker which is hard coded.  

Making the logic of determining if a role is a super user pluggable while allow users more flexibility in setting super users which can be a dynamic list of users held in an external source.

### Modifications

Made the mechanism of determining super users pluggable

PR #24714, opened 2025-09-08
[fix][broker] replace thenComposeAsync with thenCompose method in AuthorizationService

oneby-wang:
### Motivation

I think we should not use any `CompletableFuture.xxxAsync()` method in Pulsar broker, unless we specify a thread pool using Netty `DefaultThreadlFactory`. I encountered `DataSketchesSummaryLogger` memory leak due to FastThreadLocal data not released after thread terminated when using mop.
https://github.com/streamnative/mop/blob/06b9763c1842915c3fd0133874088de96d821293/mqtt-broker/src/main/java/io/streamnative/pulsar/handlers/mqtt/broker/processor/MQTTBrokerProtocolMethodProcessor.java#L395-L402

`CompletableFuture.xxxAsync()` use new Thread or ForkJoinPool by default(base on CPU cores), Pulsar has many metrics loggers using FastThreadLocal, which need FastThreadLocalThread to release the memory.

### Modifications

Before: use `thenComposeAsync` method
After: use `thenCompose` method

### Verifying this change

This change is already covered by existing tests, see AuthorizationTest.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/2

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java
@@ -219,7 +219,7 @@ public class AuthorizationService {
         if (!this.conf.isAuthorizationEnabled()) {
             return CompletableFuture.completedFuture(true);
         }
-        return provider.isSuperUser(role, authenticationData, conf).thenComposeAsync(isSuperUser -> {
+        return provider.isSuperUser(role, authenticationData, conf).thenCompose(isSuperUser -> {
             if (isSuperUser) {
                 return CompletableFuture.completedFuture(true);
             } else {
@@ -244,7 +244,7 @@ public class AuthorizationService {
         if (!this.conf.isAuthorizationEnabled()) {
             return CompletableFuture.completedFuture(true);
         }
-        return provider.isSuperUser(role, authenticationData, conf).thenComposeAsync(isSuperUser -> {
+        return provider.isSuperUser(role, authenticationData, conf).thenCompose(isSuperUser -> {
             if (isSuperUser) {
                 return CompletableFuture.completedFuture(true);
             } else {
@@ -326,7 +326,7 @@ public class AuthorizationService {
         if (!this.conf.isAuthorizationEnabled()) {
             return CompletableFuture.completedFuture(true);
         }
-        return provider.isSuperUser(role, authenticationData, conf).thenComposeAsync(isSuperUser -> {
+        return provider.isSuperUser(role, authenticationData, conf).thenCompose(isSuperUser -> {
             if (isSuperUser) {
                 return CompletableFuture.completedFuture(true);
             } else {


oneby-wang:
There are still some `thenComposeAsync` calls in the code that do not specify an executor, and I think this will pose potential performance and memory leak risks. 
I'm not sure whether to replace these `thenComposeAsync` calls with `thenCompose` or specify an executor. However, if I need to specify an executor, I also don't know which one to choose.

lhotari:
> I think we should not use any `CompletableFuture.xxxAsync()` method in Pulsar broker, unless we specify a thread pool using Netty `DefaultThreadlFactory`. I encountered `DataSketchesSummaryLogger` memory leak due to FastThreadLocal data not released after thread terminated when using mop.

@oneby-wang What memory leak are you talking about? Is it a Netty ByteBuf leak?
Have you tried setting `io.netty.allocator.cacheTrimIntervalMillis` to some value?
([Netty 4.1.x source code](https://github.com/netty/netty/blob/b7d832d4599acd8f9aff3609f7d17af232395715/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L145-L146))

Threads in an executor will use the per-thread cache and release the cached memory after the JVM runs the finalizer ([ref 1](https://github.com/netty/netty/blob/b7d832d4599acd8f9aff3609f7d17af232395715/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L536-L538), [ref 2](https://github.com/netty/netty/blob/b7d832d4599acd8f9aff3609f7d17af232395715/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java#L103), [ref 3](https://github.com/netty/netty/blob/b7d832d4599acd8f9aff3609f7d17af232395715/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java#L486-L498)).

Are you talking about the PooledByteBufAllocator per-thread cache that is the "leaked memory"?

lhotari:
Similar changes in #24383. I don't think that we should blindly change from thenComposeAsync to thenCompose since some implementations might be blocking the callback completion thread. This change might break some implementations.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java
```diff
@@ -219,7 +219,7 @@ public CompletableFuture<Boolean> canProduceAsync(TopicName topicName, String ro
         if (!this.conf.isAuthorizationEnabled()) {
             return CompletableFuture.completedFuture(true);
         }
-        return provider.isSuperUser(role, authenticationData, conf).thenComposeAsync(isSuperUser -> {
+        return provider.isSuperUser(role, authenticationData, conf).thenCompose(isSuperUser -> {
```
lhotari, line 222: This is a breaking change to the interface. That's why we should change from `thenComposeAsync` to `thenComple`. Switching to `thenComposeAsync` with an executor is a possible acceptable solution so that we don't break the interface.

oneby-wang:
Hi, I'm talking about FastThreadLocal memory leak, `FastThreadLocal'remove()` method is called by `FastThreadLocalThread ` when thread is completed.
https://github.com/apache/pulsar/blob/56399ee448d7c3f3ba132206bb9425b8371f3301/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/DataSketchesSummaryLogger.java#L117-L120

If `DataSketchesSummaryLogger.ThreadLocalAccessor` is accessed by normal thread(not netty thread), map element will not be removed when thread is completed.
https://github.com/netty/netty/blob/255ad95343adcbc41b90dd3316560b6bd3af134c/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java#L302-L307

I encountered this problem in our production cluster using mop. After I replace `canConsumeAsync` method with `allowTopicOperationAsync` method in mop, the memory leak problem is solved.

And I haven't tried `io.netty.allocator.cacheTrimIntervalMillis`, I'll check the purpose of this parameter later.

oneby-wang:
> Similar changes in #24383. I don't think that we should blindly change from thenComposeAsync to thenCompose since some implementations might be blocking the callback completion thread. This change might break some implementations.

I see, I think at least we should specify an executor(created by netty DefaultThreadFactory) for thenComposeAsync method.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java
oneby-wang: I think in AuthorizationService, we can do this change, because all operations after thenComposeAsync are asynchronous (async) operations. I'm not sure if my understanding is correct.
https://github.com/apache/pulsar/blob/56399ee448d7c3f3ba132206bb9425b8371f3301/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java#L241-L254
and
https://github.com/apache/pulsar/blob/56399ee448d7c3f3ba132206bb9425b8371f3301/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java#L216-L229
and 
https://github.com/apache/pulsar/blob/56399ee448d7c3f3ba132206bb9425b8371f3301/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java#L324-L336

lhotari: There might be custom implementations that behave in a different way.

oneby-wang: I would assume that the implementer of the interface should implement the async semantics, rather than using the thenComposeAsync method to ensure that custom methods without async implementation can run asynchronously.
https://github.com/apache/pulsar/blob/c7dff63e8d38bfdc7ac881b5fec8c1b486346e6e/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationProvider.java#L112-L113

My opinion: it is the duty of implementer to implement the async semantics to make caller simpler, rather than leaving it to the caller to handle. At least currently, both implementations(PulsarAuthorizationProvider and MultiRolesTokenAuthorizationProvider) in Pulsar are non-blocking, the async operation is hold by MetaCache executor.
https://github.com/apache/pulsar/blob/c7dff63e8d38bfdc7ac881b5fec8c1b486346e6e/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/NamespaceResources.java#L136-L138

If we need to consider such an assumption, then some places in the project may need to be replaced with thenComposeAsync method.

BewareMyPower:
> If DataSketchesSummaryLogger.ThreadLocalAccessor is accessed by normal thread(not netty thread), map element will not be removed when thread is completed.

I've suffered the same memory leak before. The key point is that we relies too much on the so-called "fast" thread local class from Netty but the thread local object is used in non-Netty threads (not `FastThreadLocalThread`) in many places. Especially when things go to the unique `onRemoval` method from `FastThreadLocal`, there is an **implicit** requirement that this thread local object must be used in `FastThreadLocalThread`. It's very unfriendly to users.

I think to resolve this issue thoroughly, we should not rely on the `onRemoval` mechanism for Netty threads specially. Instead, we should call `map.remove` explicitly when detected the key is ready to remove.

@lhotari We'd better migrate the internal fix that from the private KSN project to Apache Pulsar.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authorization/AuthorizationService.java
BewareMyPower: > I would assume that the implementer of the interface should implement the async semantics,

It never makes sense to assume the implementation from 3rd party libraries has a certain behavior. There could be a chance that the downstream application has implemented a blocking method and it works well.

Now you changed the way of upstream side for how to call this API and you have to clarify that their implementation is now illegal though it has been working for long time.

That's what is called "breaking change".

oneby-wang: I got your point, but I see many direct calls to AuthorizationProvider in AuthorizationService, without wrapping a thenXxxAsync method. How do we guarantee non-blocking behaviour of these calls in that case?

Or we are just talking about "don't break change in existing non-blocking behaviours"?

BewareMyPower: Yes I'm talking about not breaking existing behavior.

> I see many direct calls to AuthorizationProvider in AuthorizationService

That makes sense. I think we should revisit the existing code usages for `xxxAsync`. But it should be a separated topic rather than a bug fix.

oneby-wang: > Yes I'm talking about not breaking existing behavior.

I see, you may close this pull request once you have made a decision.

oneby-wang:
> We'd better migrate the internal fix that from the private KSN project to Apache Pulsar.

May kick me when you are in progess? I'm interested in this memory leak problem.

nodece (changes requested):
This path correctly uses `thenCompose` instead of `thenComposeAsync`, which avoids unnecessary thread switching.

However, we still need to handle synchronized calls in Pulsar(This is the primary task). For example, `org.apache.pulsar.broker.authorization.AuthorizationService#canLookup`, which should be deprecated. We cannot use synchronous calls inside an async callback thread, as this will cause thread blocking. In the past, we often overlooked such issues.

oneby-wang:
> This path correctly uses `thenCompose` instead of `thenComposeAsync`, which avoids unnecessary thread switching.
> 
> However, we still need to handle synchronized calls in Pulsar(This is the primary task). For example, `org.apache.pulsar.broker.authorization.AuthorizationService#canLookup`, which should be deprecated

Is it ok if I replace canLookup with canLookupAsync? I searched canLookup, canConsume, canProduce, only find this call usage.
https://github.com/apache/pulsar/blob/90b2aac0fa61f90b3ddf3095da1652decf18602a/pulsar-websocket/src/main/java/org/apache/pulsar/websocket/admin/WebSocketWebResource.java#L152-L157

Or, like comments above, we should open a separated topic to discuss xxxAsync calls? Avoiding "breaking change".

oneby-wang:
Seems there are so many codes that convert async to sync operations using `get(int, TimeUnit)` in pulsar-websocket module.

nodece:
> Is it ok if I replace canLookup with canLookupAsync?

Yes, you can make this replacement. After that, you can try removing the `thenComposeAsync`, but you still need to handle the fact that `canLookup`, `canConsume`, and `canProduce` are synchronized methods, since they are public.

oneby-wang:
> but you still need to handle the fact that `canLookup`, `canConsume`, and `canProduce` are synchronized methods, since they are public.

Confused about handling synchronized methods `canLookup`, `canConsume`, and `canProduce`. Only one place uses `canLookup`, `canConsume`, and `canProduce` are not used in Pulsar(except in unit tests). How do I handle these three synchronized methods since I don't know where and how they will be called.

nodece:
> > but you still need to handle the fact that `canLookup`, `canConsume`, and `canProduce` are synchronized methods, since they are public.
> 
> Confused about handling synchronized methods `canLookup`, `canConsume`, and `canProduce`. Only one place uses `canLookup`, `canConsume`, and `canProduce` are not used in Pulsar(except in unit tests). How do I handle these three synchronized methods since I don't know where and how they will be called.

You can add the `@deprecated` to these methods, and improve the `get(int, TimeUnit)` in the synchronized method.

lhotari (commented):
Thanks for digging into this, @oneby-wang, and sorry for the very long silence. Your analysis was mostly right, and my one-line objection was vaguer than it deserved, so let me give you a proper answer.

**Where my objection was too vague.** I said "there might be custom implementations that behave in a different way." That is true in the abstract, but it overstates the protection `thenComposeAsync` actually provides here. In all three of these methods `provider.isSuperUser(role, authenticationData, conf)` is invoked *directly on the caller's thread* — there is no hop before it. `thenComposeAsync` only defers the *second* provider call. A provider that blocks is therefore already blocking a broker thread today, in these exact methods. The same shape a few methods further down (`allowFunctionOpsAsync`, `allowSourceOpsAsync`, `allowSinkOpsAsync`) already uses plain `thenCompose`, and `allowTopicOperationAsync`, `allowNamespaceOperationAsync` and most of the rest of the class call the provider with no hop at all. There is no class-wide "always hop before calling the provider" contract to break — these three `canXxxAsync` methods are the anomaly, and the `thenComposeAsync` in them arrived incidentally in #3383 (Jan 2019) when the super-user short-circuit was added. Before that commit, `provider.canProduceAsync(...)` was called directly on the caller's thread, which is exactly what your patch restores.

**Where you were clearly right.** Every in-tree `isSuperUser` returns an already-completed future: the default implementation in `AuthorizationProvider` and the only production override, in `MultiRolesTokenAuthorizationProvider`, return `CompletableFuture.completedFuture(...)` on every branch. `thenComposeAsync` on an already-completed future *always* dispatches to `ForkJoinPool.commonPool()`, whose `ForkJoinWorkerThread`s are not `FastThreadLocalThread`s. So today this really is an unconditional, unnecessary hop onto exactly the kind of thread that caused your leak. Your diagnosis holds up.

**Why I still don't want to merge it.** Two things have moved since September:

1. *The leak is fixed on master.* @BewareMyPower's #24719 landed the day after this discussion. `ThreadLocalAccessor` now keeps a weak reference to the owner thread and prunes entries whose owner is dead or garbage-collected, instead of depending on `FastThreadLocal#onRemoval`; `ThreadLocalAccessorTest` covers the plain-`Thread` case explicitly. The motivation in the PR description no longer applies to Pulsar itself.

2. *The in-tree benefit is zero.* `canProduceAsync` and `canConsumeAsync` have no production callers in this repository at all, and `canLookupAsync` is reached from exactly one place — `WebSocketWebResource#isAuthorized`, through the blocking `canLookup` wrapper. The broker's hot authorization path goes through `allowTopicOperationAsync`, which this PR does not touch.

So both the remaining benefit and the residual compatibility risk land entirely on out-of-tree code: protocol handlers such as MoP that call these methods, and custom `AuthorizationProvider` implementations. Neither is something our CI can observe, and nothing in the test suite pins the threading either way. Changing six-year-old published behaviour on a public API for a benefit no in-tree caller can see is not a trade I want to make, so I am going to close this without merging. That is a cost/benefit judgement, not a verdict on your analysis.

**What would be worth doing**, if you are still interested: the point @nodece raised is the real problem. `canProduce`, `canConsume` and `canLookup` are public, synchronous, and block on `get(timeout)`. Deprecating those three and converting the single real caller — `WebSocketWebResource#isAuthorized` — to the async path is a self-contained change with a benefit that is visible in-tree, and you have already established that it is the only call site. I would be glad to review that. The wider `get(int, TimeUnit)` usage you noticed across `pulsar-websocket` is worth its own issue.

Thanks again for the careful work, and for your patience with the delay.

lhotari:
Belated answer to this: the fix landed the following day as #24719. `ThreadLocalAccessor` now holds a weak reference to the owner thread and prunes entries whose owner has died or been garbage-collected, rather than depending on `FastThreadLocal#onRemoval`, so cleanup no longer requires the recording thread to be a `FastThreadLocalThread`. `ThreadLocalAccessorTest` covers the plain-`Thread` case. Sorry you did not get a ping at the time.

PR #24719, opened 2025-09-09
[fix][broker] Fix memory leak when metrics are updated in a thread other than FastThreadLocalThread

BewareMyPower:
### Motivation

See https://github.com/apache/pulsar/pull/24714#issuecomment-3266369903

Currently this issue rarely happens because most metrics updates happen in Netty I/O threads when handling requests. However, `DataSketchesOpStatsLogger` and `DataSketchesSummaryLogger` are public and can be used by downstream plugins (e.g. protocol handlers).

### Modifications

- Refactor the code by adding a `record` method to `LocalData` and `ThreadLocalAccessor` responsible for updating the metrics.
- Then, add `ThreadLocalAccessorTest` to verify when the metrics are updated in a normal thread (not `FastThreadLocalRunnable`), the `LocalData` objects won't be removed from the map in `ThreadLocalAccessor`
- Fix this issue by introducing a weak reference that holds the current thread when creating a `LocalData` in a non-Netty thread and remove it from the map if it's garbage collected or not alive when iterating the whole map called by `rotateLatencyCollection`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/DataSketchesOpStatsLogger.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/DataSketchesOpStatsLogger.java
@@ -165,8 +165,7 @@ public class DataSketchesOpStatsLogger implements OpStatsLogger {
         return s != null ? s.getQuantile(quantile) : Double.NaN;
     }
 
-    @VisibleForTesting
-    static class LocalData {
+    private static class LocalData {
         private final DoublesSketch successSketch = new DoublesSketchBuilder().build();
         private final DoublesSketch failSketch = new DoublesSketchBuilder().build();
         private final StampedLock lock = new StampedLock();
@@ -184,7 +183,7 @@ public class DataSketchesOpStatsLogger implements OpStatsLogger {
             }
         }
 
-        boolean shouldRemove() {
+        private boolean shouldRemove() {
             if (ownerThreadReference == null) {
                 // the owner is a FastThreadLocalThread which handles the removal using FastThreadLocal#onRemoval
                 return false;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/DataSketchesOpStatsLogger.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/DataSketchesOpStatsLogger.java
@@ -18,19 +18,11 @@
  */
 package org.apache.pulsar.broker.stats.prometheus.metrics;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.yahoo.sketches.quantiles.DoublesSketch;
-import com.yahoo.sketches.quantiles.DoublesSketchBuilder;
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM

BewareMyPower:
cc @oneby-wang 

BTW, the archived open source KoP fixes this issue by remembering the Netty I/O executor everywhere and updating the metrics in that executor. See https://github.com/streamnative/kop/pull/2006/files

However, this solution is error-prone because it's easy to miss such pattern.

pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/DataSketchesOpStatsLogger.java
lhotari, line 1: There's also org.apache.pulsar.broker.stats.prometheus.metrics.DataSketchesSummaryLogger class with a similar pattern. I guess that should also be addressed?

lhotari: Perhaps the duplication could be reduced by refactoring?

BewareMyPower: Addressed, PTAL again

pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/prometheus/metrics/ThreadLocalAccessor.java
```diff
@@ -0,0 +1,113 @@
...
+    static class LocalData {
+
+        final DoublesSketch successSketch = new DoublesSketchBuilder().build();
+        final DoublesSketch failSketch = new DoublesSketchBuilder().build();
+        final StampedLock lock = new StampedLock();
+        // Keep a weak reference to the owner thread so that we can remove the LocalData when the thread
+        // is not alive anymore or has been garbage collected.
+        // This reference isn't needed when the owner thread is a FastThreadLocalThread and will be null in that case.
+        // The removal is handled by FastThreadLocal#onRemoval when the owner thread is a FastThreadLocalThread.
+        final WeakReference<Thread> ownerThreadReference;
```
lhotari: It could be useful to continue the refactoring and make these fields private.

there should be public methods `updateSuccess` and `updateFail` to get rid of duplicate code that is using these field from outside:

```
        long stamp = localData.lock.readLock();
        try {
            localData.successSketch.update(valueMillis);
        } finally {
            localData.lock.unlockRead(stamp);
        }
```


```
        long stamp = localData.lock.readLock();
        try {
            localData.failSketch.update(valueMillis);
        } finally {
            localData.lock.unlockRead(stamp);
        }
```

BewareMyPower: Good suggestion, addressed

lhotari (approved):

PR #24722, opened 2025-09-10
[fix][ml] Negative backlog & acked positions does not exist & message lost when concurrently occupying topic owner

poorbarcode:
### Motivation

Our pulsar cluster encountered an issue where the response of the command `pulsar-admin topics stats --get-precise-backlog --get-subscription-backlog-size  <topic>` has a negative backlog.

And topic internal stats show that a consumer acknowledged messages that do not exist in `topic.ledgers`
- ledgerInfo:  `{"ledgerId":1140377,"entries":5,"size":45573,"offloaded":false,"metadata":null,"underReplicated":false}`
- But the acknowledged more entries: `(1140377:-1..1140377:6]`

---

The issue happened this way:

- Configurations
  - `managedLedgerDefaultEnsembleSize=2`
  - `managedLedgerDefaultWriteQuorum=2`
  - `managedLedgerDefaultAckQuorum=2`
- Topic state
  - `ledgers`: `[3: {entries: 1}]`
- Scenarios
  -  `bookie-2` is slow
  - ZK connections are unstable, which may cause a race condition on owning topics

| time | `broker-1` | `broker-2` | `bookie-1` | `bookie-2` |
| --- | --- | --- | --- | --- |
| 1 | owned the topic |
| 2 | start publish message `3:1`(has not completed yet) |
| 3 | | | receives the write request `3:1` | receives the write request `3:1` |
| 4 | | | write disk finished | write disk finished |
| 5 | | | responds to `broker-1` | 
| 2 | start publish message `3:2`(has not sent to Bookies yet) |
| 6 | metadata store connection unstable |
| 7 | | onwed the topic |
| 8 | assumes itself is still the owner of the topic |
| 9 | | start to close the ledger `3` |
| 10 | | | receives the close request of the ledger `3` | receives the close request of the ledger `3` |
| 11 | | | fence & close the ledger `3` |
| 12 | | | receives the write request `3:2` | receives the write request `3:2` |
| 13 | | | responds `fenced error` since the ledger has been closed |
| 14 | received a `fenced error`|
| 15 | switch ledger to `4` and write topic metadata with `[3: {entries: 1}]` since `bookie-2` has not responded to the request of writing `3:1` |
| 16 | | | | responds to `broker-1` for the writing `3:1` |
| 17 | | | | fence & close the ledger `3` |
| 18 | | closed ledger `3`, and it has `2` entries |
| 19 | received the response of writing `3:1` |

Highlight: The issue also leads to messages being lost.

You can reproduce the issue with the new test `testConcurrentlyCloseCurrentLedger`

### Modifications

Re-check the ledger entries from the metadata store if received a `fenced error` when writing Bookies

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -1878,7 +1878,7 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
         }
     }
 
-    void ledgerFailedWriteBecauseClosedByOthers(final LedgerHandle currentLedger) {
+    void ledgerAddFailedDueToConcurrentlyModified(final LedgerHandle currentLedger) {
         bookKeeper.asyncOpenLedger(currentLedger.getId(), digestType, config.getPassword(), (rc, lh, ctx) -> {
             if (rc == Code.OK) {
                 log.warn("[{}] Successfully opened ledger {} to check the last add confirmed position when the ledger"
@@ -1909,7 +1909,7 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
     // //////////////////////////////////////////////////////////////////////
     // Private helpers
 
-    synchronized void ledgerClosed(final LedgerHandle lh, @Nullable Long forcedLastAddConfirmed) {
+    synchronized void ledgerClosed(final LedgerHandle lh, @Nullable Long lastAddConfirmed) {
         final State state = STATE_UPDATER.get(this);
         LedgerHandle currentLedger = this.currentLedger;
         if (currentLedger == lh && (state == State.ClosingLedger || state == State.LedgerOpened)) {
@@ -1924,7 +1924,7 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
             return;
         }
 
-        long entriesInLedger = forcedLastAddConfirmed != null ? forcedLastAddConfirmed.longValue() + 1
+        long entriesInLedger = lastAddConfirmed != null ? lastAddConfirmed.longValue() + 1
                 : lh.getLastAddConfirmed() + 1;
         if (log.isDebugEnabled()) {
             log.debug("[{}] Ledger has been closed id={} entries={}", name, lh.getId(), entriesInLedger);
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -1878,38 +1878,21 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
         }
     }
 
-    void ledgerAddFailedDueToConcurrentlyModified(final LedgerHandle currentLedger) {
-        bookKeeper.asyncOpenLedger(currentLedger.getId(), digestType, config.getPassword(), (rc, lh, ctx) -> {
-            if (rc == Code.OK) {
-                log.warn("[{}] Successfully opened ledger {} to check the last add confirmed position when the ledger"
-                    + " was concurrent modified(it is an unexpected behaviour, which happens when the load-balancer"
-                    + " does not work as expected). The add confirmed position in memory is {}, and the value"
-                    + " stored in metadata store is {}. When you get this log, the latest several entries may be"
-                    + " repeated.", name, lh.getId(), currentLedger.getLastAddConfirmed(), lh.getLastAddConfirmed());
-                ledgerClosed(currentLedger, lh.getLastAddConfirmed());
-            } else {
-                log.error("[{}] Going to fence the topic because failed opened ledger {} to check the last add"
-                        + " confirmed position when the ledger was concurrent modified(it is an unexpected behaviour,"
-                        + " which happens when the load-balancer does not work as expected). The add confirmed position"
-                        + " in memory is {}, and the error code {}. Fecing the topic to avoid messages lost.",
-                        name, lh.getId(), currentLedger.getLastAddConfirmed(), rc);
-                // Stop switching ledger and write topic metadata, to avoid messages lost. The doc of
-                // LedgerHandle also mentioned this: https://github.com/apache/bookkeeper/blob/release-4.17.2/
-                // bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L2047-L2048
-                handleBadVersion(new BadVersionException("Failed opened ledger {} to check the last add confirmed"
-                        + " position when the ledger was concurrent modified, error code: " + rc));
-            }
-        }, null);
-    }
-
-    synchronized void ledgerClosed(final LedgerHandle lh) {
-        ledgerClosed(lh, null);
+    void addEntryFailedDueToConcurrentlyModified(final LedgerHandle currentLedger, int rc) {
+        log.error("[{}] Fencing the topic because the current ledger was concurrent modified but a other"
+                + " bookie client, which is not expcted. Current ledger: {}, lastAddConfirmed: {}, error coder: {}.",
+                name, currentLedger.getId(), currentLedger.getLastAddConfirmed(), rc);
+        // Stop switching ledger and write topic metadata, to avoid messages lost. The doc of
+        // LedgerHandle also mentioned this: https://github.com/apache/bookkeeper/blob/release-4.17.2/
+        // bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L2047-L2048.
+        handleBadVersion(new BadVersionException("Failed opened ledger {} to check the last add confirmed"
+                + " position when the ledger was concurrent modified."));
     }
 
     // //////////////////////////////////////////////////////////////////////
     // Private helpers
 
-    synchronized void ledgerClosed(final LedgerHandle lh, @Nullable Long lastAddConfirmed) {
+    synchronized void ledgerClosed(final LedgerHandle lh) {
         final State state = STATE_UPDATER.get(this);
         LedgerHandle currentLedger = this.currentLedger;
         if (currentLedger == lh && (state == State.ClosingLedger || state == State.LedgerOpened)) {
@@ -1924,8 +1907,7 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
             return;
         }
 
-        long entriesInLedger = lastAddConfirmed != null ? lastAddConfirmed.longValue() + 1
-                : lh.getLastAddConfirmed() + 1;
+        long entriesInLedger = lh.getLastAddConfirmed() + 1;
         if (log.isDebugEnabled()) {
             log.debug("[{}] Ledger has been closed id={} entries={}", name, lh.getId(), entriesInLedger);
         }
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -136,6 +136,7 @@ import org.apache.bookkeeper.mledger.util.CallbackMutex;
 import org.apache.bookkeeper.mledger.util.Futures;
 import org.apache.bookkeeper.mledger.util.ManagedLedgerImplUtils;
 import org.apache.bookkeeper.net.BookieId;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.pulsar.common.api.proto.CommandSubscribe.InitialPosition;
 import org.apache.pulsar.common.policies.data.EnsemblePlacementPolicyConfig;
@@ -261,6 +262,10 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
     protected static final int DEFAULT_LEDGER_DELETE_RETRIES = 3;
     protected static final int DEFAULT_LEDGER_DELETE_BACKOFF_TIME_SEC = 60;
     private static final String MIGRATION_STATE_PROPERTY = "migrated";
+    // If "currentLedger" was closed by the component "auto-replication", Managed ledger will determine the LAC from
+    // ledger metadata stored. This variable used to record the result of the latest once re-checking.
+    private Pair<Long, CompletableFuture<Integer>> ledgerRecheckInProgress = new ImmutablePair<>(-1L,
+            CompletableFuture.completedFuture(Code.OK));
 
     public enum State {
         None, // Uninitialized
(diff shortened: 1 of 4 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
+
+    synchronized void ledgerClosed(final LedgerHandle lh) {
+        ledgerClosed(lh, null);
+    }
+
     // //////////////////////////////////////////////////////////////////////
     // Private helpers
 
-    synchronized void ledgerClosed(final LedgerHandle lh) {
+    synchronized void ledgerClosed(final LedgerHandle lh, @Nullable Long forcedLastAddConfirmed) {
```
Technoboy-: `forcedLastAddConfirmed` ->`lastAddConfirmed`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
         }
     }
 
+    void ledgerFailedWriteBecauseClosedByOthers(final LedgerHandle currentLedger) {
```
Technoboy-: `ledgerFailedWriteBecauseClosedByOthers` -> `ledgerAddFailed`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Renamed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Renamed to `ledgerAddFailedDueToConcurrentlyModified`

lhotari:
@poorbarcode Regarding the title "Negative backlog & acked positions does not exist & message lost when concurrently occupying topic owner" is describing the current symptoms and impact. Perhaps it should be emphasized in the title that this PR is addressing fundamentals of how Pulsar Brokers should be handling topic ownership? It's surprising that this hasn't already been implemented properly by now. Thanks for addressing this issue!

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
         }
     }
 
+    void ledgerAddFailedDueToConcurrentlyModified(final LedgerHandle currentLedger) {
+        bookKeeper.asyncOpenLedger(currentLedger.getId(), digestType, config.getPassword(), (rc, lh, ctx) -> {
+            if (rc == Code.OK) {
+                log.warn("[{}] Successfully opened ledger {} to check the last add confirmed position when the ledger"
+                    + " was concurrent modified(it is an unexpected behaviour, which happens when the load-balancer"
+                    + " does not work as expected). The add confirmed position in memory is {}, and the value"
```
lhotari: Is it necessary to mention "it is an unexpected behavior, which happens when the load-balancer does not work as expected" ? What is that when "the load-balancer does not work as expected"?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
+                    + " stored in metadata store is {}. When you get this log, the latest several entries may be"
+                    + " repeated.", name, lh.getId(), currentLedger.getLastAddConfirmed(), lh.getLastAddConfirmed());
```
lhotari: What does this mean: "When you get this log, the latest several entries may be repeated." ?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
+                ledgerClosed(currentLedger, lh.getLastAddConfirmed());
+            } else {
+                log.error("[{}] Going to fence the topic because failed opened ledger {} to check the last add"
+                        + " confirmed position when the ledger was concurrent modified(it is an unexpected behaviour,"
+                        + " which happens when the load-balancer does not work as expected). The add confirmed position"
```
lhotari: again, is it necessary mention the load-balancer

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
+                        + " in memory is {}, and the error code {}. Fecing the topic to avoid messages lost.",
```
lhotari: Fecing -> Fencing
It could be more general. "Fencing the topic to ensure durability and consistency."

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
 
+    void ledgerAddFailedDueToConcurrentlyModified(final LedgerHandle currentLedger) {
+        bookKeeper.asyncOpenLedger(currentLedger.getId(), digestType, config.getPassword(), (rc, lh, ctx) -> {
+            if (rc == Code.OK) {
+                log.warn("[{}] Successfully opened ledger {} to check the last add confirmed position when the ledger"
+                    + " was concurrent modified(it is an unexpected behaviour, which happens when the load-balancer"
+                    + " does not work as expected). The add confirmed position in memory is {}, and the value"
+                    + " stored in metadata store is {}. When you get this log, the latest several entries may be"
+                    + " repeated.", name, lh.getId(), currentLedger.getLastAddConfirmed(), lh.getLastAddConfirmed());
+                ledgerClosed(currentLedger, lh.getLastAddConfirmed());
```
lhotari, line 1900: `ledgerClosed` doesn't seem to close the new LedgerHandle that is passed to this call back in the `lh` parameter. Is it the correct method to call?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
+            } else {
+                log.error("[{}] Going to fence the topic because failed opened ledger {} to check the last add"
+                        + " confirmed position when the ledger was concurrent modified(it is an unexpected behaviour,"
+                        + " which happens when the load-balancer does not work as expected). The add confirmed position"
+                        + " in memory is {}, and the error code {}. Fecing the topic to avoid messages lost.",
+                        name, lh.getId(), currentLedger.getLastAddConfirmed(), rc);
+                // Stop switching ledger and write topic metadata, to avoid messages lost. The doc of
+                // LedgerHandle also mentioned this: https://github.com/apache/bookkeeper/blob/release-4.17.2/
+                // bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L2047-L2048
```
lhotari: Just wondering if the information in the LedgerHandle `if (getLedgerMetadata().getState() == LedgerMetadata.State.IN_RECOVERY) {` is consistent:
https://github.com/apache/bookkeeper/blob/75f65df7b487a9ee29817dc849a40f627c9857b7/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L2046-L2048

Due to similar reasons as https://github.com/apache/pulsar/pull/24665, is it really reliable?

Uses cache:
https://github.com/apache/pulsar/blob/a53279837fae8b1c8bd16bec966cbc39823ed210/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/bookkeeper/PulsarLedgerManager.java#L198-L222

In BK, there's no ZK "sync", but it goes directly to ZK each time:
https://github.com/apache/bookkeeper/blob/2eb70b1f8216b2c6621d0e57cd491a2067824316/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L469-L471

Not necessarily relevant here, but it's good to be aware of the PulsarLedgerManager impact with Pulsar.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,10 +1878,38 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
         }
     }
 
+    void ledgerAddFailedDueToConcurrentlyModified(final LedgerHandle currentLedger) {
```
lhotari: Does this get called multiple times for each failure? would that result in attempts to open the ledger multiple times in a row?

poorbarcode: I have changed the behaviour: the topic should be fenced if the ledger was modified by the other.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Thanks for the suggestion. I have improved the log.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Removed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Removed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Removed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: It is correct. BK client guarantees the ledgerHandle will be closed eventually.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: > Due to similar reasons as https://github.com/apache/pulsar/pull/24665, is it really reliable?

They are not related.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
lhotari: Does it cause a memory leak? Could you add a comment about it that it's intentionally left open. When does the BK client close it "eventually"?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,6 +1878,18 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
         }
     }
 
+    void addEntryFailedDueToConcurrentlyModified(final LedgerHandle currentLedger, int rc) {
+        log.error("[{}] Fencing the topic to ensure durability and consistency(the current ledger was concurrent"
+                + " modified by a other bookie client, which is not expcted)."
```
lhotari: nit
```suggestion
        log.error("[{}] Fencing the topic to ensure durability and consistency (the current ledger was concurrent"
                + " modified by a other bookie client, which is not expected)."
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1878,6 +1878,18 @@ protected synchronized void updateLedgersIdsComplete(@Nullable LedgerHandle orig
...
+                + " Current ledger: {}, lastAddConfirmed: {} (the value stored may be larger), error coder: {}.",
+                name, currentLedger.getId(), currentLedger.getLastAddConfirmed(), rc);
+        // Stop switching ledger and write topic metadata, to avoid messages lost. The doc of
+        // LedgerHandle also mentioned this: https://github.com/apache/bookkeeper/blob/release-4.17.2/
+        // bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L2047-L2048.
+        handleBadVersion(new BadVersionException("Failed opened ledger {} to check the last add confirmed"
+                + " position when the ledger was concurrent modified."));
```
lhotari: The exception message isn't very clear. Could it be improve? It also contains `{}` without a meaning.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
```diff
@@ -361,7 +361,16 @@ void handleAddFailure(final LedgerHandle lh) {
...
+            // If we received a "MetadataVersionException" or a "LedgerFencedException", we should tell the ML that
+            // the ledger has been closed by others, and the entries count in the ledger may is not correct. The ML
+            // will handle it.
             finalMl.ledgerClosed(lh);
+            if (rc != null && (rc.intValue() == BKException.Code.MetadataVersionException
+                    || rc.intValue() == BKException.Code.LedgerFencedException)) {
+                finalMl.addEntryFailedDueToConcurrentlyModified(lh, rc);
+            } else {
+                finalMl.ledgerClosed(lh);
+            }
```
lhotari: There's now 2 calls to `finalMl.ledgerClosed(lh);` it seems that the intention has been to remove the one on line 367.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Fixed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: Fixed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java
poorbarcode: Sorry, I added the line-367 for test and forgot to remove it

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
poorbarcode: I have changed the behaviour: the topic should be fenced if the ledger was modified by someone else.

After reviewing the code, I got another case: auto-replication may close an active ledger if it is not in write-in-flight state. So Pulsar need to work continuously if the LAC re-checking is successful.

> Does this get called multiple times for each failure? would that result in attempts to open the ledger multiple times in a row?

I have added a mechanism to prevent it. And I have added more tests

Issue #19705, opened 2023-03-03
PIP-254: Support configuring client version with a description suffix


BewareMyPower:
### Motivation

In the `CommandConnect` of the Pulsar API protocol, there is a `client_version` field, which is configured by the client with a string that describes the client SDK. This description can be queried via the `pulsar-admin topics stats` command.

https://github.com/apache/pulsar/blob/0e6ef109322058e7834ca23d968f66f5632cf2b1/pulsar-common/src/main/proto/PulsarApi.proto#L269

However, this field is hard-coded (though generated according to the version) in each client AFAIK. It could bring an issue that if client SDK B depends on client SDK A directly, the `client_version` field is still the description of SDK A.

Currently the issue exists in the Python and the Node.js clients, which are both based on the C++ client. It means that when an application written in the Python or Node.js Pulsar SDK connects to Pulsar, the stats show that they are C++ client.

Further more, the Java client can also be called by other languages directly, including JVM languages like Scala, and script languages like JRuby. Here is an example that a Scala Pulsar client uses the official Java client as the dependency:

https://github.com/CleverCloud/pulsar4s/blob/e16d5cc778e4038efed030b6d74ba2b1b06172d4/build.sbt#L187

However, the `PulsarVersion` class is generated automatically and the version is fixed. It would be meaningful to differ the wrapper and the base library.

### Goal

Make the client version string configurable for each `PulsarClientImpl` instance. Then the library that depends on the `pulsar-client` dependency can wrap the `PulsarClientImpl` with its own client version.

### API Changes

Add the following method to `ClientBuilderImpl`:

```java
    /**
     * Set the description.
     *
     * <p> By default, when the client connects to the broker, a version string like "Pulsar-Java-v<x.y.z>" will be
     * carried and saved by the broker. The client version string could be queried from the topic stats.
     *
     * <p> This method provides a way to add more description to a specific PulsarClient instance. If it's configured,
     * the description will be appended to the original client version string, with '-' as the separator.
     *
     * <p>For example, if the client version is 3.0.0, and the description is "forked", the final client version string
     * will be "Pulsar-Java-v3.0.0-forked".
     *
     * @param description the description of the current PulsarClient instance
     * @throws IllegalArgumentException if the length of description exceeds 64
     */
    public ClientBuilder description(String description) { /* ... */ }
```

Then the library provider should cast the `ClientBuilder` to `ClientBuilderImpl` before setting other fields:

```java
[...]

PR #24729, opened 2025-09-11
[improve][client] Allow adding custom description to User-Agent header

nodece:
### Motivation

[PIP-254](https://github.com/apache/pulsar/issues/19705) only improved the pulsar-client. The pulsar-admin was not covered. This PR extends that work to also support the pulsar-admin. While not explicitly included in [PIP-254](https://github.com/apache/pulsar/issues/19705), this can be considered a natural part of it.

### Modifications

- Update the user agent construction to append the client description if provided.
- Add `description(String)` to the `PulsarAdminBuilder` and `PulsarAdminBuilderImpl`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/PulsarAdminBuilderImpl.java
+++ b/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/PulsarAdminBuilderImpl.java
@@ -283,4 +283,10 @@ public class PulsarAdminBuilderImpl implements PulsarAdminBuilder {
         this.conf.setConnectionMaxIdleSeconds(connectionMaxIdleSeconds);
         return this;
     }
+
+    @Override
+    public PulsarAdminBuilder description(String description) {
+        this.conf.setDescription(description);
+        return this;
+    }
 }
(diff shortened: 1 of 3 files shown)


pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/PulsarAdminBuilderImpl.java
```diff
@@ -283,4 +283,10 @@ public PulsarAdminBuilder connectionMaxIdleSeconds(int connectionMaxIdleSeconds)
         this.conf.setConnectionMaxIdleSeconds(connectionMaxIdleSeconds);
         return this;
     }
+
+    @Override
+    public PulsarAdminBuilder description(String description) {
+        this.conf.setDescription(description);
+        return this;
+    }
```
lhotari, line 291: https://github.com/apache/pulsar/blob/e916457f9055cf3e292d98ffcb677518a27699f6/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java#L430-L434

Where is the check for the max 64 character limit? Please add a test that it's not allowed.

nodece: I will make a PR to add this test.

BewareMyPower: ~~@nodece It seems that you merged without checking this comment~~

I didn't see your comment when I replied

nodece: @BewareMyPower Good catch. I should add the character limit in the builder.

nodece: @lhotari @BewareMyPower I submitted https://github.com/apache/pulsar/pull/24734 to fix this.

PR #24730, opened 2025-09-11
[fix][broker] Ensure KeyShared sticky mode consumer respects assigned ranges

shibd:
### Motivation

This PR fixes a critical bug where the `KeyShared` sticky mode consumer in Pulsar would incorrectly consume messages from hash ranges not explicitly assigned to it. 

Specifically, if a single `KeyShared` consumer is configured with non-contiguous sticky ranges, e.g.:
- `[0, 9999]`
- `[20000, 29999]`
- `[40000, 49999]`

The consumer would incorrectly receive messages for keys falling into the gaps (e.g., `15000` which is between `9999` and `20000`). 

The root cause lies in the `HashRangeExclusiveStickyKeyConsumerSelector`'s `select` method, which previously stored only the start/end points of ranges in its `rangeMap`. This led to an erroneous interpretation where any hash between a range's `end` and the next range's `start` would be assigned to the previous range's consumer.

https://github.com/apache/pulsar/blob/84b834f95c83e2385d8ca9bccb4cb78120ed582c/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java#L108-L116

In the scenario above, this consumer would actually receive messages in the range of **0 ~ 49999** (when there is only one consumer).

You can reproduce by testing from this PR:
- The `testConsumerSelectWithMultipRanges` unit test in this PR.
- The `testCustomStickyRange` integration test in this PR.

BTW: v3.0 has the same issue.

### Modifications

- Core Data Structure Refactor: The rangeMap now stores the complete Range object instead of just the range's start key. This provides a foundation for precise containment checks and conflict detection.
- Conflict Detection: The algorithm for detecting range conflicts between different consumers has been optimized for better accuracy and efficiency.

### Verifying this change

This change added new unit tests and an integration test and can be verified as follows:

-   **`HashRangeExclusiveStickyKeyConsumerSelectorTest` (Unit Tests)**:
    * `testConsumerSelect`: Updated to verify `select(hash)` returns the correct consumer only when `hash` is within an explicitly defined range, and `null` for hashes in gaps.
    * `testConsumerSelectWithMultipRanges`: Added to confirm a single consumer with multiple distinct sticky ranges correctly selects messages only within those ranges and returns `null` for hashes in gaps.
    * `testOneConsumerRangeConflict`: Added to verify that a consumer cannot be added if its own `KeySharedMeta` contains internally conflicting or invalid (`start > end`) ranges.
    * `testSingleRangeConflict` and `testMultipleRangeConflict`: Updated to correctly assert expected conflicts and non-conflicts based on the new strict range overlap detection logic.
-   **`KeySharedSubscriptionTest.testCustomStickyRange` (Integration Test)**:
    * A new end-to-end integration test has been added. It simulates the reported scenario with a partitioned topic and two consumers assigned non-overlapping sticky ranges. This test verifies that each consumer *only* receives messages for keys falling within its *explicitly assigned ranges*, confirming the fix for the unintended auto-split behavior.

### Documentation

- [ ] `doc`
- [ ] `doc-required` 
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR in forked repository: ```

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
@@ -137,7 +137,7 @@ public class HashRangeExclusiveStickyKeyConsumerSelector implements StickyKeyCon
             // 2. Validate: check for overlaps with the next range in the sorted list
             if (i < sortedRanges.size() - 1) {
                 IntRange nextRange = sortedRanges.get(i + 1);
-                if (checkRangesOverlap(currentRange, nextRange)) {
+                if (areRangesOverlapping(currentRange, nextRange)) {
                     return FutureUtil.failedFuture(
                             new BrokerServiceException.ConsumerAssignException("Consumer's own ranges conflict: "
                                     + "[" + currentRange.getStart() + "," + currentRange.getEnd() + "] "
@@ -159,7 +159,7 @@ public class HashRangeExclusiveStickyKeyConsumerSelector implements StickyKeyCon
             Map.Entry<Integer, Pair<Range, Consumer>> conflictBeforeStart = rangeMap.floorEntry(newRange.getStart());
             if (conflictBeforeStart != null) {
                 Range existingRange = conflictBeforeStart.getValue().getLeft();
-                if (checkRangesOverlap(newRange, existingRange)) {
+                if (areRangesOverlapping(newRange, existingRange)) {
                     return conflictBeforeStart.getValue().getRight();
                 }
             }
@@ -167,7 +167,7 @@ public class HashRangeExclusiveStickyKeyConsumerSelector implements StickyKeyCon
             Map.Entry<Integer, Pair<Range, Consumer>> conflictAfterStart = rangeMap.ceilingEntry(newRange.getStart());
             if (conflictAfterStart != null) {
                 Range existingRange = conflictAfterStart.getValue().getLeft();
-                if (checkRangesOverlap(newRange, existingRange)) {
+                if (areRangesOverlapping(newRange, existingRange)) {
                     return conflictAfterStart.getValue().getRight();
                 }
             }
@@ -176,11 +176,11 @@ public class HashRangeExclusiveStickyKeyConsumerSelector implements StickyKeyCon
     }
 
 
-    private static boolean checkRangesOverlap(IntRange range1, Range range2) {
+    private static boolean areRangesOverlapping(IntRange range1, Range range2) {
         return Math.max(range1.getStart(), range2.getStart()) <= Math.min(range1.getEnd(), range2.getEnd());
     }
 
-    private static boolean checkRangesOverlap(IntRange range1, IntRange range2) {
+    private static boolean areRangesOverlapping(IntRange range1, IntRange range2) {
         return Math.max(range1.getStart(), range2.getStart()) <= Math.min(range1.getEnd(), range2.getEnd());
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
@@ -156,36 +156,19 @@ public class HashRangeExclusiveStickyKeyConsumerSelector implements StickyKeyCon
     private synchronized Consumer findConflictingConsumer(List<IntRange> newConsumerRanges) {
         for (IntRange newRange : newConsumerRanges) {
             // 1. Check for potential conflicts with existing ranges that start before newRange's start.
-            //    Example: newRange = [50, 100], existingRange = [0, 70]
-            //    floorEntry finds an entry whose key is less than or equal to newRange.getStart().
             Map.Entry<Integer, Pair<Range, Consumer>> conflictBeforeStart = rangeMap.floorEntry(newRange.getStart());
             if (conflictBeforeStart != null) {
                 Range existingRange = conflictBeforeStart.getValue().getLeft();
                 if (checkRangesOverlap(newRange, existingRange)) {
-                    // Conflict found, return the consumer associated with the conflicting range
                     return conflictBeforeStart.getValue().getRight();
                 }
             }
-
-            // 2. Iterate through existing ranges that start at or after newRange's start.
-            //    ConcurrentSkipListMap.tailMap(fromKey, inclusive) returns a sub-map
-            //    containing entries whose keys are greater than or equal to fromKey.
-            //    This allows efficient iteration over all existing ranges that start at or after newRange.getStart().
-            for (Map.Entry<Integer, Pair<Range, Consumer>> entry :
-                    rangeMap.tailMap(newRange.getStart(), true).entrySet()) {
-                Range existingRange = entry.getValue().getLeft();
-                Consumer existingConsumer = entry.getValue().getRight();
-
-                // If the start of the current existingRange is already beyond the end of newRange,
-                // no further existing ranges (which are ordered by start key) can overlap, so we can break.
-                if (existingRange.getStart() > newRange.getEnd()) {
-                    break;
-                }
-
-                // Check if the current existingRange overlaps with newRange
+            // 2. Check for potential conflicts with existing ranges that start after newRange's start.
+            Map.Entry<Integer, Pair<Range, Consumer>> conflictAfterStart = rangeMap.ceilingEntry(newRange.getStart());
+            if (conflictAfterStart != null) {
+                Range existingRange = conflictAfterStart.getValue().getLeft();
                 if (checkRangesOverlap(newRange, existingRange)) {
-                    // Conflict found, return the consumer associated with the conflicting range
-                    return existingConsumer;
+                    return conflictAfterStart.getValue().getRight();
                 }
             }
         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
@@ -20,14 +20,15 @@ package org.apache.pulsar.broker.service;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListMap;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.pulsar.client.api.Range;
 import org.apache.pulsar.common.api.proto.IntRange;
-import org.apache.pulsar.common.api.proto.KeySharedMeta;
 import org.apache.pulsar.common.util.FutureUtil;
 
 /**
@@ -38,7 +39,7 @@ import org.apache.pulsar.common.util.FutureUtil;
 public class HashRangeExclusiveStickyKeyConsumerSelector implements StickyKeyConsumerSelector {
     private final int rangeSize;
     private final Range keyHashRange;
-    private final ConcurrentSkipListMap<Integer, Consumer> rangeMap;
+    private final ConcurrentSkipListMap<Integer, Pair<Range, Consumer>> rangeMap;
 
     public HashRangeExclusiveStickyKeyConsumerSelector() {
         this(DEFAULT_RANGE_SIZE);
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
```diff
@@ -73,47 +74,41 @@ private synchronized Optional<ImpactedConsumersResult> internalAddConsumer(Consu
...
-            } else {
-                if (prev.getValue().equals(entry.getValue())) {
-                    result.add(new HashRangeAssignment(Range.of(prev.getKey(), entry.getKey()), entry.getValue()));
-                }
-                prev = null;
-            }
+        for (Map.Entry<Integer, Pair<Range, Consumer>> entry : rangeMap.entrySet()) {
+            Range assignedRange = entry.getValue().getLeft();
+            Consumer assignedConsumer = entry.getValue().getRight();
+            result.add(new HashRangeAssignment(assignedRange, assignedConsumer));
```
poorbarcode, line 94: Since the method `getConsumerKeyHashRanges()` is the unique caller of the method `getConsumerHashAssignmentsSnapshot()`, we can overwrite the method `getConsumerKeyHashRanges()` in this class, which is simpler: loop consumers and put all `consumer.getKeySharedMeta().getHashRangesList()`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
```diff
@@ -129,10 +124,25 @@ private synchronized CompletableFuture<Void> validateKeySharedMeta(Consumer cons
...
+            // 1. Validate: check if start > end for the current range
+            if (currentRange.getStart() > currentRange.getEnd()) {
                 return FutureUtil.failedFuture(
-                        new BrokerServiceException.ConsumerAssignException("Fixed hash range start > end"));
+                        new BrokerServiceException.ConsumerAssignException("Fixed hash range start > end for range: "
+                                + "[" + currentRange.getStart() + "," + currentRange.getEnd() + "]"));
+            }
+            // 2. Validate: check for overlaps with the next range in the sorted list
+            if (i < sortedRanges.size() - 1) {
+                IntRange nextRange = sortedRanges.get(i + 1);
```
poorbarcode, line 139: When `i == 0 and sortedRanges.size == 1`, the line may face an `IndexOutOfBoundsException`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
```diff
@@ -143,34 +153,38 @@ private synchronized CompletableFuture<Void> validateKeySharedMeta(Consumer cons
...
                 }
             }
         }
         return null;
     }
 
-    Map<Integer, Consumer> getRangeConsumer() {
+
+    private boolean checkRangesOverlap(IntRange range1, Range range2) {
+        return Math.max(range1.getStart(), range2.getStart()) <= Math.min(range1.getEnd(), range2.getEnd());
```
poorbarcode, line 180: Can we allow the intersection of one point between two ranges like follows?
- `range 1`: `[0-100]`
- `range 2`: `[100-200]`

lhotari: It shouldn't be allowed unless the old code behaved that way? I guess it would be useful to have a test for this and run it with the old implementation unless it's already tested.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
```diff
@@ -143,34 +153,38 @@ private synchronized CompletableFuture<Void> validateKeySharedMeta(Consumer cons
...
+                    return conflictAfterStart.getValue().getRight();
                 }
             }
         }
         return null;
     }
 
-    Map<Integer, Consumer> getRangeConsumer() {
+
+    private boolean checkRangesOverlap(IntRange range1, Range range2) {
```
lhotari: make it static since it's not using instance fields

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelectorTest.java
```diff
@@ -89,6 +89,38 @@ public void testConsumerSelect() throws ExecutionException, InterruptedException
         Assert.assertNull(selectedConsumer);
     }
 
+
+    @Test
+    public void testConsumerSelectWithMultipRanges() throws ExecutionException, InterruptedException {
```
lhotari: Multip -> Multiple

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelectorTest.java
```diff
@@ -189,6 +221,32 @@ public void testGetConsumerKeyHashRangesWithSameConsumerName() throws Exception
         }
     }
 
+    @Test
+    public void testOneConsumerRangeConflict() throws ExecutionException, InterruptedException {
```
lhotari: It's a bit unclear from the test name what this is testing. Perhaps testAddingConsumerWithMultipleRangeConflicts is what this does and how this differs from following tests?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
shibd: > When i == 0 and sortedRanges.size == 1

This case will not go this line.

```
if (i < sortedRanges.size() - 1) {

i = 0
sortedRanges.size - 1 = 0
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
shibd: But we still need to implement the `getConsumerHashAssignmentsSnapshot` method, right?

If we were to remove it from the interface, we would have to modify all three implementing classes. I don't think it's necessary to do that in this PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
shibd: I'm wondering if allowing this might create some ambiguity. For instance, which range would a value like 100 fall into?

I tested it, and this wasn't allowed before this PR, and the client-side also prevents this. So, the behavior is consistent with how it was before. 


```
java.lang.IllegalArgumentException: Ranges for KeyShared policy with overlap between [0, 9999] and [9999, 29999]

	at org.apache.pulsar.client.api.KeySharedPolicy$KeySharedPolicySticky.validate(KeySharedPolicy.java:118)
	at org.apache.pulsar.client.impl.ConsumerBuilderImpl.keySharedPolicy(ConsumerBuilderImpl.java:525)
	at org.apache.pulsar.client.api.KeySharedSubscriptionTest.testCustomStickyRange(KeySharedSubscriptionTest.java:2579)
```

https://github.com/apache/pulsar/blob/14129e352e4bb4e06cbf2c6c581edcdbaf9f6775/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/KeySharedPolicy.java#L105-L123

lhotari: @shibd I think `<=` should be `<` in the checkRangesOverlap method to be consistent with the `Range.intersect` method logic. Please make the method static and add tests for it too so that you test the overlap rule for that method.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
poorbarcode: I am thinking that maybe we can cache the `hash range assignments` after adding or removing consumers, which may help for GC(to avoid re-creating objects typed `HashRangeAssignment `)

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
poorbarcode: > Math.max(range1.getStart(), range2.getStart()) <= Math.min(range1.getEnd(), range2.getEnd());

After discussing with @shibd, we consider that this code has no bugs

shibd: `end >=  start` is the same as `start <= end`

lhotari: @poorbarcode @shibd  the method isn't tested. The point is that `<=` allows the boundary to overlap if I'm not mistaken. That's why I think the method should be tested and `<=` should be changed to `<`.

shibd: hi,  @lhotari  Let's say we have two ranges.

- range1:   [2, 5]
- range2:  [5, 10]

```
    private static boolean checkRangesOverlap(IntRange range1, IntRange range2) {
     
                                  5                                            5
        return Math.max(range1.getStart(), range2.getStart()) <= Math.min(range1.getEnd(), range2.getEnd());
    }
```


5  == 5

So, this method return `true`, that means two ranges is `Overlap`


If we change to `<`

```
    private static boolean checkRangesOverlap(IntRange range1, IntRange range2) {
     
                                  5                                            5
        return Math.max(range1.getStart(), range2.getStart()) <= Math.min(range1.getEnd(), range2.getEnd());
    }
```

5 !< 5

return false, This is unexpected.



I added `testShouldConflictConsumerWithBoundaryRanges` to cover this case.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelectorTest.java
shibd: I changed name to `testShouldConflictConsumerWithSelfOverlappingRanges`

This test is to ensure that a single consumer is not allowed if its own hash ranges conflict with each other. the client-side already prevents this case, but on the broker-side, having an additional layer of validation makes it safer.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/HashRangeExclusiveStickyKeyConsumerSelector.java
lhotari: Yes that's clear now. 

What I meant is to have a unit test direct for the method. making the method static and default visibility allows testing in isolation to test all cases. For example:

```java
@Test
    public void testCheckRangesOverlap() {
        // Overlap in the middle
        Assert.assertTrue(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(1).setEnd(5),
                new IntRange().setStart(3).setEnd(4)));

        // One contains the other
        Assert.assertTrue(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(0).setEnd(10),
                new IntRange().setStart(2).setEnd(8)));

        // Touching at boundary (shared endpoint counts as overlap)
        Assert.assertTrue(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(1).setEnd(2),
                new IntRange().setStart(2).setEnd(4)));

        // Disjoint - gap between ranges
        Assert.assertFalse(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(1).setEnd(2),
                new IntRange().setStart(3).setEnd(4)));

        // Disjoint - larger gap between ranges
        Assert.assertFalse(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(1).setEnd(2),
                new IntRange().setStart(4).setEnd(5)));

        // Disjoint - first after second
        Assert.assertFalse(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(8).setEnd(10),
                new IntRange().setStart(1).setEnd(7)));

        // Touching at upper/lower boundary in reversed order
        Assert.assertTrue(HashRangeExclusiveStickyKeyConsumerSelector.checkRangesOverlap(
                new IntRange().setStart(8).setEnd(10),
                new IntRange().setStart(10).setEnd(12)));
    }
```

lhotari: The name `checkRangesOverlap` is a bit confusing. I first understood that it's working in a way where `true` means that it doesn't overlap and `false` means that it overlaps. It's the other way around.
Renaming the method to `doesOverlap` would address that confusion.

shibd: This is a private static method.

I don't think a separate test is necessary for this. The test coverage within the addConsumer test should be sufficient.

lhotari: Well the naming is confusing at least for code maintenance. renaming from `checkRangesOverlap` to `doRangesOverlap` would resolve this confusion.

lhotari: Although it's a private method, testing such utility methods in isolation is useful. It improves clarity for code reviews too when you can see from the test code what something is intended to do.

BewareMyPower: I think `areRangesOverlapped` or `rangesOverlapped` might be a better name.  `doRangesOverlap` is still confusing to me.

The `checkXxx` names are also widely used in the code base to return a boolean, e.g. `Consumer#checkAndApplyTopicMigration`.

image

shibd: Thanks for all comments, I changed name to `areRangesOverlapping`

lhotari (approved):
LGTM, great job @shibd

Issue #24697, opened 2025-09-03
[Bug] Consumer stops consuming due to inconsistent state between dispatcher and cursor


dragonls:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

Broker version: 3.0.7

### Issue Description

We observed a case where a Pulsar consumer suddenly stops consuming messages (throughput drops to 0) while the system remains stable.

<img width="1872" height="537" alt="Image" src="https://github.com/user-attachments/assets/d641c64c-4023-49f6-b05b-49edf311ef76" />

I have caught a memory dump of broker and seems find a state inconsistency between `PersistentDispatcherSingleActiveConsumer` and `NonDurableCursorImpl`:
1. The dispatcher's `havePendingRead` flag remains `true` (indicating an ongoing read operation)
2. However, the cursor shows:
    - `pendingReadOps = 0`
    - `waitingReadOp = null`
    - No visible read operations in progress

There is nothing wrong with the consumer/reader, which is just waiting for the broker to send message to the consumer with no  `incomingMessages`.

If I unload the specific topic partition, everything turns back to normal.

### Error messages

```text
2025-09-01T16:20:33,644+0800 [broker-topic-workers-OrderedExecutor-5-0] ERROR org.apache.bookkeeper.common.util.SingleThreadExecutor - Error while running task: Object has been recycled already.
java.lang.IllegalStateException: Object has been recycled already.
	at io.netty.util.Recycler$DefaultHandle.toAvailable(Recycler.java:284) ~[io.netty-netty-common-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.util.Recycler$LocalPool.release(Recycler.java:337) ~[io.netty-netty-common-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.util.Recycler$DefaultHandle.recycle(Recycler.java:257) ~[io.netty-netty-common-4.1.113.Final.jar:4.1.113.Final]
[...]

PR #20522, opened 2023-06-07
[improve][broker] Choose random thread for consumerFlow in PersistentDispatcherSingleActiveConsumer

AnonHxy:
### Motivation

Currently, all subscriptions of one topic will do `consuemrFlow` action in a single thread, which is chosen by topicName:
```
this.topicExecutor = topic.getBrokerService().getTopicOrderedExecutor().chooseThread(topicName);
```

If there is a large number of subscriptions in a topic,  all the work will focus on one thread ---- the chosen thread, which will reduce the consume performance.  So this this patch , I'd like to choose a ramdom thread for `consumerFlow` in `PersistentDispatcherSingleActiveConsumer` to improve the consume performance.

### Modifications

*  `topic.getBrokerService().getTopicOrderedExecutor().chooseThread(topicName);` -> `topic.getBrokerService().getTopicOrderedExecutor().chooseThread();`
*  `this.topicExecutor ` -> `this.executor`

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository 

PR in forked repository: https://github.com/AnonHxy/pulsar/pull/42

PR #24741, opened 2025-09-15
[fix][broker] Prevent unexpected recycle failure in dispatcher's read callback

BewareMyPower:
Main Issue: #24697

### Motivation

Using callback-based read API with a Netty recyclable object is buggy that once the `ReadEntriesCtx` object has been recycled twice.

### Modifications

- Add a future-based API wrapper on `asyncReadEntriesWithSkipOrWait`
- Add `testConcurrentRead` to verify concurrent read will fail
- Modify the `CompactedTopicUtils#asyncReadCompactedEntries` API to return a future as well
- Remove `ReadEntriesCtx`

In extreme cases like #24697, it's safe to deliver messages to the consumer after that.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


BewareMyPower:
While the reason of https://github.com/apache/pulsar/issues/24697 is still in investigation, this PR aims at the hard-to-debug recycling issue that did happen in real world by eliminating the recycler in read path

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumer.java
```diff
@@ -458,29 +454,26 @@ protected Pair<Integer, Long> calculateToRead(Consumer consumer) {
...
     }
 
-    @Override
-    public void readEntriesFailed(ManagedLedgerException exception, Object ctx) {
-        executor.execute(() -> internalReadEntriesFailed(exception, ctx));
-    }
-
-    private synchronized void internalReadEntriesFailed(ManagedLedgerException exception, Object ctx) {
+    @VisibleForTesting
+    public synchronized Void readEntriesFailed(Throwable throwable, Consumer consumer) {
```
lhotari: what is the reason for using `Void` return type?

BewareMyPower: Just saves a `return null` in `exceptionallyAsync`

lhotari: It doesn't seem to be a good idea to use `Void` return type for public methods.

BewareMyPower: It makes sense. `whenCompleteAsyn` combined with an `exceptionally` similar to `catch (Throwable)` might be good here, let me revisit the code later

BewareMyPower: I've updated the code, PTAL again

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumer.java
```diff
@@ -154,20 +153,16 @@ protected void cancelPendingRead() {
...
-        Consumer readConsumer = readEntriesCtx.getConsumer();
-        long epoch = readEntriesCtx.getEpoch();
-        readEntriesCtx.recycle();
+    private synchronized void readEntriesComplete(List<Entry> entries, Consumer readConsumer, long epoch) {
         if (log.isDebugEnabled()) {
             log.debug("[{}-{}] Got messages: {}", name, readConsumer, entries.size());
         }
 
+        if (!havePendingRead) {
+            log.warn("[{}-{}] There is no pending read when entries {} are called", name, readConsumer,
```
nodece: This log could fire very frequently, do you consider log.debug?

BewareMyPower: When could `havePendingRead` be false in `readEntriesComplete`?

BewareMyPower: I choose to remove these logs, `havePendingRead` seems never helpful and maybe we can replace it with a better design in future. For now, adding such logs for debugging is also not meaningful

nodece (approved):
LGTM

coderzc (approved):

lhotari:
Cherry-picking to branch-3.0 causes merge conflicts mainly due to #22285 being missing in branch-3.0, perhaps others. It was fairly easy to backport to branch-3.3. @BewareMyPower can you please handle cherry-picking to branch-3.0? I already resolved other merge conflicts in branch-3.3 commit 65ddd44 so you should be able to cherry-pick that to branch-3.0 once the preconditions in branch-3.0 have been addressed.

BewareMyPower:
Okay, I will do that

BewareMyPower:
@lhotari I've completed the cherry-picking, as well as cherry-picking https://github.com/apache/pulsar/pull/20522 before this PR

PR #22598, opened 2024-04-26
[improve][broker] Don't use forkjoin pool by default for deleting partitioned topics

lhotari:
### Motivation

While investigating some recent test failures and Pulsar CI OOME issues, #22588, the heap dumps revealed that
the ForkJoin pool's work queue is retaining a lot of CompletableFuture related instances. 

image

The heap dump confirms that these are originating from `NamespaceResources.PartitionedTopicResources#runWithMarkDeleteAsync` method in this location:
https://github.com/apache/pulsar/blob/bbdc173e80157296ff0475cfdc4e36f5d063a7df/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/NamespaceResources.java#L345-L374

The assumption is that graceful shutdown in tests without causing memory leaks will be more achievable with the provided executor since the instance isn't shared like the Fork join common pool executor.

### Modifications

Instead of using the forkjoin pool by default in `NamespaceResources.PartitionedTopicResources#runWithMarkDeleteAsync`, use a specific executor. For Pulsar broker, this would be the executor returned by `PulsarService#getExecutor()`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #22589, opened 2024-04-25
[improve][broker] Close TopicPoliciesService to allow Pulsar broker graceful shutdown

lhotari:
### Motivation

There doesn't seem to be a proper handling for TopicPoliciesService shutdown.
This would be needed to allow the Pulsar broker to shutdown gracefully.
This also prevents memory leaks in tests. This problem was detected during the investigation of test OOMEs.
See #22588 for details.

### Modifications

- Add close method to TopicPoliciesService and implement it
- Prevent creating new cached writers or readers after the service has been closed
- Use Pulsar's executor for the AsyncLoader in TopicPoliciesService
- make shutdown synchronous
- Improve the close method of SystemTopicTxnBufferSnapshotService

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24745, opened 2025-09-16
[improve][broker] Remove ForkJoin common pool in caffeine cache

zhanghaou:
### Motivation
Same to https://github.com/apache/pulsar/pull/24383

Currently, the usage of Caffeine in pulsar code is inconsistent. Some asynchronous builds use the default ForkJoin common pool, while others specify a custom thread pool, and there are also synchronous builds.

The ForkJoin pool has a specific bug that can prevent tasks from being executed after being submitted to the thread pool. This issue has been fixed in Java 21.0.8, but it still exists in Java 17.
https://bugs.openjdk.org/browse/JDK-8330017
https://bugs.openjdk.org/browse/JDK-8351933

If the ForkJoin common pool gets blocked, it can prevent all dependent caches from fetching data.

### Modifications

1. When executor in pulsarService is available, use pulsarService executor
2. Add new executor in AbstractMetadataStore.java, which use 4 corePoolSize(hard code, need set to be configured?)
3. Use MoreExecutors.directExecutor() for other caffeines.

**The choice of thread pool may not be the most appropriate. Please provide suggestions for improvement, thanks.**

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
@@ -77,6 +77,7 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
     private final CopyOnWriteArrayList<Consumer<SessionEvent>> sessionListeners = new CopyOnWriteArrayList<>();
     protected final String metadataStoreName;
     protected final ScheduledExecutorService executor;
+    protected final ScheduledExecutorService cacheExecutor;
     private final AsyncLoadingCache<String, List<String>> childrenCache;
     private final AsyncLoadingCache<String, Boolean> existsCache;
     private final CopyOnWriteArrayList<MetadataCacheImpl<?>> metadataCaches = new CopyOnWriteArrayList<>();
@@ -100,9 +101,12 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
         this.executor = new ScheduledThreadPoolExecutor(1,
                 new DefaultThreadFactory(
                         StringUtils.isNotBlank(metadataStoreName) ? metadataStoreName : getClass().getSimpleName()));
+        this.cacheExecutor = new ScheduledThreadPoolExecutor(4,
+                new DefaultThreadFactory("metadata-cache-loader"));
         registerListener(this);
 
         this.childrenCache = Caffeine.newBuilder()
+                .executor(cacheExecutor)
                 .recordStats()
                 .refreshAfterWrite(CACHE_REFRESH_TIME_MILLIS, TimeUnit.MILLISECONDS)
                 .expireAfterWrite(CACHE_REFRESH_TIME_MILLIS * 2, TimeUnit.MILLISECONDS)
@@ -126,6 +130,7 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
         CacheMetricsCollector.CAFFEINE.addCache(metadataStoreName + "-children", childrenCache);
 
         this.existsCache = Caffeine.newBuilder()
+                .executor(cacheExecutor)
                 .recordStats()
                 .refreshAfterWrite(CACHE_REFRESH_TIME_MILLIS, TimeUnit.MILLISECONDS)
                 .expireAfterWrite(CACHE_REFRESH_TIME_MILLIS * 2, TimeUnit.MILLISECONDS)
@@ -249,7 +254,7 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
         JavaType typeRef = TypeFactory.defaultInstance().constructSimpleType(clazz, null);
         String cacheName = StringUtils.isNotBlank(metadataStoreName) ? metadataStoreName : getClass().getSimpleName();
         MetadataCacheImpl<T> metadataCache =
-                new MetadataCacheImpl<T>(cacheName, this, typeRef, cacheConfig, this.executor);
+                new MetadataCacheImpl<T>(cacheName, this, typeRef, cacheConfig, this.cacheExecutor);
         metadataCaches.add(metadataCache);
         return metadataCache;
     }
@@ -258,7 +263,7 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
     public <T> MetadataCache<T> getMetadataCache(TypeReference<T> typeRef, MetadataCacheConfig cacheConfig) {
         String cacheName = StringUtils.isNotBlank(metadataStoreName) ? metadataStoreName : getClass().getSimpleName();
         MetadataCacheImpl<T> metadataCache =
-                new MetadataCacheImpl<T>(cacheName, this, typeRef, cacheConfig, this.executor);
+                new MetadataCacheImpl<T>(cacheName, this, typeRef, cacheConfig, this.cacheExecutor);
         metadataCaches.add(metadataCache);
         return metadataCache;
     }
@@ -267,8 +272,7 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
     public <T> MetadataCache<T> getMetadataCache(String cacheName, MetadataSerde<T> serde,
                                                  MetadataCacheConfig cacheConfig) {
         MetadataCacheImpl<T> metadataCache =
-                new MetadataCacheImpl<>(cacheName, this, serde, cacheConfig,
-                        this.executor);
+                new MetadataCacheImpl<>(cacheName, this, serde, cacheConfig, this.cacheExecutor);
         metadataCaches.add(metadataCache);
         return metadataCache;
     }
@@ -556,6 +560,8 @@ public abstract class AbstractMetadataStore implements MetadataStoreExtended, Co
 
     @Override
     public void close() throws Exception {
+        cacheExecutor.shutdownNow();
+        cacheExecutor.awaitTermination(10, TimeUnit.SECONDS);
         executor.shutdownNow();
         executor.awaitTermination(10, TimeUnit.SECONDS);
         this.metadataStoreStats.close();
--- a/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/JwksCache.java
+++ b/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/JwksCache.java
@@ -34,6 +34,7 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.github.benmanes.caffeine.cache.AsyncCacheLoader;
 import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.util.concurrent.MoreExecutors;
 import io.kubernetes.client.openapi.ApiCallback;
 import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.openapi.ApiException;
@@ -88,6 +89,7 @@ public class JwksCache {
             }
         };
         this.cache = Caffeine.newBuilder()
+                .executor(MoreExecutors.directExecutor())
                 .recordStats()
                 .maximumSize(maxSize)
                 .refreshAfterWrite(refreshAfterWriteSeconds, TimeUnit.SECONDS)
--- a/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/OpenIDProviderMetadataCache.java
+++ b/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/OpenIDProviderMetadataCache.java
@@ -31,6 +31,7 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.github.benmanes.caffeine.cache.AsyncCacheLoader;
 import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.util.concurrent.MoreExecutors;
 import io.kubernetes.client.openapi.ApiCallback;
 import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.openapi.ApiException;
@@ -78,6 +79,7 @@ class OpenIDProviderMetadataCache {
             }
         };
         this.cache = Caffeine.newBuilder()
+                .executor(MoreExecutors.directExecutor())
                 .recordStats()
                 .maximumSize(maxSize)
                 .refreshAfterWrite(refreshAfterWriteSeconds, TimeUnit.SECONDS)
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/ManagedLedgerClientFactory.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/ManagedLedgerClientFactory.java
@@ -21,6 +21,7 @@ package org.apache.pulsar.broker;
 import com.github.benmanes.caffeine.cache.AsyncCache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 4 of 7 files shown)


lhotari:
> This issue has been fixed in Java 21.0.8, but it still exists in Java 17.
> https://bugs.openjdk.org/browse/JDK-8330017
> https://bugs.openjdk.org/browse/JDK-8351933

Java 17.0.17 will include [the fix](https://bugs.openjdk.org/browse/JDK-8351935) and the planned release date is October 21st: https://wiki.openjdk.org/display/JDKUpdates/JDK+17u

lhotari:
> Currently, the usage of Caffeine in pulsar code is inconsistent. Some asynchronous builds use the default ForkJoin common pool, while others specify a custom thread pool, and there are also synchronous builds.

why is this a problem if there are different ways to use Caffeine?
 
> The ForkJoin pool has a specific bug that can prevent tasks from being executed after being submitted to the thread pool. This issue has been fixed in Java 21.0.8, but it still exists in Java 17. https://bugs.openjdk.org/browse/JDK-8330017 https://bugs.openjdk.org/browse/JDK-8351933

Since the fix for Java 17 will be released in 17.0.17 on October 21st, it will soon be addressed.

> If the ForkJoin common pool gets blocked, it can prevent all dependent caches from fetching data.

I understand that it's severe when that happens. How often have you observed this bug? Does it require that that JVM has been running for very long or is it just random when there's heavy load?

zhanghaou:
> why is this a problem if there are different ways to use Caffeine?

It’s not necessarily a problem, but I’d like to highlight that certain parts of the code explicitly specify a thread pool, whereas other parts continue to use the default thread pool.

> I understand that it's severe when that happens. How often have you observed this bug? Does it require that that JVM has been running for very long or is it just random when there's heavy load?

It occurs very rarely, only after running for a long time, and it doesn’t seem to be related to the load.

lhotari:
> It occurs very rarely, only after running for a long time, and it doesn’t seem to be related to the load.

what is "long time" here? days? weeks?

lhotari:
> It’s not necessarily a problem, but I’d like to highlight that certain parts of the code explicitly specify a thread pool, whereas other parts continue to use the default thread pool.

For blocking IO, ForkJoinPool is a reasonable solution to avoid blocking Netty IO threads or the Metadata store thread. VirtualThreads in Java 21 could be a better option eventually. Instead of removing ForkJoinPool usage blindly, there should be more analysis of each case so that we don't cause performance regressions. In Pulsar, the execution model isn't clean and more effort should be put in designing how thread pools and task execution should behave in general. This problem has existed for a long time so it's waiting for contributors. :)

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
```diff
@@ -100,9 +101,12 @@ protected AbstractMetadataStore(String metadataStoreName, OpenTelemetry openTele
         this.executor = new ScheduledThreadPoolExecutor(1,
                 new DefaultThreadFactory(
                         StringUtils.isNotBlank(metadataStoreName) ? metadataStoreName : getClass().getSimpleName()));
+        this.cacheExecutor = new ScheduledThreadPoolExecutor(4,
```
nodece: Do you consider using CPU cores?

zhanghaou:
> what is "long time" here? days? weeks?

We’ve gone through both weeks and months.

Add more info：
When ForkJoin common pool stop executing tasks, use arthas can get info of common pool. 
Then common pool is running, but have some submissions in queue.

@ForkJoinPool[
    SWIDTH=@Integer[16],
    SMASK=@Integer[65535],
    MAX_CAP=@Integer[32767],
    SQMASK=@Integer[126],
    UNSIGNALLED=@Integer[-2147483648],
    SS_SEQ=@Integer[65536],
    QLOCK=@Integer[1],
    OWNED=@Integer[1],
    FIFO=@Integer[65536],
    SHUTDOWN=@Integer[262144],
    TERMINATED=@Integer[524288],
    STOP=@Integer[-2147483648],
    QUIET=@Integer[1073741824],
    DORMANT=@Integer[-1073741824],
    INITIAL_QUEUE_CAPACITY=@Integer[8192],
    MAXIMUM_QUEUE_CAPACITY=@Integer[67108864],
    TOP_BOUND_SHIFT=@Integer[10],
    defaultForkJoinWorkerThreadFactory=@DefaultForkJoinWorkerThreadFactory[java.util.concurrent.ForkJoinPool$DefaultForkJoinWorkerThreadFactory@43f5e9a3],
    modifyThreadPermission=@RuntimePermission[("java.lang.RuntimePermission" "modifyThread")],
    **common=@ForkJoinPool[java.util.concurrent.ForkJoinPool@2253dae0[Running, parallelism = 3, size = 2, active = 32770, running = 0, steals = 181625, tasks = 0, submissions = 27]],**
    COMMON_PARALLELISM=@Integer[3],
    COMMON_MAX_SPARES=@Integer[256],
    poolNumberSequence=@Integer[0],
    DEFAULT_KEEPALIVE=@Long[60000],
    TIMEOUT_SLOP=@Long[20],
    DEFAULT_COMMON_MAX_SPARES=@Integer[256],
    SEED_INCREMENT=@Integer[-1640531527],
    SP_MASK=@Long[4294967295],
    UC_MASK=@Long[-4294967296],
    RC_SHIFT=@Integer[48],
    RC_UNIT=@Long[281474976710656],
    RC_MASK=@Long[-281474976710656],
    TC_SHIFT=@Integer[32],
    TC_UNIT=@Long[4294967296],
    TC_MASK=@Long[281470681743360],
    ADD_WORKER=@Long[140737488355328],
    stealCount=@Long[181603],
    keepAlive=@Long[60000],
    indexSeed=@Integer[-1126400000],
    bounds=@Integer[16842750],
    mode=@Integer[3],
    workQueues=@WorkQueue[][isEmpty=false;size=8],
    workerNamePrefix=@String[ForkJoinPool.commonPool-worker-],
    factory=@DefaultForkJoinWorkerThreadFactory[java.util.concurrent.ForkJoinPool$DefaultForkJoinWorkerThreadFactory@43f5e9a3],
    ueh=null,
    saturate=null,
    ctl=@Long[9223372035728539649],
    CTL=@FieldInstanceReadWrite[java.lang.invoke.VarHandleLongs$FieldInstanceReadWrite@1732b0fd],
    MODE=@FieldInstanceReadWrite[java.lang.invoke.VarHandleInts$FieldInstanceReadWrite@3b3414a8],
    QA=@Array[java.lang.invoke.VarHandleObjects$Array@221b627],
    $assertionsDisabled=@Boolean[true],
]

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
zhanghaou: Updated.

zhanghaou:
> For blocking IO, ForkJoinPool is a reasonable solution to avoid blocking Netty IO threads or the Metadata store thread. VirtualThreads in Java 21 could be a better option eventually. 

I agree that VirtualThreads is better. However, given the current situation, ForkJoin common pool is not very suitable.

> Instead of removing ForkJoinPool usage blindly, there should be more analysis of each case so that we don't cause performance regressions. 

https://github.com/apache/pulsar/pull/22589  this pr change executor from ForkJoinPool to pulsar service executor. It seems there’s no analysis regarding performance regression. 

> In Pulsar, the execution model isn't clean and more effort should be put in designing how thread pools and task execution should behave in general. This problem has existed for a long time so it's waiting for contributors. :)

Look forward to it.

lhotari (changes requested):
I don't support making this change, since it's blindly making changes across the code base. For example in JwksCache, it's necessary to have an actual executor to run tasks in background. The `refreshAfterWrite` will handle that. This PR breaks existing behavior.

zhanghaou:
> I don't support making this change, since it's blindly making changes across the code base. For example in JwksCache, it's necessary to have an actual executor to run tasks in background. The `refreshAfterWrite` will handle that. This PR breaks existing behavior.

Added a new executor to address it. Please take a review again. Thanks.

BewareMyPower (approved):

BewareMyPower:
> Since the fix for Java 17 will be released in 17.0.17 on October 21st, it will soon be addressed.

I saw this comment just now, it should be fine to wait for the JDK patch, so I dismissed my approval.

Using a dedicated executor is reasonable. But the `ThreadPoolExecutor` based executor is not better than ForkJoinPool, which uses work-stealing and is leveraged by the implementation of virtual threads. A simple difference is that if a task takes too long to complete, the queued tasks in the same thread don't have a chance to execute.

`ForkJoinPool` is a better choice for independent tasks like those tasks to avoid blocking Netty IO threads or the Metadata store thread in Pulsar.

zhanghaou:
> I saw this comment just now, it should be fine to wait for the JDK patch, so I dismissed my approval.

Waiting for the JDK upgrade and using the latest version can indeed solve this problem, but wouldn't it be better if we could solve this issue without relying on a JDK upgrade? 

If the user doesn't know that they need to upgrade the JDK to a specified version, there will always be a risk of blocking. That's really scary!

> Using a dedicated executor is reasonable. 

That's the purpose of this PR.

> But the `ThreadPoolExecutor` based executor is not better than ForkJoinPool, which uses work-stealing and is leveraged by the implementation of virtual threads. A simple difference is that if a task takes too long to complete, the queued tasks in the same thread don't have a chance to execute.
> 
> `ForkJoinPool` is a better choice for independent tasks like those tasks to avoid blocking Netty IO threads or the Metadata store thread in Pulsar.

Use `ThreadPoolExecutor` based executor is refer to the existing code https://github.com/apache/pulsar/pull/22589 

@BewareMyPower

BewareMyPower:
Technically, there are many references in Pulsar that use `xxxAsync` without the executor argument. We have to pass a dedicated executor in all places.  This will change the code base a lot, which is redundant once we upgraded to a newer JDK that includes that fix.

To prevent this use cases, we also have to disallow any `CompletableFuture`'s `Async` method without specifying the executor argument. It's hard to add this limit for a rarely happened case which will be fixed soon.

It's a JDK standard API that should be widely used. Many projects should be affected rather than Pulsar itself. So I guess it's why @lhotari asked `what is "long time" here? days? weeks?` before.

BewareMyPower:
A simple question, if I'm going to change the dedicated thread pool to `ForkJoinedPool` in future, could it be blocked due to the bug that is fixed in recent JDK versions?

zhanghaou:
> A simple question, if I'm going to change the dedicated thread pool to `ForkJoinedPool` in future, could it be blocked due to the bug that is fixed in recent JDK versions?

If use ForkJoinPool as the dedicated thread pool but not the common pool,  even if it is blocked, the impact is localized; Unlike now when using the ForkJoin common pool, where the impact is global.

zhanghaou:
> Many projects should be affected rather than Pulsar itself. 

Hi, @BewareMyPower 
Could you share which Apache project use the same ForkJoin common pool in various places?

BewareMyPower:
In principal, a Java project's code base should assume the JDK is reliable.

`ForkJoinPool` is widely used within JDK, not only for `CompletableFuture`:
- Parallel stream (`parallelStream()`) uses `ForkJoinPool` to execute sub-tasks (see `ForkJoinTask`)
- Virtual threads uses `ForkJoinPool` as its default scheduler (see `VirtualThread#createDefaultScheduler`)
- (not sure if there are some other similar classes provided by JDK)
- 3rd party Java infra projects like Caffine that use default `ForkJoinPool` as the default executor

Now, your question might become which Apache project uses parallel stream, virtual threads, Caffine cache (or any unknown dependency that uses the common `ForkJoinPool` as the default executor) without changing the default scheduler?

I don't have a tool to detect usages of `ForkJoinPool` in all open sourced or even only Apache projects, but this issue is beyond the scope of a Java project like Apache Pulsar. If you have suffered the bug in production and is eager to get a fix ASAP, you should:
- Fork the project and push the "fix" internally (I believe you have also done that)
- Build a JDK that includes the fix by yourself or turn to an existing JDK vendor

However, such a bug from JDK that rarely happens should not be a key factor that changes the code base a lot.

BewareMyPower:
In short, using a dedicated pool is okay. Continue using `ForkJoinPool` is okay as well. I'm not against using a different thread pool for Caffine cache only, so I didn't leave a comment or requested change to block merging this PR as well. But this bug of JDK should not be a reason to remove `ForkJoinPool` everywhere. 

I'd like to hear @lhotari's voice again

lhotari:
> In short, using a dedicated pool is okay. Continue using `ForkJoinPool` is okay as well. I'm not against using a different thread pool for Caffine cache only, so I didn't leave a comment or requested change to block merging this PR as well. But this bug of JDK should not be a reason to remove `ForkJoinPool` everywhere.
> 
> I'd like to hear @lhotari's voice again

@BewareMyPower I think your previous comment already covered most and was a really good answer. 

Just adding a few items:
* There's already a problem in Pulsar that existing executors aren't properly used and designed.
  * For example, Netty IO threads should never be blocked, but they are: https://github.com/apache/pulsar/issues/23865
  * `synchronized` blocks under heavy concurrency will lead to contention that blocks threads. When a Netty IO thread is blocked, it will cause a "noisy neighbor" problem to all connections that share that connection
* An ordinary executor behaves in a different way than the ForkJoinPool. ForkJoinPool can expand the number of threads when it detects blocking and handle this efficiently. This might be relevant in some of the current usage of executors. Switching to use an ordinary executor could cause deadlocks in certain cases where the existing code doesn't deadlock because of ForkJoinPool.
* `MoreExecutor.directExecutor` shouldn't be used at all when `refreshAfterWrite` is used with Caffeine, since it would defeat the purpose of using background refreshing of the cached items. I don't see a reason why someone would want to make an async Caffeine cache synchronous. That doesn't make sense.

zhanghaou:
@lhotari
 
> * An ordinary executor behaves in a different way than the ForkJoinPool. ForkJoinPool can expand the number of threads when it detects blocking and handle this efficiently. This might be relevant in some of the current usage of executors. Switching to use an ordinary executor could cause deadlocks in certain cases where the existing code doesn't deadlock because of ForkJoinPool.

The purpose of this PR is to achieve thread-pool isolation, rather than to compare whether the ForkJoin pool is better than a regular thread pool. For example, in your previous PR https://github.com/apache/pulsar/pull/22589 , you also replaced the ForkJoin common pool with Pulsar’s executor.

> * `MoreExecutor.directExecutor` shouldn't be used at all when `refreshAfterWrite` is used with Caffeine, since it would defeat the purpose of using background refreshing of the cached items. I don't see a reason why someone would want to make an async Caffeine cache synchronous. That doesn't make sense.

https://github.com/apache/pulsar/pull/24745#issuecomment-3298091512  Maybe you miss the explanation here，I had added a new executor.

zhanghaou:
> I don't have a tool to detect usages of `ForkJoinPool` in all open sourced or even only Apache projects, but this issue is beyond the scope of a Java project like Apache Pulsar. If you have suffered the bug in production and is eager to get a fix ASAP, you should:
> 
> * Fork the project and push the "fix" internally (I believe you have also done that)
> * Build a JDK that includes the fix by yourself or turn to an existing JDK vendor
> 
> However, such a bug from JDK that rarely happens should not be a key factor that changes the code base a lot.

I’ve already fixed this issue internally and deployed it to production. I’m submitting this PR to share the bug with the community and show how it can be resolved.

Although this issue may occur infrequently, I think it’s still important that we should address it.

zhanghaou:
https://github.com/apache/pulsar/pull/22598  
The problem encountered in this PR is likely caused by the bug in the ForkJoin pool. And the fix is to replace it, same to this pr.

lhotari:
Java 17.0.17 has been released. This change is no longer needed as a workaround for the Java bug. I'll initiate 3.0.15 release so that we'll have Java 17.0.17 used in the 3.0.15 docker image.

PR #24753, opened 2025-09-17
[fix][ml]Fix EOFException after enabled topics offloading

poorbarcode:
### Motivation

The steps of reproducing the issue:
- We have a ledger, which contains `5` entries, the last confirmed is `4`
- Enable offloading for a topic
- Read the first round: `[0 ~ 3]`
- Read the second round `[0 ~ 4]`, the error will occur.

```
2025-09-11T11:22:09,210+0000 [offloader-OrderedScheduler-1-0] ERROR org.apache.bookkeeper.mledger.offload.jcloud.impl.BlobStoreBackedReadHandleImpl - Failed to read entries 0 - 10 from the offloader in ledger 1670362
java.io.EOFException: null
	at java.base/java.io.DataInputStream.readFully(Unknown Source) ~[?:?]
	at java.base/java.io.DataInputStream.readInt(Unknown Source) ~[?:?]
	at org.apache.bookkeeper.mledger.offload.jcloud.impl.BlobStoreBackedReadHandleImpl.lambda$readAsync$2(BlobStoreBackedReadHandleImpl.java:171) ~[?:?]
	at org.apache.bookkeeper.common.util.SingleThreadSafeScheduledExecutorService$SafeRunnable.run(SingleThreadSafeScheduledExecutorService.java:45) ~[io.streamnative-bookkeeper-common-4.17.2.2.jar:4.17.2.2]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) ~[?:?]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty-netty-common-4.1.124.Final.jar:4.1.124.Final]
```

### Root cause

After the previous round of reading, the `offset` of the stream is the start position of the latest entry. If Pulsar tries to read `[0 ~ LAC]`, it will encounter the bottom logic branch: `skip the latest entry`, then the next round look will receive an `EOFException` because the stream is empty.

https://github.com/apache/pulsar/blob/v4.1.0/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImpl.java#L210

```java
if (entryId == nextExpectedId) {
    entries.add(LedgerEntryImpl.create(ledgerId, entryId, length, buf));
} else if (entryId > nextExpectedId && entryId < lastEntry) {
    seekToEntry(nextExpectedId);
} else if (entryId < nextExpectedId
    && !index.getIndexEntryForEntry(nextExpectedId).equals(index.getIndexEntryForEntry(entryId))) {
    seekToEntry(nextExpectedId);
} else if (entryId > lastEntry) {
    seekToEntry(nextExpectedId);
} else {
    long ignore = inputStream.skip(length);
}
```

### Modifications
- Fix the issue.
- Improve logs.
- Add `hashCode and equals` for `OffloadIndexEntryImpl`, since Pulsar calls `equals` here: https://github.com/apache/pulsar/blob/v4.1.0/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImpl.java#L193
- Seek first before the reading loop, to avoid another case of EOFException, which is thrown by `DataInputStream.readInt` if the value is unexpected.

```java
 public final int readInt() throws IOException {
    int ch1 = in.read();
    int ch2 = in.read();
    int ch3 = in.read();
    int ch4 = in.read();
    if ((ch1 | ch2 | ch3 | ch4) < 0)
        throw new EOFException(); // if the value is not an int, it throws an error here.
    return ((ch1 << 24) + (ch2 << 16) + (ch3 << 8) + (ch4 << 0));
}
```

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImpl.java
+++ b/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImpl.java
@@ -146,7 +146,40 @@ public class BlobStoreBackedReadHandleImpl implements ReadHandle, OffloadedLedge
 
             List<LedgerEntry> entryCollector = new ArrayList<LedgerEntry>();
             try {
-                readEntries(entryCollector);
+                if (firstEntry > lastEntry
+                        || firstEntry < 0
+                        || lastEntry > getLastAddConfirmed()) {
+                    promise.completeExceptionally(new BKException.BKIncorrectParameterException());
+                    return;
+                }
+                long entriesToRead = (lastEntry - firstEntry) + 1;
+                long expectedEntryId = firstEntry;
+                seekToEntry(firstEntry);
+                seekedAndTryTimes++;
+
+                while (entriesToRead > 0) {
+                    long currentPosition = inputStream.getCurrentPosition();
+                    int length = dataStream.readInt();
+                    if (length < 0) { // hit padding or new block
+                        seekToEntry(expectedEntryId);
+                        continue;
+                    }
+                    long entryId = dataStream.readLong();
+                    if (entryId == expectedEntryId) {
+                        entryOffsetsCache.put(ledgerId, entryId, currentPosition);
+                        ByteBuf buf = PulsarByteBufAllocator.DEFAULT.buffer(length, length);
+                        entryCollector.add(LedgerEntryImpl.create(ledgerId, entryId, length, buf));
+                        int toWrite = length;
+                        while (toWrite > 0) {
+                            toWrite -= buf.writeBytes(dataStream, toWrite);
+                        }
+                        entriesToRead--;
+                        expectedEntryId++;
+                    } else {
+                        handleUnexpectedEntryId(expectedEntryId, entryId);
+                    }
+                }
+                promise.complete(LedgerEntriesImpl.create(entryCollector));
             } catch (Throwable t) {
                 log.error("Failed to read entries {} - {} from the offloader in ledger {}",
                         firstEntry, lastEntry, ledgerId, t);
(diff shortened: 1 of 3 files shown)


tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImpl.java
```diff
@@ -121,108 +123,133 @@ public CompletableFuture<Void> closeAsync() {
...
-                        log.info("Expected to read {}, but read {}, which is greater than last entry {}",
-                            nextExpectedId, entryId, lastEntry);
-                        throw new BKException.BKUnexpectedConditionException();
-                    } else {
-                        long ignore = inputStream.skip(length);
-                    }
-                }
-
-                promise.complete(LedgerEntriesImpl.create(entries));
+                readEntries(entryCollector);
```
Technoboy-: Only this param seems confuse.

poorbarcode: Improved the code, please review again

Technoboy- (approved):

dao-jun (approved):

coderzc (approved):

Issue #24754, opened 2025-09-17
[Bug] Failure in topK bundle selection due to the transitivity Contract Violation in namespaceBundleStats' & BundleData's compareTo()


Meet0861:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

Pulsar version : 3.0.12 + cherry-picked - https://github.com/apache/pulsar/pull/22765
But issue is there in master as well.

### Issue Description

The **topK bundle** selection fails due to the sorting failure in the partition sort algo being used [here](https://github.com/apache/pulsar/blob/437d9f6a3ccfdfd327ac032c298728fcc43fc263/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/models/TopKBundles.java#L115). 
Partition sort uses, NamespaceBundleStats' + BundleData's custom ccompararable implementations where it checks throughput, connections, cache size etc against the defined threshold and when values are within the defined thresholds, they are considered "equal" (return 0), but this creates transitivity violations. 
It violates the transitivity property required by Java's Comparable interface, causing [Collections.sort()](https://github.com/apache/pulsar/blob/437d9f6a3ccfdfd327ac032c298728fcc43fc263/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/models/TopKBundles.java#L148) to potentially throw IllegalArgumentException with "Comparison method violates its general contract" error.
**Error Log**:
`2025-09-12T22:29:58.832386492+05:30 16:59:58.832 [pulsar-load-manager-1-1] WARN  org.apache.pulsar.broker.loadbalance.LoadResourceQuotaUpdaterTask - Error write resource quota - java.lang.IllegalArgumentException: Comparison method violates its general contract`

[...]

PR #24759, opened 2025-09-18
[fix][broker] transitivity violation during bundle comparison in topK bundles by NamespaceBundleStats,BundleData Comparator

Meet0861:
Fixes: [#24754](https://github.com/apache/pulsar/issues/24754)

### Motivation

The topK bundle selection fails due to the sorting failure in the partition sort algo being used [here](https://github.com/apache/pulsar/blob/437d9f6a3ccfdfd327ac032c298728fcc43fc263/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/models/TopKBundles.java#L115).
Partition sort uses, NamespaceBundleStats' + BundleData's custom ccompararable implementations where it checks throughput, connections, cache size etc against the defined threshold and when values are within the defined thresholds, they are considered "equal" (return 0), but this creates transitivity violations.
It violates the transitivity property required by Java's Comparable interface, causing [Collections.sort()](https://github.com/apache/pulsar/blob/437d9f6a3ccfdfd327ac032c298728fcc43fc263/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/models/TopKBundles.java#L148) to potentially throw IllegalArgumentException with "Comparison method violates its general contract" error.
Error Log:
`2025-09-12T22:29:58.832386492+05:30 16:59:58.832 [pulsar-load-manager-1-1] WARN  org.apache.pulsar.broker.loadbalance.LoadResourceQuotaUpdaterTask - Error write resource quota - java.lang.IllegalArgumentException: Comparison method violates its general contract`

Due to this failure, the job - writeBundleDataOnZooKeeper(leader broker writes bundle data aggregated from all brokers to metadata store) in modularLoadManager may fail [link](https://github.com/apache/pulsar/blob/437d9f6a3ccfdfd327ac032c298728fcc43fc263/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/ModularLoadManagerImpl.java#L1173) and can cause degradation in productions due to failure/inconsistencies in LB decisions

### Modifications

added strict NamespaceBundleStats comparator and BundleData comparator to be used for sorting for topk bundle comparison which is being used by ModularLoadManager and ExtensibleLoadManager

### Verifying this change

  - *Added unit tests to validate the fix*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ *] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/Meet0861/pulsar/tree/fix-transitivityViolationInNamespaceBundleStatsAndBundledataComparator

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/models/BundleDataComparator.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/models/BundleDataComparator.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.broker.loadbalance.extensions.models;
+
+import java.util.Comparator;
+import java.util.Map;
+import org.apache.pulsar.policies.data.loadbalancer.BundleData;
+import org.apache.pulsar.policies.data.loadbalancer.TimeAverageMessageData;
+
+/**
+ * A strict comparator for BundleData that provides deterministic ordering
+ * without threshold-based comparisons to avoid transitivity violations.
+ */
+public class BundleDataComparator implements Comparator<BundleData> {
+
+    /**
+     * Compares two bundle entries using BundleData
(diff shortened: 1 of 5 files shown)


lhotari:
@Meet0861 Pull requests should target the master branch. I changed the base, but that messed up things. You can rebase on your side and do a force-push. One way is to do this is:
```shell
git fetch origin # assuming "origin" is apache/pulsar
git reset --hard origin/master
git cherry-pick 8ce1fc4 # your PR commit
```
then push changes to the PR branch

lhotari:
@Meet0861 Thanks for the effort of debugging and drilling down to the root cause.
Since the compareTo contracts were violated in NamespaceBundleStats, TimeAverageMessageData and ResourceUnitRanking classes, I have addressed those in an alternative approach to address the issue, #24772. It wouldn't have been useful to keep the invalid implementations since they could have caused similar problems in the future when sorting would have been used. The main purpose of `compareTo` methods is sorting.
The original reason for the `Math.abs` was to compare with a coarse granularity (or resolution). I have addressed that without `Math.abs` in the PR so that the value is divided by the comparison resolution/granularity at comparison. The behavior isn't exactly the same as `Math.abs`, but it does retain the behavior of comparing at the desired level of granularity.
Since the `compareTo` methods are now fixed with #24772, it won't be necessary to have custom comparators that this PR provides.

PR #13043, opened 2021-11-30
Transparent batching of ZK operations

merlimat:
### Motivation

It is more efficient to do operations to ZK in batch, using the `ZooKeeper.multi()` operation instead of individual calls. 
This reduces the RPC traffic between ZK client and servers and it also reduces the number of write transactions, since each `multi()` correspond to a single ZK transaction, containing multiple write operations.

### Modifications

 * Added a base class `AbstractBatchedMetadataStore` which contains the batching logic
 * Added configurable batching parameters
 * Using 2 different transaction sets, one for read operations and the other just for write operations
 * Added code of (very) basic benchmark

### Preliminary results

This is done with the included basic benchmark, making request to the embedded ZK running on a laptop:

image

image

PR #24760, opened 2025-09-18
[improve][broker] PIP-444: Rate limit for deleting ledger to alleviate the zk pressure.

thetumbled:
PIP: #24771 

### Motivation

When the retention of a large topic is reduced, a significant number of ledgers need to be deleted. Since this deletion operation is not rate-limited, it results in ZooKeeper (ZK) latency of several minutes, as shown below:
image

### Modifications

Add rate limit feature for deleting ledgers.

### Verifying this change

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/75

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -2208,12 +2208,6 @@ public class ServiceConfiguration implements PulsarConfiguration {
     )
     private int managedLedgerDeleteMaxConcurrentRequests = 1000;
     @FieldContext(
-            category = CATEGORY_STORAGE_ML,
-            doc = "Number of threads to be used for deleting ledgers at broker level, "
-                    + "default is the number of CPU cores"
-    )
-    private int managedLedgerDeleteThreadPoolSize = Runtime.getRuntime().availableProcessors();
-    @FieldContext(
         category = CATEGORY_STORAGE_ML,
         dynamic = true,
         doc = "Allow automated creation of topics if set to true (default value)."
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -475,14 +475,11 @@ public class BrokerService implements Closeable {
 
         this.bundlesQuotas = new BundlesQuotas(pulsar);
         if (pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests() > 0) {
-            log.info("Setting managed ledger deletion max concurrent requests to {} and executor pool size to {}",
-                    pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests(),
-                    pulsar.getConfiguration().getManagedLedgerDeleteThreadPoolSize());
+            log.info("Setting managed ledger deletion max concurrent requests to {}",
+                    pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests());
             this.ledgerDeletionSemaphore = new Semaphore(
                     pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests());
-            this.ledgerDeletionExecutorProvider = new ExecutorProvider(
-                    pulsar.getConfiguration().getManagedLedgerDeleteThreadPoolSize(),
-                    "pulsar-ledger-deletion");
+            this.ledgerDeletionExecutorProvider = new ExecutorProvider(1, "pulsar-ledger-deletion");
         } else {
             this.ledgerDeletionSemaphore = null;
             this.ledgerDeletionExecutorProvider = null;


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -569,6 +576,7 @@ protected synchronized void initializeBookKeeper(final ManagedLedgerInitializeLe
             public void operationComplete(Void v, Stat stat) {
                 ledgersStat = stat;
                 emptyLedgersToBeDeleted.forEach(ledgerId -> {
+                    deleteLedgerRateLimiter.acquire();
```
merlimat: This is a blocking operation. We cannot block the thread in this callback.

lhotari: @thetumbled Since you have placed the rate limiter's acquire before each call to asyncDeleteLedger, I'd recommend extracting a method that removes the code duplication (includes the logic for rate limiting and deleting) and makes it easier to address the problem of blocking code, since that needs more changes. 

There would need to be a separate executor for triggering the asyncDeleteLedger jobs so that the BK callback thread isn't blocked.

Rate limiting is not necessarily alone the most useful way of handling the problem. 
It might be necessary to also limit concurrency. A semaphore would be needed for that purpose. There should be both a rate limit and semaphore for fully addressing the problem. One might want to disable rate limiting, but allow 100 concurrent deletions to happen at a time; or have a combination.

Instead of having **ledger** level rate limiting and concurrency limits, I'd recommend placing this in the ManagedLedgerFactoryImpl level so that it would be applied across the whole broker. I don't see how the limits could be reliable when they are at the ledger level.

thetumbled: Good point, i will refactor the implementation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -468,6 +472,19 @@ public BrokerService(PulsarService pulsar, EventLoopGroup eventLoopGroup) throws
...
 
         this.bundlesQuotas = new BundlesQuotas(pulsar);
+        if (pulsar.getConfiguration().getManagedLedgerDeleteRateLimit() > 0) {
+            log.info("Setting managed ledger deletion rate limit to {}, concurrency {}",
+                    pulsar.getConfiguration().getManagedLedgerDeleteRateLimit(),
+                    pulsar.getConfiguration().getManagedLedgerDeleteConcurrency());
+            this.ledgerDeletionRateLimiter = RateLimiter.create(
+                    pulsar.getConfiguration().getManagedLedgerDeleteRateLimit());
+            this.ledgerDeletionExecutorProvider = new ExecutorProvider(
+                    pulsar.getConfiguration().getManagedLedgerDeleteConcurrency(),
```
lhotari: This solution doesn't work for the concurrency. `bookKeeper.asyncDeleteLedger` is an asynchronous call and the purpose should be to limit how many deletions are "in flight".
A semaphore could handle this so that you acquire the permit before making the call and release it after the deletion has completed.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerConfig.java
```diff
@@ -61,6 +64,8 @@ public class ManagedLedgerConfig {
     private int metadataMaxEntriesPerLedger = 50000;
     private int ledgerRolloverTimeout = 4 * 3600;
     private double throttleMarkDelete = 0;
+    private RateLimiter ledgerDeletaRateLimiter;
```
lhotari: typo here

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
thetumbled: You are right, I have used the semaphore to limit the concurrency.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2058,6 +2077,8 @@ public CompletableFuture<ManagedLedgerConfig> getManagedLedgerConfig(@NonNull To
             managedLedgerConfig.setThrottleMarkDelete(persistencePolicies.getManagedLedgerMaxMarkDeleteRate() >= 0
                     ? persistencePolicies.getManagedLedgerMaxMarkDeleteRate()
                     : serviceConfig.getManagedLedgerDefaultMarkDeleteRateLimit());
+            managedLedgerConfig.setLedgerDeletionSemaphore(this.ledgerDeletionSemaphore);
+            managedLedgerConfig.setLedgerDeleteExecutor(this.ledgerDeletionExecutorProvider.getExecutor());
```
Denovo1998: When `getManagedLedgerDeleteMaxConcurrentRequests()` is 0 or negative, `ledgerDeletionExecutorProvider` is set to null in the constructor. However, there is no null check here when calling `this.ledgerDeletionExecutorProvider.getExecutor()`, which will cause a `NullPointerException`?
```java
managedLedgerConfig.setLedgerDeleteExecutor(this.ledgerDeletionExecutorProvider != null ?this.ledgerDeletionExecutorProvider.getExecutor() : null);
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -402,6 +408,10 @@ public ManagedLedgerImpl(ManagedLedgerFactoryImpl factory, BookKeeper bookKeeper
         this.minBacklogEntriesForCaching = config.getMinimumBacklogEntriesForCaching();
         this.maxBacklogBetweenCursorsForCaching = config.getMaxBacklogBetweenCursorsForCaching();
         this.managedLedgerAttributes = new ManagedLedgerAttributes(this);
+        if (config.getLedgerDeletionSemaphore() != null) {
+            this.deleteLedgerSemaphore = config.getLedgerDeletionSemaphore();
+            this.deleteLedgerExecutor = config.getLedgerDeleteExecutor();
```
Denovo1998, line 413: The initialization logic should validate that both the semaphore and executor are provided together?

Defensive programming:

```java
if (config.getLedgerDeletionSemaphore() != null) {
    this.deleteLedgerSemaphore = config.getLedgerDeletionSemaphore();
    this.deleteLedgerExecutor = config.getLedgerDeleteExecutor();
    if (this.deleteLedgerExecutor == null) {
        throw new IllegalArgumentException("LedgerDeleteExecutor must be provided when LedgerDeletionSemaphore is configured");
    }
}
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -3408,6 +3418,32 @@ private void asyncDeleteLedgerWithRetry(CompletableFuture<Void> future, long led
...
+                deleteLedgerSemaphore.release();
+                cb.deleteComplete(rc, ctx1);
+            };
+            deleteLedgerExecutor.execute(() -> {
+                try {
+                    deleteLedgerSemaphore.acquire();
+                } catch (InterruptedException e) {
+                    log.error("[{}] Interrupted while waiting to delete ledger {}", name, ledgerId, e);
+                }
+                bookKeeper.asyncDeleteLedger(ledgerId, cbWrapper, ctx);
```
Denovo1998: When semaphore acquisition is interrupted, the code logs the error but continues with the deletion. This could lead to exceeding the concurrency limit.

Looks a bit strange?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
thetumbled: fixed.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
thetumbled: This is a point that needs to be discussed: Should we proceed to send the delete request? @lhotari

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerConfig.java
```diff
@@ -23,7 +23,11 @@
 import java.time.Clock;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+
+import com.google.common.util.concurrent.RateLimiter;
```
BewareMyPower: seems it violates the checkstyle check?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
BewareMyPower: I suggest logging an error and exit, and there is no need to print the whole stack.

An `InterruptedException` is thrown when:
- the future returned by `Executor#submit` is called with `.cancel(true)` or the semaphore is called.
- or when the `acquire` returns, the thread has already exited (e.g. `executor.shutdown()` is called)

P.S. you can check the following code snippet to see the behavior:

```java
        final var executor = Executors.newSingleThreadExecutor();
        final var semaphore = new Semaphore(1);
        semaphore.acquire();
        final var future = executor.submit(() -> {
            try {
                semaphore.acquire();
                System.out.println("done");
                System.out.println("interrupted: " + Thread.currentThread().isInterrupted());
            }
        });
        Thread.sleep(300);
        future.cancel(true);
        executor.shutdown();
        semaphore.release();
        Thread.sleep(300);
        semaphore.release();
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -468,6 +474,19 @@ public BrokerService(PulsarService pulsar, EventLoopGroup eventLoopGroup) throws
...
         this.bundlesQuotas = new BundlesQuotas(pulsar);
+        if (pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests() > 0) {
+            log.info("Setting managed ledger deletion max concurrent requests to {} and executor pool size to {}",
+                    pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests(),
+                    pulsar.getConfiguration().getManagedLedgerDeleteThreadPoolSize());
+            this.ledgerDeletionSemaphore = new Semaphore(
+                    pulsar.getConfiguration().getManagedLedgerDeleteMaxConcurrentRequests());
+            this.ledgerDeletionExecutorProvider = new ExecutorProvider(
+                    pulsar.getConfiguration().getManagedLedgerDeleteThreadPoolSize(),
+                    "pulsar-ledger-deletion");
```
BewareMyPower: We should shutdown the `ExecutorProvider`

codelipenghui:
We do have batch operation to the metadata service. If the deletion option happened concurrently, it should be already batched in one request?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
thetumbled: fixed.

thetumbled:
> We do have batch operation to the metadata service. If the deletion option happened concurrently, it should be already batched in one request?

I am afraid that there is no batch feature for `bookKeeper.asyncDeleteLedger`. Do you know anything related to this? @lhotari @BewareMyPower

BewareMyPower:
I think @codelipenghui mentioned this feature: https://github.com/apache/pulsar/pull/13043, while `asyncDeleteLedger` involves a metadata store operation.

But from my perspective, the batching feature seems to be enabled by default, so I don't think it can resolve the issue mentioned in PIP-444

thetumbled:
> I think @codelipenghui mentioned this feature: #13043, while `asyncDeleteLedger` involves a metadata store operation.
> 
> But from my perspective, the batching feature seems to be enabled by default, so I don't think it can resolve the issue mentioned in PIP-444

Only when the zk request is issued by broker can this feature takes effect and batch them to alleviate the zk pressure. 

However, in this case, the zk requests are sent by the bookie client. Therefore, I'm afraid feature https://github.com/apache/pulsar/pull/13043 cannot help resolve this issue.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -821,6 +840,9 @@ public CompletableFuture<Void> closeAsync() {
         try {
             log.info("Shutting down Pulsar Broker service");
 
+            // shutdown executor for ledger deletion
+            ledgerDeletionExecutorProvider.shutdownNow();
```
nodece: Please add NPE check.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -2202,6 +2202,16 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    @FieldContext(
+            category = CATEGORY_STORAGE_ML,
+            doc = "Max number of concurrent requests for deleting ledgers at broker level"
+    )
+    private int managedLedgerDeleteMaxConcurrentRequests = 1000;
+    @FieldContext(
+            category = CATEGORY_STORAGE_ML,
+            doc = "Number of threads to be used for deleting ledgers at broker level"
+    )
+    private int managedLedgerDeleteThreadPoolSize = 5;
```
nodece: You should use the CPU cores.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
thetumbled: fixed.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -2202,6 +2202,17 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+            category = CATEGORY_STORAGE_ML,
+            doc = "Max number of concurrent requests for deleting ledgers at broker level"
+    )
+    private int managedLedgerDeleteMaxConcurrentRequests = 1000;
+    @FieldContext(
+            category = CATEGORY_STORAGE_ML,
+            doc = "Number of threads to be used for deleting ledgers at broker level, "
+                    + "default is the number of CPU cores"
+    )
+    private int managedLedgerDeleteThreadPoolSize = Runtime.getRuntime().availableProcessors();
```
BewareMyPower: @nodece I noticed your comment [here](https://github.com/apache/pulsar/pull/24760#discussion_r2387080113). However, using CPU cores as the default thread number is not always a silver bullet.

The only task executed by this executor is:

```java
                    deleteLedgerSemaphore.acquire();
                    bookKeeper.asyncDeleteLedger(ledgerId, cbWrapper, ctx);
```

`asyncDeleteLedger` is an asynchronous operation. Running asynchronous tasks in parallel is not a good practice. We only need a separated thread to avoid blocking the metadata store I/O thread by `acquire`. In this case, even if there are N threads, all of them will be blocked at `acquire`, which does not get the benefits from multi-threading as well.

thetumbled: Yes, we need an appropriate number of threads. How about setting it back to 5?

nodece: It looks like we should use a single thread.

BewareMyPower: Yes, single thread would be better, and I don't think we need to make it configurable. (question: in which cases will you increase the number of threads)

nodece: Built-in is better.

thetumbled: updated.

BewareMyPower:
Please update the original document #24771 as well, then you can start the vote

nodece (approved):

thetumbled:
> Please update the original document #24771 as well, then you can start the vote

Updated, and here is the link to the vote thread: https://lists.apache.org/thread/2dop6kyvrpt6ztz9pvylyl8o2syvh9kw. 
Thanks.

Issue #23897, opened 2025-01-26
[Bug] set-bookie-affinity-group overwrites "number of bundles" passed during namespace creation


programmerahul:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that unsupported versions don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### Version

OS: MacOs Sonoma 14.6.1 
Java: 17
Pulsar version: 4.0.2

### Minimal reproduce step

1.  pulsar-admin namespaces create t/n --bundles 10
2.  pulsar-admin namespaces set-bookie-affinity-group --primary-group temp t/n

### What did you expect to see?

The namespace "n" should have 10 bundles

### What did you see instead?

The namespace "n" have 4 bundles

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #24762, opened 2025-09-19
[fix][admin] set local policies overwrites "number of bundles" passed during namespace creation

oneby-wang:
Fixes [#xyz]([[Bug] set-bookie-affinity-group overwrites "number of bundles" passed during namespace creation · Is](https://github.com/apache/pulsar/issues/23897))

### Motivation

Set local policies will overwrite "number of bundles" passed during namespace creation. If we don't call any get namespace policies operation, local policies will not be created, so it will be overwritten.
[pulsar/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/NamespaceBundleFactory.java at 88](https://github.com/apache/pulsar/blob/8879b3ba9daa6de9b8b78bbce89bab68cc37cf0b/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/NamespaceBundleFactory.java#L164-L182)

The following steps can reproduce the problem.
1. pulsar-admin namespaces create t/n --bundles 10
2. pulsar-admin namespaces set-bookie-affinity-group --primary-group temp t/n
3. pulsar-admin namespaces bundles t/n

But the following step will not reproduce the problem.
1. pulsar-admin namespaces create t/n --bundles 10
2. pulsar-admin namespaces bundles t/n
3. pulsar-admin namespaces set-bookie-affinity-group --primary-group temp t/n
4. pulsar-admin namespaces bundles t/n

### Modifications

If  local policies is not present, first try to use namespace policies bundle data, then fallback to `config().getDefaultNumberOfNamespaceBundles()`

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -19,7 +19,6 @@
 package org.apache.pulsar.broker.admin.impl;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
-import static org.apache.pulsar.common.policies.data.PoliciesUtil.defaultBundle;
 import static org.apache.pulsar.common.policies.data.PoliciesUtil.getBundles;
 import com.google.common.collect.Sets;
 import java.lang.reflect.Field;
@@ -962,9 +961,7 @@ public abstract class NamespacesBase extends AdminResource {
                                 bookieAffinityGroup,
                                 policies.namespaceAntiAffinityGroup,
                                 policies.migrated))
-                        .orElseGet(() -> new LocalPolicies(getBundles(config().getDefaultNumberOfNamespaceBundles()),
-                                bookieAffinityGroup,
-                                null));
+                        .orElseGet(() -> new LocalPolicies(getDefaultBundleData(), bookieAffinityGroup, null));
                 log.info("[{}] Successfully updated local-policies configuration: namespace={}, map={}", clientAppId(),
                         namespaceName, localPolicies);
                 return localPolicies;
@@ -973,6 +970,8 @@ public abstract class NamespacesBase extends AdminResource {
             log.warn("[{}] Failed to update local-policy configuration for namespace {}: does not exist", clientAppId(),
                     namespaceName);
             throw new RestException(Status.NOT_FOUND, "Namespace does not exist");
+        } catch (RestException re) {
+            throw re;
         } catch (Exception e) {
             log.error("[{}] Failed to update local-policy configuration for namespace {}", clientAppId(), namespaceName,
                     e);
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -2907,5 +2908,16 @@ protected CompletableFuture<Set<String>> internalGetNamespaceAllowedClustersAsyn
                 .thenApply(policies -> policies.allowed_clusters);
     }
 
-
+    private BundlesData getDefaultBundleData() {
+        Optional<Policies> nsPolicies;
+        try {
+            nsPolicies = namespaceResources().getPolicies(namespaceName);
```
lhotari: This is a blocking method, please refactor the solution so that blocking methods aren't used

oneby-wang:
> From the PR description, it seems possible to add a test?

Thanks for advice, I'll add unit tests.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
oneby-wang: There are many synchronous calls in the admin v2 Namespaces. I had no choice but to use a synchronous call here because the upstream methods are using synchronous calls. I plan to submit several PRs to replace the synchronous operations with asynchronous ones after this PR in admin v2 Namespaces. WDYT?

lhotari: Oh yes, I see that now. I guess one way to prepare for this would be to already add a `getDefaultBundleDataAsync` method and then have the `getDefaultBundleData` delegate to the async method. That's a pattern that has been used in other locations.

oneby-wang: Nice suggestion, already applied.

oneby-wang:
ExtensibleLoadManagerTest.testAntiaffinityPolicy() unit test failure in fork repository PR.
https://github.com/oneby-wang/pulsar/actions/runs/17970886848/job/51114222121?pr=5

Seems this default bundle config affects this unit test. Previous: static default 1 numBundles. After: namespace policies numBundles.
https://github.com/apache/pulsar/blob/7c8d3c9210dcaf1d0c3b487af325b7b0a9dc3095/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java#L1792-L1793

PoliciesUtil.defaultBundle() method.
https://github.com/apache/pulsar/blob/7c8d3c9210dcaf1d0c3b487af325b7b0a9dc3095/pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/PoliciesUtil.java#L33-L41

Seems some bugs in ExtensibleLoadManager or ExtensibleLoadManagerTest? 
https://pulsar.apache.org/docs/3.0.x/administration-load-balance/#distribute-anti-affinity-namespaces-across-failure-domains

Not familiar with ExtensibleLoadManager, may give me some entry point, may be I can help fix this problem? Seems here?
https://github.com/apache/pulsar/blob/7c8d3c9210dcaf1d0c3b487af325b7b0a9dc3095/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java#L224-L237

oneby-wang:
Seems the anti-affinity works at the bundle level.
https://github.com/apache/pulsar/blob/7c8d3c9210dcaf1d0c3b487af325b7b0a9dc3095/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/LoadManagerShared.java#L588-L620

 I think this unit test may work in a wrong way due to this PR related issue, so I changed numBundles to 1 in createNamespace method.
https://github.com/apache/pulsar/blob/7c8d3c9210dcaf1d0c3b487af325b7b0a9dc3095/tests/integration/src/test/java/org/apache/pulsar/tests/integration/loadbalance/ExtensibleLoadManagerTest.java#L315-L321

PR #24766, opened 2025-09-19
[improve][pip] PIP-443: Stop using Netty Recycler in new code

BewareMyPower:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-443.md
+++ b/pip/pip-443.md
@@ -1,0 +1,236 @@
+# PIP-443: Stop using Netty Recycler in new code
+
+# Background knowledge
+
+Netty [`Recycler`](https://github.com/netty/netty/blob/netty-4.1.127.Final/common/src/main/java/io/netty/util/Recycler.java) is widely used in many places to reduce GC pressure by reusing objects. From the outputs of the following command:
+
+```bash
+find . -name "*.java" | grep src/main | xargs grep -n "new .*Recycler"
+```
+
+You can find:
+- 5 usages in `pulsar-common`
+- 10 usages in `pulsar-broker`
+- 10 usages in `pulsar-client`
+
+However, it's a pre-mature optimization without any real world evidence. Many of them are only the result of cargo cult programming. Hence, we should understand how it works first.
+
+Recycler is basically a thread-local object pool that provides
+- a `get` method to allocate an object from the pool or call the user-provided `newObject` method to create a new one if the pool is empty or full.
+- a `recycle` method to return an object back to the pool.
+
+Here is a typical implementation that wraps a `Recycler` to manage a triple:
+
+```java
+@Data // support getter and setter for all fields
+public class RecyclablePair {
+
+    private final Recycler.Handle<RecyclablePair> handle;
+    private Object x;
+    private Object y;
+
+    private RecyclablePair(Handle<RecyclablePair> handle) {
+        this.handle = handle;
+    }
+
+    public void recycle() {
+        this.x = null;
+        this.y = null;
+        handle.recycle(this);
+    }
+
+    public static RecyclablePair create(Object x, Object y) {
+        final RecyclablePair s = RECYCLER.get();
+        s.x = x;
+        s.y = y;
+        return s;
+    }
+
+    private static final Recycler<RecyclablePair> RECYCLER = new Recycler<RecyclablePair>() {
+        @Override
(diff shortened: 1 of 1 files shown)


pip/pip-443.md
```diff
@@ -0,0 +1,233 @@
...
+```bash
+find . -name "*.java" | grep src/main | xargs grep -n "new .*Recycler"
+```
+
+You can find:
+- 5 usages in `pulsar-common`
+- 10 usages in `pulsar-broker`
+- 10 usages in `pulsar-client`
+
+However, it's a pre-mature optimization without any real world evidence. Many of them are only the result of cargo cult programming. Hence, we should understand how it works first.
```
lhotari: It most likely was useful to some extent when it was introduced a long time ago. :) Java JVMs have developed a lot over the last 10+ years. Generational ZGC is pretty amazing and makes it unnecessary to use object recycling.
(Generational ZGC blog posts: [by Netflix](https://netflixtechblog.com/bending-pause-times-to-your-will-with-generational-zgc-256629c9386b), [by Gunnar Morling](https://www.morling.dev/blog/lower-java-tail-latencies-with-zgc/))

BewareMyPower: I mentioned the GC evolution a bit in `Issue 3` section. I can add this info to that section as well

pip/pip-443.md
```diff
@@ -0,0 +1,233 @@
...
+
+Now, if the recycler design on `OpReadEntry` saved 1 of 1 heap memory allocation for each read operation, now it just saves 1 of 2 heap memory allocations. What's more, this wrapped callback is passed to `PendingReadsManager` and eventually used as a field of a new record type `ReadEntriesCallbackWithContext` to an `ArrayList` in https://github.com/apache/pulsar/blob/d833b8bef21cb9e85f0f313eb9d49c7ca550fbbd/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/PendingReadsManager.java#L256
+
+Now, each read operation has:
+- 1 object allocated by Netty recycler (`OpReadEntry`)
+- 1 object allocated from heap memory (the wrapped callback in `RangeEntryCacheImpl`)
+- 1 object allocated from heap memory (the `ReadEntriesCallbackWithContext` object that has a field of the wrapped callback)
+  - This object is added to an `ArrayList` allocated from heap memory with at least 1 element
+
+Eventually, only saving the 1st heap memory allocation should not make a difference.
```
lhotari: btw. I guess allocations in the JVM aren't only about the heap allocations. If JVM's JIT/hotspot escape analysis determines an object doesn't escape its scope, the JVM can optimize it away through scalar replacement or stack allocation, eliminating the need for GC completely. Such optimizations in the JVM make it even more justified to get rid of the recycler.

pip/pip-443.md
```diff
@@ -0,0 +1,233 @@
...
+
+For existing code, generally we should not pay efforts to remove recyclable classes. But it's acceptable to remove them especially when:
+- A known recycler issue is reported but the reason is hard to analyze. (example: [#24741](https://github.com/apache/pulsar/pull/24741))
+- The existing recyclable object makes code refactoring hard
+
+# Alternatives
+
+## Add an option to allow whether to use Netty recycler
+
+This would be helpful to figure out if the GC pressure can be reduced once it's enabled. However, the code would be much more complicated.
```
lhotari: Tracking GC pressure isn't that easy since we currently have very misleading metrics. In many dash boards we mistakenly call the metric "GC Pause time", although, the metrics are usually about GC time where there's no actual "pause". When objects aren't recycled, it's expected for some use cases that more CPU will be spent in GC. However the Netty Recycler consumes a significant share of CPU too and there's no metric, so just looking at GC metrics increasing wouldn't be a way how to see if getting rid of Netty Recyclers is useful.

lhotari (approved):
LGTM 🚀

pip/pip-443.md
BewareMyPower: Yes, escape analysis should be here even with an older version of GC, especially for short-lived objects.

I've made some minor adjustments in the latest commit.

coderzc (approved):

PR #24772, opened 2025-09-22
[fix][misc] Fix compareTo contract violation for NamespaceBundleStats, TimeAverageMessageData and ResourceUnitRanking

### LINKED ISSUE #24754: [Bug] Failure in topK bundle selection due to the transitivity Contract Violation in namespaceBundleStats' & BundleData's compareTo()

lhotari:
Fixes #24754

### Motivation

- JVM's default sorting algorithm "TimSort" will fail with `java.lang.IllegalArgumentException: Comparison method violates its general contract!` if `compareTo` doesn't meet [the contract of `Comparable.compareTo` interface](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Comparable.html#compareTo(T))

The contract of `Comparable.compareTo`:

* The implementor must ensure [signum](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Integer.html#signum(int))(x.compareTo(y)) == -signum(y.compareTo(x)) for all x and y. (This implies that x.compareTo(y) must throw an exception if and only if y.compareTo(x) throws an exception.)

* The implementor must also ensure that the relation is transitive: (x.compareTo(y) > 0 && y.compareTo(z) > 0) implies x.compareTo(z) > 0.

* Finally, the implementor must ensure that x.compareTo(y)==0 implies that signum(x.compareTo(z)) == signum(y.compareTo(z)), for all z.

### Modifications

- address the compareTo violation by using "resolution" instead of "difference" when comparing values since it's not possible to fulfill the [`Comparable.compareTo` contract](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Comparable.html#compareTo(T)) by using a difference.
- add tests for `compareTo` for NamespaceBundleStats and TimeAverageMessageData
- Rename the ResourceUnitRanking's `compareTo` method to `compareToOtherRanking` and remove the `Comparable` interface since ResourceUnitRanking cannot fulfill the contract and there isn't a need to do so. Removing the interface and new comments in the `compareToOtherRanking` method will help prevent possible misuse later. The current usage of ResourceUnitRanking's method is fine, it's just not requiring Comparable.compareTo semantics and contract. That's why it's better to rename the method.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/SimpleLoadManagerImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/SimpleLoadManagerImpl.java
@@ -813,7 +813,7 @@ public class SimpleLoadManagerImpl implements LoadManager, Consumer<Notification
                     minLoadPercentage = loadPercentage;
                 } else {
                     if ((unboundedRanks ? ranking.compareMessageRateTo(selectedRanking)
-                            : ranking.compareTo(selectedRanking)) < 0) {
+                            : ranking.compareToOtherRanking(selectedRanking)) < 0) {
                         minLoadPercentage = loadPercentage;
                         selectedRU = candidate;
                         selectedRanking = ranking;
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/loadbalance/extensions/models/TopKBundlesTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/loadbalance/extensions/models/TopKBundlesTest.java
@@ -319,4 +319,29 @@ public class TopKBundlesTest {
             }
         }
     }
+
+    // Issue https://github.com/apache/pulsar/issues/24754
+    @Test
+    public void testPartitionSortCompareToContractViolationIssue() {
+        Random rnd = new Random(0);
+        ArrayList<NamespaceBundleStats> stats = new ArrayList<>();
+        for (int i = 0; i < 1000; ++i) {
+            NamespaceBundleStats s = new NamespaceBundleStats();
+            s.msgThroughputIn = 4 * 75000 * rnd.nextDouble();  // Just above threshold (1e5)
+            s.msgThroughputOut = 75000000 - (4 * (75000 * rnd.nextDouble()));
+            s.msgRateIn = 4 * 75 * rnd.nextDouble();
+            s.msgRateOut = 75000 - (4 * 75 * rnd.nextDouble());
+            s.topics = i;
+            s.consumerCount = i;
+            s.producerCount = 4 * rnd.nextInt(375);
+            s.cacheSize = 75000000 - (rnd.nextInt(4 * 75000));
+            stats.add(s);
+        }
+        List<Map.Entry<String, ? extends Comparable>> bundleEntries = new ArrayList<>();
+
+        for (NamespaceBundleStats s : stats) {
+            bundleEntries.add(Map.entry("bundle-" + s.msgThroughputIn, s));
+        }
+        TopKBundles.partitionSort(bundleEntries, 100);
+    }
 }
--- a/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer/NamespaceBundleStats.java
+++ b/pulsar-client-admin-api/src/main/java/org/apache/pulsar/policies/data/loadbalancer/NamespaceBundleStats.java
@@ -38,13 +38,13 @@ public class NamespaceBundleStats implements Comparable<NamespaceBundleStats>, S
     public long cacheSize;
 
     // Consider the throughput equal if difference is less than 100 KB/s
-    private static final double throughputDifferenceThreshold = 1e5;
+    private static final double throughputComparisonResolution = 1e5;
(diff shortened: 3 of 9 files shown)


merlimat (approved):

lhotari:
After merging, I noticed that NamespaceBundleStats uses a slightly different implementation for `compareDoubleWithResolution`. I didn't use `org.apache.pulsar.common.util.CompareUtil#compareDoubleWithResolution` in that case since the module doesn't currently depend on pulsar-common. 
The difference is that NamespaceBundleStats.compareDoubleWithResolution will use rounding to closest integer value `Long.compare(Math.round(v1 / resolution), Math.round(v2 / resolution))` and CompareUtil.compareDoubleWithResolution will use just `Long.compare((long) (v1 / resolution), (long) (v2 / resolution))`. In practice, this doesn't make a difference since both will compare with similar resolution/granularity and the compareTo contract is fulfilled in both cases.
This is a small inconsistency, but I don't see a need to address it.

PR #24780, opened 2025-09-25
[improve][broker] Replace isServiceUnitActiveAsync with checkTopicNsOwnership

BewareMyPower:
### Motivation

I'm working on optimizing the long topic loading path recently and I found these two methods just work the same. The ownership validation does not have a single source of truth, which makes code hard to read.

1. `BrokerService#checkTopicNsOwnership`

This method returns a **void** future that **fails** if the topic is not owned by the current broker. It's called by:
- `TransactionMetadataStoreService#handleTcClientConnect`
- `AbstractTopic#addProducer`
- `BrokerService#createNonPersistentTopic`
- `BrokerService#loadOrCreatePersistentTopic`
- `BrokerService#createPendingLoadTopic`
- `NonPersistentTopic#checkTopicNsOwnership`
- `PersistentTopic#internalSubscribe`

2. `NamespaceService#isServiceUnitActiveAsync`

This method returns a **boolean** future **that returns false** if the topic is not owned by the current broker. It's only used in `BrokerService#checkOwnershipAndCreatePersistentTopic`.
The implementations of these two methods are slightly different, which might also mislead users that they are different.

Here is how `isServiceUnitActiveAsync` handles topics' bundle: https://github.com/apache/pulsar/blob/1b74fe07bd24e079434769de2c52222761d88ca1/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java#L1249-L1253

However, the logic is duplicated with

https://github.com/apache/pulsar/blob/1b74fe07bd24e079434769de2c52222761d88ca1/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/OwnershipCache.java#L147-L153

https://github.com/apache/pulsar/blob/1b74fe07bd24e079434769de2c52222761d88ca1/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java#L1280-L1286

which is called by `checkTopicNsOwnership`: https://github.com/apache/pulsar/blob/1b74fe07bd24e079434769de2c52222761d88ca1/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L2376

The only place that calls `isServiceUnitActiveAsync` is here: https://github.com/apache/pulsar/blob/1b74fe07bd24e079434769de2c52222761d88ca1/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1739

when it returns false, it just completes another future exceptionally: https://github.com/apache/pulsar/blob/1b74fe07bd24e079434769de2c52222761d88ca1/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1761-L1764

where the error message does not have bundle info included like `checkTopicNsOwnership`.

### Modifications

Use `checkTopicNsOwnership` instead of `isServiceUnitActiveAsync` in `BrokerService#checkOwnershipAndCreatePersistentTopic`.

Remove the `isServiceUnitActiveAsync` and `isServiceUnitActive` methods from `NamespaceService`. The same function can be implemented easily by composing existing methods like:

```java
    public static CompletableFuture<Boolean> isServiceUnitActiveAsync(NamespaceService namespaceService,
                                                                      TopicName topicName) {
        return namespaceService.getBundleAsync(topicName).thenCompose(bundle ->
                namespaceService.checkBundleOwnership(topicName, bundle));
    }
```

### Verifying this change

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
@@ -43,9 +43,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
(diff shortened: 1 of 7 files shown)


lhotari (approved):
LGTM, thanks for taking actions to cleanup Pulsar code!

Issue #10473, opened 2021-05-03
PulsarStandaloneBuilder: make it a public API


eolivelli:
PulsarStandaloneBuilder looks like a great API for users to launch an embedded Pulsar service for unit tests.

Currently (2.8.0) there are a few problems that need to be fixed:
- Provide convenience methods to build a Pulsar Client and a Pulsar Admin
- Provide a way to start the server without creating a real file on disk
- Add tests that ensure that PulsarStandaloneBuilder can be called by a Java program, not only by the means of the scripts bundled with Pulsar

codelipenghui:
The issue had no activity for 30 days, mark with Stale label.

MananJethwani:
@eolivelli is this issue still open, would like to contribute

eolivelli:
I think so.

I suggest you to try it and fix it in a way that you can start Pulsar in a JVM. The main goal is to use it to run unit tests of applications based on Pulsar that cannot use docker/testcontainers

PavelZeger:
Is the linked PR still relevant?

PR #24781, opened 2025-09-25
Make PulsarStandaloneBuilder a public API (#10473)

leocavalhere6:
## Description
- Added convenience methods to build `PulsarClient` and `PulsarAdmin`.
- Enabled starting the server without creating real files on disk.
- Added unit tests to ensure usage via Java program.

**Fixes:** #10473

## Motivation
Make `PulsarStandaloneBuilder` usable as a public API for unit tests, without requiring Docker/Testcontainers.

## Modifications
- Added `buildClient()` and `buildAdmin()` methods.
- Create temporary directories for ZK and BK to avoid writing on disk.
- Added `PulsarStandaloneBuilderTest.java` unit test.

## Verifying this change
- Ran `mvn test` to ensure all unit tests pass, including `PulsarStandaloneBuilderTest`.

- [x] The public API

## Documentation
- [x] `doc-not-needed`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/PulsarStandaloneBuilder.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/PulsarStandaloneBuilder.java
@@ -21,6 +21,10 @@ package org.apache.pulsar;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import org.apache.pulsar.broker.ServiceConfiguration;
 import org.apache.pulsar.broker.ServiceConfigurationUtils;
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.admin.PulsarAdmin;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 public final class PulsarStandaloneBuilder {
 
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/PulsarStandaloneBuilder.java
```diff
@@ -96,6 +100,31 @@ public PulsarStandaloneBuilder withAdvertisedAddress(String advertisedAddress) {
...
+            .build();
+    }
+
+    private PulsarStandaloneBuilder() throws IOException {
+        pulsarStandalone = new PulsarStandalone();
+        pulsarStandalone.setWipeData(true);
+        pulsarStandalone.setNoFunctionsWorker(true);
+
+        Path tempZkDir = Files.createTempDirectory("zk");
+        Path tempBkDir = Files.createTempDirectory("bk");
```
lhotari, line 121: These files would be left around in the temp directories. `wipeData` currently wipes data only at startup time.

lhotari: These directories would also be created when it's not desired.

PR #24787, opened 2025-09-25
[improve][broker] Add tests for using absolute FQDN for advertisedAddress and remove extra dot from brokerId

lhotari:
### Motivation

When brokers are exposed directly to clients, there might be a need to make the `advertisedAddress` absolute so that clients with unoptimal ndots configuration won't do unnecessary DNS lookups on the client side when an `advertisedAddress` with less dots than ndots is returned to the client from topic lookup.

### Modifications

- remove extra dot from brokerId so that the brokerId format doesn't change when advertisedAddress is made absolute
- add tests for brokerId
- add integration tests to ensure that client can connect when TLS is enabled. This ensures that TLS hostname validation doesn't fail with the extra trailing dot.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


nodece (approved):
LGTM

dao-jun (approved):

PR #24794, opened 2025-09-27
[fix][client] Fix thread leak in reloadLookUp method which is used by ServiceUrlProvider

lhotari:
### Motivation

LookupService is replaced by calling reloadLookUp. However the previous instance is not closed. This causes a thread leak.

### Modifications

Call the close method before replacing LookupService.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
@@ -1206,6 +1206,13 @@ public class PulsarClientImpl implements PulsarClient {
     }
 
     public void reloadLookUp() throws PulsarClientException {
+        if (lookup != null) {
+            try {
+                lookup.close();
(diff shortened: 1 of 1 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
```diff
@@ -1206,6 +1206,13 @@ public LookupService getLookup() {
     }
 
     public void reloadLookUp() throws PulsarClientException {
+        if (lookup != null) {
```
nodece: @lhotari The `lookup` variable is not declared as `volatile`, which may lead to a visibility issue.

lhotari: Addressed

PR #24799, opened 2025-09-29
[improve][broker] Part-2 of PIP-434: Use ServerCnxThrottleTracker, instead of modifying channel.readable directly

poorbarcode:
### Motivation

See the PIP-434
- https://github.com/apache/pulsar/pull/24510

### Modifications
- Use `ServerCnxThrottleTracker`, instead of modifying `channel.readable` directly
- Changes of `ServerCnxThrottleTracker`
  - Merge multiple tracker states into one.
  - Split the counter for each throttling type, to avoid them affecting each other, and make troubleshooting easier.
  - Instead of modifying per tracker state with CAS, let all changes in the same thread(`pulsar-io-thread`)
  - Add a new test to confirm that the limited state will be resumed after a period of the rate limiter, even though the producer was closed

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -501,7 +501,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
     public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
         if (pauseReceivingRequestsIfUnwritable && ctx.channel().isWritable()) {
             log.info("[{}] is writable, turn on channel auto-read", this);
-            getThrottleTracker().markThrottled(ThrottleType.ConnectionOutboundBufferFull);
+            getThrottleTracker().unmarkThrottled(ThrottleType.ConnectionOutboundBufferFull);
             requestRateLimiter.timingOpen(pauseReceivingCooldownMilliSeconds, TimeUnit.MILLISECONDS);
         } else if (pauseReceivingRequestsIfUnwritable && !ctx.channel().isWritable()) {
             final ChannelOutboundBuffer outboundBuffer = ctx.channel().unsafe().outboundBuffer();
@@ -515,8 +515,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                     PAUSE_RECEIVING_LOG.debug("[{}] is not writable, turn off channel auto-read", this);
                 }
             }
-            getThrottleTracker().unmarkThrottled(ThrottleType.ConnectionOutboundBufferFull);
-            ctx.channel().config().setAutoRead(false);
+            getThrottleTracker().markThrottled(ThrottleType.ConnectionOutboundBufferFull);
         }
         ctx.fireChannelWritabilityChanged();
     }


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
```diff
@@ -18,125 +18,439 @@
...
+     * <p>Metrics are automatically recorded to track throttling events and connection state changes.
+     *
+     * @param type the type of throttling condition to activate
+     * @throws IllegalArgumentException if type is null
+     *
+     * @see #unmarkThrottled(ThrottleType)
+     * @see ThrottleType
+     */
+    public void markThrottled(ThrottleType type) {
+        ThrottleRes res = doMarkThrottled(type);
```
lhotari, line 306: At the beginning of the method, could we add a check here to ensure that this is executed within the event loop of the connection?

something like
```
assert serverCnx.ctx().executor().inEventLoop() : "This method should be called in serverCnx.ctx().executor()";
```

Another possibility would be to always accept the call and if it's not in the eventloop, then run the call to the event loop.
```
if (!serverCnx.ctx().executor().inEventLoop()) {
   serverCnx.ctx().executor().execute(() -> markThrottled(type));
}
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/resourcegroup/ResourceGroupPublishLimiter.java
```diff
@@ -30,7 +31,13 @@ public class ResourceGroupPublishLimiter extends PublishRateLimiterImpl  {
     private volatile long publishMaxByteRate;
 
     public ResourceGroupPublishLimiter(ResourceGroup resourceGroup, MonotonicClock monotonicClock) {
-        super(monotonicClock);
+        super(monotonicClock, producer -> {
+            producer.getCnx().getThrottleTracker().markThrottled(
+                ServerCnxThrottleTracker.ThrottleType.BrokerPublishRate);
+        }, producer -> {
+            producer.getCnx().getThrottleTracker().unmarkThrottled(
+                ServerCnxThrottleTracker.ThrottleType.BrokerPublishRate);
```
lhotari: should use ResourceGroupPublishRate here? 

just wondering if the ResourceGroup solution allows multiple ResourceGroupPublishLimiters being active at once. In that case a single ResourceGroupPublishRate wouldn't be sufficient.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
```diff
@@ -18,125 +18,439 @@
...
+     * <p>Metrics are automatically recorded to track unthrottling events and connection state changes.
+     *
+     * @param type the type of throttling condition to deactivate
+     * @throws IllegalArgumentException if type is null
+     *
+     * @see #markThrottled(ThrottleType)
+     * @see ThrottleType
+     */
+    public void unmarkThrottled(ThrottleType type) {
+        ThrottleRes res = doUnmarkThrottled(type);
```
lhotari, line 338: At the beginning of the method, could we add a check here to ensure that this is executed within the event loop of the connection?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
```diff
@@ -18,125 +18,439 @@
...
+        switch (throttleType) {
+            case TopicPublishRate: {}
+            case ResourceGroupPublishRate: {
+                states[throttleType.ordinal()] = value + 1;
+                return throttled ? ThrottleRes.TypeStateChanged : ThrottleRes.ConnectionStateChanged;
+            }
+            default: {
+                states[throttleType.ordinal()] = 1;
+                if (value != 0) {
+                    return ThrottleRes.Dropped;
```
lhotari: it would be useful to log this case just so that the ThrottleTypes aren't misused later. Let's say if there would be a case in the future where a since ThrottleType has multiple instances behind it. That should be handled by adding a counter for ThrottleType. I guess we don't have that currently unless the ResourceGroupPublishType turns out to be one.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
```diff
@@ -18,125 +18,439 @@
...
+        int value = states[throttleType.ordinal()];
+        switch (throttleType) {
+            case TopicPublishRate: {}
+            case ResourceGroupPublishRate: {
+                states[throttleType.ordinal()] = value - 1;
+                return hasThrottled() ? ThrottleRes.TypeStateChanged : ThrottleRes.ConnectionStateChanged;
+            }
+            default: {
+                if (value != 1) {
+                    return ThrottleRes.Dropped;
```
lhotari: log this case

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
poorbarcode: Sure, added the check for both `markThrottled` and `unmarkThrottled`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
poorbarcode: Added the check, see also https://github.com/apache/pulsar/pull/24799#discussion_r2415448306

pulsar-broker/src/main/java/org/apache/pulsar/broker/resourcegroup/ResourceGroupPublishLimiter.java
poorbarcode: > should use ResourceGroupPublishRate here?

Thanks for mentioning this mistake. I have changed it

> just wondering if the ResourceGroup solution allows multiple ResourceGroupPublishLimiters being active at once. In that case a single ResourceGroupPublishRate wouldn't be sufficient.

Since the `markThrottled` is reentrant if the type is `ResourceGroupPublishRate`, it should work. Could you explain more details?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
poorbarcode: > it would be useful to log this case just so that the ThrottleTypes aren't misused later.

I added a boolean field named `reentrant` for the `ThrottleType`, to manually define a value for the new field is needed when a new type is added.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
poorbarcode: Improved. See also https://github.com/apache/pulsar/pull/24799#discussion_r2415469581

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
```diff
@@ -18,125 +18,442 @@
...
+         * Throttling due to excessive pending publish requests on the connection.
+         *
+         * <p>This throttling is activated when the number of in-flight publish requests
+         * exceeds the configured limit. It helps prevent memory exhaustion and ensures
+         * fair resource allocation across connections.
+         *
+         * <p><b>Type:</b> Non-reentrant
+         * <p><b>Configuration:</b> {@link ServiceConfiguration#getMaxPendingPublishRequestsPerConnection()}
+         */
+        ConnectionMaxQuantityOfInFlightPublishing(false),
```
lhotari: the name `ConnectionMaxQuantityOfInFlightPublishing` is not great. It's better to align more closely with `maxPendingPublishRequestsPerConnection` configuration setting.
For example, `ConnectionMaxPendingPublishRequestsExceeded` (could also omit `Exceeded` suffix).

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
```diff
@@ -18,125 +18,442 @@
...
+         * Throttling due to excessive memory usage by in-flight publish operations on the IO thread.
+         *
+         * <p>This throttling is activated when the total memory used by pending publish operations
+         * on a shared IO thread exceeds the configured limit. Multiple connections may share the
+         * same IO thread, so this limit applies across all connections on that thread.
+         *
+         * <p><b>Type:</b> Non-reentrant
+         * <p><b>Configuration:</b> {@link ServiceConfiguration#getMaxMessagePublishBufferSizeInMB()}
+         */
+        IOThreadMaxBytesOfInFlightPublishing(false),
```
lhotari: The name `IOThreadMaxBytesOfInFlightPublishing` isn't great. It's better to more closely aligned with the configuration setting `maxMessagePublishBufferSizeInMB`. 
For example `IOThreadMaxPendingPublishBytesExceeded`. This would also be aligned with the possible new name `ConnectionMaxPendingPublishRequestsExceeded`.
(could also omit `Exceeded` suffixes)

pulsar-broker/src/main/java/org/apache/pulsar/broker/resourcegroup/ResourceGroupPublishLimiter.java
lhotari: This looks fine now.

poorbarcode:
Rebase master

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnxThrottleTracker.java
BewareMyPower: This assertion breaks the protocol handler feature in https://github.com/streamnative/kop/pull/1589

The root cause is that the public `incrementPublishCount` method is called directly by the protocol handler but it's not guaranteed to be called in the `ServerCnx`'s event loop thread. In KoP, the `ServerCnx` object is just a mock object that does not serve any connection.

Currently, I made a workaround like:

```java
            // TODO: Here we cannot leverage the `PersistentTopic#incrementPublishCount` method because it could call
            //   `markThrottled` and `unmarkThrottled` from `ServerCnxThrottleTracker`, which has a strong assertion on
            //   the thread that calls this method. See https://github.com/apache/pulsar/pull/24799 for details.
            try {
                final var topic = (AbstractTopic) producer.getTopic();
                final var bytesInCounter = AbstractTopic.class.getDeclaredField("bytesInCounter");
                bytesInCounter.setAccessible(true);
                ((LongAdder) bytesInCounter.get(topic)).add(readableBytes);
                final var msgInCounter = AbstractTopic.class.getDeclaredField("msgInCounter");
                msgInCounter.setAccessible(true);
                ((LongAdder) msgInCounter.get(topic)).add(numMessages);
            } catch (NoSuchFieldException | ClassCastException | IllegalAccessException e) {
                if (errorLoggedForReflection.compareAndSet(false, true)) {
                    log.error("Unable to update bytes and msg for topic", e);
                }
            }
```

Though the publish rate limiter will not be supported by the downstream protocol handler. /cc @tjiuming @Demogorgon314

BewareMyPower: For now, it's okay to keep this change. Because if calling `setAutoRead` out of the Netty thread does not work, the previous publish rate limiting feature in KoP might actually never work as expected.

PR #22064, opened 2024-02-17
[fix][broker] Set ServiceUnitStateChannel topic compaction threshold explicitly, improve getOwnerAsync, and fix other bugs

heesung-sn:
### Motivation

We better set the compaction threshold of the serviceUnitStateChannel topic instead of relying on other components' init.

Also, we need to fix the issue that lookup and unload often timeout with  500 error.
### Modifications

- Set the compaction threshold of the serviceUnitStateChannel topic in the monitor thread
- Release metadataAndPayload buffer in RawBatchMessageContainerImpl.
- Automatically restart load data store producer and table view if inactive.
- Updated getOwnerAsync for better synchronization of ownership checks, deferred lookups and active broker checks
- added minor retry logic in the ExtensibleLoadManager tests

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/heesung-sn/pulsar/pull/61

PR #24809, opened 2025-10-02
[feat][client] PIP-445: Add Builder Methods to Create Message-based TableView

### LINKED ISSUE #24744: API to allow access to raw message for TableView
### Search before reporting

### Motivation

We use message properties to enable client side message filtering.

However this means we cannot use table view, as it only provides access to the de-serialized message.

### Solution

An API to allow access to raw message object in combination with table view

### Alternatives

Currently using a reader to find the last message ID, then re-reading from the beginning until that message ID.

### Anything else?

_No response_

namest504:
Fixes: #24744

Main Issue: #24744

PIP: 445 #24842

### Motivation

The current `TableView` API only exposes the deserialized message value, which limits access to essential message metadata like properties, event time, or the raw message object itself.

This PR introduces a flexible, non-breaking mechanism to create value-mapped views over a topic. It replaces the initial proposal of adding a `getRawMessage()` method, which would have performance implications for all users. Instead, it adds `createMapped` methods to the `TableViewBuilder`, allowing users to transform a full `Message<T>` into any custom object `V` that suits their needs. This provides maximum flexibility, from accessing the full raw message to creating custom, memory-efficient objects.

Since arbitrary user code now runs inside the table view's message handling, the PR also defines what happens when that code fails. Previously an exception thrown while extracting a message's value escaped into the reader loop and was treated as a reader failure: during the initial replay it failed table view creation and leaked the reader, and while tailing the message was silently dropped with a log message that blamed the reader.

### Modifications

- **Added `TableViewMessageMapper<T, V>` to the public client API**: a functional interface with `V map(Message<T>)`, which may throw, and a `default boolean onMappingError(Message<T>, Throwable)` callback. The default callback returns `false`, which makes the table view log the failure at `ERROR` level with the topic, key and message id; an implementation can override it to report the failure its own way (metrics, signalling another thread, ...) and return `true` to suppress the log. Further callbacks can be added later as `default` methods without breaking implementations.
- **Added `createMapped` and `createMappedAsync` to `TableViewBuilder`**: they take a `TableViewMessageMapper<T, V>`. Lambdas work directly, e.g. `createMapped(msg -> msg.getProperty("region"))`, and `createMapped(msg -> msg)` builds a `TableView<Message<T>>` exposing the complete messages. A `null` mapper is rejected with `IllegalArgumentException` (failed future in the async variant).
- **Defined the mapping contract on the public API**: a keyed message with an empty payload is a tombstone that removes the key without calling the mapper; a `null` return value from the mapper is also a tombstone; if the mapper throws, the message is skipped — the key keeps its previous value, listeners are not notified, a pending `refresh()` still completes — and `onMappingError` is called. Exceptions thrown by the callback are logged and ignored.
- **Rejected `topicCompactionStrategyClassName` for mapped views** in both builder methods and, defensively, in the `MappedTableViewImpl` constructor before the reader is created: a `TopicCompactionStrategy` compares values of the topic's schema type, which a mapped view does not store, so the combination would fail with a `ClassCastException` on the reader thread.
- **Moved the core logic to `AbstractTableViewImpl<T, V>`**: `TableViewImpl` keeps its name and becomes a thin subclass, so the behavior for existing users (including message pooling and release) is unchanged, with one exception: a message whose lazily decoded value throws is now skipped and logged consistently in both the initial replay and the tailing phase, instead of failing `create()` in the former and being dropped silently in the latter. This also applies to the broker's own `create()` user, the extensible load balancer's `ServiceUnitStateTableViewImpl`: an undecodable `ServiceUnitStateData` record no longer prevents the channel from starting; it is skipped and logged at `ERROR`, as it already was while tailing. The trade-off is documented in the PIP-445 compatibility section.
- **Introduced package-private `MappedTableViewImpl<T, V>`**: implements the mapped views and delegates mapping failures to the mapper's `onMappingError`. Message pooling is disabled for mapped views since the mapper may retain the `Message` instance; the `createMapped` javadoc notes that a retained `Message` also keeps its metadata, schema and connection reference, so a value object is preferable to `msg -> msg` for topics with many keys.
- **Closed the reader when `start()` fails**: the reader created for the table view is no longer leaked when the initial replay fails. A failed `hasMessageAvailableAsync()` during the replay now fails `start()` too, instead of leaving `create()` waiting forever with the reader open.

### Verifying this change

- Unit tests in `TableViewBuilderImplTest`: `testCreateMapped*` (mapped view creation, `null` mapper validation, and rejection of a configured compaction strategy for both the sync and async variants).
- New unit tests in `MappedTableViewImplTest` with a mocked reader: a mapping failure skips the message and keeps the previous value, notifies `onMappingError`, does not notify listeners and does not block `refreshAsync()` (including a refresh issued from the callback); the default callback and a throwing callback both leave the view working; a mapping failure during the initial replay no longer fails `start()`; the reader is closed when the initial replay fails; a failed `hasMessageAvailableAsync()` fails `start()` instead of hanging it; and the constructor rejects a configured compaction strategy without creating a reader.
- Integration tests in `TableViewTest`: `testCreateMapped` (custom mapping, value updates, tombstone via `null` mapping), `testCreateMappedWithIdentityMapper` (`TableView<Message<T>>` with metadata access) and `testCreateMappedWithFailingMapper` (a mapper that fails during the initial replay and while tailing; the key keeps its previous value and later messages are still applied).
- Verified locally with `./gradlew quickCheck` and the test classes above.

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: [fork-repo](https://github.com/namest504/pulsar)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
```diff
@@ -18,40 +18,29 @@
...
+
+import static org.apache.pulsar.common.topics.TopicCompactionStrategy.TABLE_VIEW_TAG;
 
 @Slf4j
 public class TableViewImpl<T> implements TableView<T> {
 
     private final TableViewConfigurationData conf;
 
     private final ConcurrentMap<String, T> data;
+    private final ConcurrentMap<String, Message<?>> rawMessages;
```
lhotari: this would change the run time behavior of TableView and use more memory and CPU.

lhotari (changes requested):
Thanks for the contribution!

Since this changes the Pulsar API, we'd need to first go through the [PIP process](https://github.com/apache/pulsar/tree/master/pip#pulsar-improvement-proposal-pip) to modify the public API.

Instead of adding a separate ConcurrentMap for storing the messages, I'd suggest going with an approach where the TableViewBuilder would have separate methods for creating a view for messages. 

```java
TableView<Message<T>> createForMessages() throws PulsarClientException;
CompletableFuture<TableView<Message<T>>> createForMessagesAsync() throws PulsarClientException;
```

I'm not exactly sure if there's any obstacles with this approach, but it seems that it could be a better way forward so that the existing runtime behavior of TableView implementation wouldn't change (like it does with the current PR changes).

namest504:
Hi @lhotari,

Thanks for the review and the great suggestion about using the `TableViewBuilder`! 

I completely agree that we should avoid impacting the runtime behavior for existing users.

As you recommended, I've created a PIP for this API change, which you can find here: #24842

Let's continue the design discussion over on the PIP PR.

lhotari:
> Hi @lhotari,
> 
> Thanks for the review and the great suggestion about using the `TableViewBuilder`!
> 
> I completely agree that we should avoid impacting the runtime behavior for existing users.
> 
> As you recommended, I've created a PIP for this API change, which you can find here: #24842
> 
> Let's continue the design discussion over on the PIP PR.

@namest504 you seemed to ignore the suggestion in the comment. I'd suggest revisiting the PIP accordingly and renaming it. You can reply to the comment on the PIP PR, https://github.com/apache/pulsar/pull/24842#pullrequestreview-3335117069.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewImpl.java
namest504: I agree it. 😄

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/TableViewTest.java
```diff
@@ -53,7 +53,6 @@
 import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.client.api.Reader;
 import org.apache.pulsar.client.api.Schema;
-import org.apache.pulsar.client.api.TableView;
```
lhotari, line 56: removing this import adds unnecessary changes to the diff. It's better to keep the original way there the code doesn't have to reference the full class name.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java
```diff
@@ -22,14 +22,10 @@
...
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.pulsar.client.api.ConsumerCryptoFailureAction;
-import org.apache.pulsar.client.api.CryptoKeyReader;
-import org.apache.pulsar.client.api.PulsarClientException;
-import org.apache.pulsar.client.api.Schema;
-import org.apache.pulsar.client.api.TableView;
```
lhotari, line 31: don't remove this import so that the code doesn't have to reference the full class name.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/TableViewImplTest.java
```diff
@@ -24,7 +24,6 @@
 import static org.testng.Assert.assertNotNull;
 import org.apache.pulsar.client.api.CryptoKeyReader;
 import org.apache.pulsar.client.api.Schema;
-import org.apache.pulsar.client.api.TableView;
```
lhotari, line 46: don't remove this import so that the code doesn't have to reference the full class name.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/TableViewTest.java
namest504: Fixed. The import removal was caused by naming the impl class `TableView`, which shadowed the API interface in the same package. I went back to the `TableViewImpl` / `AbstractTableViewImpl` / `MessageMapperTableViewImpl` naming from your experiment branch, so the imports and the related diff churn are gone. The remaining changes in this file are the reflection target (`readAllExistingMessages` now lives in `AbstractTableViewImpl`) and the two new tests.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java
namest504: Fixed, same root cause as above. The explicit import list is restored and the wildcard import is gone.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/TableViewImplTest.java
namest504: Fixed. This file now has no changes compared to master.

namest504:
@lhotari I've addressed the review comments and resolved the conflicts with master (rebuilt the refactoring on top of the current Gradle/slog codebase).

Main changes since your last review:
- Kept the original `TableViewImpl` class name and moved the shared logic to a package-private `AbstractTableViewImpl<T, V>`, following the naming in your experiment branch. No imports are removed and there are no fully qualified class name references anymore.
- Removed the unused `MessageTableView` class left over from the earlier `createForMessages` approach.
- Disabled message pooling for mapped views: the mapper may retain the `Message` instance, so the pooled payload buffer must not be reused or released. The classic `TableView` path keeps pooling and releasing as before.
- Added mapper null-validation and unit tests in `TableViewBuilderImplTest`.

The vote result for PIP-445 has been posted on the dev mailing list (3 binding +1s), and I've also updated the PIP document in #24842 to match the final class names. PTAL when you have a chance.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        // The message instance is passed to the user-provided mapper function, which may keep a reference
+        // to it (e.g. when Function.identity() is used as the mapper). Pooled messages must not be used
+        // since there is no way to know when the message could be released.
+        super(client, schema, conf, false);
+        this.mapper = mapper;
+    }
+
+    @Override
+    protected V getValue(Message<T> msg) {
+        return mapper.apply(msg);
```
lhotari: **[BUG] a throwing mapper is misdiagnosed as a reader failure: the tail update is silently lost, and during replay it fails createMapped() and leaks the reader**

`mapper.apply(msg)` runs unguarded — `handleMessage` has no `catch` around the value extraction ([`AbstractTableViewImpl.java:228`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L228)); only per-listener exceptions are isolated ([`AbstractTableViewImpl.java:259`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L259)). So a user mapper that throws is handled as though the *reader* had failed:

- **Tail path** ([`AbstractTableViewImpl.java:437-464`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L437-L464)): the throw propagates out of `thenAccept` into `exceptionally` at [`:442`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java#L442). It is not an `AlreadyClosedException`, so the else branch sleeps 50 ms **on the client callback thread** ([`:454`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java#L454)), logs `"Reader was interrupted while reading tail messages. Retrying.."` ([`:458`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java#L458)), and calls `readTailMessages(reader)` — which reads the *next* message. `readNextAsync()` has already acknowledged the failed one, and nothing seeks backward (reconnect, `hasMessageAvailable` and refresh all move forward only), so that key's update is lost permanently and the log blames the reader rather than the mapper.
- **Replay path** ([`AbstractTableViewImpl.java:397-423`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L397-L423)): the same throw reaches [`AbstractTableViewImpl.java:412`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L412) and does `future.completeExceptionally(ex)` at [`AbstractTableViewImpl.java:421`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L421). `start()` ([`AbstractTableViewImpl.java:134-143`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L134-L143)) has no `exceptionally`/`whenComplete`, so the `Reader` created upstream is never closed — `createMapped()` fails *and* leaks the reader.

**Failure scenario:** a mapper such as `m -> m.getProperty("x").length()` NPEs on one message that lacks the property. While tailing, that key silently diverges from the topic for the life of the table view. The same message during initial replay instead throws out of `createMapped()` and leaves an orphaned reader behind.

To be fair: on master this handler already caught lazy `Message.getValue()` schema-decode failures, so the shape is not new. What is new is that arbitrary user code now flows into it, and neither `TableViewBuilder.createMapped` nor the abstract `getValue` javadoc says the mapper must not throw.

Could the mapper call be wrapped so a failure is attributable and does not masquerade as a reader error — and could the javadoc state the contract either way? Whichever you choose, closing the reader when `start()` fails looks worth doing regardless.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java
```diff
@@ -0,0 +1,466 @@
...
+                String key = msg.getKey();
+                V cur = getValueIfPresent(msg);
+                    log.debug().attr("key", key)
+                            .attr("value", cur)
+                            .log("Applying message");
+
+                boolean update = true;
+                if (compactionStrategy != null) {
+                    V prev = data.get(key);
+                    update = !compactionStrategy.shouldKeepLeft(prev, cur);
```
lhotari: **[BUG] a configured TopicCompactionStrategy is fed mapper outputs, not payload values**

`compactionStrategy` is declared `TopicCompactionStrategy<V>` ([`:72`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L72)) but loaded reflectively by class name ([`:109-110`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L109-L110)), so `<V>` is an unchecked claim rather than something the compiler or the loader verifies. On the classic path `V == T`, so a strategy written for the payload type receives payload values and all is well. On a mapped view, `V` is the mapper's output while the configured strategy class is still written against `T` — its synthetic bridge method casts to `T` and throws `ClassCastException`.

The combination is reachable through public API: `topicCompactionStrategyClassName` has no builder setter, but `TableViewBuilder.loadConf(Map<String, Object>)` is public ([`TableViewBuilder.java:58`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TableViewBuilder.java#L58)), and `createMapped` / `createMappedAsync` ([`TableViewBuilderImpl.java:72-86`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java#L72-L86)) validate only that the mapper is non-null before passing the same `conf` straight through. Nothing rejects the pair.

**Failure scenario:**

```java
client.newTableViewBuilder(schema)
      .topic(topic)
      .loadConf(Map.of("topicCompactionStrategyClassName", MyStrategy.class.getName()))
      .createMapped(msg -> msg.getKey());   // V = String; MyStrategy expects the payload type
```

The `ClassCastException` lands inside `shouldKeepLeft` on the reader thread — where the handler in my other comment then reports it as a reader interruption, so the real cause is hidden.

No in-tree caller combines the two today, so this is a trap for future users rather than a live break. My suggestion is to reject the pair rather than try to define it: `TopicCompactionStrategy<T>` also declares `Schema<T> getSchema()`, so by the interface's own contract its type parameter is the topic's *payload* type — exactly the type a mapped view does not store.

https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-common/src/main/java/org/apache/pulsar/common/topics/TopicCompactionStrategy.java#L57-L61

That makes the combination meaningless rather than merely unimplemented, and failing fast in the builder is the honest answer.

**Where the guard belongs:** beside the mapper null-validation you already added in `TableViewBuilderImpl` ([`TableViewBuilderImpl.java:73`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java#L73) and [`TableViewBuilderImpl.java:83-85`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java#L83-L85)) — and in **both** methods, for the same reason that check is duplicated. `createMapped` raises `IllegalArgumentException` synchronously through `checkArgument`; `createMappedAsync` returns `FutureUtil.failedFuture(new IllegalArgumentException(...))`. If you guard only the async method, `createMapped` surfaces it as a bare `PulsarClientException` instead: `PulsarClientException.unwrap` has no `IllegalArgumentException` branch, so that cause falls through to the generic `new PulsarClientException(t)` ([`PulsarClientException.java:1142-1143`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java#L1142-L1143)), which no longer matches what `testCreateMappedWhenMapperIsNull` ([`TableViewBuilderImplTest.java:161-164`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/test/java/org/apache/pulsar/client/impl/TableViewBuilderImplTest.java#L161-L164)) asserts for its sibling. The builder is also the right layer because the strategy is instantiated in the `AbstractTableViewImpl` constructor ([`:109-110`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L109-L110)), so a builder-level check runs before anything is constructed. Both `checkArgument` and `FutureUtil` are already imported ([`TableViewBuilderImpl.java:21`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java#L21), [`TableViewBuilderImpl.java:36`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TableViewBuilderImpl.java#L36)), so nothing else moves.

<details>
<summary>Suggested guard and tests</summary>

```java
// TableViewBuilderImpl
private static final String COMPACTION_STRATEGY_UNSUPPORTED =
        "topicCompactionStrategyClassName is not supported for mapped table views: a "
                + "TopicCompactionStrategy is defined over the topic's payload type, not the "
                + "mapper's output type";

// in createMapped, next to checkArgument(mapper != null, "mapper cannot be null")
checkArgument(conf.getTopicCompactionStrategyClassName() == null, COMPACTION_STRATEGY_UNSUPPORTED);

// in createMappedAsync, next to the existing mapper null check
if (conf.getTopicCompactionStrategyClassName() != null) {
    return FutureUtil.failedFuture(new IllegalArgumentException(COMPACTION_STRATEGY_UNSUPPORTED));
}
```

One trap worth flagging for the tests: the strategy class name has to be one that actually loads. With a made-up name the test passes whether or not the guard is there, because `TopicCompactionStrategy.load` itself throws `IllegalArgumentException` for a class it cannot instantiate ([`TopicCompactionStrategy.java:88-90`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-common/src/main/java/org/apache/pulsar/common/topics/TopicCompactionStrategy.java#L88-L90)), and `PulsarClientException.unwrap` rethrows a `RuntimeException` cause as-is. A tiny nested implementation keeps the guard as the only possible source:

```java
// TableViewBuilderImplTest — alongside testCreateMappedWhenMapperIsNull
public static class NoopStrategy implements TopicCompactionStrategy<String> {
    @Override
    public Schema<String> getSchema() {
        return Schema.STRING;
    }

    @Override
    public boolean shouldKeepLeft(String prev, String cur) {
        return false;
    }
}

@Test(expectedExceptions = IllegalArgumentException.class)
public void testCreateMappedRejectsCompactionStrategy() throws PulsarClientException {
    tableViewBuilderImpl.topic(TOPIC_NAME)
            .loadConf(Map.of("topicCompactionStrategyClassName", NoopStrategy.class.getName()))
            .createMapped(Message::getKey);
}

@Test
public void testCreateMappedAsyncRejectsCompactionStrategy() {
    assertTrue(tableViewBuilderImpl.topic(TOPIC_NAME)
            .loadConf(Map.of("topicCompactionStrategyClassName", NoopStrategy.class.getName()))
            .createMappedAsync(Message::getKey)
            .isCompletedExceptionally());
}
```

</details>

If you would rather leave the door open than close it, the missing piece is a way to supply a strategy typed over `V`: a class name resolved by `Class.forName` cannot carry the mapper's output type, so it would have to be handed over as an object instead. That is more than this PR needs, and the guard above does not stand in its way.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TableViewBuilder.java
```diff
@@ -76,6 +77,48 @@ public interface TableViewBuilder<T> {
...
+     * <p>If the {@code mapper} function returns {@code null}, it is treated as a tombstone message, and the
+     * corresponding key will be removed from the {@link TableView}.
+     *
+     * @param mapper a function that takes a {@link Message} and returns a custom object of type {@code V}
+     * @param <V> the type of the values in the {@link TableView}
+     * @return the {@link TableView} instance
+     * @throws PulsarClientException
+     *              if the tableView creation fails
+     */
+    <V> TableView<V> createMapped(Function<Message<T>, V> mapper) throws PulsarClientException;
```
lhotari: **[QUALITY] empty-payload messages bypass the mapper entirely, which the public javadoc does not say**

`getValueIfPresent` ([`AbstractTableViewImpl.java:282`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L282)) returns `null` when `msg.size() == 0` **without calling the mapper at all**, and `handleMessage` then removes the key. The precondition is written down on the package-private abstract `getValue` ([`AbstractTableViewImpl.java:289-294`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L289-L294)), but not here on the public API.

That matters because this javadoc advertises mapping from "properties and other metadata" — precisely the case where a keyed message with an empty payload is meaningful rather than a tombstone. A mapper deriving its value only from the key or properties will never see such a message, and the entry is silently deleted instead. With `Function.identity()` the caller does not observe the tombstone `Message` either.

Treating an empty payload as a tombstone is very likely the right semantics for a compacted topic — I am not asking you to change it, only to say so here, so nobody writes a metadata-only mapper and is surprised.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.pulsar.client.api.Schema;
+
+/**
+ * {@link org.apache.pulsar.client.api.TableView} implementation that applies a user-provided mapper
+ * function to each message to produce the value stored in the view.
+ *
+ * @param <T> the message schema type
+ * @param <V> the value type returned by the mapper function
+ */
+public class MessageMapperTableViewImpl<T, V> extends AbstractTableViewImpl<T, V> {
```
lhotari: **[QUALITY] MessageMapperTableViewImpl is public without needing to be**

The class is `public`, but its supertype `AbstractTableViewImpl` is package-private ([`AbstractTableViewImpl.java:57`](https://github.com/apache/pulsar/blob/7e61a1c338d5fbdb627423c4c236de620b8d6cbb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java#L57)) and its constructor is package-private too, so externally it cannot be constructed and exposes no reachable members. Only `TableViewBuilderImpl` builds it, and there is no cross-package test, reflective lookup or `ServiceLoader` entry that needs the name.

`TableViewImpl` is public only because it already was. This class has no such constraint, and public surface is much easier to give away than to take back — package-private now costs nothing.

lhotari (changes requested):
Sorry for the long silence — you asked for another look on 10 July and waited ten weeks for it. Congratulations on getting PIP-445 through the vote — the PIP document (#24842) is on master now too.

All four of my earlier points still check out: the naming is back to `TableViewImpl` / `AbstractTableViewImpl` / `MessageMapperTableViewImpl`, the import churn is gone, and `MessageTableView` is removed. Since then your fork merged upstream `master` again, and that merge correctly ports #26566 ("Complete table view refresh after applying messages") into the moved `handleMessage` in `AbstractTableViewImpl`: `lastReadPositions` is now recorded in each completion branch, after the value has been produced, instead of unconditionally before processing starts. That also closes a point I was going to raise — a throwing mapper leaving the read position advanced, so `refreshAsync` could report stale data as fresh — so I have dropped it.

I paid particular attention to the pooling decision, because it is the part most likely to go wrong, and **it is right**. `TableViewImpl` passes `super(..., true)` and `MessageMapperTableViewImpl` passes `false`; `AbstractTableViewImpl` releases in a `finally` only `if (poolMessages)`, and the same flag drives `.poolMessages(...)` on the reader builder, so allocation and release cannot drift apart. I also diffed the classic path against master: it already used `.poolMessages(true)` with an unconditional `release()`, so classic `TableView` behaviour is preserved exactly. Your reasoning in the PR description — that the mapper may retain the `Message`, so the pooled buffer must not be reused — is the correct call.

What I would like changed is not the design but the **contract around the mapper**, and it matters more here than it would elsewhere because `createMapped` is new public API: once released, the behaviour is fixed. Four inline comments below; the first is the real question — *what happens when a user's mapper throws?* — and the other three are smaller.

That behaviour is **pre-existing** and I want to be fair about it. On master, `TableViewImpl` already routed lazy `Message.getValue()` schema-decode failures through the same handler. That is not something you introduced. What changes is the blast radius: a schema-decode failure is rare and roughly deterministic, whereas a mapper is arbitrary user code that this PR invites people to write. That is why I think the new API should either guard or document it rather than inherit it silently.

One smaller thing I considered and decided **not** to raise as a finding, because I could not substantiate the impact: the shared logger moved from `Logger.get(TableViewImpl.class)` to `Logger.get(AbstractTableViewImpl.class)`, so classic-TableView log lines change category. Pulsar documents no compatibility guarantee for client logger names and I found no in-tree config keyed on the old one, so this is only worth doing if you think it is cheap — `Logger.get(getClass())` in the constructor would keep per-subclass categories.

Nothing here questions PIP-445 or the shape of the change. It is close.

lhotari:
Thanks for the detailed summary, and sorry for the ten-week wait — that was on me, not on anything missing from your side.

I have gone through it again at the current head. All four of my earlier points check out, and the pooling decision you describe is right: I diffed the classic path against master and it keeps `.poolMessages(true)` with its unconditional `release()`, while the mapped path opts out, so nothing can release a buffer your mapper still holds.

My remaining comments are about the contract around the mapper rather than the design — mainly what should happen when a user's mapper throws, since `createMapped` is new public API and that behaviour becomes fixed once released. One of the behaviours I flag is pre-existing on master and I have said so in that comment; the concern is that a mapper is arbitrary user code where a schema-decode failure was not. Congratulations on the PIP-445 vote.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java
lhotari: Addressed in a final refactoring pass that revised the design slightly to give the mapper an explicit error model: `createMapped` now takes a `TableViewMessageMapper<T, V>` whose `map` may throw and whose `default onMappingError` reports the failure (logged at `ERROR` with topic, key and message id unless the callback returns `true`). A throwing mapper now skips only that message in both the replay and the tail phase — the key keeps its previous value, listeners are not notified, and the refresh position still advances — and `start()` closes the reader when the initial replay fails. Covered by `MappedTableViewImplTest` and `TableViewTest#testCreateMappedWithFailingMapper`.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/AbstractTableViewImpl.java
lhotari: Addressed in the final refactoring pass: both `createMapped` and `createMappedAsync` reject a configured `topicCompactionStrategyClassName` with `IllegalArgumentException` (a failed future in the async variant) before anything is constructed. The tests use a loadable `NoopStrategy`, so the guard is the only possible source of the exception.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TableViewBuilder.java
lhotari: Addressed in the final refactoring pass: the `createMapped` javadoc and `TableViewMessageMapper#map` now state that a keyed message with an empty payload is a tombstone that removes the key without calling the mapper, that a `null` return is also a tombstone, and what happens when the mapper throws. The semantics are unchanged; they are now part of the public contract.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java
lhotari: Addressed in the final refactoring pass: the class is now package-private and renamed to `MappedTableViewImpl` to match the `createMapped` builder methods. PIP-445 was updated to the same name.

lhotari:
@namest504 thank you for your persistence and patience in getting PIP-445 through the design discussion and the vote, and now finally into Pulsar — this took far longer than it should have, and that was largely on the review side.

To get it over the line I pushed a final refactoring pass to your branch (maintainer edits are enabled) that revises the design slightly to tackle the error model: `createMapped` takes a `TableViewMessageMapper` (`map` may throw, `onMappingError` reports failures), mapped views reject compaction strategies, the impl is the package-private `MappedTableViewImpl`, tombstones are documented, and PIP-445 plus the PR description are updated to match. I also merged `master` to resolve the conflict with #26626 and ported that fix into `AbstractTableViewImpl`. Each review thread has a short note on how it was handled; please take a look and let me know if anything looks off.

lhotari (approved):
LGTM

david-streamlio:
Nice work — the refactor is a faithful extraction. I diffed `AbstractTableViewImpl` line-by-line against the base `TableViewImpl` at `a846a45953` and the only core-logic deltas are the ones the PIP claims: generify to `<T, V>`, the `poolMessages` flag, close-on-start-failure, and the guarded value extraction. `createMappedAsync` correctly returns failed futures rather than throwing synchronously (`TableViewBuilderImpl.java:88-93`), there's no new reflection into private state, no blocking on the reader thread, and no new `Recycler` usage.

I ran `./gradlew :pulsar-client-original:test --tests MappedTableViewImplTest --tests TableViewBuilderImplTest -PtestRetryCount=0` → 22/22 green. I did not run the `pulsar-broker` tests (`TableViewTest`, `ServiceUnitStateCompactionTest`).

Everything below is non-blocking; the PR already has an approval. Two points I'd like your read on, then some nits.

### 1. The `create()` behavior change lands on the broker's own load-balancer path

`AbstractTableViewImpl.java:249-262` now swallows a decode failure for plain table views too, not just mapped ones. `ServiceUnitStateTableViewImpl.java:94` builds its channel view with `create()`, so a `ServiceUnitStateData` message that fails to decode used to fail channel start loudly and now leaves a silent hole in the ownership view, observable only as an ERROR log.

The PIP documents the general change in its compatibility section, but doesn't name this consumer. Is that trade deliberate? Skipping is probably the better operational default — a broker that won't start its load-balancer channel is worse than one that drops a corrupt record — but `create()` users have no programmatic signal at all. Worth either calling the consumer out explicitly in the PIP, or considering whether non-mapped views should get an equivalent of `onMappingError` at some point.

### 2. `createMapped(msg -> msg)` retains more than the message

The pooling claim itself checks out: with `poolMessage == false`, `MessageImpl.java:207` does `Unpooled.copiedBuffer(payload)`, so retaining the `Message` is safe. But `MessageImpl.java:196` also assigns `msg.cnx = cnx`, and the instance holds `msgMetadata`, `brokerEntryMetadata` and `schema`. So an identity-mapped view over a large keyspace pins a `ClientCnx` reference per stored key, keeping old connection objects reachable across reconnects.

Safe, just not cheap. A sentence in the `TableViewBuilder#createMapped` javadoc recommending extraction of the needed fields over `msg -> msg` for high-cardinality topics would set the right expectation.

### Nits

- **`AbstractTableViewImpl.java:135-149`** — the reader-leak fix doesn't cover the adjacent path. `readAllExistingMessages` (line 434) has no `.exceptionally` on `hasMessageAvailableAsync()`, so if that call fails, `future` never completes, `start()` hangs, and the new `whenComplete` never fires — reader still leaked. Pre-existing, but you're fixing the leak right next to it; one `.exceptionally(ex -> { future.completeExceptionally(ex); return null; })` would close the case.
- **`AbstractTableViewImpl.java:251`** — `catch (Throwable t)` routes `OutOfMemoryError` / `StackOverflowError` into `onMappingError` and keeps reading. `map` declares `throws Exception`, so `catch (Exception e)` matches the contract and lets a `VirtualMachineError` escape.
- **`AbstractTableViewImpl.java:259`** — `"Failed to map message, skipping it"` reads oddly for plain `TableViewImpl`, where there is no mapper and the cause is a decode failure.
- **`TableViewBuilderImpl.java:77-78`** — the two `checkArgument` calls are redundant: `PulsarClientException.unwrap` rethrows `RuntimeException` unchanged (`PulsarClientException.java:1058-1059`), so the async guards already surface `IllegalArgumentException` from `createMapped()`. Harmless, and it does fail fast before a reader is created; just two places to keep in sync.
- **`MappedTableViewImpl.java:37-44`** — the compaction-strategy invariant is enforced only in the builder. A defensive check in the constructor would keep it local to the class that would otherwise hit a `ClassCastException` on the reader thread.
- Making `AbstractTableViewImpl` package-private while `TableViewImpl` stays public is what forced `readDeclaredField` → `readField` in `TableViewTest` and `ServiceUnitStateCompactionTest`. Fine inside `client.impl`, just noting the reflective-access consequence is real for anyone doing the same downstream.

Adding two abstract methods to `TableViewBuilder` looks fine to me — it's `@InterfaceStability.Evolving` and `TableViewBuilderImpl` is the only implementation in-tree.

lhotari:
@david-streamlio thanks for the thorough second pass — the line-by-line diff against the base `TableViewImpl` is exactly the kind of check this extraction needed. Here is what I did with each point.

**1. `create()` behaviour change on the load balancer channel.** Deliberate, and now called out by name. The PIP-445 compatibility section names `ServiceUnitStateTableViewImpl` and spells out the trade: on master a `ServiceUnitStateData` that fails to decode was *already* skipped (with a log that blamed the reader) when it arrived while tailing; only the replay phase changes, from "channel does not start" to "record skipped, logged at `ERROR` with topic, key and message id". A broker that cannot start its load balancer channel because of one undecodable record is the worse outcome. A programmatic failure signal for plain `create()` views (an `onMappingError` equivalent) is noted as follow-up material rather than added here — it would be new public API on the classic path.

**2. `msg -> msg` retains more than the payload.** Correct — `MessageImpl.create` assigns `cnx`, and the instance also carries `msgMetadata`, `brokerEntryMetadata` and the schema. The `createMapped` javadoc now says so and recommends copying the needed fields into a value object for topics with many keys.

**Nits**

- **`hasMessageAvailableAsync()` failure hangs `create()`** — fixed. The outer chain in `readAllExistingMessages` now fails the replay future, so `start()` completes exceptionally and the `whenComplete` closes the reader. New test `testStartFailsWhenHasMessageAvailableFails`; with the `.exceptionally` removed it fails with a `TimeoutException` after 5 s, which is the hang.
- **`catch (Throwable)` around the mapper** — kept, on purpose. The mapper runs inside `CompletableFuture` callbacks, and `CompletableFuture` catches `Throwable` regardless, so an `Error` cannot escape to the JVM either way: the only choice is whether it is attributed to the mapper (with key and message id, via `onMappingError`) or misdiagnosed as a reader failure and retried — which is the misattribution the guard exists to remove. `StackOverflowError` from a recursive mapper is the realistic case, and by the time it is caught the stack has unwound. It also matches how the client treats every other user callback: `ConsumerBase` catches `Throwable` around `MessageListener.received`, and the table view listener loop has always caught `Throwable`.
- **Log wording** — changed to "Skipping message whose value could not be decoded or mapped", which reads correctly for both the classic and the mapped path.
- **Redundant `checkArgument` in `createMapped`** — they are load-bearing, not redundant. `unwrap` rethrows a `RuntimeException` only when it is the *top-level* throwable; `.get()` throws a checked `ExecutionException`, and in the cause-unwrapping branch there is no `IllegalArgumentException` case, so it falls through to the generic `new PulsarClientException(t)`. I verified by deleting the sync null check: `testCreateMappedWhenMapperIsNull` then fails with `Expected exception of type IllegalArgumentException but got PulsarClientException: java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: mapper cannot be null`. Without them `createMapped()` would break the documented `@throws IllegalArgumentException` contract.
- **Compaction-strategy invariant local to `MappedTableViewImpl`** — done. The constructor rejects a configured strategy itself, via a static helper evaluated in the `super(...)` argument list so it runs before the reader is created (the client targets release 17, so no flexible constructor bodies). The class now owns the message constant and the builder reuses it. New test `testConstructorRejectsCompactionStrategyBeforeCreatingReader` asserts the `IllegalArgumentException` and `verify(client, never()).newReader(...)`.
- **`readDeclaredField` → `readField`** — agreed it is a real consequence for downstream reflective access; it stays as is since `AbstractTableViewImpl` is package-private on purpose.

namest504:
@lhotari Thank you for taking this over the line, and for the notes on each thread. I had been going back and forth on how a throwing mapper should be handled, and the direction you took makes the trade-off clear to me: skipping the message keeps a long-running view alive, and `onMappingError` means the skip is not silent. Having that spelled out in the javadoc as a contract is something I would not have arrived at on my own. I went through the changes and the follow-ups from @david-streamlio's review as well, and nothing looks off to me.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageMapperTableViewImpl.java
namest504: One thing from this comment that is outside the scope of the PR: the 50 ms `Thread.sleep` in the retry branch predates this change (#22064) and is still there after the refactoring. I have filed it separately as #26671, with a fix that schedules the retry on the client's scheduled executor instead.

lhotari: @namest504 thanks for catching the #26671 issue, yes it's worth fixing in a separate PR. I'll merge this current PR so that you can work on it.

PR #21080, opened 2023-08-28
[improve][pip] PIP-264: Enhanced OTel-based metric system

asafm:
### Motivation

Add PIP-264 (copied from https://github.com/apache/pulsar/issues/20197)

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change is a trivial rework / code cleanup without any test coverage.
### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24815, opened 2025-10-03
[feat][monitor] Add ML write latency histogram and entry size histogram as OTel metrics

equanz:
PIP: #21080

### Motivation

Please see the [PIP-264 doc](https://github.com/apache/pulsar/blob/v4.1.1/pip/pip-264.md).

ref.
* https://github.com/apache/pulsar/blob/v4.1.1/pip/pip-264.md#moving-topic-level-histograms-to-namespace-and-broker-level-only
  * Note:
  > Histograms are problematic in that sense, since they yet to have an asynchronous version - alas they are only synchronous. Hence, if we use them for topic instruments, used attributes can never be cleared from memory for them thus we’ll have an “attributes” (topic) leak, as over time it will only grow.
  * Therefore, even if a namespace is deleted, the namespace-level histogram metrics persist
* https://github.com/apache/pulsar/blob/v4.1.1/pip/pip-264.md#which-summaries--histograms-are-effected

### Modifications

* Added ML write latency histogram and entry size histogram as OTel metrics

### Verifying this change

- Added a simple E2E test to verify the expected metrics is exposed

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics
  - Introducing the new OTel metrics
    - `pulsar.broker.managed_ledger.message.outgoing.latency`
    - `pulsar.broker.managed_ledger.message.outgoing.ledger.latency`
    - `pulsar.broker.managed_ledger.ledger.switch.latency`
    - `pulsar.broker.managed_ledger.entry.size`

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/equanz/pulsar/pull/14

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedLedgerStats.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedLedgerStats.java
@@ -20,11 +20,17 @@ package org.apache.bookkeeper.mledger.impl;
 
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.metrics.BatchCallback;
+import io.opentelemetry.api.metrics.DoubleHistogram;
+import io.opentelemetry.api.metrics.LongHistogram;
 import io.opentelemetry.api.metrics.ObservableLongMeasurement;
+import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.mledger.ManagedLedger;
 import org.apache.pulsar.opentelemetry.Constants;
 
 public class OpenTelemetryManagedLedgerStats implements AutoCloseable {
+
+    // ml-level metrics
 
     // Replaces pulsar_ml_AddEntryMessagesRate
     public static final String ADD_ENTRY_COUNTER = "pulsar.broker.managed_ledger.message.outgoing.count";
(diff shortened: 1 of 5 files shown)


equanz:
Note:
In my understanding, implementing AutoConfigurationCustomizerProvider allows you to customize the histogram buckets.

Here is an example.
```java
import com.google.auto.service.AutoService;
import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
import io.opentelemetry.sdk.metrics.Aggregation;
import io.opentelemetry.sdk.metrics.InstrumentSelector;
import io.opentelemetry.sdk.metrics.View;
import java.util.Arrays;

@AutoService(AutoConfigurationCustomizerProvider.class)
public class MyCustomizerProvider implements AutoConfigurationCustomizerProvider {

    @Override
    public void customize(AutoConfigurationCustomizer customizer) {
        customizer.addMeterProviderCustomizer(((sdkMeterProviderBuilder, configProperties) -> {
            sdkMeterProviderBuilder.registerView(InstrumentSelector.builder()
                            .setName("pulsar.broker.topic.publish.latency").build(),
                    View.builder()
                            .setAggregation(Aggregation.explicitBucketHistogram(Arrays.asList(0.001, 30.0))).build());
            return sdkMeterProviderBuilder;
        }));
    }
}
```

```
$ PULSAR_EXTRA_CLASSPATH='path/to/custom-provider.jar' PULSAR_EXTRA_OPTS=' -Dotel.sdk.disabled=false -Dotel.metrics.exporter=prometheus ' ./bin/pulsar standalone
```

ref.
* https://opentelemetry.io/docs/languages/java/sdk/#views
* https://opentelemetry.io/docs/languages/java/configuration/#zero-code-sdk-autoconfigure
  * The `pulsar-opentelemetry` uses AutoConfiguredOpenTelemetrySdk. So, we can use AutoConfigurationCustomizerProvider
  https://github.com/apache/pulsar/blob/v4.1.1/pulsar-opentelemetry/src/main/java/org/apache/pulsar/opentelemetry/OpenTelemetryService.java#L72
* https://opentelemetry.io/ja/docs/languages/java/configuration/#autoconfigurationcustomizerprovider

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedLedgerStats.java
```diff
@@ -151,4 +213,24 @@ private void recordMetrics(ManagedLedger ml) {
         markDeleteCounter.record(stats.getMarkDeleteTotal(), attributes);
         readEntryCacheMissCounter.record(stats.getReadEntriesOpsCacheMissesTotal(), attributes);
     }
+
+    void recordAddEntryLatency(long latency, TimeUnit unit, ManagedLedger ml) {
+        final var attributes = ml.getManagedLedgerAttributes().getAttributesOnlyNamespace();
```
lhotari, line 223: Is there a specific reason to track add namespace level? Could you add comments in the source code to explain the reasoning?

lhotari: Now I got it from the referenced PIP-264 section. "In OTel, the API doesn't support remove() for an attribute set on an instrument (counter, histogram, etc.).". I didn't understand the asynchronous vs. synchronous description and the lack of being able to remove the instruments makes more sense to me. Perhaps summarizing the reason to use `getAttributesOnlyNamespace` in a code comment would be useful.

equanz: > Perhaps summarizing the reason to use getAttributesOnlyNamespace in a code comment would be useful.

Added. PTAL

https://github.com/apache/pulsar/compare/f65a473b30130cbd7bc21b6fe02c085473396b36..14d025cc1899b33337afa6623113e69be4c7d12e

lhotari (approved):
LGTM

equanz:
Rebased to the master to resolve conflicts.

PR #23136, opened 2024-08-07
[fix] [broker] Fix compatibility issues for PIP-344

poorbarcode:
### Motivation & Modifications

After [PIP-344 add a public API pulsarClient.getPartitionsForTopic(String topic, boolean metadataAutoCreationEnabled)](https://github.com/apache/pulsar/pull/22182), there are `3` compatibility issues:

---

#### 1. Using DLQ( brokers’ version is  `<3.0.6`, and clients’ version is `>=3.0.6`)
PIP-344 fixed a bug that the Retry/DLQ consumer will still trigger a Retry/DLQ topic with the old rule `{namespace}/{subscription}-RETRY/DLQ`, but the rule was changed to `{namespace}/{topic}-{subscription}-RETRY/DLQ` after `2.8.0`. This behavior has been defined in the PIP-344, so it is fine.

**Modifications for compatibility**: modify the error message to let users know what happened.

---

#### 2. Using Geo-Replication, the version of the local cluster and the remote cluster is not the same ( local clusters are `>=3.0.6` , and remote clusters are `<3.0.6` )

1. This error will only occur when using Geo-Replication, and one version of the two clusters is larger or equal to `3.0.6` and "3.3.1" and another is smaller than `3.0.6` and `3.3.1`.
2. **Reason why getting the not supported error when creating internal Replicator:** The feature method above was supported at `3.0.6` and `3.3.1`, before that the API  "getPartitionedTopicMetadata" will trigger a creation for partitioned topic metadata automatically even if you just want to query it. So the brokers whose version is less than `3.0.6` and `3.3.1` do not support the new API.
3. **Modifications for compatibility:** Skip the check of comparing of topic's partitions, and force connect to the non-partitioned topic,  it may cause both the partitioned topic and non-partitioned topic to exist at the same time. But this is still better than the behavior before fix #22983, without fix #22838, there is an issue that may cause replication stuck and topics to be created in confusion, see more details in #22838's motivation.

---

#### 3. Using non-persistent topics, the versions of the brokers in the same cluster are different( some are less than `3.0.6` )

1. **Reason why getting the not supported error when creating a non-persistent consumer/producer:** The feature method `getPartitionedTopicMetadata(topic, false)` was supported at `3.0.6` and `3.3.1`, before that the API `getPartitionedTopicMetadata(topic)` will trigger a creation for partitioned topic metadata automatically even if you just want query it. So the brokers whose version is less than `3.0.6` and `3.3.1` do not support the new API.
[...]

PR #22773, opened 2024-05-23
[improve] [client] PIP-344 support feature flag supportsGetPartitionedMetadataWithoutAutoCreation

poorbarcode:
### Motivation & Modifications

[PIP-344 Correct the behavior of the public API pulsarClient.getPartitionsForTopic(topicName)](https://github.com/apache/pulsar/pull/22182)

Section [Backward & Forward Compatibility](https://github.com/apache/pulsar/blob/master/pip/pip-344.md#backward--forward-compatibility)

- Old version client and New version Broker: The client will call the old API.

- New version client and Old version Broker: The feature flag `supports_binary_api_get_partitioned_meta_with_param_created_false` will be `false`. The client will get a not-support error if the param `createIfAutoCreationEnabled` is `false`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: x

PR #24822, opened 2025-10-04
[fix][client] Make auto partitions update work for old brokers without PIP-344

BewareMyPower:
### Motivation

See https://lists.apache.org/thread/tpd2kgl4v3x8bkw0skv0nj38rzrk56t7

### Modifications

Set the 3rd argument of `getPartitionedTopicMetadata` to `true` in `PulsarClientImpl#getPartitionsForTopic`, which is used to query the number of partitions.

Add `testAutoPartitionsUpdate` to cover the case. Specially, it tests a consumer that subscribes two partitioned topics to cover the path here: https://github.com/apache/pulsar/blob/008e4ebbaf5808c0f859dddabd8382991f32d3aa/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java#L182

P.S. This test is only added to `ClientTest25`, which runs tests against `apachepulsar/pulsar:2.5.0` image. The current backward compatibility test strategy is ambiguous that only 2.2.0, 2.3.0, 2.4.0, 2.5.0 are tested. It's hard to cover all old versions, but only choose these 4 versions does not make sense. This topic should be out of the scope of this PR. Anyway, the test on 2.5.0 can already prevent regression on the fix.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
@@ -1291,7 +1291,7 @@ public class PulsarClientImpl implements PulsarClient {
 
     @Override
     public CompletableFuture<List<String>> getPartitionsForTopic(String topic, boolean metadataAutoCreationEnabled) {
-        return getPartitionedTopicMetadata(topic, metadataAutoCreationEnabled, false).thenApply(metadata -> {
+        return getPartitionedTopicMetadata(topic, metadataAutoCreationEnabled, true).thenApply(metadata -> {
             if (metadata.partitions > 0) {
                 TopicName topicName = TopicName.get(topic);
                 List<String> partitions = new ArrayList<>(metadata.partitions);
--- a/tests/integration/src/test/java/org/apache/pulsar/tests/integration/backwardscompatibility/ClientTest25.java
+++ b/tests/integration/src/test/java/org/apache/pulsar/tests/integration/backwardscompatibility/ClientTest25.java
@@ -19,8 +19,19 @@
 package org.apache.pulsar.tests.integration.backwardscompatibility;
 
 
+import java.util.List;
+import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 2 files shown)


lhotari (approved):
LGTM, thanks for addressing this

lhotari:
There seems to be a [test that explicitly tests for the behavior that connecting to an older broker fails](https://github.com/apache/pulsar/blob/24f38145e7beeee35d28f871ac9be65aa65fb1bc/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java#L236-L273):
```
  Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 32.05 s <<< FAILURE! -- in org.apache.pulsar.client.impl.ClientCnxTest
  Error:  org.apache.pulsar.client.impl.ClientCnxTest.testSupportsGetPartitionedMetadataWithoutAutoCreation -- Time elapsed: 0.255 s <<< FAILURE!
  java.lang.AssertionError: Expected an error that the broker version is too old.
  	at org.apache.pulsar.client.impl.ClientCnxTest.testSupportsGetPartitionedMetadataWithoutAutoCreation(ClientCnxTest.java:264)
```

This test was added in #22773 and it seems to test what is defined in [PIP-344 Backward and Forward compatibility](https://github.com/apache/pulsar/blob/master/pip/pip-344.md#backward--forward-compatibility):

> - New version client and Old version Broker: The feature flag `supports_binary_api_get_partitioned_meta_with_param_created_false` will be `false`. The client will get a not-support error if the param `createIfAutoCreationEnabled` is false.

@poorbarcode would it be possible to explain the reason why the client should fail in this case?

BewareMyPower:
I think it has been explained in https://github.com/apache/pulsar/pull/23136. I updated the test and the PIP document now.

PR #24830, opened 2025-10-08
[fix][client] Fix getPendingQueueSize for PartitionedTopicProducerStatsRecorderImpl: avoid NPE and implement aggregation

lhotari:
### Motivation

When getPendingQueueSize is called for stats returned from a multi-topics producer, it will result in NPE

### Modifications

- add a null check to avoid NPE
- implement aggregation for getPendingQueueSize

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerStatsRecorderImpl.java
```diff
@@ -332,7 +332,7 @@ public double getSendLatencyMillisMax() {
 
     @Override
     public int getPendingQueueSize() {
-        return producer.getPendingQueueSize();
+        return producer != null ? producer.getPendingQueueSize() : -1;
```
nodece:
```suggestion
        return producer != null ? producer.getPendingQueueSize() : 0;
```

I think we should return 0 if producer is null.

PR #18998, opened 2022-12-20
[improve][broker] Add javadoc for PulsarDecoder handle* method parameter lifecycle

lhotari:
### Motivation

The lifecycle of the PulsarDecoder handle* method parameters isn't currently documented.

This PR adds this kind of paragraph:
>  Please be aware that the decoded protocol command instance passed to a handle* method is cleared and reused for the next protocol command after the method completes. This is done in order to minimize object allocations for performance reasons. <b>It is not allowed to retain a reference to the handle* method parameter command instance after the method returns.</b> If you need to pass an instance of the command instance to another thread or retain a reference to it after the handle* method completes, you must make a deep copy of the command instance.

See #18987 or #10215 for a sample issues where a related problem was fixed.

### Modifications

Add Javadoc to PulsarDecoder and reference it in the implementation classes.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/lhotari/pulsar/pull/115

PR #24832, opened 2025-10-08
[fix] Fix mixed lookup/partition metadata requests causing reliability issues and incorrect responses

lhotari:
### Motivation

PulsarDecoder unnecessarily keeps a reference to the last parsed message. It's better to clear it immediately to avoid unnecessary memory growth when there are multiple connections.
This mainly impacts `GET_TOPICS_OF_NAMESPACE_RESPONSE` which can be large.

While making this change, it was noticed that this revealed a major reliability issue in Pulsar. Lookup requests and Partition metadata requests were getting reused and potentially mixed up in cases where there would be multiple concurrent lookup requests in progress on the same connection. The impact of this has been that lookup or partition metadata requests have failed and also that they have returned incorrect information.
The problem in these cases has been a violation of the rule documented in #18998. The implementations of `handle*` methods shouldn't keep references to the command after the method returns. When the command is handle asynchronously, a copy is usually needed if the asynchronous handler makes a reference to the command.

### Modifications

- call `cmd.clear()` in the finally block of `channelRead`.
- fix bugs in `handleLookup`, `handlePartitionMetadataRequest` and `handleCommandWatchTopicList`
  - make copy of command before handling it asynchronously

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
+++ b/pulsar-client/src/test/java/org/apache/pulsar/client/impl/ClientCnxTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.pulsar.client.impl;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
@@ -334,7 +335,9 @@ public class ClientCnxTest {
                     .setRequestId(7)
                     .setWatcherId(5).setTopicsHash("f00");
             cnx.handleCommandWatchTopicListSuccess(success);
-            assertEquals(result.getNow(null), success);
+            assertThat(result.getNow(null))
+                    .usingRecursiveComparison()
+                    .comparingOnlyFields("requestId", "watcherId", "topicsHash");
         });
     }
 


pulsar-common/src/main/java/org/apache/pulsar/common/protocol/PulsarDecoder.java
```diff
@@ -483,6 +483,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
             }
         } finally {
             buffer.release();
+            cmd.clear();
```
BewareMyPower, line 490: It would be better to add a comment here.

`cmd.parseFrom` will be called each time in `channelRead`, calling `clear` seems to be unnecessary. However, `CommandGetTopicsOfNamespaceResponse` has a `topics` field, whose size could be large. Calling `clear()` mainly makes sense for this specific response.

lhotari: I'll add a comment. This change actually revealed a major bug that has been impacting Pulsar's reliability for a long time. The command is essentially reused and no references should be kept to it. This was violated in multiple locations and for example lookups and partition metadata requests have been failing because of this bug. 

Yes, I'm aware that cmd.parseFrom calls clear. Calling clear twice is not going to be a performance problem. The flags will get reset and it's just going to be some ifs that are going to execute. I'd assume that a modern CPU can execute 100s of millions of such code lines per second, so this is not going to be a problem. In this case, it's essential to call clear so that possible bugs where the cmd reference is kept in other threads will come to the surface in CI.

lhotari: #18998 added javadoc about the restriction, but this change will help enforce that.

lhotari: comment added

BewareMyPower:
> I'm still concerned that asynchronous handling may lead to accessing variables after cmd have been cleared (for example, in the interceptor).

Do you mean `cmd` is published to a different thread? If so, we should not do that. In the interceptor case, the interceptor should be responsible to make a deep copy on `cmd`. A better approach is to copy necessary fields to a new object.

BTW, I checked the code in Pulsar that `cmd` is never accessed out of the `channelRead` method. Please correct me if I'm wrong.

nodece:
Deep copy is a good idea.

PR #24833, opened 2025-10-09
[feat] PIP-442: Add memory limits for CommandGetTopicsOfNamespace

lhotari:
[PIP-442](https://github.com/apache/pulsar/blob/master/pip/pip-442.md)

### Motivation

See [PIP-442](https://github.com/apache/pulsar/blob/master/pip/pip-442.md).
Apache Pulsar brokers currently lack memory limits for topic listing operations (`CommandGetTopicsOfNamespace` and `CommandWatchTopicList`), creating a significant gap in the broker's otherwise comprehensive memory management framework. This can lead to:

- **Broker/Proxy Memory Exhaustion**: Large namespaces with thousands of topics can trigger OutOfMemoryErrors when multiple clients simultaneously request topic lists
- **Unpredictable Resource Usage**: Operators cannot reliably predict or limit total memory consumption due to this unbounded allocation path
- **Performance Degradation**: Large topic list operations cause GC pressure and latency spikes affecting all broker operations

While Pulsar has robust memory limits for message publishing (`maxMessagePublishBufferSizeInMB`), managed ledger operations (`managedLedgerMaxReadsInFlightSizeInMB`, `managedLedgerCacheSizeMB`), and concurrent requests (`maxConcurrentLookupRequest`), topic listing operations remain unbounded.

## Modifications

This PR implements PIP-442 by introducing memory-aware flow control for topic listing operations:

### Core Components

1. **AsyncSemaphore Interface & Implementation**
   - Generic asynchronous semaphore abstraction with configurable timeouts and queue limits
   - Supports permit acquisition, updates, and cancellation
   - Tracks queue size and wait times for monitoring

2. **AsyncDualMemoryLimiter Interface & Implementation**
   - Dual memory pool tracking for heap memory (topic list assembly) and direct memory (network buffers)
   - Separate limits and queues for each memory type
   - Helper methods for common patterns (acquire-execute-release)

3. **AsyncDualMemoryLimiterUtil**
   - Utility methods for acquiring permits and writing responses
   - Handles serialization and network buffer management
   - Ensures permits are released after write completion or failure

4. **TopicListMemoryLimiter**
   - Extends AsyncDualMemoryLimiterImpl with metrics integration
   - Exposes both Prometheus and OpenTelemetry metrics
   - Tracks memory usage, queue sizes, wait times, and timeouts

### Integration Points

**Broker Integration:**
- Modified `ServerCnx.handleGetTopicsOfNamespace()` to acquire initial heap permits (1KB estimate), update to actual size after topic retrieval, then acquire direct memory permits for response serialization
- Modified `PulsarCommandSenderImpl` to acquire direct memory permits before serialization and release after write completion
- Created `TopicListMemoryLimiter` in `BrokerService` with configured limits

**Proxy Integration:**
- Modified `LookupProxyHandler.handleGetTopicsOfNamespace()` with similar permit acquisition flow
- Handles both heap memory (for topic list from broker) and direct memory (for response to client)
- Created `TopicListMemoryLimiter` in `ProxyService` with configured limits

### Configuration

Added six new configuration parameters to `broker.conf`, `proxy.conf`, and `standalone.conf`:

```properties
maxTopicListInFlightHeapMemSizeMB=100
maxTopicListInFlightDirectMemSizeMB=100
maxTopicListInFlightHeapMemSizePermitsAcquireTimeoutMillis=25000
maxTopicListInFlightHeapMemSizePermitsAcquireQueueSize=10000
maxTopicListInFlightDirectMemSizePermitsAcquireTimeoutMillis=25000
maxTopicListInFlightDirectMemSizePermitsAcquireQueueSize=10000
```

### Metrics

Added 12 new metrics per instance (broker/proxy) with both Prometheus and OpenTelemetry support:

- Memory usage and limits (heap and direct)
- Queue sizes and limits
- Wait time summaries (P50, P95, P99, Max)
- Timeout counters

## Verifying this change

- [x] Make sure that the change passes the CI checks

- **Added `PatternConsumerBackPressureMultipleConsumersTest`**: Integration test that creates 8,192 partitioned topics and sends 500 concurrent `getTopicsOfNamespace` requests from 200 different client connections. The test intentionally reduces available direct memory to reproduce memory pressure scenarios and verifies all requests complete successfully with memory limiting in place.

- **Added `ProxyPatternConsumerBackPressureMultipleConsumersTest`**: Extended version of the broker test that routes requests through the Pulsar Proxy to verify memory limiting works correctly in proxy scenarios.

- **Added `AsyncSemaphoreImplTest`**: Comprehensive unit tests for the async semaphore implementation covering:
  - Basic acquire/release operations
  - Queueing behavior when permits unavailable
  - Timeout and queue full exceptions
  - Cancellation support
  - Permit updates (increase/decrease)
  - Concurrent operations

- **Added `AsyncDualMemoryLimiterImplTest`**: Unit tests for the dual memory limiter covering:
  - Independent heap and direct memory limiting
  - Permit acquisition and updates for both types
  - Queue behavior and timeout handling
  - Concurrent operations with mixed memory types
  - Proper resource cleanup

- **Added `AsyncDualMemoryLimiterUtilTest`**: Tests for utility methods covering:
  - Successful permit acquisition and response writing
  - Error handling for permit acquisition failures
  - Write failures and proper permit release
  - Cancellation scenarios
  - Concurrent operations

- **Verified metrics**: Tests validate that all 26 metrics are properly registered and report correct values.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
@@ -18,48 +18,30 @@
  */
 package org.apache.pulsar.broker.service;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Executor;
-import java.util.concurrent.LinkedBlockingDeque;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
-import java.util.function.BooleanSupplier;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Supplier;
 import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.namespace.NamespaceService;
 import org.apache.pulsar.broker.resources.TopicResources;
-import org.apache.pulsar.broker.topiclistlimit.TopicListMemoryLimiter;
-import org.apache.pulsar.broker.topiclistlimit.TopicListSizeResultCache;
-import org.apache.pulsar.common.api.proto.CommandGetTopicsOfNamespace;
 import org.apache.pulsar.common.api.proto.CommandWatchTopicListClose;
 import org.apache.pulsar.common.api.proto.ServerError;
 import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.TopicName;
-import org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiter;
-import org.apache.pulsar.common.semaphore.AsyncSemaphore;
 import org.apache.pulsar.common.topics.TopicList;
 import org.apache.pulsar.common.topics.TopicsPattern;
 import org.apache.pulsar.common.topics.TopicsPatternFactory;
-import org.apache.pulsar.common.util.Backoff;
-import org.apache.pulsar.common.util.FutureUtil;
-import org.apache.pulsar.common.util.Runnables;
 import org.apache.pulsar.common.util.collections.ConcurrentLongHashMap;
 import org.apache.pulsar.metadata.api.NotificationType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class TopicListService {
+
+
     public static class TopicListWatcher implements BiConsumer<String, NotificationType> {
 
         /** Topic names which are matching, the topic name contains the partition suffix. **/
@@ -68,21 +50,15 @@ public class TopicListService {
         private final long id;
         /** The regexp for the topic name(not contains partition suffix). **/
         private final TopicsPattern topicsPattern;
-        private final Executor executor;
-        private volatile boolean closed = false;
-        private boolean sendTopicListSuccessCompleted = false;
-        private BlockingDeque<Runnable> sendTopicListUpdateTasksBeforeInit = new LinkedBlockingDeque<>();
 
         /***
          * @param topicsPattern The regexp for the topic name(not contains partition suffix).
          */
         public TopicListWatcher(TopicListService topicListService, long id,
-                                TopicsPattern topicsPattern, List<String> topics,
-                                Executor executor) {
+                                TopicsPattern topicsPattern, List<String> topics) {
             this.topicListService = topicListService;
             this.id = id;
             this.topicsPattern = topicsPattern;
-            this.executor = executor;
             this.matchingTopics = TopicList.filterTopics(topics, topicsPattern);
         }
 
@@ -95,9 +71,6 @@ public class TopicListService {
          */
         @Override
         public void accept(String topicName, NotificationType notificationType) {
-            if (closed) {
-                return;
-            }
             String partitionedTopicName = TopicName.get(topicName).getPartitionedTopicName();
             String domainLessTopicName = TopicList.removeTopicDomainScheme(partitionedTopicName);
 
@@ -114,40 +87,8 @@ public class TopicListService {
                     matchingTopics.add(topicName);
                 }
                 String hash = TopicList.calculateHash(matchingTopics);
-                sendTopicListUpdate(hash, deletedTopics, newTopics);
+                topicListService.sendTopicListUpdate(id, hash, deletedTopics, newTopics);
             }
-        }
-
-        private synchronized void sendTopicListUpdate(String hash, List<String> deletedTopics, List<String> newTopics) {
-            if (closed) {
-                return;
-            }
-            Runnable task = () -> topicListService.sendTopicListUpdate(id, hash, deletedTopics, newTopics);
-            if (sendTopicListSuccessCompleted) {
-                executor.execute(task);
-            } else {
-                // if sendTopicListSuccess hasn't completed, add to a queue to be executed after it completes
-                sendTopicListUpdateTasksBeforeInit.add(task);
-            }
-        }
-
-        @VisibleForTesting
-        synchronized void sendTopicListSuccessCompleted() {
-            if (closed) {
-                sendTopicListUpdateTasksBeforeInit.clear();
-                return;
-            }
-            // Drain all pending sendTopicListUpdate tasks
-            Runnable task;
-            while ((task = sendTopicListUpdateTasksBeforeInit.poll()) != null) {
-                executor.execute(task);
-            }
-            sendTopicListSuccessCompleted = true;
-        }
-
-        public synchronized void close() {
-            closed = true;
-            sendTopicListUpdateTasksBeforeInit.clear();
         }
     }
 
@@ -156,18 +97,15 @@ public class TopicListService {
 
     private final NamespaceService namespaceService;
     private final TopicResources topicResources;
-    private final PulsarService pulsar;
     private final ServerCnx connection;
     private final boolean enableSubscriptionPatternEvaluation;
     private final int maxSubscriptionPatternLength;
     private final ConcurrentLongHashMap<CompletableFuture<TopicListWatcher>> watchers;
-    private final Backoff retryBackoff;
 
 
     public TopicListService(PulsarService pulsar, ServerCnx connection,
                             boolean enableSubscriptionPatternEvaluation, int maxSubscriptionPatternLength) {
         this.namespaceService = pulsar.getNamespaceService();
-        this.pulsar = pulsar;
         this.connection = connection;
         this.enableSubscriptionPatternEvaluation = enableSubscriptionPatternEvaluation;
         this.maxSubscriptionPatternLength = maxSubscriptionPatternLength;
@@ -176,10 +114,6 @@ public class TopicListService {
                 .concurrencyLevel(1)
                 .build();
         this.topicResources = pulsar.getPulsarResources().getTopicResources();
-        this.retryBackoff = new Backoff(
-                100, TimeUnit.MILLISECONDS,
-                25, TimeUnit.SECONDS,
-                0, TimeUnit.MILLISECONDS);
     }
 
     public void inactivate() {
@@ -228,12 +162,34 @@ public class TopicListService {
         CompletableFuture<TopicListWatcher> existingWatcherFuture = watchers.putIfAbsent(watcherId, watcherFuture);
 
         if (existingWatcherFuture != null) {
-            log.info("[{}] Watcher with the same watcherId={} is already created.", connection, watcherId);
-            // use the existing watcher if it's already created
-            watcherFuture = existingWatcherFuture;
+            if (existingWatcherFuture.isDone() && !existingWatcherFuture.isCompletedExceptionally()) {
+                TopicListWatcher watcher = existingWatcherFuture.getNow(null);
+                log.info("[{}] Watcher with the same id is already created:"
+                                + " watcherId={}, watcher={}",
+                        connection.toString(), watcherId, watcher);
+                watcherFuture = existingWatcherFuture;
+            } else {
+                // There was an early request to create a watcher with the same watcherId. This can happen when
+                // client timeout is lower the broker timeouts. We need to wait until the previous watcher
+                // creation request either completes or fails.
+                log.warn("[{}] Watcher with id is already present on the connection,"
+                        + " consumerId={}", connection.toString(), watcherId);
+                ServerError error;
+                if (!existingWatcherFuture.isDone()) {
+                    error = ServerError.ServiceNotReady;
+                } else {
+                    error = ServerError.UnknownError;
+                    watchers.remove(watcherId, existingWatcherFuture);
+                }
+                connection.getCommandSender().sendErrorResponse(requestId, error,
+                        "Topic list watcher is already present on the connection");
+                lookupSemaphore.release();
+                return;
+            }
         } else {
             initializeTopicsListWatcher(watcherFuture, namespaceName, watcherId, topicsPattern);
         }
+
 
         CompletableFuture<TopicListWatcher> finalWatcherFuture = watcherFuture;
         finalWatcherFuture.thenAccept(watcher -> {
@@ -247,8 +203,7 @@ public class TopicListService {
                                 "[{}] Received WatchTopicList for namespace [//{}] by {}",
                                 connection.toString(), namespaceName, requestId);
                     }
-                    sendTopicListSuccessWithPermitAcquiringRetries(watcherId, requestId, topicList, hash,
-                            watcher::sendTopicListSuccessCompleted);
+                    connection.getCommandSender().sendWatchTopicListSuccess(requestId, watcherId, hash, topicList);
                     lookupSemaphore.release();
                 })
                 .exceptionally(ex -> {
@@ -263,99 +218,30 @@ public class TopicListService {
                 });
     }
 
-    private void sendTopicListSuccessWithPermitAcquiringRetries(long watcherId, long requestId, List<String> topicList,
-                                                                String hash,
-                                                                Runnable successfulCompletionCallback) {
-        performOperationWithPermitAcquiringRetries(watcherId, "topic list success", permitAcquireErrorHandler ->
-                () -> connection.getCommandSender()
-                        .sendWatchTopicListSuccess(requestId, watcherId, hash, topicList, permitAcquireErrorHandler)
-                        .whenComplete((__, t) -> {
-                            if (t != null) {
-                                // this is an unexpected case
-                                log.warn("[{}] Failed to send topic list success for watcherId={}. Watcher will be in "
-                                        + "inconsistent state.", connection, watcherId, t);
-                            } else {
-                                // completed successfully, run the callback
-                                successfulCompletionCallback.run();
-                            }
-                        }));
-    }
-
     /***
      * @param topicsPattern The regexp for the topic name(not contains partition suffix).
      */
     public void initializeTopicsListWatcher(CompletableFuture<TopicListWatcher> watcherFuture,
             NamespaceName namespace, long watcherId, TopicsPattern topicsPattern) {
-        BooleanSupplier isPermitRequestCancelled = () -> !connection.isActive() || !watchers.containsKey(watcherId);
-        if (isPermitRequestCancelled.getAsBoolean()) {
-            return;
-        }
-        TopicListSizeResultCache.ResultHolder listSizeHolder = pulsar.getBrokerService().getTopicListSizeResultCache()
-                .getTopicListSize(namespace.toString(), CommandGetTopicsOfNamespace.Mode.PERSISTENT);
-        AsyncDualMemoryLimiter maxTopicListInFlightLimiter = pulsar.getBrokerService().getMaxTopicListInFlightLimiter();
-
-        listSizeHolder.getSizeAsync().thenCompose(initialSize -> {
-            // use heap size limiter to avoid broker getting overwhelmed by a lot of concurrent topic list requests
-            return maxTopicListInFlightLimiter.withAcquiredPermits(initialSize,
-                    AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY, isPermitRequestCancelled, initialPermits -> {
-                        AtomicReference<TopicListWatcher> watcherRef = new AtomicReference<>();
-                        return namespaceService.getListOfPersistentTopics(namespace).thenCompose(topics -> {
-                            long actualSize = TopicListMemoryLimiter.estimateTopicListSize(topics);
-                            listSizeHolder.updateSize(actualSize);
-                            // register watcher immediately so that we don't lose events
-                            TopicListWatcher watcher =
-                                    new TopicListWatcher(this, watcherId, topicsPattern, topics,
-                                            connection.ctx().executor());
-                            watcherRef.set(watcher);
-                            topicResources.registerPersistentTopicListener(namespace, watcher);
-                            // use updated permits to slow down responses so that backpressure gets applied
-                            return maxTopicListInFlightLimiter.withUpdatedPermits(initialPermits, actualSize,
-                                    isPermitRequestCancelled, updatedPermits -> {
-                                        // reset retry backoff
-                                        retryBackoff.reset();
-                                        // just return the watcher which was already created before
-                                        return CompletableFuture.completedFuture(watcher);
-                                    }, CompletableFuture::failedFuture);
-                        }).whenComplete((watcher, exception) -> {
-                            if (exception != null) {
-                                if (watcherRef.get() != null) {
-                                    watcher.close();
-                                    topicResources.deregisterPersistentTopicListener(watcherRef.get());
-                                }
-                                // triggers a retry
-                                throw FutureUtil.wrapToCompletionException(exception);
-                            } else {
-                                if (!watcherFuture.complete(watcher)) {
-                                    log.warn("[{}] Watcher future was already completed. Deregistering "
-                                            + "watcherId={}.", connection, watcherId);
-                                    watcher.close();
-                                    topicResources.deregisterPersistentTopicListener(watcher);
-                                    watchers.remove(watcherId, watcherFuture);
-                                }
-                            }
-                        });
-                    }, CompletableFuture::failedFuture);
-        }).exceptionally(t -> {
-            Throwable unwrappedException = FutureUtil.unwrapCompletionException(t);
-            if (!isPermitRequestCancelled.getAsBoolean() && (
-                    unwrappedException instanceof AsyncSemaphore.PermitAcquireTimeoutException
-                            || unwrappedException instanceof AsyncSemaphore.PermitAcquireQueueFullException)) {
-                // retry with backoff if permit acquisition fails due to timeout or queue full
-                long retryAfterMillis = this.retryBackoff.next();
-                log.info("[{}] {} when initializing topic list watcher watcherId={} for namespace {}. Retrying in {} "
-                                + "ms.", connection, unwrappedException.getMessage(), watcherId, namespace,
-                        retryAfterMillis);
-                connection.ctx().executor()
-                        .schedule(() -> initializeTopicsListWatcher(watcherFuture, namespace, watcherId, topicsPattern),
-                                retryAfterMillis, TimeUnit.MILLISECONDS);
-            } else {
-                log.warn("[{}] Failed to initialize topic list watcher watcherId={} for namespace {}.", connection,
-                        watcherId, namespace, unwrappedException);
-                watcherFuture.completeExceptionally(unwrappedException);
-            }
-            return null;
-        });
+        namespaceService.getListOfPersistentTopics(namespace).
+                thenApply(topics -> {
+                    TopicListWatcher watcher = new TopicListWatcher(this, watcherId, topicsPattern, topics);
+                    topicResources.registerPersistentTopicListener(namespace, watcher);
+                    return watcher;
+                }).
+                whenComplete((watcher, exception) -> {
+                    if (exception != null) {
+                        watcherFuture.completeExceptionally(exception);
+                    } else {
+                        if (!watcherFuture.complete(watcher)) {
+                            log.warn("[{}] Watcher future was already completed. Deregistering watcherId={}.",
+                                    connection.toString(), watcherId);
+                            topicResources.deregisterPersistentTopicListener(watcher);
+                        }
+                    }
+                });
     }
+
 
     public void handleWatchTopicListClose(CommandWatchTopicListClose commandWatchTopicListClose) {
         long requestId = commandWatchTopicListClose.getRequestId();
@@ -365,7 +251,7 @@ public class TopicListService {
     }
 
     public void deleteTopicListWatcher(Long watcherId) {
-        CompletableFuture<TopicListWatcher> watcherFuture = watchers.remove(watcherId);
+        CompletableFuture<TopicListWatcher> watcherFuture = watchers.get(watcherId);
         if (watcherFuture == null) {
             log.info("[{}] TopicListWatcher was not registered on the connection: {}",
                     watcherId, connection.toString());
@@ -379,23 +265,20 @@ public class TopicListService {
             // create operation will complete, the new watcher will be discarded.
             log.info("[{}] Closed watcher before its creation was completed. watcherId={}",
                     connection.toString(), watcherId);
+            watchers.remove(watcherId);
             return;
         }
-
-        // deregister topic listener while avoiding race conditions
-        watcherFuture.whenComplete((watcher, t) -> {
-            if (watcher != null) {
-                topicResources.deregisterPersistentTopicListener(watcher);
-                watcher.close();
-            }
-        });
 
         if (watcherFuture.isCompletedExceptionally()) {
             log.info("[{}] Closed watcher that already failed to be created. watcherId={}",
                     connection.toString(), watcherId);
+            watchers.remove(watcherId);
             return;
         }
 
+        // Proceed with normal watcher close
+        topicResources.deregisterPersistentTopicListener(watcherFuture.getNow(null));
+        watchers.remove(watcherId);
         log.info("[{}] Closed watcher, watcherId={}", connection.toString(), watcherId);
     }
 
@@ -405,67 +288,8 @@ public class TopicListService {
      */
     public void sendTopicListUpdate(long watcherId, String topicsHash, List<String> deletedTopics,
                                     List<String> newTopics) {
-        performOperationWithPermitAcquiringRetries(watcherId, "topic list update", permitAcquireErrorHandler ->
-                () -> connection.getCommandSender()
-                        .sendWatchTopicListUpdate(watcherId, newTopics, deletedTopics, topicsHash,
-                                permitAcquireErrorHandler)
-                        .whenComplete((__, t) -> {
-                            if (t != null) {
-                                // this is an unexpected case
-                                log.warn("[{}] Failed to send topic list update for watcherId={}. Watcher will be in "
-                                        + "inconsistent state.", connection, watcherId, t);
-                            }
-                        }));
+        connection.getCommandSender().sendWatchTopicListUpdate(watcherId, newTopics, deletedTopics, topicsHash);
     }
 
-    // performs an operation with infinite permit acquiring retries.
-    // If acquiring permits fails, it will retry after a backoff period
-    private void performOperationWithPermitAcquiringRetries(long watcherId, String operationName,
-                                                            Function<Consumer<Throwable>,
-                                                                    Supplier<CompletableFuture<Void>>>
-                                                                    asyncOperationFactory) {
-        // holds a reference to the operation, this is to resolve a circular dependency between the error handler and
-        // the actual operation
-        AtomicReference<Runnable> operationRef = new AtomicReference<>();
-        // create the error handler for the operation
-        Consumer<Throwable> permitAcquireErrorHandler =
-                createPermitAcquireErrorHandler(watcherId, operationName, operationRef);
-        // create the async operation using the factory function. Pass the error handler to the factory function.
-        Supplier<CompletableFuture<Void>> asyncOperation = asyncOperationFactory.apply(permitAcquireErrorHandler);
-        // set the operation to run into the operation reference
-        operationRef.set(Runnables.catchingAndLoggingThrowables(() -> {
-            if (!connection.isActive() || !watchers.containsKey(watcherId)) {
-                // do nothing if the connection has already been closed or the watcher has been removed
-                return;
-            }
-            asyncOperation.get().thenRun(() -> retryBackoff.reset());
-        }));
-        // run the operation
-        operationRef.get().run();
-    }
 
-    // retries acquiring permits until the connection is closed or the watcher is removed
-    private Consumer<Throwable> createPermitAcquireErrorHandler(long watcherId, String operationName,
-                                                                AtomicReference<Runnable> operationRef) {
-        ScheduledExecutorService scheduledExecutor = connection.ctx().channel().eventLoop();
-        AtomicInteger retryCount = new AtomicInteger(0);
-        return t -> {
-            Throwable unwrappedException = FutureUtil.unwrapCompletionException(t);
-            if (unwrappedException instanceof AsyncSemaphore.PermitAcquireCancelledException
-                    || unwrappedException instanceof AsyncSemaphore.PermitAcquireAlreadyClosedException
-                    || !connection.isActive()
-                    || !watchers.containsKey(watcherId)) {
-                return;
-            }
-            long retryDelay = retryBackoff.next();
-            log.info("[{}] Cannot acquire direct memory tokens for sending {}. Retry {} in {} ms. {}", connection,
-                    operationName, retryCount.get(), retryDelay, t.getMessage());
-            scheduledExecutor.schedule(operationRef.get(), retryDelay, TimeUnit.MILLISECONDS);
-        };
-    }
-
-    @VisibleForTesting
-    CompletableFuture<TopicListWatcher> getWatcherFuture(long watcherId) {
-        return watchers.get(watcherId);
-    }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PulsarCommandSender.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PulsarCommandSender.java
@@ -96,11 +96,8 @@ public interface PulsarCommandSender {
 
     void sendEndTxnErrorResponse(long requestId, TxnID txnID, ServerError error, String message);
 
-    CompletableFuture<Void> sendWatchTopicListSuccess(long requestId, long watcherId, String topicsHash,
-                                                      List<String> topics,
-                                                      Consumer<Throwable> permitAcquireErrorHandler);
+    void sendWatchTopicListSuccess(long requestId, long watcherId, String topicsHash, List<String> topics);
 
-    CompletableFuture<Void> sendWatchTopicListUpdate(long watcherId,
-                                  List<String> newTopics, List<String> deletedTopics, String topicsHash,
-                                  Consumer<Throwable> permitAcquireErrorHandler);
+    void sendWatchTopicListUpdate(long watcherId,
+                                         List<String> newTopics, List<String> deletedTopics, String topicsHash);
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PulsarCommandSenderImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PulsarCommandSenderImpl.java
@@ -366,30 +366,27 @@ public class PulsarCommandSenderImpl implements PulsarCommandSender {
 
     /***
      * @param topics topic names which are matching, the topic name contains the partition suffix.
-     * @return
      */
     @Override
-    public CompletableFuture<Void> sendWatchTopicListSuccess(long requestId, long watcherId, String topicsHash,
-                                                             List<String> topics,
-                                                             Consumer<Throwable> permitAcquireErrorHandler) {
+    public void sendWatchTopicListSuccess(long requestId, long watcherId, String topicsHash, List<String> topics) {
         BaseCommand command = Commands.newWatchTopicListSuccess(requestId, watcherId, topicsHash, topics);
-        safeIntercept(command, cnx);
-        return acquireDirectMemoryPermitsAndWriteAndFlush(cnx.ctx(), maxTopicListInFlightLimiter, () -> !cnx.isActive(),
-                command, permitAcquireErrorHandler);
+        interceptAndWriteCommand(command);
     }
 
     /***
      * {@inheritDoc}
-     * @return
      */
     @Override
-    public CompletableFuture<Void> sendWatchTopicListUpdate(long watcherId, List<String> newTopics,
-                                                            List<String> deletedTopics, String topicsHash,
-                                                            Consumer<Throwable> permitAcquireErrorHandler) {
+    public void sendWatchTopicListUpdate(long watcherId,
+                                         List<String> newTopics, List<String> deletedTopics, String topicsHash) {
         BaseCommand command = Commands.newWatchTopicUpdate(watcherId, newTopics, deletedTopics, topicsHash);
+        interceptAndWriteCommand(command);
+    }
+
+    private void interceptAndWriteCommand(BaseCommand command) {
         safeIntercept(command, cnx);
-        return acquireDirectMemoryPermitsAndWriteAndFlush(cnx.ctx(), maxTopicListInFlightLimiter, () -> !cnx.isActive(),
-                command, permitAcquireErrorHandler);
+        ByteBuf outBuf = Commands.serializeWithSize(command);
+        writeAndFlush(outBuf);
     }
 
     private void writeAndFlush(ByteBuf outBuf) {
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/TopicListServiceTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/TopicListServiceTest.java
@@ -18,62 +18,31 @@
  */
 package org.apache.pulsar.broker.service;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.anyLong;
-import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.timeout;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
-import io.netty.channel.Channel;
-import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.EventLoop;
-import io.netty.util.concurrent.EventExecutor;
-import io.netty.util.concurrent.ImmediateEventExecutor;
-import io.netty.util.concurrent.ScheduledFuture;
 import java.net.InetSocketAddress;
-import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Consumer;
-import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.collections4.ListUtils;
 import org.apache.pulsar.broker.PulsarServerException;
 import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.namespace.NamespaceService;
 import org.apache.pulsar.broker.resources.PulsarResources;
 import org.apache.pulsar.broker.resources.TopicResources;
-import org.apache.pulsar.broker.topiclistlimit.TopicListSizeResultCache;
 import org.apache.pulsar.common.api.proto.CommandWatchTopicListClose;
 import org.apache.pulsar.common.api.proto.ServerError;
 import org.apache.pulsar.common.naming.NamespaceName;
-import org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiter;
-import org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiterImpl;
-import org.apache.pulsar.common.semaphore.AsyncSemaphore;
 import org.apache.pulsar.common.topics.TopicList;
 import org.apache.pulsar.common.topics.TopicsPattern;
-import org.apache.pulsar.metadata.api.MetadataStore;
-import org.apache.pulsar.metadata.api.Notification;
-import org.apache.pulsar.metadata.api.NotificationType;
-import org.awaitility.Awaitility;
 import org.testng.Assert;
-import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
-@Slf4j
 public class TopicListServiceTest {
 
     private TopicListService topicListService;
@@ -83,25 +52,13 @@ public class TopicListServiceTest {
     private TopicResources topicResources;
     private final TopicsPattern.RegexImplementation topicsPatternImplementation =
             TopicsPattern.RegexImplementation.RE2J_WITH_JDK_FALLBACK;
-    private EventLoop eventLoop;
-    private PulsarCommandSender pulsarCommandSender;
-    private Consumer<Notification> notificationConsumer;
-    private AsyncDualMemoryLimiterImpl memoryLimiter;
 
     @BeforeMethod(alwaysRun = true)
     public void setup() throws Exception {
         lookupSemaphore = new Semaphore(1);
         lookupSemaphore.acquire();
         topicListFuture = new CompletableFuture<>();
-
-        AtomicReference<Consumer<Notification>> listenerRef = new AtomicReference<>();
-        MetadataStore metadataStore = mock(MetadataStore.class);
-        doAnswer(invocationOnMock -> {
-            listenerRef.set(invocationOnMock.getArgument(0));
-            return null;
-        }).when(metadataStore).registerListener(any());
-        topicResources = spy(new TopicResources(metadataStore));
-        notificationConsumer = listenerRef.get();
+        topicResources = mock(TopicResources.class);
 
         PulsarService pulsar = mock(PulsarService.class);
         when(pulsar.getNamespaceService()).thenReturn(mock(NamespaceService.class));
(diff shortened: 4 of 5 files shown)


pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncDualMemoryLimiterUtil.java
```diff
@@ -0,0 +1,111 @@
...
+                        ctx.writeAndFlush(outBuf).addListener(future -> {
+                            // Release permits after the response has been written to the socket
+                            dualMemoryLimiter.release(permits);
+                        });
+                    } catch (Exception e) {
+                        // Return permits if an exception occurs before writeAndFlush is called successfully
+                        dualMemoryLimiter.release(permits);
+                        throw e;
+                    }
+                }).thenAccept(__ -> {
```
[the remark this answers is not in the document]
lhotari: The purpose is to map to CompletableFuture<Void>

org/apache/pulsar/common/util/memory/AsyncSemaphore.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.apache.pulsar.common.util.memory;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Interface for asynchronous semaphore operations.
+ * Used for managing concurrent access to limited resources.
+ */
+public interface AsyncSemaphore {
```
Denovo1998: Are the five files in this directory redundant?

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    CompletableFuture<AsyncSemaphorePermit> acquire(long permits, BooleanSupplier isCancelled);
+
+    /**
+     * Acquire or release permits for previously acquired permits by updating the permits.
+     * Returns a future that completes when permits are available.
+     * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
+     * and exceptionally with AsyncSemaphorePermitAcquireQueueFullException when queue full
+     * @return CompletableFuture that completes with permit when available
+     */
```
Denovo1998, line 52: This should be synchronized with the modifications in the PIP documentation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2606,6 +2577,77 @@ protected void handleGetTopicsOfNamespace(CommandGetTopicsOfNamespace commandGet
...
+                                        BrokerServiceException.getClientErrorCode(new ServerMetadataException(ex)),
+                                        ex.getMessage());
+                                return null;
+                            });
+                }, t -> {
+                    log.warn("[{}] Failed to acquire initial heap memory permits for GetTopicsOfNamespace: {}",
+                            remoteAddress, t.getMessage());
+                    writeAndFlush(Commands.newError(requestId, ServerError.TooManyRequests,
+                            "Failed due to heap memory limit exceeded"));
+                    return CompletableFuture.completedFuture(null);
```
Denovo1998: Should we call `lookupSemaphore.release();` here?

org/apache/pulsar/common/util/memory/AsyncSemaphore.java
lhotari: No, what do you mean?

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java
lhotari: The PIP document has been updated to match this implementation. What's missing?

org/apache/pulsar/common/util/memory/AsyncSemaphore.java
lhotari: Oh yes, now I realized the problem. Deleted the files.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: yes, added.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,296 @@
...
+
+        this.directWaitTimeMs = register(Summary.build(prometheusPrefix + "topic_list_direct_wait_time_ms",
+                        "Wait time for direct memory permits")
+                .quantile(0.50, 0.01)
+                .quantile(0.95, 0.01)
+                .quantile(0.99, 0.01)
+                .quantile(1, 0.01)
+                .create());
+        this.otelDirectWaitTime = openTelemetryMeter.histogramBuilder("topic.list.direct.wait.time.ms")
+                .setUnit("s")
```
Denovo1998: The Prometheus metric name uses _wait_time_ms (milliseconds) but the OpenTelemetry histogram unit is set to "s" (seconds).
While the actual recording in recordDirectWaitTime correctly converts to seconds, the metric name .wait.time.ms suggests milliseconds, which is confusing.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,296 @@
...
+                        "Maximum heap memory limiter queue size")
+                .create());
+        this.heapQueueMaxSize.set(maxHeapQueueSize);
+        LongGauge otelHeapQueueMaxSize = openTelemetryMeter
+                .gaugeBuilder("topic.list.heap.queue.max.size")
+                .setDescription("Maximum heap memory limiter queue size")
+                .setUnit("1")
+                .ofLongs()
+                .build();
+        otelHeapQueueMaxSize.set(maxHeapQueueSize);
```
Denovo1998, line 160: The `otelHeapQueueMaxSize` gauge is created as a local variable and cannot be closed in the `close()` method, potentially causing a resource leak.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,296 @@
...
+                        "Maximum direct memory limiter queue size")
+                .create());
+        this.directQueueMaxSize.set(maxDirectQueueSize);
+        LongGauge otelDirectQueueMaxSize = openTelemetryMeter
+                .gaugeBuilder("topic.list.direct.queue.max.size")
+                .setDescription("Maximum direct memory limiter queue size")
+                .setUnit("1")
+                .ofLongs()
+                .build();
+        otelDirectQueueMaxSize.set(maxDirectQueueSize);
```
Denovo1998, line 189: The `otelDirectQueueMaxSize` gauge is created as a local variable and cannot be closed in the `close()` method, potentially causing a resource leak.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,296 @@
...
+        unregister(heapQueueMaxSize);
+        unregister(directQueueSize);
+        unregister(directQueueMaxSize);
+        unregister(heapWaitTimeMs);
+        unregister(directWaitTimeMs);
+        unregister(heapTimeoutTotal);
+        unregister(directTimeoutTotal);
+        otelHeapMemoryUsedGauge.close();
+        otelDirectMemoryUsedGauge.close();
+        otelHeapQueueSize.close();
```
Denovo1998, line 295: Close all OpenTelemetry instruments to prevent resource leaks.
```java
otelHeapMemoryLimitGauge.close();
otelDirectMemoryLimitGauge.close();
otelHeapQueueMaxSize.close();
otelDirectQueueMaxSize.close();
otelHeapWaitTime.close();
otelDirectWaitTime.close();
otelHeapTimeoutTotal.close();
otelDirectTimeoutTotal.close();
```

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        CompletableFuture<AsyncSemaphorePermit> future = new CompletableFuture<>();
+
+        if (closed.get()) {
+            future.completeExceptionally(new PermitAcquireAlreadyClosedException("Semaphore is closed"));
+            return future;
+        }
+
+        if (queue.size() >= maxQueueSize) {
```
Denovo1998: The size() method on `ConcurrentLinkedQueue` iterates through all elements, making it an `O(n)` operation.
Maintaining a separate AtomicInteger counter to track queue size in O(1) time?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -288,8 +294,13 @@ public void deleteTopicListWatcher(Long watcherId) {
...
                                     List<String> newTopics) {
-        connection.getCommandSender().sendWatchTopicListUpdate(watcherId, newTopics, deletedTopics, topicsHash);
+        connection.getCommandSender().sendWatchTopicListUpdate(watcherId, newTopics, deletedTopics, topicsHash,
+                t -> {
+                    // TODO add retry with backoff
+                    log.warn(
+                            "[{}] Cannot acquire direct memory tokens for sending topic list update. State will be "
+                                    + "inconsistent on the client. {}",
+                            connection.toString(), t.getMessage());
+                });
```
Denovo1998: Extracting the error handling into a reusable method.
```java
private void handleSendFailure(String operation, Throwable t, long watcherId) {
    // TODO: add retry with backoff
    log.warn("[{}] Cannot acquire direct memory tokens for sending topic list update. State will be "
                                    + "inconsistent on the client. {}",
                            connection.toString(), t.getMessage());
}

...
// Line 206
connection.getCommandSender().sendWatchTopicListSuccess(requestId, watcherId, hash, topicList,
        t -> handleSendFailure("topic list success", t, watcherId));
...
// Line 297
connection.getCommandSender().sendWatchTopicListUpdate(watcherId, newTopics, deletedTopics, topicsHash,
        t -> handleSendFailure("topic list update", t, watcherId));
...
```

pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerBackPressureMultipleConsumersTest.java
```diff
@@ -0,0 +1,205 @@
...
+                                    log.error("Failed to get topic list.", ex);
+                                }
+                                log.info("latch-count: {}, succeed: {}, available direct mem: {} MB", latch.getCount(),
+                                        success.get(),
+                                        (DirectMemoryUtils.jvmMaxDirectMemory() - JvmMetrics.getJvmDirectMemoryUsed())
+                                                / (1024 * 1024));
+                                latch.countDown();
+                            });
+                } catch (Exception e) {
+                    semaphore.release();
```
Denovo1998, line 152: Should add this?
```java
latch.countDown();
log.error("Failed to execute getTopicsUnderNamespace request.", e);
```

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.function.BooleanSupplier;
+
+/**
+ * An abstraction for a generic asynchronous semaphore.
+ */
+public interface AsyncSemaphore {
+    /**
+     * Acquire permits from the semaphore.
+     * Returned future completes when permits are available.
+     * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireTimeoutException`

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java
```diff
@@ -0,0 +1,127 @@
...
+     * @param permits     number of permits to acquire
+     * @param isCancelled supplier that returns true if acquisition should be cancelled
+     * @return CompletableFuture that completes with permit when available
+     */
+    CompletableFuture<AsyncSemaphorePermit> acquire(long permits, BooleanSupplier isCancelled);
+
+    /**
+     * Acquire or release permits for previously acquired permits by updating the permits.
+     * Returns a future that completes when permits are available.
+     * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireTimeoutException`

pip/pip-442.md
```diff
@@ -132,24 +134,48 @@ public interface AsyncSemaphore {
      * Returned future completes when permits are available.
      * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireTimeoutException`

pip/pip-442.md
```diff
@@ -132,24 +134,48 @@ public interface AsyncSemaphore {
...
+     * @param isCancelled supplier that returns true if acquisition should be cancelled
      * @return CompletableFuture that completes with permit when available
      */
-    CompletableFuture<AsyncSemaphorePermitResult> acquire(long permits);
+    CompletableFuture<AsyncSemaphorePermit> acquire(long permits, BooleanSupplier isCancelled);
 
     /**
      * Acquire or release permits for previously acquired permits by updating the permits.
      * Returns a future that completes when permits are available.
      * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireTimeoutException`

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * An abstraction for a generic asynchronous semaphore.
+ */
+public interface AsyncSemaphore {
+    /**
+     * Acquire permits from the semaphore.
+     * Returned future completes when permits are available.
+     * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
+     * and exceptionally with AsyncSemaphorePermitAcquireQueueFullException when queue full
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireQueueFullException`

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java
```diff
@@ -0,0 +1,127 @@
...
+     * @param isCancelled supplier that returns true if acquisition should be cancelled
+     * @return CompletableFuture that completes with permit when available
+     */
+    CompletableFuture<AsyncSemaphorePermit> acquire(long permits, BooleanSupplier isCancelled);
+
+    /**
+     * Acquire or release permits for previously acquired permits by updating the permits.
+     * Returns a future that completes when permits are available.
+     * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
+     * and exceptionally with AsyncSemaphorePermitAcquireQueueFullException when queue full
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireQueueFullException`

pip/pip-442.md
```diff
@@ -132,24 +134,48 @@ public interface AsyncSemaphore {
...
      * and exceptionally with AsyncSemaphorePermitAcquireQueueFullException when queue full
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireQueueFullException`

pip/pip-442.md
```diff
@@ -132,24 +134,48 @@ public interface AsyncSemaphore {
      * Returned future completes when permits are available.
      * It will complete exceptionally with AsyncSemaphorePermitAcquireTimeoutException on timeout
      * and exceptionally with AsyncSemaphorePermitAcquireQueueFullException when queue full
```
Denovo1998: typo: `AsyncSemaphore.PermitAcquireQueueFullException`

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
```diff
@@ -357,6 +357,59 @@ private void performGetTopicsOfNamespace(long clientRequestId,
...
+                                                                   BooleanSupplier isCancelled) {
+        BaseCommand responseCommand = Commands.newGetTopicsOfNamespaceResponseCommand(
+                r.getNonPartitionedOrPartitionTopics(), r.getTopicsHash(), r.isFiltered(),
+                r.isChanged(), clientRequestId);
+        return acquireDirectMemoryPermitsAndWriteAndFlush(proxyConnection.ctx(), maxTopicListInFlightLimiter,
+                isCancelled, responseCommand, t -> {
+                    log.warn("[{}] Failed to acquire actual direct memory permits for GetTopicsOfNamespace: {}",
+                            clientAddress, t.getMessage());
+                    writeAndFlush(Commands.newError(clientRequestId, ServerError.TooManyRequests,
+                            "Failed due to heap memory limit exceeded"));
```
Denovo1998: "Failed due to direct memory limit exceeded"));

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
```diff
@@ -357,6 +357,59 @@ private void performGetTopicsOfNamespace(long clientRequestId,
...
+        maxTopicListInFlightLimiter.withAcquiredPermits(INITIAL_TOPIC_LIST_HEAP_PERMITS_SIZE,
+                AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY, isPermitRequestCancelled, initialPermits -> {
+                    return clientCnx.newGetTopicsOfNamespace(command, requestId).whenComplete((r, t) -> {
+                        if (t != null) {
+                            log.warn("[{}] Failed to get TopicsOfNamespace {}: {}", clientAddress, namespaceName,
+                                    t.getMessage());
+                            writeAndFlush(Commands.newError(clientRequestId, getServerError(t), t.getMessage()));
+                        } else {
+                            long actualSize =
+                                    r.getNonPartitionedOrPartitionTopics().stream().mapToInt(String::length).sum();
```
Denovo1998: Java strings are UTF-16, so 2 bytes per character?

pip/pip-442.md
```diff
@@ -163,227 +189,409 @@ public interface AsyncDualMemoryLimiter {
...
                     // Serialize the response
                     ByteBuf outBuf = Commands.serializeWithPrecalculatedSerializedSize(command, serializedSize);
                     // Write the response
-                    cnx.ctx().writeAndFlush(outBuf).addListener(future -> {
+                    ctx.writeAndFlush(outBuf).addListener(future -> {
                         // Release permits after the response has been written to the socket
-                        maxTopicListInFlightLimiter.release(permits);
+                        dualMemoryLimiter.release(permits);
                     });
                 } catch (Exception e) {
```
Denovo1998: The `withPermitsFuture` method catches `Throwable (line 71)` to ensure permits are released even for Error subclasses.

```java
} catch (Throwable e) {
```

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
lhotari: there are no such methods on Otel instruments that aren't observable.

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: I'll replace the queue implementation with ArrayBlockingQueue and use `offer` to add to the queue. (it returns false if the queue size limit has been reached)

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: I'm handling this with changes to address the retry with backoff which is currently a TODO. I'll push the changes very soon.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
lhotari: Thanks for pointing this out. I'll remove the ".ms" suffix which is wrong as you have pointed out. In Otel, the metrics are in seconds. That's part of the specification.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
lhotari: there is no close method in Otel for instruments that aren't "observable".

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
lhotari: there is no close method in Otel for instruments that aren't "observable".

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: PTAL

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
lhotari: My assumption is that String are encoded as UTF-8 in the heap. There could be multi-byte characters which make the byte count larger than the character count. I've revisited the calculation to use byte count for UTF-8 encoding of characters.
I also added 32 byte overhead for each entry in the calculation.

The accuracy isn't that relevant in this case, since the limiter will introduce backpressure. The limit should be set to a low enough value so that actual resources don't get exhausted when the limit is reached.

pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerBackPressureMultipleConsumersTest.java
lhotari: done

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
```diff
@@ -357,6 +358,62 @@ private void performGetTopicsOfNamespace(long clientRequestId,
...
+                    return clientCnx.newGetTopicsOfNamespace(command, requestId).whenComplete((r, t) -> {
+                        if (t != null) {
+                            log.warn("[{}] Failed to get TopicsOfNamespace {}: {}", clientAddress, namespaceName,
+                                    t.getMessage());
+                            writeAndFlush(Commands.newError(clientRequestId, getServerError(t), t.getMessage()));
+                        } else {
+                            long actualSize =
+                                    r.getNonPartitionedOrPartitionTopics().stream()
+                                            .mapToInt(ByteBufUtil::utf8Bytes) // convert character count to bytes
+                                            .map(n -> n + 32) // add 32 bytes overhead for each entry
```
Denovo1998: Can be changed to
```java
.mapToLong(topic -> ByteBufUtil.utf8Bytes(topic) + 32) // UTF-8 bytes + 32 bytes overhead per entry
```

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    private void processQueue() {
+        if (closed.get()) {
+            return;
+        }
+        executor.execute(processQueueRunnable);
+    }
+
+    private void internalProcessQueue() {
```
Denovo1998: Is there a race condition here?
- Line 187: current = 10
- Request needs 8 permits
- Another thread calls update(), reducing availablePermits to 3
- Line 205: Check passes (8 <= 10 using stale current)
- Line 206: availablePermits = 3 - 8 = -5 (negative!)
This violates the semaphore invariant, allowing more permits to be acquired than maxPermits.

pip/pip-442.md
```diff
@@ -163,227 +189,409 @@ public interface AsyncDualMemoryLimiter {
...
+                                                               BooleanSupplier isCancelled) {
+    BaseCommand responseCommand = Commands.newGetTopicsOfNamespaceResponseCommand(
+            r.getNonPartitionedOrPartitionTopics(), r.getTopicsHash(), r.isFiltered(),
+            r.isChanged(), clientRequestId);
+    return acquireDirectMemoryPermitsAndWriteAndFlush(proxyConnection.ctx(), maxTopicListInFlightLimiter,
+            isCancelled, responseCommand, t -> {
+                log.warn("[{}] Failed to acquire actual direct memory permits for GetTopicsOfNamespace: {}",
+                        clientAddress, t.getMessage());
+                writeAndFlush(Commands.newError(clientRequestId, ServerError.TooManyRequests,
+                        "Failed due to heap memory limit exceeded"));
```
Denovo1998:
```java
"Failed due to direct memory limit exceeded"
```

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
lhotari: Sure, there's multiple ways to use the Streams API. I don't see why that would be any better.

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: This method is always run in a single thread by a single threaded executor. I guess it would need to be documented that the executor needs to be single threaded.

Denovo1998: Oh! I understand. This does indeed need to be recorded.

lhotari: I've documented the requirement of a single threaded executor and made the method synchronized to avoid any future confusion in maintaining the code

lhotari:
To properly handle a large amount of requests, it was necessary to add a solution for keeping a cache of the estimated size of the topic listing. Without the solution heap memory would run out in the ProxyPatternConsumerBackPressureMultipleConsumersTest test after increasing concurrency. While debugging, I noticed that this happened due to the low initial permits. Since a lot of inflight requests were accepted, it caused heap memory to grow very quickly. The solution is explained [in the javadoc of TopicListSizeResultCache](https://github.com/lhotari/pulsar/blob/lh-implement-PIP-442/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java). It lets the first request to proceed with the initial permits and concurrent requests will wait until the first request updates the topic list size. This will ensure that all subsequent requests are using a sane value.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,296 @@
...
+        this.directQueueMaxSize.set(maxDirectQueueSize);
+        LongGauge otelDirectQueueMaxSize = openTelemetryMeter
+                .gaugeBuilder("topic.list.direct.queue.max.size")
+                .setDescription("Maximum direct memory limiter queue size")
+                .setUnit("1")
+                .ofLongs()
+                .build();
+        otelDirectQueueMaxSize.set(maxDirectQueueSize);
+
+        this.heapWaitTimeMs = register(Summary.build(prometheusPrefix + "topic_list_heap_wait_time_ms",
```
codelipenghui, line 191: Based on https://github.com/apache/pulsar/pull/21080, we should change to use a histogram instead of a summary.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,296 @@
...
+                .quantile(0.95, 0.01)
+                .quantile(0.99, 0.01)
+                .quantile(1, 0.01)
+                .create());
+        this.otelHeapWaitTime = openTelemetryMeter.histogramBuilder("topic.list.heap.wait.time")
+                .setUnit("s")
+                .setDescription("Wait time for heap memory permits")
+                .build();
+
+        this.directWaitTimeMs = register(Summary.build(prometheusPrefix + "topic_list_direct_wait_time_ms",
```
codelipenghui, line 203: Same as the above comment

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2610,6 +2581,99 @@ protected void handleGetTopicsOfNamespace(CommandGetTopicsOfNamespace commandGet
...
+                listSizeHolder = service.getTopicListSizeResultCache().getTopicListSize(namespaceName.toString(), mode);
+        listSizeHolder.getSizeAsync().thenAccept(initialSize -> {
+            maxTopicListInFlightLimiter.withAcquiredPermits(initialSize,
+                    AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY, isPermitRequestCancelled, initialPermits -> {
+                        return getBrokerService().pulsar().getNamespaceService()
+                                .getListOfUserTopics(namespaceName, mode)
+                                .thenAccept(topics -> {
+                                    long actualSize = topics.stream()
+                                            .mapToInt(ByteBufUtil::utf8Bytes) // convert character count to bytes
+                                            .map(n -> n + 32) // add 32 bytes overhead for each entry
```
codelipenghui: Is there any reason to add extra 32 bytes?

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
```diff
@@ -357,6 +359,76 @@ private void performGetTopicsOfNamespace(long clientRequestId,
...
+                            if (t != null) {
+                                log.warn("[{}] Failed to get TopicsOfNamespace {}: {}", clientAddress, namespaceName,
+                                        t.getMessage());
+                                listSizeHolder.resetIfInitializing();
+                                writeAndFlush(Commands.newError(clientRequestId, getServerError(t), t.getMessage()));
+                            } else {
+                                long actualSize =
+                                        r.getNonPartitionedOrPartitionTopics().stream()
+                                                .mapToInt(ByteBufUtil::utf8Bytes) // convert character count to bytes
+                                                .map(n -> n + 32) // add 32 bytes overhead for each entry
```
codelipenghui: Same as the above comment

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
lhotari: I assume you are referencing to ["Summary is widely used, but not aggregatable across brokers / labels"](https://github.com/apache/pulsar/blob/master/pip/pip-264.md#summary-is-widely-used-but-not-aggregatable-across-brokers--labels). In this case, it's a Prometheus metric and we have been using summaries. I think that there's only very few locations where Prometheus histograms are used in the Pulsar code base. IIRC, it's only for a few metadata store metrics.
Just wondering if we should just keep on using a summary in this case since there's not really a need to aggregate this metric. 

For the Open Telemetry metric `topic.list.heap.wait.time`, `io.opentelemetry.api.metrics.DoubleHistogram` is already used.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: In JVMs, an object header typically consumes 16 bytes and a reference consumes 4 bytes or 8 bytes. This is improve the accuracy of the estimation of consumed heap. I guess there's also padding involved. 
I think that there's at least 20 bytes of overhead per entry. 

One detail here is that the memory limiter doesn't do an accurate calculation about the actual consumed heap. The purpose is to have a way to ensure that there is a limit for heap memory. In many cases the referenced objects are shared and therefore they won't duplicate the usage of heap memory. It would be possible to address that, but it would unnecessarily complicate the solution.

lhotari: I also realized that java.lang.String contains several fields that would be consuming some amount of bytes.

lhotari: I have refactored the code and moved the calculation to a separate utility method in org.apache.pulsar.broker.topiclistlimit.TopicListMemoryLimiter#estimateTopicListSize . That contains a comment about the 16 byte object header and additional overhead of java.lang.String fields.

lhotari:
I noticed a separate heap memory issue related to topic listing operations: the childrenCache in AbstractMetadataStore. There's a separate PR to address the unbounded cache: https://github.com/apache/pulsar/pull/24868.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
```diff
@@ -0,0 +1,137 @@
...
+    record CacheKey(String namespaceName, CommandGetTopicsOfNamespace.Mode mode) {
+    }
+
+    /**
+     * Holds the topic list size estimate and future for the topic list size.
+     * The size is returned by calling {@link #getSizeAsync()} method which is asynchronous.
+     * The estimate is updated by calling {@link #updateSize(long)} method.
+     */
+    public static class ResultHolder {
+        private final AtomicReference<CompletableFuture<Long>> topicListSizeFuture =
```
poorbarcode, line 55: The value of the topic list size seems never to be modified asynchronously, so the type being CompletableFuture is unnecessary.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
```diff
@@ -0,0 +1,137 @@
...
+                new AtomicReference<>(null);
+        private final AtomicLong existingSizeRef = new AtomicLong(-1L);
```
poorbarcode: Since the value of the topic list size seems never to be modified asynchronously, the unnecessary field `topicListSizeFuture` can be removed.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+
+        /**
+         * Update the topic list size estimate. The new estimated size will be updated by calculating the average
+         * of the existing and the new size. If the difference between the new and the existing size is less than 1,
+         * no update will be done.
+         * @param actualSize the actual size of the topic list
+         */
+        public void updateSize(long actualSize) {
```
poorbarcode, line 81: We need to mention that the current modifying operation may be dropped at line 103

```java
if (currentFuture != null && !currentFuture.isDone()) {
    currentFuture.complete(existingSizeValue);
} else if (currentFuture == null || currentFuture.join().longValue() != existingSizeValue) {
    topicListSizeFuture.compareAndSet(currentFuture, CompletableFuture.completedFuture(existingSizeValue));
} else {
    // The operation may be dropped here
}
```

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    @Override
+    public CompletableFuture<AsyncSemaphorePermit> acquire(long permits, BooleanSupplier isCancelled) {
+        return internalAcquire(permits, permits, isCancelled);
+    }
+
+    private CompletableFuture<AsyncSemaphorePermit> internalAcquire(long permits, long acquirePermits,
+                                                                    BooleanSupplier isCancelled) {
+        if (!isPermitsValidForAcquiring(permits)) {
```
poorbarcode: We'd better separate the following two cases:
- `permits <0`: it may be caused by a bug
- `permits > maxPermits`: users should reset the configuration

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2610,6 +2581,93 @@ protected void handleGetTopicsOfNamespace(CommandGetTopicsOfNamespace commandGet
...
+                listSizeHolder = service.getTopicListSizeResultCache().getTopicListSize(namespaceName.toString(), mode);
+        listSizeHolder.getSizeAsync().thenAccept(initialSize -> {
+            maxTopicListInFlightLimiter.withAcquiredPermits(initialSize,
+                    AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY, isPermitRequestCancelled, initialPermits -> {
+                        return getBrokerService().pulsar().getNamespaceService()
+                                .getListOfUserTopics(namespaceName, mode)
+                                .thenAccept(topics -> {
+                                    long actualSize = TopicListMemoryLimiter.estimateTopicListSize(topics);
+                                    listSizeHolder.updateSize(actualSize);
+                                    maxTopicListInFlightLimiter.withUpdatedPermits(initialPermits, actualSize,
```
poorbarcode, line 2599: Instead of updating and acquiring additional permits if the actual size is larger than estimated, we'd better improve as follows:
- The current request does not need to acquire permits again; only updating the permits that were borrowed out is enough, even though the permits were over-acquired
- Updates the permits that were borrowed out, to limit the following permits acquisition;

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2610,6 +2581,93 @@ protected void handleGetTopicsOfNamespace(CommandGetTopicsOfNamespace commandGet
...
+    private void internalHandleGetTopicsOfNamespace(String namespace, NamespaceName namespaceName, long requestId,
+                                                    CommandGetTopicsOfNamespace.Mode mode,
+                                                    Optional<String> topicsPattern, Optional<String> topicsHash,
+                                                    Semaphore lookupSemaphore) {
+        BooleanSupplier isPermitRequestCancelled = () -> !ctx().channel().isActive();
+        TopicListSizeResultCache.ResultHolder
+                listSizeHolder = service.getTopicListSizeResultCache().getTopicListSize(namespaceName.toString(), mode);
+        listSizeHolder.getSizeAsync().thenAccept(initialSize -> {
+            maxTopicListInFlightLimiter.withAcquiredPermits(initialSize,
+                    AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY, isPermitRequestCancelled, initialPermits -> {
```
poorbarcode, line 2593: Just add a tip: if the data does not exist in the cache, it may query ZK, which will also use direct memory

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -165,8 +173,66 @@ protected CompletableFuture<Void> internalCreateNamespace(Policies policies) {
...
+                AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY, isPermitRequestCancelled).exceptionally(t -> {
+            throw new CompletionException(
+                    new RestException(Status.TOO_MANY_REQUESTS, "Failed due to heap memory limit exceeded"));
+        }).thenCompose(initialPermits -> {
+            initialPermitsRef.set(initialPermits);
+            // perform the actual get list of topics operation
+            return doInternalGetListOfTopics(policies, mode).thenCompose(topicList -> {
+                long actualSize = TopicListMemoryLimiter.estimateTopicListSize(topicList);
+                listSizeHolder.updateSize(actualSize);
+                return maxTopicListInFlightLimiter.update(initialPermits, actualSize, isPermitRequestCancelled)
```
poorbarcode, line 219: The same comment as https://github.com/apache/pulsar/pull/24833#discussion_r2444987116

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -101,6 +112,7 @@ public void accept(String topicName, NotificationType notificationType) {
     private final boolean enableSubscriptionPatternEvaluation;
     private final int maxSubscriptionPatternLength;
     private final ConcurrentLongHashMap<CompletableFuture<TopicListWatcher>> watchers;
+    private final Backoff retryBackoff;
 
 
     public TopicListService(PulsarService pulsar, ServerCnx connection,
```
poorbarcode, line 106: This function is a notification for the client, to let the client know which topic was changed; it should not be limited by the new feature. the change https://github.com/apache/pulsar/pull/24833/files#diff-5cb94e51fda5445d6fc829fda24eeccda65bf166643a7bf6957fa195516bca28L363-R392 is also related

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,312 @@
...
+import java.util.concurrent.TimeUnit;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiterImpl;
+import org.apache.pulsar.common.semaphore.AsyncSemaphore;
+
+/**
+ * Topic list memory limiter that exposes both Prometheus metrics and OpenTelemetry metrics.
+ */
+@Slf4j
+public class TopicListMemoryLimiter extends AsyncDualMemoryLimiterImpl {
```
poorbarcode, line 44: Do we need to indicate in this class name that this is a class related to metrics?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: > Instead of updating and acquiring additional permits if the actual size is larger than estimated, we'd better improve as follows:
> 
> * The current request does not need to acquire permits again; only updating the permits that were borrowed out is enough, even though the permits were over-acquired
> * Updates the permits that were borrowed out, to limit the following permits acquisition;

This is already encapsulated in the logic of the `update` implementation. Please check the implementation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: > Just add a tip: if the data does not exist in the cache, it may query ZK, which will also use direct memory

I agree that there might be other sources of memory allocations. One example is AbstractMetadataStore's childrenCache which consumes heap memory without bounds. #24868 addresses that. In the case of ZK, direct memory usage itself is fairly low.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
lhotari: similar response.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: > This function is a notification for the client, to let the client know which topic was changed; it should not be limited by the new feature. 

I disagree. To properly limit memory usage, it's also necessary to handle CommandWatchTopicListSuccess / CommandWatchTopicUpdate as explained in https://github.com/apache/pulsar/blob/master/pip/pip-442.md .

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
lhotari: > Do we need to indicate in this class name that this is a class related to metrics?

I don't think so. There are many classes in Pulsar that contain metrics and it's not part of the class name.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
lhotari: This is necessary for the behavior that the first request passes through to resolve the size. This behavior is explained in the javadoc of the getSizeAsync method.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
lhotari: It is needed. Please read the javadoc for getSizeAsync

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
lhotari: It's intentional that it's dropped, I guess adding a comment explaining the logic could be helpful.

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: ok, will add separate exception messages for these cases

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
lhotari: Adding comments

lhotari:
@codelipenghui @poorbarcode I've addressed the review comments. PTAL

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListSizeResultCache.java
poorbarcode: Okay, the first request will get a constant value, and the second will wait for the first request to be finished if it has not completed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: No, the implementation is like follows, which never covered the case I said above
- update
- call `internalAcquire` if `realSize > estimatedSize`, see also https://github.com/apache/pulsar/pull/24833/files#diff-1daf74806e4d1f752d9576bd63d71678a7a46cbc7a1cf56f71ba88719f87d346R173-R175
- `internalAcquire` will pending the current thread or throws errors, see also https://github.com/apache/pulsar/pull/24833/files#diff-1daf74806e4d1f752d9576bd63d71678a7a46cbc7a1cf56f71ba88719f87d346R117-R122

- **method update**
```java
    public CompletableFuture<AsyncSemaphorePermit> update(AsyncSemaphorePermit permit, long newPermits,
                                                          BooleanSupplier isCancelled) {
        long oldPermits = permit.getPermits();
        long additionalPermits = newPermits - oldPermits;
        if (additionalPermits > 0) {
            return internalAcquire(newPermits, additionalPermits, isCancelled);
        }
}
```

- **pending requests**
```java
 private CompletableFuture<AsyncSemaphorePermit> internalAcquire(long permits, long acquirePermits,
                                                                    BooleanSupplier isCancelled) {
        validatePermits(permits);
        PendingRequest request = new PendingRequest(permits, acquirePermits, future, isCancelled);
        if (!queue.offer(request)) {
            future.completeExceptionally(new PermitAcquireQueueFullException(
                    "Semaphore queue is full"));
            return future;
        }
}
```

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
```diff
@@ -0,0 +1,330 @@
...
+        // mark the old permits as released without adding the permits to availablePermits
+        castToImplementation(permit).releasePermits();
+        if (additionalPermits > 0) {
+            return internalAcquire(newPermits, additionalPermits, isCancelled);
+        } else {
+            // new permits are less than the old ones, so we return the difference
+            availablePermits.addAndGet(-additionalPermits);
+            processQueue();
+            // return the new permits immediately
+            return CompletableFuture.completedFuture(new SemaphorePermit(newPermits));
```
poorbarcode: There should be a bug:

- `withAcquiredPermits`
  - acquired estimated permits ( `100` )
-  `withUpdatedPermits`
  - update permits when got a real size( `50`)
  - release `50` permits here
- finally the permits acquired will be released
  - `withAcquiredPermits -> releaser`:  releases `100` permits
  - `withUpdatedPermits -> releaser`: releases `50` permits
- Issue occurs: the permits is larger than the expected value `100`

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
```diff
@@ -0,0 +1,330 @@
...
+            queueLatencyRecorder.accept(ageNanos);
+        }
+    }
+
+    @Override
+    public CompletableFuture<AsyncSemaphorePermit> update(AsyncSemaphorePermit permit, long newPermits,
+                                                          BooleanSupplier isCancelled) {
+        validatePermits(newPermits);
+        if (isUnbounded()) {
+            return CompletableFuture.completedFuture(new SemaphorePermit(newPermits));
```
poorbarcode, line 175: https://github.com/apache/pulsar/pull/24833/files#r2476939392 The same bug will happen here

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
poorbarcode: Please don't close this comment for now. I will review this piece of code again. I will solve it as I have completed the review

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
lhotari: > No, the implementation is like follows, which never covered the case I said above
> 
> * update
> * call `internalAcquire` if `realSize > estimatedSize`, see also https://github.com/apache/pulsar/pull/24833/files#diff-1daf74806e4d1f752d9576bd63d71678a7a46cbc7a1cf56f71ba88719f87d346R173-R175
> * `internalAcquire` will pending the current thread or throws errors, see also https://github.com/apache/pulsar/pull/24833/files#diff-1daf74806e4d1f752d9576bd63d71678a7a46cbc7a1cf56f71ba88719f87d346R117-R122

I don't know exactly what you are referring to. I made some modifications so that permits on the original instance are only released after successfully acquiring the new permits. If you have something else in mind, it could be useful to reveal more details.

> Instead of updating and acquiring additional permits if the actual size is larger than estimated, we'd better improve as follows:

This is intentionally designed in the way it is. I'll explain in the following paragraph.

> * The current request does not need to acquire permits again; only updating the permits that were borrowed out is enough, even though the permits were over-acquired

In the current design the original permit instance is reseted when the permits are moved to a new instance successfully (that was now fixed). This is intended behavior so that any later logic can always release the original permit without side-effects. This makes it easier to prevent permits leaks when such pattern is applied. For example, the org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiterUtil#withPermitsFuture and org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiterUtil#acquireDirectMemoryPermitsAndWriteAndFlush helpers are simpler to implement when there's 2 separate permit instances. You can also review the implementation code and see why I came to this design decision.

> * Updates the permits that were borrowed out, to limit the following permits acquisition;

I wonder if you meant to point out the bug that there was previously where original permits were released before the new permit (with updated amount) was successfully acquired?

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: When permits are transferred to another instance, the permits are released from the instance. Calling release again won't release the permits again. 
There was a bug where permits where released before the updated permits were successfully acquired. That is fixed and there's new test cases to cover it.

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: The above link doesn't resolve anymore. In the unbounded case, the semaphore is like a no-op. If someone would like to turn off the semaphore behavior, it's possible by setting maxPermits to 0. That makes it just hand out the permits immediately. There's no bug in this case.

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: I added test cases to cover your concerns. There's no bug.

poorbarcode: You are right(there is no bug), I added a new test as follows

```java
public void testWithAcquiredAndUpdatedPermitsDoesntLeakPermitsWhenUpdatedSizeIsLarger() throws Exception {
    assertThat(limiter.getLimiter(LimitType.HEAP_MEMORY).getAvailablePermits()).isEqualTo(10000);
    CompletableFuture<Void> result = limiter.withAcquiredPermits(
            90,  // estimated permits
            LimitType.HEAP_MEMORY,
            () -> false,
            firstPermit -> {
                // Simulate getting actual size and updating permits
                return limiter.withUpdatedPermits(
                        firstPermit,
                        100,
                        () -> false,
                        secondPermit -> {
                            System.out.println(0/0);
                        },
                        throwable -> CompletableFuture.failedFuture(throwable)
                );
            },
            throwable -> CompletableFuture.failedFuture(throwable)
    );
    try {
        result.get();
    } catch (Exception e) {
        // it is expected.
    }
    assertThat(limiter.getLimiter(LimitType.HEAP_MEMORY).getAcquiredPermits()).isEqualTo(0);
    assertThat(limiter.getLimiter(LimitType.HEAP_MEMORY).getAvailablePermits()).isEqualTo(10000);
}
```

----

Sorry, I truly understand the entire process now. 
- Util method: acquire permits
  - Creates the first object AsyncDualMemoryLimiterPermit (`p1`) 
- Util method: acquire permits
  - Creates the second object AsyncDualMemoryLimiterPermit (`p2`) 
  - Sets the value of `p1` to `0`
- Releases permits
  - Release `p2`, which maintains the full permits
  - Release `p1`, it has been set to `0`

And another concept is also important:
- `SemaphorePermit.releasePermits` releases nothing, it just updates the permits it acquired
- `AsyncDualMemoryLimiter.release(permits)` does release the permits

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
poorbarcode: Mark the comment as solved with the comment https://github.com/apache/pulsar/pull/24833#discussion_r2487231319

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: I just want to throw an improvement: Instead of updating and acquiring additional permits if the actual size is larger than estimated, we can skip requesting the additional permits and just mark it as "beyond the original value". 

For example:
- max permits available: `100`
- estimated `100`
  - acquire `100`
- get the real size `110`
  - Apply directly without checking if the remaining permission is sufficient
  - permits left: `-10`

---

- Benefit: An already applied 100 permits will not block subsequent requests
- Disadvantage: It has broken through the limitations
 
---

Anyway, I feel that the current implementation is also fine

lhotari: > Anyway, I feel that the current implementation is also fine

Ok, I'll keep the current implementation.

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java
lhotari: Tests added in 578f027

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
poorbarcode: @lhotari The command should not be limited, because `MultiTopicsConsumer` and `MultiTopicsProducer` need to receive this command once the topics' partition is changed. The regexp consumer is also relies on the command.

lhotari: > @lhotari The command should not be limited, because `MultiTopicsConsumer` and `MultiTopicsProducer` need to receive this command once the topics' partition is changed. The regexp consumer is also relies on the command.

@poorbarcode There's perhaps some misunderstanding. Memory limiting won't prevent this command from being executed unless it times out or the queue fills up. It is absolutely necessary to handle also CommandWatchTopicListSuccess / CommandWatchTopicUpdate related commands so that the broker doesn't run out of memory and that's also explained in [PIP-442](https://github.com/apache/pulsar/blob/master/pip/pip-442.md). Handling CommandWatchTopicListSuccess / CommandWatchTopicUpdate related commands with the memory limiter will help the broker remain stable regardless of a large amount of regex consumers connecting at once. When the limit is reached, it will queue up the requests until other requests have completed. 
Does this make sense?

poorbarcode: >  unless it times out or the queue fills up

That situation is what I want to say. It will cause consumers to be unable to receive messages that were sent to the new partitions until the client application restarts.

lhotari: > That situation is what I want to say. It will cause consumers to be unable to receive messages that were sent to the new partitions until the client application restarts.

That doesn't hold since the solution will retry with back off in case of errors.

There's some additional changes that I'll need to make. 
* There is a need to stop retries 
  * if the watcher is closed
  * the connection is closed
* operations should be handled in order
  * retrying could break the ordering currently

I'll follow up with changes to address the above issues.

lhotari: @poorbarcode I've addressed the required changes. Topic list watchers will work fine with this solution. In case of running out of permits, there will be infinite retries (until the watcher is closed by closing it's connection). Backpressure is about slowing down responses in this case to reduce the load on the system. I'll continue to add a test case where the problems of the existing solution would be demonstrated in the similar way as there's org.apache.pulsar.client.api.PatternConsumerBackPressureMultipleConsumersTest, but using actual client consumers that use topic list watchers.

poorbarcode: > That doesn't hold since the solution will retry with back off in case of errors.

It absolutely holds, because you set a `MAX_RETRY_COUNT (default value is 10) `, right?

> In case of running out of permits, there will be infinite retries (until the watcher is closed by closing it's connection).

This new change can solve the issue I mentioned. But I still think that we'd better remove this limitation because the topic change event uses less memory, and the event will rarely occur

lhotari: @poorbarcode It appears that previous changes have broke the topic list watcher functionality. In the current Pulsar client code, the CommandWatchTopicListSuccess response is ignored:
https://github.com/apache/pulsar/blame/1300f0a4b1c015e278ff321d58c12a7e85ba12e7/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TopicListWatcher.java#L136-L152

This was changed in #21853 while fixing the lost topic list events issue. While working on the code, I found the actual root cause that causes topic list events to be lost. I'll fix that in a separate PR.

What I'll do now is that I'll split the Topic list watcher memory limiting to another PR from this current PR. That is useful since the PR to fix the topic list events issue is not related to PIP-442. After that is fixed, it makes sense to continue adding PIP-442 to cover also topic list events. 

This way we can also make progress on getting this current PR merged, without the topic list watcher limiting changes.

lhotari: Topic list watcher changes have been reverted. @poorbarcode PTAL

pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncDualMemoryLimiter.java
```diff
@@ -0,0 +1,114 @@
...
+                                                         Function<AsyncDualMemoryLimiterPermit,
+                                                                 CompletableFuture<T>> function,
+                                                         Function<Throwable, CompletableFuture<T>>
+                                                                 permitAcquireErrorHandler) {
+        return AsyncDualMemoryLimiterUtil.withPermitsFuture(acquire(memorySize, limitType, isCancelled), function,
+                permitAcquireErrorHandler, this::release);
+    }
+
+    /**
+     * Executed the specified function with updated permits and release the permits after the returned future completes.
```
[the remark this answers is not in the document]
lhotari: I'll fix in follow up PR.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/topiclistlimit/TopicListMemoryLimiter.java
```diff
@@ -0,0 +1,312 @@
...
+     * @return the estimated heap memory size in bytes
+     */
+    public static long estimateTopicListSize(List<String> topicList) {
+        return topicList.stream()
+                .mapToLong(ByteBufUtil::utf8Bytes) // convert character count to bytes
+                // add 32 bytes overhead for each entry
+                // 16 bytes for object header, 16 bytes for java.lang.String fields
+                .map(n -> n + 32)
+                .sum();
+    }
```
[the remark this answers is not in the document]
lhotari: It says "convert character count to bytes" which should be clear.

lhotari:
Cherry-picked to branch-4.0. Mailing list discussion thread was https://lists.apache.org/thread/0n0nf2g7jm75lv3n8ymlxcvx7o2x4mnw

PR #24839, opened 2025-10-13
[improve][offload]use dedicated bookkeeper-ml-offload-scheduler to avoid block core service

ethqunzhong:
### Motivation
We use HDFS as the offload remote storage in our online scenarios. During an HDFS namenode switchover, the normal service of the broker was disrupted due to the failure and retry of a large number of offload operations. The logs are as follows:
image

Offload operations in Pulsar's managed ledger share the same scheduler thread pool with core services. When offload operations block or take too long, they can impact critical managed ledger operations like data read/write and metadata management, causing system performance degradation.

This change introduces a dedicated offload scheduler to isolate offload operations from core services, preventing potential blocking issues.

### Modifications

1. **Added configuration support**:
   - New `managedLedgerNumOffloadSchedulerThreads` parameter in `ServiceConfiguration` and `broker.conf`
   - Default thread count set to available CPU cores

2. **Created dedicated offload scheduler**:
   - Added `offloadScheduler` field in `ManagedLedgerFactoryImpl` using `OrderedScheduler`
   - Proper shutdown handling in factory cleanup

3. **Isolated offload operations**:
   - Migrated all offload-related async operations from `scheduledExecutor` to dedicated `offloadScheduler`
   - Affected operations: `maybeOffloadInBackground()`, `maybeOffload()`, `cleanupOffloaded()`, and `tryTransformLedgerInfo()`

This ensures offload operations don't interfere with core managed ledger functionality, improving overall system stability and performance.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:  https://github.com/ethqunzhong/pulsar/pull/11

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1249,6 +1249,9 @@ managedLedgerDigestType=CRC32C
 # Number of threads to be used for managed ledger scheduled tasks
 managedLedgerNumSchedulerThreads=
 
+# Number of threads to be use for managed ledger scheduled offload operations.
+managedLedgerNumOffloadSchedulerThreads=
+
 # Amount of memory to use for caching data payload in managed ledger. This memory
 # is allocated from JVM direct memory and it's shared across all the topics
 # running  in the same broker. By default, uses 1/5th of available direct memory
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerFactoryConfig.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerFactoryConfig.java
@@ -148,6 +148,11 @@ public class ManagedLedgerFactoryConfig {
     private String managedCursorInfoCompressionType = MLDataFormats.CompressionType.NONE.name();
 
     /**
+     * Number of threads to use for ML offload operations.
+     */
+    private int numManagedLedgerOffloadSchedulerThreads = Runtime.getRuntime().availableProcessors();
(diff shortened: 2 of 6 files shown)


dao-jun:
Offloader already have a separate threadpool, the configuration key is `managedLedgerOffloadMaxThreads`, and it's default value is 2. Why add a new  threadpool?

ethqunzhong:
> Offloader already have a separate threadpool, the configuration key is `managedLedgerOffloadMaxThreads`, and it's default value is 2. Why add a new threadpool?

Strictly speaking, the thread pool configured by `managedLedgerOffloadMaxThreads`(use at `FSOffloader.scheduler`) is used in the offloader instance to perform cold storage operations for the sequential reading and writing of specific ledgers. The thread pool newly added in this modification is used for scheduling the offload operations of all ManagedLedgers at the Broker level—a capability that was previously handled by `ManagedLedgerImpl#scheduledExecutor`.

dao-jun:
Maybe add a Backoff to https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3702-L3704 is better? To avoid frequent retries.

hangc0276:
@ethqunzhong, could you provide more stack to show how the ordered-scheduler threads run into the Hadoop retry?

PR #24848, opened 2025-10-14
[improve][client]Add null check for Pulsar client clock configuration

3pacccccc:
### Motivation
Currently, when setting a clock for the Pulsar client, there is no null check for the clock parameter. This can cause a NullPointerException  when sending messages if a null clock is passed.[1]

[1]: https://github.com/apache/pulsar/blob/45d71f8fcdd58d8715ada7e418f3acd8b43c141a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L710

### Modifications
This PR adds a null check for the clock parameter to provide a clear error message and prevent NPEs.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:  https://github.com/3pacccccc/pulsar/pull/26

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
@@ -332,6 +332,17 @@ public class SimpleProducerConsumerTest extends ProducerConsumerBase {
         });
     }
 
+    @Test(timeOut = 5000)
+    public void pulsarClientClockCheckTest() {
+        assertThatThrownBy(
+                () -> PulsarClient.builder()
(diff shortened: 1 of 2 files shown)


3pacccccc:
BTW, For custom clock implementations, if the ```millis()``` method returns abnormal data (such as negative numbers(eg. -1), extremely large values(eg. 9999999999)) or throws exceptions directly, these cases are not validated. Such validations should be considered? any suggestions?

PR #24857, opened 2025-10-16
[feat][pip] PIP-446: Support Native OpenTelemetry Tracing in Pulsar Java Client

Technoboy-:
### Motivation

Support Native OpenTelemetry Tracing in Pulsar Java client

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-446.md
+++ b/pip/pip-446.md
@@ -520,5 +520,5 @@ User documentation provided in:
 
 # Links
 
-* Mailing List discussion thread: [To be added]
-* Mailing List voting thread: [To be added]
+* Mailing List discussion thread: https://lists.apache.org/thread/61f5zxsc8ydcbjllwbp2x6y5g4f1xj5m
+* Mailing List voting thread: https://lists.apache.org/thread/yv455tyyyrpr3j6o1lzx8k9bq4d0ko5m


pip/pip-446.md
```diff
@@ -0,0 +1,532 @@
...
+}
+```
+
+**ClientBuilder new methods**:
+
+```java
+/**
+ * Set the OpenTelemetry instance for observability.
+ */
+ClientBuilder openTelemetry(OpenTelemetry openTelemetry);
```
crossoverJie: This method already exists:

https://github.com/apache/pulsar/blob/master/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java#L616

pip/pip-446.md
```diff
@@ -0,0 +1,532 @@
...
+
+No CLI changes. This is a client library feature.
+
+### Metrics
+
+This PIP focuses on distributed tracing, not metrics. No new metrics are added.
+
+# Monitoring
+
+Users can monitor tracing effectiveness through their OpenTelemetry backend (e.g., Jaeger, Zipkin, Grafana Tempo):
```
crossoverJie, line 367: Where can users configure the backend address? I don't see any related settings. 

 Is it through environment variables?

pip/pip-446.md
```diff
@@ -0,0 +1,532 @@
...
+1. **Producer tracing**: Create spans for message send operations with automatic trace context injection
+2. **Consumer tracing**: Create spans for message receive/process operations with automatic trace context extraction
+3. **Trace context propagation**: Inject and extract W3C Trace Context via message properties
+4. **Programmatic API**: Enable tracing via `ClientBuilder` API
+5. **Interceptor-based design**: Implement using Pulsar's existing interceptor mechanism
+6. **Cumulative acknowledgment support**: Properly handle span lifecycle for cumulative acks
+7. **Multi-topic consumer support**: Track spans across multiple topic partitions
+8. **Agent compatibility**: Ensure compatibility with OpenTelemetry Java Agent
+9. **Semantic conventions**: Follow OpenTelemetry messaging semantic conventions
+10. **Zero overhead when disabled**: No performance impact when tracing is not enabled
```
lhotari, line 56: One additional point to consider in the design is the behavior when sampling is configured (I believe that's the default) and a specific span isn't recorded at all. In those cases, an implementation can choose to reduce the overhead by omitting certain calls to the OTel tracing API. 
This is explained in the [Span IsRecording](https://opentelemetry.io/docs/specs/otel/trace/api/#isrecording) and [Sampling](https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampling) documentation.
Although a span isn't sampled, it's still necessary to create child spans to propagate the span context to downstream, so optimizations based on span.isRecording() might not be very useful in the end.

I also noticed that the current Otel API that we have in Pulsar doesn't contain isRecording. We should also make sure to upgrade to latest stable versions of Otel libraries when implementing this PIP.

pip/pip-446.md
lhotari: Yes, that's handled with OTel configuration. Usually OTel configuration is handled with environment variables, although there are programmatic ways to construct and configure the `OpenTelemetry` instance.

pip/pip-446.md
```diff
@@ -0,0 +1,532 @@
...
+
+### 9. Span Attributes
+
+Following [OpenTelemetry messaging semantic conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/):
+
+**Producer spans**:
+- `messaging.system`: "pulsar"
+- `messaging.destination.name`: Topic name
+- `messaging.operation.name`: "send"
+- `messaging.message.id`: Message ID (added on ack)
```
lhotari, line 261: how is Message ID serialized?

pip/pip-446.md
```diff
@@ -0,0 +1,532 @@
...
+- `TopicMessageIdImpl`: Delegates `TraceableMessageId` methods to wrapped message ID
+- `ConsumerBase`: Provides `getSubscriptionType()` for interceptors
+- `ConsumerBuilderImpl`: Auto-adds consumer interceptor when enabled
+- `ProducerBuilderImpl`: Auto-adds producer interceptor when enabled
+- `PulsarClientImpl`: Stores and provides OpenTelemetry configuration
+- `ClientConfigurationData`: Stores OpenTelemetry and enableTracing settings
+
+### 9. Span Attributes
+
+Following [OpenTelemetry messaging semantic conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/):
```
lhotari, line 255: Does the scope of this work include publishing semantic conventions for Pulsar directly to OpenTelemetry project? For example contribution custom attributes to https://github.com/open-telemetry/semantic-conventions/blob/main/model/messaging/registry.yaml / https://github.com/open-telemetry/semantic-conventions/blob/main/model/messaging/spans.yaml and contributing the specification documentation?

pip/pip-446.md
```diff
@@ -0,0 +1,532 @@
...
+
+**Producer spans**:
+- `messaging.system`: "pulsar"
+- `messaging.destination.name`: Topic name
+- `messaging.operation.name`: "send"
+- `messaging.message.id`: Message ID (added on ack)
+
+**Consumer spans**:
+- `messaging.system`: "pulsar"
+- `messaging.destination.name`: Topic name
```
lhotari, line 265: Where's the subscription name and type? 

The subscription name could be considered part of the destination. This is how it's handled in the RabbitMQ semantic conventions:
https://github.com/open-telemetry/semantic-conventions/blob/a7d4b413b133c60704c334bff2a4053c89dce791/model/messaging/spans.yaml#L121-L136

A similar approach could work for Pulsar where the `messaging.destination.name` is in the format `[topicname]:[subscriptionname]`. However, that might not be the best possible solution since I guess a Pulsar topic name and subscription name can contain `:`.

lhotari: Never mind encoding the subscription name in the destination name. It looks like there's a subscription name too. 
[`messaging.destination.subscription.name`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/messaging/#messaging-destination-subscription-name) at https://opentelemetry.io/docs/specs/semconv/registry/attributes/messaging/#general-messaging-attributes

pip/pip-446.md
Technoboy-: removed

pip/pip-446.md
Technoboy-: only store MessageId.toString()

pip/pip-446.md
Technoboy-: Thanks @lhotari .
Added

pip/pip-446.md
Technoboy-: I think so, whatever please help review the implementation to avoid missing something.

poorbarcode (approved):

Issue #24850, opened 2025-10-14
Flaky-test: SimpleProducerConsumerTest.testReceiveAsyncCompletedWhenClosing


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/apache/pulsar/actions/runs/18382798391/job/52374149209#step:11:1777

### Exception stacktrace

```text
org.testng.internal.thread.ThreadTimeoutException: Method org.apache.pulsar.client.api.SimpleProducerConsumerTest.testReceiveAsyncCompletedWhenClosing() didn't finish within the time-out 10000
	at java.base/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base/java.util.concurrent.locks.LockSupport.park(LockSupport.java:221)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:754)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1099)
	at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:230)
	at org.apache.pulsar.client.api.SimpleProducerConsumerTest.testReceiveAsyncCompletedWhenClosing(SimpleProducerConsumerTest.java:4095)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:11)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
[...]

PR #24858, opened 2025-10-16
[fix][test]fix flaky SimpleProducerConsumerTest.testReceiveAsyncCompletedWhenClosing

3pacccccc:
Fixes https://github.com/apache/pulsar/issues/24850

### Motivation
The test ```testReceiveAsyncCompletedWhenClosing``` was experiencing concurrency race conditions that caused flaky test failures. The issue occurred due to unpredictable timing between two operations:

- ```consumer.receiveAsync()``` or ```batchReceiveAsync()```

- ```consumer.close()```

#### Expected behavior:
- receiveAsync() should happen before close(), allowing the async operation to be properly interrupted by the close.

#### Flaky scenario:
- When ```close()``` executes before receiveAsync(), the subsequent ```receiveAsync()``` call immediately returns a ```CompletableFuture``` that completes with a ```PulsarClientException``` containing the message **"Consumer already closed"**. This differs from the expected interruption behavior and causes the test assertion to fail.
### Modifications
ensure ```consumer.receiveAsync()```  happens before ```consumer.close()```
#### Solution
Added proper synchronization using CountDownLatch to ensure:

- The async operation (receiveAsync/batchReceiveAsync) starts first

- The close() operation executes only after the async operation has begun
### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/27

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
@@ -4032,17 +4032,17 @@ public class SimpleProducerConsumerTest extends ProducerConsumerBase {
         // 1) Test receiveAsync is interrupted
         CountDownLatch countDownLatch = new CountDownLatch(1);
         new Thread(() -> {
-            CountDownLatch subCoundDownLatch = new CountDownLatch(1);
+            CountDownLatch subCountDownLatch = new CountDownLatch(1);
             try {
                 new Thread(() -> {
                     try {
-                        subCoundDownLatch.await();
+                        subCountDownLatch.await();
                         consumer.close();
                     } catch (PulsarClientException | InterruptedException ignore) {
                     }
                 }).start();
                 CompletableFuture<Message<String>> futhre = consumer.receiveAsync();
-                subCoundDownLatch.countDown();
+                subCountDownLatch.countDown();
                 futhre.get();
                 Assert.fail("should be interrupted");
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
```diff
@@ -4032,14 +4032,18 @@ public void testReceiveAsyncCompletedWhenClosing() throws Exception {
         // 1) Test receiveAsync is interrupted
         CountDownLatch countDownLatch = new CountDownLatch(1);
         new Thread(() -> {
+            CountDownLatch subCoundDownLatch = new CountDownLatch(1);
```
Denovo1998: type: `subCountDownLatch`

pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
Denovo1998, line 4051: This test is not so easy to reproduce locally. What if we add countDown in finally?

pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
Denovo1998: typo: `subCountDownLatch`

3pacccccc: > typo: `subCountDownLatch`

thanks for pointing this out, fixed

pulsar-broker/src/test/java/org/apache/pulsar/client/api/SimpleProducerConsumerTest.java
3pacccccc: > This test is not so easy to reproduce locally. What if we add countDown in finally?

you can add ```thread.sleep(100)``` before ```consumer2.batchReceiveAsync().get();``` in the origin test to reproduce this issue. 
I don't think add countDown in finally will solve this issue. it's concurrency race conditions between ```consumer.receiveAsync()``` and ```consumer.close()```

PR #22378, opened 2024-03-28
[feat][broker] PIP-321 Introduce allowed-cluster at the namespace level

liangyepianzhou:
PIP: https://github.com/apache/pulsar/pull/21648

### Motivation

Because the configurations of clusters at the namespace level are unclear, some flexible topic policies can not work as expected, e.g. geo-replication at the topic level. 

### Modifications

Use `replication-clusters`, `allowed-clusters` to replace a single `replication-clusters` originally in the namespace policy.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24860, opened 2025-10-16
[fix][broker] Fix wrong behaviour when using namespace.allowed_clusters, such as namespace deletion and namespace policies updating

poorbarcode:
### Motivation
Before [PIP-321 Introduce allowed-cluster at the namespace level](https://github.com/apache/pulsar/pull/22378), Pulsar does not support enabling topic level Geo Replication without enabling namespace level Geo Replication, because the policies `namespace.replication_clusters` has two meanings: 
1. Which cluster is allowed to access the namespace?
2. Which clusters enabled Geo Replication?

[PIP-321 Introduce allowed-cluster at the namespace level](https://github.com/apache/pulsar/pull/22378) defined a new policy `namespace.allowed_clusters`, which splits the two definitions.
- if `allowed_clusters` is not empty
  - `allowed_clusters` can be used to define which cluster is allowed to access the namespace, if it is set.
- otherwise: the `replication_clusters` defines both statuses.

---

### 3 Issues

PIP-321 did not complete all the adaptations of codes, such as follows

- 1. Expected behaviour: when a cluster permission was removed from a namespace, Pulsar will unload the namespace to prevent the topics under the namespace from being loaded up again. 
  - Issue: the cluster is still defined `allowed` by `namespace.allowed_clusters`, the namespace is also unloaded. More importantly, in this case, all brokers will trigger an unloading, which will cause this namespace to remain unavailable for a long time. Every time any policy of the namespace is updated, unload will be triggered again for a round, which will cause the problem to become so serious as to be uncontrollable. The test `testUpdateNamespacePolicies` is used to reproduce the issue.

- 2. Expected behaviour: a namespace can only be removed if only one cluster has permission to access it.
  - Issue:  the namespace can be deleted even if `allowed_cluster` is defined as `2` clusters are allowed to access.  The test `testDeleteNamespaceIfTwoClustersAllowed` is used to reproduce the issue.

- 3. Expected behaviour: broker will unload the namespace if the isolation policy is changed, to make the new policy apply.
  - Issue: the namespace will not be unloaded if the allowed permission is defined by `namespace.allowed_cluster`.  The test `testUpdateNamespaceIsolationPolicy` is used to reproduce the issue.

---

### Modifications
- Fix the 3 issues
- Rather than check `replication_clusters` and `allowed_clusters` anywhere, use a common method to deal with them, including the following checks
  - Whether a cluster is allowed to access a namaspace.
  - New policy checking.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/namespace/NamespaceServiceTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/namespace/NamespaceServiceTest.java
@@ -936,7 +936,9 @@ public class NamespaceServiceTest extends BrokerTestBase {
             namespaces.setNamespaceReplicationClusters(namespace, replicationClustersExcel);
             fail();
             //Todo: The status code in the old implementation is confused.
-        } catch (PulsarAdminException.NotAuthorizedException ignore) {}
+        } catch (PulsarAdminException ignore) {
+            assertTrue(ignore.getMessage().contains("allowed clusters list"));
+        }
 
         // 2.2 Peer cluster can not be a part of the allowed clusters.
         LinkedHashSet<String> peerCluster = new LinkedHashSet<>();
@@ -951,6 +953,7 @@ public class NamespaceServiceTest extends BrokerTestBase {
 
         // CleanUp: Namespace with replication clusters can not be deleted by force.
         namespaces.setNamespaceReplicationClusters(namespace, Set.of(conf.getClusterName()));
+        namespaces.setNamespaceAllowedClusters(namespace, Set.of(conf.getClusterName()));
         admin.namespaces().deleteNamespace(namespace, true);
         admin.tenants().deleteTenant(tenant, true);
         for (String cluster : clusters) {


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -1967,13 +1965,17 @@ protected BundlesData validateBundlesData(BundlesData initialBundles) {
...
     private CompletableFuture<Void> validatePoliciesAsync(NamespaceName ns, Policies policies) {
-        if (ns.isV2() && policies.replication_clusters.isEmpty()) {
-            // Default to local cluster
-            policies.replication_clusters = Collections.singleton(config().getClusterName());
+        if (!policies.checkAllowedAndReplicationClusters()) {
+            String msg = String.format("[%s] All replication clusters should be included in allowed clusters."
+                    + " Repl clusters: %s, allowed clusters: %s",
+                    ns.toString(), policies.replication_clusters, policies.allowed_clusters);
+            log.info(msg);
+            throw new RestException(Status.FORBIDDEN, msg);
```
lhotari: `FORBIDDEN` seems to be the incorrect status code for input validation. The correct one would be `BAD_REQUEST` (400).

lhotari:
please check this test failure:
```
  Error:  Tests run: 9, Failures: 1, Errors: 0, Skipped: 4, Time elapsed: 3.943 s <<< FAILURE! -- in org.apache.pulsar.broker.lookup.http.HttpTopicLookupv2Test
  Error:  org.apache.pulsar.broker.lookup.http.HttpTopicLookupv2Test.testValidateReplicationSettingsOnNamespace -- Time elapsed: 0.018 s <<< FAILURE!
  java.lang.AssertionError: expected [404] but found [412]
  	at org.testng.Assert.assertEqualsImpl(Assert.java:149)
  	at org.testng.Assert.assertEquals(Assert.java:131)
  	at org.testng.Assert.assertEquals(Assert.java:1418)
  	at org.testng.Assert.assertEquals(Assert.java:1382)
  	at org.testng.Assert.assertEquals(Assert.java:1428)
  	at org.apache.pulsar.broker.lookup.http.HttpTopicLookupv2Test.testValidateReplicationSettingsOnNamespace(HttpTopicLookupv2Test.java:281)
```

poorbarcode:
> please check this test failure:

Sure

Technoboy- (approved):

codelipenghui (approved):

lhotari:
@poorbarcode Should this fix target also branch-3.0?

poorbarcode:
@lhotari 

> Should this fix target also branch-3.0?

Following the rule described on the website, I think we can skip to cherry-pick it into branch-3.0 since it is not s security issue. If other committers need it, they can drive cherry-picking.

Screenshot 2025-10-31 at 20 01 55

lhotari:
> Following the rule described on the website, I think we can skip to cherry-pick it into branch-3.0 since it is not s security issue. If other committers need it, they can drive cherry-picking.

@poorbarcode The support policy is about Apache Pulsar PMC's commitment to perform releases. If there's a severe bug that impacts branch-3.0, it makes sense to target the branch-3.0 as long as necessary. 
Since this seems to be an important bug fix, wouldn't it make sense to cherry-pick also to branch-3.0 ? There are a lot of community users that haven't yet upgraded to 4.0.x and beyond.

PR #14494, opened 2022-02-28
[PIP-74] Support auto scaled consumer receiver queue 

Jason918:
### Motivation
1. Pick a proper receiver queue size is not an easy thing. If the value is too small, it impact the throughput, and if the value is too large, it consumes too many memory. With default set up, the queue size is 1000, and the max message size is 5MB, this means maximum of 5GB memory occupation.

3. This is part of the work for [PIP 74](https://github.com/apache/pulsar/wiki/PIP-74%3A-Pulsar-client-memory-limits). We need auto scale `currentReceiverQueue` to control client memory. 

### Modifications

Add optional autoScaledReceiverQueueSizeEnabled for consumer client.

Previous `receiverQueueSize` is the max value of this auto scaled queue size.

Every time the client will try to double the size of `currentReceiverQueue` if it limits message throughput. Currently, it's determined by the following two conditions in exact order:
A) Current receiver queue (`ConsumerBase#incomingMessages`) is full after we put a message into it. (it's marked by `scaleReceiverQueueHint` as true).
B) Application wants process more messages but the receiver queue is empty. (`expectMoreIncomingMessages` is called in this PR).

The queue size won't grow if we got new messages during A and B. So if assume current receiver queue size is 10, and the timeline would be like
1. Consumer send flow command to client, and client receive 10 messages, so `scaleReceiverQueueHint` is marked as true. 
2. Application calls `receive()` repeatedly and processed all 10 messages. And in the meanwhile no new message is sent to client[1].
3. Receiver queue size will be doubled to 20.
4. ...

NOTE: 
`Condition A` here is slightly different from original design in PIP-74 described as "there are messages pending to be sent for this consumer" (let's refer it as Condition X). This is proposed with these consideration: 
a) We can assume that when receiver queue is full, the chance of broker have no more messages is insignificant in practice.
b) If we accept a), then `Condition A` implies `Condition X`, as new messages are pending because local queue size is full.  But it's NOT vice versa.
c) Once fact we should accept is that we don't need expand queue size every time it limits the throughput as some slight and occasional delay is acceptable and won't affect overall throughput. And by replacing `Condition A` with `Condition X`, we can reduce the sensitivity of the expansion. 

### Verifying this change

- [x] Make sure that the change passes the CI checks.

This change added tests and can be verified as follows:

*(example:)*
  - org.apache.pulsar.client.impl.AutoScaledReceiverQueueSizeTest
  
### Does this pull request potentially affect one of the following parts:

[...]

PR #24867, opened 2025-10-17
[improve][client] Add independent multiTopicsSinglePartitionReceiverQueueSize config for single partition consumer in multi-topics consumer to reduce memory consumption

oneby-wang:
### Motivation

There no independent `receiverQueueSize` config for single partition consumer in `MultiTopicsConsumerImpl` and `PatternMultiTopicsConsumerImpl`. Although `maxTotalReceiverQueueSizeAcrossPartitions` config can limit in-memory messages of a single topic with multi-partitions, but it can't limit in-memory messages of multi-topics.

For example, if we subscribe to a regex pattern that matches 1000 non-partitioned topics. Before this PR, each non-partitioned topic consumer's `receiverQueueSize` is 1000(`ConsumerImpl` uses the same `receiverQueueSize` value as `PatternMultiTopicsConsumerImpl`), the max messages in memory is 1000 + 1000 * 1000 = 1001000. Let's ignore the insignificant number 1000, if each message size is 8Kb, then we need 1000000 * 8Kb = 7,812.5MB memory to boot our application in catch-up read situation, which is unnecessary.

https://github.com/apache/pulsar/blob/e6560657e20d30103f2f01c3a24600dad1ba9ab6/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java#L236-L249

For backward compatibility, add `multiTopicsSinglePartitionReceiverQueueSizeEnable` switch, the default value is set to false.

### Modifications

Add independent `multiTopicsSinglePartitionReceiverQueueSize` config for single consumer in `MultiTopicsConsumerImpl` to reduce memory consumption.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/7

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java
@@ -454,7 +454,7 @@ public interface ConsumerBuilder<T> extends Cloneable {
      * multi-topics consumer.
      *
      * <p>For partitioned-topic, each partition consumer's receiverQueueSize is the min value of receiverQueueSize and
-     * (maxTotalReceiverQueueSizeAcrossPartitions() / numPartitions)
+     * (maxTotalReceiverQueueSizeAcrossPartitions / numPartitions)
      *
      * @param multiTopicsSinglePartitionReceiverQueueSize
      *            the receiverQueueSize of single partition consumer in multi-topics consumer


oneby-wang:
If this PR is ok, I'll add unit tests to cover the cde change.

lhotari (changes requested):
> There no independent `receiverQueueSize` config for single partition consumer in `MultiTopicsConsumerImpl` and `PatternMultiTopicsConsumerImpl`. Although `maxTotalReceiverQueueSizeAcrossPartitions` config can limit in-memory messages of a single topic with multi-partitions, but it can't limit in-memory messages of multi-topics.
> 
> For example, if we subscribe to a regex pattern that matches 1000 non-partitioned topics. Before this PR, each non-partitioned topic consumer's `receiverQueueSize` is 1000(`ConsumerImpl` uses the same `receiverQueueSize` value as `PatternMultiTopicsConsumerImpl`), the max messages in memory is 1000 + 1000 * 1000 = 1001000. Let's ignore the insignificant number 1000, if each message size is 8Kb, then we need 1000000 * 8Kb = 7,812.5MB memory to boot our application in catch-up read situation, which is unnecessary.

This is a valid problem. However, in this case, since it requires public API changes, it's better to start a discussion on the dev mailing list to ask for feedback from other Pulsar community members. Changing the public API will require a PIP eventually. Before creating a PIP, it's better to ask feedback on ways to address this problem and what approach would others recommend. Adding more options to the API isn't great from usability perspective since it causes more overhead in understanding the usage. It would be better if the multi-topics consumer could automatically tune itself to handle the situation. That might have been the purpose of PIP-74 receiver queue autoscaling ([PR 14494](https://github.com/apache/pulsar/pull/14494)), so it's worth checking that out too.

Are you already using [PIP-74](https://github.com/apache/pulsar/blob/master/pip/pip-74.md) memory limiter (enabled by default) and receiver queue autoscaling?

oneby-wang:
Not tried autoscaling yet. I noticed that we can auto scale receiverQueueSize using memory limiter by turning on `autoScaledReceiverQueueSizeEnabled` switch, but after some code analysis, I think autoscaling may probably not work in multi-topics situation. 

Every single partition consumer starts with currentReceiverQueueSize flow action, meaning that the consumer will receive currentReceiverQueueSize messages.
https://github.com/apache/pulsar/blob/e6560657e20d30103f2f01c3a24600dad1ba9ab6/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java#L236-L249

`MultiTopicsConsumerImpl` creates single partition consumer using there own receiverQueueSize and set batch receive maxNumMessages to (receiverQueueSize / 2).
https://github.com/apache/pulsar/blob/678db6b34587b0042235f0d31e4c406bb0263bcd/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java#L1216-L1231

If `autoScaledReceiverQueueSizeEnabled` switch is on, we set `currentReceiverQueueSize` to `minReceiverQueueSize()`.
https://github.com/apache/pulsar/blob/678db6b34587b0042235f0d31e4c406bb0263bcd/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java#L220-L226

`minReceiverQueueSize()` is (2 * batchReceivePolicy.getMaxNumMessages() - 2≈ receiverQueueSize) according to above analysis. So, the application boots with about (n * receiverQueueSize) messages, n is the topic nums.
https://github.com/apache/pulsar/blob/678db6b34587b0042235f0d31e4c406bb0263bcd/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L502-L509

Every single partition consumer's `minReceiverQueueSize()` is about receiverQueueSize due to batch receive maxNumMessages is (receiverQueueSize / 2), so it can't be reduced below (2 * batchReceivePolicy.getMaxNumMessages() - 2 ≈ receiverQueueSize).
https://github.com/apache/pulsar/blob/678db6b34587b0042235f0d31e4c406bb0263bcd/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java#L250-L259

This PR can tune single partition consumer's receiverQueueSize and batch receive maxNumMessages using multiTopicsSinglePartitionReceiverQueueSize config, which may help to solve this problem.

EDIT: tested, memory limiter can't solve this problem, pulsar client version 4.0.6.

And more, I think multi-topics consumer and it's inner single partition consumers just like parent container and child containers relation, and their receiverQueueSize should be tuned independently in someway.

PR #24868, opened 2025-10-17
[improve][meta] Add byte size limit to AbstractMetadataStore's childrenCache

lhotari:
### Motivation

AbstractMetadataStore contains an unbounded childrenCache which might consume a lot of heap memory when a lot of namespaces with a lot of topics are listed. The cache has a timeout of 10 minutes so the memory usage problem is limited to certain use cases. For example, when all topics are listed for all namespaces in less than 10 minutes when there are a lot of namespaces and topics.

### Modifications

Add a limit to the childrenCache so that oldest entries start to get expired when the cache consumes more than 20% of the maximum heap size.

Tests are TBD.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,7 +247,7 @@ flexible messaging model and an intuitive client API.</description>
     <json-smart.version>2.5.2</json-smart.version>
     <opensearch.version>2.16.0</opensearch.version>
     <elasticsearch-java.version>8.15.3</elasticsearch-java.version>
-    <debezium.version>3.2.2.Final</debezium.version>
+    <debezium.version>3.2.4.Final</debezium.version>
     <debezium.postgresql.version>${postgresql-jdbc.version}</debezium.postgresql.version>
     <debezium.mysql.version>9.4.0</debezium.mysql.version>
     <jsonwebtoken.version>0.11.1</jsonwebtoken.version>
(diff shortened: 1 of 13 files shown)


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
```diff
@@ -152,6 +180,20 @@ public CompletableFuture<Boolean> asyncReload(String key, Boolean oldValue,
...
 
+    /**
+     * Return the maximum size of the children cache in bytes.
+     * @return maximum size of the children cache in bytes.
+     */
+    protected long getChildrenCacheMaxSizeBytes() {
+        long heapMaxSizeBytes = Runtime.getRuntime().maxMemory();
+        // default 20% of max heap size, this should be sufficient to prevent OOME in the use case
+        // when a lot of namespaces with lots of topics are listed in the metadata store.
+        long defaultSizeBytes = heapMaxSizeBytes / 5;
```
oneby-wang: Hi, @lhotari Should we add a config in broker.conf, so developers can tune this memory size by their self?

nodece (approved):

dao-jun (approved):
LGTM

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java
```diff
@@ -152,6 +180,20 @@ public CompletableFuture<Boolean> asyncReload(String key, Boolean oldValue,
...
+        // min size 20MB
+        int minSizeBytes = 1024 * 1024 * 20;
```
Technoboy-, line 193: It would be better if it were configurable

PR #24873, opened 2025-10-18
[feat][client] PIP-446: Support Native OpenTelemetry Tracing in Pulsar Java Client

Technoboy-:
### Motivation

Implementation of [PIP-446](https://github.com/apache/pulsar/pull/24857)

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OpenTelemetryTracingIntegrationTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OpenTelemetryTracingIntegrationTest.java
@@ -185,6 +185,8 @@ public class OpenTelemetryTracingIntegrationTest extends BrokerTestBase {
         assertNotNull(msg);
         consumer.negativeAcknowledge(msg);
 
+        Thread.sleep(3000);
+
         // Close to ensure negative ack is processed
         producer.close();
         consumer.close();


lhotari:
@asafm Do you have a chance to review this PR?

asafm:
@lhotari yes. This was the only comment I saw.

asafm:
> @lhotari yes. This was the only comment I saw.

Sorry @lhotari I confused with the other one. This one I didn't review

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java
```diff
@@ -615,6 +615,44 @@ ClientBuilder authentication(String authPluginClassName, Map<String, String> aut
...
+     *     .serviceUrl("pulsar://localhost:6650")
+     *     .enableTracing(true)
+     *     .build();
+     * }</pre>
+     *
+     * @param tracingEnabled whether to enable tracing (default: false)
+     * @return the client builder instance
+     * @since 4.2.0
+     */
+    ClientBuilder enableTracing(boolean tracingEnabled);
```
poorbarcode: 收否区分类型？

poorbarcode (approved):
I have one suggestion, it should be useful to determine client-side publish latency issue
- Should we add more events for producers, to let us to record more spans? Such as follows
  - `users call send` -> `cache message with message container`
  - `publish message to broker` -> `receive broker response`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/tracing/OpenTelemetryProducerInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+
+            this.tracer = instrumentProvider.getTracer();
+            this.propagator = GlobalOpenTelemetry.getPropagators().getTextMapPropagator();
+            this.initialized = true;
+        }
+    }
+
+    @Override
+    public void close() {
+        // No cleanup needed - spans are attached to messages
```
poorbarcode:
```suggestion
        // Producer will fail pending messages when it being closed, which will trigger the `onSendAcknowledgement` events
```

pulsar-client/src/main/java/org/apache/pulsar/client/impl/tracing/OpenTelemetryProducerInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public boolean eligible(Message message) {
+        return tracer != null && propagator != null;
+    }
+
+    @Override
+    public Message beforeSend(Producer producer, Message message) {
+        // Initialize tracer from producer on first call
+        initializeIfNeeded(producer);
```
poorbarcode, line 88: Shouldn't we treat the first and last parts of a chunked message differently?

poorbarcode: Please ignore this comment channel, because `sendComplete` already covered the case here: https://github.com/apache/pulsar/blob/v4.1.2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L1372-L1381

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TraceableMessage.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Set the OpenTelemetry span associated with this message.
+     * <p>
+     * This method is called by tracing interceptors to attach a span to the message
+     * for later retrieval when completing the span.
+     *
+     * @param span the span to associate with this message, or null to clear
+     */
+    void setTracingSpan(Span span);
```
lhotari, line 47: It doesn't seem to make sense to have this method on the public Pulsar client API. Isn't this a method that should only be used internally?

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TraceableMessageId.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Set the OpenTelemetry span associated with this message ID.
+     * <p>
+     * This method is called by tracing interceptors to attach a span to the message ID
+     * for later retrieval in acknowledge callbacks.
+     *
+     * @param span the span to associate with this message ID, or null to clear
+     */
+    void setTracingSpan(Span span);
```
lhotari, line 48: It doesn't seem to make sense to have this method on the public Pulsar client API. Isn't this a method that should only be used internally?

pulsar-client/src/main/java/org/apache/pulsar/client/impl/tracing/TracingContext.java
```diff
@@ -0,0 +1,192 @@
...
+                .startSpan();
+    }
+
+    /**
+     * Mark a span as successful and end it.
+     *
+     * @param span the span to end
+     */
+    public static void endSpan(Span span) {
+        if (span != null && span.isRecording()) {
```
lhotari: the usage of `span.isRecording()` doesn't make sense here. It should only be used to skip adding some costly values to the span.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/tracing/TracingContext.java
```diff
@@ -0,0 +1,192 @@
...
+     *
+     * @param span the span to end
+     * @param throwable the exception that caused the failure
+     */
+    public static void endSpan(Span span, Throwable throwable) {
+        if (span != null && span.isRecording()) {
+            span.setStatus(StatusCode.ERROR, throwable.getMessage());
+            span.recordException(throwable);
+            span.end();
+        }
```
lhotari: Please revisit the usage of `span.isRecording()`

```suggestion
        if (span != null) {
            span.setStatus(StatusCode.ERROR, throwable.getMessage());
            if (span.isRecording() {
                span.recordException(throwable);
            }
            span.end();
        }
```

lhotari (commented):
Please check the comments about exposing internals on the public API.

pulsar-client/TRACING.md
```diff
@@ -0,0 +1,405 @@
...
+    processMessage(msg.getValue());
+    span.setStatus(StatusCode.OK);
+} catch (Exception e) {
+    span.recordException(e);
+    span.setStatus(StatusCode.ERROR);
+    throw e;
+} finally {
+    span.end();
+    consumer.acknowledge(msg);
+}
```
lhotari, line 205: I'm just wondering how could this work. Does this PR contains tests that the custom processing span becomes a child span of the span that is created by the PulsarClient tracing.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OpenTelemetryTracingIntegrationTest.java
```diff
@@ -0,0 +1,752 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Integration tests for OpenTelemetry tracing with real broker.
+ * Note: These tests may be timing-dependent and could be flaky in CI environments.
+ * They verify end-to-end tracing functionality with actual Pulsar broker.
+ */
+@Test(groups = "broker")
+public class OpenTelemetryTracingIntegrationTest extends BrokerTestBase {
```
lhotari, line 56: This test class doesn't seem to test that "Custom Span Creation" feature described in TRACING.md works as expected. I believe that it would be expected that the custom span would be the child span of the span created by the Pulsar client's consume tracing.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/tracing/TracingContext.java
Technoboy-: yes, right. updated

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TraceableMessage.java
Technoboy-: So how could we add the metrics like `messaging.pulsar.acknowledgment.type` when call `onAcknowledge`

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TraceableMessageId.java
Technoboy-: It's for the implementation like `onSendAcknowledgement` in OpenTelemetryProducerInterceptor to add some metrics.
Do you have any better idea?

lhotari: An internal interface could be used, not an interface that is exposed on the public API. I'm assuming that it doesn't make much sense to allow users of the API to call `setTracingSpan` since it's something that only the PIP-446 tracing implementation should be calling internally. Is this the correct assumption?

pulsar-client/TRACING.md
Technoboy-: this is coming from `io.opentelemetry.api.trace.Tracer` and `io.opentelemetry.api.trace.Span`

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TraceableMessageId.java
Technoboy-: could you give the demo to help for the best practice? 
For the message metrics, i don't have a better way to do it right now

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OpenTelemetryTracingIntegrationTest.java
Technoboy-: add related test

PR #24878, opened 2025-10-20
[improve][broker] Extract authentication logic from ServerCnx to dedicated class

nodece:
### Motivation

The current authentication flow in broker is tightly coupled with the `ServerCnx` class, making it harder to extend authentication logic (e.g., for proxy authentication, role forwarding). This PR refactors the authentication handling into dedicated classes to make the flow more modular, extensible, and easier to maintain.

### Modifications

- Decouple authentication from server connection handling for cleaner separation of concerns
- Enable reuse of authentication logic for proxy authentication

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authentication/AuthenticationService.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authentication/AuthenticationService.java
@@ -245,6 +245,15 @@ public class AuthenticationService implements Closeable {
         }
     }
 
+    /**
+     * Creates a binary authentication session for Pulsar's binary protocol.
+     * <p>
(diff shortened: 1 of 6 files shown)


lhotari:
When adding new classes like BinaryAuthSession and BinaryAuthContext, please pay attention to add class level javadocs to explain the abstraction. It should also be considered whether an interface is used for BinaryAuthContext to make it easier to keep the API backwards compatible in the future.

It's currently not clear what the extension model is and how a custom implementation could help extend the authentication logic.

nodece:
This implementation aims to unify authentication handling between the broker and proxy. It’s not currently designed for external extension or customization. I plan to test this in our forked Pulsar.

Issue #24796, opened 2025-09-27
[feat] Pulsar Admin Client should support PIP-234 PulsarClientSharedResources for event loop, timer and DNS resolver sharing


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Pulsar Admin client creates it's own AsyncHttpClient with a Netty event loop, timer and DNS resolver and cache. In use cases where it is desired to reduce resource usage and share the DNS settings in the admin client, it would be useful if the Pulsar Admin client would also support PulsarClientSharedResources. PIP-234 implementation #24790 adds support for sharing thread pools and DNS resolver/cache across multiple client instances with PulsarClientSharedResources API.

### Solution

Adding sharedResources to the builder and using the resources for the AsyncHttpClient instance.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

3pacccccc:
@lhotari Hi, Lari,I think we can close this issue

lhotari:
Implemented by #24893

PR #24893, opened 2025-10-25
[feat][client] PIP-234: Support shared resources in PulsarAdmin to reduce thread usage

3pacccccc:
Main Issue: https://github.com/apache/pulsar/issues/24796

### Motivation
When creating multiple PulsarAdmin clients in the same JVM, each client currently creates its own set of resources (event loops, timers, DNS resolvers). This leads to excessive thread consumption and memory usage when many admin clients are instantiated.

This change allows PulsarAdmin clients to share resources through the `PulsarClientSharedResources` mechanism, similar to what's already available for  PulsarClient.

### Modifications
1. Added `sharedResources(PulsarClientSharedResources)` method to `PulsarAdminBuilder` API
2. Modified `PulsarAdminBuilderImpl` to accept and propagate shared resources configuration
3. Updated `PulsarAdminImpl` constructor to pass shared resources to HTTP connector
4. Enhanced `AsyncHttpConnector` to utilize shared resources:
   - Event loop group for I/O operations
   - Netty timer for timeouts and scheduling
   - DNS resolver for name resolution
5. Added test to verify shared resources are properly reused across multiple admin clients.

### Verifying this change

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/29

### DIFF
--- a/.github/workflows/ci-pulsarbot.yaml
+++ b/.github/workflows/ci-pulsarbot.yaml
@@ -22,6 +22,10 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  actions: write
+  contents: read
+
 jobs:
   pulsarbot:
     runs-on: ubuntu-24.04
@@ -31,5 +35,5 @@ jobs:
       - name: Execute pulsarbot command
         id: pulsarbot
         env:
-          GITHUB_TOKEN: ${{ secrets.PULSARBOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         uses: apache/pulsar-test-infra/pulsarbot@master
--- a/.github/workflows/pulsar-ci.yaml
+++ b/.github/workflows/pulsar-ci.yaml
@@ -318,6 +318,8 @@ jobs:
             group: BROKER_GROUP_3
           - name: Brokers - Broker Group 4
             group: BROKER_GROUP_4
+          - name: Brokers - Broker Group 5
+            group: BROKER_GROUP_5
           - name: Brokers - Client Api
             group: BROKER_CLIENT_API
           - name: Brokers - Client Impl
--- a/README.md
+++ b/README.md
@@ -98,38 +98,66 @@ components in the Pulsar ecosystem, including connectors, adapters, and other la
 
 ## Pulsar Runtime Java Version Recommendation
 
-- pulsar ver > 4.0 and master branch
+> **Note**:
+>
+> When using Java versions, it is recommended to
+> use [a recent version of a particular Java release (17 or 21)](https://adoptium.net/en-GB/temurin/releases?version=21&os=linux&arch=any)
+> with the most recent bug fixes and security patches.
+> For example, the JVM bug [JDK-8351933](https://bugs.openjdk.org/browse/JDK-8351933) can cause stability issues in
+> Pulsar.
+> [JDK-8351933](https://bugs.openjdk.org/browse/JDK-8351933) was fixed in Java 17.0.17+ and Java 21.0.8+.
+> Pulsar Docker images come with the most recent Java version at the time of release.
 
-| Components     | Java Version |
-|----------------|:------------:|
-| Broker         |      17      |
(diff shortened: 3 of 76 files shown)


pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
```diff
@@ -112,33 +120,57 @@ public AsyncHttpConnector(Client client, ClientConfigurationData conf, int autoC
...
+        if (sharedResources != null && sharedResources.getDnsResolverGroup() != null) {
+            EventLoopGroup eventLoopGroupReference;
+            if (sharedResources.getIoEventLoopGroup() != null) {
+                eventLoopGroupReference = sharedResources.getIoEventLoopGroup();
+            } else {
+                // build an EventLoopGroup with default value
+                eventLoopGroupReference = EventLoopUtil.newEventLoopGroup(
+                        Runtime.getRuntime().availableProcessors(), false,
+                        new ExecutorProvider.ExtendedThreadFactory("pulsar-admin-client-io",
+                                Thread.currentThread().isDaemon()));
```
lhotari, line 167: In this case the EventLoopGroup wouldn't get closed and it would leak resources. It would be better to move this logic earlier so that the created event loop group is also used for AsyncHttpClient in this case and it would be closed when the instance is closed.

Please also add a test for this case. Restricting sharing of resources can  be achieved by using the `shareConfigured` method (org.apache.pulsar.client.api.PulsarClientSharedResourcesBuilder#shareConfigured ) so that the eventloopgroup isn't shared by not configuring it, but just calling `configureDnsResolver` to configure the shared dns resolver.

3pacccccc: OK, good catch, I'll fix it. @lhotari

3pacccccc: @lhotari  Hi Lari, could you please review this when you have a moment? Thanks!

pulsar-client-admin/src/test/java/org/apache/pulsar/client/admin/internal/PulsarAdminBuilderImplTest.java
```diff
@@ -188,6 +195,96 @@ public void testClientDescription() throws PulsarClientException {
...
+
+        EventLoopGroup eventLoopGroup1 =
+                pulsarAdminImpl1.getAsyncHttpConnector().getHttpClient().getConfig().getEventLoopGroup();
+        EventLoopGroup eventLoopGroup2 =
+                pulsarAdminImpl2.getAsyncHttpConnector().getHttpClient().getConfig().getEventLoopGroup();
+        Timer nettyTimer1 = pulsarAdminImpl1.getAsyncHttpConnector().getHttpClient().getConfig().getNettyTimer();
+        Timer nettyTimer2 = pulsarAdminImpl2.getAsyncHttpConnector().getHttpClient().getConfig().getNettyTimer();
+        boolean b1 = eventLoopGroup1 == eventLoopGroup2;
+        boolean b2 = nettyTimer1 == nettyTimer2;
+        assertThat(b1 && b2).isTrue();
```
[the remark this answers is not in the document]
lhotari: This is a good suggestion from Copilot. The goal of AssertJ is to provide proper error messages when something fails and the suggestion shows how AssertJ should be used.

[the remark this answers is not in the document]
3pacccccc: fixed

Issue #23027, opened 2024-07-12
[improve][broker] cursor read entry would trigger readMoreEntry() one more time when addWaitingCursor and notify


TakaHiR07:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

As the code shown, even if we disable txn, we use topic.getMaxReadPosition() which is topic.lastConfirmedEntry as OpReadEntry's maxPosition. 

However, there exist the situation:
1. create OpReadEntry, using current topic's lastConfirmedEntry
2. cursor hasNoMoreEntry, addWaitingCursor. opReadEntry become waiting.
3. topic add new entry, trigger notifyEntriesAvailable() to notify opReadEntry
4. Since opReadEntry.maxPosition is the previous lastConfirmedEntry, and now the lastConfirmedEntry has been changed. This op would enter opReadEntry.checkReadCompletion and trigger callback directly without reading any entry.
5. Therefore, it would cause one more time to execute readEntry request. 

So this is where we can improve, actually we may be able to execute the opReadEntry directly.

By the way, in version-2.9, if we disable txn, topic.getMaxReadPosition() is Position.latest, so opReadEntry.maxPosition is Position.latest too. And it would not have this issue since op.maxPosition is always > op.readPosition. 

https://github.com/apache/pulsar/blob/c160cc9c3d44c1df073d63b325b45b9bc9bff8c7/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TransactionBufferDisable.java#L109-L113

https://github.com/apache/pulsar/blob/c160cc9c3d44c1df073d63b325b45b9bc9bff8c7/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherSingleActiveConsumer.java#L368-L381

https://github.com/apache/pulsar/blob/5dc030431a60b49e81d577cd06a1ae63dbee0293/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L934-L979

https://github.com/apache/pulsar/blob/5dc030431a60b49e81d577cd06a1ae63dbee0293/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2051-L2056

[...]

PR #21466, opened 2023-10-30
[fix][txn] Fix getting last message ID when there are ongoing transactions

liangyepianzhou:
### Motivation
1. Consumer/Reader API `getLastMessageIds` should return the max read position instead of `lastConfirmedEntry`.
https://github.com/apache/pulsar/blob/2322004069f23ecca1a12a4ced898fed854275fb/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L2117-L2122
2. When the transaction buffer is recovering, the `MaxReadPosition` of the topic is unknown. Wait it recovers completely is needed.
3. The `MaxReadPosition` should be the previous position of the first ongoing transaction position instead of `EntryID -1`. Because compacted decide whether the topic's managedLedger is null by `EntryID == -1`
https://github.com/apache/pulsar/blob/2322004069f23ecca1a12a4ced898fed854275fb/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L2161-L2176

### Modifications
1. Return `max read position` when calling `getLastMessageIds`.
2. Call `checkIfTBRecoverCompletely` before getting the max read position.
3. If transaction buffer recovery fails, return `Unknown` error to the client. 
### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #24898, opened 2025-10-28
[fix][broker] fix getMaxReadPosition in TransactionBufferDisable should return latest

TakaHiR07:
Fixes https://github.com/apache/pulsar/issues/23027

### Motivation

As shown in the issue, getMaxReadPosition in TransactionBufferDisable should return latest. This implementation is changed in https://github.com/apache/pulsar/pull/21466

### Modifications

getMaxReadPosition in TransactionBufferDisable return latest

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -4140,6 +4140,12 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
         if (lastDispatchablePosition != null) {
             return CompletableFuture.completedFuture(lastDispatchablePosition);
         }
+        Position lastPosition;
+        if (transactionBuffer instanceof TransactionBufferDisable) {
+            lastPosition = getLastPosition();
+        } else {
+            lastPosition = getMaxReadPosition();
+        }
         return ledger.getLastDispatchablePosition(entry -> {
             MessageMetadata md = entry.getMessageMetadata();
             if (md == null) {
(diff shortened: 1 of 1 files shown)


lhotari (approved):
The change makes sense to me since there's other logic in ManagedLedger/ManagedCursor to avoid reading beyond last confirmed entry.

@TakaHiR07 Would it be possible to add some tests?

TakaHiR07:
> The change makes sense to me since there's other logic in ManagedLedger/ManagedCursor to avoid reading beyond last confirmed entry.
> 
> @TakaHiR07 Would it be possible to add some tests?

It is hard to add test for the case in  https://github.com/apache/pulsar/issues/23027. 

Besides, This pr just revert the getMaxReadPosition() implementation to version-2.9.x, but there is a potential issue may be need to consider. DispatcherRateLimiter may not work well in a corner case. 

For example, one topic set a small rate for topic-level dispatcherRateLimiter, and then multiple subscription try to consume one by one. But topic has no new message writing, so all the subscription would become waitingCursor and wait to be notify once new message publish successful. So if new message publish, all the subscription can read message in the same time, maybe exceed the  rate of dispatch.

To some degree, getMaxReadPosition() return lastConfirmedEntry can avoid this corner case . But it is not reasonable because when disable transaction, the maxReadPosition should be latest. Other code implementation should obey this point or it may bring unexpected error.

In all, this potential corner issue should not block this pr

lhotari:
Oddly, this change seems to trigger a lot of test failures which might be test flakiness, invalid tests or some other underlying bug.

lhotari:
@TakaHiR07 A lot of tests break consistently with this change. Do you have a chance to investigate?

TakaHiR07:
> @TakaHiR07 A lot of tests break consistently with this change. Do you have a chance to investigate?

@lhotari Have fixed, The reason is also the change in pr-21466. Previous pr try to fix a issue about transaction, but there is no need to  change the getMaxReadPosition() about non-transaction. So the ServerCnx#getLastMessageId() should return lastConfirmedEntry when transaction disable, as the same as the implementation in before pulsar version.

PR #24916, opened 2025-10-29
[improve][ci] Replace PULSARBOT_TOKEN with GITHUB_TOKEN configured with permissions

lhotari:
### Motivation

The use of `PULSARBOT_TOKEN` secret is unnecessary and should be replaced with `GITHUB_TOKEN` that is configured with sufficient permissions in the workflow file.

### Modifications

- add sufficient permissions to `GITHUB_TOKEN`
- pass `GITHUB_TOKEN` to pulsarbot action

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #24918, opened 2025-10-29
[fix][ci] Add missing permissions to ci-pulsarbot.yaml

lhotari:
### Motivation

- follow up on #24916
- If the permission isn't specified, there will be no permission
  - https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defining-access-for-the-github_token-scopes
- it seems that this request fails without pull-request read permission: https://github.com/apache/pulsar-test-infra/blob/a0ef898276e191789153842618f854a39f972c61/pulsarbot/entrypoint.sh#L52

### Modifications

- Add more permissions so that pulsarbot could work

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-pulsarbot.yaml
+++ b/.github/workflows/ci-pulsarbot.yaml
@@ -25,6 +25,9 @@ on:
 permissions:
   actions: write
   contents: read
+  pull-requests: read
+  checks: read
+  statuses: read
 
 jobs:
   pulsarbot:


lhotari:
Actually, it might be a bug in pulsarbot. URL encoding should be used for branch name at https://github.com/apache/pulsar-test-infra/blob/a0ef898276e191789153842618f854a39f972c61/pulsarbot/entrypoint.sh#L63 .
I'll test with another PR.

lhotari:
This PR isn't needed. It turns out to be a bug in Pulsar bot as mentioned in https://github.com/apache/pulsar/pull/24918#issuecomment-3461353520 . Restarting worked fine in this case: https://github.com/apache/pulsar/actions/runs/18908512581/job/53972649999

lhotari:
Pulsarbot bug is fixed by https://github.com/apache/pulsar-test-infra/commit/8ff059e49446fff5bb9baf2de4a12bc05c2d57ab . The problem came up with PRs submitted from a repository different from the PR creator.

PR #24921, opened 2025-10-30
[improve][test] Add filtering solution for fast selection of TestNG groups

lhotari:
### Motivation

When executing tests with Surefire & TestNG, there's a lot of overhead caused by the test filtering. It takes 1-2 seconds to evaluate whether each test class belongs to the test group since TestNG instantiates each test class, runs configured listeners and then decides whether the class belongs to the test group.
Since there's a lot of test classes in pulsar-broker, a lot of unnecessary time is spent in each broker group test job just to filter out tests.

This solution will reduce the duration of each test job and allow further splitting of pulsar-broker tests to be run in separate test jobs in CI.

### Modifications

- add a utility class for test filtering using Byte-buddy directly at the byte code level
- add maven profiles for filtering and test execution
- adapt build/run_unit_group.sh script to use the test filtering solution

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/build/run_unit_group.sh
+++ b/build/run_unit_group.sh
@@ -99,6 +99,8 @@ function mvn_test() {
           create_filter_targets+=("$target" "-DskipTests")
           # no need to run install for test run
           target="verify"
+      else
+          create_filter_targets+=("test-compile")
       fi
       create_filter_targets+=("exec:exec")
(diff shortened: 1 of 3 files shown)


lhotari:
There's a difference in the number of executed tests for some reason. Would have to check further to find out the difference.

lhotari:
Obsolete after switching to gradle build

Issue #24600, opened 2025-08-04
[Enhancement] Future optimization and defect fixing of the delayed message module


Denovo1998:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

## Future

**Currently, I am deeply researching the delayed message module of Pulsar and want to discuss some things with everyone.**

### Flexible control of delayed messages

Before this, I researched both the community and offline sources. I found that the delayed message feature is very important for the business, and the business requirements demand flexible control over delayed messages. We know that for a message system, once a message is sent, it is immutable, but for delayed messages, the delay itself is actually variable.

Currently, Pulsar's delay message module does not have this variability, and to my knowledge, other messaging systems also do not have similar functionality. However, it is actually achievable.

---

1. Initially, I tried the feature for canceling delayed messages.

https://github.com/apache/pulsar/blob/f66f1b5439f0a9809e84a1720855468ccfedc24c/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/BucketDelayedDeliveryTracker.java

Currently, we can indirectly implement the cancellation of delayed messages by acknowledging the messages.
See PR：https://github.com/apache/pulsar/pull/24370

2. Then @thetumbled proposed that the business needs to achieve delayed messages on the consumer side, but there will be issues such as write amplification.

I also tried to implement it, see the PR for details：https://github.com/apache/pulsar/pull/24372

These two PRs are actually implemented in the same way. We only need ledgerId, entryId, and delayed timestamp to operate on delayed messages, but they all have a common problem.

> If a delayed message in the LastMutableBucket has not been flushed to the Bookie, and a failure occurs, the data in the LastMutableBucket will be lost. However, this will not have any impact, as after restarting, the messages will still be read from the MackDelete position onwards, and the Bucket will be rebuilt. This is why the data in the Bucket can be deleted as long as it is read (without requiring client Ack).
> 
> If we send a command to add or cancel a delayed message from the consumer side, and it fails to be persisted (sealBucketAndAsyncPersistent) in the LastMutableBucket, and the Broker crashes, the command will be lost. We cannot wait until the Seal Bucket condition is triggered to return the delayed message command cancellation success, because we do not know how long it will take.

---

**So we need to solve the problem that commands are lost.**

- We need an operation command type for delayed messages

CANCEL（Delayed message cancellation）
DELAY（Delay message on the consumer side）
LOOP（Loop/Timed delay）
[...]

PR #24927, opened 2025-11-01
[feat][broker] Implement topic-level delayed delivery tracking with in-memory manager

Denovo1998:
Fixes #xyz

Main Issue: #24600

PIP: #24928

### Motivation

The primary motivation for this proposal is to address the high memory consumption caused by the current per-subscription delayed message tracking mechanism. For topics with hundreds or thousands of subscriptions, the memory footprint for delayed messages becomes prohibitively large. Each delayed message's position is duplicated across every subscription's tracker, leading to a memory usage pattern of O(num_delayed_messages * num_subscriptions).

This excessive memory usage can cause:

Increased memory pressure on Pulsar brokers.
More frequent and longer Garbage Collection (GC) pauses, impacting broker performance.
Potential OutOfMemoryErrors, leading to broker instability.
Limited scalability for use cases that rely on many subscriptions per topic, such as IoT or large-scale microservices with shared subscriptions.
By optimizing the delayed message tracking to be more memory-efficient, we can enhance broker stability and scalability, allowing Pulsar to better support these critical use cases.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/Denovo1998/pulsar/pull/16

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/AbstractPersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/AbstractPersistentDispatcherMultipleConsumers.java
@@ -19,9 +19,13 @@
 package org.apache.pulsar.broker.service.persistent;
 
 import java.util.Map;
+import java.util.Optional;
 import org.apache.bookkeeper.mledger.AsyncCallbacks;
 import org.apache.bookkeeper.mledger.ManagedCursor;
+import org.apache.bookkeeper.mledger.Position;
 import org.apache.pulsar.broker.ServiceConfiguration;
+import org.apache.pulsar.broker.delayed.DelayedDeliveryTracker;
+import org.apache.pulsar.broker.delayed.InMemoryTopicDelayedDeliveryTracker;
 import org.apache.pulsar.broker.service.AbstractDispatcherMultipleConsumers;
 import org.apache.pulsar.broker.service.Dispatcher;
 import org.apache.pulsar.broker.service.Subscription;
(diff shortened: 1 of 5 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryTopicDelayedDeliveryTrackerManager.java
```diff
@@ -0,0 +1,648 @@
...
+        timerLock.lock();
+        try {
+            updateTimerLocked();
+        } finally {
+            timerLock.unlock();
+        }
+        if (log.isDebugEnabled()) {
+            log.debug("Updated tickTimeMillis for topic-level delayed delivery manager to {} ms", newTickTimeMillis);
+        }
+    }
```
Denovo1998, line 263: @lhotari @codelipenghui @coderzc @Apurva007 @thetumbled @dao-jun @BewareMyPower 

Currently, getScheduledMessagesForSub does not mark the returned position as "Scheduled for this subscription" nor does it delete it in the topic-level index (can only wait for unified pruning after all subscriptions mark-delete). As a result, before the ack, the next getScheduledMessagesForSub will still return the same position, causing duplicate scheduling and duplicate reading, which in severe cases can lead to duplicate delivery.

The original per-sub InMemoryDelayedDeliveryTracker usually "takes out and removes" from the subscription structure when getScheduledMessages, avoiding duplicate scheduling; after the shared index, a per-sub "pending/scheduled but unconfirmed" collection or cursor is also needed, used to filter out the positions that have been scheduled but not confirmed when getScheduledMessagesForSub, and will not return these positions again before mark-delete.

**This needs optimization, but it feels like there's no low-cost way to handle it.**

1.Index bucketing uses a fixed `indexGranularityMillis` (can be the initial tick or a fixed constant of 10ms/1ms), no longer changes with resetTickTime; resetTickTime only affects the triggering frequency of the Timer.

2.Deduplication on the read side. In SubContext, maintain a pending collection of ledgerId -> Roaring64Bitmap(entryId); get ScheduledMessagesForSub records the returned positions in pending, and subsequent calls filter pending; when mark-delete is advanced, remove from pending; Timer judgment also filters pending to avoid empty rotation.

3.Block and rebuild delayedMessageMap here at onTickTimeUpdated.

PR #24739, opened 2025-09-14
[improve][broker] Optimize fine-grained concurrency control for BucketDelayedDeliveryTracker

Denovo1998:
Fixes #24603

Main Issue: #24600

PIP: #xyz 

### Motivation
- Reduce lock contention and fix thread-safety issues in BucketDelayedDeliveryTracker.
- Remove blocking I/O from the publish path.
- Decouple the tracker from dispatcher for easier testing and benchmarking.

### Modifications
- Concurrency
  - Replace StampedLock and synchronized with ReentrantReadWriteLock (read/write separation).
  - addMessage uses “read-lock check → write-lock modify” to minimize write lock time.
- Asynchronous snapshots
  - Seal-and-swap the current bucket and persist snapshots on a single-thread executor.
  - Ensure only one snapshot is created at a time.
- Scheduling and timer
  - getScheduledMessages triggers async loading of next snapshot segment instead of blocking.
  - Reworked timer handling with a dedicated lock and scheduleImmediateRun; no synchronization on dispatcher.
- Decoupling
  - Introduce DelayedDeliveryContext:
    - DispatcherDelayedDeliveryContext (production).
    - NoopDelayedDeliveryContext (tests/benchmarks).
  - Trackers and factories accept dispatcher or context; added constructors for name+cursor.
- Merging and lifecycle
  - asyncMergeBucketSnapshot uses write lock; merges bitmaps; removes old buckets; improved stats/logging.
  - clear() and close() under write lock; wait for snapshot futures outside locks; shutdown snapshot executor.
- Tests and benchmarks
  - Expanded unit and thread-safety tests; made MockManagedCursor public.
  - New JMH suite (BucketDelayedDeliveryTrackerBenchmark) and MockBucketSnapshotStorage; removed old simple benchmark.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
[...]

PR #24928, opened 2025-11-01
[feat][pip] PIP-448: Topic-level Delayed Message Tracker for Memory Optimization

Denovo1998:
Main Issue: #24600

### Motivation

The primary motivation for this proposal is to address the high memory consumption caused by the current per-subscription delayed message tracking mechanism. For topics with hundreds or thousands of subscriptions, the memory footprint for delayed messages becomes prohibitively large. Each delayed message's position is duplicated across every subscription's tracker, leading to a memory usage pattern of `O(num_delayed_messages * num_subscriptions)`.

This excessive memory usage can cause:
*   Increased memory pressure on Pulsar brokers.
*   More frequent and longer Garbage Collection (GC) pauses, impacting broker performance.
*   Potential OutOfMemoryErrors, leading to broker instability.
*   Limited scalability for use cases that rely on many subscriptions per topic, such as IoT or large-scale microservices with shared subscriptions.

By optimizing the delayed message tracking to be more memory-efficient, we can enhance broker stability and scalability, allowing Pulsar to better support these critical use cases.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pip/pip-448.md
+++ b/pip/pip-448.md
@@ -4,15 +4,13 @@
 
 In Apache Pulsar, **Delayed Message Delivery** allows producers to specify a delay for a message, ensuring it is not delivered to any consumer until the specified time has passed. This is a useful feature for implementing tasks like scheduled reminders or retry mechanisms with backoff.
 
-Currently, the default mechanism for handling delayed messages is the **In-Memory Delayed Delivery Tracker** (`InMemoryDelayedDeliveryTracker`). This tracker is instantiated on a *per-subscription* basis within the broker. When a topic has multiple subscriptions (e.g., in a shared subscription model), each subscription gets its own independent `InMemoryDelayedDeliveryTracker` instance.
+The legacy default mechanism for handling delayed messages is the `InMemoryDelayedDeliveryTracker`. This tracker is instantiated on a *per-subscription* basis within the broker. When a topic has multiple subscriptions, each subscription gets its own independent `InMemoryDelayedDeliveryTracker` instance.
 
-The consequence of this design is that if a delayed message is published to a topic with 'N' subscriptions, that message's metadata (its position: ledgerId and entryId) is stored 'N' times in the broker's memory, once for each subscription's tracker. This leads to significant memory overhead, especially for topics with a large number of subscriptions, as the memory usage scales linearly with the number of subscriptions.
-
-The **`DelayedDeliveryTrackerFactory`** is responsible for creating these tracker instances whenever a new subscription dispatcher is initialized.
+The consequence of this per-subscription design is that if a delayed message is published to a topic with 'N' subscriptions, that message's metadata (its position) is stored 'N' times in the broker's memory. This leads to significant memory overhead, especially for topics with a large number of subscriptions, as the memory usage scales linearly with the number of subscriptions.
 
 # Motivation
 
-The primary motivation for this proposal is to address the high memory consumption caused by the current per-subscription delayed message tracking mechanism. For topics with hundreds or thousands of subscriptions, the memory footprint for delayed messages becomes prohibitively large. Each delayed message's position is duplicated across every subscription's tracker, leading to a memory usage pattern of `O(num_delayed_messages * num_subscriptions)`.
+The primary motivation for this proposal is to address the high memory consumption caused by the legacy per-subscription delayed message tracking mechanism. For topics with hundreds or thousands of subscriptions, the memory footprint for delayed messages can become prohibitively large. Each delayed message's position is duplicated across every subscription's tracker, leading to a memory usage pattern of `O(num_delayed_messages * num_subscriptions)`.
 
 This excessive memory usage can cause:
 *   Increased memory pressure on Pulsar brokers.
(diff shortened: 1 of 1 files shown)


Technoboy-:
I think this is a good improvement, but based on this improvement, is there a test or benchmark that can visually demonstrate the results?

Denovo1998:
Yes, as in https://github.com/apache/pulsar/pull/24739, a JMH benchmarks is needed:

https://github.com/Denovo1998/pulsar/blob/738d0e2c41cdf7cc727789f348ebeda4089b9353/microbench/src/main/java/org/apache/pulsar/broker/delayed/bucket/BucketDelayedDeliveryTrackerBenchmark.java

Denovo1998:
I recently made some simple adjustments to OpenMessaging Benchmark to support the delayed message feature, while testing many PRs related to the delayed message module.

https://github.com/apache/pulsar/issues/24600#issuecomment-3591687762

For this PR, we must first monitor memory, but the most important thing is to test whether a problem with one subscription under a large number of subscriptions affects other subscriptions.

Any good suggestions for the result comparison part? Which indicators should we visualize?
@Technoboy- @codelipenghui @lhotari @coderzc

pip/pip-448.md
```diff
@@ -0,0 +1,166 @@
...
+*   Preserve the existing per-subscription `InMemoryDelayedDeliveryTrackerFactory` as the default for backward compatibility, requiring operators to opt-in to use the new topic-level tracker.
+
+## Out of Scope
+*   This proposal does not modify the persistent, bucket-based delayed delivery tracker (`BucketDelayedDeliveryTracker`).
+*   No changes will be made to the public-facing client APIs, REST APIs, or the wire protocol. This is a broker-internal optimization.
+*   The semantic behavior of delayed messages from a user's perspective will remain identical.
+
+# High Level Design
+
+The core idea is to introduce a new, opt-in `DelayedDeliveryTrackerFactory` that implements a shared, topic-level tracking strategy. This is achieved with two new components: a `TopicDelayedDeliveryTrackerManager` and a subscription-scoped `InMemoryTopicDelayedDeliveryTracker`.
```
coderzc, line 39: Why not reference `DelayedDeliveryTracker` at the interface level, which would also apply to `BucketDelayedDeliveryTracker`?

Denovo1998: What you mean is why not to implement DelayedDeliveryTracker directly, but to add a TopicDelayedDeliveryTrackerManager?

pip/pip-448.md
```diff
@@ -0,0 +1,166 @@
...
+*   **Data Structure**: Uses a `ConcurrentSkipListMap<Long, Long2ObjectRBTreeMap<Roaring64Bitmap>>` as its core index, mapping delivery timestamps to message positions efficiently.
+*   **Subscription Context (`SubContext`)**: Holds per-subscription state, including the `markDeletePosition`.
+*   **Pruning Logic**: Pruning is throttled and tunable:
+    *   The minimum time between pruning attempts is controlled by `delayedDeliveryPruneMinIntervalMillis`. If not set, it uses an adaptive interval based on the tick time.
+    *   Opportunistic pruning is triggered during the delivery check if the ratio of subscriptions ready for delivery exceeds `delayedDeliveryPruneEligibleRatio`.
+*   **`createOrGetTracker`**: This method creates the per-subscription `InMemoryTopicDelayedDeliveryTracker` instance.
+
+### `InMemoryTopicDelayedDeliveryTracker` (New Class)
+This class implements the `DelayedDeliveryTracker` interface for a single subscription.
+*   **Role**: Acts as a lightweight proxy, forwarding all operations (e.g., `addMessage`, `getScheduledMessages`) to the shared `InMemoryTopicDelayedDeliveryTrackerManager`.
```
coderzc, line 92: Whether it's possible to consider using a special cursor to replay delayed messages to build shared delayed message tracker, instead of relying on the subscription-level DelayedDeliveryTracker to forward messages?

We can check and poll the scheduled message IDs from the shared delayed message tracker to add to the replay queue (MessageRedeliveryController) of all subscriptions.

pip/pip-448.md
coderzc: Oh I see，you have designed a new data structure for the InMemoryTopicDelayedDeliveryTrackerManager, unable to adapt to other DelayedDeliveryTracker.

Denovo1998: Yes.

At the code level, the new TopicDelayedDeliveryTrackerManager already exposes createOrGetTracker(AbstractPersistentDispatcherMultipleConsumers) returning the DelayedDeliveryTracker interface:

```java
public interface TopicDelayedDeliveryTrackerManager extends AutoCloseable {
    DelayedDeliveryTracker createOrGetTracker(AbstractPersistentDispatcherMultipleConsumers dispatcher);
    // ...
}
```

so the dispatcher only depends on the DelayedDeliveryTracker interface and not on the concrete in-memory implementation. The InMemoryTopicDelayedDeliveryTracker is just one implementation that acts as a proxy to the shared topic-level manager in this PIP.

In the PIP text I focused on the in-memory implementation (e.g. InMemoryTopicDelayedDeliveryTracker*) because this proposal explicitly keeps BucketDelayedDeliveryTracker out of scope (see the “Out of Scope” section). The goal of PIP-448 is to address the memory footprint of the legacy in-memory tracker first, without changing any semantics of the persistent, bucket-based tracker.

PR #24929, opened 2025-11-02
[fix][test] Stabilize testMsgDropStat by reliably triggering non-persistent publisher drop

### LINKED ISSUE #24908: Flaky-test: NonPersistentTopicTest.testMsgDropStat
### Search before reporting

### Example failure

https://github.com/apache/pulsar/actions/runs/18747937935/job/53480679625#step:11:1392

### Exception stacktrace

```text
java.lang.AssertionError: expected [true] but found [false]
	at org.apache.pulsar.client.api.NonPersistentTopicTest.testMsgDropStat(NonPersistentTopicTest.java:905)
```

vinkal-chudgar:
Fixes #24908

### Motivation

`NonPersistentTopicTest.testMsgDropStat` fails intermittently on CI:

```
java.lang.AssertionError: expected [true] but found [false]
    at org.apache.pulsar.client.api.NonPersistentTopicTest.testMsgDropStat(NonPersistentTopicTest.java:905)
```
**Failure point:** `assertTrue(latch.await(5, TimeUnit.SECONDS))` times out because no publisher drop is observed within 5 seconds (no `MessageIdImpl` with `entryId == -1`).

**Root cause (in the original test loop)**
https://github.com/apache/pulsar/blob/fbc50b06ada0af9edf23b4b75b7bf2f96b0b1c67/pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java#L882-L905

**Why it flakes:**
- The test sets `maxConcurrentNonPersistentMessagePerConnection = 1`. In `ServerCnx#handleSend`, the broker drops only when `nonPersistentPendingMessages > maxNonPersistentMessagePerConnection`. With `maxConcurrentNonPersistentMessagePerConnection = 1`, the first concurrent send increments the counter from 0 to 1 and is accepted. The second concurrent send checks 1 > 1 (false), then increments the counter to 2 and is accepted. Only a third overlapping send sees 2 > 1 and is dropped. Therefore, at least three overlapping sends are required to trigger any drop. The original loop does not guarantee that level of overlap.
https://github.com/apache/pulsar/blob/fbc50b06ada0af9edf23b4b75b7bf2f96b0b1c67/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L1999-L2013
- The executor threads that call `producer.send()` are not coordinated to start together, so overlap of `send()` calls on the same connection is left to scheduler timing and is therefore probabilistic. 
- The loop uses `Thread.sleep(10)`, introducing timing assumptions that vary under CI load.
- Latch gating creates an additional race: The latch is decremented only when a drop is observed after 20% of messages have been sent. If drops occur before the 20% threshold and later sends do not drop, the latch never opens and the test times out at the await. (See NonPersistentTopicTest.java lines 0887-0894 for the latch condition and line 0905 for the timeout.)

This PR makes the publisher-drop trigger reliable without changing the test scenario or API usage.

### Modifications

Test-only changes in `NonPersistentTopicTest.testMsgDropStat`. No production code or docs changed. 

-  **Replace the unsynchronized, sleep-based loop with a small synchronized batch of `send()` calls.**
Use a `CyclicBarrier` to start a fixed number of `producer.send()` calls at the same time, explicitly creating the required overlap without sleeps or ad-hoc timing. Execute the batch inside a bounded `Awaitility` wait until at least one `send()` returns a `MessageIdImpl` with `entryId == -1`, then assert that condition.

- **Changed `maxConcurrentNonPersistentMessagePerConnection` from 1 to 0**

  `conf.setMaxConcurrentNonPersistentMessagePerConnection(0);`
  
  **Rational:** In `ServerCnx#handleSend` for non-persistent topics, a publish is dropped only when `nonPersistentPendingMessages > maxNonPersistentMessagePerConnection`. Lowering the limit from 1 to 0 reduces the overlap requirement from three in-flight sends to two, which, combined with the synchronized start, makes the publisher drop reliably detectable in this test.

### Verifying this change

**Personal CI Results**

Tested in Personal CI fork: https://github.com/vinkal-chudgar/pulsar/pull/3

Status: All checks have passed (48 successful checks, 3 skipped)

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/vinkal-chudgar/pulsar/pull/3

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
@@ -921,11 +921,9 @@ public class NonPersistentTopicTest extends ProducerConsumerBase {
                 }
 
                 // Wait for all sends to complete.
-                completionLatch.await();
+                assertTrue(completionLatch.await(20, TimeUnit.SECONDS));
 
-                if (error.get() != null) {
-                    throw new AssertionError("Concurrent send encountered an exception", error.get());
-                }
+                assertNull(error.get(), "Concurrent send encountered an exception");
                 return publisherDropSeen.get();
             });
 


pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
```diff
@@ -873,36 +879,57 @@ public void testMsgDropStat() throws Exception {
...
-                        throw new RuntimeException(e);
-                    } catch (InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        throw new RuntimeException(e);
-                    }
-                });
-            }
-            assertTrue(latch.await(5, TimeUnit.SECONDS));
+                // Wait for all sends to complete.
+                completionLatch.await();
```
lhotari:
```suggestion
                assertTrue(completionLatch.await(20, TimeUnit.Seconds));
```

pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
```diff
@@ -873,36 +879,57 @@ public void testMsgDropStat() throws Exception {
...
+            /*
+             * Trigger at least one publisher drop through concurrent send() calls.
+             *
+             * Uses CyclicBarrier to ensure all threads send simultaneously, creating overlap.
+             * With maxConcurrentNonPersistentMessagePerConnection = 0, ServerCnx#handleSend
+             * drops any send while another is in-flight, returning MessageId with entryId = -1.
+             * Awaitility repeats whole bursts (bounded to 20s) until a drop is observed.
+             */
+            AtomicBoolean publisherDropSeen = new AtomicBoolean(false);
+            Awaitility.await().atMost(Duration.ofSeconds(20)).until(() -> {
```
lhotari, line 897: I don't think using Awaitility is needed here. Awaitility is only used when there's some sort of polling and retry loop.

pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
```diff
@@ -873,36 +879,57 @@ public void testMsgDropStat() throws Exception {
...
-                    }
-                });
-            }
-            assertTrue(latch.await(5, TimeUnit.SECONDS));
+                // Wait for all sends to complete.
+                completionLatch.await();
+
+                if (error.get() != null) {
+                    throw new AssertionError("Concurrent send encountered an exception", error.get());
+                }
```
lhotari: use `assertNull` instead

pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
vinkal-chudgar: Agreed. updated in bfcd6965f7e4f562dac148068fda724f05aed0f8

pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
vinkal-chudgar: Agreed. updated in bfcd6965f7e4f562dac148068fda724f05aed0f8

pulsar-broker/src/test/java/org/apache/pulsar/client/api/NonPersistentTopicTest.java
lhotari: I guess this is fine if the intention is to retry for 20 second until something is dropped.

vinkal-chudgar: Thanks for the follow-up @lhotari . Yes, that is the intention: the test retries for up to 20 seconds with Awaitility until a publisher drop is observed.

vinkal-chudgar:
@lhotari - Thank you for the review. I’ve addressed the comments. Could you please take a look when convenient?

lhotari (approved):

PR #24920, opened 2025-10-30
[fix][broker] Add `internalAddConsumer` to help track StackOverflowError issue

Technoboy-:
### Motivation

We see the below StackOverflowError log, but we don't have the heapdump for it, so try to add `internalAddConsumer` to log to help find the root cause.
```
2025-10-28T22:54:08,431+0000 [configuration-metadata-store-14-1] ERROR org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/default/test-partition-0] Failed to create subscription: test-subscription
java.util.concurrent.CompletionException: java.lang.StackOverflowError
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315) ~[?:?]
	at java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1194) ~[?:?]
	at java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309) ~[?:?]
	at org.apache.pulsar.broker.service.AbstractDispatcherSingleActiveConsumer.addConsumer(AbstractDispatcherSingleActiveConsumer.java:171) 
	at org.apache.pulsar.broker.service.AbstractDispatcherSingleActiveConsumer.lambda$addConsumer$2(AbstractDispatcherSingleActiveConsumer.java:176) 
	at java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187) ~[?:?]
	at java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309) ~[?:?]
	at org.apache.pulsar.broker.service.AbstractDispatcherSingleActiveConsumer.addConsumer(AbstractDispatcherSingleActiveConsumer.java:171)
	at org.apache.pulsar.broker.service.AbstractDispatcherSingleActiveConsumer.lambda$addConsumer$2(AbstractDispatcherSingleActiveConsumer.java:176) 
	at java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187) ~[?:?]
	at java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309) ~[?:?]
	at org.apache.pulsar.broker.service.AbstractDispatcherSingleActiveConsumer.addConsumer(AbstractDispatcherSingleActiveConsumer.java:171) 
	at 
...
[...]

PR #24934, opened 2025-11-03
[fix][broker] Fix stack overflow caused by race condition when closing a connection

BewareMyPower:
### Motivation

See https://github.com/apache/pulsar/pull/24920#discussion_r2486182394

### Modifications

- Add `testOverrideInactiveConsumer` to simulate the race condition by injecting a delay on `channelInactive` via the broker interceptor.
- Add a retry logic to `addConsumer` to wait for `channelInactive` is done. When the retry count reaches the limit, fail the  method

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java
@@ -192,7 +192,7 @@ public abstract class AbstractDispatcherSingleActiveConsumer extends AbstractBas
                             // A race condition happened in `ServerCnx#channelInactive`
                             // 1. `isActive` was set to false
                             // 2. `consumer.close()` is called
-                            // We should wait for the
+                            // We should wait until the consumer is closed, retry for some times
                             log.warn("[{}] race condition happened that cnx of the active consumer ({}) is inactive "
                                     + "but it's not removed, retrying", getName(), actConsumer);
                             final var future = new CompletableFuture<Void>();


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java
```diff
@@ -161,7 +163,21 @@ private NavigableMap<Integer, Integer> makeHashRing(int consumerSize) {
...
+    public CompletableFuture<Void> addConsumer(Consumer consumer) {
+        return internalAddConsumer(consumer, 0);
+    }
+
+    private synchronized CompletableFuture<Void> internalAddConsumer(Consumer consumer, int retryCount) {
+        if (retryCount >= MAX_RETRY_COUNT_FOR_ADD_CONSUMER_RACE) {
+            log.warn("[{}] The active consumer's connection is still inactive after all retries, remove {} by force",
+                    getName(), consumer);
+            try {
+                removeConsumer(consumer);
```
lhotari: I don't think that `removeConsumer` could be used directly to remove a consumer here since it won't signal in any way to the client that the consumer has been closed and other other data structures in parent levels don't get updated consistently.

BewareMyPower: Okay, so I think in this case, we should return an error. This is a rare case. In case it happens, we could unload the topic to remove the consumer by force.

lhotari: Instead of removing, adding should simply fail after the 5 attempts?

BewareMyPower: Yes, please check the latest test for the updated behavior

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java
```diff
@@ -161,7 +163,21 @@ private NavigableMap<Integer, Integer> makeHashRing(int consumerSize) {
         return Collections.unmodifiableNavigableMap(hashRing);
     }
 
-    public synchronized CompletableFuture<Void> addConsumer(Consumer consumer) {
+    public CompletableFuture<Void> addConsumer(Consumer consumer) {
+        return internalAddConsumer(consumer, 0);
+    }
+
+    private synchronized CompletableFuture<Void> internalAddConsumer(Consumer consumer, int retryCount) {
```
lhotari, line 170: Is this ever called with retryCount+1?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java
```diff
@@ -171,12 +187,26 @@ public synchronized CompletableFuture<Void> addConsumer(Consumer consumer) {
...
+                            // A race condition happened in `ServerCnx#channelInactive`
+                            // 1. `isActive` was set to false
+                            // 2. `consumer.close()` is called
+                            // We should wait for the
+                            log.warn("[{}] race condition happened that cnx of the active consumer ({}) is inactive "
+                                    + "but it's not removed, retrying", getName(), actConsumer);
+                            final var future = new CompletableFuture<Void>();
+                            CompletableFuture.delayedExecutor(100, TimeUnit.MILLISECONDS)
+                                    .execute(() -> future.complete(null));
+                            return future.thenCompose(__ -> addConsumer(consumer));
```
lhotari, line 206:
```suggestion
                            return future.thenCompose(__ -> internalAddConsumer(consumer, retryCount + 1));
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractDispatcherSingleActiveConsumer.java
```diff
@@ -171,12 +187,26 @@ public synchronized CompletableFuture<Void> addConsumer(Consumer consumer) {
...
+                        } else {
+                            return addConsumer(consumer);
```
lhotari, line 208:
```suggestion
                            return internalAddConsumer(consumer, retryCount + 1);
```

PR #24938, opened 2025-11-03
[fix][broker]Wrong backlog: expected 0 but got 1

poorbarcode:
### Motivation
- The backlog was calculated by `managedLedger.entriesAddedCounter - cursor.messagesConsumedCounter`
- When resetting a subscription, the broker resets the value `cursor.messagesConsumedCounter` this way: `{new value} = {old value} - {how many positions the old value is larger than the new value}`. 
- The method `managedLegder.getNumberOfEntries` is used to calculate the number between two positions, but it has a bug, which will get a larger value than expected if the entry ID is a negative value.
- You can reproduce the issue by the new test

- **[1]** : https://github.com/apache/pulsar/blob/v4.0.7/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3730C5-L3764C6
```java
public long getNumberOfEntries(Range<Position> range) {
    long count = 0;
    // If the from & to are pointing to different ledgers, then we need to :
    // 1. Add the entries in the ledger pointed by toPosition
    count += toPosition.getEntryId(); // When the entry id is -1, it plush "-1" in mistake
    count += toIncluded ? 1 : 0;
    ...
    ...
    return count;
}
```

### Modifications

- Fix the issue
- Improve the method `position.comparate`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
@@ -4994,9 +4994,11 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
         assertNotEquals(ledger4.getLedgerId(), ml2.currentLedger.getId());
         Range<Position> range131 = Range.closed(positions.get(0), PositionFactory.create(ledger4.getLedgerId(), 100));
         assertEquals(ml2.getNumberOfEntries(range131), 131);
-        Range<Position> range132 = Range.openClosed(positions.get(0), PositionFactory.create(ledger4.getLedgerId(), 100));
+        Range<Position> range132 = Range.openClosed(positions.get(0), PositionFactory.create(ledger4.getLedgerId(),
+            100));
         assertEquals(ml2.getNumberOfEntries(range132), 130);
-        Range<Position> range133 = Range.closedOpen(positions.get(0), PositionFactory.create(ledger4.getLedgerId(), 100));
+        Range<Position> range133 = Range.closedOpen(positions.get(0), PositionFactory.create(ledger4.getLedgerId(),
+            100));
         assertEquals(ml2.getNumberOfEntries(range133), 130);
 
         // cleanup.


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1453,6 +1453,23 @@ public Position getFirstPosition() {
...
+    private int comparePositions(Position pos1, Position pos2) {
+        if (pos1 == null || pos2 == null) {
+            throw new IllegalArgumentException("Positions must not be null");
+        }
+        if (pos1.getLedgerId() == pos2.getLedgerId() && pos1.getEntryId() == pos2.getEntryId()) {
+            return 0;
+        }
+        if (!ledger.isValidPosition(pos1) || !ledger.isValidPosition(pos2)) {
+            return ledger.getNextValidPosition(pos1).compareTo(ledger.getNextValidPosition(pos2));
+        }
```
lhotari: Why is getNextValidPosition call made for both positions? Could this be handled in way so that getNextValidPosition is called for pos1 and pos2 separately only if it's not valid? The result would be slightly different, for example in cases where pos2 is already a valid position and pos1 becomes the same as pos2 only by calling getNextValidPosition.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1453,6 +1453,23 @@ public Position getFirstPosition() {
         return firstLedgerId == null ? null : PositionFactory.create(firstLedgerId, 0);
     }
 
+    /**
+     * Compare two positions. It is different with {@link Position#compareTo(Position)} when the params are invalid.
+     * For example: position-1 is "1:{latest entry}", and position-2 is "2:0", they are the same position.
```
lhotari: how is this achieved? could you add a test to validate this?

poorbarcode: Sorry, the comment should be `position-1 is "1:{latest entry}", and position-2 is "2:-1"`, I have corrected it and added more and more tests

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1489,7 +1506,10 @@ public void operationComplete() {
                 // modify mark delete and read position since we are able to persist new position for cursor
                 lock.writeLock().lock();
                 try {
-                    if (markDeletePosition.compareTo(newMarkDeletePosition) >= 0) {
+                    int compareRes = comparePositions(markDeletePosition, newMarkDeletePosition);
+                    if (compareRes == 0) {
```
Technoboy-: So we can remove this line

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
Technoboy-: yes, + 1

poorbarcode: > Why is getNextValidPosition call made for both positions? Could this be handled in way so that getNextValidPosition is called for pos1 and pos2 separately only if it's not valid? The result would be slightly different, for example in cases where pos2 is already a valid position and pos1 becomes the same as pos2 only by calling getNextValidPosition.

I can not fully understand the comment, could you write a demo code?

lhotari:
```suggestion
        boolean isPos1Valid = ledger.isValidPosition(pos1);
        boolean isPos2Valid = ledger.isValidPosition(pos2);
        if (!isPos1Valid && !isPos2Valid) {
            return ledger.getNextValidPosition(pos1).compareTo(ledger.getNextValidPosition(pos2));
        } else if (!isPos1Valid) {
            return ledger.getNextValidPosition(pos1).compareTo(pos2);
        } else if (!isPos2Valid) {
            return pos1.compareTo(ledger.getNextValidPosition(pos2));
        }
```

@poorbarcode this is what I mean with my previous comment.

lhotari: > I can not fully understand the comment, could you write a demo code?

Please see https://github.com/apache/pulsar/pull/24938/files#r2517541713

poorbarcode: I think the current implementation is simpler and clearer. And the following code you shared has changed the object that was used to compare, which is wrong

```java
            return ledger.getNextValidPosition(pos1).compareTo(pos2);
...
            return pos1.compareTo(ledger.getNextValidPosition(pos2));
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: Removed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: > And the following code you shared has changed the object that was used to compare, which is wrong

Why is it wrong? Explaining that in more detail could help understand the purpose of your code.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: That's a very useful detail. The comparison logic makes more sense now. I asked one more question there just to get a clarification.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
lhotari: I see, this detail might resolve this: https://github.com/apache/pulsar/pull/24938#discussion_r2516930977.

poorbarcode: > Why is it wrong? Explaining that in more detail could help them understand the purpose of your code.

Regarding [the code you wrote above](https://github.com/apache/pulsar/pull/24938#discussion_r2517541713) , there is an example:

- ledgers: `[{ledger-1: 10 entris}, {ledger-2: 10 entries}]`
- `position-1`: `2 : -1`, which is the same as `1: 9`
- `position-2`: `2 : 0`
- the correct result of comparing is `-1`
- but your code will return `0`

poorbarcode: Added more test cases

Technoboy- (approved):

PR #24396, opened 2025-06-08
[improve][broker] enable consumer throtting and track unacked message for exlusive or failover subscription

berg223:
Fixes https://github.com/apache/pulsar/issues/24159

Main Issue: https://github.com/apache/pulsar/issues/24159

### Motivation

1. I have tried to fix the issue by https://github.com/apache/pulsar/pull/24376. However, it's not work when consumer use cumulative ack, batching mode, abort individual transaction. So I want to fix it further.
2. I have found another issue that flowcontrol of exclusive or failover consumer is not work. Since the issues are highly correlated, I want to fix them at the same time. 

<details>
<summary>We can reproduce flow control issue by this unit test</summary>

```
@Test(timeOut = 30000)
    public void testMaxUnackedMessagesOnExclusiveConsumer() throws Exception {
        final String topicName = testTopic + System.currentTimeMillis();
        final String subscriberName = "test-sub-exclusive" + System.currentTimeMillis();
        final int unackMsgAllowed = 100;
        final int receiverQueueSize = 10;
        final int totalProducedMsgs = 300;

        ConsumerBuilder<String> consumerBuilder = pulsarClient.newConsumer(Schema.STRING).topic(topicName)
                .subscriptionName(subscriberName).receiverQueueSize(receiverQueueSize)
                .ackTimeout(1, TimeUnit.MINUTES)
                .subscriptionType(SubscriptionType.Exclusive);
        @Cleanup
        Consumer<String> consumer = consumerBuilder.subscribe();
        // 1) Produced Messages
        @Cleanup
        Producer<String> producer = pulsarClient.newProducer(Schema.STRING).topic(topicName).create();
        for (int i = 0; i < totalProducedMsgs; i++) {
            String message = "my-message-" + i;
            producer.send(message);
        }
        // 2) Unlimited, so all messages can be consumed
        int count = 0;
        List<Message<String>> list = new ArrayList<>(totalProducedMsgs);
        for (int i = 0; i < totalProducedMsgs; i++) {
            Message<String> message = consumer.receive(1, TimeUnit.SECONDS);
            if (message == null) {
                break;
            }
            count++;
            list.add(message);
        }
        assertEquals(count, totalProducedMsgs);
        list.forEach(message -> {
            try {
                consumer.acknowledge(message);
            } catch (PulsarClientException e) {
            }
        });
        // 3) Set restrictions, so only part of the data can be consumed
        waitCacheInit(topicName);
[...]

PR #24944, opened 2025-11-04
[feat][client] oauth2 trustcerts file and timeouts

### LINKED ISSUE #24437: OAuth2 client auth PEM truststore
### Search before reporting

### Motivation

When using the OAuth2 client authentication plugin, there is no option to define the truststore.

Currently the only way is to set the Java truststore with `-Djavax.net.ssl.trustStore`.

The pulsar helm chart has recently switch from Java keystore to combined PEM files, so the truststore is dynamically updated without needed to restart the pulsar component

### Solution

Add a new option to define the truststore in `brokerClientAuthenticationParameters`, and the format.

### Alternatives

_No response_

### Anything else?

_No response_

gulecroc:
Fixes #24437

### Motivation

For client oauth2 authentication, allow to configure :
* connect timeout
* read timeout
* trust certs file

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [] `doc-required` 
- [ ] `doc-not-needed` 
- [x] `doc-complete` 

PR in forked repository:

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -567,6 +567,30 @@ allowOverrideEntryFilters=false
 # Max number of concurrent lookup request broker allows to throttle heavy incoming lookup traffic
 maxConcurrentLookupRequest=50000
 
+# Maximum heap memory for inflight topic list operations (MB).
+# Default: 100 MB (supports ~1M topic names assuming 100 bytes each)
+maxTopicListInFlightHeapMemSizeMB=100
+
+# Maximum direct memory for inflight topic list responses (MB).
+# Default: 100 MB (network buffers for serialized responses)
+maxTopicListInFlightDirectMemSizeMB=100
+
+# Timeout for acquiring heap memory permits (milliseconds).
+# Default: 25000 (25 seconds)
+maxTopicListInFlightHeapMemSizePermitsAcquireTimeoutMillis=25000
+
+# Maximum queue size for heap memory permit requests.
+# Default: 10000 (prevent unbounded queueing)
+maxTopicListInFlightHeapMemSizePermitsAcquireQueueSize=10000
+
+# Timeout for acquiring direct memory permits (milliseconds).
+# Default: 25000 (25 seconds)
+maxTopicListInFlightDirectMemSizePermitsAcquireTimeoutMillis=25000
+
+# Maximum queue size for direct memory permit requests.
+# Default: 10000 (prevent unbounded queueing)
+maxTopicListInFlightDirectMemSizePermitsAcquireQueueSize=10000
+
 # Max number of concurrent topic loading request broker allows to control number of zk-operations
 maxConcurrentTopicLoadRequest=5000
 
--- a/conf/proxy.conf
+++ b/conf/proxy.conf
@@ -228,6 +228,30 @@ maxConcurrentInboundConnectionsPerIp=0
 # Max concurrent outbound connections. The proxy will error out requests beyond that.
 maxConcurrentLookupRequests=50000
 
+# Maximum heap memory for inflight topic list operations (MB).
+# Default: 100 MB (supports ~1M topic names assuming 100 bytes each)
+maxTopicListInFlightHeapMemSizeMB=100
+
+# Maximum direct memory for inflight topic list responses (MB).
+# Default: 100 MB (network buffers for serialized responses)
+maxTopicListInFlightDirectMemSizeMB=100
+
+# Timeout for acquiring heap memory permits (milliseconds).
+# Default: 25000 (25 seconds)
+maxTopicListInFlightHeapMemSizePermitsAcquireTimeoutMillis=25000
+
+# Maximum queue size for heap memory permit requests.
+# Default: 10000 (prevent unbounded queueing)
+maxTopicListInFlightHeapMemSizePermitsAcquireQueueSize=10000
+
+# Timeout for acquiring direct memory permits (milliseconds).
+# Default: 25000 (25 seconds)
+maxTopicListInFlightDirectMemSizePermitsAcquireTimeoutMillis=25000
+
+# Maximum queue size for direct memory permit requests.
+# Default: 10000 (prevent unbounded queueing)
+maxTopicListInFlightDirectMemSizePermitsAcquireQueueSize=10000
+
 ##### --- TLS --- #####
 
 # Deprecated - use servicePortTls and webServicePortTls instead
--- a/conf/standalone.conf
+++ b/conf/standalone.conf
@@ -320,6 +320,30 @@ preciseDispatcherFlowControl=false
 # Max number of concurrent lookup request broker allows to throttle heavy incoming lookup traffic
 maxConcurrentLookupRequest=50000
 
+# Maximum heap memory for inflight topic list operations (MB).
+# Default: 100 MB (supports ~1M topic names assuming 100 bytes each)
+maxTopicListInFlightHeapMemSizeMB=100
+
+# Maximum direct memory for inflight topic list responses (MB).
+# Default: 100 MB (network buffers for serialized responses)
+maxTopicListInFlightDirectMemSizeMB=100
+
+# Timeout for acquiring heap memory permits (milliseconds).
+# Default: 25000 (25 seconds)
+maxTopicListInFlightHeapMemSizePermitsAcquireTimeoutMillis=25000
+
+# Maximum queue size for heap memory permit requests.
+# Default: 10000 (prevent unbounded queueing)
+maxTopicListInFlightHeapMemSizePermitsAcquireQueueSize=10000
+
+# Timeout for acquiring direct memory permits (milliseconds).
+# Default: 25000 (25 seconds)
+maxTopicListInFlightDirectMemSizePermitsAcquireTimeoutMillis=25000
+
+# Maximum queue size for direct memory permit requests.
+# Default: 10000 (prevent unbounded queueing)
+maxTopicListInFlightDirectMemSizePermitsAcquireQueueSize=10000
+
 # Max number of concurrent topic loading request broker allows to control number of zk-operations
 maxConcurrentTopicLoadRequest=5000
 
--- a/pip/pip-442.md
+++ b/pip/pip-442.md
@@ -14,8 +14,8 @@ Pulsar brokers already implement comprehensive memory management for most operat
 - `maxMessagePublishBufferSizeInMB` (default: 50% direct memory): Limits memory used for buffering messages during publishing, providing backpressure when producers exceed broker capacity
 
 **Managed Ledger Memory Limits:**
-- `managedLedgerMaxReadsInFlightSizeInMB` (default: 0, disabled): Controls memory allocation for concurrent read operations from BookKeeper, preventing excessive memory usage during high read loads. This limit extends to cover buffers that were read from BookKeeper and are waiting in channel outbound buffers to be written to client sockets. 
-- `managedLedgerCacheSizeMB` (default: 20% of direct memory): Limits cache memory for recently read ledger entries, ensuring predictable memory usage for read caching. This limit extends to cover buffers that were read from the cache and are waiting in channel outbound buffers to be written to client sockets. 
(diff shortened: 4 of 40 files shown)


gulecroc:
Doc : https://github.com/apache/pulsar-site/pull/1052

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
```diff
@@ -43,23 +43,51 @@ public static Authentication clientCredentials(URL issuerUrl, URL credentialsUrl
...
+     *                           and each string adds an additional access range to the requested scope.
+     *                           From here: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2
+     * @param connectTimeout     the connect timeout in milliseconds
+     * @param readTimeout        the read timeout in milliseconds
+     * @param trustCertsFilePath the path to the file that contains trusted certificates
+     * @return an Authentication object
+     */
+    public static Authentication clientCredentials(URL issuerUrl, URL credentialsUrl, String audience, String scope,
+                                                   Integer connectTimeout, Integer readTimeout,
+                                                   String trustCertsFilePath) {
```
lhotari: Since there are so many parameters, this method is problematic.

Passing a parameter object built using a builder would be a better approach. Since ClientCredentialsFlow is package private and shouldn't be exposed externally, it would require a new public parameter object class + builder to resolve this.

lhotari (commented):
Besides having an API for instantiating the `Authentication` instance, it is necessary to consider how the `Authentication` instance would be instantiated when parameters are passed in `authParams` (for example in `client.conf`). 

You should also update https://github.com/apache/pulsar/blob/b402985db972d32021eb354f160ce7a5ed443e39/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/ClientCredentialsFlow.java#L117-L134 so that https://github.com/apache/pulsar/blob/671994f74cea4496a95c5a033b2d1ae2cfae5347/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2.java#L67-L86 would work.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
gulecroc: builder added

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
```diff
@@ -43,23 +44,53 @@ public static Authentication clientCredentials(URL issuerUrl, URL credentialsUrl
...
+     *                           case-sensitive strings. The strings are defined by the authorization server.
+     *                           If the value contains multiple space-delimited strings, their order does not matter,
+     *                           and each string adds an additional access range to the requested scope.
+     *                           From here: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2
+     * @param connectTimeout     the connect timeout in milliseconds
+     * @param readTimeout        the read timeout in milliseconds
+     * @param trustCertsFilePath the path to the file that contains trusted certificates
+     * @return an Authentication object
+     */
+    @Builder(builderMethodName = "clientCredentialsBuilder")
```
lhotari: It would be preferable not to use Lombok's Builder for Pulsar's client API since it's harder to maintain. 

We can use Lombok's Builder if we set the builderClassName to something meaningful.

By default, the builderClassName would be (ReturnTypeName)Builder, so that would be AuthenticationBuilder unless the return type is changed or builderClassName is set explicitly. The builder classname wouldn't contain the relevant context and maintaining the API backwards compatibility later would be harder if the builder class name isn't set to something useful.
In this case, since the builder is for AuthenticationOAuth2 with client credentials flow, one possibility would be to set `builderClassName` to `OAuth2WithClientCredentialsBuilder` (to keep it shorter).

gulecroc: I replace lombok builder. Since we are already in AuthenticationFactoryOAuth2, I named it only ClientCredentialsBuilder

lhotari (approved):
LGTM

pulsar-client/src/test/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2Test.java
lhotari, line 1: The license header is missing

gulecroc: licence added

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/DefaultMetadataResolver.java
```diff
@@ -109,4 +71,30 @@ public static URL getWellKnownMetadataUrl(URL issuerUrl) {
...
+                    .toCompletableFuture()
+                    .get();
+
+            Metadata metadata;
+            try (InputStream inputStream = response.getResponseBodyAsStream()) {
+                metadata = this.objectReader.readValue(inputStream);
+            }
+            return metadata;
+
+        } catch (IOException | InterruptedException | ExecutionException e) {
```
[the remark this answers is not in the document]
gulecroc: InterruptedException was not restored here too : https://github.com/gulecroc/pulsar/blob/edd1739f61fe75b8bb87ed4484a94fc89dbff8fa/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/TokenClient.java#L113
I will fix both

gulecroc:
I fix Copilot review

lhotari:
@gulecroc for future PRs, you might want to use ["Personal CI"](https://pulsar.apache.org/contribute/personal-ci/) to run Pulsar CI in your own fork to get early CI feedback when working on a larger PR like this one. The benefit is that you have full control over the CI and wouldn't have to wait for one of the project maintainers to approve a CI run.

Regarding flaky tests, in apache/pulsar CI, there is a way for anyone to trigger a retry run. That happens by adding "/pulsarbot rerun-failure-checks" comment on the PR. We have [a lot of flaky tests](https://github.com/apache/pulsar/issues?q=is%3Aissue%20state%3Aopen%20flaky) and retrying 1 to 3 times for a PR is common. Retrying can only be requested after all jobs in the workflow have completed.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
```diff
@@ -18,31 +18,80 @@
...
         this.issuerUrl = issuerUrl;
+        this.httpClient = defaultHttpClient(readTimeout, connectTimeout, trustCertsFilePath);
+    }
+
+    private AsyncHttpClient defaultHttpClient(Integer readTimeout, Integer connectTimeout, String trustCertsFilePath) {
+        DefaultAsyncHttpClientConfig.Builder confBuilder = new DefaultAsyncHttpClientConfig.Builder();
+        confBuilder.setCookieStore(null);
+        confBuilder.setUseProxyProperties(true);
+        confBuilder.setFollowRedirect(true);
+        confBuilder.setConnectTimeout(
```
lhotari, line 72: please check https://github.com/apache/pulsar-site/pull/1052/files#r2509758571 and the earlier comment about adding javadoc for connectTimeout/readTimeout

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
```diff
@@ -31,35 +31,101 @@ public final class AuthenticationFactoryOAuth2 {
...
+            this.readTimeout = readTimeout;
+            return this;
+        }
+
+        public ClientCredentialsBuilder trustCertsFilePath(String trustCertsFilePath) {
+            this.trustCertsFilePath = trustCertsFilePath;
+            return this;
+        }
+
+        public Authentication build() {
```
lhotari, line 171: please add javadoc to these methods. One useful detail is to document the unit for connectTimeout/readTimeout, that the unit is in seconds or milliseconds.

https://github.com/apache/pulsar-site/pull/1052/files#r2509758571

gulecroc:
Hi @lhotari, as I see AHC is moving to [Duration ](https://github.com/AsyncHttpClient/async-http-client/blob/43826107728bde0443202af2908a65fa3fa79473/client/src/main/java/org/asynchttpclient/config/AsyncHttpClientConfigDefaults.java#L118) for timeouts in v3.x, I update my code and documentation in this way

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
gulecroc: java doc added and timeouts moved to Duration

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
gulecroc: java doc added and timeouts moved to Duration

gulecroc:
Thank you @lhotari !
Can I carry over these changes to version 4.0.x LTS?

lhotari:
> Thank you @lhotari! Can I carry over these changes to version 4.0.x LTS?

@gulecroc Usually new features aren't immediately added to LTS versions to ensure that changes don't cause regressions. 
New features can be added to an LTS release by starting a discussion on the dev mailing list. In most cases, it's better to have a PIP available for referencing the changes.

I'm just wondering whether this feature is completed by this PR. **Is the intention to support updates for `trustCertsFilePath`?** In that case, more changes would be needed to add support for `trustCertsFilePath` changes. I think that decoupling from AsyncHttpClient would be useful and instead having an internal abstraction that hides implementation details.

One detail to take into account is that there is [PIP-337: SSL Factory Plugin to customize SSLContext/SSLEngine generation](https://github.com/apache/pulsar/blob/master/pip/pip-337.md), which is expected to be used for SSLContext/SSLEngine instance creation. For consistency, that solution should be used. It's also possible that the PIP-337 solution might require some changes to support this new use case.

I think it would be useful to write a PIP document to capture the requirements and at least the high-level design. A PIP will also help when starting the community discussion and decision-making to include changes to 4.0.x LTS regarding this area.

This might help [in creating a PIP](https://github.com/apache/pulsar/tree/master/pip#pulsar-improvement-proposal-pip): ([how to create an LLM generated draft](https://github.com/apache/pulsar/pull/24396#pullrequestreview-2909244330))

Btw. Regarding the AsyncHttpClient used in OAuth authentication, I'd also like it support PIP-234 changes so that the client doesn't create it's own set of threads, but has a way to reuse them from the PulsarClient / PulsarAdminClient. The solution for that is still missing. PIP-234 was planned a long time ago and it wasn't implemented until recently. The implementation is in PRs #24790, #24784 and #24893 and will be available in 4.1.2 version. More changes are needed so that there would be an interface to be used by authentication plugins for getting access to the shared resources. 
For authentication plugins, one possibility would be that there would be a way to directly get a HTTP client with given configuration (such as trustCertsFilePath) that is sufficiently abstracted so that there's no need to couple to AsyncHttpClient in authentication plugin implementations.

lhotari:
It should be fine to include in 4.0.9/4.1.3.

PR #24947, opened 2025-11-05
[fix][broker][branch-4.0] Fix failed testFinishTakeSnapshotWhenTopicLoading due to topic future cache conflicts

BewareMyPower:
### Motivation

The cherry-picking on a399b5ccd317282e8624135e49e08843c7196887 inserts the topic future cache duplicatedly.

1st time:
https://github.com/apache/pulsar/blob/74931c9e5f2f4384ed3b048101051dd8e5ea4a9a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1129

or

https://github.com/apache/pulsar/blob/74931c9e5f2f4384ed3b048101051dd8e5ea4a9a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1141

2nd time: https://github.com/apache/pulsar/blob/74931c9e5f2f4384ed3b048101051dd8e5ea4a9a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1146

In this case, topic future might fail with timeout. This error was exposed after https://github.com/apache/pulsar/commit/dd149d4ba4b9bc59bbc793fb083abe09252be5bf because the future completed with `LowOverheadTimeoutException` won't trigger the cache invalidation.

### Modifications

Move the `fetchPartitionedTopicMetadataAsync` check to `loadOrCreatePersistentTopic`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -1117,31 +1117,7 @@ public class BrokerService implements Closeable {
                     // The topic level policies are not needed now, but the meaning of calling
                     // "getTopicPoliciesBypassSystemTopic" will wait for system topic policies initialization.
                     getTopicPoliciesBypassSystemTopic(topicName, TopicPoliciesService.GetType.LOCAL_ONLY)
-                            .thenCompose(optionalTopicPolicies -> {
-                        if (topicName.isPartitioned()) {
-                            final TopicName topicNameEntity = TopicName.get(topicName.getPartitionedTopicName());
-                            return fetchPartitionedTopicMetadataAsync(topicNameEntity)
-                                    .thenCompose((metadata) -> {
-                                        // Allow creating non-partitioned persistent topic that name includes
-                                        // `partition`
-                                        if (metadata.partitions == 0
-                                                || topicName.getPartitionIndex() < metadata.partitions) {
-                                            return topics.computeIfAbsent(topicName.toString(), (tpName) ->
-                                                    loadOrCreatePersistentTopic(context));
-                                        } else {
-                                            final String errorMsg =
-                                                    String.format("Illegal topic partition name %s with max allowed "
-                                                            + "%d partitions", topicName, metadata.partitions);
-                                            log.warn(errorMsg);
-                                            return FutureUtil.failedFuture(
-                                                    new BrokerServiceException.NotAllowedException(errorMsg));
-                                        }
-                                    });
-                        } else {
-                            return topics.computeIfAbsent(topicName.toString(), (tpName) ->
-                                    loadOrCreatePersistentTopic(context));
-                        }
-                    }).thenRun(() -> {
+                            .thenRun(() -> {
                         final var inserted = new MutableBoolean(false);
                         final var cachedFuture = topics.computeIfAbsent(topicName.toString(), ___ -> {
                             inserted.setTrue();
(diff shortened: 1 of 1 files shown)


lhotari:
> In this case, topic future might fail with timeout. This error was exposed after [dd149d4](https://github.com/apache/pulsar/commit/dd149d4ba4b9bc59bbc793fb083abe09252be5bf) because the future completed with `LowOverheadTimeoutException` won't trigger the cache invalidation.

Just wondering if there should be code to handle this. Similar as in `createNonPersistentTopic`
https://github.com/apache/pulsar/blob/74931c9e5f2f4384ed3b048101051dd8e5ea4a9a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1389-L1395

BewareMyPower:
Well, there has been a debate before in https://github.com/apache/pulsar/pull/24500/files#r2199584548

The timeout failure usually represents a unknown state. Removing it from the cache cannot cancel the ongoing topic loading process in background. There could be two concurrent topic loading tasks if we added the logic, some tests might fail (you can see the changes in #24500). We have to take care of this case.

Topic loading for non-persistent topics is much more simple than persistent topics, so actually the timeout exception hardly happen.

lhotari:
> Well, there has been a debate before in https://github.com/apache/pulsar/pull/24500/files#r2199584548
> 
> The timeout failure usually represents a unknown state. Removing it from the cache cannot cancel the ongoing topic loading process in background. There could be two concurrent topic loading tasks if we added the logic, some tests might fail (you can see the changes in #24500). We have to take care of this case.
> 
> Topic loading for non-persistent topics is much more simple than persistent topics, so actually the timeout exception hardly happen.

@BewareMyPower Yes, I agree. I came to the same conclusion when looking into the code. However, there's also problems in the current CompletableFuture & timeout based solution. For example, if the topic has already opened the ledger when the timeout occurs, it won't be able to complete the future successfully. The topic needs to be closed after it just got successfully opened:

https://github.com/apache/pulsar/blob/39bb67542f2a7b849acaff681d408c693e1a2a18/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L1896-L1914

This doesn't make much sense. Instead of using CompletableFuture values in maps, a better approach would be to have some Entry class that holds more related state. There could be a CompletableFuture inside the Entry class.
In the topic loading timeout scenario, it doesn't make sense to terminate the topic loading if it's already opening the ledger.
The system would just get more loaded due to closing the topic again and reopening it in a retry.

BewareMyPower:
IMO, we should not apply a timeout on topic loading. When the timeout happens, there must be a bug or the CPU resource is exhausted. You can see `topicLoadTimeoutSeconds` is 60, which is already too long. Just imagine creating a producer and the future does not complete in 60 seconds, or a reconnection due to topic unloading didn't succeed in 60 seconds.

In production env, the timeout is usually caused by topic replay because the metadata store operation should not be such slow:
1. `__change_events`
2. or message deduplication replay on the topic

The 1st case is usually caused by a bug with system topic based topic policies service. It is hard to say whether retrying helps.

The 2nd case is usually caused by the snapshot was not taken in time. Retry the topic replay never helps for the 2nd case. That's also the motivation of https://github.com/apache/pulsar/pull/23004, which introduces `testFinishTakeSnapshotWhenTopicLoading`.

I've been working on this part before but I didn't have time recently. Here are some local notes I wrote before. https://gist.github.com/BewareMyPower/c7f26eb2b5741c94d8cf771c4dc35a97

PR #24965, opened 2025-11-10
[fix][client] Fix deduplication for getPartitionedTopicMetadata to include method parameters

### LINKED ISSUE #24963: [Bug] Deduplication logic for getPartitionedTopicMetadata doesn't take method parameters into account
### Search before reporting

### Read release policy

### User environment

All Pulsar client version since 3.0.6 / 3.3.1 with #22206 changes

### Issue Description

The deduplication logic in org.apache.pulsar.client.impl.BinaryProtoLookupService#getPartitionedTopicMetadata doesn't take the method parameters into account:
https://github.com/apache/pulsar/blob/0896c0a5fe23c57179faf8752b416bee5e43b6a2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/BinaryProtoLookupService.java#L188-L206

The impact of this is that `boolean metadataAutoCreationEnabled, boolean useFallbackForNonPIP344Brokers` parameters are ignored for any other requests which are made while there's a request in progress.

### Error messages

```text

```

### Reproducing the issue

Should be straightforward in a unit test

### Additional information

_No response_

vinkal-chudgar:
Fixes #24963

### Motivation

The in-flight deduplication for `BinaryProtoLookupService#getPartitionedTopicMetadata `used only `TopicName`. Calls that differ in `metadataAutoCreationEnabled` or `useFallbackForNonPIP344Brokers` were coalesced, which could return a future that does not match the requested behavior.

**Concrete example:**
- Thread A calls `getPartitionedTopicMetadata(topic, true, false)` and creates a future
- Thread B calls `getPartitionedTopicMetadata(topic, false, true) `and receives the same future from A
- Thread B gets a result that reflects `metadataAutoCreationEnabled=true`, which is incorrect

This PR fixes deduplication by keying in-flight requests on all method parameters.

### Modifications
- Implemented composite key class `PartitionedTopicMetadataKey` (static inner class) to uniquely identify requests by `topicName`, `metadataAutoCreationEnabled`, `useFallbackForNonPIP344Brokers`
- Changed `partitionedMetadataInProgress` to use `PartitionedTopicMetadataKey` as the map key instead of `TopicName`
- Used `computeIfAbsent` with the new key and removed the entry on completion with `remove(key, future)`
- Tests
   - Added BinaryProtoLookupServiceTest#testPartitionedMetadataDeduplicationAndCleanup
   - Added BinaryProtoLookupServiceTest#testPartitionedMetadataDeduplicationDifferentParameterCombinations

### Verifying this change

- Added BinaryProtoLookupServiceTest#testPartitionedMetadataDeduplicationAndCleanup
- Added BinaryProtoLookupServiceTest#testPartitionedMetadataDeduplicationDifferentParameterCombinations

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/vinkal-chudgar/pulsar/pull/5

### DIFF
--- a/pulsar-client/src/test/java/org/apache/pulsar/client/impl/BinaryProtoLookupServiceTest.java
+++ b/pulsar-client/src/test/java/org/apache/pulsar/client/impl/BinaryProtoLookupServiceTest.java
@@ -24,9 +24,13 @@ import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNotSame;
+import static org.testng.Assert.assertSame;
 import static org.testng.Assert.assertTrue;
 import static org.testng.Assert.fail;
 import io.netty.buffer.ByteBuf;
@@ -37,6 +41,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.pulsar.client.api.PulsarClientException.LookupException;
@@ -46,6 +51,7 @@ import org.apache.pulsar.client.impl.metrics.InstrumentProvider;
 import org.apache.pulsar.common.api.proto.BaseCommand;
 import org.apache.pulsar.common.api.proto.BaseCommand.Type;
 import org.apache.pulsar.common.naming.TopicName;
+import org.apache.pulsar.common.partition.PartitionedTopicMetadata;
 import org.apache.pulsar.common.protocol.Commands;
 import org.awaitility.Awaitility;
 import org.testng.annotations.AfterMethod;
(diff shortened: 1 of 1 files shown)


Technoboy-:
Why could take the method params into account ? There will be more request to the broker, but the result is the same.

lhotari:
> Why could take the method params into account ? There will be more request to the broker, but the result is the same.

@Technoboy- I think that this is more about correctness than the possibility that the broker response might be the same in most cases. It's strictly not the same in all cases. Therefore I think that such bugs should be fixed.

Technoboy-:
> > Why could take the method params into account ? There will be more request to the broker, but the result is the same.
> 
> @Technoboy- I think that this is more about correctness than the possibility that the broker response might be the same in most cases. It's strictly not the same in all cases. Therefore I think that such bugs should be fixed.

This is not a bug

lhotari:
> > > Why could take the method params into account ? There will be more request to the broker, but the result is the same.
> > 
> > 
> > @Technoboy- I think that this is more about correctness than the possibility that the broker response might be the same in most cases. It's strictly not the same in all cases. Therefore I think that such bugs should be fixed.
> 
> This is not a bug

@Technoboy-  I think that depends on how you define a bug. The impact of the issue  is that `boolean metadataAutoCreationEnabled, boolean useFallbackForNonPIP344Brokers` parameters for the `getPartitionedTopicMetadata` method are ignored for any other requests which are made while there's a request in progress. Let's say that the first request is made with `metadataAutoCreationEnabled=false` and while it's in progress a request with `metadataAutoCreationEnabled=true` is made. This will definitely result in an inconsistent situation. However, I agree that the likelihood that this would happen is low. Regardless of that, such issues should be fixed and I would consider them bugs. Your definition of a bug could vary and I think it's fine.

vinkal-chudgar:
> Why could take the method params into account ? There will be more request to the broker, but the result is the same.


@Technoboy- The reason we must include these parameters in the key is that they can lead to fundamentally different outcomes for the caller, breaking the API contract if not handled separately. The logic within `getPartitionedTopicMetadataAsync` demonstrates this.

```
if (!metadataAutoCreationEnabled && !clientCnx.isSupportsGetPartitionedMetadataWithoutAutoCreation()) {
    if (useFallbackForNonPIP344Brokers) {
        // Fall back to old behavior
        finalAutoCreationEnabled = true;
    } else {
        // This path fails immediately
        partitionFuture.completeExceptionally(new PulsarClientException.FeatureNotSupportedException(...));
        return;
    }
}
// This path sends a request to the broker
ByteBuf request = Commands.newPartitionMetadataRequest(topicName.toString(), requestId, finalAutoCreationEnabled);
```

Here is a practical example of why caching them separately is critical, as two calls with the same `TopicName` but different parameters can produce completely different results:

- **When the broker does not support PIP-344**, the outcome depends entirely on the `useFallbackForNonPIP344Brokers` flag:
    *   One call might fail immediately with a `FeatureNotSupportedException`.
    *   Another might fall back to the old behavior and succeed.

    If we only deduped by `TopicName`, a caller that expects a failure could incorrectly receive a successful result from the other caller. **This breaks the method contract for that caller.**

- **Even when the broker supports PIP-344**, the `metadataAutoCreationEnabled` flag changes the request sent. Coalescing requests with different values for this flag hides the caller’s chosen behavior and side effects (like topic auto-creation).

Regarding your point about increasing request volume: this change still correctly deduplicates all concurrent calls that use the exact same parameters. A new request is only sent when the parameters are different, ensuring we respect the caller's intent while still preventing redundant traffic for identical calls.

Technoboy-:
> > > > Why could take the method params into account ? There will be more request to the broker, but the result is the same.
> > > 
> > > 
> > > @Technoboy- I think that this is more about correctness than the possibility that the broker response might be the same in most cases. It's strictly not the same in all cases. Therefore I think that such bugs should be fixed.
> > 
> > 
> > This is not a bug
> 
> @Technoboy- I think that depends on how you define a bug. The impact of the issue is that `boolean metadataAutoCreationEnabled, boolean useFallbackForNonPIP344Brokers` parameters for the `getPartitionedTopicMetadata` method are ignored for any other requests which are made while there's a request in progress. Let's say that the first request is made with `metadataAutoCreationEnabled=false` and while it's in progress a request with `metadataAutoCreationEnabled=true` is made. This will definitely result in an inconsistent situation. However, I agree that the likelihood that this would happen is low. Regardless of that, such issues should be fixed and I would consider them bugs. Your definition of a bug could vary and I think it's fine.

Make sense

PR #24971, opened 2025-11-11
[fix][broker]Leaving orphan schemas and topic-level policies after partitioned topic is deleted by GC

poorbarcode:
### Motivation

**How to reproduce the issue**
- Start two clusters with a global configuration metadata store
- Enable `brokerDeleteInactiveTopicsEnabled` and `brokerDeleteInactivePartitionedTopicMetadataEnabled`
- After topic partitions and topics are deleted from the GC mechanism, the following events occur
  - `cluster 1`:  deletes all partitions
  - `cluster 2`: deletes all partitions
  - `cluster 1`: deletes partitioned topic metadata(the schema and topic-level policies will also be deleted with the operation)
  - `cluster 2`: since the partitioned topic metadata was deleted by `cluster 1`(both clusters share the same metadata), it deletes nothing, then the issue occurs: topics were deleted, but schemas and topic-level policies are still there

### Modifications

- fix the issue

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x


lhotari (approved):
LGTM

PR #24974, opened 2025-11-11
[fix][client] Fix retry logic for BinaryProtoLookupService.getTopicsUnderNamespace method

lhotari:
### Motivation

BinaryProtoLookupService.getTopicsUnderNamespace method contains retry logic, but it only applies to exceptions that occur when acquiring the connection fails.
A getTopicsUnderNamespace might fail later if the connection gets closed before the response has been successfully received. That's why it's better that the retry logic would also cover the actual operation to be effective.

### Modifications

- Fix the retry logic and add tests

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/utils/ClientChannelHelper.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/utils/ClientChannelHelper.java
@@ -21,7 +21,6 @@ package org.apache.pulsar.broker.service.utils;
 import com.google.common.collect.Queues;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.embedded.EmbeddedChannel;
-import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
 import java.util.Queue;
 import org.apache.pulsar.common.api.proto.BaseCommand;
 import org.apache.pulsar.common.api.proto.CommandAck;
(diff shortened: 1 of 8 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/utils/ClientChannelHelper.java
```diff
@@ -61,7 +61,7 @@ public class ClientChannelHelper {
 
     public ClientChannelHelper() {
         int maxMessageSize = 5 * 1024 * 1024;
```
Denovo1998, line 63: This can be removed.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/BinaryProtoLookupService.java
```diff
@@ -433,50 +435,56 @@ public CompletableFuture<GetTopicsResult> getTopicsUnderNamespace(NamespaceName
...
-                    getTopicsResultFuture.completeExceptionally(t);
+            ByteBuf request = Commands.newGetTopicsOfNamespaceRequest(namespace.toString(), requestId, mode,
+                            topicsPattern, topicsHash);
+            return clientCnx.newGetTopicsOfNamespace(request, requestId).whenComplete((r, t) -> {
+                client.getCnxPool().releaseConnection(clientCnx);
+            });
+        }, lookupPinnedExecutor).whenComplete((r, t) -> {
+            if (t != null) {
+                Throwable cause = FutureUtil.unwrapCompletionException(t);
+                if (cause instanceof PulsarClientException && !PulsarClientException.isRetriableError(cause)) {
```
Denovo1998, line 454: Is it covered in the test here?

PR #24975, opened 2025-11-12
[improve][broker]Improve error response of failed to delete topic if it has replicators connected

poorbarcode:
### Motivation

1. Create two clusters: `["c1", "c2"]`,  enable Geo-Replication `c1 -> c2`
2. Create a partitioned topic on both clusters.
3. On cluster `c2`: the response of `pulsar-admin topics stats <topic>` shows no consumers/producers connected.
4. On cluster `c2`: Call `pulsar-admin topics delete-partitioned-topic <topic>` will get the following error

```
pulsar-admin topics delete-partitioned-topic public/default/test-topic-1
2025-11-11T13:30:19,725+0000 [AsyncHttpClient-9-2] WARN  org.apache.pulsar.client.admin.internal.BaseResource - [http://local:8080/admin/v2/persistent/public/default/test-topic-1/partitions?force=false&deleteSchema=true] Failed to perform http delete request: javax.ws.rs.ClientErrorException: HTTP 412 {"reason":"Topic has active producers/subscriptions"}
```

Root cause: there is a replicator connected(from the cluster `c1`)

### Modifications

Improve the log
- before: `Topic has active producers/subscriptions`
- after: `Topic has active clients connected, including {num} consumers, {num} producers, {num} replicators`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -2780,7 +2780,7 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
             .whenComplete(((txnID, ex) -> {
                 if (ex == null) {
                     if (log.isDebugEnabled()) {
-                        log.debug("Send response {} for new txn request {}", tcId.getId(), requestId);
+                        log.debug("Send response {} for new txn request {}", txnID, requestId);
                     }
                     commandSender.sendNewTxnResponse(requestId, txnID, tcId.getId());
                 } else {


shibd (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1520,8 +1517,23 @@ private CompletableFuture<Void> delete(boolean failIfHasSubscriptions,
...
+                    if (!producers.isEmpty()) {
+                        replicatorCount = producers.values().stream().filter(Producer::isRemote).count();
+                        if (producers.size() > replicatorCount) {
+                            producerCount = producers.size() - replicatorCount;
+                        }
+                    }
+                    errorMsg.append(" Including").append(" ").append(consumerCount).append(" consumers,")
+                        .append(" ").append(producerCount).append(" producers,").append(" and")
+                        .append(" ").append(replicatorCount).append(" replicators.");
+                    return FutureUtil.failedFuture(new TopicBusyException(errorMsg.toString()));
```
Technoboy-, line 1536: Better to use String.format("xxx", x)

poorbarcode: Ah, they are the same. "String.format" throws an exception when placeholders do not match, while stringBuffer is relatively safe.

PR #24992, opened 2025-11-17
[fix][sec] Upgrade avro to 1.12.2

### LINKED ISSUE #25783: [sec] org.apache.avro:avro contains CVE-2025-33042
### Search before reporting

### Motivation

Summary
There is a reported security vulnerability in org.apache.avro:avro, which is used as a dependency in Apache Pulsar.

CVE: CVE-2025-33042
Affected dependency version in Pulsar:
[pulsar-3.0/pom.xml](https://github.com/apache/pulsar/blob/branch-3.0/pom.xml#L189)
[pulsar-4.0/pom.xml](https://github.com/apache/pulsar/blob/branch-4.0/pom.xml#L236)

 <avro.version>1.11.4</avro.version>

### Solution

Upgrade avro version to 1.12.1 or 1.11.5

### Alternatives

_No response_

### Anything else?

Could you please share the planned approach for handling this vulnerability in Apache Pulsar,

lhotari:
Fixes #25783

### Motivation

Avro 1.12.1 contains 4 security fixes: https://avro.apache.org/blog/2025/10/16/avro-1.12.1/

The upgrade was blocked on AVRO-4209, a `StackOverflowError` in `ReflectData.getSchema` for a POJO with a
recursive field, introduced by AVRO-3940 in 1.12.1. That is fixed in 1.12.2, so this goes from 1.12.0
straight to 1.12.2 rather than stopping at 1.12.1.

Going to 1.12.2 pulls in three behaviour changes that Pulsar has to handle.

**1. Avro 1.12.2 refuses to reflect over classes that are not explicitly trusted (AVRO-4189).**

It is worth separating the two releases, since the change that causes this landed in 1.12.2 rather than in
1.12.1. In 1.12.0 and 1.12.1 the trusted-class allow-list was consulted in exactly one place,
`SpecificDatumReader.findStringClass`, so it only guarded classes named by the `java-class` /
`java-key-class` schema properties. AVRO-4189 introduced `ClassSecurityValidator` and moved the check into
`ClassUtils.forName`, so it now guards *every* reflective class resolution, and the default trusted
packages are gone: what is left is 16 hardcoded JDK classes and no packages at all.

That matters because `SpecificData.getClass(Schema)` resolves a class for every RECORD, ENUM and FIXED
schema, and `ReflectData.populateEncoderCache` calls it from `getCustomEncoding` on both
`ReflectDatumWriter.write` and `ReflectDatumReader.read` (those `getCustomEncoding` call sites arrived
separately, in 1.12.1, via AVRO-4165). Pulsar's `AvroWriter` and `AvroReader` are built on
`ReflectDatum*`, so without an allow-list every `Schema.AVRO(...)` encode and decode throws
`SecurityException` — including the schemas Pulsar uses internally for its own system topics.

Scope, since it is narrower than it first looks: `Schema.JSON` is **not** affected — it derives an Avro
schema for the `SchemaInfo` but reads and writes through Jackson, so it never resolves a class
reflectively. `Schema.PROTOBUF` is affected, but at schema-construction time rather than on produce.

**2. The fast reader is enabled by default (AVRO-3230, 1.12.1).**

`GenericData.fastReaderEnabled` now defaults to `true`. The fast path reports malformed data as
`AvroTypeException` where the classic path raised `IndexOutOfBoundsException`, so decoding errors escaped
`GenericAvroReader.read` instead of being wrapped in `SchemaSerializationException`.

**3. avro-protobuf emits a `default` for enum types (AVRO-4133, 1.12.1).**

The schema JSON generated from a protobuf message with an enum field gains a `"default"` attribute.

### Modifications

- Upgrade Avro from 1.12.0 to 1.12.2 in the version catalog, and update the jar names in the
  `distribution/server` and `distribution/shell` `LICENSE.bin.txt` files.

- Add `org.apache.pulsar.client.schema.AvroTrustedClasses`, and **declare the application's classes
  automatically**. Passing a class to `Schema.AVRO(...)` is the application naming that class, so Pulsar
  trusts it and every type the derived schema references — nested records and enums, including ones in
  other packages, and the declared collection and `@Stringable` types the fields carry. Most applications
  therefore need no code change at all. Pulsar Functions and connectors are covered the same way, since
  they build their schemas through the same call, and so is the broker for the types it writes to its own
  system topics — no separate declaration is needed for either. `Schema.PROTOBUF` declares its message
  class too, before deriving the schema, since deriving is itself a reflective resolution.

  The hook is in `AvroSchema.of`, and only when the `SchemaDefinition` carries a POJO *and* no
  `jsonDef`. That condition is the trust boundary rather than a convenience: `SchemaUtil.createAvroSchema`
  gives `jsonDef` precedence, and `AutoConsumeSchema` feeds it raw schema-registry bytes, so expanding
  from a schema Pulsar did not derive would let whoever supplied that document choose which class names
  become trusted. Expansion works from the `SchemaInfo` just derived rather than by re-deriving from the
  class, since a differently configured `ReflectData` does not name the same types, and is cached by the
  schema that was walked.

  For what auto-registration deliberately does not cover, `AvroTrustedClasses` is the supported entry
  point: `trust(Class...)` (expands through the schema), `trustExactly(Class...)` (only what you named),
  `trustClasses(String...)` for classes you cannot reference at compile time, `trustPackages(String...)`
  and `trust(Predicate<Class<?>>)`. Its signatures use only `Class`, `String` and
  `java.util.function.Predicate` — deliberately no Avro types, because Avro's own API is
  unusable from the shaded clients (see below).

  Declarations are process-wide and accumulate, so code making a temporary change takes a `snapshot()`
  and hands it back to `restore(State)` afterwards; `resetToDefaults()` returns to the built-in
  baseline. The declarations live in a nested `AvroTrustedClasses.State` that the facade swaps
  atomically, and both `snapshot()` and `restore(...)` copy, so a snapshot never shares mutable state
  with what is live and can be restored more than once. It installs by composing with the validator already in
  place, so Avro's defaults, the `SERIALIZABLE_*` system properties and any policy the application
  installed itself all keep working. It also seeds the protobuf runtime types and the collection types
  Avro records as `java-class` properties, so those never have to be rediscovered by a caller.

  `ClassSecurityValidator` only exists from Avro 1.12.2, and an application using
  `pulsar-client-original` can pin an older Avro, so every reference to it sits behind a lazily-loaded
  holder: on such an Avro there is nothing to enforce and declaring trust is a no-op, rather than a
  `NoClassDefFoundError` out of `Schema.AVRO(...)`.

- Catch `AvroRuntimeException` in both `GenericAvroReader.read` overloads so fast-reader decoding failures
  are still reported as `SchemaSerializationException`. The `InputStream` overload is the one consumers
  reach, via `MessageImpl.decodeBySchema` → `AbstractStructSchema.decode` → `AbstractMultiVersionReader`.

- Update `ProtobufSchemaTest` for the enum `default` attribute.

- The test JVMs are granted **no** Avro trust. Declaring classes where the application hands one over
  covers the fixture POJOs the suites serialize just as it covers an application's own, so granting the
  whole Pulsar namespace would only give tests a safety net production does not have — a path that failed
  to declare something would pass in CI and fail for users. Reverting a change is the test's own
  responsibility rather than shared test infrastructure's: a test that needs a known baseline takes a
  `snapshot()`, pins the global validator, and puts both back afterwards, so declarations belonging to
  a broker shared across test classes survive.

### Upgrade note (for the release notes)

**Avro 1.12.2 no longer reflects over arbitrary classes**, but for most applications there is nothing to
do: passing a class to `Schema.AVRO(...)` declares it, along with everything the derived schema
references. Pulsar Functions and connectors are covered the same way.

Where a class is named *only* by a schema that arrived from the registry, Pulsar deliberately does not
declare it on the application's behalf — that is the case the allow-list exists to constrain. The main
one is `Schema.AUTO_CONSUME()` against a topic whose schema is a bare enum, which resolves the enum class
named in the writer's schema. An application that hits it declares the class itself, once, before the
first producer or consumer:

```java
import org.apache.pulsar.client.schema.AvroTrustedClasses;

AvroTrustedClasses.trust(Colour.class);
```

If something is rejected that should not be, the exception names the class, and `trust(...)` on it is the
fix.

`Schema.JSON` is unaffected. `Schema.PROTOBUF` is affected at schema-construction time.

**A warning about Avro's own error message,** because it is the first thing users will read. It says to
set `org.apache.avro.SERIALIZABLE_CLASSES` / `SERIALIZABLE_PACKAGES` "or set them via the API (see
`org.apache.avro.util.ClassSecurityValidator`)". For anyone on the shaded client — `pulsar-client` or
`pulsar-client-all`, the usual dependency — both are wrong: shading relocates the class, so the snippet
does not compile (and adding `org.apache.avro:avro` to make it compile is worse — it compiles, runs,
configures a second copy, and the exception fires unchanged), and the property name is relocated with it,
so the `-D` is silently inert. Following the exception produces a byte-identical failure.

The system properties do work with the right name for the artifact:

```bash
# pulsar-client / pulsar-client-all (shaded - the usual dependency)
-Dorg.apache.pulsar.shade.org.apache.avro.SERIALIZABLE_PACKAGES=com.example.model

# pulsar-client-original (unshaded), broker, proxy and other components
-Dorg.apache.avro.SERIALIZABLE_PACKAGES=com.example.model
```

They are read in a static initializer, so they must be set before anything in the JVM has touched Avro — a
command-line `-D` always works, `System.setProperty()` only if it runs first. Note that a package alone
will not cover the `java.util.List` above; that is the bookkeeping `trust(Class...)` does for you.

Setting `SERIALIZABLE_PACKAGES=*` (same two spellings) restores the pre-1.12.2 behaviour, if the upgrade
breaks an application and you need it running while working out which classes to trust. Whether to keep it
that way is a policy decision for each deployment to make on its own assessment.

Two things that are easy to miss: nested classes must be named with the binary `$` form
(`com.example.Outer$Inner`), and `Schema.AUTO_CONSUME()` against a topic whose schema is
`Schema.JSON(SomeEnum.class)` needs that enum trusted — an enum is the one POJO shape that produces a
non-RECORD top-level schema, and the consumer resolves the class named in the writer's schema.

### Verifying this change

- `AvroTrustedClassesTest` pins the global validator to Avro's hardcoded `DEFAULT_TRUSTED_CLASSES`, so
  what is trusted is only ever what the test declares. It covers: `Schema.AVRO(Order.class)` round-tripping
  with no declaration at all, where `Order` reaches an enum, a record in another package, a `List` and a
  `URI`; that a definition built with `withJsonDef` — how `AutoConsumeSchema` passes a registry-fetched
  schema — does not expand trust; that a schema is walked only once; `trustExactly` not following
  references; interfaces, including one with no derivable schema; expansion through an `@Union` of
  implementations; a nested enum registering under its binary `$` name; the predicate form; the seeded
  collection types; `snapshot()`/`restore(...)` round-tripping and `resetToDefaults()` clearing both the
  declarations and the walk cache; and that declaring more trust does not compose into the global
  validator again.
- `AvroTrustedClassesWithoutValidatorTest` loads the class in a class loader that hides
  `ClassSecurityValidator`, so the older-Avro path is exercised from a JVM whose own Avro is current.
- `PulsarInternalAvroTypesTrustTest` round-trips `MetadataEvent`, `PulsarEvent` with a fully populated
  `TopicPolicies`, and both transaction buffer snapshot formats under that same pinned validator — the
  guard that the broker's own system-topic types stay serializable with nothing declared in advance. It
  also pins that `Schema.JSON` over the broker's load-balancer types needs no declaration, and that trust
  does not leak to unrelated application classes.
- Personal CI on this branch is green (40/40 jobs), including `CI - System - Pulsar Connectors - Thread`
  and `- Process`, which run a function using `Schema.AVRO` over a function-owned POJO.

- [x] The public API

Adds `org.apache.pulsar.client.schema.AvroTrustedClasses`. Client applications using `Schema.AVRO` with
their own POJOs must declare those classes after this upgrade — see the upgrade note above. Pulsar's own
types, and Pulsar Functions and connectors, are handled automatically.

### Documentation

- [ ] `doc`
- [x] `doc-required`
- [ ] `doc-not-needed`
- [ ] `doc-complete`


nodece (approved):

dao-jun (approved):

Technoboy- (approved):

lhotari:
The change https://github.com/apache/avro/pull/3304 causes the StackOverflowError issue.

```
  Caused by: java.lang.StackOverflowError
  	at org.apache.avro.specific.SpecificData.createSchema(SpecificData.java:492)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:673)
  	at org.apache.avro.reflect.ReflectData.createNonStringMapSchema(ReflectData.java:548)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:642)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:649)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createNonStringMapSchema(ReflectData.java:549)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:642)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:649)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createNonStringMapSchema(ReflectData.java:549)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:642)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:649)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createNonStringMapSchema(ReflectData.java:549)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:642)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:649)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createNonStringMapSchema(ReflectData.java:549)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:642)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:649)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
  	at org.apache.avro.reflect.ReflectData.createNonStringMapSchema(ReflectData.java:549)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:642)
  	at org.apache.avro.reflect.ReflectData.createFieldSchema(ReflectData.java:894)
  	at org.apache.avro.reflect.ReflectData$AllowNull.createFieldSchema(ReflectData.java:98)
  	at org.apache.avro.reflect.ReflectData.createSchema(ReflectData.java:744)
```

lhotari:
Issue reported to Avro project: https://issues.apache.org/jira/browse/AVRO-4209

crossoverJie (approved):

dao-jun (approved):

lhotari:
waiting for Avro 1.12.2 release

nodece (approved):

merlimat (approved):

PR #25002, opened 2025-11-20
[fix][test] Fix flaky EndToEndMetadataTest.testPublishConsume

hanmz:
### Motivation
When executing this test case in a 16C32G Linux environment, the following error will be encountered. The problem is that Jetty requires more than 4 threads. Therefore, we cannot configure it to 4 here; instead, we must use the default value `Math.max(8, 2 * Runtime.getRuntime().availableProcessors())`

Clipboard_Screenshot_1763638875

### Modifications 
Use the default values ​​configured for `numHttpServerThreads`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/EmbeddedPulsarCluster.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/EmbeddedPulsarCluster.java
@@ -120,7 +120,6 @@ public class EmbeddedPulsarCluster implements AutoCloseable {
         conf.setManagedLedgerNumSchedulerThreads(1);
         conf.setWebSocketNumIoThreads(1);
         conf.setNumTransactionReplayThreadPoolSize(1);
-        conf.setNumHttpServerThreads(4);
 
         if (numBookies < 2) {
             conf.setManagedLedgerDefaultEnsembleSize(1);


Technoboy- (approved):

nodece (approved):

lhotari (commented):
> Therefore, we cannot configure it to 4 here; instead, we must use the default value Math.max(8, 2 * Runtime.getRuntime().availableProcessors())

One of the goals in setting the value is to use minimal amount of threads. Removing of the value would set the value to 64 on a 32 core machine like yours. That seems a lot for testing.
Have you checked what the minimal value is?

We could consider passing `-XX:ActiveProcessorCount=4` to the test JVM args to address test behavior differences across environments. The reason why that would make sense is that the tests would behave closer to how they behave in GitHub Actions based CI ([GitHub Actions VM specs](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) when the number of threads are similar across environments.

By default, the Pulsar maven build has the setting `testForkCount=4` which will run tests in 4 JVMs in parallel.

PR #25008, opened 2025-11-21
[fix][client] Fix double recycling of the message in isValidConsumerEpoch method

lhotari:
### Motivation

There's a bug in the isValidConsumerEpoch method:

https://github.com/apache/pulsar/blob/807dcaf5d928f8202c1bf8b8402cfcf72a41e63d/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java#L1307-L1319

There's a call to both `message.release()` and `message.recycle()`. 

`message.release()` will call `recycle()` when it's a pooled message:
https://github.com/apache/pulsar/blob/807dcaf5d928f8202c1bf8b8402cfcf72a41e63d/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L738-L744

This is why it's a problem if both `release()` and `recycle()` are called.

### Modifications

Remove the `message.recycle()` call.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
@@ -1312,7 +1312,6 @@ public abstract class ConsumerBase<T> extends HandlerState implements Consumer<T
             log.info("Consumer filter old epoch message, topic : [{}], messageId : [{}], messageConsumerEpoch : [{}], "
                     + "consumerEpoch : [{}]", topic, message.getMessageId(), message.getConsumerEpoch(), consumerEpoch);
             message.release();
-            message.recycle();
             return false;
         }
         return true;


lhotari:
> Is it possible to add a test? This issue should always happen when polled message is enabled

It would be useful, but I don't currently have time to add it.

BewareMyPower:
I tried to add a test and found it's hard to test.

This bug does not affect correctness. It just throws an `IllegalStateException`, which will be handled by the `UncaughtExceptionHandler` from `ExtendedThreadFactory`. It's basically equivalent with returning false in this case.

BewareMyPower:
Eventually I added a test to verify no exception would be propagated to the internal executor: https://github.com/apache/pulsar/pull/25013

It could help avoid the possible regression in future to some extent.

PR #25010, opened 2025-11-24
[improve][broker] Add counter to topic stats for delayed messages which exceed TTL at publish time

zjxxzjwang:
### Motivation

Currently, Pulsar Broker lacks a metric to track instances where delayed messages exceed their TTL. This results in messages set with delayed delivery times exceeding the TTL expiring before being consumed by users, with no mechanism to detect this occurrence. Therefore, a counter is needed to track such instances.

### Modifications

1. Add a new member variable Rate exceedTTLDelayMessage = new Rate() to PersistentTopic to track the number of delayed messages exceeding the TTL.
2. Within the message sending logic of handleSend, add a check specifically for delayed messages whose delay time exceeds the TTL. If exceeded, call exceedTTLDelayMessage.recordEvent().
3. In TopicStatsImpl, add the exceedTTLDelayMessages variable to retrieve and display the count of delayed messages exceeding the TTL.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -644,7 +644,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             decrementPendingWriteOpsAndCheck();
             return;
         }
-        if (isExceedMaximumDeliveryDelay(headersAndPayload)) {
+        if (isExceedMaximumDeliveryDelay(headersAndPayload, publishContext)) {
             publishContext.completed(
                     new NotAllowedException(
                             String.format("Exceeds max allowed delivery delay of %s milliseconds",
@@ -652,6 +652,8 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             decrementPendingWriteOpsAndCheck();
             return;
         }
+        // Count exceed ttl delayed messages.
+        checkDelayedMessageExceededTTL(headersAndPayload, publishContext);
 
         MessageDeduplication.MessageDupStatus status =
                 messageDeduplication.isDuplicate(publishContext, headersAndPayload);
@@ -4552,7 +4554,7 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             decrementPendingWriteOpsAndCheck();
             return;
         }
-        if (isExceedMaximumDeliveryDelay(headersAndPayload)) {
+        if (isExceedMaximumDeliveryDelay(headersAndPayload, publishContext)) {
             publishContext.completed(
                     new NotAllowedException(
                             String.format("Exceeds max allowed delivery delay of %s milliseconds",
@@ -4560,6 +4562,8 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
             decrementPendingWriteOpsAndCheck();
             return;
         }
+        // Count exceed ttl delayed messages.
+        checkDelayedMessageExceededTTL(headersAndPayload, publishContext);
 
         MessageDeduplication.MessageDupStatus status =
                 messageDeduplication.isDuplicate(publishContext, headersAndPayload);
@@ -4792,29 +4796,32 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
         return Optional.ofNullable(shadowSourceTopic);
     }
 
-    protected boolean isExceedMaximumDeliveryDelay(ByteBuf headersAndPayload) {
+    protected boolean isExceedMaximumDeliveryDelay(ByteBuf headersAndPayload, PublishContext publishContext) {
         if (isDelayedDeliveryEnabled()) {
             long maxDeliveryDelayInMs = getDelayedDeliveryMaxDelayInMillis();
-            if (maxDeliveryDelayInMs <= 0) {
-                return false;
+            if (maxDeliveryDelayInMs > 0) {
+                MessageMetadata msgMetadata = publishContext.peekMessageMetadata(headersAndPayload);
+                if (msgMetadata == null || !msgMetadata.hasDeliverAtTime()) {
+                    return false;
+                }
+                return msgMetadata.getDeliverAtTime() - msgMetadata.getPublishTime() > maxDeliveryDelayInMs;
             }
-            headersAndPayload.markReaderIndex();
-            MessageMetadata msgMetadata = Commands.parseMessageMetadata(headersAndPayload);
-            headersAndPayload.resetReaderIndex();
-            if (!msgMetadata.hasDeliverAtTime()) {
-                return false;
-            }
-            long deliverAtTime = msgMetadata.getDeliverAtTime();
-            // count exceed ttl delayed messages
-            Integer messageTTLInSeconds = topicPolicies.getMessageTTLInSeconds().get();
-            if (messageTTLInSeconds != null && messageTTLInSeconds > 0
-                    && deliverAtTime >= (messageTTLInSeconds * 1000L) + System.currentTimeMillis()) {
-                this.incrementTtlExceededDelayedMessages();
-            }
-
-            return deliverAtTime - msgMetadata.getPublishTime() > maxDeliveryDelayInMs;
         }
         return false;
+    }
+
+    private void checkDelayedMessageExceededTTL(ByteBuf headersAndPayload, PublishContext publishContext) {
+        if (isDelayedDeliveryEnabled()) {
+            Integer messageTTLInSeconds = topicPolicies.getMessageTTLInSeconds().get();
+            if (messageTTLInSeconds != null && messageTTLInSeconds > 0) {
+                MessageMetadata msgMetadata = publishContext.peekMessageMetadata(headersAndPayload);
+                if (msgMetadata != null && msgMetadata.hasDeliverAtTime()) {
+                    if (msgMetadata.getDeliverAtTime() >= (messageTTLInSeconds * 1000L) + System.currentTimeMillis()) {
+                        this.incrementTtlExceededDelayedMessages();
+                    }
+                }
+            }
+        }
     }
 
     @Override
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
@@ -406,6 +406,8 @@ public class Producer {
 
         private long entryTimestamp;
 
+        private MessageMetadata messageMetadata;
+
         @Override
         public long getLedgerId() {
             return ledgerId;
@@ -686,6 +688,17 @@ public class Producer {
             }
         };
 
+        @Override
+        public MessageMetadata peekMessageMetadata(ByteBuf entryData) {
+            if (messageMetadata == null) {
+                entryData.markReaderIndex();
+                messageMetadata = new MessageMetadata();
+                Commands.parseMessageMetadata(entryData, messageMetadata);
+                entryData.resetReaderIndex();
+            }
+            return messageMetadata;
+        }
+
         public void recycle() {
             producer = null;
             sequenceId = -1L;
@@ -701,6 +714,10 @@ public class Producer {
             isMarker = false;
             if (propertyMap != null) {
                 propertyMap.clear();
+            }
+            if (messageMetadata != null) {
+                messageMetadata.clear();
+                messageMetadata = null;
             }
             recyclerHandle.recycle(this);
         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Topic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Topic.java
@@ -35,6 +35,7 @@ import org.apache.pulsar.client.api.transaction.TxnID;
 import org.apache.pulsar.common.api.proto.CommandSubscribe.InitialPosition;
 import org.apache.pulsar.common.api.proto.CommandSubscribe.SubType;
 import org.apache.pulsar.common.api.proto.KeySharedMeta;
+import org.apache.pulsar.common.api.proto.MessageMetadata;
 import org.apache.pulsar.common.policies.data.BacklogQuota;
 import org.apache.pulsar.common.policies.data.BacklogQuota.BacklogQuotaType;
 import org.apache.pulsar.common.policies.data.EntryFilters;
@@ -129,6 +130,10 @@ public interface Topic {
 
         default boolean supportsReplDedupByLidAndEid() {
             return false;
+        }
+
+        default MessageMetadata peekMessageMetadata(ByteBuf entryData) {
+            return null;
         }
     }
 


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2042,6 +2042,13 @@ protected void handleSend(CommandSend send, ByteBuf headersAndPayload) {
             producer.publishMessage(send.getProducerId(), send.getSequenceId(), headersAndPayload,
                     send.getNumMessages(), send.isIsChunk(), send.isMarker(), position);
         }
+
+        // count delayed message times exceeding the ttl policy time
+        MessageMetadata msgMetadata = Commands.parseMessageMetadata(headersAndPayload);
+        if (msgMetadata.hasDeliverAtTime() && service.getPulsar()
+                .isMessageDelayTimeExceedTTL(producer.getTopic(), msgMetadata.getDeliverAtTime())) {
+            producer.getTopic().incrementExceedTTLDelayMessages();
+        }
```
Denovo1998: Here we might want to put it directly in PersistentTopic.publishMessage to use the parsed metadata + TTL to complete the statistics.

Perhaps it can be done in `org.apache.pulsar.broker.service.persistent.PersistentTopic#isExceedMaximumDeliveryDelay`:
```java
protected boolean isExceedMaximumDeliveryDelay(ByteBuf headersAndPayload) {
        if (isDelayedDeliveryEnabled()) {
            long maxDeliveryDelayInMs = getDelayedDeliveryMaxDelayInMillis();
            if (maxDeliveryDelayInMs > 0) {
                headersAndPayload.markReaderIndex();
                MessageMetadata msgMetadata = Commands.parseMessageMetadata(headersAndPayload);
                headersAndPayload.resetReaderIndex();
                return msgMetadata.hasDeliverAtTime()
                        && msgMetadata.getDeliverAtTime() - msgMetadata.getPublishTime() > maxDeliveryDelayInMs;
            }
        }
        return false;
    }
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
```diff
@@ -2252,4 +2252,17 @@ public HealthChecker getHealthChecker() {
...
     }
+
+    /**
+     * Check if message delay time exceeds TTL
+     *
+     * @param topic
+     * @param deliverAtTime
+     * @return true if message delay time exceeds TTL, false otherwise
+     */
+    public boolean isMessageDelayTimeExceedTTL(Topic topic, long deliverAtTime) {
```
Denovo1998: Constraints/statistics related to delay are more unified in `PersistentTopic`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
```diff
@@ -2252,4 +2252,17 @@ public HealthChecker getHealthChecker() {
...
+        return deliverAtTime
+                >= topic.getHierarchyTopicPolicies().getMessageTTLInSeconds().get() * 1000 + System.currentTimeMillis();
```
Denovo1998: When TTL is not enabled, i.e., TTL <= 0. The condition here will become deliverAtTime >= now. For any normal delayed message (deliverAtTime > now), it will be true, meaning "all delayed messages" will be counted as "expired TTL".

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -297,6 +295,8 @@ protected TopicStatsHelper initialValue() {
     @Getter
     private final PersistentTopicMetrics persistentTopicMetrics = new PersistentTopicMetrics();
 
+    private final Rate exceedTTLDelayMessage = new Rate();
```
Denovo1998: In this pr all parameters or function names regarding "delay message" should be changed to "delayed messages".

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4913,4 +4872,14 @@ public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
...
     }
+
+    @Override
+    public void incrementExceedTTLDelayMessages() {
+        this.exceedTTLDelayMessage.recordEvent();
+    }
+
+    @Override
+    public long getExceedTTLDelayMessages() {
+        return this.exceedTTLDelayMessage.getCount();
```
Denovo1998: The return value of getCount() is "the number of events in the most recent period" (i.e., the number of events accumulated since the last call to calculateRate());

Here is perhaps to count "how many delayed messages exceeded the TTL"?
```java
exceedTTLDelayMessage.getTotalCount()
```

Denovo1998 (changes requested):
Left some comments.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
zjxxzjwang: Alright, I'll make some adjustments

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
zjxxzjwang: Alright, I'll make some adjustments

pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
zjxxzjwang: The new version of TTL means that messages will not expire, I will make modifications

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
zjxxzjwang: Alright, I'll make some adjustments

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
zjxxzjwang: It's also possible, but taking the value for a period of time like this can only be recalculated

zjxxzjwang:
@Denovo1998  Hello, could you please help me review again

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4913,4 +4876,30 @@ public void readEntryFailed(ManagedLedgerException exception, Object ctx) {
...
+     * Check if the message deliver time is expired by message TTL.
+     *
+     * @param msgMetadata the message metadata
+     * @return true if the message deliver time is expired by message TTL, false otherwise
+     */
+    protected boolean isExceedMessageTTL(MessageMetadata msgMetadata) {
+        Integer messageTTLInSeconds = topicPolicies.getMessageTTLInSeconds().get();
+        if (messageTTLInSeconds == null || messageTTLInSeconds <= 0) {
+            return false;
+        }
```
Denovo1998: I think it should only be necessary to parse metadata when maxDelayLimit > 0 or TTL > 0; otherwise, return false directly.

Example:
```
if (!isDelayedDeliveryEnabled()) return false;
long maxDelay = getDelayedDeliveryMaxDelayInMillis();
Integer ttl = topicPolicies.getMessageTTLInSeconds().get();
if (maxDelay <= 0 && (ttl == null || ttl <= 0)) return false;
// then parse metadata once...
```

zjxxzjwang: Good suggestion, I have made the necessary modifications

zjxxzjwang:
@Denovo1998  Hello, could you please help me review again

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4833,13 +4792,22 @@ public Optional<TopicName> getShadowSourceTopic() {
...
+            MessageMetadata msgMetadata = Commands.parseMessageMetadata(headersAndPayload);
+            headersAndPayload.resetReaderIndex();
+            if (!msgMetadata.hasDeliverAtTime()) {
+                return false;
             }
+            long deliverAtTime = msgMetadata.getDeliverAtTime();
+            // count exceed ttl delayed messages
+            if (deliverAtTime >= (messageTTLInSeconds * 1000L) + System.currentTimeMillis()) {
+                this.incrementExceedTTLDelayedMessages();
+            }
```
Denovo1998: There is no guarantee that messageTTLInSeconds > 0.

I suggest wrapping the counting logic in 

```java
if (messageTTLInSeconds != null && messageTTLInSeconds > 0)
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -297,6 +295,8 @@ protected TopicStatsHelper initialValue() {
     @Getter
     private final PersistentTopicMetrics persistentTopicMetrics = new PersistentTopicMetrics();
 
+    private final Rate exceedTTLDelayedMessage = new Rate();
```
lhotari: `exceedTTLDelayedMessage` doesn't read well. It seems that `ttlExceededDelayedMessagesRate` would be more accurate for this field.

In addition, it would be necessary to add a comment to this field to explain it.

pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/stats/TopicStatsImpl.java
```diff
@@ -186,6 +186,9 @@ public class TopicStatsImpl implements TopicStats {
     /** The last publish timestamp in epoch milliseconds. */
     public long lastPublishTimeStamp;
 
+    /** The number of delay messages that exceed TTL delay. */
+    public long exceedTTLDelayedMessages;
```
lhotari: similar comment about `exceedTTLDelayedMessages` as in the previous comment. It doesn't read well. I think that `ttlExceededDelayedMessages` would be better. In the comment, it should be made explicit that this is the number of delayed messsages that exceeded TTL at publish time.

zjxxzjwang:
@lhotari Hello, I have made the modifications. Can you help me review them

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4833,13 +4795,23 @@ public Optional<TopicName> getShadowSourceTopic() {
...
-                headersAndPayload.resetReaderIndex();
-                return msgMetadata.hasDeliverAtTime()
-                        && msgMetadata.getDeliverAtTime() - msgMetadata.getPublishTime() > maxDeliveryDelayInMs;
+            Integer messageTTLInSeconds = topicPolicies.getMessageTTLInSeconds().get();
+            if (maxDeliveryDelayInMs <= 0 && (messageTTLInSeconds == null || messageTTLInSeconds <= 0)) {
+                return false;
+            }
+            headersAndPayload.markReaderIndex();
+            MessageMetadata msgMetadata = Commands.parseMessageMetadata(headersAndPayload);
+            headersAndPayload.resetReaderIndex();
```
lhotari: One side-effect of this change is that the message metadata would get parsed when TTL is set. It has a small performance overhead compared to the previous behavior. Therefore, it would be useful to be able to disable this feature completely.

lhotari:
> Currently, Pulsar Broker lacks a metric to track instances where delayed messages exceed their TTL. This results in messages set with delayed delivery times exceeding the TTL expiring before being consumed by users, with no mechanism to detect this occurrence. Consequently, there is a significant risk of message loss.

The implementation doesn't seem to match the goal of the change ("track instances where delayed messages exceed their TTL").
It currently target the case where the message has already been expired when it gets published. How common is that case? What TTL value are you using?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
zjxxzjwang: I have made modifications to maintain the same logic as before, and only perform statistics when maxDeliveryDelayIMs is greater than 0

zjxxzjwang:
> > Currently, Pulsar Broker lacks a metric to track instances where delayed messages exceed their TTL. This results in messages set with delayed delivery times exceeding the TTL expiring before being consumed by users, with no mechanism to detect this occurrence. Consequently, there is a significant risk of message loss.
> 
> The implementation doesn't seem to match the goal of the change ("track instances where delayed messages exceed their TTL"). It currently target the case where the message has already been expired when it gets published. How common is that case? What TTL value are you using?

I have made the necessary modifications, and the current logic will not introduce additional overhead. Based on extensive user usage patterns, this counter is critical and highly valuable. Without it, users would remain unaware when messages expire due to delayed delivery exceeding the TTL. With this counter in place, we can implement alerting mechanisms to notify users and adjust the TTL accordingly.

zjxxzjwang:
@lhotari  Hello, I have made the modifications. Can you help me review them

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4833,13 +4795,24 @@ public Optional<TopicName> getShadowSourceTopic() {
...
+            if (!msgMetadata.hasDeliverAtTime()) {
+                return false;
             }
+            long deliverAtTime = msgMetadata.getDeliverAtTime();
+            // count exceed ttl delayed messages
+            Integer messageTTLInSeconds = topicPolicies.getMessageTTLInSeconds().get();
+            if (messageTTLInSeconds != null && messageTTLInSeconds > 0
```
lhotari: there should be a check before parsing the message metadata to skip the check completely if message TTL is not set at all. The reason for performing this before the parsing of metadata is to avoid a performance overhead when TTL is not set.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4833,13 +4795,24 @@ public Optional<TopicName> getShadowSourceTopic() {
     protected boolean isExceedMaximumDeliveryDelay(ByteBuf headersAndPayload) {
```
lhotari: This method has a side-effect now, it's no longer a pure function for determining whether maximum delivery delay has exceeded.

The code using this method would be misleading after adding the TTL stats logic to this method:
https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java#L645-L652

Keeping the new TTL stats together with the maximum delivery delay check does help improve performance since it's possible to eliminate parsing the message metadata multiple times. However, I think that a better solution would be to keep these separate and move the message metadata parsing to be lazily performed in `org.apache.pulsar.broker.service.Topic.PublishContext` implementations.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
lhotari: Moving the TTL stats to a new method would be preferred as mentioned in a previous comment.

zjxxzjwang:
I have made the modifications. Please review them. I have broken down the method and reduced unnecessary metadata parsing.  @lhotari

zjxxzjwang:
@lhotari   I have made the modifications. Please review them. I have broken down the method and reduced unnecessary metadata parsing.

PR #25015, opened 2025-11-25
[fix][client] Fix AutoProduceBytesSchema.clone() method

lhotari:
### Motivation

AutoProduceBytesSchema has a clone method and it would be expected that it returns a new instance which has similar state as the original instance. This is not the case currently.

### Modifications

- implement `AutoProduceBytesSchema.clone()` properly

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/AutoProduceBytesSchema.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/AutoProduceBytesSchema.java
@@ -50,7 +50,7 @@ public class AutoProduceBytesSchema<T> implements Schema<byte[]> {
     }
 
     private AutoProduceBytesSchema(AutoProduceBytesSchema<T> other) {
-        this.schema = other.schema;
+        this.schema = other.schema != null ? other.schema.clone() : null;
         this.userProvidedSchema = other.userProvidedSchema;
         this.requireSchemaValidation = other.requireSchemaValidation;
     }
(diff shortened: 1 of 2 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/AutoProduceBytesSchema.java
```diff
@@ -49,6 +49,12 @@ public AutoProduceBytesSchema(Schema<T> schema) {
                                        && schemaInfo.getType() != SchemaType.NONE;
     }
 
+    private AutoProduceBytesSchema(AutoProduceBytesSchema<T> other) {
+        this.schema = other.schema;
```
Denovo1998: @lhotari 
```java
this.schema = other.schema != null ? other.schema.clone() : null;
```
Compared to the previous implementation, the new clone() no longer clones the underlying schema instance but shares the reference instead.
If we want to preserve the previous deep-clone semantics, we might consider calling other.schema.clone() in the copy constructor when schema != null, but I don't think this is required for this fix.

lhotari: Thanks for catching. It makes sense to preserve the previous behavior, so I'll fix that.

lhotari: Addressed.

Denovo1998: OK, no problem.

Technoboy- (approved):

PR #25016, opened 2025-11-25
[fix][broker]Fix memory leak when using a customized ManagedLedger implementation

poorbarcode:
### Motivation

Issue: if a user used a customized `ManagedLedger` implementation, which does not extend `ManagedLedgerImpl`, he will encounter a memory leak issue, you can reproduce it by the new test `testNoMemoryLeakWhenExpireMessages`

https://github.com/apache/pulsar/pull/24606 improved the function `message expiration`, but forgot to release `EntryImpl` here: https://github.com/apache/pulsar/pull/24606/files#diff-5edf14cc6f25857d0cfdd26b2d3b3141230ecfb0dfa95aebf7583fd76ede4c4bR3918-R3929

**PersistentTopic.java**
```java
    public CompletableFuture<Boolean> isOldestMessageExpiredAsync(ManagedCursor cursor, int messageTTLInSeconds) {
        CompletableFuture<Boolean> res = new CompletableFuture<>();
        cursor.asyncGetNthEntry(1, IndividualDeletedEntries.Include, new AsyncCallbacks.ReadEntryCallback() {
            @Override
            public void readEntryComplete(Entry entry, Object ctx) {
                long entryTimestamp = 0;
                try {
                    res.complete(MessageImpl.isEntryExpired(
                            (int) (messageTTLInSeconds * MESSAGE_EXPIRY_THRESHOLD), entryTimestamp));
                } catch (IOException e) {
                    log.warn("[{}] [{}] Error while getting the oldest message", topic, cursor.toString(), e);
                    res.complete(false);
                }
                // (Highlight) the entry that was read out is not released here, which causes a memory leak issue
            }
        }, null);
        return res;
    }
```

```
2025-11-10T01:29:54,412+0000 [compact-task-publisher-OrderedScheduler-2-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedLedgerImplWithTS - [public/default/persistent/test-partition-1] Published compact task for streamId: 68238366, ledger 938515 from entry 113 to 115
2025-11-10T01:29:54,428+0000 [pulsar-io-9-6] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
#1:
	io.netty.buffer.AdvancedLeakAwareByteBuf.readByte(AdvancedLeakAwareByteBuf.java:401)
	org.apache.pulsar.common.api.proto.LightProtoCodec.readVarInt(LightProtoCodec.java:140)
	org.apache.pulsar.common.api.proto.MessageMetadata.parseFrom(MessageMetadata.java:1359)
	org.apache.pulsar.common.protocol.Commands.parseMessageMetadata(Commands.java:479)
	org.apache.pulsar.common.protocol.Commands.parseMessageMetadata(Commands.java:466)
	org.apache.pulsar.common.protocol.Commands.getEntryTimestamp(Commands.java:499)
	org.apache.pulsar.broker.service.persistent.PersistentTopic$13.readEntryComplete(PersistentTopic.java:3921)
	org.apache.bookkeeper.mledger.impl.cache.RangeEntryCacheImpl$1.readEntriesComplete(RangeEntryCacheImpl.java:239)
	org.apache.bookkeeper.mledger.impl.cache.PendingReadsManager$PendingRead.readEntriesComplete(PendingReadsManager.java:278)
	org.apache.bookkeeper.mledger.impl.cache.PendingReadsManager$PendingRead.lambda$attach$0(PendingReadsManager.java:244)
	org.apache.bookkeeper.common.util.SingleThreadExecutor.safeRunTask(SingleThreadExecutor.java:128)
	org.apache.bookkeeper.common.util.SingleThreadExecutor.run(SingleThreadExecutor.java:105)
	io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	java.base/java.lang.Thread.run(Unknown Source)
```

### Modifications

- fix the bug

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -338,6 +338,7 @@ public class ManagedCursorImpl implements ManagedCursor {
         void operationFailed(ManagedLedgerException exception);
     }
 
+    @VisibleForTesting
     protected ManagedCursorImpl(BookKeeper bookkeeper, ManagedLedgerImpl ledger, String cursorName) {
         this.bookkeeper = bookkeeper;
         this.cursorProperties = Collections.emptyMap();
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/ManagedLedgerClientFactory.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/ManagedLedgerClientFactory.java
@@ -54,6 +54,7 @@ public class ManagedLedgerClientFactory implements ManagedLedgerStorage {
     private static final Logger log = LoggerFactory.getLogger(ManagedLedgerClientFactory.class);
     private static final String DEFAULT_STORAGE_CLASS_NAME = "bookkeeper";
     private BookkeeperManagedLedgerStorageClass defaultStorageClass;
+    @VisibleForTesting
     protected ManagedLedgerFactory managedLedgerFactory;
     private BookKeeper defaultBkClient;
     private final AsyncCache<EnsemblePlacementPolicyConfig, BookKeeper>


lhotari:
> Issue: if a user used a customized `ManagedLedger` implementation, which does not extend `ManagedLedgerImpl`

@poorbarcode Why does this only apply to a custom implementation?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4046,6 +4046,10 @@ public void readEntryComplete(Entry entry, Object ctx) {
                 } catch (Exception e) {
                     log.warn("[{}] [{}] Error while getting the oldest message", topic, cursor.toString(), e);
                     res.complete(false);
+                } finally {
+                    if (entry != null) {
+                        entry.release();
+                    }
```
BewareMyPower, line 4052: It should be the callback's responsibility to release the entry buffer, so this change makes sense.

But I don't understand why the implementation that extends `ManagedLedgerImpl` does not have this issue. If so, should `entry.release()` here be a duplicated release?

poorbarcode: > But I don't understand why the implementation that extends ManagedLedgerImpl does not have this issue. If so, should entry.release() here be a duplicated release?

Because it will run another logic branch, see also https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java#L2168

pulsar-broker/src/test/java/org/apache/bookkeeper/mledger/impl/CustomizedManagedLedgerTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.pulsar.client.api.ProducerConsumerBase;
+import org.apache.pulsar.client.api.Schema;
+import org.apache.pulsar.common.naming.TopicName;
+import org.awaitility.Awaitility;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+@Slf4j
+@Test(groups = "broker")
```
codelipenghui, line 42: Is there any reason to add this test to broker group? All the other tests from managed ledger module don't have this group.

poorbarcode: I considered that the test is triggered by `PersistentTopic.checkMessageExpiry`, could you suggest the correct value?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
poorbarcode: > @poorbarcode Why does this only apply to a custom implementation?

@lhotari  I answered here

BewareMyPower: I see. Though it's not a good practice to add special logic when `ml instanceof ManagedLedgerImpl`, it's okay to go ahead for now. I opened a follow-up issue: https://github.com/apache/pulsar/issues/25019

Technoboy- (approved):
Better to paste the stack for the motivation.

coderzc (approved):

PR #25035, opened 2025-12-03
[improve][pip] PIP-449: Improve metadata sync to have exclusions in syncing specific config matching regex pattern across onprem and cloud clusters

vraulji567:
### Motivation

The pulsar metadata sync does not have a way to exclude specific config from syncing across clusters. For example, a cluster need to have a separate config in on-prem and cloud for it to function correctly for geo-replication. Syncing the cluster config creates issues in geo-replication. Enhance pulsar metadata synchronizer to exclude cluster config from syncing across clusters.
Apply the exclusion in a way that a specific destination cluster can ignore a config sync based on the pattern set at the destination cluster level. Pulsar metadata synchronizer will publish the metadata change event to the destination cluster, which may or may not exclude the event depending on the exclusion pattern set in the destination broker config.
Multiple exclusions patterns can be configured and sync would be excluded if any of the patterns match. 

### Modifications

- Add config `metadataSyncEventExclusions` to exclude sync. Provide regex pattern eg /admin/clusters/.*, /admin/(?!.*test).* to exclude the config sync event.
- Enhance PulsarMetadataEventSynchronizer to exclude syncing events configured with `metadataSyncEventExclusions`

### Verifying this change

  - Added tests to verify different regex patterns to apply exclusions to the metadata sync.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -36,6 +36,7 @@ metadataSyncEventTopic=
 configurationMetadataSyncEventTopic=
 
 # Exclusions for metadata sync between separate clusters on different cloud platforms
+# multiple configurations are separated by comma eg. /admin/clusters/.*,/admin/test/.*
 metadataSyncEventExclusions=
 
 # The metadata store URL for the configuration data. If empty, we fall back to use metadataStoreUrl
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -35,6 +35,9 @@ metadataSyncEventTopic=
 # Event topic to sync configuration-metadata between separate pulsar clusters on different cloud platforms.
 configurationMetadataSyncEventTopic=
 
+# Exclusions for metadata sync between separate clusters on different cloud platforms
+metadataSyncEventExclusions=
+
 # The metadata store URL for the configuration data. If empty, we fall back to use metadataStoreUrl
 configurationMetadataStoreUrl=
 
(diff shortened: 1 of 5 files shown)


conf/broker.conf
```diff
@@ -35,6 +35,9 @@ metadataSyncEventTopic=
 # Event topic to sync configuration-metadata between separate pulsar clusters on different cloud platforms.
 configurationMetadataSyncEventTopic=
 
+# Exclusions for metadata sync between separate clusters on different cloud platforms
```
Technoboy-, line 38: could you add more comments for this, `multiple configurations are separated by commas`

codelipenghui:
Hi @vraulji567 Thanks for contributing this feature. Since this feature is going to add new features to the metadata sync, could you please help create a proposal following https://github.com/apache/pulsar/tree/master/pip#readme

conf/broker.conf
vraulji567: Updated.

vraulji567:
> Hi @vraulji567 Thanks for contributing this feature. Since this feature is going to add new features to the metadata sync, could you please help create a proposal following https://github.com/apache/pulsar/tree/master/pip#readme

I have created the proposal for the change.

PR #25036, opened 2025-12-03
[improve][client] Add null checks for MessageAcknowledger methods to prevent NullPointerException

3pacccccc:
### Motivation
Currently, several methods in the ```MessageAcknowledger``` interface implementation do not validate null parameters for messages, message IDs, or message lists. Passing null values to methods such as ```acknowledge(List<MessageId>)```, ```acknowledge(Messages)```, or ```acknowledgeAsync(MessageId, Transaction)``` can lead to unexpected ```NullPointerException``` errors deep in the call stack. This PR adds explicit null validation at the entry points of these methods to provide clear, early feedback to developers.
### Modifications
1. Added ```validateMessageIds(List<MessageId>)``` and ```validateMessages(Messages)``` helper methods in ```ConsumerBase.java``` to check for null inputs and throw a descriptive ```PulsarClientException.InvalidMessageException```.

2. Integrated these validation checks into:

- ```acknowledge(List<MessageId>)```

- ```acknowledge(Messages)```

- ```acknowledgeAsync(MessageId, Transaction)```

- ```acknowledgeCumulativeAsync(MessageId, Transaction)```

- ```doAcknowledgeWithTxn(...)```

3. Added corresponding unit tests in multiple test classes to verify that null inputs are properly rejected with appropriate exception messages:

- ```TransactionTest.testAckTnxWithNullMessageId()```

- ```ConsumerAckListTest.testAckNullMessageIdList()```

- ```ConsumerBatchReceiveTest.testBatchReceiveAckWithNullMessages()```

- ```TransactionEndToEndTest.testAcknowledgeCumulativeAsyncWithNullMessageId()```
### Verifying this change

### Documentation

- [ ] `doc` 
- [] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/31

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/TransactionTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/TransactionTest.java
@@ -24,7 +24,6 @@ import static org.apache.pulsar.common.naming.SystemTopicNames.PENDING_ACK_STORE
 import static org.apache.pulsar.common.naming.SystemTopicNames.PENDING_ACK_STORE_SUFFIX;
 import static org.apache.pulsar.transaction.coordinator.impl.DisabledTxnLogBufferedWriterMetricsStats.DISABLED_BUFFERED_WRITER_METRICS;
 import static org.apache.pulsar.transaction.coordinator.impl.MLTransactionLogImpl.TRANSACTION_LOG_PREFIX;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.Mockito.doAnswer;
@@ -330,21 +329,6 @@ public class TransactionTest extends TransactionTestBase {
         } catch (PulsarAdminException e) {
             assertEquals(e.getMessage(), "Cannot create topic in system topic format!");
         }
-    }
-    @Test(timeOut = 10000)
-    public void testAckTnxWithNullMessageId() throws Exception {
-        final String topic = BrokerTestUtil.newUniqueName("persistent://tnx/ns1/testAckTnxWithNullMessageId");
-        @Cleanup final Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING)
-                .topic(topic)
-                .subscriptionName("sub1")
-                .subscribe();
-        List<MessageId> messageIdList = null;
-        assertThatThrownBy(
-                () -> consumer.acknowledgeAsync(messageIdList, null).get()
-        )
-                .isInstanceOf(ExecutionException.class)
-                .hasMessageContaining("Cannot handle messages with null messageIdList")
-                .hasCauseInstanceOf(PulsarClientException.InvalidMessageException.class);
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/TransactionTest.java
```diff
@@ -330,6 +331,21 @@ public void testCreateTransactionSystemTopic() throws Exception {
...
+    @Test(timeOut = 10000)
+    public void testAckTnxWithNullMessageId() throws Exception {
+        final String topic = BrokerTestUtil.newUniqueName("persistent://tnx/ns1/testAckTnxWithNullMessageId");
+        @Cleanup final Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING)
+                .topic(topic)
+                .subscriptionName("sub1")
+                .subscribe();
+        List<MessageId> messageIdList = null;
+        assertThatThrownBy(
+                () -> consumer.acknowledgeAsync(messageIdList, null).get()
```
BewareMyPower: This is a null `Transaction`, not message id. It seems that the test name should change?

BewareMyPower: and I think a null transaction should be accepted, though it's not formally documented. In this case, it will fall back to `acknowledgeAsync(messageIdList)`

3pacccccc: Thanks for the feedback @BewareMyPower 

Actually, in this test, the main point is that the ```messageIdList``` is null. The test is designed to verify that passing a null ```messageIdList``` to ```acknowledgeAsync()``` triggers the expected ```InvalidMessageException```, regardless of whether the ```Transaction``` parameter is null or not.

I named the test ```testAckTnxWithNullMessageId``` because the method signature includes both ```messageIdList``` and ```Transaction```, but the focus is really on the ```messageIdList``` validation.

To make it clearer, I’m happy to rename the test to something like:
```testAcknowledgeAsyncWithNullMessageIdList()``` – would that help avoid confusion?

Let me know if you'd prefer that update, or if you have other suggestions!

BewareMyPower: Oh I see the 1st argument is also null.

I think you can just put all `acknowledge` tests into a single test method because there are many overloads.

BewareMyPower: You added this test to `TransactionTest`, but it's nothing related to transaction.

3pacccccc: > Oh I see the 1st argument is also null.
> 
> I think you can just put all `acknowledge` tests into a single test method because there are many overloads.

OK, thanks for the suggestion, that's make sense, I'll do it.

BewareMyPower (commented):
Please also document this behavior to `MessageAcknowledger` so that the behavior will be clear if a null message id or list is passed to `acknowledge` family

3pacccccc:
> Please also document this behavior to `MessageAcknowledger` so that the behavior will be clear if a null message id or list is passed to `acknowledge` family

OK

Technoboy- (approved):

PR #25037, opened 2025-12-04
[fix][broker]Incorrect backlog that is larger than expected

poorbarcode:
### Motivation

By default, the backlog of a subscription is got by `{num of entries published} - {num of entries acknowleged}`
- `num of entries published`: `ManagedLedger.entriesAddedCounter`
- `num of entries acknowleged`: `ManagedLedger.messagesConsumedCounter`

### Bug

There is bug that calculates the entries count between two positons in the method `ManagedLedger.getNumberOfEntries`, which causes `ManagedLedger.messagesConsumedCounter` going to a incorrect value. BTW, this bug was introduced by https://github.com/apache/pulsar/pull/24938

It happens when the `{from position} < {currentLedger}:0` and `{to position} > {currentLedger}:0 and {top position} < {LAC}`
- For example, ledgers: `[{ ledger: 1, entries: 10 }, {current ledger: 2, entries: 10}]`
- Get entries between `1:5(exclusive) ~ 2:5(inclusive)`
- Expected value: `10`
- Actual value: `4`

### Modifications

Fix the bug.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -3959,12 +3959,12 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
         // 2. Add the entries in the ledger pointed by fromPosition.
         //    Add nothing if "toPosition.entryId < 0".
         //    Add nothing if "toPosition" does not exit in "ledgers".
-        LedgerInfo formLedger = ledgers.get(fromPosition.getLedgerId());
-        if (formLedger != null) {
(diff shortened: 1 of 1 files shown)


Technoboy- (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
lhotari: Could you also fix the typo here? `formLedger` should most likely be `fromLedger`.

lhotari (approved):

PR #25038, opened 2025-12-04
[fix][client] Fix lookup request semaphore not release problem

congbobo184:
### Motivation

fix bugs

### Bug

if lookup request timeout, `pendingLookupRequestSemaphore` will not be released, so fix it

### Modifications

Fix the bug.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/congbobo184/pulsar/pull/23

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/LookupRetryTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/LookupRetryTest.java
@@ -21,7 +21,6 @@ package org.apache.pulsar.client.impl;
 import static org.apache.pulsar.common.protocol.Commands.newLookupErrorResponse;
 import static org.apache.pulsar.common.protocol.Commands.newPartitionMetadataResponse;
 import static org.testng.AssertJUnit.fail;
-
 import com.google.common.collect.Sets;
 import java.util.Queue;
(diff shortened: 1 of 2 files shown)


dao-jun (approved):

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/LookupRetryTest.java
```diff
@@ -282,6 +313,9 @@ private Queue<LookupError> errorList(String topicName) {
         @Override
         protected void handlePartitionMetadataRequest(CommandPartitionedTopicMetadata partitionMetadata) {
             TopicName t = TopicName.get(partitionMetadata.getTopic());
+            if (t.getLocalName().equals(TEST_TIME_OUT_SEMAPHORE_RELEASE)) {
```
Technoboy-, line 316: It's better to use mock.answer, to mock the request method timeout.

congbobo184: if mock.answer, we will not get the correct result when change the timeout code

poorbarcode (approved):

Issue #24879, opened 2025-10-21
[Bug] Auto topic delete of partitioned topics has a MetadataCacheImpl race condition


darinspivey:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

We are using partitioned topics of 6 partitions, and we have auto-topic-deletion turned on.  When the GC fires to delete the partitions, as expected, they mostly fire at the same time.  This appears to sometimes cause a race condition when they all attempt to update the topic's metadata.  This has a cascading effect of:

* The metadata update never completes, even with jitter and retries
* The topic deletion eventually fails as a whole because the race conditions causes 1 or more partitions to remain
* The partition(s) becomes orphaned, and the parent topic never can be deleted because GC does not appear to try again after a period of time.

## System info
```
/pulsar $ pulsar version
Current version of pulsar is: 4.0.6
Git Revision 4538ef7645c45a3c8686092128fde6c5d61c762b
Git Branch branch-4.0
Built by Lari Hotari <lhotari@apache.org> on Laris-MBP.lan at 2025-07-30T13:37:25+0300

/pulsar $ uname -a
Linux pulsar-broker-0 6.12.46-66.121.amzn2023.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Sep 22 16:35:59 UTC 2025 x86_64 GNU/Linux
```

## Relevant helm values (Chart `pulsar-4.2.0`)
```
    brokerDeleteInactiveTopicsEnabled: 'true'
    brokerDeleteInactiveTopicsFrequencySeconds: '60'
    brokerDeleteInactiveTopicsMaxInactiveDurationSeconds: '60'
    brokerDeleteInactiveTopicsMode: delete_when_no_subscriptions
    brokerDeleteInactivePartitionedTopicMetadataEnabled: 'true'
    allowAutoTopicCreation: 'true'
    defaultNumPartitions: '6'
    allowAutoTopicCreationType: 'partitioned'
```
* Note: other settings for auto-deleting subscriptions works fine. The problem is happening when the topic attempts auto-deletion (TLDR `delete_when_no_subscriptions` is working fine).

## Relevant Logs

```
[...]

PR #25040, opened 2025-12-05
[fix][admin] Use ClientConfigurationData timeout params to set AsyncHttpConnector timeout params

oneby-wang:
### Motivation

In previous code, we use `client.getConfiguration().getProperty(ClientProperties.CONNECT_TIMEOUT)` as PulsarAdmin's `connectionTimeoutMs`, use `client.getConfiguration().getProperty(ClientProperties.READ_TIMEOUT)` as PulsarAdmin's `readTimeoutMs`, use `PulsarAdminImpl.DEFAULT_REQUEST_TIMEOUT_SECONDS * 1000` as PulsarAdmin's `requestTimeoutMs`. See:

https://github.com/apache/pulsar/blob/b71bea4c42353f9d14f817aa9a0877d495336f34/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java#L121-L128

Before this PR, we can't tune `requestTimeoutMs`, which is always 30000s = 5min. See also in: 

https://github.com/apache/pulsar/issues/24879#issuecomment-3440341666

### Modifications

Now, we use `ClientConfigurationData.requestTimeoutMs` as `AsyncHttpConnector.requestTimeoutMs` , so we can control the http retry timeout by tuning `ClientConfigurationData.requestTimeoutMs`.

AsyncHttpClient get Connector by calling `ConnectorProvider#getConnector(Client client, Configuration runtimeConfig)` method.

https://github.com/eclipse-ee4j/jersey/blob/bb8a2a1b78e39604f4d283d0176705b143355cce/core-client/src/main/java/org/glassfish/jersey/client/ClientConfig.java#L465-L467

Both `connectionTimeoutMs` and `readTimeoutMs` are the same values in jersey properties and `ClientConfigurationData`. See: 

https://github.com/apache/pulsar/blob/b71bea4c42353f9d14f817aa9a0877d495336f34/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/PulsarAdminImpl.java#L146-L150

To ensure code consistency, I would like to suggest using `ClientConfigurationData` and removing jersey config.

Side effect: default `requestTimeoutMs` changes from 30s to 60s after this PR, which I think is reasonable. If I didn't read the source code, I would assume `requestTimeoutMs` is set to 60s, and it would confuse me to find that `requestTimeoutMs` didn't take effect after I adjusted this value.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/11

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
@@ -1856,6 +1856,9 @@ public class PulsarService implements AutoCloseable, ShutdownService {
                     + ", webServiceAddress: " + webServiceAddress);
         }
         PulsarAdminBuilder builder = PulsarAdmin.builder().serviceHttpUrl(adminApiUrl);
+        // most of the admin request requires to make zk-call so, keep the max read-timeout based on
+        // zk-operation timeout. Put it before loading brokerClient_ prefix config, so user can override it
+        builder.readTimeout(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS);
 
         // Apply all arbitrary configuration. This must be called before setting any fields annotated as
         // @Secret on the ClientConfigurationData object because of the way they are serialized.
@@ -1887,9 +1890,6 @@ public class PulsarService implements AutoCloseable, ShutdownService {
                     .enableTlsHostnameVerification(conf.isTlsHostnameVerificationEnabled());
         }
 
-        // most of the admin request requires to make zk-call so, keep the max read-timeout based on
-        // zk-operation timeout
-        builder.readTimeout(conf.getMetadataStoreOperationTimeoutSeconds(), TimeUnit.SECONDS);
         return builder;
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
@@ -1653,6 +1653,10 @@ public class BrokerService implements Closeable {
                 PulsarAdminBuilder builder = PulsarAdmin.builder();
 
                 ServiceConfiguration conf = pulsar.getConfig();
+                // most of the admin request requires to make zk-call so, keep the max read-timeout based on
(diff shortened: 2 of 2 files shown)


pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
```diff
@@ -216,7 +210,6 @@ private void configureAsyncHttpClientConfig(ClientConfigurationData conf, int co
         confBuilder.setCookieStore(null);
         confBuilder.setUseProxyProperties(true);
         confBuilder.setFollowRedirect(false);
-        confBuilder.setRequestTimeout(conf.getRequestTimeoutMs());
```
lhotari, line 223: it would be more consistent to set this to `requestTimeoutMs`, since this is how `connectTimeout and ReadTimeout are also handled.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
oneby-wang, line 223: @lhotari hidden here

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/http/AsyncHttpConnector.java
oneby-wang: The `requestTimeoutMs` param is set twice in this method: once via `confBuilder.setRequestTimeout(conf.getRequestTimeoutMs())` and once via `confBuilder.setReadTimeout(readTimeoutMs)`. For this reason, I removed the line `confBuilder.setRequestTimeout(conf.getRequestTimeoutMs())`.

Issue #25041, opened 2025-12-05
[Bug] update partition process is not reasonable since branch-3.0


TakaHiR07:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

pulsar version 3.0.x

### Issue Description

step 1: admin update topic partition from 50 to 150, throw error
step 2: see  topic partition metadata has been updated to 150, and producer try to produce msg to new partition
step 3: there are 4 partition keep throw exception "topic does not exist", and other new partition can succeed.

The problem is in this pr, https://github.com/apache/pulsar/pull/19166
before pr, update partition process is: 
1. create missing partitions
2. create subscription for new partition
3. update topic metadata

after pr, it become:
1. update topic metadata
2. create missing partitions
3. create subscription for new partition

That is not reasonable, pulsar producer and consumer rely on topic partition metadata to know whether the topic partition has been updated, if metadata updated, client can add new producer or new consumer to the new partition. 

However, with the pr's change, although update partition is not complete, client regard it as success and try to construct new connection to the new topic partition. That is wrong. 

There is another high risk situation, if update topic metadata and create missing partitions succeed, but fail on create subscription. In this moment, producer can send new msg to new partition, but after a while consumer may auto create subscription on new partition, and the position is latest. That would result in consumer lost some message. It is very dangerous.

I don't understand why previous pr need to change the process, while it only illustrate clean the duplicate code

### Error messages

```text

```

### Reproducing the issue

update partition, fail on step 2 or 3

1. update topic metadata
2. create missing partitions
3. create subscription for new partition

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

lhotari:
> I don't understand why previous pr need to change the process, while it only illustrate clean the duplicate code

I think it links to another PR which refers to issue #19085 and mailing list discussion https://lists.apache.org/thread/3s7b7dh0h9qcnn8x3lclxmw75okkqf37 .

TakaHiR07:
[...]

PR #19166, opened 2023-01-10
[improve][broker] Refactor update topic partitions endpoint.

mattisonchao:
### Motivation

This PR follow-up https://github.com/apache/pulsar/pull/19086 to clean up complex logic and remove duplicate method invoke.

### Modifications

- Remove unnecessary topic owner redirect. 
- Refactor unreadable code.
- Change unreasonable status code 406 to 422.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

- [x] The REST endpoints

### Documentation

- [x] `doc-not-needed`

PR #25042, opened 2025-12-05
[fix][broker] fix error process in admin.updateTopicPartition

TakaHiR07:
Main Issue: https://github.com/apache/pulsar/issues/25041

### Motivation

fix the unreasonable update partition process

### Modifications

recover the topic partition process to :

1. create missing partitions
2. create subscription for new partition
3. update topic metadata

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -418,37 +418,31 @@ public class PersistentTopicsBase extends AdminResource {
                 } else {
                     checkFuture = CompletableFuture.completedFuture(null);
                 }
-                return checkFuture.thenCompose(topics -> {
-                    final CompletableFuture<Void> updateMetadataFuture = (expectPartitions == currentMetadataPartitions)
-                            // current metadata partitions is equals to expect partitions
-                            ? CompletableFuture.completedFuture(null)
-                            // update current cluster topic metadata
-                            : namespaceResources().getPartitionedTopicResources()
-                            .updatePartitionedTopicAsync(topicName, m ->
-                                    new PartitionedTopicMetadata(expectPartitions, m.properties));
-                return updateMetadataFuture
+
+                return checkFuture
                     // create missing partitions
-                    .thenCompose(__ -> tryCreatePartitionsAsync(expectPartitions))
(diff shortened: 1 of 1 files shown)


mattisonchao:
Hi @TakaHiR07 

Could you help write a test for it?

TakaHiR07:
> Could you help write a test for it?

@mattisonchao I find that AdminApi2Test#testFailedUpdatePartitionedTopic can not pass. The reason is some other pr such as https://github.com/apache/pulsar/pull/24118 on master-branch is implement based on the new process of admin.updateTopicPartition. 

It seems that this pr maybe is not a good way for master-branch, but only can use in branch-3.0. Since I don't know how many new pr on master-branch are designed based on new process.

Shawyeok:
@TakaHiR07 @mattisonchao 

**This pr would reintroduce a silent, unrecoverable failure mode.** Here is why:

The root cause of #25041 is that the update-partition-count operation is not atomic — it can fail mid-way with a persistent side effect. The key difference between the old and new ordering:

**Old order (pre #19166, what this PR reverts to):**
1. `tryCreatePartitionsAsync(N)` — writes managed-ledger ZK nodes for new partitions
2. `createSubscriptions(N)` — ⚠️ failure here leaves orphaned ZK nodes
3. Update partition metadata

When step 2 fails (e.g. due to bundle not yet assigned, causing 307 redirect exhaustion), ZK nodes for the new partitions already exist but the metadata still shows the old count. These orphaned nodes appear as non-partitioned topics, causing any retry to permanently fail with:
> "Already have non partition topic … could cause conflict"

**New order (introduced in #19166, what this PR reverts):**
1. `updatePartitionedTopicAsync(N)` — update metadata first
2. `tryCreatePartitionsAsync(N)` — idempotent ZK writes
3. `createSubscriptionAsync(...)` — `ConflictException` is tolerated

With metadata updated first, any mid-operation failure leaves the system in a **self-consistent, retryable state** — the metadata already reflects N partitions, so no orphaned nodes exist outside the valid range.

Although the operator can still retry if the endpoint returns an exception, the old ordering makes retries **permanently broken** rather than transiently failing. This was verified on 2.8.1 clusters in my company.

PR #25047, opened 2025-12-08
[fix][broker]Fix incorrect backlog if use multiple acknowledge types on the same subscription

poorbarcode:
### Motivation

**Background**
By default, the backlog of a subscription is got by `{num of entries published} - {num of entries acknowleged}`
- `num of entries published`: `ManagedLedger.entriesAddedCounter`
- `num of entries acknowleged`: `ManagedLedger.messagesConsumedCounter`

How it being modified when calling `reset curosr` or `seek` or `cumulative acknowledge`
- Calculate entries between `{original md-position}` and `{target md-position}`, gets `counter-1`. If `{target md-position}` is smaller than `{original md-position}`, the result of `counter-1` will be negative.
- Calculate entries that are before `{target md-position}` and were acknowledged before calling `reset curosr`, gets `counter-2`. If `{target md-position}` is smaller than `{original md-position}`, the result of `counter-2` will be `0`.
- Let `cursor.messagesConsumedCounter` add `{counter-1} - {counter-2}`

---

**Issue 1**: clears individual acknowledge status before calculate how many entries acked.
1. Call `reset cursor` or `seek` or `cumulative acknowledge`
1. Broker moves `md-position` to the target.
1. Broker clears `cursor. individualDeletedMessages`, see https://github.com/apache/pulsar/blob/v4.1.2/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L2315-L2316
1. Broker calculate how many entries that were acked, see Background for the implementation.
1. Issue: the items that mantained by `cursor. individualDeletedMessages` has been removed at setp 3.
  a. The step 4 can not get a correct  `{counter-2}`, which is described at the Background.

**Issue 2**: `reset curosr` or `seek` forgtes to reduce the counter that acknowledged messages
1. Status: `md-positiin : 3:5`, `cursor. individualDeletedMessages`: `[3:10 ~ 3:11]`
2. Call `reset cursor` or `seek` to `3: 5`, which is the same as before.
3. Broker clears `cursor. individualDeletedMessages`, but forget to reduce `cursor.messagesConsumedCounter` that `cursor. individualDeletedMessages`: `[3:10 ~ 3:11]` increased.

---

### Modifications

- fix bugs

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -210,7 +210,7 @@ public class ManagedCursorImpl implements ManagedCursor {
     @Getter
     @VisibleForTesting
     @Nullable protected final ConcurrentSkipListMap<Position, BitSet> batchDeletedIndexes;
-    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    protected final ReadWriteLock lock = new ReentrantReadWriteLock();
 
     private RateLimiter markDeleteLimiter;
     // The cursor is considered "dirty" when there are mark-delete updates that are only applied in memory,
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/NonDurableCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/NonDurableCursorImpl.java
@@ -106,8 +106,13 @@ public class NonDurableCursorImpl extends ManagedCursorImpl {
 
         MarkDeleteEntry mdEntry = new MarkDeleteEntry(newPosition, properties, callback, ctx,
                 alignAcknowledgeStatusAfterPersisted);
-        lastMarkDeleteEntry = mdEntry;
-        mdEntry.alignAcknowledgeStatus();
+        lock.writeLock().lock();
+        try {
+            lastMarkDeleteEntry = mdEntry;
+            mdEntry.alignAcknowledgeStatus();
+        } finally {
+            lock.writeLock().unlock();
+        }
         // it is important to advance cursor so the retention can kick in as expected.
         ledger.onCursorMarkDeletePositionUpdated(NonDurableCursorImpl.this, mdEntry.newPosition);
 


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1483,47 +1502,63 @@ protected void internalResetCursor(Position proposedReadPosition,
 
         final Position newMarkDeletePosition = ledger.getPreviousPosition(newReadPosition);
 
+        Runnable alignAcknowledgeStatusAfterPersisted = () -> {
+            // Correct the variable "messagesConsumedCounter".
+            // BTW, no need to change "messagesConsumedCounter" if new "markDeletePosition" is the same as the
+            // old one.
+            int compareRes = ledger.comparePositions(markDeletePosition, newMarkDeletePosition);
```
lhotari, line 1512: I think that `markDeletePosition` should be assigned to a local variable so that the logic would be correct even if it gets updated concurrently. Or is there some other way to ensure consistency?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -268,6 +286,10 @@ public void triggerComplete() {
             }
         }
 
+        public void alignAcknowledgeStatus() {
+            this.alignAcknowledgeStatusAfterPersisted.run();
+        }
```
Technoboy-, line 291: need to check if the `alignAcknowledgeStatusAfterPersisted != null`

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/NonDurableCursorImpl.java
```diff
@@ -101,11 +101,13 @@ void recover(final VoidCallback callback) {
...
     protected void internalAsyncMarkDelete(final Position newPosition, Map<String, Long> properties,
-            final MarkDeleteCallback callback, final Object ctx) {
+            final MarkDeleteCallback callback, final Object ctx, Runnable alignAcknowledgeStatusAfterPersisted) {
         // Bypass persistence of mark-delete position and individually deleted messages info
 
-        MarkDeleteEntry mdEntry = new MarkDeleteEntry(newPosition, properties, callback, ctx);
+        MarkDeleteEntry mdEntry = new MarkDeleteEntry(newPosition, properties, callback, ctx,
+                alignAcknowledgeStatusAfterPersisted);
         lastMarkDeleteEntry = mdEntry;
+        mdEntry.alignAcknowledgeStatus();
```
coderzc: It seems that lock.readLock() should be used to avoid data conflicts.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: It is a final variable, and always not be null. See the constructor method here https://github.com/apache/pulsar/pull/25047/files#diff-fad355f91bd15cc041161f9a46fce62b7fee87fbfb8f0ff8a8b724a1bd1f29eeR257

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/NonDurableCursorImpl.java
poorbarcode: Good point, added

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
poorbarcode: It already is a local variable

PR #17016, opened 2022-08-09
[fix][txn] fix ack with txn compute ackedCount error

congbobo184:
issue: 
reproduce:
1. redeliver one consumer received message of a batch message, will remove this message from this consumer pending acks
2. if this message is removed and doesn't redeliver to another consumer or this consumer, we ack another message in the same batch
3. to compute the consumer unackedCount will get an incorrect batch size, because the message don't in any consumer pending acks
### Motivation
get correct batch size to compute the consumer unackedCount
### Modification
use the correct batch size from the `messageIdDate`
### Verifying this change
add the test

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API: (no)
  - The schema: (no)
  - The default values of configurations: (no)
  - The wire protocol: (no)
  - The rest endpoints: (no)
  - The admin cli options: (no)
  - Anything that affects deployment: (no)

### Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

- [x] `doc-not-needed`

PR #25049, opened 2025-12-09
[fix][broker] Fix acked count calculation when acknowledging with transaction

3pacccccc:
### Motivation
This PR fixes an issue with the acked count calculation during individual message acknowledgment with transaction  when dealing with messages that no longer exist in the pending acknowledgment list. The problem occurs specifically when acknowledging messages that not exist, which leads to incorrect acked count calculations and potential negative unacked message counts.
you can reproduce in `TransactionEndToEndTest` like below:
```java
        final String topic = "persistent://" + NAMESPACE1 + "/my-topic";
        String subName = "my-sub";

        Consumer<String> consumer = pulsarClient
                .newConsumer(Schema.STRING)
                .topic(topic)
                .subscriptionName(subName)
                .subscribe();

        Transaction transaction = getTxn();
        for (int j = 0; j < 10; j++) {
            // mock a message that is not exist.
            MessageIdImpl messageId = new MessageIdImpl(9999L, 9999L, 0);
            try {
                consumer.acknowledgeAsync(messageId, transaction).get();
            } catch (Exception e) {

            }
            System.out.println("unack count:" + getPulsarServiceList().get(0).getBrokerService().getTopic(topic, false)
                    .get().get().getSubscription(subName).getConsumers().get(0).getUnackedMessages());
        }

```
and you can see console output like:
```
unack count:-1
unack count:-2
unack count:-3
unack count:-4
unack count:-5
unack count:-6
unack count:-7
unack count:-8
unack count:-9
unack count:-10
```

### Modifications
- Consumer.java:

  - Modified `getAckOwnerConsumerAndBatchSize` to return batch size 0 for non-existent messages

  - Added check in individualAckWithTransaction to set ackedCount = 0 when batchSize <= 0

  - Improved logging for better debugging of invalid acknowledgment attempts

- MessageIndividualAckTest.java:

  - Added tests for normal and transactional individual acknowledgments

  - Tests verify correct behavior with non-existent messages for both `Shared` and `Key_Shared` subscription types

  - Confirms that acked counts are properly calculated (0 for non-existent messages, correct counts for valid messages)

- TransactionEndToEndTest.java:
  - Updated test expectations to reflect that unacked counts should not be reduced when acknowledging already-redelivered messages(this test was introduced in https://github.com/apache/pulsar/pull/17016, which includes when a client later attempts to acknowledge these already-redelivered messages, the system was incorrectly reducing the unacked message count again, leading to double counting. redelivery logic already subtracts the entire batch size from the unacked count[1], so subsequent acknowledgments should not further reduce this count)

[1]: https://github.com/apache/pulsar/blob/968eb52a697d8566c3217fbb37a12e1340e8ef53/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java#L1172

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/33


3pacccccc:
@lhotari Hi, lari, could you take a look when you have a chance? Thanks.

PR #25050, opened 2025-12-09
[fix][admin] Refactor bookie affinity group sync operations to async in rest api

oneby-wang:
### Motivation

It is not recommended to use sync operations in NIO threads, which would block other operations and degrade the overall performance of the system.

### Modifications

Refactor bookie affinity group sync operations to async in rest api, and add tests.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [x] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/12

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -1013,45 +1013,6 @@ public abstract class NamespacesBase extends AdminResource {
                 });
     }
 
-    @Deprecated
-    protected void internalSetBookieAffinityGroup(BookieAffinityGroupData bookieAffinityGroup) {
-        validateSuperUserAccess();
-        log.info("[{}] Setting bookie-affinity-group {} for namespace {}", clientAppId(), bookieAffinityGroup,
-                this.namespaceName);
-
-        if (namespaceName.isGlobal()) {
-            // check cluster ownership for a given global namespace: redirect if peer-cluster owns it
-            validateGlobalNamespaceOwnership(namespaceName);
-        } else {
-            validateClusterOwnership(namespaceName.getCluster());
-            validateClusterForTenant(namespaceName.getTenant(), namespaceName.getCluster());
-        }
-
-        try {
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v1/Namespaces.java
```diff
@@ -1245,45 +1245,68 @@ public void setPersistence(@Suspended final AsyncResponse asyncResponse, @PathPa
     }
 
     @POST
-    @Path("/{property}/{cluster}/{namespace}/persistence/bookieAffinity")
+    @Path("/{tenant}/{cluster}/{namespace}/persistence/bookieAffinity")
```
oneby-wang: Question: in v1 `Namespaces`, many path variables are named as property, not tenant, should I change the variable name here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -1013,7 +1013,7 @@ protected CompletableFuture<Void> internalUnloadNamespaceAsync() {
                 });
     }
 
-
+    @Deprecated
     protected void internalSetBookieAffinityGroup(BookieAffinityGroupData bookieAffinityGroup) {
```
oneby-wang: Question: this method is no longer used after this PR's refactor, should I remove it or just keep it as deprecated?

Technoboy-: you can remove it

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v1/Namespaces.java
Technoboy-: this is for v1 path, keep it there,

PR #25051, opened 2025-12-09
[fix][cli] Fix some pulsar-admin topicPolicies commands exiting before async operations complete

poorbarcode:
### Background
`PulsarAdminTool` is the management of execution of pulsar admin client commans, it calls `command.run` and then calls`systm.exit`.

### Motivation

The command `pulsar-admin topicPolicies set-replication-clusters` is an asynchronize method, so it can not be executed since `system.exit` runs before the asynchronize method. See also https://github.com/apache/pulsar/blob/v4.1.2/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/PulsarAdminTool.java#L227-L233

When you call `pulsar-admin topicPolicies set-replication-clusters`, it does nothing.

### Modifications

Wrap these async topicPolicies operations with sync() to prevent premature exit
* setReplicationClusters
* setDispatcherPauseOnAckStatePersistent
* getDispatcherPauseOnAckStatePersistent
* removeDispatcherPauseOnAckStatePersistent

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopicPolicies.java
+++ b/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopicPolicies.java
@@ -1959,7 +1959,7 @@ public class CmdTopicPolicies extends CmdBase {
         @Override
         void run() throws PulsarAdminException {
             String persistentTopic = validatePersistentTopic(topicName);
-            getTopicPolicies(isGlobal).setDispatcherPauseOnAckStatePersistent(persistentTopic);
+            sync(() -> getTopicPolicies(isGlobal).setDispatcherPauseOnAckStatePersistent(persistentTopic));
         }
     }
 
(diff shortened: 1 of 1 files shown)


pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopicPolicies.java
```diff
@@ -2019,7 +2019,7 @@ private class SetReplicationClusters extends CliCommand {
         void run() throws PulsarAdminException {
             String persistentTopic = validatePersistentTopic(topicName);
             List<String> clusters = Lists.newArrayList(clusterIds.split(","));
-            getTopicPolicies(isGlobal).setReplicationClusters(persistentTopic, clusters);
+            sync(() -> getTopicPolicies(isGlobal).setReplicationClusters(persistentTopic, clusters));
```
Technoboy-, line 2023: Ah, I see `SetDispatcherPauseOnAckStatePersistent`, `getDispatcherPauseOnAckStatePersistent`, `removeDispatcherPauseOnAckStatePersistent`

lhotari: Yes, please fix also other async methods such as https://github.com/apache/pulsar/blob/f1b66aed03a3752a8a5df213cfc3115512768a0c/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopicPolicies.java#L1997

PR #25052, opened 2025-12-09
[improve][client] Make authorization server metadata path configurable in AuthenticationOAuth2

izumo27:
### Motivation

The client’s OAuth2 authentication plugin can be used for OIDC.
So, the authorization server metadata path is fixed to `/.well-known/openid-configuration`.

However, RFC 8414 defines `/.well-known/oauth-authorization-server` as the default, and some users may want to configure a different path.
https://datatracker.ietf.org/doc/html/rfc8414#section-3
This PR makes the authorization server metadata path configurable.

### Modifications

- Added the `wellKnownMetadataPath` parameter to make the authorization server metadata path configurable.
- Added the `AuthenticationOAuth2StandardAuthzServer` class and the `clientCredentialsWithStandardAuthzServerBuilder` builder, which preconfigure the standard path `/.well-known/openid-configuration` as defined in RFC 8414.

### Verifying this change

  - Added tests for configurable authorization server metadata path

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/izumo27/pulsar/pull/6

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/DefaultMetadataResolver.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/DefaultMetadataResolver.java
@@ -63,10 +63,7 @@ public class DefaultMetadataResolver implements MetadataResolver {
      */
     public static DefaultMetadataResolver fromIssuerUrl(URL issuerUrl, AsyncHttpClient httpClient,
                                                         String wellKnownMetadataPath) {
-        if (wellKnownMetadataPath != null) {
-            return new DefaultMetadataResolver(getWellKnownMetadataUrl(issuerUrl, wellKnownMetadataPath), httpClient);
-        }
-        return new DefaultMetadataResolver(getWellKnownMetadataUrl(issuerUrl), httpClient);
+        return new DefaultMetadataResolver(getWellKnownMetadataUrl(issuerUrl, wellKnownMetadataPath), httpClient);
     }
 
     /**
@@ -84,7 +81,13 @@ public class DefaultMetadataResolver implements MetadataResolver {
                 return URI.create(issuerUrl.toExternalForm() + DEFAULT_WELL_KNOWN_METADATA_PATH).normalize().toURL();
             }
             if (wellKnownMetadataPath.startsWith(WELL_KNOWN_PREFIX)) {
-                return URI.create(issuerUrl.toExternalForm() + wellKnownMetadataPath).normalize().toURL();
+                String issuerUrlString = issuerUrl.toExternalForm();
+                // For OAuth2, insert well-known path before the issuer URL path
+                URL url = new URL(issuerUrlString);
+                String path = url.getPath();
+                String basePath = issuerUrlString.substring(0,
+                        issuerUrlString.length() - (path.isEmpty() ? 0 : path.length()));
+                return URI.create(basePath + wellKnownMetadataPath + path).normalize().toURL();
             } else {
                 throw new IllegalArgumentException("Metadata path must start with '" + WELL_KNOWN_PREFIX
                         + "', but was: " + wellKnownMetadataPath);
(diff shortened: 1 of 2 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2StandardAuthzServer.java
```diff
@@ -0,0 +1,70 @@
...
+        String type = params.getOrDefault(CONFIG_PARAM_TYPE, TYPE_CLIENT_CREDENTIALS);
+        switch(type) {
+            case TYPE_CLIENT_CREDENTIALS:
+                this.flow = ClientCredentialsFlow.fromParameters(params);
+                break;
+            default:
+                throw new IllegalArgumentException("Unsupported authentication type: " + type);
+        }
+    }
+}
```
lhotari, line 70: What's the purpose of this class? Could we eliminate this?

izumo27: This class allows users to switch authentication plugins and use the standard metadata path defined in RFC 8414 without additional configuration.
It’s especially useful for CLI tool, where `ClientCredentialsBuilder` cannot be used.

lhotari:
> However, RFC 8414 defines `/.well-known/oauth-authorization-server` as the default, and some users may want to configure a different path.
> https://datatracker.ietf.org/doc/html/rfc8414#section-3
> This PR makes the authorization server metadata path configurable

@izumo27 It seems that according to RFC 8414, the logic is also different than what there is in the current implementation.

https://www.rfc-editor.org/rfc/rfc8414.html#section-3.1
More specifically this part:

>    If the issuer identifier value contains a path component, any
>    terminating "/" MUST be removed before inserting "/.well-known/" and
>    the well-known URI suffix between the host component and the path
>    component.  The client would make the following request when the
>    issuer identifier is "https://example.com/issuer1" and the well-known
>    URI suffix is "oauth-authorization-server" to obtain the metadata,
>    since the issuer identifier contains a path component:
> 
>      GET /.well-known/oauth-authorization-server/issuer1 HTTP/1.1
>      Host: example.com

Instead of appending `.well-known/oauth-authorization-server` to the path, the logic should be to prefix the path with `/.well-known/oauth-authorization-server`.

izumo27:
@lhotari Thank you for the comment. I’ve fixed it.
- Changed the logic to insert well-known path before the issuer path when `wellKnownMetadataPath` is configured.
- For OIDC, keep the existing implementation that appends well-known path after the issuer path when `wellKnownMetadataPath` is empty.
> The RP would make the following request to the Issuer https://example.com/issuer1 to obtain its configuration information, since the Issuer contains a path component:
> ```
>   GET /issuer1/.well-known/openid-configuration HTTP/1.1
>   Host: example.com
> ```
https://openid.net/specs/openid-connect-discovery-1_0.html

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2StandardAuthzServer.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        Map<String, String> params;
+        try {
+            params = AuthenticationUtil.configureFromJsonString(encodedAuthParamString);
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Malformed authentication parameters", e);
+        }
+
+        // Always set the OAuth 2.0 standard metadata path
+        params.put(FlowBase.CONFIG_PARAM_WELL_KNOWN_METADATA_PATH,
```
lhotari, line 58: should this be `putIfAbsent` to allow overriding the path?

lhotari:
Closing and reopening to start CI with latest master branch changes

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2StandardAuthzServer.java
lhotari: Marking this resolved so that this change will make it into the next set of releases. This can be handled in a follow up PR if necessary.

PR #25053, opened 2025-12-09
[improve][broker] Use atomic counter for ongoing transaction count

3pacccccc:
### Motivation
Currently, `txnMetaMap.size()` is used to determine if a new transaction can be added and for metrics reporting. However, `txnMetaMap` is a `ConcurrentSkipListMap`, and calling `size()` is an O(n) operation, which can consume significant system resources, especially under high transaction loads.

This PR replaces the use of `txnMetaMap.size()` with a dedicated atomic counter (`onGoingTxnCount`) to track the number of ongoing transactions in constant time (O(1)). This improves performance and reduces overhead in transaction management.
### Modifications
1. Introduced a `LongAdder` field `onGoingTxnCount` in `MLTransactionMetadataStore` to maintain the ongoing transaction count.
2. Updated all relevant methods to increment/decrement the counter when transactions are added or removed.
3. Refactored test cases to use `getOnGoingTxnCount()` instead of accessing the map via reflection, improving test clarity and maintainability.
4. Ensured metrics and coordinator stats now use the atomic counter instead of `txnMetaMap.size()`.
### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation
- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/34

### DIFF
--- a/pulsar-transaction/coordinator/src/main/java/org/apache/pulsar/transaction/coordinator/impl/MLTransactionMetadataStore.java
+++ b/pulsar-transaction/coordinator/src/main/java/org/apache/pulsar/transaction/coordinator/impl/MLTransactionMetadataStore.java
@@ -205,8 +205,9 @@ public class MLTransactionMetadataStore
                                     if (newStatus == TxnStatus.COMMITTED || newStatus == TxnStatus.ABORTED) {
                                         transactionLog.deletePosition(txnMetaMap
                                                 .get(transactionId).getRight()).thenAccept(v -> {
-                                                    txnMetaMap.remove(transactionId).getLeft();
-                                                    onGoingTxnCount.decrement();
+                                                    if (txnMetaMap.remove(transactionId) != null) {
+                                                        onGoingTxnCount.decrement();
+                                                    }
                                                 }
                                         );
                                     }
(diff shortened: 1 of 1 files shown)


dao-jun (approved):

pulsar-transaction/coordinator/src/main/java/org/apache/pulsar/transaction/coordinator/impl/MLTransactionMetadataStore.java
```diff
@@ -423,6 +434,7 @@ public CompletableFuture<Void> updateTxnStatus(TxnID txnID, TxnStatus newStatus,
                                         abortedTransactionCount.increment();
                                     }
                                     txnMetaMap.remove(txnID.getLeastSigBits());
+                                    onGoingTxnCount.decrement();
```
lhotari: Are race conditions possible? just wondering if there should be a check that the transaction really was removed once? 
Something like this:

```java
if (txnMetaMap.remove(txnID.getLeastSigBits()) != null) {
   onGoingTxnCount.decrement();
}

```

3pacccccc: Thanks for catching this! It's indeed a valid race condition concern. I've updated the code accordingly. PTAL when you have a moment.

PR #25066, opened 2025-12-11
[fix][broker] PIP-442: Fix race condition in async semaphore permit updates that causes memory limits to become ineffective

lhotari:
### Motivation

The current implementation has a race condition when updating permits in the async semaphore. When a permit is being updated, if it gets released before the updated permit is acquired, the semaphore's internal state becomes inconsistent, causing available permits to gradually increase over time.

This bug was triggered by improper permit handling in:
- `ServerCnx.internalHandleGetTopicsOfNamespace`
- `LookupProxyHandler.internalPerformGetTopicsOfNamespace`

In these methods, the CompletableFuture chain was broken, allowing permits to be released while an update was in progress, creating a race condition.

**Impact:** Memory limits for topic list operations become ineffective over time as the semaphore slowly accumulates extra permits, defeating the purpose of the memory limiting mechanism.

### Modifications

- Fixed the race condition in `AsyncSemaphoreImpl` to ensure correct permit accounting when updating permits
- Corrected the CompletableFuture chains in `ServerCnx.internalHandleGetTopicsOfNamespace` to properly compose futures (changed `thenAccept` to `thenCompose` and added missing `return` statement)
- Corrected the CompletableFuture chains in `LookupProxyHandler.internalPerformGetTopicsOfNamespace` to properly handle both success and failure cases (changed `whenComplete` to `handle` and ensured proper future composition)
- Added a test case that reproduces the race condition and validates the fix

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerBackPressureMultipleConsumersTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerBackPressureMultipleConsumersTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.pulsar.client.api;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.testng.Assert.assertEquals;
 import io.netty.buffer.ByteBuf;
 import java.io.Closeable;
@@ -41,6 +42,8 @@ import org.apache.pulsar.client.impl.PulsarClientImpl;
 import org.apache.pulsar.common.allocator.PulsarByteBufAllocator;
 import org.apache.pulsar.common.api.proto.CommandGetTopicsOfNamespace;
 import org.apache.pulsar.common.naming.NamespaceName;
+import org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiter;
+import org.apache.pulsar.common.semaphore.AsyncDualMemoryLimiterImpl;
 import org.apache.pulsar.common.stats.JvmMetrics;
 import org.apache.pulsar.common.util.DirectMemoryUtils;
 import org.testng.annotations.AfterMethod;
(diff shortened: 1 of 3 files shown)


lhotari:
> How do you find the issue ?

@Technoboy-  I was working on a test case for PIP-442 (part 2, in progress) and found inconsistency in the test case. I was checking whether there was a leak of tokens after the test run, but there was actually more tokens that had been introduced.

lhotari:
@Technoboy- I've added a similar test case where I noticed the issue, 210af30 .

Technoboy- (approved):

pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/LookupProxyHandler.java
```diff
@@ -392,7 +393,7 @@ private void internalPerformGetTopicsOfNamespace(long clientRequestId, String na
                                             return CompletableFuture.completedFuture(null);
                                         });
                             }
-                        }).thenApply(__ -> null);
+                        });
```
Denovo1998, line 396: @lhotari It seems there might be some issues here?

The callback of handle returns a CompletableFuture<Void> (the success branch returns withUpdatedPermits(...), and the failure branch returns completedFuture(null)), so the overall return type of handle will become nested: CompletableFuture<CompletableFuture<Void>> (the "result" of the outer future is the inner future)

This "outer future" will be completed at the moment when the broker request is finished and the handle callback returns the "inner future"; this will cause withAcquiredPermits to think that the "future returned by the function is completed", thereby prematurely releasing initial permits.

Perhaps need 
```java
}).thenCompose(future -> future);
```

Make the final return value a CompletableFuture<Void>, and it should only be considered complete after the "inner future" is also finished. This way, withAcquiredPermits will hold the initial permits until withUpdatedPermits(...) (and subsequent write-back of the response) is truly finished.

Issue #25020, opened 2025-11-26
[Bug] Topic list watcher will miss changes while metadata store connection is disconnected


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

Any released version

### Issue Description

This org.apache.pulsar.broker.service.TopicListService implementation doesn't currently handle session events ([org.apache.pulsar.metadata.api.extended.SessionEvent](https://github.com/apache/pulsar/blob/master/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/extended/SessionEvent.java#L24)) at all. 

The consequence of this is that the state on the client side can get out of sync when the metadata store session has a temporary disruption.

### Error messages

```text
None
```

### Reproducing the issue

-

### Additional information

One potential solution is to retrieve the list of topics for the namespace after the metadata store session resumes after disconnection. It could then compare the changes and reconstruct the missed add/delete events to send to the client.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

lhotari:
> One potential solution is to retrieve the list of topics for the namespace after the metadata store session resumes after disconnection. It could then compare the changes and reconstruct the missed add/delete events to send to the client.

I think this could be useful. The client implementation needs a change since by now, the hash has been ignored. 

The correct approach should be to restart the topic watcher or request a full topic listing if the hash doesn't match after applying the changes. This way the client could be kept in sync with the actual state on the broker side.

oneby-wang:
Hi, @lhotari are you working on this? If this issue is not assigned, I would like to help fix this.

lhotari:
> Hi, [@lhotari](https://github.com/lhotari) are you working on this? If this issue is not assigned, I would like to help fix this.

@oneby-wang Yes, I'm working on this. Btw. There's also PIP-442 part 2 (#25070) which contains a fix for a case where initial change events are lost.

PR #25070, opened 2025-12-12
[improve][broker] PIP-442: Add memory limits for topic list watcher (part 2)

lhotari:
[PIP-442](https://github.com/apache/pulsar/blob/master/pip/pip-442.md)

### Motivation

See [PIP-442](https://github.com/apache/pulsar/blob/master/pip/pip-442.md).

This is part 2 for the PIP-442 implementation. This covers the topic list watcher operations. 
The first part of PIP-442 implementation was #24833.

This PR implements memory limits and flow control for the `CommandWatchTopicList` feature as specified in PIP-442. Currently, topic list watcher operations can allocate unbounded amounts of memory based on namespace size, creating the final significant gap in Pulsar's otherwise comprehensive memory management system.

The lack of memory limits creates several critical issues:
- **Broker Memory Exhaustion Risk**: Large namespaces can trigger OutOfMemoryErrors when multiple clients establish topic list watchers simultaneously
- **Unpredictable Resource Usage**: Operators cannot reliably predict or limit total broker memory consumption due to this unbounded allocation path
- **Performance Degradation**: Even without OOM, large topic list operations cause GC pressure and latency spikes affecting all broker operations

This PR completes Pulsar's memory management framework by implementing memory-based flow control and backpressure specifically for topic list watcher operations.

### Modifications

- use memory limits in TopicListService / PulsarCommandSenderImpl so that topic list watcher related operations are covered
- **Modified PulsarCommandSender interface**: Changed `sendWatchTopicListSuccess` and `sendWatchTopicListUpdate` methods to return `CompletableFuture<Void>` and accept permit acquire error handlers for asynchronous memory management
- **Enhanced TopicListService**: 
  - Added permit-based memory control with retry logic for sending topic list responses
  - Implemented backoff strategy for permit acquisition with configurable timeouts and queue sizes
  - Added proper cleanup and cancellation support when watcher operations fail or are closed
- **Added comprehensive testing**: Created `PatternConsumerTopicWatcherBackPressureMultipleConsumersTest` to verify backpressure behavior with multiple concurrent consumers
- **Updated existing tests**: Modified `TopicListServiceTest` and `TopicListWatcherTest` to accommodate the new asynchronous response pattern

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/distribution/server/src/assemble/LICENSE.bin.txt
+++ b/distribution/server/src/assemble/LICENSE.bin.txt
@@ -508,11 +508,11 @@ The Apache Software License, Version 2.0
   * JCTools - Java Concurrency Tools for the JVM
     - org.jctools-jctools-core-4.0.5.jar
   * Vertx
-    - io.vertx-vertx-auth-common-4.5.22.jar
-    - io.vertx-vertx-bridge-common-4.5.22.jar
-    - io.vertx-vertx-core-4.5.22.jar
-    - io.vertx-vertx-web-4.5.22.jar
-    - io.vertx-vertx-web-common-4.5.22.jar
+    - io.vertx-vertx-auth-common-4.5.24.jar
+    - io.vertx-vertx-bridge-common-4.5.24.jar
+    - io.vertx-vertx-core-4.5.24.jar
+    - io.vertx-vertx-web-4.5.24.jar
+    - io.vertx-vertx-web-common-4.5.24.jar
   * Apache ZooKeeper
     - org.apache.zookeeper-zookeeper-jute-3.9.4.jar
   * Snappy Java
--- a/jetty-upgrade/zookeeper-with-patched-admin/pom.xml
+++ b/jetty-upgrade/zookeeper-with-patched-admin/pom.xml
@@ -102,4 +102,43 @@
       </plugin>
     </plugins>
   </build>
+  <profiles>
+    <profile>
+      <!-- Profile that is activated when using IntelliJ IDEA -->
+      <id>intellij</id>
+      <activation>
+        <property>
+          <name>idea.maven.embedder.version</name>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <!-- Add shaded jar to artifacts for IntelliJ IDEA so that IntelliJ builds
+          and running tests work -->
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>build-helper-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>attach-shade-jar</id>
+                <phase>package</phase>
+                <goals>
+                  <goal>attach-artifact</goal>
+                </goals>
+                <configuration>
+                  <artifacts>
+                    <artifact>
+                      <file>${project.build.directory}/${project.build.finalName}.jar</file>
+                      <type>jar</type>
+                      <classifier>shaded</classifier>
+                    </artifact>
+                  </artifacts>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
 </project>
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
@@ -2193,6 +2193,11 @@ public class ManagedLedgerImpl implements ManagedLedger, CreateCallback {
         return Optional.ofNullable(ledgers.get(ledgerId));
     }
 
+    public CompletableFuture<ReadHandle> reopenReadHandle(long ledgerId) {
+        invalidateReadHandle(ledgerId);
+        return getLedgerHandle(ledgerId);
+    }
+
     CompletableFuture<ReadHandle> getLedgerHandle(long ledgerId) {
         CompletableFuture<ReadHandle> ledgerHandle = ledgerCache.get(ledgerId);
         if (ledgerHandle != null) {
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImpl.java
@@ -32,6 +32,7 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.LongAdder;
+import java.util.function.Function;
 import java.util.function.IntSupplier;
 import org.apache.bookkeeper.client.api.BKException;
 import org.apache.bookkeeper.client.api.LedgerEntry;
@@ -519,6 +520,11 @@ public class RangeEntryCacheImpl implements EntryCache {
      */
     CompletableFuture<List<Entry>> readFromStorage(ReadHandle lh, long firstEntry, long lastEntry,
                                                    IntSupplier expectedReadCount) {
+        return readFromStorage(lh, firstEntry, lastEntry, expectedReadCount, true);
+    }
+
+    private CompletableFuture<List<Entry>> readFromStorage(ReadHandle lh, long firstEntry, long lastEntry,
+                                                          IntSupplier expectedReadCount, boolean allowRetry) {
         final int entriesToRead = (int) (lastEntry - firstEntry) + 1;
         CompletableFuture<List<Entry>> readResult = ReadEntryUtils.readAsync(ml, lh, firstEntry, lastEntry)
                 .thenApply(
@@ -550,17 +556,30 @@ public class RangeEntryCacheImpl implements EntryCache {
                                 ledgerEntries.close();
                             }
                         });
-        // handle LH invalidation
-        readResult.exceptionally(exception -> {
-            if (exception instanceof BKException
-                    && ((BKException) exception).getCode() == BKException.Code.TooManyRequestsException) {
-            } else {
+
+        return readResult.handle((entries, exception) -> {
+            if (exception == null) {
+                return CompletableFuture.completedFuture(entries);
+            }
+
+            Throwable cause = FutureUtil.unwrapCompletionException(exception);
+            if (allowRetry && cause instanceof ManagedLedgerException.OffloadReadHandleClosedException) {
+                log.info("[{}] Read handle closed for ledger {}, reopening", ml.getName(), lh.getId());
+                pendingReadsManager.invalidateLedger(lh.getId());
+                return ml.reopenReadHandle(lh.getId())
+                        .thenCompose(reopened -> readFromStorage(reopened, firstEntry, lastEntry, expectedReadCount,
+                                false));
+            }
+
+            if (!(cause instanceof BKException
+                    && ((BKException) cause).getCode() == BKException.Code.TooManyRequestsException)) {
                 ml.invalidateLedgerHandle(lh);
                 pendingReadsManager.invalidateLedger(lh.getId());
             }
-            return null;
-        });
-        return readResult;
+
+            CompletableFuture<List<Entry>> failedFuture = CompletableFuture.failedFuture(cause);
+            return failedFuture;
+        }).thenCompose(Function.identity());
     }
 
     @Override
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImplTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/cache/RangeEntryCacheImplTest.java
@@ -32,8 +32,12 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.IntSupplier;
+import org.apache.bookkeeper.client.api.LedgerEntries;
+import org.apache.bookkeeper.client.api.LedgerEntry;
 import org.apache.bookkeeper.client.api.ReadHandle;
+import org.apache.bookkeeper.client.impl.LedgerEntryImpl;
 import org.apache.bookkeeper.mledger.AsyncCallbacks;
 import org.apache.bookkeeper.mledger.Entry;
 import org.apache.bookkeeper.mledger.ManagedLedgerException;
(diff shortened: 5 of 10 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/client/api/PatternConsumerTopicWatcherBackPressureMultipleConsumersTest.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        validateThatTokensHaventLeakedOrIncreased();
+    }
+
+    protected void validateThatTokensHaventLeakedOrIncreased() {
+        AsyncDualMemoryLimiterImpl limiter =
+                pulsar.getBrokerService().getMaxTopicListInFlightLimiter();
+        assertThat(limiter.getLimiter(AsyncDualMemoryLimiter.LimitType.HEAP_MEMORY).getAvailablePermits())
+                .isEqualTo(pulsar.getConfiguration().getMaxTopicListInFlightHeapMemSizeMB() * 1024 * 1024);
```
Denovo1998: If `getMaxTopicListInFlightHeapMemSizeMB()` or `getMaxTopicListInFlightDirectMemSizeMB()` returns a value of 2048 or more, the expression `* 1024 * 1024` will overflow a 32-bit integer. To prevent this, use `1024L` to ensure the computation is performed with long integers.

```diff
-        .isEqualTo(pulsar.getConfiguration().getMaxTopicListInFlightHeapMemSizeMB() * 1024 * 1024);
+        .isEqualTo(pulsar.getConfiguration().getMaxTopicListInFlightHeapMemSizeMB() * 1024L * 1024);
-        .isEqualTo(pulsar.getConfiguration().getMaxTopicListInFlightDirectMemSizeMB() * 1024 * 1024);
+        .isEqualTo(pulsar.getConfiguration().getMaxTopicListInFlightDirectMemSizeMB() * 1024L * 1024);
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -288,8 +405,67 @@ public void deleteTopicListWatcher(Long watcherId) {
...
+            Throwable unwrappedException = FutureUtil.unwrapCompletionException(t);
+            if (unwrappedException instanceof AsyncSemaphore.PermitAcquireCancelledException
+                    || unwrappedException instanceof AsyncSemaphore.PermitAcquireAlreadyClosedException
+                    || !connection.isActive()
+                    || !watchers.containsKey(watcherId)) {
+                return;
+            }
+            long retryDelay = retryBackoff.next();
+            log.info("[{}] Cannot acquire direct memory tokens for sending {}. Retry {} in {} ms. {}", connection,
+                    operationName, retryCount.get(), retryDelay, t.getMessage());
```
Denovo1998, line 560: The `retryCount` variable is created but never incremented.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -218,31 +263,100 @@ public void handleWatchTopicList(NamespaceName namespaceName, long watcherId, lo
                 });
     }
 
+    private void sendTopicListSuccessWithPermitAcquiringRetries(long watcherId, long requestId, List<String> topicList,
+                                                                String hash,
+                                                                Runnable successfulCompletionCallback) {
+        performOperationWithPermitAcquiringRetries(watcherId, "topic list success", permitAcquireErrorHandler ->
```
Denovo1998, line 309: If an earlier update A is removed from the semaphore queue due to timeout/queuefull and retried, and a later update B gets the permits first and sends successfully, it is possible for B to arrive first and A to arrive later.

For the pattern consumer, out-of-order updates may lead to inconsistent final subscription sets (typical example: A=Create topic2, B=Delete topic2; if B arrives first, the client may mistakenly retain topic2).

Perhaps we need to maintain a "send chain" or queue for each watcher, ensuring that only one update/success is trying to acquire + write at any moment; if it fails, retry this "head of the queue," and subsequent updates cannot bypass it?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -50,15 +68,21 @@ public static class TopicListWatcher implements BiConsumer<String, NotificationT
         private final long id;
         /** The regexp for the topic name(not contains partition suffix). **/
         private final TopicsPattern topicsPattern;
+        private final Executor executor;
+        private volatile boolean closed = false;
+        private boolean sendTopicListSuccessCompleted = false;
+        private BlockingDeque<Runnable> sendTopicListUpdateTasksBeforeInit = new LinkedBlockingDeque<>();
```
Denovo1998: No capacity limit. Will there be any issues here? When a watcher has registered a listener (to "not miss events"), but sendWatchTopicListSuccess cannot be obtained for a long time due to pressure on direct permits (or repeatedly times out and retries), all updates will enter this queue.

Additionally, since both sendTopicListUpdate(...) and sendTopicListSuccessCompleted() are synchronized here, maybe the concurrency capability of `LinkedBlockingDeque` is redundant?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: In this case,  there isn't a possibility for a reordering problem with `sendTopicListSuccessWithPermitAcquiringRetries` since the updates will get buffered until the topic list success has been sent out. I agree that the out-of-order case is possible for `sendTopicListUpdate` method. In that case, it would be necessary to apply the queuing/buffering solution in TopicListWatcher class to all updates and only send out the next update after the previous update finishes. I'll think about handling that.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: Valid point. There could be an upper bound for the buffer so that exceeding the buffer would result in dismissing the updates and reading the complete topic listing.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/TopicListServiceTest.java
```diff
@@ -88,10 +178,37 @@ public void testCommandWatchSuccessResponse() {
...
+                13,
+                7,
+                "persistent://tenant/ns/topic\\d",
+                topicsPatternImplementation, null,
+                lookupSemaphore);
+        List<String> topics = Collections.singletonList("persistent://tenant/ns/topic1");
+        String hash = TopicList.calculateHash(topics);
+        topicListFuture.complete(topics);
+        // wait for acquisition to timeout a few times
+        Thread.sleep(2000);
```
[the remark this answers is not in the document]
lhotari: In this case Thread.sleep is used to cause a timeout.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: This has been addressed by setting an upper bound for the buffer.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: The ordering problem has been addressed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -287,9 +422,70 @@ public void deleteTopicListWatcher(Long watcherId) {
...
+                () -> connection.getCommandSender()
+                        .sendWatchTopicListUpdate(watcherId, newTopics, deletedTopics, topicsHash,
+                                permitAcquireErrorHandler)
+                        .whenComplete((__, t) -> {
+                            if (t != null) {
+                                // this is an unexpected case
+                                log.warn("[{}] Failed to send topic list update for watcherId={}. Watcher will be in "
+                                        + "inconsistent state.", connection, watcherId, t);
+                            }
+                            completionCallback.run();
```
Denovo1998, line 511: Even if update A fails due to "acquire timeout / queue full" and a retry has been scheduled, the completionCallback will still trigger TopicListWatcher.sendingCompleted(), and the queue will continue to execute the subsequent update B.

Assume there are two updates on the same watcher (in the order of event occurrence): A then B.

  - t0：update A (Create topic2) triggers, enters the sending chain, and starts attempting to acquire direct permits.
  - t1：Direct permits not enough, A acquire failed:
      - acquireDirectMemoryPermitsAndWriteAndFlush calls permitAcquireErrorHandler.accept(failure) → Schedule "retry A later"
      - At the same time, the future returned by sendWatchTopicListUpdate also completes exceptionally → sendTopicListUpdate(...).whenComplete is called
      - whenComplete still calls completionCallback.run() → sendingCompleted() is triggered → the queue releases the next item
  - t2：Update B (Delete topic2) starts executing. Maybe this time the permits are sufficient, so B successfully writes to the socket first.
  - t3：t1's scheduled "Retry A" has expired, attempting to acquire again. This time it succeeds, so A is then written back to the socket.

**Result: The sequence in which the client receives the messages becomes B (Delete) first, followed by A (Create). As a result, the client may incorrectly "retain topic2".**

---

For each watcher, perhaps we need a strict "send chain":

  - Head task (success or some update) is only allowed to advance sendingCompleted() to the next one after successfully writing (at least the write is committed)?
  - If the head of the queue enters retry due to permit acquire failure, then it is still the head of the queue: subsequent updates must not skip over it?

lhotari: Yes there are gaps. I haven't yet created a test case to clarify and validate the behavior. The `permitAcquireErrorHandler` breaks the completablefuture chain and I'll have to revisit that part. Instead of `Consumer<Throwable> permitAcquireErrorHandler`, I think it should be `Function<Throwable, CompletableFuture<Void>> permitAcquireErrorHandler` so that the chain doesn't break.

Regarding the update events, there is no guarantee that the MetadataStore implementation will send all updates. For example, with ZooKeeper, this can happen. There's issue #25020 about this. I've discussed this with Matteo and the resolution would be to keep on polling even when topic list watchers are used. Another improvement would be to handle the session events immediately when they occur.

lhotari: @Denovo1998 The ordering issue is handled now. There's also a test to verify the behavior at the TopicListService level.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -218,31 +278,103 @@ public void handleWatchTopicList(NamespaceName namespaceName, long watcherId, lo
...
+                                    isPermitRequestCancelled, updatedPermits -> {
+                                        // reset retry backoff
+                                        retryBackoff.reset();
+                                        // just return the watcher which was already created before
+                                        return CompletableFuture.completedFuture(watcher);
+                                    }, CompletableFuture::failedFuture);
+                        }).whenComplete((watcher, exception) -> {
+                            if (exception != null) {
+                                if (watcherRef.get() != null) {
+                                    watcher.close();
```
Denovo1998: Is there a potential NPE here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -71,24 +102,67 @@ public List<String> getMatchingTopics() {
...
+                    this::sendingCompleted);
+            if (!sendingInProgress) {
+                sendingInProgress = true;
+                executor.execute(task);
+            } else {
+                // if sendTopicListSuccess hasn't completed, add to a queue to be executed after it completes
+                if (!sendTopicListUpdateTasks.offer(task)) {
+                    log.warn("Dropping update for watcher id {} matching {} since queue is full.", id,
+                            topicsPattern.inputPattern());
+                }
```
Denovo1998, line 151: Is there consistency risk?

After the delta update, there is no hash check / automatic full correction mechanism on the client side (`pulsar-client/TopicListWatcher` just appends add/remove), and the final subscription set may permanently deviate from the real state.

Perhaps it needs to be changed to:

- Close watcher when the queue is full (forcing the client to rebuild and resynchronize fully)?
- Or change the "queue" from storing delta runnable to "merged": only ensure the final state (e.g., only keep one task that "requires recalculation and send diff"), to avoid infinite accumulation of delta?

lhotari: Yes, this is a valid concern. The issue #25020 is also related.

In this case, it's most likely useful to close the topic list watcher. One of the challenges is that the original protocol for Topic list watchers has some gaps. The `CommandWatchTopicListClose` command isn't designed to be sent from the broker to the client at the moment. The only way that a broker can close a watcher is by closing the connection. However, this is problematic since closing the connection will also close all producers and consumers sharing the same connection.

Another solution that could be useful would be to mark the state as dirty and perform the complete listing so that the hash of the topic listing would be in consistent state. 

Besides #25020, another related detail is that Metadata Store events could be lost. Because of that, topic list watcher should perform a full refresh with a configurable interval, for example every 5 minutes. 

Another gap in the topic list watcher protocol is that the broker cannot send the updated hash to the client. A better protocol would be to send the plain hash to the client each time the broker performs a full refresh. If there's a difference to the state on the client side, the client could request the full listing.
Currently, the client should perform a scheduled polling to ensure that the state doesn't get out of sync from the broker side state. This isn't properly implemented in the Java client at the moment. Similar issues might be present also in other Pulsar client implementations.

lhotari: I'll address this in this PR by marking the state invalid and performing a complete topic listing in that case. The client should notice the state diversion by comparing hashes. This is also missing from the client implementation.

lhotari: This is now addressed, however the code coverage in tests should be improved.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: Addressed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: I've added some test coverage to TopicListService level for refreshing all topics when the queue overflows. To fix the problem, it will be necessary to also address #25020. I think it's better to have that in a separate PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -18,77 +18,171 @@
...
+            this.matchingTopics =
+                    TopicList.filterTopics(topics, topicsPattern, Collectors.toCollection(LinkedHashSet::new));
+            // start with in progress state since topic list update will be sent first
+            this.sendingInProgress = true;
+            this.sendTopicListUpdateTasks =
+                    new LinkedBlockingDeque<>(topicListUpdateMaxQueueSize);
         }
 
-        public List<String> getMatchingTopics() {
+        public synchronized Collection<String> getMatchingTopics() {
```
BewareMyPower, line 98: Adding `synchronized` is useless without copying `matchingTopics`.

BewareMyPower (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
```diff
@@ -18,77 +18,171 @@
...
+        private final NamespaceName namespace;
         /** The regexp for the topic name(not contains partition suffix). **/
         private final TopicsPattern topicsPattern;
+        private final Executor executor;
+        private volatile boolean closed = false;
+        private boolean sendingInProgress;
+        private final BlockingDeque<Runnable> sendTopicListUpdateTasks;
+        private boolean updatingTopics;
+
+        public TopicListWatcher(TopicListService topicListService, long id,
```
BewareMyPower: Please add `@VisibleForTesting`.

BTW, when I read code recently, I found many classes have many constructors or factory methods but most of them are only used in tests, e.g. `EntryImpl`. It makes code hard to read. IMO, such method can be replaced by an util method in tests rather than the implementation:

```java
public class TopicListServiceTest {

    private TopicListService newTopicListService() {
        return new TopicListService(pulsar, connection, true, 30, 10000);
    }

    private TopicListService newTopicListService(int maxQueueSize) {
        return new TopicListService(pulsar, connection, true, 30, maxQueueSize);
    }
```

BewareMyPower:
Overall LGTM except for my comments

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: thanks for pointing this out. addressed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/TopicListService.java
lhotari: makes sense. addressed.

BewareMyPower (approved):

Denovo1998 (approved):

PR #25072, opened 2025-12-14
[feat][client] PIP-234: Support shared resources in AuthenticationOAuth2 to reduce thread usage

### LINKED ISSUE #24795: [feat] Authentication implementations such as AuthenticationOAuth2 should support sharing thread pools or DNS cache
### Search before reporting

### Motivation

PIP-234 implementation #24790 adds support for sharing thread pools and DNS resolver/cache across multiple client instances. However, authentication implementations such as AuthenticationOAuth2 create new thread pools and DNS resolvers/caches. This isn't desired, since it would be useful to be able to share the Netty Event Loop and configure the DNS settings in the client and use them for all Pulsar client DNS lookups, at least for most common use cases.

### Solution

The org.apache.pulsar.client.api.Authentication interface could be changed so that the "container" could pass a context instance with contains methods to lookup shared instances.
```java
    /**
     * Initialize the authentication provider.
     */
    default void start(AuthenticationInitContext context) throws PulsarClientException {
        start();
    }

    /**
     * Initialize the authentication provider.
     */
    void start() throws PulsarClientException;
```

The AuthenticationInitContext could contain generic "getService" and "getServiceByName" methods:
```java
public interface AuthenticationInitContext {
    <T> Optional<T> getService(Class<T> serviceClass);
    <T> Optional<T> getServiceByName(Class<T> serviceClass, String name);
}
```

This solution could be used to pass the Netty event loop, DNS resolver and Timer to be shared in the AuthenticationOAuth2 implementation.

### Alternatives

_No response_

### Anything else?

_No response_

3pacccccc:
fix: https://github.com/apache/pulsar/issues/24795

🔗 Related links
- issue: https://github.com/apache/pulsar/issues/24795
- comments: https://github.com/apache/pulsar/pull/24944#issuecomment-3520852664
### Motivation

OAuth2 authentication currently creates its own HTTP client resources for metadata discovery and token exchange. These resources are not integrated with the client resource sharing introduced by PIP-234, so OAuth2 can create separate thread pools and DNS resolver/cache state instead of using the client-provided configuration.

At the same time, authentication implementations should not expose shaded AsyncHttpClient or Netty classes through the public client API. This PR adds a small authentication-specific abstraction so authentication providers can obtain a configured HTTP client from the Pulsar client container without depending on shaded implementation types.

OAuth2 token exchange is still synchronous today, so authentication HTTP I/O must not run on the Pulsar client I/O event loop. Otherwise, broker internal clients can block during authentication and cause topic loading to time out, such as in `Oauth2PerformanceTransactionTest.testTransactionPerf`.

### Modifications

- Added `AuthenticationInitContext` and a default `Authentication.start(AuthenticationInitContext)` method.
- Added authentication HTTP abstractions:
  - `AuthenticationHttpClientFactory`
  - `AuthenticationHttpClient`
  - `AuthenticationHttpClientConfig`
  - `AuthenticationHttpRequest`
  - `AuthenticationHttpResponse`
- Added `AuthenticationInitContextImpl` in the client implementation.
- Updated `PulsarClientImpl` to provide an authentication HTTP client factory through the initialization context.
- Reused the client DNS resolver group configuration/cache for authentication HTTP requests.
- Reused the client timer for authentication HTTP requests.
- Kept authentication HTTP I/O on a dedicated lazy event loop so synchronous OAuth2 token exchange does not block Pulsar client I/O threads.
- Reused the same authentication HTTP client factory across `updateAuthentication()` calls instead of creating a new factory every time.
- Refactored OAuth2 metadata resolution and token exchange to use the new HTTP abstraction instead of directly depending on AsyncHttpClient.
- Preserved the previous fallback behavior when `trustCertsFilePath` cannot build a custom SSL context.
- Added cleanup for TLS refresh/SSL resources when HTTP client construction fails.

- [x] The public API
- [ ] The schema
- [x] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/35


lhotari (commented):
I think that there's a need to revisit the original design expressed in#24795. The comment in https://github.com/apache/pulsar/pull/24944#issuecomment-3520852664 contains some details of the direction:

> For authentication plugins, one possibility would be that there would be a way to directly get a HTTP client with given configuration (such as trustCertsFilePath) that is sufficiently abstracted so that there's no need to couple to AsyncHttpClient in authentication plugin implementations.

One of the problems in exposing Netty and AsyncHttpClient classes in the public interfaces is that the classes are different when using a shaded Pulsar client. 
That is why there should be a sufficient abstraction for the HTTP client.

For Pulsar internal authentication plugins, it wouldn't be a problem to support the solution in this PR since the classes are also shaded in the shaded Pulsar client and there wouldn't be a problem with unshaded Netty and shaded Netty classes, for example. However, it might be work trying to find a more stable abstraction that expressed in the original design.

3pacccccc:
@lhotari hi, Lari, could you please take a look again?

lhotari:
For Pulsar 5.0 there's PIP-478 in voting, https://lists.apache.org/thread/6n6q014236l4tstrxn9t4jbrp8xty997. Please check it. It also covers shared resources for auth plugins.

PR #25076, opened 2025-12-16
[fix][broker] fix delayedMessagesCount error in InMemoryDelayedDeliveryTracker

TakaHiR07:
### Motivation

Occur a NPE issue in delay message. And then find the reason is in delayedMessagesCount. When InMemoryDelayedDeliveryTracker#addMessage(), it don't judge whether the entryId is exist in roaringbitmap, that result in the delayedMessagesCount of the map size is not correct. 

企业微信截图_bb4f30f2-8829-413d-a7e8-e8746dc07adc

### Modifications

1. add test to test the duplicate entry case 
2. check whether roaring64Bitmap contains entryId
3. log error for the case of "n < 0" in getScheduledMessages(), since this case should not occur

### Verifying this change

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -130,7 +130,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
                 .computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
                 .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap());
         if (!roaring64Bitmap.contains(entryId)) {
-            roaring64Bitmap.add(entryId);
+            roaring64Bitmap.addLong(entryId);
             delayedMessagesCount.incrementAndGet();
         }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -126,10 +126,13 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
         }
 
         long timestamp = trimLowerBit(deliverAt, timestampPrecisionBitCnt);
-        delayedMessageMap.computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
-                .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap())
-                .add(entryId);
-        delayedMessagesCount.incrementAndGet();
+        Roaring64Bitmap roaring64Bitmap = delayedMessageMap
+                .computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
+                .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap());
+        if (!roaring64Bitmap.contains(entryId)) {
+            roaring64Bitmap.add(entryId);
+            delayedMessagesCount.incrementAndGet();
+        }
 
         updateTimer();
 
@@ -200,7 +203,12 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
                     delayedMessagesCount.addAndGet(-n);
                     n = 0;
                 }
-                if (n <= 0) {
+                if (n == 0) {
+                    break;
+                } else if (n < 0) {
+                    // should not go into this situation
+                    log.error("[{}] Delayed message tracker getScheduledMessages should not < 0, number is: {}",
+                            dispatcher.getName(), n);
                     break;
                 }
             }
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
@@ -274,4 +274,69 @@ public class InMemoryDeliveryTrackerTest extends AbstractDeliveryTrackerTest {
         tracker.close();
     }
 
+    @Test(dataProvider = "delayedTracker")
+    public void testDelayedMessagesCountWithDuplicateEntryId(InMemoryDelayedDeliveryTracker tracker) throws Exception {
+        assertFalse(tracker.hasMessageAvailable());
+
+        // case1: addMessage() with duplicate entryId,
+        // getScheduledMessages() enter "cardinality <= n" and make tracker empty
+        assertTrue(tracker.addMessage(1, 1, 10));
+        assertTrue(tracker.addMessage(1, 2, 20));
+        assertTrue(tracker.addMessage(1, 2, 20));
+        assertTrue(tracker.addMessage(1, 3, 40));
+        assertTrue(tracker.addMessage(1, 4, 50));
+
+        clockTime.set(50);
+        assertTrue(tracker.hasMessageAvailable());
+        assertEquals(tracker.getNumberOfDelayedMessages(), 4L);
+        assertEquals(tracker.delayedMessageMap.size(), 4L);
+
+        Set<Position> scheduled = tracker.getScheduledMessages(10);
+        assertEquals(scheduled.size(), 4);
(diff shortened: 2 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -126,10 +126,13 @@ public boolean addMessage(long ledgerId, long entryId, long deliverAt) {
...
         long timestamp = trimLowerBit(deliverAt, timestampPrecisionBitCnt);
-        delayedMessageMap.computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
-                .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap())
-                .add(entryId);
-        delayedMessagesCount.incrementAndGet();
+        Roaring64Bitmap roaring64Bitmap = delayedMessageMap
+                .computeIfAbsent(timestamp, k -> new Long2ObjectRBTreeMap<>())
+                .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap());
+        if (!roaring64Bitmap.contains(entryId)) {
+            roaring64Bitmap.add(entryId);
```
lhotari: It looks like `.addLong` should be used in the Roaring64Bitmap API.
```suggestion
            roaring64Bitmap.addLong(entryId);
```
The `.add` method works too, but the method signature takes a long array (long...). Perhaps the compiler is able to optimize that, so it might not make a difference.

It's unfortunate that `Roaring64Bitmap` doesn't have the `checkedAdd` method as there is in `RoaringBitmap`. That would eliminate the need for the `.contains` check.

lhotari:
btw. this code location was discussed in the review: https://github.com/apache/pulsar/pull/24430/changes#r2156278377

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
TakaHiR07: Yes, unfortunately. 

"add" or "addLong" I guess is the same after compile, sure it may be better to use "addLong" directly.

Besides, I think there is no need to consider concurrent situation in InMemoryDelayedDeliveryTracker. I don't see any code point out that concurrent situation would occur.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
Denovo1998, line 139: One thought: should updateTimer() and checkAndUpdateHighest(deliverAt) run only when we actually insert a new entryId?

With the current structure, duplicate addMessage() calls still update highestDeliveryTimeTracked / messagesHaveFixedDelay, which could disable the fixed-delay optimization even though the tracker state didn’t change.”

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -200,7 +203,12 @@ public NavigableSet<Position> getScheduledMessages(int maxMessages) {
                     delayedMessagesCount.addAndGet(-n);
                     n = 0;
                 }
-                if (n <= 0) {
+                if (n == 0) {
+                    break;
+                } else if (n < 0) {
+                    // should not go into this situation
+                    log.error("[{}] Delayed message tracker getScheduledMessages should not < 0, number is: {}",
+                            dispatcher.getName(), n);
```
Denovo1998, line 211: About the new n < 0 branch: this should be unreachable in normal flow. One potential way to hit it is int overflow from int cardinality = (int) entryIds.getLongCardinality().

Would it be better to keep cardinality as long (and compare cardinality <= (long) n) to eliminate overflow, instead of only logging when n < 0?

pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
```diff
@@ -274,4 +274,69 @@ public void testDelaySequence(InMemoryDelayedDeliveryTracker tracker) throws Exc
...
+        assertEquals(tracker.getNumberOfDelayedMessages(), 4L);
+        assertEquals(tracker.delayedMessageMap.size(), 4L);
+
+        Set<Position> scheduled = tracker.getScheduledMessages(10);
+        assertEquals(scheduled.size(), 4);
+        assertEquals(tracker.getNumberOfDelayedMessages(), 0L);
+
+
+
+        // case2: addMessage() with duplicate entryId,
```
Denovo1998, line 301: In case2 the comment says it enters cardinality > n, but with getScheduledMessages(10) and 4 unique entryIds it should hit the cardinality <= n branch. Could we adjust the comment to match the scenario (case3 seems to be the one exercising cardinality > n)?

Denovo1998 (changes requested):
Left some comments.

Denovo1998:
> btw. this code location was discussed in the review: https://github.com/apache/pulsar/pull/24430/changes#r2156278377

@thetumbled Do you have a chance to review the current PR?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
Denovo1998: Threading question: addMessage() and getScheduledMessages() are invoked under synchronized (this) in the dispatcher (e.g. PersistentDispatcherMultipleConsumers#trackDelayedDelivery), but clearDelayedMessages() doesn’t seem synchronized and InMemoryDelayedDeliveryTracker#clear() isn’t synchronized either.

Is clear() guaranteed to be called under the same lock, or should we align with BucketDelayedDeliveryTracker#clear() (synchronized) to avoid concurrent access to delayedMessageMap/bitmaps?

thetumbled:
> btw. this code location was discussed in the review: https://github.com/apache/pulsar/pull/24430/changes#r2156278377

Maybe we should figure out the reason why duplicate entry IDs are added multiple times if this class does not intentionally allow that behavior.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
TakaHiR07: You are right. I think it is another issue and both use long value is better. Do you think we fix it in this pr or you push another pr to fix?

pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
TakaHiR07: Yes. have changed the comment

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
TakaHiR07: Great catch. I have checked the earliest fixed delay implementation in https://github.com/apache/pulsar/pull/16609. In the earliest implementation, the issue is already exist. When duplicate addMessage(), highestDeliveryTimeTracked is ok, but messagesHaveFixedDelay would be set to false incorrectly.

I would check why exist duplicate addMessage() later. And I prefer that we open another pr to fix the additional issue.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
lhotari: @TakaHiR07 Please check the previous comment by @Denovo1998.

lhotari:
#25681 addresses the same issue

lhotari:
> > btw. this code location was discussed in the review: https://github.com/apache/pulsar/pull/24430/changes#r2156278377
> 
> Maybe we should figure out the reason why duplicate entry IDs are added multiple times if this class does not intentionally allow that behavior.

@thetumbled One reason finally came up to the surface with PR #26012 (reported issue is #25996)

lhotari:
closing this one since #25681 is merged and addressed the same problem.

PR #25077, opened 2025-12-16
[fix][broker] Fix chunked message loss when no consumers are available

zjxxzjwang:
### Motivation

When no consumers are online for a subscription, sending chunked messages will be lost.

Clipboard_Screenshot_1765871357

### Modifications
When no subscribers are online, send chunk messages to deliver them to the redeliveryMessages queue. When a subscriber comes online, it can consume the messages again.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SharedConsumerAssignorTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SharedConsumerAssignorTest.java
@@ -240,7 +240,7 @@ public class SharedConsumerAssignorTest {
     }
 
     /**
-     * When there are no consumers online, chunk messages will not be directly lost
+     * When there are no consumers online, chunk messages will not be directly lost.
      */
     @Test
     public void testChunkMessagesNotBeLostNoConsumer() {


zjxxzjwang:
> Thanks for the PR @zjxxzjwang . Please add a test that demonstrates the broken behavior and shows that the changes in this PR fix the problem.

@lhotari Okay, I will make additional comments

zjxxzjwang:
@lhotari I have submitted the test cases, please review them

zjxxzjwang:
@lhotari  Hello, please review it

hanmz (approved):
LGTM

Jason918 (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SharedConsumerAssignor.java
```diff
@@ -58,7 +62,11 @@ public Map<Consumer, List<EntryAndMetadata>> assign(final List<EntryAndMetadata>
 
         Consumer consumer = getConsumer(numConsumers);
         if (consumer == null) {
-            entryAndMetadataList.forEach(EntryAndMetadata::release);
+            if (subscription != null) {
+                log.info("No consumer found to assign in topic:{}, subscription:{}, redelivering {} messages.",
+                        subscription.getTopic().getName(), subscription.getName(), entryAndMetadataList.size());
+            }
```
lhotari, line 68: this logging could be very noisy. I think it's better to use debug level logging instead since it's a normal condition that there might not be any consumers available with permits.

lhotari (commented):
Good catch @zjxxzjwang! LGTM. Just a comment about the log level.

PR #25089, opened 2025-12-17
[fix][ml] Fix cursor backlog size to account for individual acks

codelipenghui:
Fixes the issue where cursor.getEstimatedSizeSinceMarkDeletePosition() would not subtract the size of individually acknowledged messages.

Problem:
- When messages are acknowledged out of order using asyncDelete(), the backlog count correctly decreases
- But the backlog size remains inflated because individualDeletedMessages are not accounted for

Example:
- 5 messages (500 bytes total)
- Individual ack of 3 messages
- Before fix: backlog size = 500 (incorrect)
- After fix: backlog size = 200 (correct)

Fix:
- Update ManagedCursorImpl.getEstimatedSizeSinceMarkDeletePosition() to calculate and subtract size of individual deleted entries
- Uses average entry size to estimate deleted size
- Thread-safe with proper locking

Unit test:
- testEstimatedSizeWithIndividualAcks() verifies the fix

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -1321,7 +1321,8 @@ public class ManagedCursorImpl implements ManagedCursor {
         }
 
         // Estimate size by using average entry size from the backlog range
-        long totalEntriesInBacklog = ledger.getNumberOfEntries(Range.openClosed(markDeletePosition, ledger.getLastPosition()));
+        Range<Position> backlogRange = Range.openClosed(markDeletePosition, ledger.getLastPosition());
+        long totalEntriesInBacklog = ledger.getNumberOfEntries(backlogRange);
 
         if (totalEntriesInBacklog <= deletedCount || totalEntriesInBacklog == 0) {
             // Should not happen, but avoid division by zero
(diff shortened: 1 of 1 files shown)


Technoboy- (approved):

poorbarcode (approved):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1282,7 +1282,69 @@ public int getNonContiguousDeletedMessagesRangeSerializedSize() {
...
+                        long countInRange = ledger.getNumberOfEntries(intersection);
+                        deletedCounter.addAndGet(countInRange);
+                    }
+                    return true;
+                }, recyclePositionRangeConverter);
+                deletedCount = deletedCounter.get();
+            }
+        } finally {
+            lock.readLock().unlock();
+        }
```
lhotari, line 1317: This part seems to duplicate logic that currently exists in `getNumberOfEntries` method. Avoiding code duplication would be preferred. Would it be possible to extract a common method that is used in both `getEstimatedSizeSinceMarkDeletePosition` and `getNumberOfEntries`?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1282,7 +1282,69 @@ public int getNonContiguousDeletedMessagesRangeSerializedSize() {
...
+
+        if (deletedCount == 0) {
+            return totalSize;
+        }
+
+        // Estimate size by using average entry size from the backlog range
+        Range<Position> backlogRange = Range.openClosed(markDeletePosition, ledger.getLastPosition());
+        long totalEntriesInBacklog = ledger.getNumberOfEntries(backlogRange);
```
lhotari: The duplicated `deletedCount` code above can be avoided by using the `ManagedCursorImpl#getNumberOfEntries` method instead of `ManagedLedgerImpl#getNumberOfEntries`

```suggestion
        long totalEntriesInBacklog = this.getNumberOfEntries(backlogRange); 
```

Issue #25082, opened 2025-12-16
[Enhancement] Add counter for marker messages to "pulsar-admin topics analyze-backlog"


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Pulsar Admin contains a command for analysing the topic backlog. However, the counters don't return the amount of marker messages. It would be useful to distinguish marker messages from ordinary messages. 

### Solution

* add the field to org.apache.pulsar.common.stats.AnalyzeSubscriptionBacklogResult
* add the counter logic to org.apache.pulsar.broker.service.persistent.PersistentSubscription#analyzeBacklog

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
Another related enhancement: #25083

oneby-wang:
Hi, @lhotari, I would like to contribute if this enhancement is not assigned.

lhotari:
> Hi, [@lhotari](https://github.com/lhotari), I would like to contribute if this enhancement is not assigned.

@oneby-wang Thanks for volunteering! I've assigned also #25083 to you since it is closely related.

oneby-wang:
Got.

PR #25091, opened 2025-12-18
[improve][admin] Add counter for marker messages in PersistentTopics.analyzeSubscriptionBacklog() rest api

oneby-wang:
Fixes https://github.com/apache/pulsar/issues/25082

### Motivation

`PersistentTopics.analyzeSubscriptionBacklog()` rest api returns the counters of backlog, but the counters don't return the amount of marker messages.

### Modifications

Add counter for marker messages in `PersistentTopics.analyzeSubscriptionBacklog()` rest api, add tests.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [x] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/15

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/v3/AdminApiTransactionTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/v3/AdminApiTransactionTest.java
@@ -1102,6 +1102,14 @@ public class AdminApiTransactionTest extends MockedPulsarServiceBaseTest {
                 admin.topics().analyzeSubscriptionBacklog(topic, transactionSubName, Optional.of(abortedMiddleMsgId));
         assertEquals(backlogResult.getMessages(), numMessages / 2);
         assertEquals(backlogResult.getMarkerMessages(), numMessages / 2);
+
+        Transaction txn = pulsarClient.newTransaction().build().get();
+        for (int i = 0; i < numMessages; i++) {
+            producer.newMessage(txn).value("uncommitted-msg-" + i).send();
+        }
+        backlogResult = admin.topics().analyzeSubscriptionBacklog(topic, transactionSubName, Optional.empty());
+        assertEquals(backlogResult.getMessages(), numMessages * 3);
+        assertEquals(backlogResult.getMarkerMessages(), numMessages * 2);
     }
 
     private static void verifyCoordinatorStats(String state,


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java
```diff
@@ -693,6 +694,9 @@ public CompletableFuture<AnalyzeBacklogResult> analyzeBacklog(Optional<Position>
             }
             long num = entries.incrementAndGet();
             messages.addAndGet(numMessages);
+            if (messageMetadata.hasMarkerType()) {
+                markerMessages.addAndGet(numMessages);
+            }
```
lhotari: this condition should be moved before `int numMessages = 1` line. In the case that the entry is a marker message, the filter shouldn't be evaluated and the counters shouldn't be incremented.
This is to be consistent with the behavior of the dispatcher which also skips further dispatching logic when the entry is a marker message.

https://github.com/apache/pulsar/blob/3fb52c5d7494d4020956b7a26db5e9e721c7d555/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java#L210-L229

I forgot to mention this detail in the original issue,  #25082.

Besides being consistent with the dispatcher logic, it will make the usage of analyzing the backlog easier since there wouldn't be a need to subtract the number of marker messages from the total message count and "filter accepted entries" count.

oneby-wang: > the filter shouldn't be evaluated and the counters shouldn't be incremented

When doing this improvement, I also have some doubts about marker messages and entry filters. Thanks for clearing this up for me.

So, if a message is a marker message, we just increment `markerMessages` and leave out the rest, right?

lhotari: > So, if a message is a marker message, We just increment `markerMessages` and leave out the rest, right?

exactly

oneby-wang:
Since uncommitted  or aborted messages will not be delivered to consumer, counting them into `numMessages` or`markerMessages` seems all improper. 

The aborted messages is filtered by `AbstractBaseDispatcher`. 

https://github.com/apache/pulsar/blob/449552534c1e1191056e957bc31081760560ff9a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractBaseDispatcher.java#L199-L206

The uncommitted messages will not be read and delivered before committed.  

https://github.com/apache/pulsar/blob/449552534c1e1191056e957bc31081760560ff9a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java#L4789-L4791

Maybe we should modify `OpScan` scan params and scan logic and filter aborted messages in `analyzeBacklog` method, so the backlog would be consistent with consumer backlog.

https://github.com/apache/pulsar/blob/449552534c1e1191056e957bc31081760560ff9a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java#L129-L132

@lhotari WDYT?

lhotari:
> Since uncommitted or aborted messages will not be delivered to consumer, counting them into `numMessages` or`markerMessages` seems all improper.

I agree that it would be useful to have a separate counter for such transactional messages that are part of aborted transactions. For the transaction markers themselves, it's useful to have them part of marker messages.
I think that the aborted transactional message counter could be handled in a separate PR.

lhotari:
Removed `release/4.0.9` label since the test `AdminApiTransactionTest#testAnalyzeSubscriptionBacklogWithTransactionMarker` fails in `branch-4.0` after cherry-picking this PR.

reproducing:
```
mvn -DredirectTestOutputToFile=false -DtestRetryCount=0 test -pl pulsar-broker -Dtest=AdminApiTransactionTest#testAnalyzeSubscriptionBacklogWithTransactionMarker
```
(compile the project with `mvn -Pcore-modules,-main -T 1C clean install -DskipTests -Dspotbugs.skip=true -DnarPluginPhase=none` before this)

lhotari:
> Removed `release/4.0.9` label since the test `AdminApiTransactionTest#testAnalyzeSubscriptionBacklogWithTransactionMarker` fails in `branch-4.0` after cherry-picking this PR.

I guess #25125 might fix that issue. Checking.

lhotari:
#25125 solved the issue with the test.

Issue #25092, opened 2025-12-18
Flaky-test: ManagedCursorTest.asyncMarkDeleteBlocking


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/lhotari/pulsar/actions/runs/20333509746/job/58415012379?pr=190#step:11:2444

### Exception stacktrace

```text
Error:  Tests run: 473, Failures: 1, Errors: 0, Skipped: 472, Time elapsed: 2.717 s <<< FAILURE! -- in org.apache.bookkeeper.mledger.impl.ManagedCursorTest
  Error:  org.apache.bookkeeper.mledger.impl.ManagedCursorTest.asyncMarkDeleteBlocking -- Time elapsed: 0.354 s <<< FAILURE!
  java.lang.AssertionError: expected [20:9] but found [23:-1]
  	at org.testng.Assert.fail(Assert.java:110)
  	at org.testng.Assert.failNotEquals(Assert.java:1577)
  	at org.testng.Assert.assertEqualsImpl(Assert.java:149)
  	at org.testng.Assert.assertEquals(Assert.java:131)
  	at org.testng.Assert.assertEquals(Assert.java:643)
  	at org.apache.bookkeeper.mledger.impl.ManagedCursorTest.asyncMarkDeleteBlocking(ManagedCursorTest.java:1397)
  	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
  	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
  	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
  	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
  	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
  	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:11)
  	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
  	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
  	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
[...]

PR #25087, opened 2025-12-17
[fix][broker] Fix cursor position persistence in ledger trimming

codelipenghui:
### Motivation

The issue was in `maybeUpdateCursorBeforeTrimmingConsumedLedger` where it called `onCursorMarkDeletePositionUpdated`, which only updates the in-memory cursor position without persisting it to the metadata store. If the broker crashes after this update, it will recover the old mark deletion position, which leads to confusion as the trimmed ledgers may still be referenced by the cursor.

Here is a real case in production:

```
bin/pulsar-admin topics stats-internal persistent://xxx-partition-1
{
  "entriesAddedCounter" : 19,
  "numberOfEntries" : 659,
  "totalSize" : 39413,
  "currentLedgerEntries" : 19,
  "currentLedgerSize" : 1140,
  "lastLedgerCreatedTimestamp" : "2025-12-16T20:03:38.788Z",
  "waitingCursorsCount" : 1,
  "pendingAddEntriesCount" : 0,
  "lastConfirmedEntry" : "5704709:18",
  "state" : "LedgerOpened",
  "ledgers" : [ {
    "ledgerId" : 5364980,
    "entries" : 1,
    "size" : 59,
    "offloaded" : false,
    "underReplicated" : false
  }, {
    "ledgerId" : 5689698,
    "entries" : 291,
    "size" : 17334,
    "offloaded" : false,
    "underReplicated" : false
  }, {
    "ledgerId" : 5692065,
    "entries" : 348,
    "size" : 20880,
    "offloaded" : false,
    "underReplicated" : false
  }, {
    "ledgerId" : 5704709,
    "entries" : 0,
    "size" : 0,
    "offloaded" : false,
    "underReplicated" : false
  } ],
  "cursors" : {
    "pulsar.repl.xxx" : {
      "markDeletePosition" : "5704709:18",
      "readPosition" : "5704709:19",

      "waitingReadOp" : true,
      "pendingReadOps" : 0,
      "messagesConsumedCounter" : 19,
      "cursorLedger" : 5708255,
      "cursorLedgerLastEntry" : 19,
      "individuallyDeletedMessages" : "[]",
      "lastLedgerSwitchTimestamp" : "2025-12-16T20:03:38.976Z",
      "state" : "Open",
      "active" : true,
      "numberOfEntriesSinceFirstNotAckedMessage" : 1,
      "totalNonContiguousDeletedMessagesRange" : 0,
      "subscriptionHavePendingRead" : false,
[...]

PR #25101, opened 2025-12-21
[fix][test] Fix ManagedCursorTest and NonDurableCursorTest flaky tests

oneby-wang:
Fixes https://github.com/apache/pulsar/issues/25092

### Motivation

`ManagedCursorTest.asyncMarkDeleteBlocking()` test is very flaky after PR https://github.com/apache/pulsar/pull/25087, so fix it.

### Modifications

1. When doing local debug, I see many logs like this:

```
org.apache.bookkeeper.mledger.ManagedLedgerException: org.apache.bookkeeper.mledger.impl.ManagedCursorImpl$MarkDeletingMarkedPosition: Mark deleting an already mark-deleted position. Current mark-delete: 13:9 -- attempted mark delete: 13:8
Caused by: org.apache.bookkeeper.mledger.impl.ManagedCursorImpl$MarkDeletingMarkedPosition: Mark deleting an already mark-deleted position. Current mark-delete: 13:9 -- attempted mark delete: 13:8
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.setAcknowledgedPosition(ManagedCursorImpl.java:2076)
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.asyncMarkDelete(ManagedCursorImpl.java:2212)
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger(ManagedLedgerImpl.java:2736)
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl$14.operationComplete(ManagedLedgerImpl.java:1768)
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl$14.operationComplete(ManagedLedgerImpl.java:1747)
	at org.apache.bookkeeper.mledger.impl.MetaStoreImpl.lambda$asyncUpdateLedgerIds$7(MetaStoreImpl.java:187)
```

So I modify `if (!lastAckedPosition.equals(cursor.getMarkDeletedPosition()))` to `if (lastAckedPosition.compareTo(cursor.getMarkDeletedPosition()) > 0)` in `ManagedLedgerImpl`.

2. Modify some typos, and add some todos(could be deleted if not needed) in `ManagedLedgerImpl`.

3. Add simplest test case method: `asyncMarkDeleteBlockingWithOneShot()`, see method comments. I move `c1.asyncMarkDelete` out of `asyncAddEntry#addComplete` callback method, because the two operations will compete for the same ledgerId due to race condition, which is the reason why `asyncMarkDeleteBlocking()` test is more flaky after https://github.com/apache/pulsar/pull/25087. Some logs if I put `c1.asyncMarkDelete`  in `asyncAddEntry#addComplete` callback method.

```
2025-12-19T21:58:48,996 - INFO  - [main:ManagedCursorTest@1405] - size: 100, positions: [3:0, 3:1, 3:2, 3:3, 3:4, 3:5, 3:6, 3:7, 3:8, 3:9, 5:0, 5:1, 5:2, 5:3, 5:4, 5:5, 5:6, 5:7, 5:8, 5:9, 7:0, 7:1, 7:2, 7:3, 7:4, 7:5, 7:6, 7:7, 7:8, 7:9, 9:0, 9:1, 9:2, 9:3, 9:4, 9:5, 9:6, 9:7, 9:8, 9:9, 11:0, 11:1, 11:2, 11:3, 11:4, 11:5, 11:6, 11:7, 11:8, 11:9, 13:0, 13:1, 13:2, 13:3, 13:4, 13:5, 13:6, 13:7, 13:8, 13:9, 14:0, 14:1, 14:2, 14:3, 14:4, 14:5, 14:6, 14:7, 14:8, 14:9, 16:0, 16:1, 16:2, 16:3, 16:4, 16:5, 16:6, 16:7, 16:8, 16:9, 18:0, 18:1, 18:2, 18:3, 18:4, 18:5, 18:6, 18:7, 18:8, 18:9, 19:0, 19:1, 19:2, 19:3, 19:4, 19:5, 19:6, 19:7, 19:8, 19:9]

2025-12-19T21:58:48,625 - INFO  - [main:ManagedCursorTest@1405] - size: 100, positions: [3:0, 3:1, 3:2, 3:3, 3:4, 3:5, 3:6, 3:7, 3:8, 3:9, 5:0, 5:1, 5:2, 5:3, 5:4, 5:5, 5:6, 5:7, 5:8, 5:9, 7:0, 7:1, 7:2, 7:3, 7:4, 7:5, 7:6, 7:7, 7:8, 7:9, 8:0, 8:1, 8:2, 8:3, 8:4, 8:5, 8:6, 8:7, 8:8, 8:9, 10:0, 10:1, 10:2, 10:3, 10:4, 10:5, 10:6, 10:7, 10:8, 10:9, 11:0, 11:1, 11:2, 11:3, 11:4, 11:5, 11:6, 11:7, 11:8, 11:9, 13:0, 13:1, 13:2, 13:3, 13:4, 13:5, 13:6, 13:7, 13:8, 13:9, 15:0, 15:1, 15:2, 15:3, 15:4, 15:5, 15:6, 15:7, 15:8, 15:9, 17:0, 17:1, 17:2, 17:3, 17:4, 17:5, 17:6, 17:7, 17:8, 17:9, 19:0, 19:1, 19:2, 19:3, 19:4, 19:5, 19:6, 19:7, 19:8, 19:9]
```

4.  Add `asyncMarkDeleteBlockingWithMultiShots()` test to reproduce race condition between `c1.asyncMarkDelete` and `asyncAddEntry#addComplete`, see method comments. 

6. Fix `asyncMarkDeleteBlocking()` flaky test, see method comments.
    a. Fix `MetadataStoreException$BadVersionException` due to race condition between ledger rollover and ledger recovery read.
    b. Fix assert failure due to race condition race condition between `c1.asyncMarkDelete` and `asyncAddEntry#addComplete`

7. And more tests, see PR discussion.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/16

### DIFF
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
@@ -1553,7 +1553,6 @@ public class ManagedCursorTest extends MockedBookKeeperTestCase {
         // flaky test case: c2.getMarkDeletedPosition() may be equals lastPositionLedgerId+1 or lastPositionLedgerId+2,
         // the last c1.asyncMarkDelete() operation may trigger a cursor ledger rollover
         // See PR https://github.com/apache/pulsar/pull/25087.
-        log.info("c2 markDeletePosition: {}, lastPosition: {}", c2.getMarkDeletedPosition(), lastPosition);
         Awaitility.await()
                 .untilAsserted(() -> assertThat(c2.getMarkDeletedPosition()).isGreaterThan(lastPosition.get()));
     }


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2713,26 +2713,29 @@ public void maybeUpdateCursorBeforeTrimmingConsumedLedger() {
...
             } else {
+                // TODO no ledger exists, should we move cursor mark deleted position to nextPointedLedger:-1 ?
                 log.warn("Cursor: {} does not exist in the managed-ledger.", cursor);
             }
 
-            if (!lastAckedPosition.equals(cursor.getMarkDeletedPosition())) {
+            if (lastAckedPosition.compareTo(cursor.getMarkDeletedPosition()) > 0) {
                 Position finalPosition = lastAckedPosition;
                 log.info("Reset cursor:{} to {} since ledger consumed completely", cursor, lastAckedPosition);
+                // TODO If PR https://github.com/apache/pulsar/pull/25087 is needed, maybe we should make
```
oneby-wang: Call stack trace: `OpaddEntry.closeComplete()` --> `ManagedLedgerImpl.ledgerClosed()`, which doesn't even get a chance to receive the callback.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2713,26 +2713,29 @@ public void maybeUpdateCursorBeforeTrimmingConsumedLedger() {
...
                     if (lastAckedPosition.getEntryId() != -1
-                            && lastAckedPosition.getEntryId() + 1 >= currPointedLedger.getEntries()) {
+                            && lastAckedPosition.getEntryId() + 1 >= curPointedLedger.getEntries()) {
                         lastAckedPosition = PositionFactory.create(nextPointedLedger.getLedgerId(), -1);
                     }
                 } else {
                     log.debug("No need to reset cursor: {}, current ledger is the last ledger.", cursor);
                 }
             } else {
+                // TODO no ledger exists, should we move cursor mark deleted position to nextPointedLedger:-1 ?
```
oneby-wang: I think we should move `markDeletePosition` to `nextPointedLedger:-1` here. If `ManagedLedger` has only one empty ledger, and `markDeletePosition` is `ledgerId:-1`, recover read will create a new ledger `ledgerId+1:-1`. If we don't move the cursor to `ledgerId+1:-1`, the `markDeletePosition` is still confusion.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1260,26 +1263,29 @@ void cursorPersistence() throws Exception {
...
+        ledger = factory2.open("cursor_persistence_ledger");
         c1 = ledger.openCursor("c1");
         c2 = ledger.openCursor("c2");
 
         assertEquals(c1.getMarkDeletedPosition(), p1);
-        // move mark-delete-position from 3:5 to 6:-1 since all the entries have been consumed
         ManagedCursor finalC2 = c2;
-        Awaitility.await().untilAsserted(() -> assertNotEquals(finalC2.getMarkDeletedPosition(), p2));
+        Awaitility.await()
+                .untilAsserted(() -> assertEquals(finalC2.getMarkDeletedPosition(), new ImmutablePositionImpl(6, -1)));
```
lhotari: `ImmutablePositionImpl` shouldn't be referenced in code directly. For creating instances, `PositionFactory.create` is preferred.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2713,26 +2713,29 @@ public void maybeUpdateCursorBeforeTrimmingConsumedLedger() {
...
                 log.warn("Cursor: {} does not exist in the managed-ledger.", cursor);
             }
 
-            if (!lastAckedPosition.equals(cursor.getMarkDeletedPosition())) {
+            if (lastAckedPosition.compareTo(cursor.getMarkDeletedPosition()) > 0) {
```
lhotari: Since there's a change to production code, is this PR more than a test fix?

lhotari: I'm just wondering when would this condition be fulfilled. `lastAckedPosition` is `lastAckedPosition = cursor.getPersistentMarkDeletedPosition() != null
                    ? cursor.getPersistentMarkDeletedPosition() : cursor.getMarkDeletedPosition();` .

I haven't tried it yet, but to me, it seems that the logic in #25087 might not be correct. 

I'd suppose that maybeUpdateCursorBeforeTrimmingConsumedLedger would have to be made asynchronous (return CompletableFuture<Void>) and the locations where it is being called should be evaluated.

oneby-wang: Mark deleting an already mark-deleted position will throw an Exception, and seems having no more side effects, so I modify it. I would like to hear your suggestion whether to change this PR title or move the production code change to another PR?

oneby-wang: > I'm just wondering when would this condition be fulfilled.

Race condition: if ledger rolls over with `cursor.asyncMarkDelete()` executing, and another thead(threadA) may execute `cursor.asyncMarkDelete()` at the same time.

The two `cursor.asyncMarkDelete()` operations are executed without order guarantee. Broken case:
1. rollover thread gets the old `markDeletePosition` 3:-1, then calculates `lastAckedPosition` as 3:-1.
2. threadA executes `cursor.asyncMarkDelete()` success, and updates `markDeletePosition` to 13:-1.
3. came back to rollover thread, we will see `lastAckedPosition < markDeletePosition`.

Maybe we should snapshot `cursor.getMarkDeletePosition()` into a local variable, instead of calling it every time to avoid this problem.

oneby-wang: > I'd suppose that maybeUpdateCursorBeforeTrimmingConsumedLedger would have to be made asynchronous (return CompletableFuture) and the locations where it is being called should be evaluated.

Seems not so easy to deal with these two cases, they can't bring CompletableFuture or Callback to their caller callbacks.

https://github.com/apache/pulsar/blob/3fb52c5d7494d4020956b7a26db5e9e721c7d555/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L1747-L1769

https://github.com/apache/pulsar/blob/3fb52c5d7494d4020956b7a26db5e9e721c7d555/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ShadowManagedLedgerImpl.java#L328-L346

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
oneby-wang: Done.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
lhotari: Yes, I think that production code changes should be in a separate PR. After reading your comments about the race condition, I now see that the tests themselves are wrong.

it seems that test assertions like this
```
assertEquals(c2.getMarkDeletedPosition(), lastPosition.get());
``` 

should be something like `assertThat(c2.getMarkDeletedPosition()).isGreaterThanOrEqualTo(lastPosition.get());`

I don't think that tests should contain ledgerIds since there's another race that could happen between the manager cursor storing the state to a ledger.

oneby-wang: > I haven't tried it yet, but to me, it seems that the logic in https://github.com/apache/pulsar/pull/25087 might not be correct.

Yeah, so I think we should decide whether to fix https://github.com/apache/pulsar/pull/25087 or rollback https://github.com/apache/pulsar/pull/25087. 

1. If we decide to fix https://github.com/apache/pulsar/pull/25087, then we should discuss how to fix it and it's releated flaky tests.
2. If we decide to rollback https://github.com/apache/pulsar/pull/25087, then we should modify test cases in this PR.

lhotari: The remaining `org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException` issue might be a production code bug. There's a race condition that should be prevented or handled by retrying.

lhotari: > Seems not so easy to deal with these two cases, they can't bring CompletableFuture or Callback to their caller callbacks.

That's true, however for the 3rd use case, ManagedLedgerFactoryImpl#asyncOpen, it might make a difference.

oneby-wang: > I don't think that tests should contain ledgerIds since there's another race that could happen between the manager cursor storing the state to a ledger.

I ran a lot of tests these days, and kind of figured out the logic of rollover and mark delete. The race condition cases in this PR can be exhaustively enumerated, so I think `assertEquals()` is more precise that `isGreaterThanOrEqualTo()`.

oneby-wang: > That's true, however for the 3rd use case, ManagedLedgerFactoryImpl#asyncOpen, it might make a difference.

Yes, I originally intended to submit a PR, but I don't know how to deal with these two special cases.

oneby-wang: > The remaining org.apache.bookkeeper.mledger.ManagedLedgerException$BadVersionException issue might be a production code bug. There's a race condition that should be prevented or handled by retrying.

Yes, topic could be fenced due to this race condition, which would cause client failure and retry.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1260,26 +1263,29 @@ void cursorPersistence() throws Exception {
...
+        ledger = factory2.open("cursor_persistence_ledger");
         c1 = ledger.openCursor("c1");
         c2 = ledger.openCursor("c2");
 
         assertEquals(c1.getMarkDeletedPosition(), p1);
-        // move mark-delete-position from 3:5 to 6:-1 since all the entries have been consumed
         ManagedCursor finalC2 = c2;
-        Awaitility.await().untilAsserted(() -> assertNotEquals(finalC2.getMarkDeletedPosition(), p2));
+        Awaitility.await()
+                .untilAsserted(() -> assertEquals(finalC2.getMarkDeletedPosition(), PositionFactory.create(6, -1)));
```
lhotari: I'd avoid using specific ledger ids since there's also the managed cursor persistent that could increase the ledger id and cause race conditions.
Wouldn't it be sufficient to assert that the mark delete position is greater than p2 here?

```suggestion
                .untilAsserted(() -> assertThat(finalC2.getMarkDeletedPosition()).isGreaterThan(p2));
```

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1391,182 @@ public void markDeleteComplete(Object ctx) {
...
+        // See PR https://github.com/apache/pulsar/pull/25087.
+        log.info("c2 markDeletePosition: {}, lastPosition: {}", c2.getMarkDeletedPosition(), lastPosition);
+        long lastPositionLedgerId = lastPosition.get().getLedgerId();
+        Awaitility.await().untilAsserted(() ->
+                assertTrue(
+                        c2.getMarkDeletedPosition().equals(lastPosition.get())
+                        || c2.getMarkDeletedPosition().equals(PositionFactory.create(lastPositionLedgerId + 2, -1))
+                        || c2.getMarkDeletedPosition().equals(PositionFactory.create(lastPositionLedgerId + 3, -1))
+                )
+        );
```
lhotari: It's better to simplify this assertion and just use  `assertThat(c2.getMarkDeletedPosition()).isGreaterThanOrEqualTo(lastPosition.get());`

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1391,182 @@ public void markDeleteComplete(Object ctx) {
...
+                @Override
+                public void addComplete(Position position, ByteBuf entryData, Object ctx) {
+                    lastPosition.set(position);
+                    positions.offer(position);
+                    addEntryLatch.countDown();
+                }
+            }, null);
+        }
+        addEntryLatch.await();
+        assertEquals(lastPosition.get(), PositionFactory.create(13, 0));
```
lhotari: Let's avoid specific ledger ids in the test. If there's a need to calculate the number of ledgers, that could be done by iterating the persisted entry positions and filtering the unique ledger ids.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1391,182 @@ public void markDeleteComplete(Object ctx) {
...
+        // the nextPointedLedger is 15, ledger 13 only has 1 consumed entry 13:0,
+        // so we will move markDeletePosition to 15:-1, see PR https://github.com/apache/pulsar/pull/25087.
+        ledger = factory2.open("async_mark_delete_blocking_test_ledger");
+        ManagedCursor c2 = ledger.openCursor("c1");
+
+        log.info("positions size: {}, positions: {}", positions.size(), positions);
+        // To make sure ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger() is completed, we should
+        // wait until c2.getMarkDeletedPosition() equals 15:-1, see PR https://github.com/apache/pulsar/pull/25087.
+        Awaitility.await()
+                .untilAsserted(() -> assertEquals(c2.getMarkDeletedPosition(), PositionFactory.create(15, -1)));
```
lhotari: remove assertions of specific positions. This applies to the specific ledger id value. I guess it would be fine to check that the ledger id value has changed. 

btw. The ManagerCursor interface is supposed to be generic and rollover should be an implementation detail, however these tests are very specific to the ManagerLedgerImpl/ManagedCursorImpl within Pulsar.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1391,182 @@ public void markDeleteComplete(Object ctx) {
...
+        ManagedCursor c2 = ledger.openCursor("c1");
+
+        // flaky test case: c2.getMarkDeletedPosition() may be equals lastPositionLedgerId+1 or lastPositionLedgerId+2,
+        // the last c1.asyncMarkDelete() operation may trigger a cursor ledger rollover
+        // See PR https://github.com/apache/pulsar/pull/25087.
+        log.info("c2 markDeletePosition: {}, lastPosition: {}", c2.getMarkDeletedPosition(), lastPosition);
+        long lastPositionLedgerId = lastPosition.get().getLedgerId();
+        Awaitility.await().untilAsserted(() -> assertTrue(
+                c2.getMarkDeletedPosition().equals(PositionFactory.create(lastPositionLedgerId + 1, -1))
+                        || c2.getMarkDeletedPosition().equals(PositionFactory.create(lastPositionLedgerId + 2, -1))));
```
lhotari: Could we simplify the assertion?

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
oneby-wang: I'm wondering why we should avoid specific ledger ids in the test. 

Seems we could konw the expected result or could exhaustively enumerate race condition cases. 

I think it would be more precise to do these asserts, so we could find flaky tests earlier if someone breaks this rule.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
oneby-wang: Got. I'll simplify these assertions.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
lhotari: > I'm wondering why we should avoid specific ledger ids in the test.

Unit tests shouldn't be tied to implementation details too closely. This is a good practice in unit testing, for example [mentioned here](https://martinfowler.com/articles/practical-test-pyramid.html#WhatToTest). We do have a lot of tests in Pulsar that test that the implementation doesn't change. That type of tests aren't optimal and don't necessarily provide value.

In addition, in this case, the Manager Cursor state is also stored in ledgers and could cause inconsistent results if specific ledger ids are asserted.

oneby-wang: > Unit tests should be tied to implementation details too closely. 

Got. Thanks for clearing up. Seems typo? Unit tests shouldn't be tied to implementation details too closely. 

I'll modify all these assetions.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
lhotari: > Yes, I originally intended to submit a PR, but I don't know how to deal with these two special cases.

In those other cases, it might not be necessary to make changes. In the asyncOpen case, the race condition might be causing [the BadVersionException](https://github.com/apache/pulsar/issues/25092#issuecomment-3681796991) which might be a real bug.

oneby-wang: > In the asyncOpen case, the race condition might be causing https://github.com/apache/pulsar/issues/25092#issuecomment-3681796991 which might be a real bug.

Same opinion. I'll move the production code change and try to fix `BadVersionException` in a seperate PR. 

BTW, I think this else should also be handled, see https://github.com/apache/pulsar/pull/25101#discussion_r2638752804.

> In those other cases, it might not be necessary to make changes.

So, if we don't handle those other cases, cursor position might be `lastActiveLedgerId:entryNum-1` or `lastActiveLedgerId+1:-1` if we run `cursor,asyncMarkDelete()` after `asyncAddEntry` completed, which seems not harmful to broker.

Should we ensure that the cursor is in a deterministic state, or allow it to remain in this indeterminate state?

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1390,173 @@ public void markDeleteComplete(Object ctx) {
...
+        // Three cases:
+        // 1. cursor recovered with lastPosition markDeletePosition
+        // 2. cursor recovered with (lastPositionLedgerId+1:-1) markDeletePosition, cursor ledger not rolled over, we
+        //    move markDeletePosition to (lastPositionLegderId+2:-1)
+        // 3. cursor recovered with (lastPositionLedgerId+1:-1) markDeletePosition, cursor ledger rolled over, we
+        //    move markDeletePosition to (lastPositionLegderId+3:-1)
+        // See PR https://github.com/apache/pulsar/pull/25087.
+        log.info("c2 markDeletePosition: {}, lastPosition: {}", c2.getMarkDeletedPosition(), lastPosition);
+        Awaitility.await()
+                .untilAsserted(() -> assertTrue(c2.getMarkDeletedPosition().compareTo(lastPosition.get()) >= 0));
```
lhotari: AssertJ is preferred over the usage of `assertTrue` in cases where it hides the asserted values. The benefit of AssertJ is that in the case of failures, it will show the asserted values.
For example, the usage of AssertJ in this case:

`import static org.assertj.core.api.Assertions.assertThat;`
```suggestion
                .untilAsserted(() -> assertThat(c2.getMarkDeletedPosition()).isGreaterThanOrEqualTo(lastPosition.get()));
```

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1390,173 @@ public void markDeleteComplete(Object ctx) {
...
+        // the nextPointedLedger is 15, ledger 13 only has 1 consumed entry 13:0,
+        // so we will move markDeletePosition to 15:-1, see PR https://github.com/apache/pulsar/pull/25087.
+        ledger = factory2.open("async_mark_delete_blocking_test_ledger");
+        ManagedCursor c2 = ledger.openCursor("c1");
+
+        log.info("positions size: {}, positions: {}", positions.size(), positions);
+        // To make sure ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger() is completed, we should
+        // wait until c2.getMarkDeletedPosition() equals 15:-1, see PR https://github.com/apache/pulsar/pull/25087.
+        Awaitility.await()
+                .untilAsserted(() -> assertTrue(c2.getMarkDeletedPosition().compareTo(lastPosition.get()) > 0));
```
lhotari:
```suggestion
                .untilAsserted(() -> assertThat(c2.getMarkDeletedPosition()).isGreaterThan(lastPosition.get()));
```

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1385,16 +1390,173 @@ public void markDeleteComplete(Object ctx) {
...
+        @Cleanup("shutdown") ManagedLedgerFactory factory2 = new ManagedLedgerFactoryImpl(metadataStore, bkc);
+        ledger = factory2.open("async_mark_delete_blocking_test_ledger");
+        ManagedCursor c2 = ledger.openCursor("c1");
+
+        // flaky test case: c2.getMarkDeletedPosition() may be equals lastPositionLedgerId+1 or lastPositionLedgerId+2,
+        // the last c1.asyncMarkDelete() operation may trigger a cursor ledger rollover
+        // See PR https://github.com/apache/pulsar/pull/25087.
+        log.info("c2 markDeletePosition: {}, lastPosition: {}", c2.getMarkDeletedPosition(), lastPosition);
+        Awaitility.await()
+                .untilAsserted(() -> assertTrue(c2.getMarkDeletedPosition().compareTo(lastPosition.get()) > 0));
```
lhotari:
```suggestion
                .untilAsserted(() -> assertThat(c2.getMarkDeletedPosition()).isGreaterThan(lastPosition.get()));
```

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -1260,26 +1263,28 @@ void cursorPersistence() throws Exception {
...
+        // See PR https://github.com/apache/pulsar/pull/25087.
+        ledger = factory2.open("cursor_persistence_ledger");
         c1 = ledger.openCursor("c1");
         c2 = ledger.openCursor("c2");
 
         assertEquals(c1.getMarkDeletedPosition(), p1);
-        // move mark-delete-position from 3:5 to 6:-1 since all the entries have been consumed
         ManagedCursor finalC2 = c2;
-        Awaitility.await().untilAsserted(() -> assertNotEquals(finalC2.getMarkDeletedPosition(), p2));
+        Awaitility.await().untilAsserted(() -> assertTrue(finalC2.getMarkDeletedPosition().compareTo(p2) > 0));
```
lhotari:
```suggestion
        Awaitility.await()untilAsserted(() -> assertThat(finalC2.getMarkDeletedPosition()).isGreaterThan(p2));
```

lhotari (commented):
Check the comments about preferring AssertJ

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
oneby-wang: Thanks for input, now I know the reason.

lhotari:
@oneby-wang There seems to be a few more locations where the same flakiness pattern occurs. please check at least `cursorPersistenceAsyncMarkDeleteSameThread`. `testLazyCursorLedgerCreationForSubscriptionCreation` appeared flaky too and it would be worth checking whether it's the same problem.
Please also check `NonDurableCursorTest.asyncMarkDeleteBlocking`.

lhotari:
The BadVersionExceptions seem to be another test flakiness problem in ManagedCursorTest due to async ledger rollovers. 
One possible solution is something like this:

```
    public <T> T retry(ThrowingSupplier<T> supplier) {
        for (int i = 0; i < 10; i++) {
            if (i > 0) {
                try {
                    log.info("Retrying after 100ms {}/{}", i, 10);
                    Thread.sleep(100);
                }
            }
            try {
                return supplier.get();
            } catch (Exception e) {
                log.warn("Failed to execute supplier: {}", supplier, e);
            }
        }
        throw new RuntimeException("Failed to execute supplier after 10 retries");
    }

    @FunctionalInterface
    public interface ThrowingSupplier<T> {
        T get() throws Exception;
    }
```
This could be added to `BrokerTestUtil` class.

and then using it in the test classes (for example in `testNumberOfEntriesWithReopen` method):
```
ledger = retry(() -> factory2.open("my_test_ledger", new ManagedLedgerConfig().setMaxEntriesPerLedger(1)));
```

oneby-wang:
Got the two above comments.

oneby-wang:
@lhotari Maybe we should do `asyncMarkDelete` if we reach this condition? Since this is the root casue to inconsistency.

https://github.com/apache/pulsar/blob/fbab357fdc28992b59512c355ebf06b29eef4f2e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3010-L3012

lhotari:
> @lhotari Maybe we should do `asyncMarkDelete` if we reach this condition? Since this is the root casue to inconsistency.
> 
> https://github.com/apache/pulsar/blob/fbab357fdc28992b59512c355ebf06b29eef4f2e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3010-L3012

@oneby-wang I guess that could make sense, however the problem is that there could be multiple cursors at the slowest position. `ManagedCursorContainerImpl/ManagedCursorContainer` doesn't have a sorted data structure which could be iterated. `ActiveManagedCursorContainer` has the method `getNumberOfCursorsAtSamePositionOrBefore` which solves this detail. It's currently used for `activeCursors` in `ManagedLedgerImpl` when `cacheEvictionByExpectedReadCount` is enabled. It would be possible to use it also for other use cases. The performance of `ActiveManagedCursorContainerImpl` is better than `ManagedCursorImpl` also for the ordinary use cases. There are micro benchmarks in the `microbench` module.

However, it might be better to handle it separately. Just wondering if it's already handled by the `advanceCursorsIfNecessary` method:

https://github.com/apache/pulsar/blob/fbab357fdc28992b59512c355ebf06b29eef4f2e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3109-L3116

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
lhotari: > Same opinion. I'll move the production code change and try to fix `BadVersionException` in a seperate PR.

I checked some of the BadVersionException cases and it's a valid case when opening the ledger without closing the first instance. In that case, the test code should perform the retry.

lhotari: > So, if we don't handle those other cases, cursor position might be `lastActiveLedgerId:entryNum-1` or `lastActiveLedgerId+1:-1` if we run `cursor,asyncMarkDelete()` after `asyncAddEntry` completed, which seems not harmful to broker.
> 
> Should we ensure that the cursor is in a deterministic state, or allow it to remain in this indeterminate state?

I think it's fine to use lenient checks in this case.

oneby-wang:
> Just wondering if it's already handled by the advanceCursorsIfNecessary method:

I read this code. If we mark delete 12:9, next ledger is an empty ledger: 13:-1, seems this method will only move mark delete position to 12:8. If ledger 12 is trimmed, then causing inconsistency.

I'm wondering if mark delete position with -1 entryId is a valid mark delete position? Seems some codes intend to avoid this.

https://github.com/apache/pulsar/blob/fbab357fdc28992b59512c355ebf06b29eef4f2e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3234-L3236

https://github.com/apache/pulsar/blob/fbab357fdc28992b59512c355ebf06b29eef4f2e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L622-L632

EDIT: if a cursor subscribe an empty managed ledger, then it's mark delete start with ledgerId:-1, it is a valid mark delete position.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
oneby-wang: > In that case, the test code should perform the retry.

Should we change the test to close the ledger first and then reopen it?

lhotari: > Should we change the test to close the ledger first and then reopen it?

I don't think so. I think that it's intentional in the test that the ledger is left open. That's why it would be necessary to retry for the second time it's opened so that the test wouldn't be flaky and result in BadVersionExceptions.

oneby-wang: Got. I misunderstood before, BadVersionExceptions do exist in previous code before https://github.com/apache/pulsar/pull/25087. I ran tests to verfied it.

https://github.com/apache/pulsar/pull/25087 brings us an uncertain cursor mark delete position, which I think doesn't  completely resolve cursor position and ledger inconsistency in some concurrent scenarios.

lhotari:
> There seems to be a few more locations where the same flakiness pattern occurs. please check at least `cursorPersistenceAsyncMarkDeleteSameThread`. `testLazyCursorLedgerCreationForSubscriptionCreation` appeared flaky too and it would be worth checking whether it's the same problem.
> Please also check `NonDurableCursorTest.asyncMarkDeleteBlocking`.

@oneby-wang Please address this so that we could get at least some improvements made to the flakiness. The `NonDurableCursorTest.asyncMarkDeleteBlocking` method is really flaky too. I'd like to merge this PR very soon although we haven't addressed all the details that have come up during the review.

oneby-wang:
@lhotari Working.

lhotari:
> I read this code. If we mark delete 12:9, next ledger is an empty ledger: 13:-1, seems this method will only move mark delete position to 12:8. If ledger 12 is trimmed, then causing inconsistency.
> 
> I'm wondering if mark delete position with -1 entryId is a valid mark delete position? Seems some codes intend to avoid this.

good observations. It seems that there's a lot of code locations where it's hard to determine which is the correct behavior. 
There seems to be some inconsistency in how things are handled across different parts of the ManagedLedger implementation. I guess we can focus on addressing the test flakiness.

oneby-wang:
I think maybe we can fix cursor position and ledger inconsistency by `advanceCursorsIfNecessary` method.

lhotari (approved):
LGTM, however it seems that BadVersionException could occur in all cases where the ledger is opened with a different instance without closing the previous one. I think that retry would be needed to address those cases. We could handle that in a separate PR since this PR will already improve significantly the current CI flakiness.

Thanks for the great work, @oneby-wang

oneby-wang:
I’ll revisit the BadVersionExceptions over the next few days and submit a PR to address it. Thanks for reviewing and your great suggestions. @lhotari

oneby-wang:
> however it seems that BadVersionException could occur in all cases where the ledger is opened with a different instance without closing the previous one.

I think just in rollover and reopen race condition case. If ledger is not rolling over, which will not update metastore, so reopen ledger will success to update metastore without throwing BadVersionException.

Issue #25103, opened 2025-12-22
[Bug] Non-thread safe code in ManagedCursorImpl.removeProperty


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

master

### Issue Description

The properties aren't accessed in a thread safe way in org.apache.bookkeeper.mledger.impl.ManagedCursorImpl#removeProperty

https://github.com/apache/pulsar/blob/3fb52c5d7494d4020956b7a26db5e9e721c7d555/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L521-L534

The way to modify properties should be implemented in a similar way as it is in the putProperty method.

https://github.com/apache/pulsar/blob/81aff30c4b461b2c630f413c407ba014e7a5d571/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L502-L519

The Map implementation used for the properties field of org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.MarkDeleteEntry should be immutable to avoid concurrency issues.

### Error messages

```text

```

### Reproducing the issue

This issue was found by reading the code. Since it's a concurrency issue, it would require a concurrency test to reproduce.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #25104, opened 2025-12-22
[improve][broker] Fix thread safety issue in ManagedCursorImpl.removeProperty 

3pacccccc:
Fixes https://github.com/apache/pulsar/issues/25103

### Motivation
The `ManagedCursorImpl.removeProperty` method contains a thread safety vulnerability where it directly modifies the shared `properties` Map without proper synchronization. This can lead to:

1. **ConcurrentModificationException** when `putProperty` creates a new HashMap copy while `removeProperty` modifies the original Map
2. **Data corruption** in the internal HashMap structure under concurrent access
3. **Inconsistent state** where one thread's modifications may not be visible to others

### Modifications
1. **Fixed `removeProperty` method** in `ManagedCursorImpl.java`:
   - Changed from direct Map modification (`properties.remove(key)`) to creating a new Map copy
   - Updated implementation to match the thread-safe pattern already used in `putProperty`
   - Maintains atomicity through `LAST_MARK_DELETE_ENTRY_UPDATER.updateAndGet()`

2. **Added concurrency test** in `ManagedCursorTest.java`:
   - 100-thread stress test with mixed put/remove/get operations
   - 5-second duration to simulate sustained concurrent access
   - Verifies no `ConcurrentModificationException` or data inconsistencies
   - Validates final state matches expected results

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/3pacccccc/pulsar/pull/37

### DIFF
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
@@ -20,6 +20,7 @@ package org.apache.bookkeeper.mledger.impl;
 
 import static org.apache.bookkeeper.mledger.impl.EntryCountEstimator.estimateEntryCountByBytesSize;
 import static org.apache.bookkeeper.mledger.impl.cache.RangeEntryCacheImpl.BOOKKEEPER_READ_OVERHEAD_PER_ENTRY;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.Mockito.any;
@@ -50,6 +51,7 @@ import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -62,6 +64,7 @@ import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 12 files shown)


managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -204,6 +206,81 @@ public void testOpenCursorWithNullInitialPosition() throws Exception {
...
+                    totalOperations.incrementAndGet();
+                    Random random = new Random();
+                    int operationType = random.nextInt(3);
+                    String randomKey = "key" + random.nextInt(20);
+
+                    switch (operationType) {
+                        case 0: // Put operation
+                            Long randomValue = random.nextLong();
+                            cursor.putProperty(randomKey, randomValue);
+                            records.put(randomKey, randomValue);
```
lhotari: the 2 separate operations aren't atomic together so I don't see how the behavior could be validated this way.

It's possible to have a valid data race for example between the put and remove where the `records` becomes inconsistent even though there's no problem.

t1
```
cursor.putProperty(randomKey, randomValue);
records.put(randomKey, randomValue);
```

t2
```
cursor.removeProperty(randomKey);
records.remove(randomKey);
```

For example with the order of operations:
```
t1 - cursor.putProperty(randomKey, randomValue);
t2 - cursor.removeProperty(randomKey);
t2 - records.remove(randomKey);
t1 - records.put(randomKey, randomValue);
```

3pacccccc: @lhotari Thanks for the catch! You're right about the atomicity issue. I've removed the tracking map and now the test just checks for thread safety issues directly - no exceptions and no data corruption

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java
```diff
@@ -251,17 +252,36 @@ public void testConcurrentPropertyOperationsThreadSafety() throws Exception {
                             }
                             break;
                     }
+                    completedOperations.incrementAndGet();
+
                 } catch (ConcurrentModificationException cme) {
+                    // Record ConcurrentModificationException but don't fail immediately
```
lhotari, line 262: HashMap will throw ConcurrentModificationException only in specific methods such as `forEach`, `compute`, and a few others.

I didn't run the tests without the fix to validate whether ConcurrentModificationException would be thrown. Have you seen it if you revert the fix and run the test?

I think that this catch block can be removed and just fail immediately like it's done for other Exceptions.

lhotari: I guess `new HashMap(properties)` could trigger `ConcurrentModificationException`, so it's most likely valid that it can happen.

3pacccccc: > HashMap will throw ConcurrentModificationException only in specific methods such as `forEach`, `compute`, and a few others.
> 
> I didn't run the tests without the fix to validate whether ConcurrentModificationException would be thrown. Have you seen it if you revert the fix and run the test?
> 
> I think that this catch block can be removed and just fail immediately like it's done for other Exceptions.

Yes, exactly. This `ConcurrentModificationException` does occur in the original (unfixed) version. I added the explicit catch in the test to verify that the fix resolves the issue.

PR #25105, opened 2025-12-22
[fix][broker]pulsar_ml_reads_inflight_bytes and pulsar_ml_reads_available_inflight_bytes are 0 at the same time

codelipenghui:
## Description

Fix the issue that the metric pulsar_ml_reads_inflight_bytes and pulsar_ml_reads_available_inflight_bytes are all 0 at the same time. The correct behavior should be if there is no inflight reads happened, the available inflight bytes should be the configured value from user.

## Changes
- Initialize metrics in InflightReadsLimiter constructor when enabled.
- Add testPrometheusMetrics to verify the fix.
- Change Prometheus Gauges visibility for testing.

## Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


poorbarcode (approved):

lhotari (approved):
LGTM

PR #25110, opened 2025-12-24
[fix][broker] Fix markDeletedPosition race condition in ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger() method

oneby-wang:
### Motivation

We do mark deleting operation in `ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger()` after https://github.com/apache/pulsar/pull/25087.

Mark deleting an already mark-deleted position will throw an Exception.

Please see comment for race condition case: https://github.com/apache/pulsar/pull/25101#discussion_r2639087166.

See also: 
1. https://github.com/apache/pulsar/pull/25101
2. https://github.com/apache/pulsar/pull/25101#discussion_r2639019381

### Modifications

1. Modify `ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger()` method:
  a. snapshot positions into a local variable to avoid race condition.
  b. use `markDeletePosition` instead of `persistentMarkDeletedPosition` to set `lastAckedPosition`, because `persistentMarkDeletedPosition` is updated after `markDeletePosition`.
  c. add if check logic to avoid mark deleting an already mark-deleted position.
  d. wait `maybeUpdateCursorBeforeTrimmingConsumedLedger()` completed when opening ledger.
  e. move `maybeUpdateCursorBeforeTrimmingConsumedLedger()` into synchronized block when creating a new ledger. I think new ledger's initialization work should be completed before the first entry is added to this ledger.

2. Modify `ManagedCursorImpl.getNumberOfEntries()` method: snapshot positions into a local variable to avoid race condition, error logs like this:

```
java.lang.IllegalArgumentException: Invalid range: [104:0..103:1)

    at com.google.common.collect.Range.<init>(Range.java:334)
    at com.google.common.collect.Range.create(Range.java:134)
    at com.google.common.collect.Range.closedOpen(Range.java:171)
    at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.getNumberOfEntries(ManagedCursorImpl.java:1253)
    at org.apache.bookkeeper.mledger.impl.ManagedLedgerTest.testNeverThrowsMarkDeletingMarkedPositionInMaybeUpdateCursorBeforeTrimmingConsumedLedger(ManagedLedgerTest.java:3838)
```

3. Ignore changes in `ManagedCursorImpl.asyncMarkDelete()` method, I just snapshot positions into a local variable to avoid race condition. It should be fixed by: https://github.com/apache/pulsar/pull/25117.

4. Add `testNeverThrowsMarkDeletingMarkedPositionInMaybeUpdateCursorBeforeTrimmingConsumedLedger` test in `ManagedLedgerImplTest` to verify the code change.

5. Fix tests due to this PR's code change.

6. Fix some flaky tests introduced by PR https://github.com/apache/pulsar/pull/25087.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/18

### DIFF
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
@@ -5291,9 +5291,11 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
         ManagedCursorImpl cursor = (ManagedCursorImpl) ledger.openCursor("c1");
 
         CountDownLatch latch = new CountDownLatch(1);
+        AtomicReference<Position> lastPosition = new AtomicReference<>();
         ledger.asyncAddEntry("entry-1".getBytes(Encoding), new AddEntryCallback() {
             @Override
             public void addComplete(Position position, ByteBuf entryData, Object ctx) {
+                lastPosition.set(position);
                 // Mark delete with properties
                 Map<String, Long> properties = new HashMap<>();
                 properties.put("test-property", 12345L);
@@ -5320,15 +5322,17 @@ public class ManagedLedgerTest extends MockedBookKeeperTestCase {
 
         Map<String, Long> expectedProperties = new HashMap<>();
         expectedProperties.put("test-property", 12345L);
+        assertThat(cursor.getMarkDeletedPosition()).isGreaterThanOrEqualTo(lastPosition.get());
         assertEquals(cursor.getProperties(), expectedProperties);
 
         // 3. Add Entry 2. Triggers second rollover process.
         // This implicitly calls maybeUpdateCursorBeforeTrimmingConsumedLedger due to rollover
-        ledger.addEntry(("entry-2").getBytes(Encoding));
+        Position p2 = ledger.addEntry(("entry-2").getBytes(Encoding));
 
         // Wait for background tasks (metadata callback) to complete.
         // We expect at least 2 ledgers (Rollover happened).
         Awaitility.await().atMost(5, TimeUnit.SECONDS).until(() -> ledger.getLedgersInfo().size() >= 2);
+        assertEquals(cursor.getMarkDeletedPosition(), PositionFactory.create(p2.getLedgerId(), -1));
 
         // Verify properties are preserved after cursor reset
         assertEquals(cursor.getProperties(), expectedProperties);


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1243,14 +1243,17 @@ public boolean hasMoreEntries() {
 
     @Override
     public long getNumberOfEntries() {
-        if (readPosition.compareTo(ledger.getLastPosition().getNext()) > 0) {
+        Position readPos = readPosition;
+        Position lastPosition = ledger.getLastPosition();
+        Position nextPosition = lastPosition.getNext();
```
BewareMyPower, line 1255: Copying the positions into local variables is not enough if these variables can be modified in a different thread. We should add a lock to guarantee thread safety.

Here is an example of a very simple case:

```java
class Range {
    private int start = 0;
    private int end = 0;

    public int getNumberOfEntries() {
        return end - start;
    }

    public void produce(int n) {
        CompletableFuture.runAsync(() -> {
            end += n;
        });
    }

    public void consume(int n) {
        CompletableFuture.runAsync(() -> {
            if (start + n > end) {
                throw new IllegalStateException("Cannot consume more than produced");
            }
            start += n;
        });
    }
}
```

Even if you changed `getNumberOfEntries()` to:

```java
    private volatile int start = 0;
    private volatile int end = 0;

    public int getNumberOfEntries() {
        int end = this.end;
        int start = this.start;
        return end - start;
    }
```

`getNumberOfEntries` is still not safe

oneby-wang: By default positions is created by `PositionFactory.create()`, which return an instance of `ImmutablePositionImpl`. If somewhere creates a position not using `PositionFactory.create()`, I think we should modify it.

I think this can make sure that we could catch a snapshot view if we snapshot positions into local varaiables.

I ran a lot of tests locally, and it appears that the error of `java.lang.IllegalArgumentException: Invalid range: [104:0..103:1)` no longer occurs.

BewareMyPower: Oh sorry, I misunderstood it before. I've thought `lastPosition` and `nextPosition` are computed individually, but `nextPosition` is computed based on `lastPosition`, so there should be no issue here.

oneby-wang:
Doesn't solve `ManagedLedgerTest.testTrimmerRaceCondition()` flaky test yet, I'll revisit it later. 

Seems we must ensure some happens-before order between `ManagedCursoImplr.asyncMarkDelete()` and `ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger()` after https://github.com/apache/pulsar/pull/25087.

oneby-wang:
Fix `ManagedLedgerTest.testTrimmerRaceCondition()`  flaky test. 

Add `ManagedLedgerTest.testTrimmerRaceConditionInNonDurableCursor()` and `ManagedLedgerTest.testTrimmerRaceConditionWithThrottleMarkDeleteInDurableCursor()` test.

Ready for review.

PR #25113, opened 2025-12-24
[feat][broker] PIP-451: Support Label-based Topic Subscription

coderzc:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-451.md
+++ b/pip/pip-451.md
@@ -16,186 +16,57 @@ Label-based Subscription solves these issues by decoupling "Identity" (Topic Nam
 
 ## In Scope
 * Management: Allow attaching, updating, and removing Key-Value labels to/from Topics via the Admin API.
-* Subscription: Allow Consumers to subscribe to topics matching specific Labels within specified Namespaces. Support cross-namespace subscription via an explicit namespace list in the Client API, avoiding the complexity of background metadata polling.
-
-
+* Subscription: Allow Consumers to subscribe to topics matching specific Tags within specified Namespaces. Support cross-namespace subscription via an explicit namespace list in the Client API, avoiding the complexity of background metadata polling.
 
 # High-Level Design
-The design introduces a metadata-driven approach where labels are stored in TopicPolicies. 
-The Broker maintains an in-memory index to map labels to topics. The client utilizes a "Watch" mechanism to receive real-time updates when topics matching the labels are created or updated.
-## Key points
-* Storage: Labels are stored as Map<String, String> inside TopicPolicies.
-* Indexing: The Broker maintains an In-Memory Inverted Index (LabelKey -> LabelValue -> Set<Topic>) per Namespace. This hierarchical structure ensures efficient lookups for Key-Value pairs without iterating through all topics.
-* Discovery Protocol: We extend the CommandWatchTopicList protocol (PIP-179) to accept a label_selector.
-* Client Implementation: The Client accepts a list of target namespaces and manages multiple watchers (one per namespace) to aggregate matching topics.
+
+The feature consists of three main components:
+
+* Storage:
+
+We reuse existing Topic Properties.
+
+To support Partitioned Topics, we adopt a Propagate-on-Write strategy. When properties are set on a Parent Topic, the Broker orchestrating the update will iterate and write these properties to the ManagedLedger metadata of all its sub-partitions.
+
+* Discovery (Broker Scatter-Gather):
+
+A new Binary Protocol command allows clients to query "Active Topics matching properties".
+
+The Gateway Broker (Coordinator) distributes the query to all Bundle Owners (Workers).
+
+Workers perform a fast in-memory scan of loaded topics.
+
+* Client (Polling Model):
+
+The Consumer client periodically polls the Broker using the new command.
+
+It compares the returned list with its current subscription and dynamically subscribes/unsubscribes.
 
 # Detailed Design
 
 ## Design & Implementation Details
 
-### Storage 
-
-#### Topic Labels in TopicPolicies
-
-Add a labels field to the TopicPolicies class. Since Topic Policies are propagated via the __change_events system topic, this ensures durability and consistency across brokers.
-```java
-public class TopicPolicies {
-    // New field: Key-Value labels
-    private Map<String, String> customLabels;
-}
-```
-
-#### In-Memory Inverted Index for Labels-Topic Mapping
-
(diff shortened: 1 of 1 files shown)


pip/pip-451.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### In-Memory Inverted Index for Labels-Topic Mapping
+
+The SystemTopicBasedTopicPoliciesService will maintain a nested map structure per Namespace to support efficient Key-Value lookups.
+
+**Data Structure**:
+
+```java
+// Map<Namespace, Map<LabelKey, Map<LabelValue, Set<TopicName>>>>
+Map<String, Map<String, Map<String, Set<String>>>> labelTopicInvertedIndex;
```
shibd: The `SystemTopicBasedTopicPoliciesService` constructs the cache for a specific namespace based on the assigned `namespace bundles`. 

In production scenarios, the bundles of a single namespace are highly likely to be evenly distributed across all brokers. This implies that every broker will effectively maintain a full copy of the this cache. I am concerned about the potential memory footprint in this scenario

pip/pip-451.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### TopicListWatcher State Machine
+
+Each TopicListWatcher maintains a set of currently matched topics to perform state-based updates:
+
+```java
+public void onLabelsUpdate(TopicName topicName, @Nullable Map<String, String> allLabels) {
+    // allLabels can be null if the topic is deleted
+    boolean wasMatching = matchedTopics.contains(topicName);
+    boolean matchesNow = allLabels != null && labelMatcher.matches(allLabels);
```
shibd: What does the data structure of labelMatcher look like? Could you provide an example?

pip/pip-451.md
```diff
@@ -0,0 +1,255 @@
...
+}
+```
+
+### Client Implementation Details
+The Client implementation acts as an orchestrator to support the "Cross-Namespace" requirement defined in topicsByLabel.
+
+* Multi-Watcher Orchestration:
+
+When `subscribe()` is called, the Client iterates over the provided namespaces.
+For each namespace, it initiates a CommandWatchTopicList request.
```
shibd: If can provide a exmaples for user side, that be better.

pip/pip-451.md
```diff
@@ -0,0 +1,255 @@
...
+    
+    // ... existing methods ...
+
+    /**
+     * Subscribe to topics matching the given labels within specific namespaces.
+     * * @param labels     The Key-Value pairs that a topic must match (AND semantics).
+     * @param namespaces The set of namespaces to look for topics. 
+     * If empty, defaults to the current consumer's namespace.
+     */
+    ConsumerBuilder<T> topicsByLabel(Map<String, String> labels, Set<String> namespaces);
```
shibd: This input parameter cannot represent specific labels for a specific namespace. Perhaps we need to wrap the parameter as:

```
List {
   String namespace,
   Map<String, String> labels
}
```

pip/pip-451.md
```diff
@@ -0,0 +1,255 @@
...
+**Data Structure**:
+
+```java
+// Map<Namespace, Map<LabelKey, Map<LabelValue, Set<TopicName>>>>
+Map<String, Map<String, Map<String, Set<String>>>> labelTopicInvertedIndex;
+```
+
+### Topic Labels Observation and Notification Mechanism
+
+### TopicPoliciesService Interface Changes
```
shibd:
```suggestion
#### TopicPoliciesService Interface Changes
```

pip/pip-451.md
```diff
@@ -0,0 +1,255 @@
...
+  - pulsar-admin topics remove-custom-labels <topic-name> --labels "key1,key2" (to remove specific labels)
+  - pulsar-admin topics remove-custom-labels <topic-name> --all (to remove all custom labels from the topic)
+* pulsar-admin topics remove-custom-labels <topic-name> --labels "key1,key2" (to remove specific labels)
+* pulsar-admin topics remove-custom-labels <topic-name> --all (to remove all custom labels from the topic)
+* REST API: DELETE /admin/v2/topics/{tenant}/{namespace}/{topic}/custom-labels with a query params keys=k1&keys=k2 or all=true.
+* Action: Removes the specified custom labels or all custom labels from the topic. This also updates the topic policy.
+
+#### Query topic associated with specific labels:
+
+* CLI: pulsar-admin topics list <namespace> --custome-labels "k1:v1,k2:v2"
```
shibd: I'm wondering how we should implement this feature, since not every broker maintains a full `labelTopicInvertedIndex` cache. In the REST API implementation, we would first need to determine which broker is assigned to any bundle of that namespace, and then send the request there. 

Is this lookup operation easy to achieve?

shibd (commented):
I support pushing this feature forward. It effectively simplifies the pattern subscription issue and provides a multi-dimensional approach to topic management. 

However, there are some implementation details that require further discussion.

pip/pip-451.md
coderzc: I think the topic policy under a namespace can be loaded onto any broker， since we call `prepareInitPoliciesCacheAsync(NamespaceName namespace)`.

pip/pip-451.md
coderzc: Why is it necessary to use different labels for different namespaces?

pip/pip-451.md
coderzc: Yes, each broker needs to cache the full amount of labelTopicInvertedIndex.

pip/pip-451.md
coderzc: Like this:

```
public class LabelMatcher {

    private Map<String, String> labels;

    boolean matches(Map<String, String> targetLabels) {
        for (Map.Entry<String, String> entry : labels.entrySet()) {
            String key = entry.getKey();
            String value = entry.getValue();
            if (!targetLabels.containsKey(key) || !targetLabels.get(key).equals(value)) {
                return false;
            }
        }
        return true;
    }
}
```

coderzc: This can be simplified into a tool function.

PR #25117, opened 2025-12-28
[fix][broker] Move newPosition to nextValidLedger:-1 to avoid cursor position and ledger inconsistency in ManagedCursorImpl

oneby-wang:
### Motivation

Since `nextValidLedger:-1` is a valid `markDeletePosition`, we should move `newPosition` to `nextValidLedger:-1` to avoid cursor position and ledger inconsistency in `ManagedCursorImpl` whenever possible.

PR https://github.com/apache/pulsar/pull/25087 also made this change.

### Modifications

1. Modify `ManagedCursorImpl.asyncMarkDelete()` method, move `newPosition` to `nextValidLedger:-1` if we reach the following conditions:
  a. if `lastConfirmedEntry >= newPosition`, and next ledger exists, and current ledger entries are all consumed.
  b. if `lastConfirmedEntry < newPosition`, next ledger exists, and `newPosition == nextValidLedger:-1`.

I think the previous code might have a little problem. 
  a. If `newPosition == nextValidLedger:n`, n is an non-negative number, we might set new `newPosition` to `nextValidLedger:n` which is greater than `lastConfirmedEntry`, which is the biggest problem.
  b. We can not mark delete `lac` many times if we already moved markDeletePosition to `nextValidLedger:-1`.
  c. We can not mark delete `nextValidLedger:-1` directly if markDeletePosition is less than `lac`.

https://github.com/apache/pulsar/blob/ab65faa12ab7279a726411152af44d81b6a6704b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L2184-L2195

And snapshot positions into a local variable to avoid race condition.

2. Add `testAsyncMarkDeleteMoveToNextLedgerInNonRolloverScenario`, `testAsyncMarkDeleteMayMoveToNextLedgerInRolloverScenario`, `testAsyncMarkDeleteMoveToNextLedgerOneByOne`, `testAsyncMarkDeleteNextLedgerMinusOneEntryIdPosition` tests in `ManagedCursorTest` to verify the code change.

3. Fix tests  due to this PR's code change.

4. Fix some flaky tests introduced by PR https://github.com/apache/pulsar/pull/25087.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/19

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -2257,37 +2257,51 @@ public class ManagedCursorImpl implements ManagedCursor {
             log.debug("[{}] Mark delete cursor {} up to position: {}", ledger.getName(), name, position);
         }
 
+        // Snapshot all positions into local variables to avoid race condition.
         Position newPosition = ackBatchPosition(position);
+        Position moveForwardPosition = newPosition;
         Position markDeletePos = markDeletePosition;
         Position lastConfirmedEntry = ledger.getLastConfirmedEntry();
-        if (lastConfirmedEntry.compareTo(newPosition) < 0) {
-            boolean shouldCursorMoveForward = false;
-            try {
-                long ledgerEntries = ledger.getLedgerInfo(markDeletePos.getLedgerId()).get().getEntries();
-                Long nextValidLedger = ledger.getNextValidLedger(lastConfirmedEntry.getLedgerId());
-                shouldCursorMoveForward = nextValidLedger != null
-                        && (markDeletePos.getEntryId() + 1 >= ledgerEntries)
-                        && (newPosition.getLedgerId() == nextValidLedger);
-            } catch (Exception e) {
-                log.warn("Failed to get ledger entries while setting mark-delete-position", e);
-            }
-
-            if (shouldCursorMoveForward) {
-                log.info("[{}] move mark-delete-position from {} to {} since all the entries have been consumed",
-                        ledger.getName(), markDeletePos, newPosition);
-            } else {
-                if (log.isDebugEnabled()) {
-                    log.debug("[{}] Failed mark delete due to invalid markDelete {} is ahead of last-confirmed-entry {}"
-                             + " for cursor [{}]", ledger.getName(), position, lastConfirmedEntry, name);
+        boolean shouldCursorMoveForward = false;
+        try {
+            if (lastConfirmedEntry.getLedgerId() >= newPosition.getLedgerId()) {
+                LedgerInfo curMarkDeleteLedgerInfo = ledger.getLedgerInfo(newPosition.getLedgerId()).get();
+                Long nextValidLedger = ledger.getNextValidLedger(newPosition.getLedgerId());
+                shouldCursorMoveForward = (nextValidLedger != null)
+                        && (curMarkDeleteLedgerInfo != null
+                        && newPosition.getEntryId() + 1 >= curMarkDeleteLedgerInfo.getEntries());
+                if (shouldCursorMoveForward) {
+                    moveForwardPosition = PositionFactory.create(nextValidLedger, -1);
                 }
-                callback.markDeleteFailed(new ManagedLedgerException("Invalid mark deleted position"), ctx);
-                return;
+            } else {
+                Long nextValidLedger = ledger.getNextValidLedger(lastConfirmedEntry.getLedgerId());
+                shouldCursorMoveForward = (nextValidLedger != null)
+                        && (newPosition.getLedgerId() == nextValidLedger)
+                        && (newPosition.getEntryId() == -1);
             }
+        } catch (Exception e) {
+            log.warn("Failed to get ledger entries while setting mark-delete-position", e);
+        }
+
+        if (shouldCursorMoveForward) {
+            log.info("[{}] move cursor {} mark-delete-position from {} to {} since all the entries have been consumed,"
+                    + " last-confirmed-entry: {}, attempt position: {}, ack-batch position: {}", ledger.getName(),
+                    name, markDeletePos, moveForwardPosition, lastConfirmedEntry, position, newPosition);
+        } else if (lastConfirmedEntry.compareTo(newPosition) < 0) {
+            // If newPosition>lastConfirmedEntry and current ledger entries are not fully consumed,
+            // in which case the newPosition is an invalid mark delete position.
+            if (log.isDebugEnabled()) {
+                log.debug("[{}] Failed mark delete due to invalid markDelete {} is ahead of last-confirmed-entry {}"
+                        + " for cursor [{}], attempt position: {}", ledger.getName(), newPosition, lastConfirmedEntry,
+                        name, position);
+            }
+            callback.markDeleteFailed(new ManagedLedgerException("Invalid mark deleted position"), ctx);
+            return;
         }
 
         lock.writeLock().lock();
         try {
-            newPosition = setAcknowledgedPosition(newPosition);
+            moveForwardPosition = setAcknowledgedPosition(moveForwardPosition);
         } catch (IllegalArgumentException e) {
             callback.markDeleteFailed(getManagedLedgerException(e), ctx);
             return;
@@ -2298,11 +2312,11 @@ public class ManagedCursorImpl implements ManagedCursor {
         // Apply rate limiting to mark-delete operations
         if (markDeleteLimiter != null && !markDeleteLimiter.tryAcquire()) {
             isDirty = true;
-            updateLastMarkDeleteEntryToLatest(newPosition, properties);
+            updateLastMarkDeleteEntryToLatest(moveForwardPosition, properties);
             callback.markDeleteComplete(ctx);
             return;
         }
-        internalAsyncMarkDelete(newPosition, properties, callback, ctx, null);
+        internalAsyncMarkDelete(moveForwardPosition, properties, callback, ctx, null);
     }
 
     private Position ackBatchPosition(Position position) {
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
@@ -151,6 +151,7 @@ import org.apache.pulsar.metadata.api.MetadataStoreException;
 import org.apache.pulsar.metadata.api.Stat;
 import org.apache.pulsar.metadata.api.extended.SessionEvent;
 import org.apache.pulsar.metadata.impl.FaultInjectionMetadataStore;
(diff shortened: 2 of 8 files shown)


managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
```diff
@@ -5144,20 +5158,23 @@ public void markDeleteFailed(ManagedLedgerException exception, Object ctx) {
...
 
         latch.await();
-        assertEquals(cursor.getPersistentMarkDeletedPosition(), lastPosition);
-        assertEquals(ledger.getCursors().getSlowestCursorPosition(), lastPosition);
-        assertEquals(cursor.getProperties(), properties);
+        assertThat(cursor.getPersistentMarkDeletedPosition()).isGreaterThanOrEqualTo(lastPosition);
+        assertThat(ledger.getCursors().getSlowestCursorPosition()).isGreaterThanOrEqualTo(lastPosition);
+        // cursor.asyncMarkDelete() and maybeUpdateCursorBeforeTrimmingConsumedLedger may run concurrently,
+        // so we can't assert properties equals here.
+        // assertEquals(cursor.getProperties(), properties);
```
BewareMyPower: Don't use comment to remove the code. Just remove it.

BTW, could it be replaced by `Awaitility.await()`?

BewareMyPower (commented):
>  avoid cursor position and ledger inconsistency in ManagedCursorImpl whenever possible.

Could you share which test can show the inconsistency? I see some tests are added to `ManagedCursorTest`, which only verify the new behavior of this PR. But it's still confusing to me that what's wrong with the current behavior.

oneby-wang:
This PR is inspired by https://github.com/apache/pulsar/pull/25087.

If a cursor consumes the last entry of previous ledger, I think we should move the `markDeletePosition` to `nextValidLedger:-1`. If we don't move the `markDeletePosition`, once the previous ledger is trimmed, cursor's `markDeletePosition` is inconsistent with ledger data.

> Could you share which test can show the inconsistency?

`ManagedCursorTest.testAsyncMarkDeleteMoveToNextLedgerInNonRolloverScenario()` and `ManagedCursorTest.testAsyncMarkDeleteMoveToNextLedgerOneByOne()`.

But, unfortunately, we can not ensure happens-before between `ManagedCursoImplr.asyncMarkDelete()` and `ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger()` in ledger rollover, so this is the only inconsistency reason after this PR(could be handled by ledger trim process), see also `ManagedCursorTest.testAsyncMarkDeleteMayMoveToNextLedgerInRolloverScenario`

> But it's still confusing to me that what's wrong with the current behavior.

I think current behavior is not wrong, they all worked well besides the following code(see PR description).

https://github.com/apache/pulsar/blob/ab65faa12ab7279a726411152af44d81b6a6704b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L2184-L2195

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
oneby-wang: This test is introduced by https://github.com/apache/pulsar/pull/25087, and I think there may be some bugs in this PR.

I comment it here to avoid flaky test, this test could be fixed by https://github.com/apache/pulsar/pull/25110.

BewareMyPower:
> once the previous ledger is trimmed, cursor's markDeletePosition is inconsistent with ledger data.

Could there be any real issue caused by this inconsistency? It would be helpful if you can share a test to show the impacts. For example, generally computing `(x, y+1)` as the next position of `(x, y)` works in most cases even if there is no entry located in `(x, y+1)` because the `compareTo` implementation guarantees `(x, y) < (x, y+1) < (x+1, 0)`.

I think it's a bit different from #25087. From what I understand, #25087 mainly fixes the bug that `onCursorMarkDeletePositionUpdated` is called directly without calling `asyncMarkDelete`.

oneby-wang:
I think at least we should snapshot immutable positions into local variables to avoid race condition, or phenomena and logs will not match. The volatile positions may be updated by other threads when retrieving it again or logging.

oneby-wang:
> Could there be any real issue caused by this inconsistency?

In continuous consuming, this should rarely happens.

I was intened to do this thing: make sure `markDeletePosition` is moved to `nextValidLedger:-1` once the new ledger is created. But as mentioned before, we can not ensure happens-before between `ManagedCursoImplr.asyncMarkDelete()` and `ManagedLedgerImpl.maybeUpdateCursorBeforeTrimmingConsumedLedger()`.

Now, I intend to consolidate all the logic for moving the `markDeletePosition` to `nextValidLedger:-1` into the `asyncMarkDelete()` method. Both `maybeUpdateCursorBeforeTrimmingConsumedLedger()` and the ledger trim process can reuse this part of the logic.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
oneby-wang: > BTW, could it be replaced by Awaitility.await()?

Seems can't, the race condition in mark delete and ledger rollover cause this test flaky, and may impact the properties user set in `asyncDelete` process.

BewareMyPower: Okay, so let's get #25110 merged before this PR first

oneby-wang: https://github.com/apache/pulsar/pull/25110 also has race condition mentioned by https://github.com/apache/pulsar/pull/25117#issuecomment-3723599827.

oneby-wang:
After this PR, if we want to move `markDeletePosition` to `nextValidLedger:-1`, we just do `cursor.asyncMarkDelete(cursor.getMarkDeletePosition())`. In non-move-forward case, mark deleting current mark-deleted position is safe, see also:

https://github.com/apache/pulsar/blob/d72dc04549b29d5ca73bf742aa81efa74ea107c4/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedCursorTest.java#L2069-L2080

oneby-wang:
Wrong behaviours in previous code that could cause test failures: `ManagedCursorTest.testMarkDeletePositionNeverExceedLacOrNextLedgerMinusOneEntryIdPos()` and `ManagedCursorTest.testMarkDeletePreviousLacManyTimesInRolloverScenario()`.

oneby-wang:
Another inconsistent test case is: `ManagedLedgerTest.testTrimmerRaceConditionInNonDurableCursor()`.

If we don't move `markDeletedPosition` to `nextValidLedger:-1`, non-durable cursor's `markDeletedPosition` will be inconsistent with ledger data. See assert: `assertEquals(cursor.getMarkDeletedPosition(), PositionFactory.create(emptyLedgerId, -1))`. 

Without this PR, the assert should be `assertEquals(cursor.getMarkDeletedPosition(), PositionFactory.create(p, 0))`.

oneby-wang:
Squashed commits and rebased master, ready for review.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -2257,37 +2257,51 @@ public void asyncMarkDelete(final Position position, Map<String, Long> propertie
...
+        try {
+            if (lastConfirmedEntry.getLedgerId() >= newPosition.getLedgerId()) {
+                LedgerInfo curMarkDeleteLedgerInfo = ledger.getLedgerInfo(newPosition.getLedgerId()).get();
+                Long nextValidLedger = ledger.getNextValidLedger(newPosition.getLedgerId());
+                shouldCursorMoveForward = (nextValidLedger != null)
+                        && (curMarkDeleteLedgerInfo != null
+                        && newPosition.getEntryId() + 1 >= curMarkDeleteLedgerInfo.getEntries());
+                if (shouldCursorMoveForward) {
+                    moveForwardPosition = PositionFactory.create(nextValidLedger, -1);
+                }
```
lhotari, line 2275: This won't behave in the expected way in all cases. the `getNextValidLedger` will return the given ledger id if it's already the last ledger.
Therefore `moveForwardPosition = PositionFactory.create(nextValidLedger, -1);` would result in the wrong behavior. 

btw. For a `LedgerInfo` instance, it's currently possible to detect the rollover by checking the `timestamp` field. It's `!= 0` for rolled over ledgers.

lhotari: Another detail of batch acknowledgements can be seen in `ackBatchPosition` method. If the last entry isn't fully acknowledged, the acked position would be the previous position. I'm not sure if this impacts this logic.

lhotari (changes requested):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
oneby-wang: > the `getNextValidLedger` will return the given ledger id if it's already the last ledger.

the `getNextValidLedger` method calls `ledgers.ceilingKey(ledgerId + 1)`, if `ledgerId + 1` doesn't exists in this `NavigableMap`, the method will return null, indicating the current ledger is the last ledger.

https://github.com/apache/pulsar/blob/6d415c6dbfb8bf9f2f8a5b823927cf422a6bd675/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L4195-L4197

> btw. For a LedgerInfo instance, it's currently possible to detect the rollover by checking the timestamp field. It's != 0 for rolled over ledgers.

I think `nextValidLedger != null` is ok to detect the rollover, the `timestamp` field may be modified by mistake.

> Another detail of batch acknowledgements can be seen in ackBatchPosition method. If the last entry isn't fully acknowledged, the acked position would be the previous position. I'm not sure if this impacts this logic.

Similar logic in the `ManagedLedgerImpl#maybeUpdateCursorBeforeTrimmingConsumedLedger()` method, I'm wondering if this method also affects the `ackBatchPosition` method.

https://github.com/apache/pulsar/blob/6d415c6dbfb8bf9f2f8a5b823927cf422a6bd675/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2722-L2740

PR #25119, opened 2025-12-29
[fix][broker] Fix compaction horizon might be reset to an old position when phase two is interrupted

### LINKED ISSUE #25118: [Bug] Compaction subscription could be reset back to a very early position
### Search before reporting

### Read release policy

### User environment

It happens on Pulsar 4.0.5, but this bug also exists on master branch.

### Issue Description

There is a topic (`persistent://public/__kafka/__consumer_offsets-partition-3`) whose namespace retention policy has 3 days time limit and no size limit.

It has 200+ ledgers. One day, we noticed the mark-delete position of the `__compaction` cursor was reset to a very old position that is not included in all current ledgers:

```json
  "numberOfEntries" : 21782357,
  "lastConfirmedEntry" : "2306104:0",
  "state" : "LedgerOpened",
  "ledgers" : [ {
    "ledgerId" : 2276367,
    "entries" : 50000,
    "size" : 17739504,
    "offloaded" : false,
    "underReplicated" : false
  }, {
```

```json
    "__compaction" : {
      "markDeletePosition" : "1877:52",
      "readPosition" : "2281596:35302",
```

where ledger 2281596 is the 75th ledger of all 200+ ledgers.

### Error messages

We noticed the cursor's mark-delete position was reset to an old position during compaction phase two when the broker was closing.

```text
2025-12-09T07:17:14,466+0000 [BookKeeperClientWorker-OrderedExecutor-2-0] INFO  org.apache.pulsar.compaction.AbstractTwoPhaseCompactor - Commencing phase two of compaction for persistent://public/__kafka/__consumer_offsets-partition-3, from 1877:53:3:-1 to 2304038:5078:3:-1, compacting 71 keys to ledger 2304479
2025-12-09T07:17:14,484+0000 [BookKeeperClientWorker-OrderedExecutor-1-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedCursorImpl - [public/__kafka/persistent/__consumer_offsets-partition-3] reset readPosition to 1877:53 (ackSet is null) before current read readPosition 2304038:5079 on cursor __compaction
2025-12-09T07:17:14,485+0000 [pulsar-io-9-3] INFO  org.apache.pulsar.client.impl.ConsumerImpl - [persistent://public/__kafka/__consumer_offsets-partition-3][__compaction] Successfully reset subscription to the message 1877:53:3:-1
2025-12-09T07:17:18,045+0000 [pulsar-service-shutdown] INFO  org.apache.pulsar.broker.service.BrokerService - [persistent://public/__kafka/__consumer_offsets-partition-3] Unloading topic
2025-12-09T07:17:18,047+0000 [pulsar-service-shutdown] INFO  org.apache.pulsar.broker.service.persistent.PersistentSubscription - [persistent://public/__kafka/__consumer_offsets-partition-3][__compaction] Successfully closed subscription [ManagedCursorImpl{ledger=public/__kafka/persistent/__consumer_offsets-partition-3, name=__compaction, ackPos=1877:52, readPos=269017:14425}]
2025-12-09T07:17:18,049+0000 [pulsar-service-shutdown] INFO  org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl - [public/__kafka/persistent/__consumer_offsets-partition-3] Closing managed ledger
025-12-09T07:17:18,198+0000 [broker-client-shared-internal-executor-11-1] WARN  org.apache.pulsar.broker.service.persistent.PersistentTopic - [persistent://public/__kafka/__con
sumer_offsets-partition-3] Compaction failure.
java.util.concurrent.CompletionException: java.util.concurrent.CancellationException
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(Unknown Source) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source) ~[?:?]
        at java.base/java.util.concurrent.CompletableFuture.cancel(Unknown Source) ~[?:?]
        at org.apache.pulsar.client.impl.RawReaderImpl$RawConsumerImpl.lambda$failPendingRawReceives$1(RawReaderImpl.java:213)
```

### Reproducing the issue

I will include the reproduce code in a PR.

### Additional information

The root cause is that during compaction phase two, the reader will seek to the 1st compacted entry's position to read the whole topic.

If topic was closing, then the pending `readAsync` future would fail with `CancellationException` by `CompletableFuture#cancel`. Then the cursor will update the mark-delete position with the 1st compacted entry's position, which is incorrect.

For example, given the following original ledgers:
- 0: 0 to 9
- 1: 0 to 9
- 2: 0

and compacted entries: (0, 5), (1,9), horizon is 1:9, (2,0) is not compacted.

Then in phase two, the mark-delete position will be set to (0, 5). Next time the phase one will read the original ledgers from (0, 5) to (2, 0).

BewareMyPower:
Fixes #25118

### Motivation

The steps of compaction phase two are:
1. Seek the compacted reader to the position of the 1st retained entry
2. Read to latest
3. Acknowledge the last compacted message id (horizon) as well as the compacted ledger's id in properties

However, if broker is closing during phase two, the pending `readNextAsync` call could fail with `CancellationException`. In this case, when the managed ledger is closed, the position of the 1st retained entry will be persisted as the `__compaction` cursor's mark-delete position.

As a result, during the next compaction, the horizon will be very early, and phase one will read too many entries from original ledgers. This issue could get into a bad state when the original ledgers are not deleted, typically due to improper retention policy or unexpected large backlog from another durable subscription.

### Modifications

Specially for `__compaction` cursor, do not modify the mark-delete position in `internalResetCursor` (triggered by client seek API).

Add `testPhaseTwoInterruption` to reproduce this issue by injecting topic close in compaction phase two. Additionally, speed up the whole `CompactionTest`.

It should be noted that the original behavior also makes `testCompactorReadsCompacted` incorrect. This test sends two messages to two ledgers and compacted them. Then it sends the third message and trigger the compaction. However, it assumes the 2nd ledger will be opened. Technically, the 2nd ledger has only 1 entry that has been compacted, so it should not be opened. The entries from first two ledgers should be read directly from the compacted ledger. This test is fixed by adding another more message before creating the 3rd ledger.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/compaction/StrategicCompactionTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/compaction/StrategicCompactionTest.java
@@ -20,18 +20,23 @@ package org.apache.pulsar.compaction;
 
 import static org.apache.pulsar.broker.loadbalance.extensions.channel.ServiceUnitStateTableViewImpl.MSG_COMPRESSION_TYPE;
 import static org.testng.Assert.assertEquals;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Random;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import org.apache.bookkeeper.client.BookKeeper;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.pulsar.broker.auth.MockedPulsarServiceBaseTest;
 import org.apache.pulsar.client.api.Consumer;
-import org.apache.pulsar.client.api.CryptoKeyReader;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageIdAdv;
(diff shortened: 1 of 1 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -1593,7 +1598,6 @@ public void operationFailed(ManagedLedgerException exception) {
 
         persistentMarkDeletePosition = null;
         inProgressMarkDeletePersistPosition = null;
-        lastMarkDeleteEntry = new MarkDeleteEntry(newMarkDeletePosition, getProperties(), null, null);
```
lhotari, line 1596: is there a reason for removing this for all cursors?

BewareMyPower: `lastMarkDeleteEntry` will be updated after `internalAsyncMarkDelete` is done in `alignAcknowledgeStatusAfterPersisted`:
https://github.com/apache/pulsar/blob/ff0d0eb2ba97d96c8bc760a1dfe8255efa71d5c1/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L1534

Modifying it before the start of `internalAsyncMarkDelete` is duplicated and does not make sense.

lhotari (approved):
LGTM

PR #25120, opened 2025-12-30
[fix][broker]Fixed an issue where the entire subscription would be blocked when a chunk message with an ID of zero did not exist.

zjxxzjwang:
### Motivation

1、A message is split into chunk0, chunk1, and chunk2. When chunk0 is unexpectedly lost, the current code logic will endlessly deliver chunk1 and chunk2 to redeliveryMessages, causing the entire subscription to become blocked.

2、When a chunk message corresponds to a producer whose availablePermits remain below zero for an extended period, that chunk message will also be endlessly delivered to the “redeliveryMessages” queue, similarly causing the entire subscription to become blocked.

### Modifications

1、In the event of chunk0 being unexpectedly lost, chunk1 and chunk2 messages are directly acknowledged and logged without blocking the entire subscription.

2、When sending each chunk of the same message, the availablePermits of the consumer corresponding to that chunk message are no longer checked, ensuring normal delivery.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
@zjxxzjwang please resolve the merge conflict

zjxxzjwang:
> @zjxxzjwang please resolve the merge conflict

ok

zjxxzjwang:
@lhotari Hello, please review this PR. It fixes a critical bug related to the existence of chunk messages.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SharedConsumerAssignor.java
```diff
@@ -106,6 +105,64 @@ public Map<Consumer, List<EntryAndMetadata>> assign(final List<EntryAndMetadata>
...
+    private boolean isChunkedMessage(MessageMetadata metadata) {
+        return metadata != null && metadata.hasUuid() && metadata.hasChunkId() && metadata.hasNumChunksFromMsg();
+    }
+
+    private void addEntry(Map<Consumer, List<EntryAndMetadata>> consumerToEntries, Consumer consumer,
+            EntryAndMetadata entry) {
+        consumerToEntries.computeIfAbsent(consumer, __ -> new ArrayList<>()).add(entry);
+    }
+
+    private boolean skipChunk(EntryAndMetadata entryAndMetadata, MessageMetadata metadata) {
```
lhotari, line 147: `skipChunk` sounds like a command, it's better to rename it to make it sound like a query.

```suggestion
    private boolean shouldSkipChunk(EntryAndMetadata entryAndMetadata, MessageMetadata metadata) {
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SharedConsumerAssignor.java
```diff
@@ -106,6 +105,64 @@ public Map<Consumer, List<EntryAndMetadata>> assign(final List<EntryAndMetadata>
...
+                        subscription.getTopicName(), subscription.getName(), entryAndMetadata.getPosition(),
+                        metadata.getUuid(), metadata.getChunkId(), metadata.getNumChunksFromMsg());
+                // Directly ack the message.
+                if (!(subscription instanceof PulsarCompactorSubscription)) {
+                    subscription.acknowledgeMessage(Collections.singletonList(
+                            entryAndMetadata.getPosition()), AckType.Individual, Collections.emptyMap());
+                }
+            }
+            entryAndMetadata.release();
+            return true;
```
lhotari, line 161: Since there's a potential for data loss, acknowledging the messages should be active only if `autoSkipNonRecoverableData` is set in `broker.conf`. The log message should be logged with `ERROR` level when `autoSkipNonRecoverableData` isn't set and the message shouldn't get acknowledged. I think it's fine to skip the message in that case so that processing the subscription continues, but there will be a backlog left behind due to the unacked messages.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SharedConsumerAssignor.java
```diff
@@ -106,6 +105,64 @@ public Map<Consumer, List<EntryAndMetadata>> assign(final List<EntryAndMetadata>
...
+            Map<Consumer, List<EntryAndMetadata>> consumerToEntries, int availablePermits) {
+        final String uuid = metadata.getUuid();
+        Consumer consumerForUuid = uuidToConsumer.get(uuid);
+        if (consumerForUuid == null) {
+            if (skipChunk(entryAndMetadata, metadata)) {
+                return availablePermits;
+            }
+            consumerForUuid = consumer;
+            uuidToConsumer.put(uuid, consumerForUuid);
+        }
```
lhotari, line 118: It seems that this solution would only skip the first entry of possibly multiple chunk entries. 
Let's say if entry with chunkId 0 got lost and there would be subsequent entries chunkId 1, chunkId 2 and chunkId 3. The entries with chunkId 2 and chunkId 3 would get delivered to the client, causing a similar issue.

lhotari (changes requested):
Please check the review comments. I think that this behavior should be active only when `autoSkipNonRecoverableData` is enabled in `broker.conf`. That would be a safe option for avoiding unintentional data loss.

PR #16377, opened 2022-07-04
[refactor][java-client] Reduce code duplication in admin client

andrasbeni:
###Motivation

The admin client uses GET requests frequently. Most of these code places share
a simple InvocationCallback that completes a CompletableFuture with the response.
By moving all this into a common place, the code will be cleaner, and GET requests
will look more like other requests.

###Modifications

Move the functionality implemented in InvocationCallbacks to a common place.
Hide creation of CompletableFuture to be consistent with other HTTP method
implementations (e.g. asyncPutRequest).

- [ ] Make sure that the change passes the CI checks.

This change is a trivial rework without any test coverage.

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API: (no)
  - The schema: (no)
  - The default values of configurations: (no)
  - The wire protocol: (no)
  - The rest endpoints: (no)
  - The admin cli options: (no)
  - Anything that affects deployment: (no)

Need to update docs?

- [X] `doc-not-needed`
This is pure refactoring, no behavior is changed.

PR #25124, opened 2026-01-05
[fix][admin] Fix asyncGetRequest to handle 204

nodece:
### Motivation

This pull request improves the handling and testing of asynchronous GET requests in the Pulsar admin client. The main changes ensure that both HTTP 200 (OK) and 204 (No Content) responses are treated as successful, and introduce comprehensive tests to verify correct behavior for various HTTP response scenarios.

image

### Modifications

**Enhancements to async GET request handling:**

* Modified `BaseResource.java` so that both HTTP 200 (OK) and 204 (No Content) responses are accepted as successful. For 204 responses, the future is completed with `null` since there is no response body to process.

**Testing improvements:**

* Added a new test class `AsyncGetRequestTest.java` to thoroughly test the async GET request logic. The tests cover successful responses with and without a body (200 and 204), as well as error responses like 404 and 500, and 200 with an empty body.

**Testability and code maintenance:**

* Exposed the `root` WebTarget in `PulsarAdminImpl` via a new `@VisibleForTesting` method to facilitate direct testing of request paths.
* Added the necessary import for `@VisibleForTesting` to `PulsarAdminImpl.java`.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-client-admin/src/test/java/org/apache/pulsar/client/admin/internal/AsyncGetRequestTest.java
+++ b/pulsar-client-admin/src/test/java/org/apache/pulsar/client/admin/internal/AsyncGetRequestTest.java
@@ -72,7 +72,7 @@ public class AsyncGetRequestTest {
         CompletableFuture<Object> getRequest =
                 namespaces.asyncGetRequest(admin.getRoot().path("/200-with-body"), Object.class);
 
-        assertThat(getRequest).succeedsWithin(3, TimeUnit. SECONDS);
+        assertThat(getRequest).succeedsWithin(3, TimeUnit.SECONDS);
 
         // Verify the request was made
         server.verify(getRequestedFor(urlEqualTo("/200-with-body")));
@@ -95,7 +95,7 @@ public class AsyncGetRequestTest {
         assertThat(getRequest).succeedsWithin(3, TimeUnit.SECONDS);
 
         // Verify the request was made
-        server. verify(getRequestedFor(urlEqualTo("/204")));
+        server.verify(getRequestedFor(urlEqualTo("/204")));
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


nodece:
@lhotari I noticed that `sink`, `source`, `function`, and `worker` all depend on this method. #16377 fixed the behavior for `namespace`, `topic`, and related resources, but this method is still used by other admin resources. Therefore, fixing it here is important to ensure consistent behavior across all dependent components.

`org.apache.pulsar.client.admin.internal.BaseResource.FutureCallback` should be deprecated. 

`org.apache.pulsar.client.admin.internal.BaseResource#asyncGetRequest(javax.ws.rs.client.WebTarget, java.lang.Class<? extends T>)` and `org.apache.pulsar.client.admin.internal.BaseResource#asyncGetRequest(javax.ws.rs.client.WebTarget, javax.ws.rs.core.GenericType<T>)` can handle most cases.

lhotari:
> @lhotari I noticed that `sink`, `source`, `function`, and `worker` all depend on this method. #16377 fixed the behavior for `namespace`, `topic`, and related resources, but this method is still used by other admin resources. Therefore, fixing it here is important to ensure consistent behavior across all dependent components.
> 
> `org.apache.pulsar.client.admin.internal.BaseResource.FutureCallback` should be deprecated.
> 
> `org.apache.pulsar.client.admin.internal.BaseResource#asyncGetRequest(javax.ws.rs.client.WebTarget, java.lang.Class<? extends T>)` and `org.apache.pulsar.client.admin.internal.BaseResource#asyncGetRequest(javax.ws.rs.client.WebTarget, javax.ws.rs.core.GenericType<T>)` can handle most cases.

@nodece Just wondering if something is currently broken due to the lack of properly handling 204 return code.

nodece:
I believe most use cases still work today. In my environment, I use the pulsar-admin API to get the replicator dispatch rate of a namespace. This is on a private Pulsar distribution, not upstream Apache Pulsar.

That said, Apache Pulsar itself still relies on this behavior in several admin APIs, so proper handling of HTTP 204 responses is required. This PR ensures consistent and correct behavior in those cases.

PR #25126, opened 2026-01-06
[improve][cli] Add client side looping in "pulsar-admin topics analyze-backlog" cli to avoid potential HTTP call timeout

### LINKED ISSUE #25537: [Enhancement] Add client side looping to "pulsar-admin topics analyze-backlog" cli command
### Search before reporting

### Motivation

Remaining work of https://github.com/apache/pulsar/issues/25083.

Use client-side looping instead of increasing broker settings to avoid potential HTTP call timeout in "pulsar-admin topics analyze-backlog" cli.

### Solution

PR link: https://github.com/apache/pulsar/pull/25126.

### Alternatives

_No response_

### Anything else?

_No response_

oneby-wang:
Fixes https://github.com/apache/pulsar/issues/25537

### Motivation

Use client-side looping instead of increasing broker settings to avoid potential HTTP call timeout in "pulsar-admin topics analyze-backlog" cli.

### Modifications

Add client-side looping, add test.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [x] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/21

### DIFF
--- a/pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
+++ b/pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
@@ -2190,62 +2190,76 @@ public class PulsarAdminToolTest {
                 getAnalyzeSubscriptionBacklogResult(10, true, PositionFactory.create(10, 0),
                         PositionFactory.create(10, 9));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId)).thenReturn(mockResult);
+
         cmdTopics.run(split("analyze-backlog " + topic + " -s " + subscriptionName));
+
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId);
         verifyNoMoreInteractions(mockTopics);
 
         // Test with the messagePosition parameter.
         reset(mockTopics);
+
         int partitionIndex = TopicName.get(topic).getPartitionIndex();
         startMessageId = Optional.of(new MessageIdImpl(11, 10, partitionIndex));
         String messagePosition = "11:10";
         mockResult = getAnalyzeSubscriptionBacklogResult(10, true, PositionFactory.create(11, 10),
                 PositionFactory.create(11, 19));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId)).thenReturn(mockResult);
+
         cmdTopics.run(split("analyze-backlog " + topic + " -s " + subscriptionName + " -p " + messagePosition));
+
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId);
         verifyNoMoreInteractions(mockTopics);
 
-        // Test client side loop: server returns false aborted flag.
+        // Test client side loop: the loop termination condition is that server returns false aborted flag.
         reset(mockTopics);
+
         long backlogScanMaxEntries = 30;
         startMessageId = Optional.empty();
         AnalyzeSubscriptionBacklogResult firstResult =
                 getAnalyzeSubscriptionBacklogResult(10, true, PositionFactory.create(12, 0),
                         PositionFactory.create(12, 9));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId)).thenReturn(firstResult);
+
         Optional<MessageId> secondInvocationMsgId = Optional.of(new MessageIdImpl(12, 10, partitionIndex));
         AnalyzeSubscriptionBacklogResult abortedResult =
                 getAnalyzeSubscriptionBacklogResult(10, false, PositionFactory.create(12, 10),
                         PositionFactory.create(12, 19));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, secondInvocationMsgId)).thenReturn(
                 abortedResult);
+
         cmdTopics.run(split("analyze-backlog " + topic + " -s " + subscriptionName + " -b " + backlogScanMaxEntries));
+
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId);
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, secondInvocationMsgId);
         verifyNoMoreInteractions(mockTopics);
 
-        // Test client side loop: total entries exceeds backlogScanMaxEntries.
+        // Test client side loop: the loop termination condition is that total entries exceeds backlogScanMaxEntries.
         reset(mockTopics);
+
         backlogScanMaxEntries = 25;
         startMessageId = Optional.empty();
         firstResult = getAnalyzeSubscriptionBacklogResult(10, true, PositionFactory.create(13, 0),
                 PositionFactory.create(13, 9));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId)).thenReturn(firstResult);
+
         secondInvocationMsgId = Optional.of(new MessageIdImpl(13, 10, partitionIndex));
         AnalyzeSubscriptionBacklogResult moreResult =
                 getAnalyzeSubscriptionBacklogResult(10, true, PositionFactory.create(13, 10),
                         PositionFactory.create(13, 19));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, secondInvocationMsgId)).thenReturn(
                 moreResult);
+
         Optional<MessageId> thirdInvocationMsgId = Optional.of(new MessageIdImpl(13, 20, partitionIndex));
         AnalyzeSubscriptionBacklogResult lastResult =
                 getAnalyzeSubscriptionBacklogResult(10, true, PositionFactory.create(14, 0),
                         PositionFactory.create(14, 9));
         when(mockTopics.analyzeSubscriptionBacklog(topic, subscriptionName, thirdInvocationMsgId)).thenReturn(
                 lastResult);
+
         cmdTopics.run(
                 split("analyze-backlog " + topic + " -s " + subscriptionName + " -b " + backlogScanMaxEntries + " -q"));
+
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, startMessageId);
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, secondInvocationMsgId);
         verify(mockTopics, times(1)).analyzeSubscriptionBacklog(topic, subscriptionName, thirdInvocationMsgId);
--- a/pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
+++ b/pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
@@ -29,7 +29,6 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
@@ -53,14 +52,11 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import lombok.Cleanup;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.bookkeeper.mledger.Position;
-import org.apache.bookkeeper.mledger.PositionFactory;
 import org.apache.pulsar.admin.cli.extensions.CustomCommandFactory;
 import org.apache.pulsar.admin.cli.utils.SchemaExtractor;
 import org.apache.pulsar.client.admin.Bookies;
@@ -93,7 +89,6 @@ import org.apache.pulsar.client.impl.MessageIdImpl;
 import org.apache.pulsar.client.impl.MessageImpl;
 import org.apache.pulsar.client.impl.auth.AuthenticationTls;
 import org.apache.pulsar.client.impl.conf.ClientConfigurationData;
-import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.policies.data.AuthAction;
 import org.apache.pulsar.common.policies.data.AutoSubscriptionCreationOverride;
 import org.apache.pulsar.common.policies.data.AutoTopicCreationOverride;
@@ -124,7 +119,6 @@ import org.apache.pulsar.common.policies.data.TenantInfoImpl;
 import org.apache.pulsar.common.policies.data.TopicStats;
 import org.apache.pulsar.common.policies.data.TopicType;
 import org.apache.pulsar.common.protocol.schema.PostSchemaPayload;
-import org.apache.pulsar.common.stats.AnalyzeSubscriptionBacklogResult;
 import org.apache.pulsar.common.util.ObjectMapperFactory;
 import org.mockito.ArgumentMatcher;
 import org.mockito.Mockito;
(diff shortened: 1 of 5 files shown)


pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -2999,20 +3000,86 @@ private class AnalyzeBacklog extends CliCommand {
...
+                    mergedResult.setFilterAcceptedMessages(
+                            mergedResult.getFilterAcceptedMessages() + currentResult.getFilterAcceptedMessages());
+                    mergedResult.setFilterRescheduledMessages(
+                            mergedResult.getFilterRescheduledMessages() + currentResult.getFilterRescheduledMessages());
+
+                    mergedResult.setAborted(currentResult.isAborted());
+                    mergedResult.setLastMessageId(currentResult.getLastMessageId());
+                }
+
+                if (!mergedResult.isAborted() || mergedResult.getEntries() >= backlogScanMaxEntries) {
```
lhotari: the `!mergedResult.isAborted()` part isn't correct.

https://github.com/apache/pulsar/blob/da0d11644f097ae657e79b2eb7835478d79ccd0e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java#L119-L128

The purpose of the CLI loop is to keep on iterating until the result is COMPLETED or the number of scanned entries exceeds `backlogScanMaxEntries`.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -2999,20 +3000,86 @@ private class AnalyzeBacklog extends CliCommand {
...
+
+                if (!quiet) {
+                    print("Analyze backlog progress, scanned entries: " + mergedResult.getEntries()
+                            + ", scan max entries: " + backlogScanMaxEntries);
+                }
+
+                String[] messageIdSplits = mergedResult.getLastMessageId().split(":");
+                MessageIdImpl nextScanMessageId =
+                        new MessageIdImpl(Long.parseLong(messageIdSplits[0]), Long.parseLong(messageIdSplits[1]) + 1,
+                                partitionIndex);
```
lhotari: Instead of parsing the message Id String, it would be better to use `org.apache.pulsar.client.api.MessageIdAdv` interface to get the ledgerId and entryId.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -2999,20 +3000,86 @@ private class AnalyzeBacklog extends CliCommand {
...
+                // In analyze-backlog, lastMessageId is null only when: total entries is 0,
+                // with false aborted flag returned.
+                if (StringUtils.isBlank(mergedResult.getLastMessageId())) {
+                    print("Incorrect last message id returned from server");
+                    return;
+                }
+
+                if (!quiet) {
+                    print("Analyze backlog progress, scanned entries: " + mergedResult.getEntries()
+                            + ", scan max entries: " + backlogScanMaxEntries);
```
lhotari: It would be useful to print the current value of mergedResult in json format without linefeeds so that the CLI output can be parsed as [NDJSON](https://en.wikipedia.org/wiki/JSON_streaming#Newline-delimited_JSON).

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -2999,20 +3000,86 @@ private class AnalyzeBacklog extends CliCommand {
...
+                }
+
+                String[] messageIdSplits = mergedResult.getLastMessageId().split(":");
+                MessageIdImpl nextScanMessageId =
+                        new MessageIdImpl(Long.parseLong(messageIdSplits[0]), Long.parseLong(messageIdSplits[1]) + 1,
+                                partitionIndex);
+                startPosition = Optional.of(nextScanMessageId);
+            }
+            print(mergedResult);
```
lhotari: output would have to be json format without linefeeds to make the CLI output parseable as NDJSON. The last line in the output would be the final result. Perhaps there could be a command line option to configure whether ndjson should be used since it's not pretty printed.

pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
```diff
@@ -2181,6 +2182,97 @@ public boolean matches(Long timestamp) {
 
     }
 
+    @Test
+    public void topicsAnalyzeBacklog() throws Exception {
```
lhotari: A test without mocks, similar to the tests in #25127 would be useful in addition since it would serve as an integration test.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
oneby-wang: There are two `ObjectMapper` in `CliCommand`, so we should use the one without pretty printer to print json without linefeeds, and then pass the json string to `print()` method. Is my understanding correct?

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CliCommand.java#L140-L141

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CliCommand.java#L120-L130

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
oneby-wang: > Perhaps there could be a command line option to configure whether ndjson should be used since it's not pretty printed.

A little bit confused. Does this option only take effect on the final result, or does it also apply to the intermediate results?

pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
oneby-wang: I haven't found similar examples for reference. Could you provide me some examples. 

I don't know how to invoke admin CLI in integration tests and how to receive the outputs(since they are printed to the console output).

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
oneby-wang: Could you provide me more information, I couldn't find one method that could parse `ledgerId:entryId` string to a MessageIdAdv instance.

pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
oneby-wang: Seems we need to handle the test in integration module like `CLITest`?

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/tests/integration/src/test/java/org/apache/pulsar/tests/integration/cli/CLITest.java

Is there a better approach? Running the tests in integration module requires local image building.

oneby-wang:
@lhotari Could help me solve the above questions when you have a moment? Especially about how to write integration tests easily in admin CLI module.

I'll refactor this PR using the API that PR https://github.com/apache/pulsar/pull/25127 provided once I'm back.

pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
lhotari: > Seems we need to handle the test in integration module like `CLITest`?
> 
> https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/tests/integration/src/test/java/org/apache/pulsar/tests/integration/cli/CLITest.java
> 
> Is there a better approach? Running the tests in integration module requires local image building.

Yes it's a bit clumbersome.

To build an image quickly, use this script:
```
./build/build_java_test_image.sh
```

That requires setting `export PULSAR_TEST_IMAGE_NAME=apachepulsar/java-test-image:latest` environment variable so that the `apachepulsar/java-test-image:latest` docker image is used instead of the default `apachepulsar/pulsar-test-latest-version:latest` which takes much longer to build.
If you are running the test in IntelliJ, you can edit the default value at https://github.com/apache/pulsar/blob/15daf18a877cddc3786dc0661fde0e049bac916a/tests/integration/src/test/java/org/apache/pulsar/tests/integration/containers/PulsarContainer.java#L62

For running integration tests on command line, this is the sequence I have used:

```
# compile integration test dependencies
mvn -am -pl tests/integration -Dcheckstyle.skip=true -Dlicense.skip=true -Dspotbugs.skip=true -DskipTests install
./build/build_java_test_image.sh
export PULSAR_TEST_IMAGE_NAME=apachepulsar/java-test-image:latest
# run the test
mvn -DintegrationTests -pl tests/integration -DtestRetryCount=0 -DredirectTestOutputToFile=false test \
-Dtest=TestClassNameHere
```

If you are working on the test alone, you obviously don't have to rebuild the image after each change.

lhotari:
@oneby-wang #25127 has been merged

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
lhotari: > A little bit confused. Does this option only take effect on the final result, or does it also apply to the intermediate results?

it would be useful to be able to printout the aggregated result in the loop (in `Predicate<AnalyzeSubscriptionBacklogResult> terminatePredicate` added in #25127) in ndjson format (json without linefeeds) or json format. The ndjson format would be machine readable so that's the reason to have that option.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
lhotari: yes, there would have to be an option to print out without linefeeds (no pretty printing, line feeds escaped).

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -2999,20 +3000,86 @@ private class AnalyzeBacklog extends CliCommand {
...
+            int partitionIndex = TopicName.get(persistentTopic).getPartitionIndex();
             if (isNotBlank(messagePosition)) {
-                int partitionIndex = TopicName.get(persistentTopic).getPartitionIndex();
                 MessageId messageId = validateMessageIdString(messagePosition, partitionIndex);
                 startPosition = Optional.of(messageId);
             }
-            print(getTopics().analyzeSubscriptionBacklog(persistentTopic, subName, startPosition));
 
+            AnalyzeSubscriptionBacklogResult mergedResult = null;
+            while (true) {
```
lhotari: Are you planning to reuse the new API added in #25127 ?

oneby-wang: Yes, I'll refactor this PR by reusing the new API added in https://github.com/apache/pulsar/pull/25127 and resolve the above comments once I'm back.

I'll re-request review once the refactoring is done.

oneby-wang: @lhotari PR refactored, please review.

oneby-wang:
Hi @lhotari, could you please help review this refactored PR?

pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
```diff
@@ -1738,10 +1737,6 @@ public void topics() throws Exception {
         verify(mockTopics).createSubscription("persistent://myprop/ns1/ds1", "sub1",
                 MessageId.earliest, false, null);
 
-        cmdTopics.run(split("analyze-backlog persistent://myprop/ns1/ds1 -s sub1"));
-        verify(mockTopics).analyzeSubscriptionBacklog("persistent://myprop/ns1/ds1", "sub1",
-                Optional.empty());
-
```
lhotari, line 1744: it could be useful to keep a unit test for parameter parsing since this would catch issues before the integration test is run

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -3002,24 +3003,44 @@ private class AnalyzeBacklog extends CliCommand {
...
+        @Option(names = {"--backlog-scan-max-entries",
+                "-b"}, description = "The maximum number of backlog entries the client will scan before terminating "
+                + "its loop", required = false)
+        private long backlogScanMaxEntries = -1;
+
+        @Option(names = {"--quiet", "-q"}, description = "Disable analyze-backlog progress reporting", required = false)
+        private boolean quiet = false;
+
+        @Option(names = {"--plain-print",
+                "-pp"}, description = "Plain(Non-pretty) print the final result output as NDJSON", required = false)
```
lhotari: since there's already `-p`, it's not recommended to add `-pp`. Instead, it's better to shorten `--plain-print` to `--plain` and just not add a short option name at all.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -3002,24 +3003,44 @@ private class AnalyzeBacklog extends CliCommand {
...
-            print(getTopics().analyzeSubscriptionBacklog(persistentTopic, subName, startPosition));
 
+            AnalyzeSubscriptionBacklogResult backlogResult =
+                    getTopics().analyzeSubscriptionBacklogAsync(persistentTopic, subName, startPosition, result -> {
+                        boolean terminate = result.getEntries() >= backlogScanMaxEntries;
+                        if (!quiet && !terminate) {
+                            print(result, false);
+                        }
+                        return terminate;
+                    }).get();
```
lhotari: using the async method isn't useful since `.get()` is called. There would be a need to change to `throws Exception` when using `analyzeSubscriptionBacklog` method.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
```diff
@@ -3002,24 +3003,44 @@ private class AnalyzeBacklog extends CliCommand {
...
-        @Option(names = { "--position",
-                "-p" }, description = "message position to start the scan from (ledgerId:entryId)", required = false)
+        @Option(names = {"--position",
+                "-p"}, description = "Message position to start the scan from (ledgerId:entryId)", required = false)
         private String messagePosition;
 
+        @Option(names = {"--backlog-scan-max-entries",
+                "-b"}, description = "The maximum number of backlog entries the client will scan before terminating "
+                + "its loop", required = false)
+        private long backlogScanMaxEntries = -1;
```
lhotari: --backlog-scan-max-entries default of -1 relies on a subtle side-effect for backward compat — CmdTopics.java:3019   With -1, the predicate result.getEntries() >= -1 is always true, so the loop completes on the first iteration (matching old  single-call behavior). That works, but it conflates "unset" with "terminate immediately." Preferred: treat unset as "no cap" (use  the no-predicate overload analyzeSubscriptionBacklogAsync(topic, sub, pos)) and only take the looping path when -b is supplied.
  That makes intent explicit and avoids relying on entries >= -1 semantics. Also add a validation error for -b 0 or negative user  input, since those are meaningless.

lhotari (commented):
A few minor comments

pulsar-client-tools-test/src/test/java/org/apache/pulsar/admin/cli/PulsarAdminToolTest.java
oneby-wang: Addressed.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
oneby-wang: Addressed.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
oneby-wang: Addressed.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdTopics.java
oneby-wang: Addressed.

lhotari (approved):
LGTM

lhotari:
Thanks for the contribution @oneby-wang

Issue #25083, opened 2025-12-16
[Enhancement] Add client side looping to "pulsar-admin topics analyze-backlog" so that a large backlog can be analyzed without increasing subscriptionBacklogScanMaxEntries/subscriptionBacklogScanMaxTimeMs


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

For `pulsar-admin topics analyze-backlog`, there are broker side configurable maximum limits how many entries will be read and how long the scanning can take. `subscriptionBacklogScanMaxTimeMs` (default 2 minutes) and `subscriptionBacklogScanMaxEntries` (default 10000) control this behavior.

https://github.com/apache/pulsar/blob/3efc174f42de22523c51608858aef581b8743509/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java#L783-L792

Increasing these settings is possible. However, it's possible that the HTTP request times out (also idle timeout in NAT/firewall etc.) before the command completes so increasing the limits might not be useful beyond a few minutes.

### Solution

There should be client-side looping which calls the [`analyzeSubscriptionBacklog` API](https://pulsar.apache.org/api/admin/4.1.x/org/apache/pulsar/client/admin/Topics.html#analyzeSubscriptionBacklog(java.lang.String,java.lang.String,java.util.Optional)) with the starting point to where the previous result left off. Final results should be aggregated on client side.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

oneby-wang:
Hi, @lhotari, I have some questions about this enhancement: 

1. When should the client-side looping stop? Should we move the two server params(`subscriptionBacklogScanMaxEntries ` and `subscriptionBacklogScanMaxTimeMs `) to client side(as input params), and add one more param to set `backlogScanMaxEntries` in client-side every single loop, so the client side can control the loop termination condition, and set every scan request timeout?

2. From the issue description, seems we need to add a new method to `TopicsImpl` to support setting these params, and pass the default values to new method in old method to avoid breaking changes.

https://github.com/apache/pulsar/blob/82ab144e6d38b5544ab814b02abf8c4099b94150/pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java#L1555-L1561

3. Add params for `pulsar-admin topics analyze-backlog` cli to support custom setting.

lhotari:
[...]

PR #25127, opened 2026-01-06
[improve][admin] Add client side looping to analyze-backlog in Topics to avoid potential HTTP call timeout

oneby-wang:
Fixes https://github.com/apache/pulsar/issues/25083

### Motivation

Use client-side looping instead of increasing broker settings to avoid potential HTTP call timeout in analyze-backlog method of Topics.

### Modifications

Add client-side looping, add test.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/22

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
@@ -71,12 +71,12 @@ class OpScan implements ReadEntriesCallback {
             lastSeenPosition = lastPositionForBatch;
             for (Entry entry : entries) {
                 if (remainingEntries.getAndDecrement() <= 0) {
-                    log.warn("[{}] Scan abort after reading too many entries", OpScan.this.cursor);
+                    log.info("[{}] Scan abort after reading too many entries", OpScan.this.cursor);
                     callback.scanComplete(lastSeenPosition, ScanOutcome.ABORTED, OpScan.this.ctx);
                     return;
                 }
                 if (!condition.test(entry)) {
-                    log.warn("[{}] Scan abort due to user code", OpScan.this.cursor);
+                    log.info("[{}] Scan abort due to user code", OpScan.this.cursor);
                     callback.scanComplete(lastSeenPosition, ScanOutcome.USER_INTERRUPTED, OpScan.this.ctx);
                     return;
                 }
@@ -111,12 +111,12 @@ class OpScan implements ReadEntriesCallback {
 
     public void find() {
         if (remainingEntries.get() <= 0) {
-            log.warn("[{}] Scan abort after reading too many entries", OpScan.this.cursor);
+            log.info("[{}] Scan abort after reading too many entries", OpScan.this.cursor);
             callback.scanComplete(lastSeenPosition, ScanOutcome.ABORTED, OpScan.this.ctx);
             return;
         }
         if (System.currentTimeMillis() - startTime > timeOutMs) {
-            log.warn("[{}] Scan abort after hitting the deadline", OpScan.this.cursor);
+            log.info("[{}] Scan abort after hitting the deadline", OpScan.this.cursor);
             callback.scanComplete(lastSeenPosition, ScanOutcome.ABORTED, OpScan.this.ctx);
             return;
         }


pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
```diff
@@ -1591,6 +1604,92 @@ public void failed(Throwable throwable) {
...
+                    log.warn("[{}][{}] Scanned last message id is blank, abort analyze backlog, start position is: {}",
+                            topic, subscriptionName, startPositionRef.get());
+                    future.completeExceptionally(
+                            new PulsarAdminException("Incorrect last message id returned from server"));
+                }
+
+                String[] messageIdSplits = mergedResult.getLastMessageId().split(":");
+                MessageIdImpl nextScanMessageId =
+                        new MessageIdImpl(Long.parseLong(messageIdSplits[0]), Long.parseLong(messageIdSplits[1]) + 1,
+                                partitionIndex);
```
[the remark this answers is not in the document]
oneby-wang: Actually, `lastMessageId` is returned by the server and is guaranteed to be valid. Keep it simple here.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
```diff
@@ -69,29 +69,23 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
+                    log.warn("[{}] Scan abort after reading too many entries", OpScan.this.cursor);
+                    callback.scanComplete(lastSeenPosition, ScanOutcome.ABORTED, OpScan.this.ctx);
+                    return;
+                }
+                if (!condition.test(entry)) {
+                    log.warn("[{}] Scan abort due to user code", OpScan.this.cursor);
+                    callback.scanComplete(lastSeenPosition, ScanOutcome.USER_INTERRUPTED, OpScan.this.ctx);
+                    return;
                 }
             }
```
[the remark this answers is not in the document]
oneby-wang: Filter process is already handled by `OpReadEntry`. I think filtering is not needed here.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
```diff
@@ -2220,21 +2220,88 @@ AnalyzeSubscriptionBacklogResult analyzeSubscriptionBacklog(String topic, String
...
+     * @param startPosition
+     *           the position to start the scan from (empty means the last processed message)
+     * @param backlogScanMaxEntries
+     *           the maximum number of backlog entries the client will scan before terminating its loop
+     * @return an accurate analysis of the backlog
+     */
+    CompletableFuture<AnalyzeSubscriptionBacklogResult> analyzeSubscriptionBacklogAsync(String topic,
+                                                                            String subscriptionName,
+                                                                            Optional<MessageId> startPosition,
+                                                                            long backlogScanMaxEntries);
```
lhotari, line 2327: It would be useful to have a method where `long backlogScanMaxEntries` would be replaced with `java.util.function.Predicate<AnalyzeSubscriptionBacklogResult> continuePredicate`. This would accept the accumulated `AnalyzeSubscriptionBacklogResult` as a parameter. The benefit of this is that the user of the API can choose to add logging when the accumulated result gets updated.

The method signature with the `long backlogScanMaxEntries` can delegate to this method with the `java.util.function.Predicate<AnalyzeSubscriptionBacklogResult> continuePredicate` parameter.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
```diff
@@ -1591,6 +1604,94 @@ public void failed(Throwable throwable) {
...
+            @Override
+            public void accept(AnalyzeSubscriptionBacklogResult currentResult, Throwable throwable) {
+                if (throwable != null) {
+                    future.completeExceptionally(throwable);
+                    return;
+                }
+
+                AnalyzeSubscriptionBacklogResult mergedResult = mergeBacklogResults(currentResult, resultRef.get());
+                resultRef.set(mergedResult);
+                if (!mergedResult.isAborted() || mergedResult.getEntries() >= backlogScanMaxEntries) {
```
lhotari: the `!mergedResult.isAborted()` part isn't correct.

https://github.com/apache/pulsar/blob/da0d11644f097ae657e79b2eb7835478d79ccd0e/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java#L119-L128

The purpose of the loop is to keep on iterating until the result is COMPLETED or the `java.util.function.Predicate<AnalyzeSubscriptionBacklogResult> continuePredicate` returns false (condition in continuePredicate could be `result.getEntries() < backlogScanMaxEntries`).

It seems that `conf.setSubscriptionBacklogScanMaxEntries(serverSubscriptionBacklogScanMaxEntries);` isn't effective in the tests since some of the tests should have failed.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
oneby-wang: Nice suggestion! I'd prefer to get this PR merged first, so the admin CLI in https://github.com/apache/pulsar/pull/25126 could reuse the code.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
oneby-wang: > the !mergedResult.isAborted() part isn't correct.

In `PersistentTopicsBase`, we convert `rawResult.getScanOutcome() != ScanOutcome.COMPLETED` to `boolean aborted`, and the orginal `scanOutcome` result is not returned back to the client, so I used `!aborted` to replace `scanOutcome == ScanOutcome.COMPLETED`.

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java#L1704-L1709

`OpScan` would only completes with `ScanOutcome.COMPLETED` if there is no more entries to scan.

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java#L97-L101

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java#L129-L135

Only this if block returns the result of ScanOutcome.USER_INTERRUPTED.

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java#L84-L88

In `analyzeBacklog()` method, we always return `true`, so the `scanOutcome` is `ScanOutcome.COMPLETED` or `ScanOutcome.ABORTED`.

https://github.com/apache/pulsar/blob/85625e0f100479dd95fb1311aeb52411b6b0a25d/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L657-L710

Should I add a field in `AnalyzeSubscriptionBacklogResult` to return the orginal `scanOutcome` result?

oneby-wang: > It seems that conf.setSubscriptionBacklogScanMaxEntries(serverSubscriptionBacklogScanMaxEntries); isn't effective in the tests since some of the tests should have failed.

I use `conf.setSubscriptionBacklogScanMaxEntries(serverSubscriptionBacklogScanMaxEntries)` to crontol the loop condition. 

I ran all these tests locally, and they all passed(as I expected). Please let me know which unit tests are expected to fail.

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/Topics.java
oneby-wang: Applied `java.util.function.Predicate<AnalyzeSubscriptionBacklogResult> terminatePredicate` as a parameter.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
```diff
@@ -1591,6 +1613,103 @@ public void failed(Throwable throwable) {
...
+                    return;
+                }
+
+                // To avoid infinite loops, we ensure the entry count is incremented after each loop.
+                // Should never happen.
+                if (currentResult.getEntries() <= 0) {
+                    log.warn("[{}][{}] Scanned total entry count is zero or negative, abort analyze backlog, start "
+                            + "position is: {}", topic, subscriptionName, startPositionRef.get());
+                    future.completeExceptionally(
+                            new PulsarAdminException("Incorrect total entry count returned from server"));
```
lhotari: I think that this shouldn't complete exceptionally. I would assume that it's possible to have a case where there's exactly the requested amount of entries available on the broker side and it would return the results because of reaching the max limit of an individual call. On the next call, the request would return 0 entries.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
```diff
@@ -1591,6 +1613,103 @@ public void failed(Throwable throwable) {
...
+                    return;
+                }
+
+                // In analyze-backlog, lastMessageId is null only when: total entries is 0, with false aborted flag
+                // returned. Should never happen.
+                if (StringUtils.isBlank(mergedResult.getLastMessageId())) {
+                    log.warn("[{}][{}] Scanned last message id is blank, abort analyze backlog, start position is: {}",
+                            topic, subscriptionName, startPositionRef.get());
+                    future.completeExceptionally(
```
lhotari: If the given message id as input doesn't return any results, this would be expected. Similar case as mentioned in the previous comment.

lhotari (commented):
A few comments. I'm sorry for the delay in the review.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
oneby-wang: Addressed.

> I would assume that it's possible to have a case where there's exactly the requested amount of entries available on the broker side and it would return the results because of reaching the max limit of an individual call. On the next call, the request would return 0 entries.

I think this would not happen. If `ScanOutcome != ScanOutcome.COMPLETED` in `OpScan`, there should be more entries to scan.

But I strongly agree with the comment, completing exceptionally is less developer-friendly. Since the broker scan logic could be changed in the future, and it is every reasonable to treat 0 entries or a null `lastMessageId` as no more entries.

`ScanOutcome.USER_INTERRUPTED` would not returned in `OpScan`, `ScanOutcome.COMPLETED` is returned in the following conditions:

```
if (cursor.hasMoreEntries(searchPosition)) {
    OpReadEntry opReadEntry = OpReadEntry.create(cursor, searchPosition, batchSize,
    this, OpScan.this.ctx, null, null, false);
    ledger.asyncReadEntries(opReadEntry);
} else {
    callback.scanComplete(lastSeenPosition, ScanOutcome.COMPLETED, OpScan.this.ctx);
}
```

or

```
searchPosition = ledger.getPositionAfterN(lastPositionForBatch, 1,
                    PositionBound.startExcluded);
if (searchPosition.compareTo(lastPositionForBatch) == 0) {
    // we have reached the end of the ledger, as we are not doing progress
    callback.scanComplete(lastSeenPosition, ScanOutcome.COMPLETED, OpScan.this.ctx);
    return;
}
```

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/TopicsImpl.java
oneby-wang: Addressed.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
```diff
@@ -69,29 +69,23 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
-                        return;
-                    }
-                    if (!condition.test(entry)) {
-                        log.warn("[{}] Scan abort due to user code", OpScan.this.cursor);
-                        callback.scanComplete(lastSeenPosition, ScanOutcome.USER_INTERRUPTED, OpScan.this.ctx);
-                        return;
-                    }
+            for (Entry entry : entries) {
+                if (remainingEntries.getAndDecrement() <= 0) {
+                    log.warn("[{}] Scan abort after reading too many entries", OpScan.this.cursor);
```
lhotari: It's not a problem caused by this PR, but I think that `warn` level is too high for this log message. It would be expected that the broker would scan a certain amount of entries in a single request and therefore there's no need to warn about it. 

Please change the log level to `info`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
```diff
@@ -69,29 +69,23 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
+                    callback.scanComplete(lastSeenPosition, ScanOutcome.ABORTED, OpScan.this.ctx);
+                    return;
+                }
+                if (!condition.test(entry)) {
+                    log.warn("[{}] Scan abort due to user code", OpScan.this.cursor);
```
lhotari: Please change the log level to `info`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
lhotari: Please change the log level to `info`.

lhotari: Please change the log level to `info`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpScan.java
oneby-wang: Think so. Addressed.

lhotari (approved):
LGTM. Great work @oneby-wang

oneby-wang:
Thanks for your great suggestions! @lhotari

PR #17238, opened 2022-08-23
[broker][admin]Add api for update topic properties

AnonHxy:
### Motivation

* Maybe this modification should be part of  https://github.com/apache/pulsar/issues/12629 .  As described in the PIP document "These metadata could be set during topic creation and also updated.",  we have implemented the `creation` method, but have not implement  the `updated` method. 

* This PR will implement the topic meta update

### Modifications

* Add a admin api `updateProperties`

### Verifying this change

- [x] Make sure that the change passes the CI checks.
- Add UT `org.apache.pulsar.broker.admin.AdminApi2Test#testUpdatePartitionedTopicProperties`
- Add UT `org.apache.pulsar.broker.admin.AdminApi2Test#testUpdateNonPartitionedTopicProperties`

### Documentation

Check the box below or label this PR directly.

Need to update docs? 

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)
  
- [x] `doc-not-needed` 
(Please explain why)
  
- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

PR #12818, opened 2021-11-15
PIP 110: Support Topic metadata - PART-1  create topic with properties

Technoboy-:
Fixes #12629

## Motivation
The original discussion mail :
https://lists.apache.org/thread/m9dkhq1fs6stsdwh78h84fsl5hs5v67f

Introduce the ability to store metadata about topics.

This would be very useful as with metadata you could add labels and other
pieces of information that would allow defining the purpose of a topic,
custom application-level properties.
This feature will allow application-level diagnostic tools and maintenance
tools to not need external databases to store such metadata.

Imagine that we could add a simple key value map (String keys and String
values) to the topic.
These metadata could be set during topic creation and also updated.

## Modification

- Add two new rest API reduce compatibility issue : 
  @Path("/{property}/{cluster}/{namespace}/{topic}/partitions/properties")
  @Path("/{tenant}/{cluster}/{namespace}/{topic}/properties")

- Add below new methods in Topics.java :
  createPartitionedTopic(String topic, int numPartitions, Map<String, String> properties);
  createPartitionedTopicAsync(String topic, int numPartitions, Map<String, String> properties);
  createNonPartitionedTopic(String topic, Map<String, String> properties);
  createNonPartitionedTopicAsync(String topic, Map<String, String> properties);

### Documentation
  
- [ x ] `no-need-doc`

PR #25134, opened 2026-01-11
[feat][pip] PIP-452: Customizable topic listing of namespace with properties

coderzc:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-452.md
+++ b/pip/pip-452.md
@@ -80,7 +80,7 @@ GET /admin/v2/persistent/{tenant}/{namespace}?properties=k1=v1,k2=v2
 
 CLI:
 ```
-pulsar-admin topics list <tenant>/<namespace> -p k1=v1,k2=v2
+pulsar-admin topics list <tenant>/<namespace> -p k1=v1 -p k2=v2
 ```
 
 ## Design & Implementation Details


pip/pip-452.md
```diff
@@ -0,0 +1,212 @@
...
+
+package org.apache.pulsar.broker.namespace;
+
+import org.apache.pulsar.common.naming.NamespaceName;
+import org.apache.pulsar.common.api.proto.CommandGetTopicsOfNamespace.Mode;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public interface NamespaceTopicListingStrategy {
```
BewareMyPower: It's better to extend the existing interceptor rather than introducing yet another plugin, i.e. add a new method to the existing interface rather than introducing a new config and a new interface

coderzc: Applied the suggestions

pip/pip-452.md
```diff
@@ -0,0 +1,306 @@
...
+                        return CompletableFuture.completedFuture(interceptorResult);
+                    } else {
+                        // case B: The interceptor did not handle, so the original query logic is executed.
+                        // Wrap List<String> into TopicListingResult(topics, filtered=false) for unified processing
+                        return getBrokerService().pulsar().getNamespaceService()
+                                .getListOfUserTopics(namespaceName, mode)
+                                .thenApply(topics -> TopicListingResult.success(topics, false));
+                    }
+                }).thenCompose(listingResult -> {
+                    List<String> rawTopics = listingResult.getTopics();
```
BewareMyPower: You don't have to copy so much code here, which does not focus on the key change. The only thing that matters is that what will replace the following existing code:

```java
                        return getBrokerService().pulsar().getNamespaceService()
                                .getListOfUserTopics(namespaceName, mode)
```

pip/pip-452.md
```diff
@@ -0,0 +1,306 @@
...
+        this.topics = topics;
+        this.filtered = filtered;
+    }
+
+
+    /**
+     * Return an invalid result indicating that the caller should continue with the default logic.
+     */
+    public static TopicListingResult passThrough() {
+        return PASS_THROUGH_INSTANCE;
```
BewareMyPower: This design is a bit over-abstraction. IMO, the following design could be simpler and clearer.

```java
public record TopicListingResult(List<String> topics, boolean filtered) {}
```

```java
    /**
     * @return the future of the result, if it's empty, fall back to the built-in implementation to list all topics
     */
    default CompletableFuture<Optional<TopicListingResult>> interceptGetTopicsOfNamespace(
```

BewareMyPower: Since the built-in implementation is very simple that calls method directly on the `NamespaceService`, I think the fallback logic is unnecessary.

coderzc: Since we allow the configuration of multiple interceptors, the return here means that the result of current interception is not processed this method. If all the interceptors do not process the method, the default logic will be used.

In addition, the interception configuration is empty by default. In this case, we still need to fall back to the default logic.

coderzc: > `default CompletableFuture<Optional<TopicListingResult>> interceptGetTopicsOfNamespace()`

Applied this suggestion

pip/pip-452.md
coderzc: Simplified pseudocode

pip/pip-452.md
```diff
@@ -0,0 +1,245 @@
...
+This implementation limits the flexibility required for complex multi-tenant scenarios:
+
+No Client Context: The broker cannot distinguish who is asking for the topics or why. It cannot filter topics based on client properties (these client properties may correspond to, or be derived from topic properties).
+
+Inefficient Filtering: For namespaces with millions of topics, the broker must fetch the full list into memory before applying the topics_pattern regex. There is no way to "push down" the filtering to the data source (e.g., a database with an index).
+
+To address these issues, I propose making the topic listing logic pluggable and extending the protocol to accept client properties.
+
+# Goals
+Protocol: Add a properties field to `CommandGetTopicsOfNamespace` to carry client-side context.
```
lhotari, line 18: How about topic list watchers? There are `CommandWatchTopicList`, `CommandWatchTopicListSuccess` and `CommandWatchTopicUpdate` which would require properties support as well so that the solution would be consistent.

coderzc: Topic watch is not supported first.

If you want to use this feature, you need to disable topic wach.

lhotari: Please mention this in the PIP-452 document since this is relevant context.

pip/pip-452.md
```diff
@@ -0,0 +1,245 @@
...
+# Backward Compatibility
+Protocol: Adding an optional field (properties) to the Protobuf definition is non-breaking. Old clients will not send this field; old brokers will ignore it.
+
+Behavior: The default strategy mimics the current behavior exactly. The Broker retains the final filtering logic, ensuring that even custom strategies cannot return topics that violate the client's requested pattern.
+
+# Security Considerations
+Input Validation: The properties map is user-supplied input. Implementers of the customizable strategy MUST validate and sanitize these inputs before using them, especially if they are used to construct database queries.
+
+Authorization: This PIP only controls the discovery (listing) of topics. It does not bypass the Authorization Service for consuming or producing to those topics.
+
```
lhotari, line 276: "Alternative solutions" section is missing.

I'm wondering if it would have made sense to extend the MetadataStore interfaces to support properties. 
The problem that this PIP-452 causes is that it will be a burden to add additional custom interfaces to add properties to topics and so on. If the MetadataStore interface would support properties, it would be a cleaner approach to close the gaps.

Has the solution for adding properties to MetadataStore been considered? What are the reasons for rejecting this alternative?

pip/pip-452.md
coderzc: Sure, I update doc and add configuration `enableBrokerTopicListWatcher`

pip/pip-452.md
lhotari: Commenting on my own question: I guess topic policy contains the properties at the moment? I guess that PIP-452 only makes sense when using a custom TopicPoliciesService implementation configured with `topicPoliciesServiceClassName`?

lhotari: TopicPolicies (https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/TopicPolicies.java#L46) doesn't contain the properties. How would the properties be added, modified and deleted for topics?

coderzc: This will be maintained by the implementer of the customizable topic listing plugin, and the relationship between properties and topic does not need to be maintained in pulsar.

lhotari: Answering my own question about the topic properties. They were added with [PIP-110](https://github.com/apache/pulsar/blob/master/pip/pip-110.md), #12818 . 
The topic properties for partitioned topics are stored in PartitionedTopicMetadata, 
https://github.com/apache/pulsar/blob/a373258d0430aa04cc05c4c46b0ac5e0e38153cb/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/partition/PartitionedTopicMetadata.java#L33-L34
However, they are stored in LedgerMetadata for non-partitioned topics:
https://github.com/apache/pulsar/blob/75658cc353e74e132ef30090767fe062eae9ca05/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java#L665-L666
(PR #17238)

pip/pip-452.md
```diff
@@ -0,0 +1,263 @@
...
+}
+```
+
+### REST API & CLI Changes
+
+Add `properties` parameter to the REST API endpoint for listing topics in a namespace to list topic with specific properties for customizable topic listing.
+
+REST API (URL encoding is required for values):
+```
+GET /admin/v2/persistent/{tenant}/{namespace}?properties=k1=v1,k2=v2
```
codelipenghui, line 78: It seems to be "&"? the topic has both label k1=v1 and k2=v2? Or we need to mention Pulsar will not define this behavior, it depends on the plugin implementation?

pip/pip-452.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+CLI:
+```
+pulsar-admin topics list <tenant>/<namespace> -p k1=v1 -p k2=v2
```
codelipenghui:
```suggestion
pulsar-admin topics list <tenant>/<namespace> -p k1=v1,k2=v2
```

Just make it consistent with the REST API

pip/pip-452.md
```diff
@@ -0,0 +1,263 @@
...
+     * based on the provided client context properties.
+     *
+     * @param namespace     The namespace being queried.
+     * @param mode          The query mode (PERSISTENT, NON_PERSISTENT, or ALL).
+     * @param topicsPattern Optional regex pattern provided by the client.
+     * @param properties    Context properties provided by the client.
+     * @return A CompletableFuture containing the result:
+     * If the future completes with {@code Optional.empty()}, proceed to the next interceptor or broker's default logic to list all topics.
+     */
+    default CompletableFuture<Optional<TopicListingResult>> interceptGetTopicsOfNamespace(
```
codelipenghui:
```suggestion
    default CompletableFuture<Optional<TopicListingResult>> onGetTopicsOfNamespace(
```

pip/pip-452.md
```diff
@@ -0,0 +1,263 @@
...
+            NamespaceName namespace,
+            CommandGetTopicsOfNamespace.Mode mode,
+            Optional<String> topicsPattern,
+            Map<String, String> properties) {
+        return CompletableFuture.completedFuture(Optional.empty());
```
codelipenghui: We need to keep the pulsar's implementation by default? Otherwise, this change will introduce compatibility issues if user has their own broker interceptor plugin but no implementation for this method.

pip/pip-452.md
```diff
@@ -0,0 +1,263 @@
...
+# Detailed Design
+
+## Public-facing Changes
+
+### Configuration
+
+broker.conf
+```properties
+# Enables watching topic add/remove events on broker side. It is separated from enableBrokerSideSubscriptionPatternEvaluation.
+enableBrokerTopicListWatcher = true
```
codelipenghui, line 46: How this configuration will be used? Here defined the configuration name, but there is no detailed behavior definition for this configuration.

pip/pip-452.md
```diff
@@ -0,0 +1,263 @@
...
+                BrokerInterceptor brokerInterceptor = getBrokerService().getPulsar().getBrokerInterceptor();
+                CompletableFuture<Optional<TopicListingResult>> interceptorFuture =
+                        brokerInterceptor == null ? CompletableFuture.completedFuture(Optional.empty()) :
+                            brokerInterceptor.interceptGetTopicsOfNamespace(namespaceName, mode, topicsPattern, properties);
+
+                return interceptorFuture.thenCompose(interceptorResult -> {
+                    // Decision branch: Use the interceptor result OR fall back to the default logic
+                    if (interceptorResult != null && interceptorResult.isPresent()) {
+                        // case A: The interceptor has taken over the request
+                        return CompletableFuture.completedFuture(interceptorResult.get());
```
codelipenghui: It seems incorrect for the interceptor implementation that intentionally to return an empty topic list to the client, for example, no properties match.

pip/pip-452.md
coderzc: Yes, this query behavior is determined by the plug-in, and it is not certain how many keys are contained in the properties.

pip/pip-452.md
coderzc: > ## Out of Scope

>Not support topic list watching for customized topic listing in this PIP. It can be considered in future work.
>If you want to use this feature, you need to disable topic list watcher by set `enableBrokerTopicListWatcher = false`.

pip/pip-452.md
coderzc: return an empty topic list? 

Do you mean that return `Optional.empty()`? `Optional.empty` is invalid result, It means that the interceptor did not handle this request.

https://github.com/apache/pulsar/pull/25134#discussion_r2684569996

pip/pip-452.md
coderzc: `Optional.empty()` is invalid result, proceed to the next interceptor or fall back to broker's default logic to list all topics

pip/pip-452.md
coderzc: done

coderzc:
After discussions with @codelipenghui and @BewareMyPower, we found that the current working method of `BrokerInterceptor` cannot accurately specify which result to use when multiple BrokerInterceptors are configured. Therefore, a new pluggable interface `PulsarResourcesExtended` has been introduced to implement this feature.

`PulsarResourcesExtended` is defined as a general interface used as an extension for  customized PulsarResources

pip/pip-452.md
coderzc: I think the current definition is more friendly to cli.

pip/pip-452.md
codelipenghui: Ok, so can we remove all the related information for `enableBrokerTopicListWatcher`? I mean, after we supported `topic list watching for customized topic listing`, this configuration will be useless?

codelipenghui: If it's required to be added, please help provide more information about how this configuration works. I'm not pretty clear about this part.

pip/pip-452.md
codelipenghui: Just checked the Kubernetes style

```
kubectl get pods -l app=nginx,environment=production
```

coderzc:
Since the vote has been passed, merge this PIP

PR #25158, opened 2026-01-17
[fix][broker] Align bundles logic in getNamespacePoliciesAsync with sync method

### LINKED ISSUE #25128: [Bug] Logic for setting policies.bundle is different in AdminResource#getNamespacePolicies compared to AdminResource#getNamespacePoliciesAsync
### Search before reporting

### Read release policy

### User environment

Pulsar master branch, reading the code

### Issue Description

The logic for setting `policies.bundles` is slightly different in `org.apache.pulsar.broker.admin.AdminResource#getNamespacePolicies` compared to the logic in the asynchronous method `org.apache.pulsar.broker.admin.AdminResource#getNamespacePoliciesAsync`.

https://github.com/apache/pulsar/blob/e041fab73c1b9a6bb93457d186fc0a937beafd53/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java#L318-L369

### Error messages

```text
No errors
```

### Reproducing the issue

n/a

### Additional information

_No response_

zooo-code:
Fixes #25128

### Motivation

The logic for setting `policies.bundles` is different between `getNamespacePolicies` (sync) and `getNamespacePoliciesAsync`:

- `getNamespacePolicies` (sync):
    - bundles source: `NamespaceBundleFactory.getBundles().getBundlesData()`
    - fallback: `bundleData != null ? bundleData : policies.bundles`
    - migrated fallback: `false` when LocalPolicies is absent

- `getNamespacePoliciesAsync`:
    - bundles source: `LocalPolicies.bundles`
    - fallback: none (only sets if `localPolicies.isPresent()`)
    - migrated fallback: none

This inconsistency can cause different results when calling sync vs async methods for the same namespace.

### Modifications

- Refactored `getNamespacePolicies` (sync) to call `getNamespacePoliciesAsync` and wait for the result, eliminating code duplication
- `getNamespacePoliciesAsync` now uses `NamespaceBundleFactory.getBundlesAsync()` instead of `LocalPolicies.bundles`
- Applied the same fallback logic: `bundleData != null ? bundleData : policies.bundles` 
- Ensured `migrated` field defaults to `false` when LocalPolicies is  absent

### Verifying this change

Added testGetNamespacePoliciesSyncAsyncBundlesConsistency test in AdminApiTest to verify that sync (getPolicies) and async (getPoliciesAsync) methods return consistent bundles data.

Existing tests only verify sync getPolicies().bundles behavior (e.g., AdminApiTest:816,822,3223, NamespacesTest:1266). There was no test that verifies async getPoliciesAsync returns the same bundles data as sync method. This new test explicitly validates the consistency between sync and async methods, which is the core fix of this PR.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/zooo-code/pulsar

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/AdminResource.java
@@ -333,7 +333,12 @@ public abstract class AdminResource extends PulsarWebResource {
                             .getBundlesAsync(namespaceName)
                             .thenApply(bundles -> bundles.getBundlesData());
                     CompletableFuture<Optional<LocalPolicies>> localPoliciesFuture =
-                            getLocalPolicies().getLocalPoliciesAsync(namespaceName);
+                            getLocalPolicies().getLocalPoliciesAsync(namespaceName)
+                                    .exceptionally(ex -> {
+                                        log.warn("Failed to load local policies for {}, using empty: {}",
+                                                namespaceName, ex.getMessage());
+                                        return Optional.empty();
+                                    });
 
                     return bundlesFuture.thenCombine(localPoliciesFuture, (bundleData, localPolicies) -> {
                         policies.bundles = bundleData != null ? bundleData : policies.bundles;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/NamespaceBundleFactory.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/common/naming/NamespaceBundleFactory.java
@@ -177,7 +177,12 @@ public class NamespaceBundleFactory {
                     return pulsar.getPulsarResources().getLocalPolicies()
                             .createLocalPoliciesAsync(namespace, localPolicies)
                             .thenApply(stat -> getBundles(namespace,
-                                    Optional.of(Pair.of(localPolicies, 0L))));
+                                    Optional.of(Pair.of(localPolicies, 0L))))
+                            .exceptionally(ex -> {
+                                // If LocalPolicies already exists (possibly with empty content),
+                                // just use the bundles from global policies
+                                return getBundles(namespace, Optional.empty());
+                            });
                 });
     }
 
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java
@@ -136,7 +136,15 @@ public class MetadataCacheImpl<T> implements MetadataCache<T>, Consumer<Notifica
 
                     try {
                         GetResult res = optRes.get();
-                        T obj = serde.deserialize(path, res.getValue(), res.getStat());
+                        byte[] value = res.getValue();
+
+                        // Handle empty content before attempting deserialization
+                        if (value == null || value.length == 0) {
+                            log.warn("Empty content found for key '{}', treating as non-existent", path);
+                            return FutureUtils.value(Optional.empty());
+                        }
+
+                        T obj = serde.deserialize(path, value, res.getStat());
                         return FutureUtils
                                 .value(Optional.of(new CacheGetResult<>(obj, res.getStat())));
                     } catch (Throwable t) {


pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AdminApiTest.java
```diff
@@ -3429,6 +3429,36 @@ public void testCreateAndDeleteNamespaceWithBundles() throws Exception {
         deleteNamespaceWithRetry(ns, false);
     }
 
+    @Test
+    public void testGetNamespacePoliciesSyncAsyncBundlesConsistency() throws Exception {
```
lhotari: Did this test fail before making the change to production code? If not, I think that it would be more useful to verify that the behavior is correct for the async method regarding the modified detail.

lhotari:
> ### Matching PR in forked repository
> PR in forked repository: https://github.com/zooo-code/pulsar

@zooo-code  the intention is that contributors would setup ["Personal CI"](https://pulsar.apache.org/contribute/personal-ci/) so that they could run the Pulsar CI without dependency on apache/pulsar maintainers approving CI runs.

zooo-code:
> > ### Matching PR in forked repository
> > PR in forked repository: https://github.com/zooo-code/pulsar
> 
> @zooo-code the intention is that contributors would setup ["Personal CI"](https://pulsar.apache.org/contribute/personal-ci/) so that they could run the Pulsar CI without dependency on apache/pulsar maintainers approving CI runs.

Thanks for the pointer! I'll enable GitHub Actions on my forked repository right away to run the CI independently.

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AdminApiTest.java
zooo-code: @lhotari Thanks for the feedback!

You're right—the current test would likely pass even before the change because LocalPolicies.bundles and NamespaceBundleFactory.getBundles() often return the same value in normal scenarios.

To address this, I'll update the test to explicitly verify that the bundles are retrieved via NamespaceBundleFactory. Would mocking NamespaceBundleFactory and verifying the call to getBundlesAsync() be the preferred approach here?

zooo-code: @lhotari 

I've updated the test to `testGetNamespacePoliciesAsyncReflectsBundleSplit`.
This test explicitly verifies that `getPoliciesAsync` returns the latest bundles from `NamespaceBundleFactory` after a bundle split:

1. Create namespace with 1 bundle
2. Split the bundle → 2 bundles
3. Verify `getPoliciesAsync()` returns 2 bundles

**Before the fix**: `LocalPolicies.bundles` would still show 1 bundle (stale data) → test fails
**After the fix**: `NamespaceBundleFactory.getBundlesAsync()` provides the latest 2 bundles → test passes

This approach verifies the actual behavior change rather than just comparing sync/async consistency, which would pass even before the fix.

zooo-code:
@lhotari  Could you please approve the CI run? I fixed the checkstyle violations.

lhotari:
> @lhotari Could you please approve the CI run? I fixed the checkstyle violations.

@zooo-code As the first step, please enable [Personal CI](https://pulsar.apache.org/contribute/personal-ci/) and complete a run in your fork by opening a PR for this same branch to your own forked repository. You will just need to enable GitHub Actions in your own repository before this. It's free. This will be helpful if you plan to make more contributions.

There might be more CI failures and it's better that you can run CI on your own. Just be aware of many flaky tests in Pulsar. You can rerun failed build steps to get pass them.

zooo-code:
Hi @lhotari,

I've refactored the code so that sync calls async as you suggested. However, I'm encountering a test failure in CI.

**Test failure:**

* Test: `AdminApiSchemaTest.createKeyValueSchema[V2]`
* CI run: [https://github.com/zooo-code/pulsar/actions/runs/21166460377/job/60873078404](https://github.com/zooo-code/pulsar/actions/runs/21166460377/job/60873078404)
* Error:
  Failed to deserialize payload for key
  `/admin/local-policies/schematest/test`
  Caused by: `MismatchedInputException: No content to map due to end-of-input`

**Root cause:**
When `LocalPolicies` has corrupted/empty data in the metadata store, deserialization fails. The `.exceptionally()` handler doesn't catch this because the exception occurs inside `MetadataCacheImpl`'s `CompletableFuture` chain.

**Possible solutions:**

1. Add exception handling to all `getLocalPoliciesAsync()` call sites (8 places)
2. Modify `LocalPoliciesResources.getLocalPoliciesAsync()` to handle deserialization failures gracefully
3. Different approach?

Could you advise on the preferred solution?

Thanks!

zooo-code:
Hi @lhotari

I've refactored the code as suggested so that the sync method now calls the async one.

However, I'm hitting a failure in `AdminApiSchemaTest.createKeyValueSchema`.

It's caused by `LocalPolicies` having empty content (0 bytes) in the metadata store, which triggers a deserialization error in the new async path:

```bash
ContentDeserializationException: Failed to deserialize payload for key '/admin/local-policies/schematest/test'
Caused by: MismatchedInputException: No content to map due to end-of-input
```

The refactoring exposed a pre-existing edge case. The original code likely handled data access differently, while the new async approach triggers full object deserialization, which fails on empty content.

I tried catching `ContentDeserializationException` in
`LocalPoliciesResources.getLocalPoliciesAsync()` to return `Optional.empty()`, but the error still propagates.

Should I handle this at the `MetadataCacheImpl` level (treating empty content as non-existent globally), or is there a better approach?

zooo-code:
Hi @lhotari,

I've resolved the `ContentDeserializationException` issue triggered by empty `LocalPolicies` content.

### Summary of Changes
1. **MetadataCacheImpl:** Added a check to treat empty content as `Optional.empty()` to prevent deserialization errors.
2. **AdminResource & NamespaceBundleFactory:** Added exception handling to gracefully fallback (using empty/global policies) if `LocalPolicies` are corrupted or fail to load.

### Verification
* **Local:** `AdminApiSchemaTest` and compatibility tests now pass.
* **Personal CI:** Core broker tests passed (verified unrelated flaky/infra failures). https://github.com/zooo-code/pulsar/pull/1

Ready for review!
Thanks

PR #25163, opened 2026-01-19
[fix][admin] Allow defaultNumPartitions for non-partitioned autoTopicCreation override

### LINKED ISSUE #24998: API usability: defaultNumPartitions validation is too strict for non-partitioned topic type
## Problem Description

When setting `autoTopicCreation` on namespace policy, the API validation prevents setting `defaultNumPartitions` when the topic type is `non-partitioned`. This limitation makes the API harder to use and less flexible.

**Current behavior:**
- Setting `defaultNumPartitions` with `topicType="non-partitioned"` results in a validation error: `"[defaultNumPartitions] is not allowed to be set when the type is non-partition."`

**Code location:**
https://github.com/apache/pulsar/blob/53162ff745f3381ad97afeebd613c291e7b4bd2e/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/policies/data/impl/AutoTopicCreationOverrideImpl.java#L54-L59

## Use Cases Affected

1. **Configuration management**: Users managing configurations programmatically or via infrastructure-as-code tools may want to set a default value even for non-partitioned topics, especially if they plan to switch to partitioned topics later.

2. **API consistency**: The validation requires users to conditionally include/exclude the `defaultNumPartitions` field based on topic type, making the API less consistent and more error-prone.

3. **Configuration templates**: Shared configuration templates or defaults become harder to maintain when they need to conditionally exclude fields.

## Proposed Solution

Consider relaxing this validation to allow `defaultNumPartitions` to be set (but ignored) when `topicType="non-partitioned"`. This would:
- Make the API more user-friendly
- Allow for more flexible configuration management
- Maintain backward compatibility (non-partitioned topics would still be created correctly)
- Simplify client code that doesn't need to conditionally set fields

Alternatively, if there's a technical reason this field must not be set, it would be helpful to document this clearly and potentially provide a clearer error message explaining why.

## Question

Is this limitation necessary from a technical standpoint, or could it be relaxed to improve API usability?

## Environment

- Pulsar version: **3.x**, **4.x**
- Component: Admin API / Namespace Policy

ChimdumebiNebolisa:
Fixes #24998

### Motivation

The namespace `autoTopicCreation` override validation rejects `defaultNumPartitions` whenever `topicType=non-partitioned`. This forces clients and infrastructure-as-code tools to conditionally omit the field, making shared configuration templates harder to maintain.

The relaxation is intentionally narrow. For non-partitioned overrides, `defaultNumPartitions` may be `null`, `0`, or `1`. Values greater than `1` remain invalid.

Accepted values are persisted as submitted but ignored during runtime topic creation. Auto-created topics therefore remain non-partitioned.

### Modifications

- Updated `AutoTopicCreationOverrideImpl.validateOverride` to allow `defaultNumPartitions` for `NON_PARTITIONED` only when the value is `null`, `0`, or `1`.
- Added unit coverage confirming that `0` and `1` are accepted and `2` is rejected.
- Added broker-level coverage confirming that:
  - an accepted value of `1` is persisted but ignored when the topic is auto-created
  - a value of `5` is rejected by the admin REST API with HTTP 412

### Verifying this change

```bash
./gradlew :pulsar-common:test \
  --tests "org.apache.pulsar.common.policies.data.AutoTopicCreationOverrideTest" \
  -PtestRetryCount=0

./gradlew :pulsar-broker:test \
  --tests "org.apache.pulsar.broker.admin.AdminApi2Test.testAutoTopicCreationOverrideNonPartitionedDefaultNumPartitionsIgnored" \
  --tests "org.apache.pulsar.broker.admin.AdminApi2Test.testAutoTopicCreationOverrideNonPartitionedDefaultNumPartitionsRejected" \
  -PtestRetryCount=0

./gradlew quickCheck
```

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [x] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [x] `doc-not-needed`


pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/policies/data/impl/AutoTopicCreationOverrideImpl.java
```diff
@@ -51,11 +51,6 @@ public static ValidateResult validateOverride(AutoTopicCreationOverride override
                 if (override.getDefaultNumPartitions() <= 0) {
                     return ValidateResult.fail("[defaultNumPartitions] cannot be less than 1 for partition type.");
                 }
-            } else if (TopicType.NON_PARTITIONED.toString().equals(override.getTopicType())) {
-                if (override.getDefaultNumPartitions() != null) {
-                    return ValidateResult.fail("[defaultNumPartitions] is not allowed to be"
```
lhotari, line 56: Instead of removing this completely, I'd suggesting checking that `override.getDefaultNumPartitions()` is null, `0` or `1`, so that it would be allowed only in that case. The more consistent solution would be to only allow `0` since that's what value is used internally when the topic is a non-partitioned topic. However, if we make this change, it seems to make sense to allow also `1`.

ChimdumebiNebolisa:
Updated `validateOverride` to keep topic type validation and allow `defaultNumPartitions` for `NON_PARTITIONED` only when `null`, `0`, or `1`. Updated unit tests to cover `0`, `1`, and reject `2`.

lhotari:
> Updated `validateOverride` to keep topic type validation and allow `defaultNumPartitions` for `NON_PARTITIONED` only when `null`, `0`, or `1`. Updated unit tests to cover `0`, `1`, and reject `2`.

Please add this information to the PR description.

lhotari (approved):
LGTM, thanks for the contribution @ChimdumebiNebolisa

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AdminApi2Test.java
```diff
@@ -2458,6 +2458,32 @@ public void testAutoTopicCreationOverrideWithMaxNumPartitionsLimit() throws Exce
...
+        TenantInfoImpl tenantInfo = new TenantInfoImpl(Set.of("role1", "role2"), Set.of("test"));
+        admin.tenants().createTenant(tenantName, tenantInfo);
+        admin.namespaces().createNamespace(namespaceName, Set.of("test"));
+
+        AutoTopicCreationOverride overridePolicy = AutoTopicCreationOverride.builder()
+                .allowAutoTopicCreation(true)
+                .topicType(TopicType.NON_PARTITIONED.toString())
+                .defaultNumPartitions(5)
+                .build();
+        admin.namespaces().setAutoTopicCreation(namespaceName, overridePolicy);
```
lhotari, line 2583: shouldn't this fail?

ChimdumebiNebolisa: Updated this to address the review comment. The acceptance test now uses `1`, and I added a separate test confirming that `5` is rejected with HTTP 412. The targeted tests and `quickCheck` pass locally. CI is currently awaiting maintainer approval.

lhotari (commented):
The updated tests now match the intended narrow validation: non-partitioned overrides accept `null`, `0`, or `1`, while the REST-level test verifies that `5` is rejected with HTTP 412. The accepted value is also verified as persisted but ignored at topic creation, producing a non-partitioned topic. I found no remaining issue in the current 125-line PR diff. The tracker delta contains unrelated upstream merges since the previous approval, so those were excluded in favor of the canonical three-file PR diff.

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AdminApi2Test.java
lhotari: Thanks, this now tests both sides of the intended contract: `1` is accepted but ignored for topic creation, while `5` is rejected with HTTP 412.

ChimdumebiNebolisa:
Fixed the Broker Group 3 CI failure in 1bdd1b6d0d by updating the stale validation-message assertion.

The exact failing test now passes locally, and git diff --check is clean.

The new CI runs are awaiting maintainer approval.

PR #25170, opened 2026-01-21
[improve][broker]Part-2 of PIP-433 add validation when enabling namespace-level Geo-Replication

poorbarcode:
### Motivation

This is part-2 of [PIP-433](https://github.com/apache/pulsar/pull/24485).

### Modifications

Checks compatibility between two clusters when enabling namespace-level replication, which includes the following
- All topics’ partitions that have been created should be the same
- Auto-creation policies between both clusters should be the same, including broker-level and namespace-level.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
@@ -1967,8 +1967,10 @@ public class OneWayReplicatorTest extends OneWayReplicatorTestBase {
         admin1.topics().createPartitionedTopic(tp, 4);
         admin2.topics().createNonPartitionedTopic(tp);
 
-        admin1.namespaces().setNamespaceReplicationClusters(ns, new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
-        admin2.namespaces().setNamespaceReplicationClusters(ns, new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
+        admin1.namespaces().setNamespaceReplicationClusters(ns,
+                new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
+        admin2.namespaces().setNamespaceReplicationClusters(ns,
+                new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
 
         Awaitility.await().untilAsserted(() -> {
             PersistentTopic persistentTopic =
@@ -2037,8 +2039,10 @@ public class OneWayReplicatorTest extends OneWayReplicatorTestBase {
         admin1.topics().createPartitionedTopic(tp, 4);
         admin2.topics().createPartitionedTopic(tp, 8);
 
-        admin1.namespaces().setNamespaceReplicationClusters(ns, new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
-        admin2.namespaces().setNamespaceReplicationClusters(ns, new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
+        admin1.namespaces().setNamespaceReplicationClusters(ns,
+                new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
+        admin2.namespaces().setNamespaceReplicationClusters(ns,
+                new HashSet<>(Arrays.asList(cluster1, cluster2)), false);
 
         Awaitility.await().untilAsserted(() -> {
             PersistentTopic persistentTopic =


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -867,12 +869,331 @@ protected CompletableFuture<Void> internalSetNamespaceReplicationClusters(List<S
...
+            }
+            return null;
+        });
+    }
+
+    /**
+     * Validates that the effective auto-topic creation policies are the same between local and remote clusters.
+     * The effective policy is computed by: namespace-level policy overrides broker-level if it exists.
+     */
+    private CompletableFuture<Void> validateAutoTopicCreationCompatibility(PulsarAdmin remoteAdmin,
```
Denovo1998, line 1149: validateAutoTopicCreationCompatibility(...) does not validate allowAutoTopicCreation?

Should we perhaps include allowAutoTopicCreation in the effective policy and participate in the comparison?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -867,12 +869,331 @@ protected CompletableFuture<Void> internalSetNamespaceReplicationClusters(List<S
...
+                                                                            String remoteCluster) {
+        String namespaceStr = namespaceName.toString();
+
+        // Get local broker config
+        ServiceConfiguration localConfig = pulsar().getConfiguration();
+        TopicType localBrokerAutoCreationType = localConfig.getAllowAutoTopicCreationType();
+        int localBrokerDefaultPartitions = localConfig.getDefaultNumPartitions();
```
Denovo1998, line 1156: No matter whether the effective topicType is partitioned or not, will it compare the defaultNumPartitions?

When both sides are effective non-partitioned, but defaultNumPartitions is different, why are they treated as incompatible?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -867,12 +869,331 @@ protected CompletableFuture<Void> internalSetNamespaceReplicationClusters(List<S
...
+        // Get remote broker config
+        CompletableFuture<Map<String, String>> remoteBrokerConfigFuture =
+                remoteAdmin.brokers().getRuntimeConfigurationsAsync();
+
+        // Get remote namespace policy
+        CompletableFuture<AutoTopicCreationOverride> remoteNsPolicyFuture =
+                remoteAdmin.namespaces().getAutoTopicCreationAsync(namespaceStr)
+                        .exceptionally(ex -> {
+                            // If namespace doesn't have override, return null
+                            if (ex.getCause() instanceof PulsarAdminException.NotFoundException
```
Denovo1998: Do we need FutureUtil.unwrapCompletionException(e) here?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SetReplicationClustersValidationTest.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    @Override
+    @AfterClass(alwaysRun = true, timeOut = 300000)
+    public void cleanup() throws Exception {
+        super.cleanup();
+    }
+    /**
+     * Helper method to clean up a namespace properly.
+     * First removes replication clusters (if any), then deletes the namespace.
```
Denovo1998: Comment says "then deletes the namespace", the implementation is to reset the replication clusters to a single cluster.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -867,12 +869,331 @@ protected CompletableFuture<Void> internalSetNamespaceReplicationClusters(List<S
...
+                                    || ex instanceof PulsarAdminException.NotFoundException) {
+                                return null;
+                            }
+                            throw new CompletionException(ex);
```
Technoboy-: duplicate exception logic, better to add a method for this.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
poorbarcode: [PIP-433](https://github.com/apache/pulsar/pull/24485) will always allow replicator to create topics

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
poorbarcode: > When both sides are effective non-partitioned, but defaultNumPartitions is different, why are they treated as incompatible?

Pulsar does not allow to set a special default partition number with a default topic type Non-Partitioned

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
poorbarcode: Good suggestion. `FutureUtil.unwrapCompletionException(e)` is clearer

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SetReplicationClustersValidationTest.java
poorbarcode: Removed this comment

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
poorbarcode: Used `FutureUtil.unwrapCompletionException(e)` to make it clearer

Technoboy- (approved):

Denovo1998 (approved):

lhotari:
> Should we consider concurrency limiting (semaphore), batching, or state in the document/comments that this is an admin operation and may take time?

This is the biggest concern I have about this chance.

PR #25177, opened 2026-01-22
[fix][ml] Fix NoSuchElementException in EntryCountEstimator caused by a race condition

### LINKED ISSUE #25174: [Bug][Broker] Escaped NoSuchElementException exception throw in EntryCountEstimator causes dispatcher to stop dispatching messages
### Search before reporting

### Read release policy

### User environment

3.0.14

### **Issue Summary**

After upgrading to Pulsar 3.0.14, we observed `NoSuchElementException` errors in the broker logs that cause the dispatcher to stop dispatching messages to consumers.

### **Error Log**

```
2026-01-19T07:50:56.200Z [broker-topic-workers-OrderedExecutor-10-0] ERROR org.apache.bookkeeper.common.util.SingleThreadExecutor - Error while running task: null
java.util.NoSuchElementException
	at java.base/java.util.concurrent.ConcurrentSkipListMap.firstKey(ConcurrentSkipListMap.java:1859)
	at org.apache.bookkeeper.mledger.impl.EntryCountEstimator.internalEstimateEntryCountByBytesSize(EntryCountEstimator.java:93)
	at org.apache.bookkeeper.mledger.impl.EntryCountEstimator.estimateEntryCountByBytesSize(EntryCountEstimator.java:51)
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.applyMaxSizeCap(ManagedCursorImpl.java:3849)
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.asyncReadEntriesWithSkipOrWait(ManagedCursorImpl.java:1051)
	at org.apache.bookkeeper.mledger.impl.ManagedCursorImpl.asyncReadEntriesOrWait(ManagedCursorImpl.java:1031)
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.readMoreEntries(PersistentDispatcherMultipleConsumers.java:371)
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.lambda$readEntriesComplete$8(PersistentDispatcherMultipleConsumers.java:628)
	at org.apache.bookkeeper.common.util.SingleThreadExecutor.safeRunTask(SingleThreadExecutor.java:137)
	at org.apache.bookkeeper.common.util.SingleThreadExecutor.run(SingleThreadExecutor.java:113)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

### **Root Cause Analysis**

The bug is in `EntryCountEstimator.internalEstimateEntryCountByBytesSize()` at line 93:

```java
} else if (readPosition.getLedgerId() < ledgersInfo.firstKey()) {
    readPosition = PositionImpl.get(ledgersInfo.firstKey(), 0);
}
```

The code calls `ledgersInfo.firstKey()` on a `ConcurrentSkipListMap` without first checking if the map is empty. When `ledgersInfo` is empty (which can occur due to race conditions during ledger trimming/compaction), `firstKey()` throws `NoSuchElementException`.

Similarly, line 89 also has a potential issue:
```java
} else if (lastLedgerId == null && readPosition.getLedgerId() > ledgersInfo.lastKey()) {
```

Maybe safer alternatives like `firstEntry()` with null checks should be used.

### **Impact**

- The uncaught exception causes the dispatcher to stop dispatching messages to consumers
- The subscription appears stuck with `subscriptionHavePendingRead: true` but `waitingReadOp: false`
- Consumers will not receive any new messages until the broker/topic is restarted/unloaded.

### **Managed Ledger Internal Stats**

This shows the state of an affected topic. Note that the cursor's `readPosition` (20144233:0) references a ledger that no longer exists in the ledgers list (ledgers start from 20223153):

```json
{
  "entriesAddedCounter": 3,
  "numberOfEntries": 3,
  "totalSize": 1598,
  "lastConfirmedEntry": "20227219:0",
  "state": "LedgerOpened",
  "ledgers": [
    {"ledgerId": 20223153, "entries": 1, "size": 533},
    {"ledgerId": 20227219, "entries": 1, "size": 532},
    {"ledgerId": 20251815, "entries": 0, "size": 0}
  ],
  "cursors": {
    "<subname>": {
      "markDeletePosition": "19058261:1",
      "readPosition": "20144233:0",
      "waitingReadOp": false,
      "subscriptionHavePendingRead": true,
      "subscriptionHavePendingReplayRead": false,
      "active": true
    }
  }
}
```

### Error messages

```text

```

### Reproducing the issue

I don't have a stable reproducer yet.

### Additional information

_No response_

lhotari:
Fixes #25174

### Motivation

An NoSuchElementException could occur in EntryCountEstimator, causing dispatching to stop.
This condition could happen as a result of a race condition caused by ledger trimming. The ledgersInfo map appear to be be empty.

### Modifications

- check if ledgersInfo map is empty
- catch NoSuchElementException when firstKey or lastKey is accessed to avoid race conditions after checking if the map was empty
- add tests

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimator.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimator.java
@@ -192,13 +192,17 @@ class EntryCountEstimator {
         // to the first available ledger or the last active ledger, respectively.
         if (lastLedgerId != null && readPosition.getLedgerId() > lastLedgerId.longValue()) {
             return PositionFactory.create(lastLedgerId, Math.max(lastLedgerTotalEntries - 1, 0));
-        } else if (lastLedgerId == null && readPosition.getLedgerId() > ledgersInfo.lastKey()) {
+        }
+        long lastKey = ledgersInfo.lastKey();
+        if (lastLedgerId == null && readPosition.getLedgerId() > lastKey) {
             Map.Entry<Long, MLDataFormats.ManagedLedgerInfo.LedgerInfo> lastEntry = ledgersInfo.lastEntry();
-            if (lastEntry != null) {
+            if (lastEntry != null && lastEntry.getKey() == lastKey) {
                 return PositionFactory.create(lastEntry.getKey(), Math.max(lastEntry.getValue().getEntries() - 1, 0));
             }
-        } else if (readPosition.getLedgerId() < ledgersInfo.firstKey()) {
-            return PositionFactory.create(ledgersInfo.firstKey(), 0);
+        }
+        long firstKey = ledgersInfo.firstKey();
+        if (readPosition.getLedgerId() < firstKey) {
+            return PositionFactory.create(firstKey, 0);
         }
         return readPosition;
     }
--- a/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimatorTest.java
+++ b/managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimatorTest.java
@@ -310,6 +310,7 @@ public class EntryCountEstimatorTest {
         ledgersInfo = mock(NavigableMap.class);
         when(ledgersInfo.isEmpty()).thenReturn(false);
         when(ledgersInfo.firstKey()).thenThrow(NoSuchElementException.class);
+        when(ledgersInfo.lastKey()).thenReturn(1L);
         int result = estimateEntryCountByBytesSize(5_000_000);
         // expect that result is 1 because the estimation couldn't be done
         assertEquals(result, 1);


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimator.java
```diff
@@ -183,4 +182,24 @@ static int internalEstimateEntryCountByBytesSize(int maxEntries, long maxSizeByt
...
+                                               NavigableMap<Long, MLDataFormats.ManagedLedgerInfo.LedgerInfo>
+                                                       ledgersInfo,
+                                               Long lastLedgerId, long lastLedgerTotalEntries) {
+        // Adjust the read position to ensure it falls within the valid range of available ledgers.
+        // This handles special cases such as EARLIEST and LATEST positions by resetting them
+        // to the first available ledger or the last active ledger, respectively.
+        if (lastLedgerId != null && readPosition.getLedgerId() > lastLedgerId.longValue()) {
+            return PositionFactory.create(lastLedgerId, Math.max(lastLedgerTotalEntries - 1, 0));
+        } else if (lastLedgerId == null && readPosition.getLedgerId() > ledgersInfo.lastKey()) {
+            Map.Entry<Long, MLDataFormats.ManagedLedgerInfo.LedgerInfo> lastEntry = ledgersInfo.lastEntry();
```
Shawyeok, line 198: The method call `ledgersInfo.lastEntry()` and `ledgersInfo.lastKey()` in the previous line may return different ledgers due to race condition.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimator.java
```diff
@@ -183,4 +182,24 @@ static int internalEstimateEntryCountByBytesSize(int maxEntries, long maxSizeByt
...
+            if (lastEntry != null) {
+                return PositionFactory.create(lastEntry.getKey(), Math.max(lastEntry.getValue().getEntries() - 1, 0));
+            }
+        } else if (readPosition.getLedgerId() < ledgersInfo.firstKey()) {
+            return PositionFactory.create(ledgersInfo.firstKey(), 0);
```
Shawyeok: Same here, those two `ledgersInfo.firstKey()` method calls may return different ledger, I think we can reuse the method call result in a local variable and switch to safe alternate `firstEntry()` and `lastEntry()` with nullness check.

Shawyeok:
```java
var firstEntry = ledgersInfo.firstEntry();
...
} else if (firstEntry != null && readPosition.getLedgerId() < firstEntry.getKey()) {
    return PositionFactory.create(firstEntry.getKey(), 0);
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimator.java
Shawyeok:
```
var lastEntry = ledgersInfo.lastEntry();
...
} else if (lastLedgerId == null && lastEntry != null && readPosition.getLedgerId() > lastEntry.getKey()) {
~~    Map.Entry<Long, MLDataFormats.ManagedLedgerInfo.LedgerInfo> lastEntry = ledgersInfo.lastEntry();~~
```

lhotari: addressed

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryCountEstimator.java
lhotari: addressed

PR #25178, opened 2026-01-23
[fix][client] ControlledClusterFailover avoid unnecessary reconnection.

thetumbled:
### Motivation

In `ControlledClusterFailover`, the `currentControlledConfiguration` field is not initialized when the class is constructed. This causes the first execution of the scheduled task to always trigger an unnecessary reconnection, even when the fetched configuration from the URL provider is identical to the client's current configuration.

The condition check:
```java
if (controlledConfiguration != null
        && !Strings.isNullOrEmpty(controlledConfiguration.getServiceUrl())
        && !controlledConfiguration.equals(currentControlledConfiguration))
```

When `currentControlledConfiguration` is `null`, `!controlledConfiguration.equals(null)` always returns `true`, causing:
1. `pulsarClient.updateServiceUrl(serviceUrl)` to be called
2. `pulsarClient.reloadLookUp()` to be called - which **disconnects all existing connections and reconnects**

This unnecessary reconnection on client startup can cause connection disruptions and performance issues.

I reproduce this issue in my test enviroment:
```
12:06:04.617 [pulsar-service-provider-1-1] INFO  org.apache.pulsar.client.impl.ControlledClusterFailover - Switch Pulsar service url from null to ControlledClusterFailover.ControlledConfiguration(serviceUrl=pulsar://XXX:6650, tlsTrustCertsFilePath=, authPluginClassName=org.apache.pulsar.client.impl.auth.AuthenticationToken, authParamsString=XXX)
```

### Modifications

Initialize `currentControlledConfiguration` in the `initialize()` method with all four configuration fields from the client's current configuration:

- `serviceUrl` - from `defaultServiceUrl` passed to the builder
- `tlsTrustCertsFilePath` - from `ClientConfigurationData`
- `authPluginClassName` - from `ClientConfigurationData`
- `authParamsString` - from `ClientConfigurationData.authParams`

This ensures that on the first scheduled check, if the fetched configuration matches the client's current configuration, no unnecessary reconnection will occur.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/thetumbled/pulsar/pull/76

### DIFF
--- a/pulsar-client/src/test/java/org/apache/pulsar/client/impl/ControlledClusterFailoverTest.java
+++ b/pulsar-client/src/test/java/org/apache/pulsar/client/impl/ControlledClusterFailoverTest.java
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 import lombok.Cleanup;
 import org.apache.pulsar.client.api.Authentication;
 import org.apache.pulsar.client.api.ServiceUrlProvider;
+import org.apache.pulsar.client.impl.conf.ClientConfigurationData;
 import org.asynchttpclient.Request;
 import org.awaitility.Awaitility;
 import org.mockito.Mockito;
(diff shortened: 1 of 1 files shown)


thetumbled:
PTAL, thanks. @hangc0276 @lhotari @BewareMyPower @dao-jun @nodece

Issue #25071, opened 2025-12-12
Update logic behind AuthenticateOriginalAuthData to better support multiple authentication Providers


shasank112001:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Currently if you utilize Pulsar Proxies, then to properly support JWT authentication and client disconnects after token expiry, one must enable authenticateOriginalAuthData on brokers.
If disabled, then connected clients with expired tokens can still produce/consume on topics as discussed in this issue #25062.
However, authenticateOriginalAuthData is not supported by Authentication methods like AuthenticationProviderTLS as the proxy performs a reencrypt of the TLS  connection and therefore, authenticating the original principal always leads to the proxy identity which is not useful when authorizing.

### Solution

1. AuthenticateOriginalAuthData can be made as a property of the AuthenticationProvider itself. With JWTs the original auth data must be authenticated, as otherwise connections will be kept open after token expiry. Therefore, it makes sense to include the property on the provider itself.

3. Terminate connections on Proxy if a token expires: I am not familiar with all the intricacies of how the proxy closes connections, but I do know that the Proxy does send a Disconnect to the client when it asserts the token to have expired. Therefore, if we can add a hook to the disconnect command, to close all proxied connections, we can get rid of authenticateOriginalAuthData completely.

### Anything else?

I am willing to work on this, however if solution 2 is possible, I would like a bit of help understanding the logic behind the proxy connections.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

nodece:
@shasank112001 We're focusing specifically on the case where **both** `forwardAuthorizationCredentials` and `authenticateOriginalAuthData` are set to `true`. 

The issue is:  **if either of these is disabled, the broker won't properly handle token expiration**, meaning expired client connections won't be closed as expected.

The current proxy implementation: 

https://github.com/apache/pulsar/blob/1050f48035c97a14094b812467f046f6d37ff6f4/pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/ProxyConnection.java#L522-L525

Which assumes the broker will always validate token expiration.  However: 

- **When `authenticateOriginalAuthData=false`:** The broker only validates the proxy's credentials, not the original client's token
- **Result:** Expired client tokens slip through undetected, and stale connections persist

A workaround: So both settings must be enabled together for proper token expiration handling through the broker and proxy.

I will open a PR to fix this issue.

PR #25179, opened 2026-01-23
[fix][proxy] Close client connection immediately when credentials expire and forwardAuthorizationCredentials is disabled

nodece:
Fixes https://github.com/apache/pulsar/issues/25071

### Motivation

This pull request enhances the authentication credential refresh logic in the Pulsar Proxy and adds comprehensive tests to ensure correct behavior when client credentials expire, especially when authorization credentials are not forwarded to the broker. The changes improve reliability and correctness in handling authentication expiration and provide a robust test for the new logic.

### Modifications

**Proxy authentication refresh and expiration logic:**

* Refactored the authentication credential refresh logic in `ProxyConnection.java` by introducing the `startAuthRefreshTaskIfNotStarted()` helper, ensuring the refresh task is only scheduled when needed and in the correct states. The refresh logic now checks if the connection is in an authenticated state and properly handles cases where credentials expire and authorization credentials are not forwarded (`forwardAuthorizationCredentials=false`). [[1]](diffhunk://#diff-bbca5aac9dede7618187e91b91ffd0c6c8ffb836cd79ff2d104439e8cf5fc0daL152-R157) [[2]](diffhunk://#diff-bbca5aac9dede7618187e91b91ffd0c6c8ffb836cd79ff2d104439e8cf5fc0daL415-R420) [[3]](diffhunk://#diff-bbca5aac9dede7618187e91b91ffd0c6c8ffb836cd79ff2d104439e8cf5fc0daR436-R439) [[4]](diffhunk://#diff-bbca5aac9dede7618187e91b91ffd0c6c8ffb836cd79ff2d104439e8cf5fc0daR521-R558)

**Test suite improvements:**

* Added a new test `testClientDisconnectWhenCredentialsExpireWithoutForwardAuth` in `ProxyAuthenticationTest.java` to verify that the client is disconnected when credentials expire and the proxy is not forwarding authorization credentials. The test uses Awaitility to assert proper disconnection behavior.
* Updated and expanded the mock authentication provider and state in the test suite to simulate credential expiration and support the new authentication flow, including parsing and checking expiration within the mock state. [[1]](diffhunk://#diff-e8bd8b62a59dff014e3084ab83180fcdcc8a81f637a3373c51ca49b634be22eaR134-R198) [[2]](diffhunk://#diff-e8bd8b62a59dff014e3084ab83180fcdcc8a81f637a3373c51ca49b634be22eaL138-R222)

**Minor test utility and import updates:**

* Added necessary imports and updated HTTP header handling in the test authentication data provider for improved correctness and clarity. [[1]](diffhunk://#diff-e8bd8b62a59dff014e3084ab83180fcdcc8a81f637a3373c51ca49b634be22eaR21-R46) [[2]](diffhunk://#diff-e8bd8b62a59dff014e3084ab83180fcdcc8a81f637a3373c51ca49b634be22eaR55-R59) [[3]](diffhunk://#diff-e8bd8b62a59dff014e3084ab83180fcdcc8a81f637a3373c51ca49b634be22eaR94)

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-proxy/src/test/java/org/apache/pulsar/proxy/server/ProxyRefreshAuthTest.java
+++ b/pulsar-proxy/src/test/java/org/apache/pulsar/proxy/server/ProxyRefreshAuthTest.java
@@ -179,37 +179,25 @@ public class ProxyRefreshAuthTest extends ProducerConsumerBase {
 
         PulsarClientImpl pulsarClientImpl = (PulsarClientImpl) pulsarClient;
         pulsarClient.getPartitionsForTopic(topic).get();
+
+        // Verify initial connection state
         Set<CompletableFuture<ClientCnx>> connections = pulsarClientImpl.getCnxPool().getConnections();
 
-        Awaitility.await().during(5, SECONDS).untilAsserted(() -> {
-            pulsarClient.getPartitionsForTopic(topic).get();
-            assertTrue(connections.stream().allMatch(n -> {
-                try {
-                    ClientCnx clientCnx = n.get();
-                    long timestamp = clientCnx.getLastDisconnectedTimestamp();
-                    return timestamp == 0;
-                } catch (Exception e) {
-                    throw new RuntimeException(e);
-                }
-            }));
-        });
-
-        // Force all connections from proxy to broker to close and therefore require the proxy to re-authenticate with
-        // the broker. (The client doesn't lose this connection.)
-        restartBroker();
-
-        // Rerun assertion to ensure that it still works
-        Awaitility.await().during(5, SECONDS).untilAsserted(() -> {
-            pulsarClient.getPartitionsForTopic(topic).get();
-            assertTrue(connections.stream().allMatch(n -> {
-                try {
-                    ClientCnx clientCnx = n.get();
-                    long timestamp = clientCnx.getLastDisconnectedTimestamp();
-                    return timestamp == 0;
-                } catch (Exception e) {
-                    throw new RuntimeException(e);
-                }
-            }));
-        });
+        Awaitility.await()
+                .during(5, SECONDS)
+                .untilAsserted(() -> {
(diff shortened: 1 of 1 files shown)


pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/ProxyConnection.java
```diff
@@ -436,6 +433,10 @@ private void handleBrokerConnected(DirectProxyHandler directProxyHandler, Comman
             final ByteBuf msg = Commands.newConnected(connected.getProtocolVersion(), maxMessageSize,
                     connected.hasFeatureFlags() && connected.getFeatureFlags().isSupportsTopicWatchers());
             writeAndFlush(msg);
+            // Start auth refresh task only if we are not forwarding authorization credentials
+            if (!service.getConfiguration().isForwardAuthorizationCredentials()) {
+                startAuthRefreshTaskIfNotStarted();
+            }
```
lhotari, line 439: It is misleading to call this "auth refresh task" in this case. 

There's a clear reason why the auth refresh task has been applied only for lookup connections since when the the connection is in `ProxyConnectionToBroker` state, the proxy doesn't intercept the Pulsar protocol messages.

It seems that this added logic would disconnect the connection after `authenticationRefreshCheckSeconds`. It would be better to reflect this in the naming of methods and not reuse the existing auth refresh logic at all.

lhotari: I might take this back. It seems that this change makes sense. now noticed `if (!authState.isExpired()) {` and if the authorization credentials aren't forwarded, this makes sense. resolving this comment.

Issue #25184, opened 2026-01-26
Add configuration option to require explicit authentication method specification


omarkj:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

Currently, when a client connects to a Pulsar broker without specifying an authentication method (via the X-Pulsar-Auth-Method-Name header in HTTP or the auth method field in the Pulsar protocol), the broker attempts to authenticate using all configured authentication providers sequentially until one succeeds or all fail.

While this fallback behaviour provides flexibility, it can create security concerns in deployments where operators want to enforce that clients must explicitly specify which authentication method they're using. This explicit requirement helps prevent:
1. Unintended authentication against the wrong provider
2. Ambiguity about which authentication mechanism was actually used
3. Potential security issues where fallback behaviour differs from operator expectations

### Solution

Add a new broker configuration option `strictAuthMethod` (default: `false`) that, when enabled, requires clients to explicitly specify their authentication method. When enabled:
- Authentication requests without an explicit method will immediately fail with an "Authentication method missing" error
- Clients must specify which authentication provider they intend to use
- No fallback to trying multiple providers will occur

This provides operators with better control over authentication behaviour in security-sensitive deployments while maintaining backward compatibility through the default false value.

### Alternatives

No, this is an additive feature that doesn't change existing behaviour when disabled.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #25185, opened 2026-01-26
[improve][broker] Add strictAuthMethod to require explicit authentication method

omarkj:
Add a new `strictAuthMethod` configuration option to enforce that clients must specify their authentication method. When enabled, authentication requests without an explicit method will immediately fail with "Authentication method missing" instead of attempting to authenticate with all configured providers.

This improves security by preventing unintended authentication fallback behaviour when operators want to require specific authentication methods for their deployment.

Main Issue: #25184

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -1796,6 +1796,14 @@ public class ServiceConfiguration implements PulsarConfiguration {
         doc = "Enable authentication"
     )
     private boolean authenticationEnabled = false;
+
+    @FieldContext(
+        category = CATEGORY_AUTHENTICATION,
+        doc = "Strictly enforce authentication method. If specified, Pulsar will only attempt to authenticate with "
+                + "the provided method. If no method is provided, authentication fails."
+    )
+    private boolean strictAuthMethod = false;
+
     @FieldContext(
         category = CATEGORY_AUTHENTICATION,
         doc = "Authentication provider name list, which is a list of class names"
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authentication/AuthenticationService.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/authentication/AuthenticationService.java
@@ -47,6 +47,7 @@ import org.slf4j.LoggerFactory;
 public class AuthenticationService implements Closeable {
     private static final Logger LOG = LoggerFactory.getLogger(AuthenticationService.class);
     private final String anonymousUserRole;
+    private final boolean strictAuthMethod;
 
     private final Map<String, AuthenticationProvider> providers = new LinkedHashMap<>();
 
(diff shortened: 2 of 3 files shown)


lhotari:
Thanks for the contribution @omarkj. Yes, a PIP would be useful in documenting the change.

Some questions:
1. does this require also a similar change for 
  - the Pulsar binary protocol?
  - the Pulsar Proxy?

2. Since the client can specify the authentication method, how does this improve security?
- Just wondering if it be useful to limit authentication methods for a particular "advertised listener"?
  - Pulsar broker can bind to multiple addresses/ports and limiting the authentication method for an external port binding could be an useful addition

omarkj:
I will get a PIP out if relevant.

@lhotari Regarding your questions:

> 1. does this require also a similar change for
> * the Pulsar binary protocol?

I believe the Pulsar Protocol authentication only tries the specified authentication provider unlike the HTTP authentication which iterates through all available providers if none is specified.

> * the Pulsar Proxy?

This is a good point, if the proxy is authenticating HTTP requests then yes, might be useful there.

> 2. Since the client can specify the authentication method, how does this improve security?

This should make it harder to probe the system for possibly misconfigured authentication providers.

lhotari:
> I will get a PIP out if relevant.
> 
> @lhotari Regarding your questions:
> 
> > 1. does this require also a similar change for
> > 
> > 
> > * the Pulsar binary protocol?
> 
> I believe the Pulsar Protocol authentication only tries the specified authentication provider unlike the HTTP authentication which iterates through all available providers if none is specified.

Oh yes, you are correct.

> 
> > * the Pulsar Proxy?
> 
> This is a good point, if the proxy is authenticating HTTP requests then yes, might be useful there.
> 
> > 2. Since the client can specify the authentication method, how does this improve security?
> 
> This should make it harder to probe the system for possibly misconfigured authentication providers.

Makes sense

lhotari (approved):
LGTM, thanks for the contribution!

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -1796,6 +1796,14 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
         doc = "Enable authentication"
     )
     private boolean authenticationEnabled = false;
+
```
lhotari, line 1799: Please also add this new configuration key to `conf/broker.conf` and `conf/standalone.conf` files.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -1796,6 +1796,14 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
...
+    @FieldContext(
+        category = CATEGORY_AUTHENTICATION,
+        doc = "Strictly enforce authentication method. If specified, Pulsar will only attempt to authenticate with "
+                + "the provided method. If no method is provided, authentication fails."
```
lhotari, line 1803: Perhaps this should be more specific about this being applied to HTTP authentication and requiring `X-Pulsar-Auth-Method-Name` header to be set when this setting is enabled.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari: I'll merge now so that this makes it to the next set of releases. This could be addressed in a follow up PR.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
lhotari: I'll merge now so that this makes it to the next set of releases. This could be addressed in a follow up PR.

PR #25189, opened 2026-01-27
[improve][broker] Add http produce backlog quota check

Dream95:
### Motivation

The binary protocol already performs check backlog quota in `ServerCnx` when creating producers, but the HTTP REST API path was missing this validation.

~~~Java
      .thenAccept(t -> {
            // TODO: Check message backlog and fail if backlog too large.
            if (!t.isPresent()) {
                // Topic not found, and remove from owning partition list.
                publishResult.completeExceptionally(new BrokerServiceException.TopicNotFoundException("Topic not "
                        + "owned by current broker."));
                TopicName topicName = TopicName.get(topic);
                pulsar().getBrokerService().getOwningTopics().get(topicName.getPartitionedTopicName())
                        .remove(topicName.getPartitionIndex());
~~~
### Modifications

- Added backlog quota checks in `publishSingleMessageToPartition` method for both `destination_storage` and `message_age` quota types

The implementation follows the same pattern used in `ServerCnx.handleProducer` where backlog quota checks are performed before allowing message production.

### Verifying this change

- Existing backlog quota tests should cover this functionality
- HTTP produce endpoint tests should verify the behavior

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [x] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

**Changes to REST endpoints:**
- The HTTP produce endpoint (`/persistent/{tenant}/{namespace}/{topic}/partitions/{partition}`) now properly enforces backlog quota limits before publishing messages. When backlog quota is exceeded, the request will fail with an appropriate error, consistent with the binary protocol behavior.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/Dream95/pulsar/pull/4

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/rest/TopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/rest/TopicsBase.java
@@ -87,6 +87,7 @@ import org.apache.pulsar.common.api.proto.MessageMetadata;
 import org.apache.pulsar.common.compression.CompressionCodecProvider;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.partition.PartitionedTopicMetadata;
+import org.apache.pulsar.common.policies.data.BacklogQuota;
 import org.apache.pulsar.common.policies.data.TopicOperation;
 import org.apache.pulsar.common.protocol.Commands;
 import org.apache.pulsar.common.protocol.schema.SchemaData;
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/rest/TopicsBase.java
```diff
@@ -272,35 +273,41 @@ private void internalPublishMessages(TopicName topicName, ProducerMessages reque
...
+                        pulsar().getBrokerService().getOwningTopics().get(tn.getPartitionedTopicName())
+                                .remove(tn.getPartitionIndex());
+                        return CompletableFuture.completedFuture(null);
+                    }
+                    Topic topicObj = tOpt.get();
+                    CompletableFuture<Void> backlogQuotaCheckFuture = CompletableFuture.allOf(
+                            topicObj.checkBacklogQuotaExceeded(defaultProducerName,
+                                    BacklogQuota.BacklogQuotaType.destination_storage),
+                            topicObj.checkBacklogQuotaExceeded(defaultProducerName,
+                                    BacklogQuota.BacklogQuotaType.message_age));
```
Denovo1998: Should we pass in message.getProducerName(); here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/rest/TopicsBase.java
```diff
@@ -272,35 +273,41 @@ private void internalPublishMessages(TopicName topicName, ProducerMessages reque
...
+                })
+                .exceptionally(ex -> {
+                    if (!publishResult.isDone()) {
+                        publishResult.completeExceptionally(FutureUtil.unwrapCompletionException(ex));
                     }
-                } catch (Exception e) {
                     if (log.isDebugEnabled()) {
-                        log.debug("Fail to publish single messages to topic  {}: {} ",
-                                topicName, e.getCause());
+                        log.debug("Fail to publish single message to topic {}: {}", topic, ex.getMessage());
```
Denovo1998: Here you can first use FutureUtil.unwrapCompletionException.

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/TopicsTest.java
```diff
@@ -931,4 +933,80 @@ public void testProduceWithAutoConsumeSchema() throws Exception {
...
+        admin.namespaces().setBacklogQuota(namespaceName, backlogQuota,
+                BacklogQuota.BacklogQuotaType.message_age);
+        admin.topics().createSubscription(topicName, "time-quota-sub", MessageId.earliest);
+
+        @Cleanup
+        Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
+                .topic(topicName)
+                .create();
+        producer.send("backlog-message");
+        Thread.sleep(2000);
```
Denovo1998: Perhaps it can be replaced with a "poll until quota exceeded (with timeout)" method.

Denovo1998 (changes requested):
Now it does a backlog quota check for each message, maybe it can be optimized?

1. In internalPublishMessages(...), first calculate which partition(s) the current request will write to (actually, you have already calculated in round-robin which partition each message goes to).
2. Perform once for each partition:
- getTopic(partitionTopic)
- checkBacklogQuotaExceeded(...destination_storage + message_age)
3. After all checks pass, then enter the actual publish loop
- If the check for a certain partition fails: mark all the message results corresponding to that partition as failed (no longer attempt to publish), other partitions can continue.

Dream95:
@Denovo1998 Thank you for your review. I have solved the problems you mentioned.Regarding the optimiz,per-message checks are more precise？

PR #24617, opened 2025-08-09
[improve][broker] Upgrade avro version to 1.12.0

hangc0276:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Motivation

Upgrade the Avro version.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #25193, opened 2026-01-29
[fix][broker] Use compatible Avro name validator to allow '$' in schema record names

mattisonchao:
### Motivation

After #24617 upgraded Avro to a newer version, the default `Schema.Parser()` 
uses `UTF_VALIDATOR` which rejects the `$` character in record names. This 
breaks Protobuf schemas whose Avro representation contains `$` in generated 
nested type names (e.g. inner classes).

### Modifications

- Introduce a `CompatibleNameValidator` that allows `$` in addition to letters, 
  digits, and underscores, matching the previous Avro behavior.
- Apply the custom validator to all broker-side `Schema.Parser()` instances 
  that handle Protobuf schemas:
  - `StructSchemaDataValidator`
  - `SchemaRegistryServiceImpl`
  - `AvroSchemaBasedCompatibilityCheck`
- Add `DataRecord.proto` for test reproduction.
- Add unit tests for the `CompatibleNameValidator` and Protobuf schema 
  compatibility.

> **Note:** The client-side `SchemaUtil` already uses `NameValidator.NO_VALIDATION`. 
> A shared solution across broker and client (e.g. factory in `pulsar-common`) 
> can be addressed in a follow-up.

### Verifying this change

- Unit tests added for `CompatibleNameValidator` (valid names, invalid names, 
  edge cases, error messages).
- Integration test with `ProtobufSchema<DataRecord>` to reproduce the original issue.
- CI checks pass.

### Documentation

- [x] `doc-not-needed`


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/schema/validator/StructSchemaDataValidator.java
```diff
@@ -49,7 +52,7 @@ public void validate(SchemaData schemaData) throws InvalidSchemaDataException {
         byte[] data = schemaData.getData();
 
         try {
-            Schema.Parser avroSchemaParser = new Schema.Parser();
+            Schema.Parser avroSchemaParser = new Schema.Parser(COMPATIBLE_NAME_VALIDATOR);
```
lhotari, line 54: The Pulsar code base includes `new Schema.Parser()` in many locations. Wouldn't we have to use the custom name validator in all cases?

mattisonchao: Yes, I'll change them also.

lhotari: Some factory class to `pulsar-common`? There's no avro dependency currently, but I guess it could justified.

mattisonchao: after the evaluation. I think this PR should focus solely on fixing the issue and eliminating breaking changes in the broker versions. To align the usage. We could try another PR on the client side to support it.

lhotari: `pulsar-common` is also used on broker side, so how could it be separated? I would assume that it would make sense to cover both client + broker to use the `COMPATIBLE_NAME_VALIDATOR` in the avro schema parser in all locations. It should be a task that Claude Code could handle without much intervention.

lhotari: Oh yes, `pulsar-common` doesn't currently depend on avro libraries (which could be a useful thing) so it would require a solution to share the name validator across broker and client.

lhotari: I think it's fine to handle broker-side first. Please rename the PR title so that it captures this detail.

mattisonchao: applied.

pulsar-broker/src/test/proto/DataRecord.proto
lhotari, line 1: is this file for tests? I think it should be in `pulsar-broker/src/test/proto` directory in that case.

mattisonchao: ah. you are right...  Let me fix it now.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/schema/AvroSchemaBasedCompatibilityCheck.java
```diff
@@ -51,11 +52,12 @@ public void checkCompatible(Iterable<SchemaData> from, SchemaData to, SchemaComp
         checkArgument(from != null, "check compatibility list is null");
         try {
             for (SchemaData schemaData : from) {
-                Schema.Parser parser = new Schema.Parser();
+                Schema.Parser parser =
+                        new Schema.Parser(StructSchemaDataValidator.COMPATIBLE_NAME_VALIDATOR);
```
codelipenghui, line 56: Please also check the JsonSchemaCompatibilityCheck.java which still use `Schema.Parser fromParser = new Schema.Parser();` in isAvroSchema check.

```
    private boolean isAvroSchema(SchemaData schemaData) {
        try {

            Schema.Parser fromParser = new Schema.Parser();
            fromParser.setValidateDefaults(false);
            Schema fromSchema = fromParser.parse(new String(schemaData.getData(), UTF_8));
            return true;
        } catch (Exception e) {
            return false;
        }
    }
```

mattisonchao: Good catch!  I'll fix it too. :))

PR #25194, opened 2026-01-29
[improve][build] Build docker image only once when pushing

nodece:
### Motivation

When pushing images to Docker Hub, the `docker-maven-plugin` currently builds the image twice: once during the `build` goal and again during the `push` goal. This results in redundant work and unnecessary build time.

This change makes the Docker goal configurable so that the image is built only once when pushing.

### Modifications

* Replaced the hardcoded Docker goals (`build` and `push`) in the default execution of the Docker Maven plugin with the `${docker.goal}` property in both `docker/pulsar/pom.xml` and `docker/pulsar-all/pom.xml`, allowing the active goal to be configured.
* Set the default value of the `docker.goal` property to `build` in the root `pom.xml`, preserving the existing behavior for normal builds.
* Added the `docker.goal=push` property to the `docker-push` Maven profile in both `docker/pulsar/pom.xml` and `docker/pulsar-all/pom.xml`, so activating this profile pushes the image without triggering an extra build.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docker/pulsar-all/pom.xml
+++ b/docker/pulsar-all/pom.xml
@@ -152,8 +152,7 @@
                 <id>default</id>
                 <phase>package</phase>
                 <goals>
-                  <goal>build</goal>
-                  <goal>push</goal>
+                  <goal>${docker.goal}</goal>
                 </goals>
                 <configuration>
                   <images>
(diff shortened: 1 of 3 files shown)


nodece:
The `push` goal depends on the `build` goal, so this PR should be reverted.

PR #25196, opened 2026-01-29
[feat][pip] PIP-454: Metadata Store Migration Framework

merlimat:
### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-454.md
+++ b/pip/pip-454.md
@@ -39,7 +39,7 @@ Provide a safe, automated framework for migrating Apache Pulsar's metadata from 
 - Developing new metadata store implementations (Oxia, Etcd support already exists)
 - Cross-cluster metadata synchronization (different use case)
 - Automated rollback after COMPLETED phase (requires manual intervention)
-- Migration of configuration metadata store (can be done separately)
+- Migration of configuration metadata store and geo-replicated clusters (can be done separately)
 
 ## High Level Design
 


pip/pip-454.md
```diff
@@ -0,0 +1,416 @@
...
+   ```json
+   {
+     "phase": "PREPARATION",
+     "targetUrl": "oxia://oxia1:6648"
+   }
+   ```
+
+**Broker/Bookie actions (automatic, triggered by watching the flag):**
+1. Detect migration flag via watch on `/pulsar/migration-coordinator/migration`
+2. Defer non-critical metadata writes (e.g., ledger rollovers, bundle ownership changes)
```
lhotari, line 111: would this use the existing solution? a `SessionEvent.ConnectionLost`/`SessionEvent.SessionLost` event sets a flag `metadataServiceAvailable` that is used for this purpose in many locations. 

https://github.com/apache/pulsar/blob/d630394cdd02792b2dbc3a55443637a5d593a137/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java#L148-L152

https://github.com/apache/pulsar/blob/1617bb22173a117f24d47ac6f11cc2f7c68de635/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerFactoryImpl.java#L288-L291

It seems that currently ledger trimming, ledger rollover and loadbalancer load shedding are using the the `metadataServiceAvailable` flag in `ManagedLedgerFactoryImpl`.

There's also a dependency on the event directly:
https://github.com/apache/pulsar/blob/38807b1511ba3b8c150d69c16a0c3ae36f321dac/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/ModularLoadManagerImpl.java#L1137-L1150

**would the coordinator send a `SessionEvent.ConnectionLost` event when migration starts so that it remains compatible with the existing solution?**

`AbstractMetadataStore` has a flag `isConnected` which could also be useful? it's not currently used within Pulsar, just for metadata store caching decisions. I guess it would be necessary to skip cache refreshs while the migration is on-going.

pip/pip-454.md
```diff
@@ -0,0 +1,416 @@
...
+(If errors occur):
+FAILED ← Rollback to source store, writes ENABLED
+```
+
+### Phase 1: NOT_STARTED → PREPARATION
+
+**Participant Registration (at startup):**
+Each broker and bookie registers itself as a migration participant by creating a sequential ephemeral node:
+- Path: `/pulsar/migration-coordinator/participants/id-NNNN` (sequential)
+- This allows the coordinator to know how many participants exist before migration starts
```
lhotari, line 93: Which node is selected as the migration coordinator? 
If it's the broker, what if the migration takes a lot more memory than the broker usually does and causes OOME.
Would it be possible to deploy a dedicated coordinator or run the coordinator in-process, let's say in a pod with sufficient resources, running in a Pulsar cluster?

lhotari: Some additional questions:
Does the same single coordinator instance migrate both Pulsar and Bookkeeper metadata? Why does the coordinator need to know how many participants exist?

lhotari (approved):
LGTM, just a few comments / questions.

merlimat:
Good questions.

> 1. is it possible to also migrate back from Oxia to ZooKeeper ?

While the mechanism can be quite generic, the tricky part to make it work seamlessly is to carry over the same version ids on each metadata record. That allows brokers and bookies conditional writes to not fail in the migration. 

In Oxia we were able to add explicit override for this specific purpose. With ZK it's unfortunately not as easy.

> 2. this PIP is not about the metadata store to coordinate geo replication, should we state it explicitly ?

Good point. I'll highlight this more.

PR #25197, opened 2026-01-30
[fix][misc] Allow JWT tokens in OpenID auth without nbf claim

### LINKED ISSUE #25190: Remove nbf as a required claim
### Search before reporting

### Motivation

When using OIDC as the authentication provider, `nbf` is a required claim as seen in code [here](https://github.com/apache/pulsar/blob/master/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/AuthenticationProviderOpenID.java#L448)

When using Auth0 as a provider, which doesn't return the `nbf` field, it results in OIDC being unusable:
https://community.auth0.com/t/jwt-token-does-not-contain-nbf-claim-again/62350

### Solution

I think the field should be removed or optional.

### Alternatives

I don't know.

### Anything else?

I am not sure what the required claims are based on but according to the [comments](https://github.com/apache/pulsar/blob/master/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/AuthenticationProviderOpenID.java#L442C57-L442C119) above the required claims, it should mirror https://openid.net/specs/openid-connect-basic-1_0.html#IDToken but in that doc, nbf doesn't show up.

I also did find this [Issue ](https://github.com/apache/pulsar/issues/20829)which is similar but instead for allowing `aud` to be optional but it was closed and I couldn't find the relevant changes made.

lhotari:
Fixes #25190

### Motivation

"nbf" (not before) claim isn't mandatory in OpenID auth JWT tokens. However, it's currently required in Pulsar.

### Modifications

- remove "nbf" from mandatory claims
- add unit test

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/AuthenticationProviderOpenID.java
+++ b/pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/AuthenticationProviderOpenID.java
@@ -445,7 +445,6 @@ public class AuthenticationProviderOpenID implements AuthenticationProvider {
                 .withAnyOfAudience(allowedAudiences)
                 .withClaimPresence(RegisteredClaims.ISSUED_AT)
                 .withClaimPresence(RegisteredClaims.EXPIRES_AT)
-                .withClaimPresence(RegisteredClaims.NOT_BEFORE)
                 .withClaimPresence(RegisteredClaims.SUBJECT);
 
         if (isRoleClaimNotSubject) {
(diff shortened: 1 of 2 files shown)


pulsar-broker-auth-oidc/src/main/java/org/apache/pulsar/broker/authentication/oidc/AuthenticationProviderOpenID.java
```diff
@@ -445,7 +445,6 @@ DecodedJWT verifyJWT(PublicKey publicKey,
                 .withAnyOfAudience(allowedAudiences)
                 .withClaimPresence(RegisteredClaims.ISSUED_AT)
                 .withClaimPresence(RegisteredClaims.EXPIRES_AT)
-                .withClaimPresence(RegisteredClaims.NOT_BEFORE)
```
michaeljmarshall, line 448: I verified that https://openid.net/specs/openid-connect-basic-1_0.html#IDToken does not indicate that the `nbf` claim is required and https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5 (the rfc responsible for the json web token spec) indicates the `nbf` claim is optional.

michaeljmarshall (approved):
LGTM

PR #25203, opened 2026-02-03
[revert] "[improve][build] Build docker image only once when pushing"

nodece:
Reverts apache/pulsar#25194

Reason: The `push` goal depends on the `build` goal, so #25194 should be reverted.

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


lhotari:
@nodece Please add an explanation to the description about why this is reverted.

PR #25207, opened 2026-02-03
[fix][client] Fix producer synchronous retry handling in failPendingMessages method

### LINKED ISSUE #25201: [Bug] Producer synchronous retries can cause retry sendAsync future to never complete
### Search before reporting

### Read release policy

### User environment

- master

### Issue Description

There is a reentrancy bug in the Pulsar producer send path where `pendingMessages.clear()` can be executed after a retry message has already been added to `pendingMessages`. This results in the retry send’s CompletableFuture never being completed.

This can occur when a retry `sendAsync` is triggered synchronously from within a handleSync callback of a failed send, while holding the producer mutex.

This happens in the [failPendingMessages](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L2315) method which usually runs on the timer thread.
As the [pendingMessages.clear()](https://github.com/apache/pulsar/blob/d630394cdd02792b2dbc3a55443637a5d593a137/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L2327) is after the completeExceptionally, the retry logic as the code below will add the retryMessage to pendingMessages first and then the clear is called.

```
CompletableFuture<MessageId> firstSend = producer.sendAsync(message);

CompletableFuture<MessageId> retrySend =
                firstSend.handleAsync((msgId, ex) -> {
                    assertNotNull(ex, "First send must timeout");
                    assertTrue(ex instanceof PulsarClientException.TimeoutException);
                    return producer.sendAsync(retryMessage);
                }).thenCompose(f -> f);
```

### Error messages

```text

```

### Reproducing the issue

Set a low timeout value and use synchronous retries as given in the above example.

### Additional information

_No response_

sandeep-mst:
Fixes #25201

### Motivation

Fix a re-entrancy bug in `ProducerImpl.failPendingMessages`. While executing on the timer, `failPendingMessages` invokes `sendComplete(ex)` on pending messages, which can synchronously trigger a retry from client code. The subsequent `pendingMessages.clear()` then removes the newly enqueued retry operation, leaving the retry’s CompletableFuture unresolved and the client in a limbo state.

### Modifications

Updated `failPendingMessages` to first drain the `pendingMessages` queue into a local list and clear the queue before iterating. This prevents re-entrant retries triggered during sendComplete from being inadvertently cleared.

### Verifying this change

  - *org.apache.pulsar.client.impl.ProducerImplTest#testFailPendingMessagesSyncRetry*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: 
https://github.com/cognitree/pulsar/pull/29

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerSyncRetryTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/ProducerSyncRetryTest.java
@@ -71,7 +71,7 @@ public class ProducerSyncRetryTest extends ProducerConsumerBase {
 
         MessageMetadata retryMessageMetadata = new MessageMetadata();
         retryMessageMetadata.setUncompressedSize(1);
-        MessageImpl<byte[]> retryMessage = MessageImpl.create(messageMetadata, payload, Schema.BYTES, topic);
+        MessageImpl<byte[]> retryMessage = MessageImpl.create(retryMessageMetadata, payload, Schema.BYTES, topic);
 
         // First send is expected to fail
         CompletableFuture<MessageId> firstSend = producer.sendAsync(message);


pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -2298,11 +2304,16 @@ public void run(Timeout timeout) throws Exception {
...
      * ProducerImpl object mutex.
      */
-    private synchronized void failPendingMessages(ClientCnx cnx, PulsarClientException ex) {
+    @VisibleForTesting
+    synchronized void failPendingMessages(ClientCnx cnx, PulsarClientException ex) {
         if (cnx == null) {
             final AtomicInteger releaseCount = new AtomicInteger();
             final boolean batchMessagingEnabled = isBatchMessagingEnabled();
-            pendingMessages.forEach(op -> {
+            final ArrayList<OpSendMsg> failingMessages = new ArrayList<>(pendingMessages.size());
```
dlg99: how many pending messages there can be? is this a good idea to allocate new ArrayList all the time?
method is synchronized, so it should be safe to create a static ArrayList of size e.g. 100 and drain min(100, pendingMessages.size()) and process and repeat until pendingMessages.size() == 0

sandeep-mst: At this point, the pending messages should consist only the timed out messages from the producer timeout config. So, it relatively will be low but it makes sense to size it down. Will update the code.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -2298,11 +2304,16 @@ public void run(Timeout timeout) throws Exception {
      * This fails and clears the pending messages with the given exception. This method should be called from within the
      * ProducerImpl object mutex.
      */
-    private synchronized void failPendingMessages(ClientCnx cnx, PulsarClientException ex) {
+    @VisibleForTesting
+    synchronized void failPendingMessages(ClientCnx cnx, PulsarClientException ex) {
```
heesung-sohn, line 2303: do we access pendingMessages without monitor lock? can you share the reentrance code without the monitor lock?

sandeep-mst: We do not access `pendingMessages` without the monitor lock.

Even though, it has monitor lock this method is executed by a timer thread and when [op.sendComplete(ex)](https://github.com/cognitree/pulsar/blob/cab92f0164947e3cbb0da1a014951561f1599add/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L2315) is called within that method it switches to client logic where on the same thread they can send a retry message synchronously (`producer.sendAsync`) which will get enqueued into the pendingMessages [here](https://github.com/apache/pulsar/blob/cab92f0164947e3cbb0da1a014951561f1599add/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L2468). 

So, monitor lock is not preventing the issue from occurring because of reentrancy as it's the same thread holding the lock.

heesung-sohn: would this work, instead of the snapshot array?
int pendingMessageCountToFail = pendingMessages.size();
for(int i =0; i< pendingMessageCountToFail; i++){
 var op = pendingMessages.remove()
  // fail op;
}

Then, we don't need to pendingMessages.clear(); at the line 2327.

sandeep-mst: yes, it would. Doing the same.

lhotari:
great work @sandeep-mst .

Regarding the tests, would it be possible to add a reproducer similar as described in #25201? (test without mocking the client, just using `MockedPulsarServiceBaseTest` base class)

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
sandeep-mst: done

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
sandeep-mst: Went for a different approach altogether. Not creating a new ArrayList.

sandeep-mst:
> LGTM, great work @sandeep-mst . Just a wish to add a test mentioned in the description.

Will add it if possible. That test will go to `pulsar-broker` module. Should I remove the currently added test afterwards?

lhotari:
> > LGTM, great work @sandeep-mst . Just a wish to add a test mentioned in the description.
> 
> Will add it if possible. That test will go to `pulsar-broker` module. Should I remove the currently added test afterwards?

I think the current unit tests are fine. The additional test would be a bit like an integration test. In Pulsar, the tests which use the `MockedPulsarServiceBaseTest` base class are in most cases integration tests although we call them "unit tests" in CI.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/ProducerImplTest.java
```diff
@@ -67,4 +70,79 @@ public void testPopulateMessageSchema() {
...
+                .semaphoreRelease(Mockito.anyInt());
+        Mockito.doReturn(false)
+                .when(producer)
+                .isBatchMessagingEnabled();
+
+        // Real pending queue
+        ProducerImpl.OpSendMsgQueue pendingQueue = new ProducerImpl.OpSendMsgQueue();
+        Field pendingField = ProducerImpl.class.getDeclaredField("pendingMessages");
+        pendingField.setAccessible(true);
+        pendingField.set(producer, pendingQueue);
```
lhotari: In many cases we use `org.apache.commons.lang3.reflect.FieldUtils` or add a  getter to the field (could use `@Getter(value = AccessLevel.PACKAGE, onMethod_ = @__({@VisibleForTesting}))` to add package protected getter in Lombok.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -2298,11 +2299,17 @@ public void run(Timeout timeout) throws Exception {
      * This fails and clears the pending messages with the given exception. This method should be called from within the
```
Denovo1998: The comment perhaps can be updated a bit.
"fails the messages that were pending at the start of the call, without dropping newly enqueued retry messages"

sandeep-mst: yes makes sense. Done

pulsar-client/src/test/java/org/apache/pulsar/client/impl/ProducerImplTest.java
sandeep-mst: done. Used `FieldUtils` as the `@Getter` approach is not possible for this producer mock as the `pendingMessages` field is `final` .

sandeep-mst:
@lhotari Added the new integration test in `pulsar-broker` module. Please review and let me know if anything else is required.

heesung-sohn (approved):

lhotari (approved):
LGTM, great work @sandeep-mst

Issue #25145, opened 2026-01-15
[Bug] Java consumer occasionally missing one message of a batched entry


YinY1:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

broker and client are in a same machine use same java

-  OS: Linux version 5.4.241-1-tlinux4-0017.16 (mockbuild@VM-81-152-TS3) (gcc version 8.5.0 20210514 (Tencent 8.5.0-23) (GCC))  SMP Thu Dec 12 21:16:37 CST 2024
- Broker version: 4.1.1
- Client version:  4.1.1 (Java)
- Java version:
   - openjdk version "24.0.2" 2025-07-15
   - OpenJDK Runtime Environment Temurin-24.0.2+12 (build 24.0.2+12)
   - OpenJDK 64-Bit Server VM Temurin-24.0.2+12 (build 24.0.2+12, mixed mode, sharing) 

### Issue Description

# What happened
## What I did
Continuously produced a certain amount of  messages to a **partitioned topic**, and consumed those messages with **two subscriptions [sub-1, sub-2]** *individually*, and check their consistency. Consistent means a message has been sent and **received** from all subscriptions, so it will be checked after received and **before acknowledged**.
options i used (use default if not mentioned)
- producer
  - batchingMaxMessagess = 1000
- consumer
  - enableBatchIndexAcknowledgment = true
  - ack mode = Individual
  - acknowledgmentGroupTime = 100ms
  - maxAcknowledgmentGroupSize = 1000

## Expected
All messages produced could be received from all subscriptions.
If produced 1000000, I should get
```
--- Total Acked Messages per Subscription ---
2026-01-14 11:56:07.351 [main] INFO  org.example.Main - Subscription [sub-2]: 1000000 acks
2026-01-14 11:56:07.351 [main] INFO  org.example.Main - Subscription [sub-1]: 1000000 acks 
```
## Happened
**Occasionally**, missed one message with a random batch index
```
--- Total Acked Messages per Subscription ---
2026-01-14 11:56:07.351 [main] INFO  org.example.Main - Subscription [sub-2]: 999999 acks
[...]

Issue #24724, opened 2025-09-10
[Bug] Batch index ACK might recycle ConcurrentBitSetRecyclable twice


BewareMyPower:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

- Java Client
- commit id: 7547fab3b441d9062fa6f67a3a3d50778c37e15e

### Issue Description

`PersistentAcknowledgmentsGroupingTracker#flush` could be called concurrently in different threads: one from the scheduled task according to `acknowledgementGroupTimeMicros`, the other is from the thread that calls `Consumer#acknowledgeAsync`.

When calling `flush` concurrently, the access to `pendingIndividualBatchIndexAcks` is not thread safe:

https://github.com/apache/pulsar/blob/415c6fa74e908bad2da1e7f986185c6ef17cb9a1/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java#L487-L495

- Thread 1: Add `ConcurrentBitSetRecyclable` to `entriesToAck`
- Thread 2: Add `ConcurrentBitSetRecyclable` to `entriesToAck`
- Thread 1: Remove the entry that holds that bit set
- Thread 2: Remove the entry that holds that bit set (and will fail)

Then two threads will share the same `ConcurrentBitSetRecyclable` object and call `recycle` twice.

### Error messages

```text

```

### Reproducing the issue

It's hard to reproduce without injecting delay. Here is the patch that applies on 7547fab3b441d9062fa6f67a3a3d50778c37e15e

```diff
diff --git a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
index 5f7957d7f1..f982108c1f 100644
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java
[...]

PR #22361, opened 2024-03-26
[fix][misc] Make ConcurrentBitSet thread safe

lhotari:
Fixes #22360

### Motivation

ConcurrentBitSet isn't thread safe although it claims this in the javadoc. This is causing concurrency issues in Pulsar. See #22360 for examples.

### Modifications

- Fix usage of StampedLock in ConcurrentBitSet
  - modifications must use the write lock
  - read operations must use the read lock

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #25208, opened 2026-02-04
[fix][client] Fix race condition between isDuplicate() and flushAsync() method in PersistentAcknowledgmentsGroupingTracker due to incorrect use Netty Recycler

oneby-wang:
Fixes https://github.com/apache/pulsar/issues/25145

### Motivation

See https://github.com/apache/pulsar/issues/25145#issuecomment-3798271727.

Due to the use of Netty `Recycler`, a `ConcurrentBitSetRecyclable` instance may still be used after being recycled. However, recycling clears all bits, which may cause the `isDuplicate()` method return true under some race conditions.

If we execute to the point where `bitSet != nul`l, but the thread is then switched out and `bitSet` gets recycled, the `isDuplicate()` method will return true at that point.

https://github.com/apache/pulsar/blob/cab92f0164947e3cbb0da1a014951561f1599add/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java#L135-L138

### Modifications

Replace `ConcurrentBitSetRecyclable` with `ConcurrentBitSet`, ensure that `ackSet` is only cleared when the `ackXxx` method is invoked.

As https://github.com/apache/pulsar/pull/24766 suggested, we should stop using Netty `Recycler` in new code. Since using Netty `Recycler` is kind of error prone way here, I think we should remove it.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/oneby-wang/pulsar/pull/26

### DIFF
--- a/pulsar-client/src/test/java/org/apache/pulsar/client/impl/AcknowledgementsGroupingTrackerTest.java
+++ b/pulsar-client/src/test/java/org/apache/pulsar/client/impl/AcknowledgementsGroupingTrackerTest.java
@@ -31,8 +31,10 @@ import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
+import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


lhotari:
Looks like #22353 and #22361 in the past were in this area of code, addressing similar issues. At that time, the impact of recycling wasn't considered.

lhotari:
Great work, thanks for the contribution @oneby-wang !

BewareMyPower:
Glad to see yet another recycler based class is removed. It's typically a case that recyclable should not be used (concurrent access). Actually, we should synchronize `isDuplicate` and `flushAsync`.

I drafted a similar PR half year ago (but just forgot to proceed): https://github.com/BewareMyPower/pulsar/pull/52/changes

IMO the `ConcurrentBitSet` is also over designed.

BewareMyPower (approved):
Anyway, the key change is to remove `bitSet.recycle()` from `newMultiMessageAckCommon`, so keeping `ConcurrentBitSet` not changed is okay.

lhotari:
Just referencing one more change in the same area: #24725 (issue #24724).

pulsar-client/src/test/java/org/apache/pulsar/client/impl/AcknowledgementsGroupingTrackerTest.java
```diff
@@ -405,6 +407,54 @@ public void testDoIndividualBatchAckAsync() {
...
+        }
+
+        int isDuplicateThreadCount = 10;
+        AtomicBoolean assertResult = new AtomicBoolean();
+        List<Thread> isDuplicateThreads = new ArrayList<>(isDuplicateThreadCount);
+        for (int i = 0; i < isDuplicateThreadCount; i++) {
+            Thread isDuplicateThread = new Thread(() -> {
+                for (int j = 0; j < loops; j++) {
+                    boolean duplicate = tracker.isDuplicate(batchMessageId1);
+                    assertResult.set(assertResult.get() || duplicate);
```
Denovo1998, line 440: It seems like atomicity cannot be guaranteed here.
```java
if (duplicate) {
    assertResult.set(true); 
}
```

oneby-wang: Nice catch! I'll submit a PR to address it soon.

oneby-wang: PR submitted: https://github.com/apache/pulsar/pull/25217.

PR #25219, opened 2026-02-05
[feat] PIP-454: Metadata Store Migration Framework (implementation)

merlimat:
### Motivation

Implementation of PIP-454

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/MigrationCoordinator.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/MigrationCoordinator.java
@@ -175,9 +175,10 @@ public class MigrationCoordinator {
             getChildren(path).whenComplete((res, e) -> {
                 if (e != null) {
                     exception.compareAndSet(null, e);
+                } else {
+                    workQueue.addAll(res);
                 }
 
-                workQueue.addAll(res);
                 semaphore.release();
             });
 


pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/MigrationCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+    private void setInitialMigrationPhase() throws MetadataStoreException {
+        try {
+            sourceStore.put(MigrationState.MIGRATION_FLAG_PATH,
+                    ObjectMapperFactory.getMapper().writer()
+                            .writeValueAsBytes(new MigrationState(MigrationPhase.PREPARATION, targetUrl)),
+                    Optional.of(-1L)).get();
+        } catch (Exception e) {
+            throw new MetadataStoreException(e);
+        }
+    }
```
lhotari, line 111: Is there a solution that prevents starting a new migration unless there's no previous migration? I guess retrying would be allowed so the previous state could be failed. Is there a way to do a CAS update to ensure that only a single node becomes the coordinator?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
```diff
@@ -0,0 +1,433 @@
...
+        registerAsParticipant();
+
+        // Watch for migration events
+        watchForMigrationEvents();
+    }
+
+    private void readCurrentState() throws MetadataStoreException {
+        try {
+            // Read the current state
+            var initialState = migrationStateCache.get(MigrationState.MIGRATION_FLAG_PATH).get();
```
lhotari, line 116: since this is read only at startup time, I guess it could by pass the cache.
In addition, it might be necessary to read after using [`sync`](https://github.com/apache/pulsar/blob/63220eadcfc11f43d4537327a3593603ee5be697/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStore.java#L59-L67) on Zookeeper to ensure that it's the most recent state.
I guess it would be a small chance that this could cause actual issues.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
```diff
@@ -0,0 +1,433 @@
...
+
+        if (sourceStore instanceof MetadataStoreLifecycle msl) {
+            msl.initializeCluster();
+        }
+
+        readCurrentState();
+        registerAsParticipant();
+
+        // Watch for migration events
+        watchForMigrationEvents();
```
lhotari, line 109: is there a chance that we lose an event between reading the current state and before the listener is registered?

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/MigrationCoordinator.java
merlimat: The `Optional.of(-1L)` is implying a create-if-not-exists on the z-node.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
merlimat: FIxed.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
merlimat: If we miss the watch (could happen though it should be very rare, eg: ZK reconnections), brokers might not be acting and therefore the migration would be stuck, which is ok since we can retry later.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
```diff
@@ -0,0 +1,434 @@
...
+    private final Set<String> localEphemeralPaths = ConcurrentHashMap.newKeySet();
+
+    private final ScheduledExecutorService executor;
+
+    private final MetadataCache<MigrationState> migrationStateCache;
+
+    private final Set<Consumer<Notification>> listeners = ConcurrentHashMap.newKeySet();
+    private final Set<Consumer<SessionEvent>> sessionListeners = ConcurrentHashMap.newKeySet();
+
+    private final AtomicInteger pendingSourceWrites = new AtomicInteger();
```
codelipenghui, line 87: Should we check `pendingSourceWrites` before starting the migration? Otherwise, updates might be written to ZooKeeper after the migration has begun.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
```diff
@@ -0,0 +1,434 @@
...
+            // for anything that can be deferred (eg: ledgers rollovers)
+            sessionListeners.forEach(listener -> listener.accept(SessionEvent.SessionLost));
+
+            // Initialize target store
+            initializeTargetStore(migrationState.getTargetUrl());
+
+            this.recreateEphemeralNodesInTarget();
+
+            // Acknowledge preparation by deleting the participant id
+            sourceStore.delete(participantId, Optional.empty()).get();
```
codelipenghui, line 196: If there is any exception happened before this line,  the waitForPreparation() method from MigrationCoordinator runs an infinite loop. It seems missed retry for the failures (or timeout from the waitForPreparation in MigrationCoordinator)

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/MigrationCoordinator.java
```diff
@@ -0,0 +1,221 @@
...
+            });
+
+            semaphore.acquire();
+            getChildren(path).whenComplete((res, e) -> {
+                if (e != null) {
+                    exception.compareAndSet(null, e);
+                }
+
+                workQueue.addAll(res);
+                semaphore.release();
```
codelipenghui: If getChildren fails, e is non-null and res is null. Calling workQueue.addAll(null) throws a NullPointerException. Because this throws before semaphore.release() is called, the semaphore permit is leaked. 

The coordinator's while (true) loop relies on semaphore.availablePermits() == MAX_PENDING_OPS to break out when the queue is empty. With a leaked permit, this condition will never be met, and the migration will deadlock.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/MetadataMigrationBase.java
```diff
@@ -0,0 +1,116 @@
...
+            MigrationCoordinator coordinator = new MigrationCoordinator(pulsar().getLocalMetadataStore(), targetUrl);
+
+            // Start migration in background thread
+            pulsar().getExecutor().submit(() -> {
+                try {
+                    log.info("Starting metadata migration to: {}", targetUrl);
+                    coordinator.startMigration();
+                    log.info("Metadata migration completed successfully");
+                } catch (Exception e) {
+                    log.error("Metadata migration failed", e);
```
codelipenghui, line 103: Because coordinator.startMigration() is executed asynchronously, the HTTP response is returned immediately as a success (204 No Content).  If there is any exception happened when starting the migration, we should return the corresponding exception messages to the client side.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
merlimat: Good catch. I forgot to add the check here :/

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/MetadataMigrationBase.java
merlimat: The idea is that migration could take longer time (eg: a big ZK data set) and cause an HTTP timeout here. So, we just check preconditions here and leave the migration happening background.

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/DualMetadataStore.java
merlimat: Fixed

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/MigrationCoordinator.java
merlimat: 👍  Fixed

PR #25229, opened 2026-02-08
[fix][client] Send all chunkMessageIds to broker for redelivery

### LINKED ISSUE #25220: [Bug] [java client] Nacked chunked messages do not get redelivered
### Search before reporting

### Read release policy

### User environment

Broker version - latest
Broker Java version - 21
Client library type - Java
Client library version - 4.0.8 (reproduced starting at 4.0.1)
Client Java version - 21

### Issue Description

We observed in 4.0.5 that after negative acknowledging a chunked message, the consumer stopped. We expected it to consume the redelivered chunked message as well as any future incoming messages.

This appears to be introduced when changing the `private HashMap<MessageId, Long> nackedMessages = null;` to a `private ConcurrentLongLongPairHashMap nackedMessages = null;` in this [PR](https://github.com/apache/pulsar/pull/23582/changes#diff-c6bdcfb1326d517455010ce62ef0b08b2f75d1c86773f6888abb165d6741454f). `nackedMessages` no longer supports a ChunkedMessageId as a key and is now just a LongLongPair of a single message.

When the consumer attempts to find unacked messages to redeliver [here](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1527-L1532), it searches the map using a ChunkedMessageId which is not how the chunked message was stored in the map.

### Error messages

```text

```

### Reproducing the issue

Branch with the test below that reproduces the issue: [branch with test](https://github.com/davidsealand-toast/pulsar/blob/b8dca6be92a00e0eab43fdd0e801638d7dc67e06/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/NegativeAcksTest.java#L601)

```
@Test
    public void testNegativeAckChunkedMessage() throws Exception {
        cleanup();
        setup();
        String topic = BrokerTestUtil.newUniqueName("testNegativeAcksWithChunks");

        @Cleanup
        Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING)
                .topic(topic)
                .subscriptionName("sub1")
                .acknowledgmentGroupTime(0, TimeUnit.SECONDS)
                .negativeAckRedeliveryDelay(1, TimeUnit.SECONDS)
                .subscribe();

        @Cleanup
        Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
                .topic(topic)
                .enableBatching(false)
                .enableChunking(true)
                .chunkMaxMessageSize(1024) // 1KB max - forces chunking for larger messages
                .create();
        String longMessage = "X".repeat(10 * 1024);
        producer.sendAsync(longMessage);
        producer.flush();

        // negative ack the first message
        consumer.negativeAcknowledge(consumer.receive());

        // now 2s has passed, the first message should be redelivered 1s later.
        Message<String> msg1 = consumer.receive(2, TimeUnit.SECONDS);
        assertNotNull(msg1);
    }
```

### Additional information

Two possible solutions could be:
- modify the ConsumerImpl `unAckedChunkedMessageIdSequenceMap` to use the same format of key that NegativeAcksTracker passes
- modify the NegativeAcksTracker `nackedMessages` to preserve the chunked message id sequence format

Please advise on the recommended fix.

berg223:
Fixes #25220

### Motivation

There is an issue that chunked message cannot be redeliveryed for shared subscription mode. The client should send all chunk message ids to broker for redelivery but it hasn't. It only send the last chunk messageId and that's the root cause of this issue.

The problematic code path​​ is that:
```
    NegativeAcksTracker#triggerRedelivery
        UnAckedMessageTracker#addChunkedMessageIdsAndRemoveFromSequenceMap
```
In `addChunkedMessageIdsAndRemoveFromSequenceMap` method, all chunk messageIds of same message should be fetched from `consumerBase.unAckedChunkedMessageIdSequenceMap` but it found nothing there. Actually, `unAckedChunkedMessageIdSequenceMap` has recorded the value by [key](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1527-L1532) which has type of `ChunkMessageIdImpl`.  `ChunkMessageIdImpl` has a different hashcode from `MessageIdImpl`. That's why we found nothing there. 

I have noticed that there is a `nackedMessageIds` map in `NegativeAcksTracker`. And it's convenient to record the origin messageId by `nackedMessageIds` which has a type of `ChunkMessageIdImpl`. 

### Modifications

Record the origin messageId by the map `nackedMessageIds` in `NegativeAcksTracker#add`

### Verifying this change

```java
MessageChunkingSharedTest#testNegativeAckChunkedMessage
```

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/berg223/pulsar/pull/4

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
@@ -158,7 +158,7 @@ class NegativeAcksTracker implements Closeable {
     private synchronized void add(MessageId messageId, int redeliveryCount) {
         if (messageId instanceof TraceableMessageId) {
             Span span = ((TraceableMessageId) messageId).getTracingSpan();
-            if (span != null) {
+            if (span != null || messageId instanceof ChunkMessageIdImpl) {
                 MessageIdAdv msgId = (MessageIdAdv) messageId;
                 nackedMessageIds.computeIfAbsent(msgId.getLedgerId(), k -> new Long2ObjectOpenHashMap<>())
                         .put(msgId.getEntryId(), messageId);
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/MessageChunkingSharedTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/impl/MessageChunkingSharedTest.java
@@ -193,6 +193,39 @@ public class MessageChunkingSharedTest extends ProducerConsumerBase {
         assertEquals(receivedUuidList1, Arrays.asList("A-0", "B-0", "B-1", "A-1"));
     }
 
+    // Issue #25220
+    @Test
+    public void testNegativeAckChunkedMessage() throws Exception {
+        final String topic = "persistent://my-property/my-ns/test-negative-acknowledge-with-chunk";
+
+        @Cleanup
+        Consumer<String> consumer = pulsarClient.newConsumer(Schema.STRING)
+                .topic(topic)
+                .subscriptionName("sub1")
+                .acknowledgmentGroupTime(0, TimeUnit.SECONDS)
+                .subscriptionType(SubscriptionType.Shared)
+                .negativeAckRedeliveryDelay(1, TimeUnit.SECONDS)
+                .subscribe();
+
+        @Cleanup
+        Producer<String> producer = pulsarClient.newProducer(Schema.STRING)
+                .topic(topic)
+                .enableBatching(false)
+                .enableChunking(true)
+                .chunkMaxMessageSize(1024) // 1KB max - forces chunking for larger messages
+                .create();
+        String longMessage = "X".repeat(10 * 1024);
+        producer.sendAsync(longMessage);
+        producer.flush();
+
+        // negative ack the first message
+        consumer.negativeAcknowledge(consumer.receive());
+
+        // now 2s has passed, the first message should be redelivered 1s later.
+        Message<String> msg1 = consumer.receive(2, TimeUnit.SECONDS);
+        assertNotNull(msg1);
+    }
+
     private Producer<String> createProducer(String topic) throws PulsarClientException {
         return pulsarClient.newProducer(Schema.STRING)
                 .topic(topic)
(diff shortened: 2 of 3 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java
```diff
@@ -158,7 +158,7 @@ static long trimLowerBit(long timestamp, int bits) {
     private synchronized void add(MessageId messageId, int redeliveryCount) {
         if (messageId instanceof TraceableMessageId) {
             Span span = ((TraceableMessageId) messageId).getTracingSpan();
-            if (span != null) {
+            if (span != null || messageId instanceof ChunkMessageIdImpl) {
```
lhotari, line 161: The change made in #24873 in this part of the code is slightly confusing. It adds back some logic before #23600/#23582 which broke the nacks for chunked messages. I'm just wondering why this change was made in #24873. @Technoboy- would you be able to share some more context?

lhotari: @thetumbled Do you also have a chance to take a look since this is related to "PIP-393: Improve performance of Negative Acknowledgement"?

lhotari: It seems that this change is fine from the PIP-393 optimization perspective. The additional overhead would be present only when tracing (added in #24873), and when the message id is a ChunkMessageIdImpl (added in this PR).

lhotari:
Kudos to @davidsealand-toast for investigating and reporting the issue and providing the test case that reproduces the issue. Thank you!

Issue #25218, opened 2026-02-05
Flaky-test: PulsarFunctionsK8STest.testCreateFunctionInK8sWithSecrets


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/apache/pulsar/actions/runs/21718811469/job/62651627693?pr=25207#step:12:5662

### Exception stacktrace

```text
Error:  Tests run: 2, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 126.1 s <<< FAILURE! -- in TestSuite
  Error:  org.apache.pulsar.tests.integration.functions.k8s.PulsarFunctionsK8STest.testCreateFunctionInK8sWithSecrets -- Time elapsed: 0.068 s <<< FAILURE!
  org.apache.pulsar.client.admin.PulsarAdminException: Function test-function already exists
  	at org.apache.pulsar.client.admin.PulsarAdminException.wrap(PulsarAdminException.java:252)
  	at org.apache.pulsar.client.admin.internal.BaseResource.sync(BaseResource.java:360)
  	at org.apache.pulsar.client.admin.internal.FunctionsImpl.createFunctionWithUrl(FunctionsImpl.java:199)
  	at org.apache.pulsar.tests.integration.functions.k8s.PulsarFunctionsK8STest.testCreateFunctionInK8sWithSecrets(PulsarFunctionsK8STest.java:111)
  	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
  	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
  	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:139)
  	at org.testng.internal.invokers.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:47)
  	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:76)
  	at org.testng.internal.invokers.InvokeMethodRunnable.call(InvokeMethodRunnable.java:11)
  	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
  	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
[...]

PR #25230, opened 2026-02-08
[improve][test] Set k8s test retry count to 0

pdolif:
Related to #25218

### Motivation

The PULSAR_K8S integration test group consists of one test only (PulsarFunctionsK8STest). Currently, the `testRetryCount` is set to 1. This leads to misleading logs in CI if the test fails. An example can be seen here: https://github.com/apache/pulsar/actions/runs/21718811469/job/62651627693?pr=25207#step:12:5662

```
Error:  Tests run: 2, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 126.1 s <<< FAILURE! -- in TestSuite
Error:  org.apache.pulsar.tests.integration.functions.k8s.PulsarFunctionsK8STest.testCreateFunctionInK8sWithSecrets -- Time elapsed: 0.068 s <<< FAILURE!
org.apache.pulsar.client.admin.PulsarAdminException: Function test-function already exists
 ```
 
If the first test run fails after the Pulsar function was created, there is no cleanup before the second test run, and the second run always fails because the function it tries to create already exists. 
The logs in CI only show the error from the second test run. To get the original error from the first test run, one needs to download the corresponding artifacts.

For the test retry to make sense, one would need to properly clean up existing resources after the first run. This is not easily possible. Even if the function is deleted, the package metadata will remain, and the second test run will fail due to:

```
org.apache.pulsar.client.admin.PulsarAdminException$ServerSideErrorException: Package 'function://public/default/test-function@0' metadata already exists
```

That's why I propose to disable the retry for this test. Then, if the first attempt fails, the CI logs show the actual failure reason.

### Modifications

Changed `testRetryCount` to 0 for PULSAR_K8S integration test group.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/pdolif/pulsar/pull/21

### DIFF
--- a/build/run_integration_group.sh
+++ b/build/run_integration_group.sh
@@ -208,7 +208,7 @@ test_group_sql() {
 }
 
 test_group_pulsar_k8s() {
-  mvn_run_integration_test "$@" -DintegrationTestSuiteFile=pulsar-k8s.xml -DintegrationTests
+  mvn_run_integration_test "$@" -DintegrationTestSuiteFile=pulsar-k8s.xml -DintegrationTests -DtestRetryCount=0
 }
 
 test_group_pulsar_io() {


lhotari:
btw. There's [org.apache.pulsar.tests.TestRetrySupport](https://github.com/apache/pulsar/blob/master/buildtools/src/main/java/org/apache/pulsar/tests/TestRetrySupport.java) which is used in other locations to add support for retries and cleaning up between retries.

PR #5856, opened 2019-12-13
Make standalone to advertise "localhost" to avoid failure when hostname is not resolvable

merlimat:
### Motivation

The standalone service is failing to startup when the hostname is not resolvable to an IP. This is common in laptops. 

The reason is that the function worker in standalone is trying to connect to the advertised address rather than "localhost". 

To fix this, we should make the standalone to always advertise "localhost" as the the broker address. This is ok because in any case the standalone treats the advertisement as the pulsar proxy, in that it forces the clients to always go through the same service url, so a client connecting from a different host won't have problems.

PR #25238, opened 2026-02-11
[improve][broker] Enhance advertised address resolution with fallback to localhost

nodece:
### Motivation

Currently, in standalone mode, the broker uses `localhost` as the advertised address (as seen in #5856), which works fine for local testing. However, in Kubernetes deployments, this behavior can be problematic: clients connecting via Pulsar Proxy attempt to reach `localhost:6550`, which is incorrect in a podized environment.

### Modifications

- Prefer the broker's fully qualified domain name (FQDN) as the advertised address in the standalone model
- Fallback to `localhost` only if FQDN resolution fails
- Make the standalone test with `--advertised-adress localhost`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/tests/bc_2_0_0/src/test/java/org/apache/pulsar/tests/integration/PulsarContainer.java
+++ b/tests/bc_2_0_0/src/test/java/org/apache/pulsar/tests/integration/PulsarContainer.java
@@ -37,7 +37,7 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
     public PulsarContainer(final String pulsarVersion) {
         super(pulsarVersion);
         withExposedPorts(BROKER_HTTP_PORT, PULSAR_PORT);
-        withCommand("/pulsar/bin/pulsar standalone");
+        withCommand("/pulsar/bin/pulsar standalone --advertised-address localhost");
         waitingFor(new HttpWaitStrategy()
                 .forPort(BROKER_HTTP_PORT)
                 .forStatusCode(200)
--- a/tests/bc_2_0_1/src/test/java/org/apache/pulsar/tests/integration/PulsarContainer.java
+++ b/tests/bc_2_0_1/src/test/java/org/apache/pulsar/tests/integration/PulsarContainer.java
@@ -37,7 +37,7 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
     public PulsarContainer(final String pulsarVersion) {
         super(pulsarVersion);
         withExposedPorts(BROKER_HTTP_PORT, PULSAR_PORT);
-        withCommand("/pulsar/bin/pulsar standalone");
+        withCommand("/pulsar/bin/pulsar standalone --advertised-address localhost");
         waitingFor(new HttpWaitStrategy()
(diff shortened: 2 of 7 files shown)


lhotari:
@nodece This breaks pulsar-standalone, at least for docker usage. it's now necessary to add `--advertised-address localhost` which is a breaking change. Thanks to @onobc for debugging this issue.

nodece:
@lhotari I agree this introduces a breaking change for standalone Docker usage, but I think it's a reasonable trade-off.

The previous default (`localhost`) works for local/docker testing, but it's misleading in containerized environments like Kubernetes, where it can lead to incorrect connectivity (e.g., clients resolving to `localhost:6650`). Prioritizing FQDN makes the behavior more consistent with real deployments and avoids these pitfalls.

For Docker users, the impact is straightforward and explicit; they can restore the old behavior by setting:

```
--advertised-address localhost
```

So while it's technically breaking, it improves correctness and makes the configuration more intentional rather than relying on an implicit default.

nodece:
I will update our doc to use old behavior.

nodece:
Update getting started with Docker: https://github.com/apache/pulsar-site/pull/1109

lhotari:
> @lhotari I agree this introduces a breaking change for standalone Docker usage, but I think it's a reasonable trade-off.
> 
> The previous default (`localhost`) works for local/docker testing, but it's misleading in containerized environments like Kubernetes, where it can lead to incorrect connectivity (e.g., clients resolving to `localhost:6650`). Prioritizing FQDN makes the behavior more consistent with real deployments and avoids these pitfalls.
> 
> For Docker users, the impact is straightforward and explicit; they can restore the old behavior by setting:
> 
> ```
> --advertised-address localhost
> ```
> 
> So while it's technically breaking, it improves correctness and makes the configuration more intentional rather than relying on an implicit default.

@nodece Most users use Pulsar standalone for testing and it breaks the behavior for everyone who has been doing this in the past. Since using Pulsar standalone in Kubernetes is a very rare use case, I'd suggest that we restore the previous default behavior and instead document the Kubernetes deployment of Pulsar standalone if someone really needs that.

nodece:
@lhotari While using an FQDN as the default can avoid misleading setups in containerized environments, the current default (`localhost`) is widely relied upon for local development and testing.

To avoid impacting existing users, I will revert this change and keep the current default behavior.

For the Kubernetes environment, `advertised-address` should be explicitly configured, since relying on the default may lead to incorrect connectivity (e.g., resolving to `localhost`).

I'll follow up with documentation improvements to make this requirement clearer.

merlimat:
The standalone already forces clients to go through the same address they used (same as Pulsar proxy). The address that is being advertised doesn't matter, in the sense that the client will never directly use that. This was done precisely to allow easy testing, where you could spin up a docker container with pulsar standalone and use it from outside docker, without worrying about advertised addresses.

nodece:
My case is that when using standalone and proxy in the k8s env,  the standalone must return FQDN, so I made this PR.

PR #25239, opened 2026-02-11
[feat][broker] Add adaptive publish throttle controller (disabled by default)

hemanth-19:
> **Experimental feature** — disabled by default (`adaptivePublisherThrottlingEnabled=false`).
> A broker restart is required to enable it; all tuning parameters are dynamic config.
>
> **Focused review request** — the areas most likely to have subtle bugs are:
> 1. **Concurrency** — `AdaptivePublishRateLimiter`: the `volatile boolean active` fast path on IO threads (see `handlePublishThrottling`), and the single-writer contract on the controller scheduler thread.
> 2. **Lifecycle** — `BrokerService.startAdaptivePublishThrottleController()` and the `close()` path; `ThrottleType.AdaptivePublishRate` ordinal stability and the new `states[]` slot.
> 3. **Hysteresis** — deactivation requires **both** memory **and** backlog pressure to drop below their low watermarks simultaneously; verify this is correct for your use case.

---

## Summary

- Adds an adaptive publish throttle controller that dynamically reduces producer publish rates when JVM heap usage or per-topic backlog size approaches configurable watermarks.
- **Disabled by default.** Zero code path changes when `adaptivePublisherThrottlingEnabled=false`.
- **Observe-only mode** (`adaptivePublisherThrottlingObserveOnly=true`, toggleable at runtime) computes and logs decisions but never applies throttling — safe for validating in production and usable as an emergency circuit-breaker.
- Controller never dies silently: every cycle is wrapped in `try-catch-finally`; failures increment a dedicated OTel counter.
- `ThrottleType.AdaptivePublishRate` is appended as the last enum constant to preserve existing ordinals 0–6. A guard test (`ThrottleTypeEnumTest`) fails immediately if the enum is accidentally mutated.

---

### Configuration (reviewers: verify defaults, dynamic flags, and conf entry comments)

| File | What changed |
|------|-------------|
| `ServiceConfiguration.java` | 10 new `@FieldContext` fields — `adaptivePublisherThrottling*` |
| `conf/broker.conf` | 10 new commented-out entries with inline docs and a recommended quick-start snippet |
| `README.md` | New "Experimental broker features" section with quick-start `broker.conf` snippet |

### Broker wiring (reviewers: focus on concurrency, lifecycle, and enum ordinal safety)

| File | What changed |
|------|-------------|
| `ServerCnxThrottleTracker.java` | New `ThrottleType.AdaptivePublishRate` constant (appended last, reentrant); class-level Javadoc on ordinal stability |
| `AbstractTopic.java` | New `AdaptivePublishRateLimiter` field; `handlePublishThrottling()` delegation; uses `ThrottleType.AdaptivePublishRate` |
| `BrokerService.java` | `startAdaptivePublishThrottleController()` lifecycle; `forEachPersistentTopic()` helper; `close()` teardown |
| `AdaptivePublishRateLimiter.java` *(new)* | Per-topic limiter: `volatile boolean active` fast path, asymmetric EWMA, `activate()`/`deactivate()` |
| `AdaptivePublishThrottleController.java` *(new)* | Broker-level scheduler: pressure math, bounded-step rate changes, hysteresis, `observeOnly` guard, `try-catch-finally` safety |
| `OpenTelemetryAdaptiveThrottleStats.java` *(new)* | 6 broker-level OTel metrics (3 always-on + 3 health); 6 per-topic metrics (opt-in) |

### Tests (reviewers: check the observeOnly safety tests and the enum guard tests)

| File | Key assertions |
|------|---------------|
| `AdaptivePublishRateLimiterTest.java` *(new)* | No-op guarantee, activate/deactivate, asymmetric EWMA, `observeOnly` never changes channel autoread |
| `AdaptivePublishThrottleControllerTest.java` *(new)* | `linearPressure()`, bounded-step `computeTargetRate()`, hysteresis, `observeOnly` never calls `activate()` |
| `AdaptiveThrottleEndToEndTest.java` *(new)* | Full control-loop integration: pressure → throttle → drain → deactivate; `observeOnly` IO-thread safety |
| `ThrottleTypeEnumTest.java` *(new)* | Minimum constant count, `AdaptivePublishRate` present and last, declared reentrant |

---

## Design FAQ

**Q: Why introduce a new `ThrottleType.AdaptivePublishRate` instead of reusing `TopicPublishRate`?**

`ServerCnxThrottleTracker` uses reference-counting via `states[ThrottleType.ordinal()]`.
If the adaptive limiter shared `TopicPublishRate` with the static rate limiter, a single
`unmarkThrottled()` call from the adaptive side could decrement the count that the static
limiter had incremented — leaving the connection unthrottled even though the static limit is
still exceeded. A dedicated constant keeps the two signals fully independent and eliminates an
entire class of ordering bugs.

**Q: Why does the controller not coordinate across brokers? Each broker throttles independently.**

Adaptive throttling reacts to local signals — JVM heap on this JVM, backlog on topics owned
by this broker. Cross-broker coordination would require a consensus protocol, introduce network
latency on the hot publish path, and create a single point of failure. Local-only decisions
are simpler, faster, and fail-safe: if a network partition occurs, each broker still protects
itself independently. Cluster-wide load imbalance (e.g. one hot broker) is better addressed
by Pulsar's topic-migration and load-balancer machinery than by throttle coordination.

**Q: What happens if the controller thread crashes?**

The evaluation loop is wrapped in `try-catch-finally`. A caught exception:
1. Increments `evaluationFailureCount` (surfaced as OTel counter `pulsar.broker.adaptive.throttle.controller.evaluation.failure.count`).
2. Logs the full stack trace at `ERROR` level so it appears in broker logs immediately.
3. Does **not** kill the `ScheduledExecutorService` — the scheduler reschedules the next cycle unconditionally.

If the failure is persistent, `last.evaluation.timestamp` stops advancing while `evaluation.failure.count` climbs — a clear alert signal. A full controller stall (only possible via an unchecked `Error`) requires a broker restart; the OTel staleness alert will fire within `3 × intervalMs`.

---

## Test plan

- [ ] `mvn test -pl pulsar-broker -am -Dtest="AdaptivePublishRateLimiterTest,AdaptivePublishThrottleControllerTest,AdaptiveThrottleEndToEndTest,ThrottleTypeEnumTest" -DfailIfNoTests=false --no-transfer-progress` passes
- [ ] `mvn test -pl pulsar-broker -am -Dtest="PublishRateLimiterTest" -DfailIfNoTests=false --no-transfer-progress` (regression: static rate limiter unaffected)
- [ ] Broker starts with `adaptivePublisherThrottlingEnabled=false` (default): no controller thread, no `AdaptivePublishRateLimiter` allocated, no OTel instruments registered
- [ ] `observeOnly=true`: logs show `OBSERVE-ONLY would-activate`, zero `ACTIVATED` lines, `active.topic.count` metric stays at 0
- [ ] OTel scrape shows all 6 broker-level metrics when feature is enabled

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AdaptivePublishThrottleController.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AdaptivePublishThrottleController.java
@@ -1,0 +1,472 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.broker.service;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.io.Closeable;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.broker.ServiceConfiguration;
+import org.apache.pulsar.broker.service.persistent.PersistentTopic;
+import org.apache.pulsar.common.policies.data.BacklogQuota;
+import org.apache.pulsar.common.util.SingleThreadNonConcurrentFixedRateScheduler;
+
+/**
+ * Broker-level controller that periodically evaluates system pressure signals
+ * (JVM heap usage and per-topic backlog pressure) and drives per-topic
+ * {@link AdaptivePublishRateLimiter} instances.
+ *
+ * <h3>Pressure computation</h3>
+ * <pre>
+ *   memoryPressure = clamp((heapUsed/maxHeap - lowWm) / (highWm - lowWm), 0, 1)
+ *
+ *   backlogPressure per topic:
+ *     if (quota ≤ 0)  → 0                    (no quota configured)
+ *     fraction = backlogBytes / quotaBytes
+ *     if (fraction ≤ backlogLowWm)  → 0
+ *     if (fraction ≥ backlogHighWm) → 1
+ *     else → (fraction - backlogLowWm) / (backlogHighWm - backlogLowWm)
+ *
+ *   pressureFactor = max(memoryPressure, backlogPressure)
+ * </pre>
+ *
+ * <h3>Rate computation (bounded step)</h3>
+ * <pre>
+ *   targetFactor = 1.0 - pressureFactor * (1.0 - minRateFactor)
+ *   targetMsgRate = max(1, naturalMsgRate * targetFactor)
+ *   delta = naturalMsgRate * maxRateChangeFactor
+ *   effectiveMsgRate = clamp(currentRate ± delta, towards target)
+ * </pre>
+ *
+ * <h3>Hysteresis</h3>
+ * Throttle activates when {@code pressureFactor > 0}.  It is fully released only
+ * when {@code pressureFactor == 0}, which requires both memory and backlog pressure
+ * to drop below their respective low watermarks.
+ *
+ * <h3>Observe-only mode</h3>
+ * When {@code adaptivePublisherThrottlingObserveOnly=true} (dynamic config) the
+ * controller computes everything — pressure factors, target rates, decisions — and
+ * emits metrics and OBSERVE-ONLY log lines, but never calls
+ * {@link AdaptivePublishRateLimiter#activate} or
+ * {@link AdaptivePublishRateLimiter#deactivate}.  This is the safe validation mode
+ * before going live, and also serves as an emergency circuit-breaker (flip the
+ * dynamic config at runtime to suspend all throttling without a restart).
+ */
+@Slf4j
+public class AdaptivePublishThrottleController implements Closeable {
+
+    private final BrokerService brokerService;
+    private final SingleThreadNonConcurrentFixedRateScheduler scheduler;
+
+    /**
+     * Current broker-wide memory pressure factor (0.0–1.0).
+     * Written by the controller thread; read by metrics/tests.
+     */
+    @Getter
+    private volatile double currentMemoryPressureFactor = 0.0;
+
+    /**
+     * Number of topics on this broker currently being adaptively throttled.
+     */
+    private final AtomicInteger activeThrottledTopicsCount = new AtomicInteger(0);
+
+    /**
+     * Total number of throttle activations since the controller started.
+     */
+    private final AtomicLong totalActivationCount = new AtomicLong(0L);
+
+    /**
+     * Unix epoch milliseconds when the last evaluation cycle completed (success or failure).
+     * Zero if no cycle has run yet.  Operators can alert when this stops advancing.
+     */
+    private volatile long lastEvaluationCompletedEpochMs = 0L;
+
+    /**
+     * Wall-clock duration in milliseconds of the most recent evaluation cycle.
+     * A sustained increase signals the broker has too many topics for the configured interval.
+     */
+    private volatile long lastEvaluationDurationMs = 0L;
+
+    /**
+     * Number of evaluation cycles that threw an uncaught exception.
+     * Any increment means the controller is degraded; check broker logs for
+     * "[AdaptiveThrottleController] Evaluation cycle failed".
+     */
+    private final AtomicLong evaluationFailureCount = new AtomicLong(0L);
+
+    public AdaptivePublishThrottleController(BrokerService brokerService) {
+        this.brokerService = brokerService;
+        this.scheduler =
+                new SingleThreadNonConcurrentFixedRateScheduler("pulsar-adaptive-throttle-controller");
+    }
+
+    /** Start periodic evaluation. */
+    public void start() {
+        ServiceConfiguration config = brokerService.getPulsar().getConfiguration();
+        long intervalMs = config.getAdaptivePublisherThrottlingIntervalMs();
+        log.info("[AdaptiveThrottleController] Starting with intervalMs={} observeOnly={}",
+                intervalMs, config.isAdaptivePublisherThrottlingObserveOnly());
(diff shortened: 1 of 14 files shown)


hemanth-19:
## Local build notes (for reviewers — not a code issue)

**tl;dr** — the broker module compiles cleanly and Checkstyle is green. The failures below are all environment constraints on the dev machine, not problems with this PR.

---

### What succeeded

| Step | Result |
|------|--------|
| `mvn install -Pcore-modules,-main -DskipTests -rf :pulsar-client-messagecrypto-bc` | **BUILD SUCCESS** — full core reactor installed |
| Checkstyle on `pulsar-broker` | **0 violations** |
| All 14 new/changed files compile | ✓ |

A fix was needed mid-way: the initial parallel build (`-T 4`) had a **pre-existing reactor race** where `pulsar-client-messagecrypto-bc` started compiling against `bouncy-castle-bc-4.2.0-SNAPSHOT-pkg.jar` before the `package` phase finished writing it. This is a known fragility in the Pulsar reactor and is **unrelated to this PR**. Rebuilding sequentially (no `-T` flag) resolved it immediately.

One genuine bug caught by the local build: three casts in `AdaptiveThrottleEndToEndTest.java` were written as `(long)(expr)` instead of `(long) (expr)` (missing space after typecast). Fixed and force-pushed to the branch before the Checkstyle run.

---

### Why the build stopped at `pulsar-offloader-distribution`

The `mvn install -rf :pulsar-broker` run continued downstream into `pulsar-offloader-distribution`, which needs `tiered-storage-file-system` and `tiered-storage-jcloud` — modules that weren't in the local cache because they sit outside the `core-modules` profile scope. This failure is **completely unrelated to the throttling change**; it would happen on any clean checkout without a prior full install.

---

### Why the targeted tests did not finish locally

Surefire's default configuration for `pulsar-broker` forks 4 JVMs simultaneously, each with `-Xmx1300M -XX:+UseZGC`. On this dev machine that totals ~5.2 GB of heap, which exceeds available memory. The JVM exits with `Error occurred during initialization of VM` before running a single test — no test failure, purely a resource limit.

To reproduce the test run on any machine with ≥ 2 GB free heap:

```bash
mvn test -pl pulsar-broker \
  -Dtest="AdaptivePublishRateLimiterTest,AdaptivePublishThrottleControllerTest,\
AdaptiveThrottleEndToEndTest,ThrottleTypeEnumTest" \
  -DfailIfNoTests=false \
  -DforkCount=1 -DreuseForks=true \
  -DargLine="-Xmx512m -XX:+UseSerialGC" \
  --no-transfer-progress
```

CI (GitHub Actions) runs in a clean environment with sufficient memory and will validate the full suite there.

hemanth-19:
**Doc label:** This PR adds an experimental, disabled-by-default broker feature. It includes:

- Inline Javadoc on all new public classes and the new `ThrottleType.AdaptivePublishRate` constant
- `README.md` updated with a quick-start snippet under a new "Experimental broker features" section
- `conf/broker.conf` updated with all 10 new config keys, commented out with inline descriptions
- `ADAPTIVE_THROTTLE_PR.md` at the repo root with the full config reference, metrics guide, and troubleshooting steps

When this feature graduates from experimental (i.e. `adaptivePublisherThrottlingEnabled` defaults to `true`), the Pulsar site docs (`reference-configuration-broker.md` and the "Manage topics" section) will need corresponding updates. Until then, the in-tree README and conf comments are the intended doc surface.

Label suggestion: `doc-required` (for the site update at graduation) or `doc-not-needed` if maintainers prefer to defer site docs until the feature is no longer experimental.

lhotari:
@hemanth-19 Thanks for the contribution! In Apache Pulsar, we follow the PIP process to introduce new features. Please check https://github.com/apache/pulsar/tree/master/pip#pulsar-improvement-proposal-pip for more details.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AdaptivePublishThrottleController.java
```diff
@@ -0,0 +1,472 @@
...
+    @VisibleForTesting
+    double computeMemoryPressure(ServiceConfiguration config) {
+        Runtime runtime = Runtime.getRuntime();
+        long maxMemory = runtime.maxMemory();
+        if (maxMemory == Long.MAX_VALUE) {
+            // Unbounded heap — cannot compute pressure.
+            return 0.0;
+        }
+        long usedMemory = maxMemory - runtime.freeMemory();
+        double usageFraction = (double) usedMemory / maxMemory;
```
lhotari, line 280: Pulsar uses direct memory besides heap memory. 
Here's one example of heap + direct memory  usage calculation:
https://github.com/apache/pulsar/blob/df5197212e8806c7d1907dedfcdfd9e40a4f0ea5/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/impl/LoadManagerShared.java#L300-L316
However, that example doesn't take it into account that Netty's allocator performs pooling and there would be more free direct memory available that the stats show in that example.
https://github.com/apache/pulsar/blob/019ae9f0d0ec961389bd093e7350ca6e23c3f496/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/AllocatorStatsGenerator.java#L35 contains an example with details Netty allocator stats. It would be possible to calculate the exact allocation of direct memory by subtracting the free memory in directArenas->chunkLists->freeBytes

lhotari:
> CI (GitHub Actions) runs in a clean environment with sufficient memory and will validate the full suite there.

@hemanth-19 Please setup ["Personal CI"](https://pulsar.apache.org/contribute/personal-ci/) in your fork and run Pulsar CI in that environment.

It would be advisable to reduce the AI generated noise in PR description and comments so that you spend effort on your side validating what gets generated by AI. Thank you!

lhotari:
Sharing some thoughts around publish throttling / backpressure / flow control:

In Pulsar, there's already a solution to limit publishing with memory backpressure. `maxMessagePublishBufferSizeInMB` is set to 50% of direct memory by default, which might be a too high of a default. The downside of the current solution is that the limit is a fixed limit and it doesn't take existing other direct memory consumption into account since that might vary.
Another property is that it's not possible to prioritize specific tenants/namespaces with the existing solution.

One of the problems with existing publish throttling is that there's no explicit flow control in the Pulsar protocol and publish throttling happens by pausing reads on the TCP/IP connection. This solution impacts all producers and consumers sharing the same TCP/IP connection. 

One of the improvement directions would be to introduce explicit producer flow control.
https://pulsar.apache.org/blog/2024/10/24/announcing-apache-pulsar-4-0/#enhanced-quality-of-service-qos-controls

What is currently missing from Pulsar is the ability to explicitly prioritize tenants/namespaces/topics and have dynamic limits instead of fixed rate limits. This would be useful in cases where an overall broker limit is hit. Pulsar's "Resource Groups" ([PIP-82](https://github.com/apache/pulsar/blob/master/pip/pip-82.md)) could be further developed to the direction what Confluent Kora has with the [dynamic tenant-level quotas](https://jack-vanlightly.com/analyses/2023/11/14/kora-serverless-kafka-asds-chapter-2#:~:text=Dynamic%20tenant%2Dlevel%20quotas) feature.

Regarding this PR, I don't currently understand what essential benefit it brings over the existing `maxMessagePublishBufferSizeInMB` solution and existing solution for limiting the per-topic backlog. Explaining the benefit and impact of this PR would be useful in justifying the changes.

PR #25240, opened 2026-02-11
[fix] Fix ManagedLedgerImpl's pendingAddEntries leak.

zhaizhibo:
### Motivation

There ard two scenarios where pendingAddEntries can leak in a ManagedLedger.

1. **Leak in Fenced state**
After an add operation, there may be a need to update the ledger's properties. If the ZooKeeper update fails, the callback sets the ManagedLedger state to fenced. Subsequent write operations that fail will trigger ManagedLedgerImpl#ledgerClosed. For a ManagedLedger in the fenced state, this can result in pendingAddEntries not being properly cleaned up, causing a leak.
2. **Leak in Closed state**
The ManagedLedger needs to rollover because the current ledger is full, switching its state to closedLedger or creatingLedger. New incoming add requests are appended to the pendingAddEntries queue but have not yet been initiated. If the ManagedLedger is closed (state set to closed) after the rollover is executed but before its callback returns, the requests in the pendingAddEntries queue will never be completed, resulting in a leak.

### Modifications

1. add a new function clearNotInitiatedPendingAddEntries(), which will be called after ManagedLedger state be set to closed or fenced. This method will immediately fail all OpAddEntry operations in the pendingAddEntries queue that have not yet been initiated  and remove them from the queue.
2. In the OpAddEntry#handleAddFail method, add a check for the fenced state. If the ManagedLedger is fenced, we should also clear all subsequent pending write requests in the queue to prevent leaks.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


zhaizhibo:
> ## Pull request overview
> 
> This PR addresses `pendingAddEntries` leaks in `ManagedLedgerImpl` when the managed ledger transitions to `Closed` or `Fenced` while there are queued add operations that were never initiated.
> 
> **Changes:**
> - Add `OpAddEntry.closeIfNotInitiated()` and use it to fail/remove queued-but-not-initiated add operations on close/fence.
> - Schedule the post-rollover metadata-update completion to run on the managed ledger executor thread.
> - Clear pending writes when `ledgerClosed()` is invoked while in `Fenced` state.
> 
> ### Reviewed changes
> 
> Copilot reviewed 2 out of 2 changed files in this pull request and generated 5 comments.
> 
> | File | Description |
> | ---- | ----------- |
> | managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpAddEntry.java | Adds a helper to atomically close an `OpAddEntry` if it has not been initiated. |
> | managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java | Adds selective clearing of not-initiated pending adds on close/fence and adjusts rollover callback execution/threading and fenced handling. |
> 
> 
> 
> 
> 
> ---
> 
> 💡 <a href="/apache/pulsar/new/master/.github/instructions?filename=*.instructions.md" class="Link--inTextBlock" target="_blank" rel="noopener noreferrer">Add Copilot custom instructions</a> for smarter, more guided reviews. <a href="https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot" class="Link--inTextBlock" target="_blank" rel="noopener noreferrer">Learn how to get started</a>.

Fixed

zhaizhibo:
add a test to reproduce the scenario where the ledger is closed before rollover completes(maybe ZooKeeper operations are slow).

zhaizhibo:
add a test to reproduce the scenario where the ledger is fenced before OpAddEntry.addcomplete().

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2070,6 +2087,17 @@ void clearPendingAddEntries(ManagedLedgerException e) {
         }
     }
 
+    void clearNotInitiatedPendingAddEntries(ManagedLedgerException e) {
+        Iterator<OpAddEntry> iterator = pendingAddEntries.iterator();
+        while (iterator.hasNext()) {
+            OpAddEntry op = iterator.next();
+            if (op.closeIfNotInitiated()) {
+                op.failed(e);
+                iterator.remove();
```
Denovo1998: Here, perhaps `iterator.remove()` should be called first to remove the op from the queue, and then `op.failed(e)`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1751,20 +1753,32 @@ public void operationComplete(Void v, Stat stat) {
...
+                                currentLedgerTimeoutTriggered = new AtomicBoolean();
+                                currentLedgerEntries = 0;
+                                currentLedgerSize = 0;
+                                updateLedgersIdsComplete(originalCurrentLedger);
+                                mbean.addLedgerSwitchLatencySample(System.currentTimeMillis()
+                                        - lastLedgerCreationInitiationTimestamp, TimeUnit.MILLISECONDS);
+                                // May need to update the cursor position
+                                maybeUpdateCursorBeforeTrimmingConsumedLedger();
+                            }
+                            metadataMutex.unlock();
```
Denovo1998, line 1777: Please re-check STATE_UPDATER inside the executor task and bail out if Closed/Fenced (while still releasing metadataMutex).

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1751,20 +1753,32 @@ public void operationComplete(Void v, Stat stat) {
...
-                        currentLedgerSize = 0;
-                        updateLedgersIdsComplete(originalCurrentLedger);
-                        mbean.addLedgerSwitchLatencySample(System.currentTimeMillis()
-                                - lastLedgerCreationInitiationTimestamp, TimeUnit.MILLISECONDS);
-                        // May need to update the cursor position
-                        maybeUpdateCursorBeforeTrimmingConsumedLedger();
+                    // make sure that pendingAddEntries' operations are executed in the same thread
+                    // to avoid potential concurrent issues
+                    State state = STATE_UPDATER.get(ManagedLedgerImpl.this);
+                    if (state == State.Closed || state.isFenced()) {
```
Denovo1998: The metadataMutex is acquired in updateLedgersListAfterRollover() and must always be released. In operationComplete(), the "Closed/Fenced" branch skips the ledger update but fails to call metadataMutex.unlock(), causing a lock leak that may deadlock subsequent metadata operations. Ensure unlock() is invoked in all branches, for example, by using a finally block.

Denovo1998 (changes requested):
Perhaps a unit test can be added:
Close occurs after updateLedgersListAfterRollover() has acquired the metadataMutex, but before operationComplete() has executed or just about to execute.

zhaizhibo:
@Denovo1998 Thanks for you review, I will fix the commets soon.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1751,20 +1753,32 @@ public void operationComplete(Void v, Stat stat) {
...
+                        if (log.isDebugEnabled()) {
+                            log.debug("[{}] Skipping ledger update after create complete because ledger is "
+                                + "closed or fenced", name);
+                        }
```
codelipenghui: [Critical] Two issues in this branch:

1. **`metadataMutex` deadlock**: When this branch is taken, `metadataMutex.unlock()` is never called. The mutex was acquired in `updateLedgersListAfterRollover()` via `metadataMutex.tryLock()`. Since `CallbackMutex` is a `Semaphore(1)`, the permit is permanently consumed. All future metadata operations (trimming, offloading, etc.) that need this mutex will deadlock or silently skip.

2. **LedgerHandle `lh` leak**: The newly created BookKeeper ledger handle `lh` is never closed in this branch. It's not assigned to `currentLedger` and not closed via `lh.closeAsync()`. This leaks a BK ledger handle. Note that `createComplete` (line 1712) only handles `State.Closed` with `lh.closeAsync()`, not fenced states, so the fenced path always reaches here.

Suggested fix:
```java
if (state == State.Closed || state.isFenced()) {
    metadataMutex.unlock();
    if (lh != null) {
        lh.closeAsync();
    }
    ...
}
```

codelipenghui (commented):
Thanks for the fix. The two leak scenarios identified in the PR description are real and well-described. However, I found a few issues that should be addressed before merging. See inline comments.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1751,20 +1753,32 @@ public void operationComplete(Void v, Stat stat) {
...
-                        currentLedgerEntries = 0;
-                        currentLedgerSize = 0;
-                        updateLedgersIdsComplete(originalCurrentLedger);
-                        mbean.addLedgerSwitchLatencySample(System.currentTimeMillis()
-                                - lastLedgerCreationInitiationTimestamp, TimeUnit.MILLISECONDS);
-                        // May need to update the cursor position
-                        maybeUpdateCursorBeforeTrimmingConsumedLedger();
+                    // make sure that pendingAddEntries' operations are executed in the same thread
+                    // to avoid potential concurrent issues
+                    State state = STATE_UPDATER.get(ManagedLedgerImpl.this);
```
codelipenghui: [Important] TOCTOU race: The state is read here *outside* `executor.execute()`, but the ledger update runs *inside* it asynchronously. Between this check and when the lambda executes, `asyncClose()` or `setFenced()` could change the state. The lambda would then proceed to call `updateLedgersIdsComplete()` which sets state to `LedgerOpened` and calls `initiate()` on pending ops — defeating the purpose of this fix.

The original code didn't have this race because the `synchronized` block was entered immediately.

Consider moving the state check inside the executor's synchronized block:
```java
executor.execute(() -> {
    synchronized (ManagedLedgerImpl.this) {
        State state = STATE_UPDATER.get(ManagedLedgerImpl.this);
        if (state == State.Closed || state.isFenced()) {
            lh.closeAsync();
            metadataMutex.unlock();
            return;
        }
        // ... existing update logic ...
    }
    metadataMutex.unlock();
});
```

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
```diff
@@ -764,6 +764,93 @@ public void invalidReadEntriesArg2() throws Exception {
         fail("Should have thrown an exception in the above line");
     }
 
+    @Test(timeOut = 30000)
+    public void testCloseManagedLedgerAfterRollover() throws Exception {
+        ManagedLedgerFactoryConfig config = new ManagedLedgerFactoryConfig();
+        config.setMaxCacheSize(0);
+        ManagedLedgerFactoryImpl factory = new ManagedLedgerFactoryImpl(metadataStore, bkc, config);
```
codelipenghui, line 791: [Minor] This `factory` is never shut down. Consider adding `@Cleanup("shutdown")` as the second test (`testFencedManagedLedgerAfterAdd`) does for its factory. Without cleanup, the factory's internal executors leak.

managed-ledger/src/test/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerTest.java
```diff
@@ -764,6 +764,93 @@ public void invalidReadEntriesArg2() throws Exception {
...
+        @Cleanup("shutdown")
+        ManagedLedgerFactory factory1 = new ManagedLedgerFactoryImpl(metadataStore, bkc);
+        ManagedLedgerImpl realLedger = (ManagedLedgerImpl) factory1.open("my_test_ledger");
+        ManagedLedgerImpl ledger = spy(realLedger);
+
+        int sendNum = 10;
+        CountDownLatch sendLatch = new CountDownLatch(sendNum);
+        CountDownLatch fencedLatch = new CountDownLatch(1);
+        doAnswer(invocationOnMock -> {
+            for (int i = 0; i < sendNum; ++i) {
```
codelipenghui, line 850: [Minor] `stopBookKeeper()` and `stopMetadataStore()` are called inside the loop (10 times). They should be called once before the loop — subsequent calls are no-ops since `bkc` is already null after the first `stopBookKeeper()`.

zhaizhibo:
> Thanks for the fix. The two leak scenarios identified in the PR description are real and well-described. However, I found a few issues that should be addressed before merging. See inline comments.

@codelipenghui Thanks for the review. All issues mentioned in the comments have been fixed. Please take another look.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -4416,13 +4441,22 @@ private void checkManagedLedgerIsOpen() throws ManagedLedgerException {
...
+        if (STATE_UPDATER.get(this) != State.Fenced) {
+            STATE_UPDATER.set(this, State.Fenced);
+            clearPendingAddEntries(new ManagedLedgerFencedException("ManagedLedger "
+                + name + " is fenced"));
+        }
     }
 
     synchronized void setFencedForDeletion() {
-        log.info("Moving to FencedForDeletion state");
+        log.info().attr("ledgerName", name).log("Moving to FencedForDeletion state");
```
dao-jun: It will always override FencedForDeletion to FENCED

zhaizhibo: Thanks for review, I have fixed this.

dao-jun (approved):

Denovo1998 (approved):

zhaizhibo:
@lhotari Could you please help review this PR? 😁

zhaizhibo:
@dao-jun  All the checks have passed, ready for merge. 😄

dao-jun:
> @dao-jun All the checks have passed, ready for merge. 😄

need more eyes on the PR.
@lhotari @merlimat @nodece PTAL

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1746,19 +1747,39 @@ public void operationComplete(Void v, Stat stat) {
...
-                        updateLedgersIdsComplete(originalCurrentLedger);
-                        mbean.addLedgerSwitchLatencySample(System.currentTimeMillis()
-                                - lastLedgerCreationInitiationTimestamp, TimeUnit.MILLISECONDS);
-                        // May need to update the cursor position
-                        maybeUpdateCursorBeforeTrimmingConsumedLedger();
+                        try {
+                            State state = STATE_UPDATER.get(ManagedLedgerImpl.this);
+                            if (state == State.Closed || state.isFenced()) {
+                                log.debug()
+                                        .attr("name", name)
```
nodece:
```suggestion
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -1746,19 +1747,39 @@ public void operationComplete(Void v, Stat stat) {
...
+                                        .log("skip ledger update after create complete ledger is closed or fenced");
+                                lh.closeAsync().exceptionally(e -> {
+                                    if (e != null) {
+                                        log.error()
+                                            .attr("ledgerName", name)
```
nodece:
```suggestion
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
nodece: log includes the ledger name.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -4416,13 +4441,21 @@ private void checkManagedLedgerIsOpen() throws ManagedLedgerException {
 
     @VisibleForTesting
     public synchronized void setFenced() {
-        log.info("Moving to Fenced state");
-        STATE_UPDATER.set(this, State.Fenced);
+        log.info().attr("ledgerName", name).log("Moving to Fenced state");
```
nodece:
```suggestion
        log.info().log("Moving to Fenced state");
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -4416,13 +4441,21 @@ private void checkManagedLedgerIsOpen() throws ManagedLedgerException {
...
+        if (STATE_UPDATER.get(this) != State.Fenced) {
+            STATE_UPDATER.set(this, State.Fenced);
```
nodece:
```suggestion
       State prev = STATE_UPDATER.getAndSet(this, State.Fenced);
       if (prev != State.Fenced) {
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -4416,13 +4441,21 @@ private void checkManagedLedgerIsOpen() throws ManagedLedgerException {
...
+            clearPendingAddEntries(new ManagedLedgerFencedException("ManagedLedger "
+                + name + " is fenced"));
+        }
     }
 
     synchronized void setFencedForDeletion() {
-        log.info("Moving to FencedForDeletion state");
-        STATE_UPDATER.set(this, State.FencedForDeletion);
+        log.info().attr("ledgerName", name).log("Moving to FencedForDeletion state");
```
nodece:
```suggestion
        log.info().log("Moving to FencedForDeletion state");
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -4416,13 +4441,21 @@ private void checkManagedLedgerIsOpen() throws ManagedLedgerException {
...
+        if (STATE_UPDATER.get(this) != State.FencedForDeletion) {
+            STATE_UPDATER.set(this, State.FencedForDeletion);
```
nodece:
```suggestion
       State prev = STATE_UPDATER.getAndSet(this, State.FencedForDeletion);
       if (prev != State.FencedForDeletion) {
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -2052,7 +2076,7 @@ public ManagedLedgerInterceptor getManagedLedgerInterceptor() {
         return managedLedgerInterceptor;
     }
 
-    void clearPendingAddEntries(ManagedLedgerException e) {
+    synchronized void clearPendingAddEntries(ManagedLedgerException e) {
```
nodece, line 2076: The `synchronized` keyword on `clearPendingAddEntries` also appears unnecessary, since `pendingAddEntries` is already a concurrent queue.

Additionally, the current `isEmpty()` + `poll()` sequence is not atomic and may lead to an NPE.

This can be simplified to:

```java
void clearPendingAddEntries(ManagedLedgerException e) {
    OpAddEntry op;
    while ((op = pendingAddEntries.poll()) != null) {
        op.failed(e);
    }
}
```

This avoids the race condition between `isEmpty()` and `poll()`.

nodece (changes requested):

nodece (approved):

zhaizhibo:
@nodece All checks are green, ready for merge.

PR #25242, opened 2026-02-12
[pip] PIP-455: Support Namespace Bundle Lookup and Topic Preloading

zhaizhibo:
### Modifications

Start a PIP: Support Namespace Bundle Lookup and Topic Preloading

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-455.md
+++ b/pip/pip-455.md
@@ -1,0 +1,125 @@
+
+---
+
+# PIP-455: Support Namespace Bundle Lookup and Topic Preloading
+
+
+---
+
+## Background Knowledge
+
+Apache Pulsar uses **namespace bundles** as the unit of ownership and load balancing.
+
+Key concepts:
+
+- **Namespace Bundle**: A subdivision of a namespace's hash space, representing a set of topics whose names hash into that range.
+- **Bundle Ownership**: At any given time, each bundle is owned by exactly one broker, which is responsible for serving all topics within that bundle.
+- **Lazy Topic Loading**: By default, topics are not loaded into memory until the first producer/consumer request arrives. This reduces startup overhead but increases first-call latency.
+- **PulsarAdmin & pulsar-admin CLI**: The administrative interface for managing Pulsar clusters, including operations on namespaces, topics, bundles, etc.
+
+Currently, there is **no API to proactively lookup a namespace bundle or load all topics within a bundle**. This forces users to trigger topic creation via producer/consumer requests, which is not suitable for:
+- Warm-up scenarios (preloading topics before traffic arrives)
+- Disaster recovery (forcing bundle ownership transfer and topic loading)
+- Observability (checking which broker actually owns a bundle)
+
+---
+
+## Motivation
+
+The current implementation of namespace and bundle management lacks support for **explicit lookup and preloading**. This leads to several pain points:
+
+1. **No way to warm up topics**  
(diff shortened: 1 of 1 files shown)


lhotari:
@zhaizhibo Please ensure that you have subscribed to the dev mailing list ([instructions](https://pulsar.apache.org/contact/#mailing-lists)). I released your email from the moderation queue.

zhaizhibo:
> @zhaizhibo Please ensure that you have subscribed to the dev mailing list ([instructions](https://pulsar.apache.org/contact/#mailing-lists)). I released your email from the moderation queue.

@lhotari Thanks, I have subscribed the dev mail.

pip/pip-455.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+PUT /admin/v2/namespaces/{tenant}/{namespace}/lookup
+PUT /admin/v2/namespaces/{tenant}/{namespace}/{bundle}/lookup
+```
+
+**V1 :**
+
+```
+PUT /admin/namespaces/{property}/{cluster}/{namespace}/lookup
```
codelipenghui, line 75: We don't need to support V1 API since we are working on removing the V1 endpoints from the repo.

pip/pip-455.md
```diff
@@ -0,0 +1,125 @@
...
+## High-Level Design
+
+The core idea is to extend the existing `Namespaces` admin resource to support **lookup operations at both namespace and bundle granularity**, with an optional flag to trigger topic loading.
+
+### 1. New REST Endpoints
+
+**V2 :**
+
+```
+PUT /admin/v2/namespaces/{tenant}/{namespace}/lookup
```
codelipenghui, line 68: Could you please also define the response format of this API, it will be a map which mapping the service URL for each bundle?

pip/pip-455.md
```diff
@@ -0,0 +1,125 @@
...
+CompletableFuture<LookupDataInterface> lookupNamespaceBundleAsync(...)
+```
+
+### 3. CLI Extensions
+
+**New `lookup` command in `pulsar-admin namespaces`:**
+
+```bash
+# Load all bundles in a namespace
+pulsar-admin namespaces lookup my-tenant/my-ns
```
shibd, line 110: One concern I have with the namespace-level lookup / preloading flow is bundle placement fairness before traffic actually arrives.

For pulsar-admin namespaces lookup my-tenant/my-ns, if the implementation just iterates bundles and triggers lookup/ownership acquisition one by one, I’m not sure this will produce an even bundle distribution across brokers in the cold-start case. At preload time, these bundles may have little or no traffic, so the load manager may not have enough signal to place them in a way that reflects the future workload. In the worst case, a single broker could end up owning many preloaded bundles, and then when traffic arrives later we would need another rebalance cycle, which brings us back to the same lazy-loading / cold-start problem that this PIP is trying to avoid.

Could the PIP clarify this part of the design a bit more?

How should namespace-level preload avoid concentrating too many bundles on one broker?
Is the expected behavior to reuse the normal bundle assignment logic as-is, or should there be a more explicit “spread-first” strategy for this admin-triggered warm-up path?

Should the API return the final bundle -> broker mapping so operators can verify that the warm-up actually achieved a balanced placement?

PR #17573, opened 2022-09-09
[fix][broker] Fix unexpected subscription deletion caused by the cursor last active time not updated in time

dragonls:
Fixes #17572

### Motivation

The `lastActive` in `ManagedCursorImpl` only be updated in 3 places:
- `org.apache.pulsar.broker.service.persistent.PersistentSubscription#addConsumer`
- `org.apache.pulsar.broker.service.persistent.PersistentSubscription#removeConsumer`
- `org.apache.bookkeeper.mledger.impl.ManagedCursorImpl#internalResetCursor`

If there are brokers shutdown with OOM or other situation, which make the broker not shutdown gracefully, the new broker load the topics and then do subscription expiry check soon before consumer reconnected, then the subscription may be deleted unexpectedly.

Need to update `lastActive` in `ManagedCursorImpl` while consuming stably, such as consumer acknowledged messages, which makes `lastActive` closer to the last active meaning.

### Modifications

- Add  `cursor.updateLastActive()` in `org.apache.pulsar.broker.service.persistent.PersistentSubscription#acknowledgeMessage`.
- Add some logs about cursor last active timestamp.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change added tests and can be verified as follows:
  - *`org.apache.pulsar.broker.service.persistent.PersistentTopicTest#testUpdateCursorLastActive`*
  - *`org.apache.pulsar.broker.service.persistent.PersistentSubscriptionTest#testAcknowledgeUpdateCursorLastActive`*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

PR #22794, opened 2024-05-29
[fix] [broker] maintain last active info in memory only.

thetumbled:
### Motivation

 PR https://github.com/apache/pulsar/pull/17573 has fixed cases that consumer commit offset periodically. It update `lastActive` field of cursor when consumer commit the offset.
But there is still corner case that the consumer do not commit the offset for any reason. For example,  there is no content in the topic could be read, so the consumer do not commit the offset reasonably. In such case we has no reason delete the subscription.
Anyway, **we should be cautious with the delete operation, which could result into data duplication or loss.**

This is the second solution trying to solve problem described in https://github.com/apache/pulsar/pull/21692.
PR https://github.com/apache/pulsar/pull/21692 try to update the last active info every time broker do expiration check.
But strictly speaking, we can't fix all the cases. If we fail to persist the cursor info, the incorrect sub deletion occurs too.
Incorrect deletion of subscription causes serious consequence, such as duplication or data lost. 
In my opinion, **missing deletion are better than incorrect deletion**, so this pr try to avoid all the incorrect deletion even there is posibility of missing deletion.

### Modifications

- Do not recover last active info from cursor info in zookeeper.
- Do not store the last active info into zk.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/thetumbled/pulsar/pull/54

PR #25244, opened 2026-02-12
[fix][broker] The configuration subscriptionExpirationTimeMinutes does not work if the topic was reloaded frequently

poorbarcode:
### Background
1. The attributes `cursor.active` and `cursor.lastActive` are applied in different scenarios, they are not in the same fields.
- `cursor.active: boolean`: whether the consumption catched up
- `cursor.lastActive`: the latest time that the cursor was touched, such as consumer registation, consumer unregistation, reset cursors.

---

2. The usages of `cursor.lastActive`
- Relates to the configuration `broker.conf -> subscriptionExpirationTimeMinutes`. Let Broker knows whether the cursor has been idle for a long time, to correctly delete the inactive subscriptions.
- Let users know the latest time of touching the curosr

---

3. The history of changing of `cursor.lastActive`

3-1. update `cursor.lastActive` to the current timestamp when consumers register, unregister, reset cursors. Broker persists it to metadata store when closing topics.
3-2. Pulsar supports persisting cursor state to Bookies.
3-3. https://github.com/apache/pulsar/pull/17573 removed the mechanism that recovering `cursor.lastActive` from Metadata store, to fix an issue<sup>[1]</sup>

---

- **[1]**: The issue that losses messages
  - `topic.LAC`:`3:99`
  - `subscriptionExpirationTimeMinutes`: `1 day`
  - `consumers`: `["c1"]`
    - `c1` registered at `2026-02-12 00:00`
  - `cursor.markDeletedPosition`: `3:0`
    - `c1` acknowledged message `3:0`at `2026-02-12 00:01`
    - `cursor.lastActive` was set to `2026-02-12 00:01` and was persisted into Metadata store
  - `c1` do not acknowlege any messages anymore.
  - The broker was crashed at `2026-02-13 01:00`
  - The topic was reloaded up.
    - The broker recovered `cursor.lastActive` to `2026-02-12 00:01`
    - Since the time range `2026-02-12 00:01 ~ 2026-02-13 01:00` is longer than `1 day`, the subscription was deleted, but the consumer is still connected until the broker crashed.
  - Messages `3:0(exclusive) ~ 3:99` were lost

---

### Motivation

**Issues 1**. The inactive subscription can not be deleted automaticaily if bundle rebalances frequently
  - `subscriptionExpirationTimeMinutes`: `1 day`
  - `2026-02-12 00:00`: all consumers were offline at 
  - `2026-02-12 12:00`: rebalanced: the topic was reloaded
    -  since https://github.com/apache/pulsar/pull/17573 removed the mechanism that persists `cursor.lastActive`, the value of `cursor.lastActive` will be reset to `system.currentMillis`
  - `2026-02-13 06:00`: rebalanced: the topic was reloaded
    -  since https://github.com/apache/pulsar/pull/17573 removed the mechanism that persists `cursor.lastActive`, the value of `cursor.lastActive` will be reset to `system.currentMillis`
  - The subscription can never be deleted, even though it takes much longer than the threshold time.

**Issue 2**:  when adding the new feature that Pulsar supports persisting cursor state to Bookies, it forgot to persist `cursor.lastActive` to Bookies, which will also cause the issue that inactive subscription can not be deleted automaticaily.

### Modifications
- Regarding to **Issue 2**, persist `cursor.lastActive` to Bookie also, since the field is typed `optional`, there is no compatibilty issues in multi versions
- Regarding to**Issue 1**, change the mechanism as follows:
  - Update `cursor.lastActive` to a negative value (`v = 0 - system.currentMillis`), when all consumers are unregister. 
    - It is a marker that means the inactive time.
  - Update `cursor.lastActive` when consumers register.
    - It is the original meanning of the field. 
  - How to determine whether the subscription should be deleted.
    -  `cursor.lastActive` is negative and `{current time} - Math.abs(cursor.lastActive) > {threshold}`
  - Persists `cursor.lastActive` when the first consumer registers, to avoid the issue that https://github.com/apache/pulsar/pull/17573 fixed.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java
@@ -74,9 +74,31 @@ public interface ManagedCursor {
 
     /**
      * Update the last active time of the cursor.
+     * @Deprecated call {@link #updateLastActive(boolean)} instead/
+     */
+    @Deprecated
+    void updateLastActive();
+
+    /**
+     * Update the last active time of the cursor.
+     * @param active when update latest active time, the subscription is not active(in other words, all consumers are
+     *              unregistered), please set the value to "false".
      *
      */
-    void updateLastActive();
+    default CompletableFuture<Void> updateLastActive(boolean active) {
+        updateLastActive();
+        return CompletableFuture.completedFuture(null);
+    }
+
+    /**
+     * Get the last active time or inactive time of the cursor.
+     *
+     * @return If the value is negative, it indicates the last inactive time and has not been active since then.
+     *     If the value is positive, it indicates the last active time (the last time delete was called, markDelete).
+     */
+    default long getLastActiveOrInActive() {
+        return getLastActive();
+    }
 
     /**
      * Return any properties that were associated with the last stored position.
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -310,6 +310,9 @@ public class ManagedCursorImpl implements ManagedCursor {
     private final Clock clock;
 
     // The last active time (Unix time, milliseconds) of the cursor
+    // If the value is negative, it indicates the last inactive time and has not been active since then.
+    // If the value is positive, it indicates the last active time (the last time delete was called, markDelete).
+    // If "lastActive & 1" is "1", it means that the value has been persisted, otherwise, it just was changed in memory.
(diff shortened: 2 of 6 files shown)


nodece:
#22794 makes `cursor.lastActive` memory-only, which avoids the cursor being deleted unexpectedly. This PR seems to change that behavior. It is safer not delete than to risk data loss.

nodece:
@thetumbled, could you review this PR?

lhotari:
One additional detail about `subscriptionExpirationTimeMinutes` is that it won't work if the topic isn't active (there's no connected producers or other subscriptions with connected consumers on the topic). It would also be useful to document this detail although it's not directly related to this PR.

PR #25245, opened 2026-02-13
[improve][admin] Apply all configurations instead of partial ones

nodece:
### Motivation

Both `PulsarAdminTool` and `PulsarClientTool` manually extracted a fixed subset of TLS-related properties from the config file and applied them one-by-one to the builder. This approach has two problems:

1. **Incomplete config loading** — only a hardcoded subset of properties (TLS/keystore fields) were read. Any other configuration in the file was silently ignored.
2. **Code duplication** — the same property-extraction boilerplate was duplicated across both tools.

Since `PulsarAdmin.builder()` and `PulsarClient.builder()` already support `loadConf(Map<String, Object>)` which handles all recognized config keys, the manual extraction is unnecessary.

### Modifications

- **`PulsarAdminTool`**: Replace manual property extraction with `Properties` → `Map` conversion + `PulsarAdmin.builder().loadConf(conf)`.
- **`PulsarClientTool`**: Replace manual property extraction and per-field builder calls with `Properties` → `Map` conversion + `PulsarClient.builder().loadConf(conf)`.
- Remove unused field declarations and imports in both classes.

### DIFF
--- a/pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/PulsarClientTool.java
+++ b/pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/PulsarClientTool.java
@@ -22,6 +22,8 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.FileInputStream;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Properties;
 import lombok.Getter;
 import lombok.SneakyThrows;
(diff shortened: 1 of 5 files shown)


pulsar-client-tools/src/main/java/org/apache/pulsar/client/cli/PulsarClientTool.java
```diff
@@ -96,23 +98,8 @@ public static class RootParams {
...
-    boolean useKeyStoreTls;
-    String tlsTrustStoreType;
-    String tlsTrustStorePath;
-    String tlsTrustStorePassword;
-    String tlsKeyStoreType;
-    String tlsKeyStorePath;
-    String tlsKeyStorePassword;
-    String sslFactoryPlugin;
-    String sslFactoryPluginParams;
+    protected Properties properties;
```
lhotari: Just wondering how does this get used? `initCommander` has `Properties properties` as argument.

nodece: Good catch. This PR has been updated. Both the `PulsarAdminPropertiesProvider` and `PulsarClientPropertiesProvider` provide properties.

nodece:
@lhotari, could you have a chance to review this PR?

dao-jun (approved):

PR #25187, opened 2026-01-27
[improve][meta] PIP-453: Improve the metadata store threading model

BewareMyPower:
PIP: #25173

### Modifications

1. Trigger callbacks in the underlying metadata store's internal thread directly.
2. Add a dedicated scheduler for batch flushing tasks
3. Add a dedicated thread pool for `MetadataCache`'s computing tasks
4. Add tests to verify serialization and deserialization tasks are executed in the pool whose thread count is specified by `metadataStoreSerDesThreads`

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #25246, opened 2026-02-13
[fix][meta] Metadata cache refresh might not take effect

BewareMyPower:
### Motivation

https://github.com/apache/pulsar/pull/25187 introduces a regression on `MetadataCacheImpl#refresh`. When a path is created or modified, the `refresh` method will be called by updating `objCache` with the future returned by `readValueFromStore()`. However, there is a race:
1. [current thread] Acquire the lock of the node in the `ConcurrentMap` of `objCache`. Then call `readValueFromStore`, which calls `store.get`
2. [metadata store worker thread] In the callback of `store.get`, `objCache.getIfPresent` calls the `ConcurrentMap#get` to get the cached future and check if it's the same with the future inserted to the map
3. [current thread] Insert the future returned by to the `ConcurrentMap` of `objCache` and release the lock

The updated future of step 3 is not guaranteed to be immediately visible in step 2 because `ConcurrentMap#get` is lock-free, which means it does not need to wait the lock on `path` is released after step 3.

https://github.com/apache/pulsar/blob/9db31cca40e3948182386b6930907daf449abd41/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/cache/impl/MetadataCacheImpl.java#L156-L162

The correctness of the code above is based on the fact that the `cachedFuture` must be the future of `readValueFromStore`, but not the existing cached future.

This logic was added originally because `testCloneInReadModifyUpdateOrCreate` failed. I've thought it's caused by duplicated `refresh` calls but it's actually not. However, the root cause is that when it failed, the `create` method implementation was wrong:

```java
        return serialize(path, value).thenAcceptAsync(content -> store.put(path, content, Optional.of(-1L)))
                .thenApply(stat -> objCache.get(path))
                /* ... */
                .thenApply(__ -> null);
```

It should be `thenComposeAsync` rather than `thenAcceptAsync`, otherwise `objCache.get` might not see the updated value. This bug was accidentally fixed in my later commits.

### Modifications

- Revert the `cachedFuture` related code. We might need a careful design to reduce unnecessary deserializations when `refresh` is called multiple times for an update (e.g. the callback of `MetadataCacheImpl#put` and the notification method of `AbstractMetadataStore#accept`)
- Add `testRefreshRace` to verify this race disappears.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/resourcegroup/ResourceQuotaCalculatorImplTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/resourcegroup/ResourceQuotaCalculatorImplTest.java
@@ -113,6 +113,9 @@ public class ResourceQuotaCalculatorImplTest extends MockedPulsarServiceBaseTest
 
     @Test
     public void testNeedToReportLocalUsage() {
+        // This field is 0 when the schedulers of `ResourceGroupService` are not started, then `needToReportLocalUsage`
+        // will always return true. Set it to a non-zero value to test the percentage change logic.
+        ResourceGroupService.maxIntervalForSuppressingReportsMSecs = 60 * 1000L; // 1 minute
         // If the percentage change (increase or decrease) in usage is more than 5% for
         // either bytes or messages, send a report.
         Assert.assertFalse(rqCalc.needToReportLocalUsage(1040, 1000, 104, 100, System.currentTimeMillis()));


BewareMyPower:
@lhotari Do you know why `ResourceQuotaCalculatorImplTest` now run on Broker Group 2? I checked the previous workflows and found it should run in BROKER_FLAKY workflow. https://github.com/apache/pulsar/actions/runs/21906841628/job/63249299032

Anyway, I pushed a PR to fix this broken test (not flaky): https://github.com/apache/pulsar/pull/25247

dao-jun (approved):

pulsar-metadata/src/test/java/org/apache/pulsar/metadata/MetadataCacheTest.java
```diff
@@ -739,4 +741,18 @@ public void testNoBackoffMetadataCacheConfig() {
...
+    @Test
+    public void testRefreshRace() throws Exception {
+        @Cleanup final var store = new LocalMemoryMetadataStore("memory:local", MetadataStoreConfig.builder().build());
+        final var cache = store.getMetadataCache(String.class);
+        for (int i = 0; i < 500; i++) {
+            final var key = "/key" + i;
+            assertTrue(cache.get(key).get().isEmpty());
+
+            store.put(key, "\"value\"".getBytes(StandardCharsets.UTF_8), Optional.empty()).get();
+            Awaitility.await().pollInterval(Duration.ofMillis(1)).atMost(Duration.ofSeconds(3)).untilAsserted(() ->
```
gaoran10, line 754: Why does the test need to wait here?

BewareMyPower: The `put` operation is called on the `MetadataStore` object. The `MetadataCache#accept` method is not guaranteed to be called immediately after the `put` operation is done because the notification callback is executed in the dedicated `metadata-store-event` thread.

See PIP-453 for the new thread model (though before PIP-453, it's also executed in that thread)

lhotari: btw. Wait could be avoided when `cache.put` is used. Internally, it will wait until the update has completed and then refresh the cache by reading from the store again. However I don't think it's necessary to add this test in this PR since I'd like to get 4.0.9 release candidate 2 and 4.1.3 release candidate 2 started asap.

```java
    @Test
    public void testRefreshRaceCacheOnly() throws Exception {
        @Cleanup final var store = new LocalMemoryMetadataStore("memory:local", MetadataStoreConfig.builder().build());
        final var cache = store.getMetadataCache(String.class);
        for (int i = 0; i < 500; i++) {
            final var key = "/key" + i;
            assertTrue(cache.get(key).get().isEmpty());

            cache.put(key, "\"value\"", EnumSet.noneOf(CreateOption.class)).get();
            assertTrue(cache.get(key).get().isPresent(), "Failed at key " + key);
        }
    }
```

BewareMyPower: I intentionally called `store.put` rather than `cache.put`. This test is added mainly to verify even if the side effect only comes from the notification listener of `AbstractMetadataStore#accept`, it still works. Hence it should not rely on the semantics of the MetadataCache `put` API, which refreshes the cache before the future is done.

PR #25247, opened 2026-02-14
[fix][test] Fix ResourceQuotaCalculatorImplTest#testNeedToReportLocalUsage

BewareMyPower:
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/resourcegroup/ResourceQuotaCalculatorImplTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/resourcegroup/ResourceQuotaCalculatorImplTest.java
@@ -113,6 +113,9 @@ public class ResourceQuotaCalculatorImplTest extends MockedPulsarServiceBaseTest
 
     @Test
     public void testNeedToReportLocalUsage() {
+        // This field is 0 when the schedulers of `ResourceGroupService` are not started, then `needToReportLocalUsage`
+        // will always return true. Set it to a non-zero value to test the percentage change logic.
+        ResourceGroupService.maxIntervalForSuppressingReportsMSecs = 60 * 1000L; // 1 minute
         // If the percentage change (increase or decrease) in usage is more than 5% for
         // either bytes or messages, send a report.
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/resourcegroup/ResourceQuotaCalculatorImplTest.java
```diff
@@ -113,6 +113,9 @@ public void testRQCalcGlobUsedZeroTest() throws PulsarAdminException {
 
     @Test
     public void testNeedToReportLocalUsage() {
+        // This field is 0 when the schedulers of `ResourceGroupService` are not started, then `needToReportLocalUsage`
+        // will always return true. Set it to a non-zero value to test the percentage change logic.
+        ResourceGroupService.maxIntervalForSuppressingReportsMSecs = 60 * 1000L; // 1 minute
```
Denovo1998, line 118: Should maxIntervalForSuppressingReportsMSecs be saved and restored here? (Although it appears that only this test directly depends on it)

```java
long old = ResourceGroupService.maxIntervalForSuppressingReportsMSecs;
try {
    ResourceGroupService.maxIntervalForSuppressingReportsMSecs = TimeUnit.MINUTES.toMillis(1);
    // asserts...
} finally {
    ResourceGroupService.maxIntervalForSuppressingReportsMSecs = old;
}
```

BewareMyPower: No need to apply this complicated logic here because only `needToReportLocalUsage` depends on this config and it's explained in the comments here.

Actually, we should initialize it when initializing the namespace in `ResourceGroupNamespaceConfigListener`(`registerNameSpace` -> `maybeStartSchedulers`). Then it will be set according to the config.

```
            maxIntervalForSuppressingReportsMSecs =
                    TimeUnit.SECONDS.toMillis(this.resourceUsagePublishPeriodInSeconds) * MaxUsageReportSuppressRounds;
```

But it's out of the scope of this PR. Since it's just a simple unit test, we can keep it simple as setting this variable manually

PR #25248, opened 2026-02-16
feat: support write tombstone to a compactTopic.

zhaizhibo:
### Motivation

Currently, Pulsar's topic compaction process retains all keys, including those with null values. This creates a limitation where deleted keys (tombstones) are never removed from compacted ledgers, causing them to grow indefinitely over time. This behavior consumes unnecessary storage and impacts system performance for long-running topics with frequent key deletions.

### Modifications

Added a new optional broker configuration `topicCompactionRetainNullValue` to control whether null values should be retained during compaction.

**Default behavior**: `true` (retain null values) - maintains backward compatibility
**New option**: When set to `false`, keys with null values (tombstones) will be removed during compaction
**Backward compatibility**: If the configuration is not set, the system defaults to the original behavior (retain null values)

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
@@ -748,6 +748,61 @@ public class CompactionTest extends MockedPulsarServiceBaseTest {
         }
     }
 
+    @Test(timeOut = 30000)
+    public void testNotToRetainNullValue() throws Exception {
+        conf.setTopicCompactionRetainNullValue(false);
+        conf.setTopicCompactionRetainNullKey(true);
+        restartBroker();
+        FieldUtils.writeField(compactor, "topicCompactionRetainNullValue", false, true);
+        FieldUtils.writeField(compactor, "topicCompactionRetainNullKey", true, true);
+
+        String topic = "persistent://my-tenant/my-ns/not-to-retain-null-value";
+
+        // subscribe before sending anything, so that we get all messages
+        pulsarClient.newConsumer().topic(topic).subscriptionName("sub1")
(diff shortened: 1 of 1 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
```diff
@@ -748,6 +748,61 @@ public void testKeyLessMessagesPassThrough(boolean retainNullKey) throws Excepti
         }
     }
 
+    @Test(timeOut = 30000)
+    public void testNotToRetainNullValue() throws Exception {
+        conf.setTopicCompactionRetainNullValue(false);
+        conf.setTopicCompactionRetainNullKey(true);
+        restartBroker();
+        FieldUtils.writeField(compactor, "topicCompactionRetainNullValue", false, true);
+        FieldUtils.writeField(compactor, "topicCompactionRetainNullKey", true, true);
```
lhotari, line 757: It would be better to override `restartBroker` in this class so that it reinitializes `compactor` so that there wouldn't be a need to set the fields directly.

lhotari:
I looked into the change history and it looks like there used to be a related bug which was fixed a long time with #18877.

lhotari:
There's also #24523 since Pulsar 4.1.0, implementing [PIP-429: Optimize Handling of Compacted Last Entry by Skipping Payload Buffer Parsing](https://github.com/apache/pulsar/blob/master/pip/pip-429.md)

PR #25258, opened 2026-02-25
[feat][broker] add topic-level subscription expiration time policy

codelipenghui:
### Motivation: 

Add topic-level override support for subscription expiration time on TopicPolicies. Changes: admin API/client methods, broker hierarchy cache wiring, v2 topic policy endpoints, runtime effective-value lookup for inactive subscription cleanup, topicPolicies CLI commands, and comprehensive tests for precedence/authz/disable/cli coverage. 

### Testing

focused compile passed and focused new/updated tests passed.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentTopicTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/PersistentTopicTest.java
@@ -2058,6 +2058,7 @@ public class PersistentTopicTest extends MockedBookKeeperTestCase {
 
     @Test
     public void testCheckInactiveSubscriptions() throws Exception {
+        pulsarTestContext.getConfig().setSubscriptionExpirationTimeMinutes(5);
         PersistentTopic topic = new PersistentTopic(successTopicName, ledgerMock, brokerService);
 
         final var subscriptions = new ConcurrentHashMap<String, PersistentSubscription>();
@@ -2092,8 +2093,6 @@ public class PersistentTopicTest extends MockedBookKeeperTestCase {
         pulsarTestContext.getPulsarResources().getNamespaceResources()
                 .createPolicies(TopicName.get(successTopicName).getNamespaceObject(),
                         new Policies());
-
-        pulsarTestContext.getConfig().setSubscriptionExpirationTimeMinutes(5);
 
         doReturn(System.currentTimeMillis() - TimeUnit.MINUTES.toMillis(6)).when(cursorMock).getLastActive();
 


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -3481,6 +3481,37 @@ protected CompletableFuture<Void> internalSetDeduplication(Boolean enabledToSet,
...
+                        .orElseGet(() -> {
+                            if (applied) {
+                                Integer namespacePolicy = getNamespacePolicies(namespaceName)
+                                        .subscription_expiration_time_minutes;
+                                return namespacePolicy == null
+                                        ? config().getSubscriptionExpirationTimeMinutes()
+                                        : namespacePolicy;
+                            }
+                            return null;
+                        }));
```
lhotari, line 3496: What is `applied` supposed to do?
It currently only gets used when topic policies are completely missing. 
I'd assume the logic to get applied also in the case where `subscriptionExpirationTimeInMinutes` hasn't been set in topic policies. Is that what would be the intention?

lhotari, line 3496: I'd assume that this would be resolve by something like this:
```suggestion
                .thenApply(op -> Optional.ofNullable(op.map(TopicPolicies::getSubscriptionExpirationTimeInMinutes).orElse(null))
                        .orElseGet(() -> {
                            if (applied) {
                                Integer namespacePolicy = getNamespacePolicies(namespaceName)
                                        .subscription_expiration_time_minutes;
                                return namespacePolicy == null
                                        ? config().getSubscriptionExpirationTimeMinutes()
                                        : namespacePolicy;
                            }
                            return null;
                        }));
```

lhotari: It looks like all other methods contain a similar bug. I'll file it separately since it should be handled in a separate PR.

lhotari: It looks like all other methods contain a similar bug. I'll file it separately since it should be handled in a separate PR.

lhotari: Checked this. `Optional.map` already handles this so there wasn't any issue after all.

PR #25261, opened 2026-02-25
[improve][client] Support protobuf v4 schema compatibility

### LINKED ISSUE #25202: Support for protobuf gencode v4.x
### Search before reporting

### Motivation

Pulsar's protobuf schema options are dependent on `GeneratedMessageV3`, which is [deprecated](https://www.javadoc.io/doc/com.google.protobuf/protobuf-java/4.28.0-RC1/com/google/protobuf/GeneratedMessageV3.html) and will be fully removed in version 5.x of the runtimes. Gencode >v4.x creates classes that extend `GeneratedMessage` instead, meaning anyone using the protobuf schema options are currently locked out of updating their gencode past v3.x, which will cease getting new updates this year.

If there is already an option I'm unaware of that enables the use of the newer gencode, please let me know.

Marking this as not similar to #24388 as that only discusses a solution for the runtime upgrade and not the gencode.

Thank you in advance!

### Solution

Add a topic schema option that supports `GeneratedMessage` as a type

### Alternatives

_No response_

### Anything else?

_No response_

codelipenghui:
### Motivation

Fixes #25202.

Pulsar protobuf schema APIs were bounded to GeneratedMessageV3, which breaks usage with protobuf v4 generated classes (for example Any).

### Modifications

- Widened protobuf schema-related generic bounds from GeneratedMessageV3 to Message in client API/binding and implementations. 
- Updated protobuf schema reader/writer/native classes and runtime validation checks to use Message. 
- Kept strict generated-message validation via descriptor/default-instance reflective paths.
- Added API signature regression test to lock in Message upper bounds.
- Added/updated protobuf schema tests for client and native paths.
- Updated pulsar-functions protobuf class detection to work with protobuf v4 (Message-based detection).
- Added protobuf-v4 required CI job for client + functions tests using v4 property overrides.

### Documentation 

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/pulsar-ci.yaml
+++ b/.github/workflows/pulsar-ci.yaml
@@ -337,6 +337,8 @@ jobs:
             group: CLIENT
           - name: Pulsar Metadata
             group: METADATA
+          - name: Protobuf v4
+            group: PROTOBUFV4
 
     steps:
       - name: checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/pulsar-ci.yaml
```diff
@@ -536,6 +536,104 @@ jobs:
...
+            -Dspotbugs.skip=true \
+            -Dlicense.skip=true \
+            -Dcheckstyle.skip=true \
+            -Drat.skip=true \
+            -Dprotobuf3.version=4.31.1 \
+            -Dprotoc3.version=4.31.1 \
+            -pl pulsar-client,pulsar-functions/instance \
+            -Dtest=org.apache.pulsar.client.api.ProtobufSchemaApiSignatureTest,org.apache.pulsar.client.impl.schema.ProtobufSchemaTest,org.apache.pulsar.client.impl.schema.ProtobufNativeSchemaTest,org.apache.pulsar.functions.source.TopicSchemaTest,org.apache.pulsar.functions.instance.JavaInstanceRunnableTest \
+            -Dsurefire.failIfNoSpecifiedTests=false \
+            test
```
lhotari: This should be added to `build/run_unit_group.sh` instead

.github/workflows/pulsar-ci.yaml
```diff
@@ -536,6 +536,104 @@ jobs:
         with:
           action: wait
 
+  protobuf-v4-tests:
+    name: CI - Protobuf v4 - Client + Functions
```
lhotari: There's no need for a separate step when the commands to run the test are added to `build/run_unit_group.sh`. In that case these tests could be run as part of the `unit-tests` matrix, in a similar way as other tests are run.

lhotari: I addressed this.

.github/workflows/pulsar-ci.yaml
lhotari: I addressed this.

PR #25171, opened 2026-01-21
[feat][broker] Expose configuration to allow to read marker type messages

zymap:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Motivation

When using RawReader to read messages, we want to read all the messages from the broker. And handle the marker type on the reader side.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #25270, opened 2026-02-27
[feat][broker] PIP-456: Add configuration to allow RawReader to read marker messages

zymap:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-456.md
+++ b/pip/pip-456.md
@@ -1,12 +1,14 @@
-# PIP-448: Add subscriptionPrefixToSkipServerMarkerCheck configuration to allow RawReader to read marker messages
+# PIP-456: Add enableReadingMarkerMessage option to RawReader to allow reading marker messages
 
 # Background knowledge
 
 In Apache Pulsar, **marker messages** are internal messages used by the broker for various operations such as compaction, replication, and transactions. These markers are not meant to be visible to end users in normal consumer reads.
 
-The broker filters out marker messages before delivering them to consumers in most cases. The filtering logic is implemented in `AbstractBaseDispatcher.java`, which checks if a message is a marker type and skips delivering it to regular consumers.
+The broker filters out marker messages before delivering them to consumers. The filtering logic is implemented in `AbstractBaseDispatcher.java`, which checks if a message is a marker type and decides whether to deliver it to the subscription.
 
 **RawReader** is a low-level reader API that allows reading raw messages directly from a topic without any client-side filtering or processing. It's useful for administrative tasks, debugging, and implementing custom message processing pipelines.
+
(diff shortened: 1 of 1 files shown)


zymap:
Here is the mail: https://lists.apache.org/thread/p1gzj9doq7o5fnn30yxbn5ks70ybbdf9

zymap:
Here is the impl: https://github.com/apache/pulsar/pull/25171

pip/pip-456.md
```diff
@@ -0,0 +1,190 @@
...
+
+No new broker configuration is required. The feature is controlled entirely via the client-side option.
+
+## Public-facing Changes
+
+### Public API
+
+| API | Description |
+|-----|-------------|
+| `ConsumerConfigurationData.setEnableReadingMarkerMessage(boolean)` | Set whether to read marker messages |
```
shibd, line 125: I'm wondering whether adding `enableReadingMarkerMessages` to `ConsumerConfigurationData` is the right approach. This field would become part of the `public API surface`, but it has no meaningful use case for **regular consumers** — it's only relevant for `RawReader`.

To avoid leaking an internal-only concern into the general consumer API, I'd suggest keeping it out of ConsumerConfigurationData entirely and instead exposing it directly on RawReader:


To avoid leaking an internal-only concern into the general consumer API, I'd suggest keeping it out of ConsumerConfigurationData entirely and instead exposing it directly on RawReader:


```java
// RawReader.java
public static CompletableFuture<RawReader> create(PulsarClient client, String topic, String subscription,
    boolean createTopicIfDoesNotExist, boolean retryOnRecoverableErrors,
    boolean enableReadingMarkerMessages);
```


Then set the subscription property internally inside RawReaderImpl:

```java
public RawReaderImpl(PulsarClientImpl client, String topic, String subscription,
                     CompletableFuture<Consumer<byte[]>> consumerFuture,
                     boolean createTopicIfDoesNotExist, boolean retryOnRecoverableErrors,
                     boolean enableReadingMarkerMessages) {
    consumerConfiguration = new ConsumerConfigurationData<>();
    // ...
    if (enableReadingMarkerMessages) {
        consumerConfiguration.getSubscriptionProperties()
            .put("enableReadingMarkerMessages", "true");
    }
    consumer = new RawConsumerImpl(client, consumerConfiguration, consumerFuture,
        createTopicIfDoesNotExist, retryOnRecoverableErrors);
}

```

Issue #13382, opened 2021-12-17
Setting `PULSAR_EXTRA_OPTS` causes the process to run without significant default JVM options


lhotari:
**Describe the bug**
When setting `PULSAR_EXTRA_OPTS`, the expectation would be that the JVM options would be added to the existing JVM arguments. However, this is not the case.

If `PULSAR_EXTRA_OPTS` isn't set, these parameters will get set `-Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacity.default=1000 -Dio.netty.recycler.linkCapacity=1024` .
If you set `PULSAR_EXTRA_OPTS`, these parameters will be omitted. For example, the `pulsar.allocator.exit_on_oom` setting will exit the JVM if the PulsarByteBufAllocator cannot allocate a ByteBuf.

**Expected behavior**
Setting `PULSAR_EXTRA_OPTS` shouldn't remove the default settings, but instead add to them. One possible solution would be to make the PULSAR_EXTRA_OPTS setting in pulsar_env.sh to be made this way:
```
# Extra options to be passed to the jvm
PULSAR_EXTRA_OPTS="-Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacity.default=1000 -Dio.netty.recycler.linkCapacity=1024 ${PULSAR_EXTRA_OPTS}"
```

**Additional context**
Other improvements for environment variable handling was recently done in PR #13025

sijie:
@lhotari: in the opposite, people should have the ability to overwrite the defaults. I don't think keeping the defaults is a good idea. If people want to update the `PULSAR_EXTRA_OPTS`, they should just put all the settings they need there.

eolivelli:
probably we could have two variables in order to full fill both of the needs

lhotari:
> @lhotari: in the opposite, people should have the ability to overwrite the defaults. I don't think keeping the defaults is a good idea. If people want to update the `PULSAR_EXTRA_OPTS`, they should just put all the settings they need there.

I agree.

@sijie  Do you see a problem in the behavior that setting `PULSAR_EXTRA_OPTS` causes the process to run without significant default JVM options, for example `-Dpulsar.allocator.exit_on_oom=true`? 

How does anyone know that relevant settings get overridden when setting PULSAR_EXTRA_OPTS?

Issue #25021, opened 2025-11-26
[Bug] Update netty allocator options to avoid oom when using EntryFilters


geniusjoe:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

Broker version: 3.0LTS, 4.0 LTS
Broker Operating system and hardware type: 5.4.241-1-tlinux4 (cloud vendor host Linux distributions, acted like CentOS)
Broker Java version: jdk 21
Client library type: Java
Client library version: apache-pulsar-3.0.4
Client Operating system and hardware type: 5.4.241-1-tlinux4 
Client Java version: openjdk version "17.0.12" 2024-07-23 LTS

### Issue Description

One of our maintained Pulsar instances's brokers (with a 16C32G spec.) frequently experience native memory OOM.
However, the memory usage shown in the `jcmd <pid> native_memory summary` does not match the `top` command process `RES` usage. Typically, just before an OOM occurred, `top` shows memory usage of around 31GB, while the total memory usage in the `jcmd <broker-pid> native_memory summary` is only about 20GB. Pulsar broker restarts due to OOM is abnormal and need to find out the reason.

<img width="2612" height="1208" alt="Image" src="https://github.com/user-attachments/assets/4eeb1209-ebcb-4b62-977f-6e97db3df51d" />

### Error messages
Process was killed by K8S OOMKiller and print error log in `kubectl describe pod` command  "Last State section" part: `CGroup OOM encountered, victim process: java, pid: <broker-pid>`

### Reproducing the issue

I believe OOM scenario is related to the broker frequently reading large-sized messages from bookies and then quickly release them. This scenario is more likely to occurred in broadcast use case(with large amount of subscriptions and high filter ratio). Below is one of reproduce ways:
1. Use 1 broker with 8C16G 5Gbps NIC hardware and 3 bookie, configure `E:Qw:Qa = 2:2:1` for improving bookie's message throughput 
2. Create approximately 1000 subscriptions, each subscription only creates one consumer, and add unique properties to each subscription
3. Send a 1MB message every 1 second, and each message's properties matching only one random subscription's properties among these 1000 subscriptions
[...]

PR #25274, opened 2026-02-27
[improve][broker] Update netty allocator default chunk size from 4MB to 8MB, and move internal JVM opts from PULSAR_EXTRA_OPTS to OPTS

geniusjoe:
Main Issue: #25021

### Motivation
Netty changed the default chunk size from 16M to 4M in version 4.1.76. The primary reason was that the Netty community believed that in most scenarios, each PoolArena does not require very large chunk sizes, as it could easily lead to wasted memory resources in this pr https://github.com/netty/netty/pull/12108, , which I think is not very suitable for pulsar use case:
> If we don't need a lot of memory, and certainly not compared to the number of cores on a system, then this will take up more memory than necessary, since each chunk is 16 MiB.

In practice, the Logstash community also already reverted it back to the original default value of 16MB. https://github.com/elastic/logstash/issues/15765

### Modifications
Refer to comment https://github.com/apache/pulsar/issues/25021#issuecomment-3584736093:

> Increasing the default chunk size to 8MB (-Dio.netty.allocator.maxOrder=10) should be something to consider first since that would be the minimal change to mitigate the issue.

In order to maintain compatibility with versions prior to Netty 4.1.76 as much as possible, and to improve chunk utilization (where the chunk size exceeds the default maximum single message size of 5 MB), the Netty configuration `-Dio.netty.allocator.maxOrder=10` is added in `pulsar.env` to set the chunk size to 8 MB.

### Verifying this change

testDefaultChunkSizeMatchesMaxOrder10

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation
- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/bin/bookkeeper
+++ b/bin/bookkeeper
@@ -214,7 +214,8 @@ fi
 # improving memory utilization and reducing the frequency of requesting off-heap memory from the OS
 #
 # Based on the netty source code, the allocator's default chunk size is calculated as:
-# io.netty.allocator.pageSize(default: 8192) << io.netty.allocator.maxOrder(default: 9 after Netty 4.1.76.Final version).
+# io.netty.allocator.pageSize (default: 8192) shifted left by
+# io.netty.allocator.maxOrder (default: 9 after Netty 4.1.76.Final version).
 # This equals 8192 * 2^9 = 4 MB：
 # https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L105
 #
@@ -225,7 +226,8 @@ fi
 # Therefore, when frequently producing messages whose size exceeds the chunk size,
 # Netty cannot utilize resources from the memory pool and must frequently allocate native memory.
 # This can lead to increased physical memory fragmentation and higher reclamation costs.
-# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger than chunk size(8MB) and can reuse Netty's memory pool.
+# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger
+# than chunk size (8MB) and can reuse Netty's memory pool.
 OPTS="-Dio.netty.leakDetection.level=disabled -Dio.netty.recycler.maxCapacityPerThread=4096 -Dio.netty.allocator.maxOrder=10 $OPTS"
 
 OPTS="$OPTS $BOOKIE_MEM $BOOKIE_GC $BOOKIE_GC_LOG $BOOKIE_EXTRA_OPTS"
--- a/bin/function-localrunner
+++ b/bin/function-localrunner
@@ -146,7 +146,8 @@ OPTS="-XX:+ExitOnOutOfMemoryError -Dpulsar.allocator.exit_on_oom=true $OPTS"
 # improving memory utilization and reducing the frequency of requesting off-heap memory from the OS
 #
 # Based on the netty source code, the allocator's default chunk size is calculated as:
-# io.netty.allocator.pageSize(default: 8192) << io.netty.allocator.maxOrder(default: 9 after Netty 4.1.76.Final version).
+# io.netty.allocator.pageSize (default: 8192) shifted left by
+# io.netty.allocator.maxOrder (default: 9 after Netty 4.1.76.Final version).
 # This equals 8192 * 2^9 = 4 MB：
 # https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L105
 #
@@ -157,7 +158,8 @@ OPTS="-XX:+ExitOnOutOfMemoryError -Dpulsar.allocator.exit_on_oom=true $OPTS"
 # Therefore, when frequently producing messages whose size exceeds the chunk size,
 # Netty cannot utilize resources from the memory pool and must frequently allocate native memory.
 # This can lead to increased physical memory fragmentation and higher reclamation costs.
-# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger than chunk size(8MB) and can reuse Netty's memory pool.
+# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger
+# than chunk size (8MB) and can reuse Netty's memory pool.
 OPTS="-Dio.netty.recycler.maxCapacityPerThread=4096 -Dio.netty.allocator.maxOrder=10 $OPTS"
 
 OPTS="$OPTS $PULSAR_EXTRA_OPTS $PULSAR_MEM $PULSAR_GC"
--- a/bin/pulsar
+++ b/bin/pulsar
@@ -309,7 +309,8 @@ OPTS="-XX:+ExitOnOutOfMemoryError -Dpulsar.allocator.exit_on_oom=true $OPTS"
 # improving memory utilization and reducing the frequency of requesting off-heap memory from the OS
 #
 # Based on the netty source code, the allocator's default chunk size is calculated as:
-# io.netty.allocator.pageSize(default: 8192) << io.netty.allocator.maxOrder(default: 9 after Netty 4.1.76.Final version).
+# io.netty.allocator.pageSize (default: 8192) shifted left by
+# io.netty.allocator.maxOrder (default: 9 after Netty 4.1.76.Final version).
 # This equals 8192 * 2^9 = 4 MB：
 # https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L105
 #
@@ -320,7 +321,8 @@ OPTS="-XX:+ExitOnOutOfMemoryError -Dpulsar.allocator.exit_on_oom=true $OPTS"
 # Therefore, when frequently producing messages whose size exceeds the chunk size,
 # Netty cannot utilize resources from the memory pool and must frequently allocate native memory.
 # This can lead to increased physical memory fragmentation and higher reclamation costs.
-# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger than chunk size(8MB) and can reuse Netty's memory pool.
+# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger
+# than chunk size (8MB) and can reuse Netty's memory pool.
 OPTS="-Dio.netty.recycler.maxCapacityPerThread=4096 -Dio.netty.allocator.maxOrder=10 $OPTS"
 
 OPTS="-cp $PULSAR_CLASSPATH $OPTS"
--- a/conf/pulsar_env.sh
+++ b/conf/pulsar_env.sh
@@ -89,15 +89,6 @@ if [[ -z "$PULSAR_GC_LOG" ]]; then
   fi
 fi
 
-# Extra options to be passed to the jvm, typically used for passing user-defined JVM OPTS configurations.
-# This PULSAR_EXTRA_OPTS parameter has a higher priority than the predefined JVM OPTS in the Pulsar bin files.
-# Therefore, in addition to defining extra JVM parameters here,
-# it can also be used to override JVM parameters with the same name in OPTS
-# For example, if IPv6 functionality needs to be enabled, configure:
-# PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS} -Djava.net.preferIPv4Stack=false"
-# If you need to obtain more details, you can refer to the generation strategy of `OPTS` in the `bin/pulsar` file
-# PULSAR_EXTRA_OPTS=
-
 # Add extra paths to the bookkeeper classpath
 # PULSAR_EXTRA_CLASSPATH=
 
--- a/bin/pulsar-admin-common.cmd
+++ b/bin/pulsar-admin-common.cmd
@@ -92,7 +92,8 @@ REM These settings are primarily used to modify the Netty allocator configuratio
 REM improving memory utilization and reducing the frequency of requesting off-heap memory from the OS
 REM
 REM Based on the netty source code, the allocator's default chunk size is calculated as:
-REM io.netty.allocator.pageSize(default: 8192) shifted left by io.netty.allocator.maxOrder(default: 9 after Netty 4.1.76.Final version).
+REM io.netty.allocator.pageSize (default: 8192) shifted left by
+REM io.netty.allocator.maxOrder (default: 9 after Netty 4.1.76.Final version).
 REM This equals 8192 * 2^9 = 4 MB：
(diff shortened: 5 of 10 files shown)


conf/pulsar_env.sh
```diff
@@ -90,7 +90,7 @@ if [[ -z "$PULSAR_GC_LOG" ]]; then
 fi
 
 # Extra options to be passed to the jvm
-PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS:-" -Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096"}"
+PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS:-" -Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096 -Dio.netty.allocator.maxOrder=10"}"
```
lhotari: This would result in the expected behavior that the default chunk size wouldn't always be set. If the user sets `PULSAR_EXTRA_OPTS`, this setting wouldn't be used. There's an old issue #13382 about the problem.

The solution for this problem would be to prepend the arguments to `OPTS` so that the defaults could be overridden with `PULSAR_EXTRA_OPTS` or `OPTS`.

```shell
OPTS="-Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096 -Dio.netty.allocator.maxOrder=10 $OPTS"
```

This would also resolve #13382.

A similar solution was implemented in #23846.

bin/function-localrunner
```diff
@@ -76,8 +76,11 @@ if [[ -z "$PULSAR_GC_LOG" ]]; then
   fi
 fi
 
-# Extra options to be passed to the jvm
-PULSAR_EXTRA_OPTS=${PULSAR_EXTRA_OPTS:-" -Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096"}
+# Extra options to be passed to the jvm，typically used for passing user-defined JVM OPTS configurations.
+# This PULSAR_EXTRA_OPTS parameter has a higher priority than the predefined JVM OPTS.
+# Therefore, in addition to defining extra JVM parameters here, it can also be used to override JVM parameters with the same name in OPTS
+# For example, if IPv6 functionality needs to be enabled, configure: PULSAR_EXTRA_OPTS="-Djava.net.preferIPv4Stack=false".
+PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS}"
```
lhotari: this can be removed from the function-localrunner script. `PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS}"` is not useful.
The comments in `function-localrunner` aren't needed.

bin/pulsar
```diff
@@ -300,6 +300,14 @@ OPTS="$OPTS --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/jdk
 OPTS="$OPTS --add-opens java.base/jdk.internal.platform=ALL-UNNAMED"
 # Required by RocksDB java.lang.System::loadLibrary call
 OPTS="$OPTS --enable-native-access=ALL-UNNAMED"
+# These two settings work together to ensure the Pulsar process exits immediately and predictably
+# if it runs out of either Java heap memory or its internal off-heap memory,
+# as these are unrecoverable errors that require a process restart to clear the faulty state and restore operation
+OPTS="-XX:+ExitOnOutOfMemoryError -Dpulsar.allocator.exit_on_oom=true $OPTS"
+# Netty tuning
+# These settings are primarily used to modify the Netty allocator configuration,
+# improving memory utilization and reducing the frequency of requesting off-heap memory from the OS
```
lhotari, line 309: for the `bin/pulsar` script it would be useful to explain at least the unit used for io.netty.allocator.maxOrder and the impact of it. Unfortunately there doesn't seem to be Netty documentation that could be referenced directly.

Based on the source code, the default chunk size for the allocator will be the value of io.netty.allocator.pageSize (default: 8192) << io.netty.allocator.maxOrder (default set here: 10).
8192 << 10 = 8192 * 2^10 = 8192 * 1024 = 8388608 = 8 MB
https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L105

Allocations that are larger than chunk size are considered huge allocations and don't use the pool:
https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PoolArena.java#L141-L142

Summarizing and simplifying this information would be useful.

conf/pulsar_env.sh
```diff
@@ -89,8 +89,13 @@ if [[ -z "$PULSAR_GC_LOG" ]]; then
...
 
-# Extra options to be passed to the jvm
-PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS:-" -Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096"}"
+# Extra options to be passed to the jvm, typically used for passing user-defined JVM OPTS configurations.
+# This PULSAR_EXTRA_OPTS parameter has a higher priority than the predefined JVM OPTS in the Pulsar bin files.
+# Therefore, in addition to defining extra JVM parameters here,
+# it can also be used to override JVM parameters with the same name in OPTS
+# For example, if IPv6 functionality needs to be enabled, configure:
+# PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS} -Djava.net.preferIPv4Stack=false"
+PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS}"
```
lhotari: This can be removed. The documentation for `PULSAR_EXTRA_OPTS` should be in `bin/pulsar` where it matters the most.

This line isn't useful at all
```
PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS}"
```

deployment/terraform-ansible/templates/pulsar_env.sh
```diff
@@ -48,7 +48,7 @@ PULSAR_MEM=" -Xms{{ max_heap_memory }} -Xmx{{ max_heap_memory }} -XX:MaxDirectMe
 PULSAR_GC=" -XX:+UseZGC -XX:+PerfDisableSharedMem -XX:+AlwaysPreTouch"
 
 # Extra options to be passed to the jvm
-PULSAR_EXTRA_OPTS="-Dio.netty.leakDetection.level=disabled -Dio.netty.recycler.maxCapacityPerThread=4096 ${PULSAR_EXTRA_OPTS}"
+PULSAR_EXTRA_OPTS="-Dio.netty.leakDetection.level=disabled -Dio.netty.recycler.maxCapacityPerThread=4096 -Dio.netty.allocator.maxOrder=10 ${PULSAR_EXTRA_OPTS}"
```
lhotari: This can be removed completely since `pulsar.broker.service` uses `bin/pulsar broker` to start the broker and settings in `bin/pulsar` are used. There's no need to set `PULSAR_EXTRA_OPTS`.

tests/integration/src/test/java/org/apache/pulsar/tests/integration/containers/PulsarContainer.java
```diff
@@ -301,7 +301,9 @@ protected void passNettyLeakDetectionSystemProperties() {
 
     protected void initializePulsarExtraOpts() {
         appendToEnv("PULSAR_EXTRA_OPTS",
-                "-Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096");
+                "-Dpulsar.allocator.exit_on_oom=true "
+                        + "-Dio.netty.recycler.maxCapacityPerThread=4096 "
+                        + "-Dio.netty.allocator.maxOrder=10");
```
lhotari: This can be removed. Just leave the empty method `initializePulsarExtraOpts`.

lhotari (commented):
Thanks for making good progress, @geniusjoe . I added some more review comments. 
btw. Please don't force push changes to a PR that is under review with comments since that will make it hard to follow changes between partial reviews. It's better to push individual commits after the review has started. The PR will be squashed eventually into a single commit when merging, so you don't have to worry about the git history made in the PR.

conf/pulsar_env.sh
geniusjoe: Fixed. The JVM-related parameters for `PULSAR_EXTRA_OPTS` have been migrated from `conf/bkenv.sh`, `conf/pulsar_env.sh`, and `conf/pulsar_tools_env.sh` to the corresponding `OPTS` variables in their respective `bin/` scripts.

bin/pulsar
geniusjoe: Fixed. A description of the Netty parameter `-Dio.netty.allocator.maxOrder=10` has been added to the related scripts

bin/function-localrunner
geniusjoe: The duplicate initialization code for `PULSAR_EXTRA_OPTS`, along with its corresponding comments, has been removed.

deployment/terraform-ansible/templates/pulsar_env.sh
geniusjoe: The `PULSAR_EXTRA_OPTS` configuration has been removed.

tests/integration/src/test/java/org/apache/pulsar/tests/integration/containers/PulsarContainer.java
geniusjoe: Removed

conf/pulsar_env.sh
geniusjoe: @lhotari 
First of all, thank you very much for your review, and I apologize for affecting your review by using force-push. I will not use force-push mode for commits in the future.

I think that for a Pulsar release version, users during initial use may not focus heavily on the internal JVM implementation details in the `bin/` directory. Maybe they are likely to pay more attention to custom configurations in the `conf/` directory. 

For these three environment files `bkenv.sh`, `pulsar_env.sh`, and `pulsar_tools_env.sh` in `conf/`, should we consider referencing the approach used in other configurations? Specifically, we could comment out the variables while retaining the explanatory comments, informing users that they can override JVM configurations in `OPTS` here, and do not place the relevant comments in the corresponding `bin/` files. Taking `conf/pulsar_env.sh` as an example:
```
# Extra options to be passed to the jvm, typically used for passing user-defined JVM OPTS configurations.
# This PULSAR_EXTRA_OPTS parameter has a higher priority than the predefined JVM OPTS in the Pulsar bin files.
# Therefore, in addition to defining extra JVM parameters here,
# it can also be used to override JVM parameters with the same name in OPTS
# For example, if IPv6 functionality needs to be enabled, configure:
# PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS} -Djava.net.preferIPv4Stack=false"
# If you need to obtain more details, you can refer to the generation strategy of `OPTS` in the `bin/pulsar` file
# PULSAR_EXTRA_OPTS=
```

lhotari: @geniusjoe There's already a documentation section in `bin/pulsar`:
https://github.com/apache/pulsar/blob/d992dc530bea4f4737da93f1527133b1c9389a45/bin/pulsar#L77-L95
It feels like the correct place to document the environment variable configuration.
In addition, there should be documentation in https://pulsar.apache.org/docs .

I agree that users focusing on bare metal / VM deployments would look into `conf` directory. However, 
I'd assume that the majority of Pulsar users use Kubernetes for deployment. In that case, the `pulsar_env.sh` isn't very useful. Setting environment variables directly is more convenient. That's another reason why environment variables should be documented in Pulsar's documentation.

conf/pulsar_env.sh
```diff
@@ -89,8 +89,14 @@ if [[ -z "$PULSAR_GC_LOG" ]]; then
...
-# Extra options to be passed to the jvm
-PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS:-" -Dpulsar.allocator.exit_on_oom=true -Dio.netty.recycler.maxCapacityPerThread=4096"}"
+# Extra options to be passed to the jvm, typically used for passing user-defined JVM OPTS configurations.
+# This PULSAR_EXTRA_OPTS parameter has a higher priority than the predefined JVM OPTS in the Pulsar bin files.
+# Therefore, in addition to defining extra JVM parameters here,
+# it can also be used to override JVM parameters with the same name in OPTS
+# For example, if IPv6 functionality needs to be enabled, configure:
+# PULSAR_EXTRA_OPTS="${PULSAR_EXTRA_OPTS} -Djava.net.preferIPv4Stack=false"
+# If you need to obtain more details, you can refer to the generation strategy of `OPTS` in the `bin/pulsar` file
+# PULSAR_EXTRA_OPTS=
```
lhotari: I think that this line `PULSAR_EXTRA_OPTS=` should be removed since it's not a useful example. 
```suggestion
```

bin/bookkeeper
```diff
@@ -209,6 +209,24 @@ OPTS="-Djava.net.preferIPv4Stack=true $OPTS"
...
+# https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L105
+#
+# Allocations that are larger than chunk size are considered huge allocations and don't use the pool:
+# https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PoolArena.java#L141-L142
+#
+# Currently, Pulsar defaults to a maximum single message size of 5 MB.
+# Therefore, when frequently producing messages whose size exceeds the chunk size,
+# Netty cannot utilize resources from the memory pool and must frequently allocate native memory.
+# This can lead to increased physical memory fragmentation and higher reclamation costs.
+# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger than chunk size(8MB) and can reuse Netty's memory pool.
```
lhotari: add word wrapping here for increased readability (on terminal):
```suggestion
# Thus, increasing io.netty.allocator.maxOrder to 10 to ensure that a single message is larger
than chunk size (8MB) and can reuse Netty's memory pool.
```
Please also update other locations.

conf/pulsar_env.sh
geniusjoe: @lhotari 
I think you make a valid point. I have removed the `EXTRA_OPTS` related explanations from the `*env.sh` files. I'm also considering creating a new PR later, to add descriptions for the variables in the files when deploying Pulsar instances in a bare metal environment for the following two documents:
https://pulsar.apache.org/docs/next/getting-started-standalone/
https://pulsar.apache.org/docs/next/deploy-bare-metal/#configure-brokers

conf/pulsar_env.sh
geniusjoe: Removed.

bin/bookkeeper
geniusjoe: Fixed. Line breaks have been added to single-line comments that were of excessive length.

PR #25276, opened 2026-02-28
[fix][broker] Support namespace unsubscribe when bundles are unloaded

zhanghaou:
### Motivation

Unsubscribing at namespace/bundle level fails when the target bundle is not loaded by any broker.

This PR makes unsubscribe paths handle unloaded bundles by acquiring ownership when needed and loading topics before unsubscribing.

### Modifications

1. Convert namespace/bundle unsubscribe paths to async internal APIs and update v1/v2 admin endpoints to use async response flow.
2. Allow bundle ownership acquisition for unsubscribe operations when bundle is not loaded.
3. Load topics from metadata and ensure they are in memory before removing subscriptions.
4. Add tests to verify unsubscribe works on unloaded bundles (both namespace and bundle scope).

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -1929,6 +1929,9 @@ public abstract class NamespacesBase extends AdminResource {
                     return FutureUtil.waitForAll(futures);
                 }).exceptionally(ex -> {
                     Throwable cause = FutureUtil.unwrapCompletionException(ex);
+                    if (cause instanceof RestException) {
+                        throw (RestException) cause;
+                    }
                     if (cause instanceof SubscriptionBusyException) {
                         throw new RestException(Status.PRECONDITION_FAILED,
                                 "Subscription has active connected consumers");


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -1918,32 +1899,52 @@ private void clearBacklog(NamespaceName nsName, String bundleRange, String subsc
...
+                                    if (sub == null) {
+                                        return CompletableFuture.completedFuture(null);
+                                    }
+                                    return sub.delete();
+                                }));
+                    }
+                    return FutureUtil.waitForAll(futures);
+                }).exceptionally(ex -> {
+                    Throwable cause = FutureUtil.unwrapCompletionException(ex);
+                    log.error("[{}] Failed to unsubscribe {} for namespace {}/{}", clientAppId(), subscription,
```
nodece: The same error logging block is repeated in the base class and both versioned implementations (v1 and v2).

nodece (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
zhanghaou: The redundant log has been removed. Thanks for the review.

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AdminApiTest.java
```diff
@@ -2339,6 +2339,92 @@ public void testUnsubscribeOnNamespace(Integer numBundles) throws Exception {
                 new ArrayList<>());
     }
 
+    @Test(dataProvider = "numBundles")
+    public void testUnsubscribeNamespaceBundleOnUnloadedBundle(Integer numBundles) throws Exception {
```
Denovo1998, line 2343: Perhaps we can add some unit tests.
1) partitioned topic (ensure all subscriptions are deleted on each partition);
2) Multi-topic cross-bundle;

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -1918,32 +1899,50 @@ private void clearBacklog(NamespaceName nsName, String bundleRange, String subsc
...
-            }
-            throw new RestException(e.getCause());
+    private CompletableFuture<Void> unsubscribeAsync(NamespaceName nsName, String bundleRange, String subscription) {
+        if (subscription.startsWith(pulsar().getConfiguration().getReplicatorPrefix())) {
+            return CompletableFuture.failedFuture(
+                    new RestException(Status.PRECONDITION_FAILED, "Cannot unsubscribe a replication cursor"));
         }
+
+        return pulsar().getNamespaceService().getFullListOfTopics(nsName)
+                .thenCompose(topicsInNamespace -> {
```
Denovo1998: Can the logic of NamespaceService#getOwnedTopicListForNamespaceBundle be reused here (or switch to using targetBundle.includes)?

pulsar-broker/src/test/java/org/apache/pulsar/broker/admin/AdminApiTest.java
zhanghaou: More test cases had added.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
zhanghaou: Thanks for the review. NamespaceService#getOwnedTopicListForNamespaceBundle is more appropriate.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -1918,32 +1899,42 @@ private void clearBacklog(NamespaceName nsName, String bundleRange, String subsc
...
+                                }));
+                    }
+                    return FutureUtil.waitForAll(futures);
+                }).exceptionally(ex -> {
+                    Throwable cause = FutureUtil.unwrapCompletionException(ex);
+                    if (cause instanceof SubscriptionBusyException) {
+                        throw new RestException(Status.PRECONDITION_FAILED,
+                                "Subscription has active connected consumers");
+                    }
+                    throw new RestException(cause);
```
codelipenghui, line 1939: Should also check if the `cause` is a RestException to avoid wraps it in another
  RestException

codelipenghui (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
zhanghaou: Good point. I’ve added a guard to check cause instanceof RestException and throw it directly

PR #25278, opened 2026-03-03
[cleanup] Refactored Backoff to be more consistent and intuitive to use

merlimat:
### Motivation

Refactored `Backoff` class to force consistent usage through builder and to use `Duration` instead of long, to avoid any time units mistakes.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java
@@ -50,7 +50,7 @@ import lombok.Getter;
  */
 public class Backoff {
     private static final Duration DEFAULT_INITIAL_DELAY = Duration.ofMillis(100);
-    private static final Duration DEFAULT_MAX_BACKOFF_INTERVAL = Duration.ofSeconds(30);
+    private static final Duration DEFAULT_MAX_BACKOFF_INTERVAL = Duration.ofMinutes(1);
     private static final Random random = new Random();
 
     @Getter
@@ -77,6 +77,16 @@ public class Backoff {
         if (initial.isZero() && max.isZero() && mandatoryStop.isZero()) {
             this.mandatoryStopMade = true;
         }
+    }
+
+    /**
+     * Creates a {@link Backoff} with the default configuration (initial delay 100 ms, max 1 min,
+     * no mandatory stop).
+     *
+     * @return a new Backoff with default settings
+     */
+    public static Backoff create() {
+        return new Builder().build();
     }
 
     /**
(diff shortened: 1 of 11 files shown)


pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java
```diff
@@ -19,108 +19,194 @@
...
-            if (initial == current) {
-                firstBackoffTimeInMillis = now;
+            Instant now = clock.instant();
+            Duration timeElapsedSinceFirstBackoff = Duration.ZERO;
+            if (initial.equals(current)) {
+                firstBackoffTime = now;
             } else {
-                timeElapsedSinceFirstBackoff = now - firstBackoffTimeInMillis;
+                timeElapsedSinceFirstBackoff = Duration.between(firstBackoffTime, now);
             }
```
Denovo1998, line 124: Here, as long as the returned delay (current) for this round equals initial, firstBackoffTime will be refreshed to "now".

This implies an assumption: current == initial will only occur during the "first next()". However, this assumption is likely invalid, as current == initial can occur in multiple scenarios.

---

1. **Assume building a constant backoff (initial = maximum):**

initialDelay = 100ms
maxBackoff = 100ms
mandatoryStop = 5s

So every time next():

- current = next = 100ms

- Since current.compareTo(max) < 0 is false (equal), next does not increase and remains 100ms

- In the mandatoryStop logic, initial.equals(current) is always true

- Every time firstBackoffTime = now is executed

- elapsed is always 0 (or close to 0), so elapsed + current > mandatoryStop is never true

- Result: Even if real time has passed for tens of seconds, mandatoryStop is not triggered (mandatoryStopMade remains false)

2. **reduceToHalf() halves the next delay, but not below initial. This produces an unintuitive side effect:**

- Assume initial=100ms, during a backoff process, next has increased to 200ms

- Call reduceToHalf(), it will change next back to 100ms (== initial)

- The next time next() is called, current will again equal initial

- So the code will execute firstBackoffTime = now (equivalent to "restarting the timer")

But from the semantic perspective of the comments, reduceToHalf does not say that the mandatoryStop timer window should be reset. Only reset() should reset it.

Therefore, using initial.equals(current) as the "first call determination" will cause: the firstBackoffTime might be reset even without calling reset, thus making mandatoryStop more lenient (or even never reaching it).

---

Perhaps the judgment for the first next() should be:

- If firstBackoffTime == Instant.EPOCH, it means this period has not started timing -> set it
- Otherwise, calculate elapsed using it
- When reset() is called, also reset firstBackoffTime back to Instant.EPOCH (otherwise, the old start point will be used after reset)

merlimat: @Denovo1998 This refactor does not (should not..) change the logic of the backoff. Any other changes we can do after this.

Denovo1998: +1. I think I can mention a PR about this issue.

PR #25280, opened 2026-03-04
[feat][ml] Support Bookkeeper Batch Read API

dao-jun:
### Motivation

In [BP-62](https://github.com/apache/bookkeeper/pull/4051), BookKeeper started to support a batch read API,
where a single RPC fetches multiple entries. This PR uses it on the Pulsar side for managed-ledger storage
reads, reducing the per-entry request overhead of sequential reads.

### Modifications

1. Support the BookKeeper batch read API in the managed-ledger read path:
   - `ReadEntryUtils` issues `batchReadUnconfirmedAsync` requests and assembles multi-round,
     size-limited partial results while preserving the existing "read the full range or fail"
     contract of the storage read; a result with no entries at all is handed over as-is, like the
     result of a regular read, so the existing handling of empty reads (#24515) still applies
   - `CompositeLedgerEntriesImpl` merges the per-round `LedgerEntries` into a single result
     with clear buffer ownership (closing the merged result releases all the underlying batches)
   - Both `RangeEntryCacheImpl` (entry cache enabled) and `EntryCacheDisabled` (entry cache
     disabled) use the batch read path
2. The size limit of each batch read request comes from the read that triggered it, not from a
   fixed configuration value: the `maxSizeBytes` that dispatchers already pass to
   `ManagedCursor.asyncReadEntriesOrWait` (e.g. `dispatcherMaxReadSizeBytes`) is now carried by
   `OpReadEntry` and plumbed through `EntryCache.asyncReadEntry`, `PendingReadsManager` and
   `RangeEntryCacheImpl.readFromStorage` down to the batch read request. Reads without a size limit
   let the BookKeeper client cap each request at its netty max frame size.
3. Batch read is enabled by default, both in `ManagedLedgerConfig` and through the new
   `managedLedgerBatchReadEnabled` broker setting (dynamic). It requires the v2 wire protocol and
   BookKeeper's own batch read flag: the managed ledger checks the BookKeeper client's configuration
   once when it is opened (`BookKeeperClientConfigAccessor`, in the client's package since the accessor
   is protected) and uses regular reads otherwise, so clients on the v3 protocol never issue a batch
   read (BookKeeper 4.18.1 rejects it on its executor and the read would hang). The BookKeeper client
   itself uses regular reads for striped ledgers and on a ledger when a bookie does not support batch
   reads. Each request is also bounded by the client's max frame size (`maxMessageSize` plus padding),
   so a read needing more data is split into sequential requests. The prerequisites are documented in
   `broker.conf`, `standalone.conf` and the `ServiceConfiguration` field doc.
4. Entries read in a batch are slices of one response frame, which stays allocated as long as any of
   them is referenced; the entry cache therefore copies batch-sourced entries into cache-owned buffers
   (the existing `copyEntries` branch), so cache residency matches the accounted size.

### Verifying this change

- `ReadEntryUtilsTest`: single-request and multi-request assembly, partial / empty / excessive /
  non-contiguous batch results, synchronous failures surfaced through the returned future, buffer
  release on every failure and cancellation path, the no-size-limit case, and the fallbacks when
  batch read is disabled, when a single entry is read, and when reading via a read-only managed ledger
- `CompositeLedgerEntriesImplTest`: iteration, `getEntry` lookups and idempotent close across
  multiple underlying `LedgerEntries` containers
- `RangeEntryCacheImplTest#testReadFromStorageWithBatchReadEnabled`: the storage read path
  with batch read enabled and the caller's size limit passed to the batch read request
- `ManagedLedgerTest#testBatchReadRequiresClientSupport`: the client check with the mock client
- With batch read enabled by default, the whole managed-ledger and broker test suites exercise the
  batch read path through `PulsarMockBookKeeper`; `ManagedLedgerBkTest` uses the v2 wire protocol so
  its reads go through the real BookKeeper batch read API, and
  `testBatchReadNotUsedWithV3WireProtocolClient` verifies a v3 client reads without it.

Batch read is enabled by default: the new `managedLedgerBatchReadEnabled` broker setting and
`ManagedLedgerConfig.batchReadEnabled` default to `true`.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

PR in forked repository:


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/ReadEntryUtils.java
```diff
@@ -49,6 +61,79 @@ static CompletableFuture<LedgerEntries> readAsync(ManagedLedger ml, ReadHandle h
...
+                        log.debug("Batch read partial result for ledger {}: received {}/{}, reading remaining {}-{}",
+                                handle.getId(), receivedCount, maxCount, nextEntryId, lastEntry);
+                    }
+
+                    // Recursively read remaining entries
+                    return batchReadWithAutoRefill(handle, nextEntryId, lastEntry, remainingCount, maxSize)
+                            .thenApply(remainingEntries -> {
+                                // Combine received and remaining entries
+                                List<LedgerEntry> combined = new ArrayList<>(receivedCount + remainingCount);
+                                combined.addAll(receivedList);
```
codelipenghui: The entry in the receivedList will get released in line:115?

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -2332,6 +2332,12 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
     )
     private boolean managedLedgerCacheEvictionExtendTTLOfRecentlyAccessed = true;
 
+    @FieldContext(category = CATEGORY_STORAGE_ML,
+            doc = "Enable batch read API when reading entries from bookkeeper. "
+                    + "Batch read allows reading multiple entries in a single RPC call, "
+                    + "reducing network overhead for sequential reads.")
+    private boolean managedLedgerBatchReadEnabled = true;
```
codelipenghui, line 2712: It should be false as the PR's description?

codelipenghui (dismissed):
The EntryCacheDisabled.java should also adopt the new method?

And from the test perspective, this PR has 0 test coverage for the core batch read logic.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/ReadEntryUtils.java
```diff
@@ -49,6 +61,79 @@ static CompletableFuture<LedgerEntries> readAsync(ManagedLedger ml, ReadHandle h
...
+                    if (receivedCount == 0) {
+                        // Edge case: no entries returned, use regular read
+                        entries.close();
+                        log.warn("Batch read returned 0 entries for ledger {} entries {}-{}, falling back to "
+                                + "regular read", handle.getId(), firstEntry, lastEntry);
+                        return handle.readUnconfirmedAsync(firstEntry, lastEntry);
+                    }
+
+                    // Close the original entries since we've collected them into receivedList
+                    entries.close();
```
coderzc: `receivedList` stores the original LedgerEntry references, but `entries.close()` is called before the merged result is returned. Since closing `LedgerEntriesImpl` also closes the contained entries, the merged result may end up holding entries whose buffers have already been released. Please preserve independent ownership of the retained entries (for example by retaining/duplicating them), or defer closing the original container until the merged result is consumed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2252,6 +2252,9 @@ public CompletableFuture<ManagedLedgerConfig> getManagedLedgerConfig(@NonNull To
                         serviceConfig.isCacheEvictionByMarkDeletedPosition());
                 managedLedgerConfig.setCacheEvictionByExpectedReadCount(false);
             }
+            managedLedgerConfig.setBatchReadEnabled(
```
void-ptr974, line 2255: This should probably be gated by the final BookKeeper ClientConfiguration instead of ServiceConfiguration alone. `BookKeeperClientFactoryImpl#createBkClientConfiguration` later applies `bookkeeper_` passthrough properties, so an operator can set `managedLedgerBatchReadEnabled=true` and `bookkeeper_useV2WireProtocol=false`. In that case ML enables batch reads, but the actual BK client uses the v3 protocol path where batch reads are unsupported. Please either reject this conflicting configuration or derive the ML batch-read flag from the actual BK client config.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/ReadEntryUtils.java
```diff
@@ -49,6 +57,66 @@ static CompletableFuture<LedgerEntries> readAsync(ManagedLedger ml, ReadHandle h
...
+        List<LedgerEntry> receivedEntries = new ArrayList<>(maxCount);
+        List<LedgerEntries> ledgerEntries = new ArrayList<>(4);
+        doBatchRead(handle, firstEntry, maxCount, maxSize, receivedEntries, ledgerEntries, future);
+        return future;
+    }
+
+    private static void doBatchRead(ReadHandle handle, long firstEntry, int maxCount, int maxSize,
+                                    List<LedgerEntry> receivedEntries, List<LedgerEntries> ledgerEntries,
+                                    CompletableFuture<LedgerEntries> future) {
+        handle.batchReadUnconfirmedAsync(firstEntry, maxCount - receivedEntries.size(), maxSize)
```
void-ptr974: `batchReadUnconfirmedAsync` can throw synchronously, for example when the actual BK client uses the non-v2 path where `PerChannelBookieClient` throws `UnsupportedOperationException` for batch reads. Since this method returns `CompletableFuture`, Pulsar convention is to surface failures through the returned future. Please wrap this call and complete the future exceptionally, or fallback to `readUnconfirmedAsync`.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/cache/ReadEntryUtils.java
```diff
@@ -49,6 +57,66 @@ static CompletableFuture<LedgerEntries> readAsync(ManagedLedger ml, ReadHandle h
...
+                        return;
+                    }
+                    long lastReceivedEntry = -1;
+                    int prevReceivedCount = receivedEntries.size();
+                    for (LedgerEntry entry : entries) {
+                        receivedEntries.add(entry);
+                        lastReceivedEntry = entry.getEntryId();
+                    }
+                    ledgerEntries.add(entries);
+                    if (receivedEntries.size() >= maxCount || prevReceivedCount == receivedEntries.size()) {
```
void-ptr974: This can complete successfully with fewer entries than the caller requested. BK batch read may legally return a partial prefix, but the previous `readAsync(firstEntry, lastEntry)` contract for these callers is effectively "read the full range or fail". If a later batch returns empty after some entries were already collected, this code currently returns the partial result. Please verify `receivedEntries.size() == maxCount` before completing successfully, or fallback/fail when the full range cannot be assembled.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerConfig.java
```diff
@@ -91,6 +92,36 @@ public class ManagedLedgerConfig {
...
+     * Batch read allows reading multiple entries in a single RPC call, reducing network overhead.
+     * Note: Batch read is only effective when ensembleSize equals writeQuorumSize (non-striped ledgers).
+     */
+    @Setter
+    private boolean batchReadEnabled = false;
+
+    /**
+     * Max size in bytes for per-batch read request. If set to 0 or negative,
+     * uses the netty max frame size (default 5MB).
+     * Batch read may return fewer entries if total size exceeds this limit.
```
void-ptr974, line 106: The comment says `batchReadMaxSizeBytes <= 0` uses the Netty max frame size, but `ReadEntryUtils` only enables batch read when `batchReadMaxSize > 0`. Please align the implementation and the config contract, either by passing non-positive values through to BK so it can apply its fallback, or by updating this documentation.

merlimat (approved):

PR #25282, opened 2026-03-04
[fix][misc] Fix Backoff mandatory-stop timer tracking

Denovo1998:
Fixes: https://github.com/apache/pulsar/pull/25278/changes#r2878215799

Main Issue: #xyz

PIP: #xyz 

### Motivation

Backoff's mandatory-stop timer can be unintentionally reset when `current == initial`, which happens with constant backoff (`initial == max`) and after `reduceToHalf()` brings `next` back to `initial`. This prevents `mandatoryStopMade` from ever being reached and can cause retries to run longer than the configured mandatory stop.

### Modifications

- Track the mandatory-stop start time with an explicit flag so it is set only once per backoff cycle.
- Reset the mandatory-stop timer state in `reset()`.

### Verifying this change

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: https://github.com/Denovo1998/pulsar/pull/23


pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java
```diff
@@ -95,6 +98,8 @@ public void reduceToHalf() {
 
     public void reset() {
         this.next = this.initial;
+        this.firstBackoffTimeInMillis = 0;
+        this.firstBackoffTimeInMillisSet = false;
```
codelipenghui: The PR modifies reset() to clear both firstBackoffTimeInMillis and firstBackoffTimeInMillisSet, but relies on the existing mandatoryStopTest for
  implicit coverage. An explicit test for "reset then new mandatory stop cycle" would strengthen confidence

codelipenghui (commented):
There should be similar issue on kinesis connector code.

pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/Backoff

@Denovo1998 could you please also check if the kinesis connector can use the same Backoff you fixed here? or need to apply the same fix to kinesis connector

pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java
Denovo1998: @codelipenghui Added an explicit unit test to verify that reset() starts a new mandatory-stop cycle.

Denovo1998:
> There should be similar issue on kinesis connector code.
> 
> pulsar-io/kinesis/src/main/java/org/apache/pulsar/io/kinesis/Backoff
> 
> @Denovo1998 could you please also check if the kinesis connector can use the same Backoff you fixed here? or need to apply the same fix to kinesis connector

I also applied the same fix and unit test in `pulsar-io/kinesis` Backoff.

PR #25284, opened 2026-03-04
[improve][broker] Offload web response from metadata thread for list tenants/namespaces/clusters

mattisonchao:
### Motivation

The list tenants, list namespaces, and list clusters REST endpoints currently execute `asyncResponse.resume()` on the metadata store thread. This can block the metadata thread pool and lead to deadlocks or performance degradation.

Additionally, `asyncResponse.resume()` triggers any registered JAX-RS response filters (e.g., `ContainerResponseFilter`), which can only execute synchronously. If a response filter performs metadata operations, it will block the metadata thread, further increasing the risk of deadlocks.

### Modifications

Offload the response handling (sorting, filtering, collecting) and `asyncResponse.resume()` to the web service executor thread pool using `thenAcceptAsync`.

- `TenantsBase.getTenants()`: `thenAccept` → `thenAcceptAsync(..., webServiceExecutor)`
- `ClustersBase.getClusters()`: `thenApply` + `thenAccept` → single `thenAcceptAsync(..., webServiceExecutor)`
- `Namespaces.getTenantNamespaces()`: `thenAccept` → `thenAcceptAsync(..., webServiceExecutor)`
- `WebService`: add `@Getter` on `webServiceExecutor` field

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/ClustersBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/ClustersBase.java
@@ -92,13 +92,15 @@ public class ClustersBase extends AdminResource {
     })
     public void getClusters(@Suspended AsyncResponse asyncResponse) {
         clusterResources().listAsync()
-                .thenApply(HashSet::new)
-                .thenAccept(asyncResponse::resume)
-                .exceptionally(ex -> {
-                    log.error("[{}] Failed to get clusters {}", clientAppId(), ex);
-                    resumeAsyncResponseExceptionally(asyncResponse, ex);
+                .<Void>handleAsync((clusters, ex) -> {
+                    if (ex != null) {
+                        log.error("[{}] Failed to get clusters {}", clientAppId(), ex);
+                        resumeAsyncResponseExceptionally(asyncResponse, ex);
+                        return null;
+                    }
+                    asyncResponse.resume(clusters);
                     return null;
-                });
+                }, webExecutor());
     }
 
     @GET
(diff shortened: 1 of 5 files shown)


codelipenghui (commented):
@mattisonchao Do we also need to fix the other methods? 

Another point is if we need to fix it everywhere, is it better to have a separate thread to complete the future in the metadata component?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/ClustersBase.java
```diff
@@ -93,11 +93,10 @@ public class ClustersBase extends AdminResource {
...
-                .thenApply(clusters -> clusters.stream()
+                .thenAcceptAsync(clusters -> asyncResponse.resume(clusters.stream()
                         // Remove "global" cluster from returned list
                         .filter(cluster -> !Constants.GLOBAL_CLUSTER.equals(cluster))
-                        .collect(Collectors.toSet()))
-                .thenAccept(asyncResponse::resume)
+                        .collect(Collectors.toSet())), pulsar().getWebService().getWebServiceExecutor())
                 .exceptionally(ex -> {
                     log.error("[{}] Failed to get clusters {}", clientAppId(), ex);
                     resumeAsyncResponseExceptionally(asyncResponse, ex);
```
codelipenghui: Do we also need to use handleAsync for the exceptional path?

jiangpengcheng (approved):

shibd (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/ClustersBase.java
mattisonchao: Yes. applied.

mattisonchao:
> Another point is if we need to fix it everywhere, is it better to have a separate thread to complete the future in the metadata component?

hmm. I am still considering it. since we might have multiple higher-level `XXXXResource` operations combined for a single endpoint. Frequent context switching is also not a very good approach.

I just want to use this small PR to hotfix the issue we are having.  and the following https://github.com/apache/pulsar/pull/25286/changes is to optimise my use cases so that they do not rely on the response filter.

Plus, go back to the original question. So far, we are doing well across most of our use cases. We don't need to make many changes to existing behaviour.  

However, for the future version. A better approach would be to support vthread for all admin endpoints, so we can make a blocking call directly for the vthread without needing any callback threads.

dao-jun (approved):

PR #25287, opened 2026-03-05
[fix][client] Fix async APIs to return failed futures on validation errors

zhanghaou:
### Motivation

Most async APIs in the Pulsar client (for example ReaderBuilderImpl#createAsync, ConsumerBuilderImpl#subscribeAsync, and ProducerBase#sendAsync) report failures by completing the returned CompletableFuture exceptionally.

However, ProducerBuilderImpl#createAsync() currently uses checkArgument for validation. When batching and chunking are both enabled, it throws IllegalArgumentException synchronously, bypassing the returned future. This breaks async usage patterns because callers’ exceptionally / whenComplete handlers are never invoked.

### Modifications

Replace the synchronous checkArgument with a failure path that returns FutureUtil.failedFuture.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
@@ -662,7 +662,10 @@ public abstract class ConsumerBase<T> extends HandlerState implements Consumer<T
         }
         TransactionImpl txnImpl = null;
         if (null != txn) {
-            checkArgument(txn instanceof TransactionImpl);
+            if (!(txn instanceof TransactionImpl)) {
+                return FutureUtil.failedFuture(
+                        new IllegalArgumentException("txn should be instance of TransactionImpl"));
+            }
             txnImpl = (TransactionImpl) txn;
             CompletableFuture<Void> completableFuture = new CompletableFuture<>();
            if (!txnImpl.checkIfOpen(completableFuture)) {
@@ -691,7 +694,10 @@ public abstract class ConsumerBase<T> extends HandlerState implements Consumer<T
 
         TransactionImpl txnImpl = null;
         if (null != txn) {
-            checkArgument(txn instanceof TransactionImpl);
+            if (!(txn instanceof TransactionImpl)) {
+                return FutureUtil.failedFuture(
+                        new IllegalArgumentException("txn should be instance of TransactionImpl"));
+            }
             txnImpl = (TransactionImpl) txn;
         }
         return doAcknowledgeWithTxn(messageId, AckType.Cumulative, Collections.emptyMap(), txnImpl);
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
@@ -663,8 +663,8 @@ public abstract class ConsumerBase<T> extends HandlerState implements Consumer<T
         TransactionImpl txnImpl = null;
         if (null != txn) {
             if (!(txn instanceof TransactionImpl)) {
-                return FutureUtil.failedFuture(
-                        new IllegalArgumentException("txn should be instance of TransactionImpl"));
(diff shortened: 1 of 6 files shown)


nodece (approved):
LGTM

codelipenghui (commented):
There is one more case

public CompletableFuture<Void> unsubscribeAsync(String topicName)

Please also check this one, it should has the same issue.

zhanghaou:
> There is one more case
> 
> public CompletableFuture unsubscribeAsync(String topicName)
> 
> Please also check this one, it should has the same issue.

Done. Also fixed two similar issues in ConsumerBase.

zhanghaou:
@codelipenghui 
I found more similar issues in the broker module, and I will submit a new PR to fix them.

codelipenghui (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
```diff
@@ -662,7 +661,10 @@ public CompletableFuture<Void> acknowledgeAsync(MessageId messageId,
         }
         TransactionImpl txnImpl = null;
         if (null != txn) {
-            checkArgument(txn instanceof TransactionImpl);
+            if (!(txn instanceof TransactionImpl)) {
+                return FutureUtil.failedFuture(new IllegalArgumentException(
+                        "Expected txn to be an instance of TransactionImpl, but got " + txn.getClass().getName()));
+            }
```
lhotari, line 667: Please add a `checkArgumentAsync` utility method to FutureUtil so that code doesn't get duplicated in all cases where `Preconditions.checkArgument` was eliminated. In `checkArgumentAsync`, `Preconditions.checkArgument` can be used and the exception is simple caught.

lhotari (changes requested):

lhotari:
Also in this case, please check how to run CI in your own fork so that you don't have to wait for someone to approve running of CI. The instructions for "Personal CI" are here: https://pulsar.apache.org/contribute/personal-ci/

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
zhanghaou: @lhotari  Thanks for the review, do you mean switching to asynchronous validation like in the code below? If so, that would require quite a lot of code changes since all asynchronous tasks would need to be chained after the validation logic.
```
public static CompletableFuture<Void> checkArgumentAsync(boolean expression,
                                                             String errorMessageTemplate,
                                                             @Nullable Object @Nullable ... errorMessageArgs) {
        try {
            Preconditions.checkArgument(expression, errorMessageTemplate, errorMessageArgs);
        } catch (IllegalArgumentException e) {
            return failedFuture(e);
        }
    }
```

zhanghaou:
> Also in this case, please check how to run CI in your own fork so that you don't have to wait for someone to approve running of CI. The instructions for "Personal CI" are here: https://pulsar.apache.org/contribute/personal-ci/

Thanks for the guidance. I really appreciate it.

dao-jun (approved):

lhotari (dismissed):
<Dismissed comment>

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java
lhotari: > If so, that would require quite a lot of code changes since all asynchronous tasks would need to be chained after the validation logic. Yes that's write. Extracting a method wouldn't help in this case.

lhotari (approved):
LGTM

PR #25289, opened 2026-03-05
[fix][broker] Return failed future instead of throwing exception in async methods

zhanghaou:
same to this pr https://github.com/apache/pulsar/pull/25287

### Motivation

In Java's CompletableFuture, callbacks registered via stages such as thenApply, thenCompose, etc., are executed within internal try/catch blocks. If a callback throws an exception, the framework automatically captures it and completes the dependent future exceptionally.

Therefore, throwing an exception inside a stage callback is safe and will be translated into an exceptionally completed future.

However, if an async method itself throws an exception instead of returning a failed future, the exception is raised synchronously and no CompletableFuture is returned. This breaks the expected async error propagation model.

Some async validation paths in the broker currently throw exceptions directly instead of returning a failed CompletableFuture. This breaks the async contract of methods that return CompletableFuture, because the exception is thrown synchronously at invocation time rather than being propagated through the future.

As a result, callers relying on async error handling (e.g. exceptionally, handle, or whenComplete) may not observe these failures correctly.

### Modifications

1. Convert validation synchronous throws to failed futures in broker async methods.
2. Refactor original-principal validation to an async helper and preserve original auth/authorization semantics.
3. Ensure default async UnsupportedOperationException paths return failed futures.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloader.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloader.java
@@ -28,7 +28,6 @@ import org.apache.bookkeeper.common.annotation.InterfaceAudience;
 import org.apache.bookkeeper.common.annotation.InterfaceStability;
 import org.apache.bookkeeper.mledger.proto.MLDataFormats;
 import org.apache.pulsar.common.policies.data.OffloadPolicies;
-import org.apache.pulsar.common.util.FutureUtil;
 
 /**
  * Interface for offloading ledgers to long-term storage.
@@ -166,7 +165,7 @@ public interface LedgerOffloader {
     default CompletableFuture<OffloadHandle> streamingOffload(ManagedLedger ml, UUID uid, long beginLedger,
                                                               long beginEntry,
                                                               Map<String, String> driverMetadata) {
-        return FutureUtil.failedFuture(new UnsupportedOperationException());
+        throw new UnsupportedOperationException();
     }
 
     /**
@@ -201,11 +200,11 @@ public interface LedgerOffloader {
 
     default CompletableFuture<ReadHandle> readOffloaded(long ledgerId, MLDataFormats.OffloadContext ledgerContext,
                                                         Map<String, String> offloadDriverMetadata) {
-        return FutureUtil.failedFuture(new UnsupportedOperationException());
+        throw new UnsupportedOperationException();
     }
 
     default CompletableFuture<Void> deleteOffloaded(UUID uid, Map<String, String> offloadDriverMetadata) {
-        return FutureUtil.failedFuture(new UnsupportedOperationException());
+        throw new UnsupportedOperationException();
     }
 
     /**
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloader.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloader.java
@@ -28,6 +28,7 @@ import org.apache.bookkeeper.common.annotation.InterfaceAudience;
 import org.apache.bookkeeper.common.annotation.InterfaceStability;
 import org.apache.bookkeeper.mledger.proto.MLDataFormats;
 import org.apache.pulsar.common.policies.data.OffloadPolicies;
+import org.apache.pulsar.common.util.FutureUtil;
 
 /**
  * Interface for offloading ledgers to long-term storage.
@@ -165,7 +166,7 @@ public interface LedgerOffloader {
     default CompletableFuture<OffloadHandle> streamingOffload(ManagedLedger ml, UUID uid, long beginLedger,
                                                               long beginEntry,
                                                               Map<String, String> driverMetadata) {
-        throw new UnsupportedOperationException();
+        return FutureUtil.failedFuture(new UnsupportedOperationException());
     }
 
     /**
@@ -200,11 +201,11 @@ public interface LedgerOffloader {
 
     default CompletableFuture<ReadHandle> readOffloaded(long ledgerId, MLDataFormats.OffloadContext ledgerContext,
                                                         Map<String, String> offloadDriverMetadata) {
-        throw new UnsupportedOperationException();
+        return FutureUtil.failedFuture(new UnsupportedOperationException());
     }
 
     default CompletableFuture<Void> deleteOffloaded(UUID uid, Map<String, String> offloadDriverMetadata) {
-        throw new UnsupportedOperationException();
+        return FutureUtil.failedFuture(new UnsupportedOperationException());
     }
 
     /**
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
@@ -23,7 +23,6 @@ import static org.apache.pulsar.broker.delayed.bucket.BucketDelayedDeliveryTrack
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import lombok.AllArgsConstructor;
@@ -158,7 +157,9 @@ abstract class Bucket {
     }
 
     private CompletableFuture<Void> putBucketKeyId(String bucketKey, Long bucketId) {
-        Objects.requireNonNull(bucketId);
+        if (bucketId == null) {
+            return FutureUtil.failedFuture(new NullPointerException());
+        }
         return sequencer.sequential(() -> {
             return executeWithRetry(() -> cursor.putCursorProperty(bucketKey, String.valueOf(bucketId)),
                     ManagedLedgerException.BadVersionException.class, MaxRetryTimes);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
@@ -158,7 +158,7 @@ abstract class Bucket {
 
     private CompletableFuture<Void> putBucketKeyId(String bucketKey, Long bucketId) {
         if (bucketId == null) {
-            return FutureUtil.failedFuture(new NullPointerException());
+            return FutureUtil.failedFuture(new NullPointerException("bucketId"));
         }
         return sequencer.sequential(() -> {
             return executeWithRetry(() -> cursor.putCursorProperty(bucketKey, String.valueOf(bucketId)),
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
@@ -158,7 +158,7 @@ abstract class Bucket {
 
     private CompletableFuture<Void> putBucketKeyId(String bucketKey, Long bucketId) {
         if (bucketId == null) {
-            return FutureUtil.failedFuture(new NullPointerException("bucketId"));
+            return FutureUtil.failedFuture(new NullPointerException("Expected bucketId should not be null"));
         }
         return sequencer.sequential(() -> {
             return executeWithRetry(() -> cursor.putCursorProperty(bucketKey, String.valueOf(bucketId)),
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LoadManager.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LoadManager.java
@@ -35,7 +35,6 @@ import org.apache.pulsar.broker.lookup.LookupResult;
 import org.apache.pulsar.broker.namespace.LookupOptions;
 import org.apache.pulsar.common.naming.ServiceUnitId;
 import org.apache.pulsar.common.stats.Metrics;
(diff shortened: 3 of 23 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/web/PulsarWebResource.java
```diff
@@ -186,12 +186,14 @@ public static boolean isClientAuthenticated(String appId) {
         return appId != null;
     }
 
-    private void validateOriginalPrincipal(String authenticatedPrincipal, String originalPrincipal) {
+    private CompletableFuture<Void> validateOriginalPrincipalAsync(String authenticatedPrincipal,
```
codelipenghui: This change seems to be wrong?

codelipenghui (dismissed):

pulsar-common/src/main/java/org/apache/pulsar/common/util/netty/ChannelFutures.java
```diff
@@ -41,7 +41,9 @@ private ChannelFutures() {
      *         and completes exceptionally if the channelFuture completes with a {@link Throwable}
      */
     public static CompletableFuture<Channel> toCompletableFuture(ChannelFuture channelFuture) {
-        Objects.requireNonNull(channelFuture, "channelFuture cannot be null");
+        if (channelFuture == null) {
```
codelipenghui, line 44: This will break the existing test `ChannelFuturesTest.toCompletableFuture_shouldRequireNonNullArgument` which uses `@Test(expectedExceptions = NullPointerException.class)` and expects a synchronous throw. After this change the method returns a failed future instead. The test needs to be updated to verify the future completes exceptionally with `NullPointerException`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java
```diff
@@ -1145,8 +1145,12 @@ public NamespaceBundleSplitAlgorithm getNamespaceBundleSplitAlgorithmByName(Stri
      */
     public CompletableFuture<Void> updateNamespaceBundlesForPolicies(NamespaceName nsname,
                                                                       NamespaceBundles nsBundles) {
-        Objects.requireNonNull(nsname);
-        Objects.requireNonNull(nsBundles);
+        if (nsname == null) {
+            return FutureUtil.failedFuture(new NullPointerException("Excepted NamespaceName should not be null "));
```
codelipenghui: Typo: `"Excepted"` should be `"Expected"`. Also has a trailing space before the closing quote. The sibling method `updateNamespaceBundles` below correctly uses `"Expected NamespaceName should not be null"`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
```diff
@@ -158,7 +157,9 @@ CompletableFuture<Long> asyncSaveBucketSnapshot(
     }
 
     private CompletableFuture<Void> putBucketKeyId(String bucketKey, Long bucketId) {
-        Objects.requireNonNull(bucketId);
+        if (bucketId == null) {
+            return FutureUtil.failedFuture(new NullPointerException());
```
codelipenghui: nit: `new NullPointerException()` has no message. Consider `new NullPointerException("bucketId")` to aid debugging.

codelipenghui (commented):
Thanks for the PR. The core intent is sound — methods returning `CompletableFuture` should not throw synchronous exceptions. A few issues found below.

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -287,8 +288,9 @@ public static <T> CompletableFuture<T> addTimeoutHandling(CompletableFuture<T> f
      */
     public static <T> @NonNull CompletableFuture<T> composeAsync(Supplier<CompletableFuture<T>> futureSupplier,
                                                                  Executor executor) {
-        Objects.requireNonNull(futureSupplier);
-        Objects.requireNonNull(executor);
+        if (futureSupplier == null || executor == null) {
```
codelipenghui: nit: Merging two separate null checks into one loses diagnostic specificity — you can't tell which parameter was null from the empty NPE. Consider keeping separate checks with messages.

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -231,7 +230,9 @@ public static <T> Sequencer<T> create() {
          * @throws NullPointerException NPE when param is null
```
codelipenghui: This `@throws NullPointerException` Javadoc is now inaccurate — the method returns a failed future instead of throwing. Same issue for `composeAsync` below.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LoadManager.java
```diff
@@ -70,11 +70,11 @@ default boolean started() {
 
     default CompletableFuture<Optional<LookupResult>> findBrokerServiceUrl(
             Optional<ServiceUnitId> topic, ServiceUnitId bundle, LookupOptions options) {
-        throw new UnsupportedOperationException();
+        return CompletableFuture.failedFuture(new UnsupportedOperationException());
```
codelipenghui: nit: This uses `CompletableFuture.failedFuture()` while most other changes in this PR use `FutureUtil.failedFuture()`. The usage is inconsistent across the PR — consider standardizing on one approach.

pulsar-broker/src/main/java/org/apache/pulsar/broker/web/PulsarWebResource.java
zhanghaou: Thanks for the review. The code has already been rolled back, and the changes have been kept to a minimum.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LoadManager.java
zhanghaou: All checked and fixed.

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
zhanghaou: Done.

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
zhanghaou: Done

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/Bucket.java
zhanghaou: Fixed

pulsar-common/src/main/java/org/apache/pulsar/common/util/netty/ChannelFutures.java
zhanghaou: Fixed

lhotari (changes requested):
> Some async validation paths in the broker currently throw exceptions directly instead of returning a failed CompletableFuture. This breaks the async contract of methods that return CompletableFuture, because the exception is thrown synchronously at invocation time rather than being propagated through the future.
> 
> As a result, callers relying on async error handling (e.g. exceptionally, handle, or whenComplete) may not observe these failures correctly.

I don't think that this is so simple.

In some cases, the calling code might be currently relying on exceptions being thrown before the actual asynchronous job is created. That's why converting all cases without evaluating each case is not a good approach.
Since all exception paths aren't tested, this change might cause more instability than bring benefits. 

I agree that the contract of the asynchronous method should be clear about exceptions, whether the exception gets thrown directly or returned as a "result" in the CompletableFuture. 

Please run Personal CI to validate such large changes on your side. You can run Pulsar CI in your personal fork by following these instructions: https://pulsar.apache.org/contribute/personal-ci/
The benefit is that you have full control of CI and we don't waste our quota for CI runs which could fail.

zhanghaou:
> In some cases, the calling code might be currently relying on exceptions being thrown before the actual asynchronous job is created. 

Thanks for raising this, I reviewed the call sites of the methods changed in this PR to check your concern specifically， in the current in-repo callers, I didn’t find that pattern. Most call sites are already handling failures via thenCompose / exceptionally or by awaiting the returned future (get/join), so the error is still observed through the async path. 

Could you give me some cases?

zhanghaou:
> Please run Personal CI to validate such large changes on your side. You can run Pulsar CI in your personal fork by following these instructions: https://pulsar.apache.org/contribute/personal-ci/ The benefit is that you have full control of CI and we don't waste our quota for CI runs which could fail.

Personal CI have all passed, except for a few uploaded tasks.

https://github.com/zhanghaou/pulsar/pull/3

lhotari:
> > In some cases, the calling code might be currently relying on exceptions being thrown before the actual asynchronous job is created.
> 
> Thanks for raising this, I reviewed the call sites of the methods changed in this PR to check your concern specifically， in the current in-repo callers, I didn’t find that pattern. Most call sites are already handling failures via thenCompose / exceptionally or by awaiting the returned future (get/join), so the error is still observed through the async path.
> 
> Could you give me some cases?

Thanks for checking that. IIRC, I've seen the pattern where the error handling relies on a pattern where the async job wouldn't be created at all at least for some method calls that return a CompletableFuture. However, it's possible that those methods have been eliminated over the years and the error handling has been modified to take this change in the method contract into account.

lhotari (changes requested):
I think it's necessary to remove LedgerOffloader and LoadManager interface changes from this PR since the changes would break third party implementations.
It's better to handle the interface changes for these "plugin interfaces" separately since according to our guidelines, we'd create a PIP for such changes. The other changes LGTM after checking them.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloader.java
lhotari, line 1: This interface is part of Pulsar's public API. Changes to this interface will break third party LedgerOffloaderFactor implementations.

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LoadManager.java
lhotari, line 1: This interface is part of Pulsar's public interface, the change would break third party LoadManager implementations (configurable with `loadManagerClassName` in broker.conf).

zhanghaou:
> I think it's necessary to remove LedgerOffloader and LoadManager interface changes from this PR since the changes would break third party implementations. It's better to handle the interface changes for these "plugin interfaces" separately since according to our guidelines, we'd create a PIP for such changes. The other changes LGTM after checking them.

Make sense. 

Additionally, I can’t guarantee that all problematic methods have been updated in this PR. I’ve done my best to address as many as possible.

Would it be possible to add the main purpose of this PR to the AI review rules, and prevent new errors from being introduced?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloader.java
zhanghaou: Done

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/LoadManager.java
zhanghaou: Done

lhotari (approved):

lhotari:
> Additionally, I can’t guarantee that all problematic methods have been updated in this PR. I’ve done my best to address as many as possible.

No worries. Unless CI doesn't catch the issues, we cannot do much more about it.

> Would it be possible to add the main purpose of this PR to the AI review rules, and prevent new errors from being introduced?

Do you mean the tuning of Copilot review guidance? I haven't checked how that is handled. 
For Pulsar "plugin interface" changes, there should be an automated solution in place in the CI pipeline that prevents accidential changes that break binary compatibility of interfaces. There's https://github.com/siom79/japicmp for checking all public API changes, but I don't know if it's possible to limit it to check only specific interfaces, for example the interfaces annotated with `org.apache.pulsar.common.classification.InterfaceStability.Stable`. It would be useful to add an AGENTS.md by explaining this since I'd assume that AI tools would at least be able to check changes at some level.

zhanghaou:
> Do you mean the tuning of Copilot review guidance? I haven't checked how that is handled. 

We can add a copilot-instructions.md file in the .github directory of the repository.


https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions

lhotari:
> > Do you mean the tuning of Copilot review guidance? I haven't checked how that is handled.
> 
> We can add a copilot-instructions.md file in the .github directory of the repository.
> 
> https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions

@zhanghaou Would you like to create a separate PR to add that?

zhanghaou:
> @zhanghaou Would you like to create a separate PR to add that?

Sure, I'd be happy to do.

dao-jun (approved):

PR #23062, opened 2024-07-22
[improve][broker] Reduce the CPU pressure from the transaction buffer in rolling restarts

BewareMyPower:
### Motivation

During the rolling restarts, the namespace bundle ownerships will change. Assuming there is a producer created on a single topic, and the ownership was transferred to the new broker. Assuming the namespace bundle has N topics and the namespace is `tenant/ns`,
1. All N topics in the same bundle of that topic will be loaded.
2. For each topic, the managed ledger will be initialized, when the transaction coordinator is enabled, a `TopicTransactionBuffer` will be created. A Pulsar reader will be created on `tenant/ns/__transaction_buffer_snapshot` concurrently.
3. Once all N readers are created, the owner of the snapshot topic will start dispatching messages to N readers. Each dispatcher will read messages from BookKeeper concurrently and might fail with too many requests error because BK can only have `maxPendingReadRequestsPerThread` pending read requests (default: 10000).

We have a `numTransactionReplayThreadPoolSize` config to limit the concurrency of transaction snapshot readers. However, it only limits the read loop. For example, if it's configured with 1, only 1 reader could read messages at the same time. However, N readers will be created concurrently. Each when one of these reader explicitly calls `readNext`, all N dispatchers at brokers side will dispatch messages to N readers.

The behaviors above brings much CPU pressure on the owner broker, especially for a small cluster with only two brokers.

### Modifications

Synchronize the reader creation, read loop and the following process on its result. Maintain only one reader for each namespace. The reader is now not closed unless there is no snapshot read request in 1 minute.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository: https://github.com/BewareMyPower/pulsar/pull/32

PR #25293, opened 2026-03-05
[improve][broker]Reduce the lock range of SimpleCache to enhance performance

poorbarcode:
### Motivation

https://github.com/apache/pulsar/pull/23062 improved the behaviour of initialising the Transaction Buffer: Synchronise the reader creation, read loop and the following process on its result. Maintain only one reader for each namespace. The reader is now not closed unless there is no snapshot read request in 1 minute.

However, [SimpleCache](https://github.com/apache/pulsar/pull/23062/changes#diff-874612f5aa0d34f1df3c8d48bd8d4a4b28dc826a12fd08c45e1a62e10f6b5182) is a global lock, and the lock force is too strong, causing all `pulsar-transaction-snapshot-recover` threads to get stuck in the creation of the first reader. When system resources are insufficient, the problem will be magnified infinitely

Screenshot 2026-03-05 at 22 31 03

---

Q1: The creation of the reader that reads `__transaction_buffer_snapshot` and the reading of existing messages are both executed synchronously.  Is it necessary to change it to be completed asynchronously?

A1: It is not necessary, since the initialisation of all TransactionBuffers under the same namespace requires waiting for the reader to complete the processing of all messages, whether it is asynchronous or not is not important

Q2. Will these synchronisation operations cause the thread `pulsar-transaction-snapshot-recover` to get stuck and affect other functions

A2: No. The function of this thread is quite simple: 1. Initialise Transaction Buffer. 2. After the Transaction Buffer recovery is completed, handle the accumulated transaction write operations (subsequent writes will no longer use this thread).
So the stuck functions actually all need to wait for the reader's messages to be processed, and thus, no other impacts have been caused

### Modifications

No longer wait for the initialisation of the reader to complete within the lock code block

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
@@ -21,8 +21,8 @@ package org.apache.pulsar.broker.service;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
-import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.broker.PulsarServerException;
 import org.apache.pulsar.broker.PulsarService;
@@ -41,13 +41,14 @@ public class SystemTopicTxnBufferSnapshotService<T> {
 
     protected final ConcurrentHashMap<NamespaceName, SystemTopicClient<T>> clients;
     protected final NamespaceEventsSystemTopicFactory namespaceEventsSystemTopicFactory;
+    protected final PulsarClientImpl pulsarClient;
 
     protected final Class<T> schemaType;
     protected final EventType systemTopicType;
 
     private final ConcurrentHashMap<NamespaceName, ReferenceCountedWriter<T>> refCountedWriterMap;
-    @Getter
-    private final TableView<T> tableView;
+
+    private final ConcurrentHashMap<Thread, TableView<T>> tableView = new ConcurrentHashMap<>();
 
     // The class ReferenceCountedWriter will maintain the reference count,
     // when the reference count decrement to 0, it will be removed from writerFutureMap, the writer will be closed.
@@ -103,14 +104,12 @@ public class SystemTopicTxnBufferSnapshotService<T> {
 
     public SystemTopicTxnBufferSnapshotService(PulsarService pulsar, EventType systemTopicType,
                                                Class<T> schemaType) throws PulsarServerException {
-        final var client = (PulsarClientImpl) pulsar.getClient();
-        this.namespaceEventsSystemTopicFactory = new NamespaceEventsSystemTopicFactory(client);
+        this.pulsarClient = (PulsarClientImpl) pulsar.getClient();
+        this.namespaceEventsSystemTopicFactory = new NamespaceEventsSystemTopicFactory(pulsarClient);
         this.systemTopicType = systemTopicType;
         this.schemaType = schemaType;
         this.clients = new ConcurrentHashMap<>();
         this.refCountedWriterMap = new ConcurrentHashMap<>();
-        this.tableView = new TableView<>(this::createReader,
-                client.getConfiguration().getOperationTimeoutMs(), pulsar.getExecutor());
     }
 
     public CompletableFuture<SystemTopicClient.Reader<T>> createReader(TopicName topicName) {
@@ -171,6 +170,12 @@ public class SystemTopicTxnBufferSnapshotService<T> {
             }
         }
         refCountedWriterMap.clear();
+    }
+
+    public TableView<T> getTableView(ScheduledExecutorService scheduledExecutor) {
+        return tableView.computeIfAbsent(Thread.currentThread(),
+            k -> new TableView<>(this::createReader, pulsarClient.getConfiguration().getOperationTimeoutMs(),
+                scheduledExecutor));
     }
 
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
@@ -48,7 +48,7 @@ public class SystemTopicTxnBufferSnapshotService<T> {
 
     private final ConcurrentHashMap<NamespaceName, ReferenceCountedWriter<T>> refCountedWriterMap;
 
-    private final ConcurrentHashMap<Thread, TableView<T>> tableView = new ConcurrentHashMap<>();
+    private final ThreadLocal<TableView<T>> tableViewThreadLocal = new ThreadLocal<>();
 
     // The class ReferenceCountedWriter will maintain the reference count,
     // when the reference count decrement to 0, it will be removed from writerFutureMap, the writer will be closed.
@@ -173,9 +173,13 @@ public class SystemTopicTxnBufferSnapshotService<T> {
     }
 
     public TableView<T> getTableView(ScheduledExecutorService scheduledExecutor) {
-        return tableView.computeIfAbsent(Thread.currentThread(),
-            k -> new TableView<>(this::createReader, pulsarClient.getConfiguration().getOperationTimeoutMs(),
-                scheduledExecutor));
+        TableView<T> tableView = tableViewThreadLocal.get();
+        if (tableView == null) {
+            tableView = new TableView<>(this::createReader,
+                    pulsarClient.getConfiguration().getOperationTimeoutMs(), scheduledExecutor);
+            tableViewThreadLocal.set(tableView);
+        }
+        return tableView;
     }
 
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
@@ -48,6 +48,11 @@ public class SystemTopicTxnBufferSnapshotService<T> {
 
     private final ConcurrentHashMap<NamespaceName, ReferenceCountedWriter<T>> refCountedWriterMap;
 
+    /** SystemTopicTxnBufferSnapshotService is created only three, see also
+     *  {@link TransactionBufferSnapshotServiceFactory}. At the same time, each object can only
+     * be fixed threads access, see also {@link PulsarService#transactionSnapshotRecoverExecutorProvider}.
+     * So the un-static ThreadLocal is safe.
+     */
     private final ThreadLocal<TableView<T>> tableViewThreadLocal = new ThreadLocal<>();
 
     // The class ReferenceCountedWriter will maintain the reference count,
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
@@ -308,7 +308,7 @@ public class PulsarService implements AutoCloseable, ShutdownService {
 
     private TransactionPendingAckStoreProvider transactionPendingAckStoreProvider;
     private final ExecutorProvider transactionExecutorProvider;
-    private final ExecutorProvider transactionSnapshotRecoverExecutorProvider;
+    private final OrderedScheduler transactionSnapshotRecoverExecutorProvider;
     private final MonotonicClock monotonicClock;
     private String brokerId;
     private final CompletableFuture<Void> readyForIncomingRequestsFuture = new CompletableFuture<>();
@@ -380,8 +380,10 @@ public class PulsarService implements AutoCloseable, ShutdownService {
         if (config.isTransactionCoordinatorEnabled()) {
             this.transactionExecutorProvider = new ExecutorProvider(this.getConfiguration()
                     .getNumTransactionReplayThreadPoolSize(), "pulsar-transaction-executor");
-            this.transactionSnapshotRecoverExecutorProvider = new ExecutorProvider(this.getConfiguration()
-                    .getNumTransactionReplayThreadPoolSize(), "pulsar-transaction-snapshot-recover");
+            this.transactionSnapshotRecoverExecutorProvider = OrderedScheduler.newSchedulerBuilder()
+                    .numThreads(this.getConfiguration().getNumTransactionReplayThreadPoolSize())
+                    .name("pulsar-transaction-snapshot-recover")
+                    .build();
         } else {
             this.transactionExecutorProvider = null;
             this.transactionSnapshotRecoverExecutorProvider = null;
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SingleSnapshotAbortedTxnProcessorImpl.java
@@ -21,6 +21,7 @@ package org.apache.pulsar.broker.transaction.buffer.impl;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ScheduledExecutorService;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.Position;
 import org.apache.bookkeeper.mledger.PositionFactory;
@@ -91,10 +92,13 @@ public class SingleSnapshotAbortedTxnProcessorImpl implements AbortedTxnProcesso
     public CompletableFuture<Position> recoverFromSnapshot() {
         final var future = new CompletableFuture<Position>();
         final var pulsar = topic.getBrokerService().getPulsar();
-        pulsar.getTransactionSnapshotRecoverExecutorProvider().getExecutor(this).execute(() -> {
+        final String namespace = TopicName.get(topic.getName()).getNamespace();
+        final ScheduledExecutorService scheduledExecutor = pulsar.getTransactionSnapshotRecoverExecutorProvider()
+                .chooseThread(namespace);
+        scheduledExecutor.execute(() -> {
             try {
                 final var snapshot = pulsar.getTransactionBufferSnapshotServiceFactory().getTxnBufferSnapshotService()
-                        .getTableView().readLatest(topic.getName());
+                        .getTableView(scheduledExecutor).readLatest(topic.getName());
                 if (snapshot != null) {
                     handleSnapshot(snapshot);
                     final var startReadCursorPosition = PositionFactory.create(snapshot.getMaxReadPositionLedgerId(),
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SnapshotSegmentAbortedTxnProcessorImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/SnapshotSegmentAbortedTxnProcessorImpl.java
@@ -28,6 +28,7 @@ import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
@@ -231,10 +232,14 @@ public class SnapshotSegmentAbortedTxnProcessorImpl implements AbortedTxnProcess
     public CompletableFuture<Position> recoverFromSnapshot() {
         final var pulsar = topic.getBrokerService().getPulsar();
         final var future = new CompletableFuture<Position>();
-        pulsar.getTransactionSnapshotRecoverExecutorProvider().getExecutor(this).execute(() -> {
+        final String namespace = TopicName.get(topic.getName()).getNamespace();
+        final ScheduledExecutorService scheduledExecutor = pulsar.getTransactionSnapshotRecoverExecutorProvider()
+                .chooseThread(namespace);
+        scheduledExecutor.execute(() -> {
             try {
                 final var indexes = pulsar.getTransactionBufferSnapshotServiceFactory()
-                        .getTxnBufferSnapshotIndexService().getTableView().readLatest(topic.getName());
+                        .getTxnBufferSnapshotIndexService().getTableView(scheduledExecutor)
+                        .readLatest(topic.getName());
                 if (indexes == null) {
                     // Try recovering from the old format snapshot
                     future.complete(recoverOldSnapshot());
@@ -342,6 +347,10 @@ public class SnapshotSegmentAbortedTxnProcessorImpl implements AbortedTxnProcess
 
     // This method will be deprecated and removed in version 4.x.0
     private Position recoverOldSnapshot() throws Exception {
+        final String namespace = TopicName.get(topic.getName()).getNamespace();
+        final ScheduledExecutorService scheduledExecutor = topic.getBrokerService().getPulsar()
+                .getTransactionSnapshotRecoverExecutorProvider()
+                .chooseThread(namespace);
         final var pulsar = topic.getBrokerService().getPulsar();
         final var topicName = TopicName.get(topic.getName());
         final var topics = wait(pulsar.getPulsarResources().getTopicResources().listPersistentTopicsAsync(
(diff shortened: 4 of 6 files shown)


codelipenghui (commented):
@poorbarcode Is it better just fix SimpleCache so get() does not run valueSupplier under a global synchronized lock? It will avoids per-thread TableView growth/leak risk.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -89,16 +103,13 @@ private T internalReadLatest(String topic) throws Exception {
...
-                throw new RuntimeException(e);
-            }
-        }, __ -> __.closeAsync().exceptionally(e -> {
-            log.warn("Failed to close reader {}", e.getMessage());
-            return null;
-        }));
+        NamespaceName ns = topicName.getNamespaceObject();
+        CompletableFuture<Reader<T>> readerFuture = readers.get(ns, () -> readerCreator.apply(topicName), expiration);
+        try {
+            return wait(readerFuture, "create reader");
```
poorbarcode: @codelipenghui 

> @poorbarcode Is it better just fix SimpleCache so get() does not run valueSupplier under a global synchronized lock? It will avoids per-thread TableView growth/leak risk.

Good suggestion, I moved out the logic for waiting for the reader to complete its creation

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -48,7 +49,20 @@ public class TableView<T> {
...
+        if (!readerFuture.isDone()) {
+            readerFuture.handle((reader, t) -> {
+                if (reader != null) {
+                    closeReader(reader);
+                }
+                return null;
+            });
+        } else if (!readerFuture.isCompletedExceptionally()) {
+            // Since the future has done, it will not wait anymore.
+            closeReader(readerFuture.join());
```
Denovo1998: This shortened the lock duration but appears to introduce a new implicit constraint: clientOperationTimeoutMs must be less than CACHE_EXPIRE_TIMEOUT_MS. Otherwise, while getReader() awaits the future, the cache entry could expire and register a "close immediately upon completion" action, causing the reader retrieved by the caller to already be in the process of closing.

Denovo1998 (changes requested):
Can we add targeted tests for the new future lifecycle here?

The risks in this change lie not in the expected path, but in the ownership and expiration transitions:

- the cache entry is removed or expired before the future completes
- the future completes successfully after expiration and should close the reader exactly once
- the future completes exceptionally and should not try to close anything

Denovo1998 (changes requested):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
poorbarcode: The default value of `clientOperationTimeoutMs` is 30s(we have not change it), and `CACHE_EXPIRE_TIMEOUT_MS` is 60s

The code was modified to ensure that the original behavior would not change and that there would never be `clientOperationTimeoutMs` situation where `CACHE_EXPIRE_TIMEOUT_MS` is greater than B


> the cache entry is removed or expired before the future completes
> the future completes successfully after expiration and should close the reader exactly once

These two cases will never happen

poorbarcode: > the future completes exceptionally and should not try to close anything

It seems that this is very difficult to achieve because you cannot determine whether a non-existent object has ever been closed.

Added a simple test to cover the case that failed to create reader

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
lhotari, line 32: Just wondering if this we could use `ConcurrentHashMap` here to reduce locking further. In that case, it should be possible to eliminate `synchronized` for the `getWithCacheInfo` and `get` methods.

Since there's need for locking for expiration handling, it should be done in the `ExpirableValue` class. The `ExpirableValue` could hold expired state which could be reset if the `ExpirableValue` isn't removed from the cache before it gets reactivated.

lhotari (commented):
Please check the comment about using ConcurrentHasMap and moving locking to ExpirableValue class to avoid race conditions with expiration.

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TableViewTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Test
+    public void testFailedCreateReader() throws Exception {
+        ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+        Function<TopicName, CompletableFuture<SystemTopicClient.Reader<Object>>> readerCreator = topicName -> {
+            return FutureUtil.failedFuture(new PulsarClientException("test failed to create reader"));
+        };
+        TableView<Object> tableView = new TableView<>(readerCreator, 60000, executor);
+        try {
+            tableView.readLatest("public/default/tp");
```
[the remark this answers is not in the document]
poorbarcode: Improved

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TableViewTest.java
```diff
@@ -0,0 +1,47 @@
...
+        } catch (Exception e) {
+            Assert.assertTrue(e.getMessage().contains("Failed to create reader"));
```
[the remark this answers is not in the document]
poorbarcode: Improved

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
```diff
@@ -48,7 +48,7 @@ boolean tryExpire() {
             }
         }
 
-        void updateDeadline() {
+        public void updateDeadline() {
             deadlineMs = System.currentTimeMillis() + timeoutMs;
         }
```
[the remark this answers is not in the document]
poorbarcode: marked `deadlineMs` as volatile

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
```diff
@@ -33,9 +33,9 @@ public class SimpleCache<K, V> {
     private final long timeoutMs;
 
     @RequiredArgsConstructor
-    private class ExpirableValue<V> {
+    public class ExpirableValue<V> {
```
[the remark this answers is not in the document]
poorbarcode: Improved

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -89,16 +105,16 @@ private T internalReadLatest(String topic) throws Exception {
...
+        NamespaceName ns = topicName.getNamespaceObject();
+        SimpleCache<NamespaceName, CompletableFuture<Reader<T>>>.ExpirableValue<CompletableFuture<Reader<T>>>
+                cachedReaderFuture = readers.getWithCacheInfo(ns, () -> readerCreator.apply(topicName), expiration);
+        try {
+            return wait(cachedReaderFuture.value, "create reader");
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        } finally {
+            cachedReaderFuture.updateDeadline();
+        }
```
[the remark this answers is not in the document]
poorbarcode: Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -48,13 +49,28 @@ public class TableView<T> {
...
+            readerFuture.handle((reader, t) -> {
+                if (reader != null) {
+                    closeReader(reader);
+                }
+                return null;
+            });
+        } else if (!readerFuture.isCompletedExceptionally()) {
+            // Since the future has done, it will not wait anymore.
+            closeReader(readerFuture.join());
+        }
```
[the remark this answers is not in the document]
poorbarcode: Improved

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
poorbarcode: ConcurrentHashMap also uses `synchronized` to guarantee thread safety, right?

lhotari: > ConcurrentHashMap also uses `synchronized` to guarantee thread safety, right?

No, it's very different. Some explanation in the [Javadoc of ConcurrentHashMap](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html). `get` operations rarely block and modification operations won't be locked on a global lock.

lhotari: `computeIfAbsent` would be used to ensure that duplicates aren't added to the map and locking could be handled in `ExpirableValue` level to avoid shared locks and just have a per-key lock.

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/TableViewTest.java
```diff
@@ -97,6 +102,17 @@ protected void cleanup() throws Exception {
         super.internalCleanup();
     }
 
+    @Test
+    public void testFailedCreateReader() throws Exception {
```
shibd: This test needs to assert a throw exception. 

In fact, this test duplication with [broker/transaction/buffer/impl/TableViewTest.java‎](https://github.com/apache/pulsar/pull/25293/changes#diff-edfc112d734b7be5af2074033fa84f7f49e22ca47223cf064ae2bc19208eabb4)

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TableViewTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.pulsar.client.api.PulsarClientException;
+import org.apache.pulsar.common.naming.TopicName;
+import org.apache.pulsar.common.util.FutureUtil;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class TableViewTest {
+
+    @Test
+    public void testFailedCreateReader() throws Exception {
```
shibd: This test covers the failure path, but it does not exercise the contention scenario that this PR is actually fixing. Could we add a targeted concurrency test where one namespace caches a blocked `CompletableFuture<Reader<T>>`, and then a second `readLatest()` on another namespace is started before the first future completes?

shibd: A more targeted test could focus on the actual contention this PR fixes:

- `ns1` gets a blocked `CompletableFuture<Reader<T>>`
- thread A calls `readLatest("public/ns1/t1")`
- before `future1` completes, thread B calls `readLatest("public/ns2/t2")`
- assert that `ns2` can still reach `readerCreator` and cache its own future

That is the observable difference from the old code: previously thread B would be stuck behind the global `SimpleCache.get()` lock while thread A was waiting inside the synchronized block.

Pseudo-code:
```java
CompletableFuture<Reader<Object>> future1 = new CompletableFuture<>();
CompletableFuture<Reader<Object>> future2 = new CompletableFuture<>();
CountDownLatch ns1Called = new CountDownLatch(1);
CountDownLatch ns2Called = new CountDownLatch(1);

readerCreator = topic -> {
    if (topic.getNamespaceObject().toString().equals("public/ns1")) {
        ns1Called.countDown();
        return future1;
    } else {
        ns2Called.countDown();
        return future2;
    }
};

runAsync(() -> tableView.readLatest("public/ns1/t1"));
assertTrue(ns1Called.await(1, SECONDS));
runAsync(() -> tableView.readLatest("public/ns2/t2"));
assertTrue(ns2Called.await(1, SECONDS)); // this is the key assertion

future1.complete(mockReaderReturningNoEvents());
future2.complete(mockReaderReturningNoEvents());
```

If `ns2Called` cannot fire until `future1` completes, the old lock amplification behavior is still present.

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
poorbarcode: > computeIfAbsent would be used to ensure that duplicates aren't added to the map and locking could be handled in ExpirableValue level to avoid shared locks and just have a per-key lock.

Since the previous implementation initialised the reader synchronously, the method `computeIfAbsent` will be stuck, and following requests will also be stuck until the first request is done, right?

pulsar-broker/src/test/java/org/apache/pulsar/client/impl/TableViewTest.java
poorbarcode: Removed duplicated test

pulsar-broker/src/test/java/org/apache/pulsar/broker/transaction/buffer/impl/TableViewTest.java
poorbarcode: Added the test

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
lhotari: > Since the previous implementation initialised the reader synchronously, the method `computeIfAbsent` will be stuck, and following requests will also be stuck until the first request is done, right?

There are multiple locks behind `ConcurrentHashMap`'s `computeIfAbsent`, so it's not getting "stuck" for all. I agree that having a solution like there is in this PR is useful and it reduces the need for further optimization as long as the operation is really fully asynchronous. It doesn't really matter in the current use case, but if SimpleCache is used for many other use cases, it would make sense to optimize further, to get rid of a global lock on SimpleCache and move the locking to the value holder class (ExpirableValue). The value wouldn't be initialized in `computeIfAbsent`, but when the key is retrieved the first key.

Denovo1998: @lhotari @poorbarcode 
Currently, it seems only `TableView` is used. The scoped fix in this PR seems reasonable.
Perhaps in the future when `SimpleCache` is reused more widely, moving to `ConcurrentHashMap` plus per-entry synchronization/state in `ExpirableValue` would make sense as a separate follow-up?

lhotari (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -48,13 +49,22 @@ public class TableView<T> {
...
+    };
 
     public TableView(Function<TopicName, CompletableFuture<Reader<T>>> readerCreator, long clientOperationTimeoutMs,
                      ScheduledExecutorService executor) {
         this.readerCreator = readerCreator;
         this.clientOperationTimeoutMs = clientOperationTimeoutMs;
-        this.readers = new SimpleCache<>(executor, CACHE_EXPIRE_TIMEOUT_MS, CACHE_EXPIRE_CHECK_FREQUENCY_MS);
+        this.readers = new SimpleCache<>(executor,
+                Math.max(clientOperationTimeoutMs + 30 * 1000, CACHE_EXPIRE_TIMEOUT_MS),
+                CACHE_EXPIRE_CHECK_FREQUENCY_MS);
```
codelipenghui: Is it risky to close the reader immediately after the cache expires? If the recovery task takes longer than the expiration period, it will consistently fail with a `ReaderAlreadyClosedException`.

codelipenghui (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -89,16 +99,17 @@ private T internalReadLatest(String topic) throws Exception {
...
-            log.warn("Failed to close reader {}", e.getMessage());
-            return null;
-        }));
+        NamespaceName ns = topicName.getNamespaceObject();
+        SimpleCache<NamespaceName, CompletableFuture<Reader<T>>>.ExpirableValue
+                cachedReaderFuture = readers.getWithCacheInfo(ns, () -> readerCreator.apply(topicName), expiration);
+        try {
+            return wait(cachedReaderFuture.value, "create reader");
+        } catch (Exception e) {
+            readers.remove(ns);
```
codelipenghui: It's better to compare the value here to avoid removing a new, valid entry for the same namespace

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
codelipenghui, line 68: It's better to catch exception for this block, to avoid any exception happened from cancel() method caused expiration processing stops silently

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
```diff
@@ -108,4 +119,11 @@ private <R> R wait(CompletableFuture<R> future, String msg) throws Exception {
             throw new CompletionException("Failed to " + msg, e.getCause());
         }
     }
+
+    private void closeReader(Reader<T> reader) {
+        reader.closeAsync().exceptionally(ex -> {
+            log.warn("Failed to close reader {}", ex.getMessage());
```
codelipenghui:
```suggestion
            log.warn("Failed to close reader ", ex);
```

Denovo1998 (approved):

shibd (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
poorbarcode: I have modified it to the new plan: the reader of a namespace will always only be executed in one thread, and locks are no longer used at all

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
poorbarcode: I have modified it to the original plan: the reader of a namespace will always only be executed in one thread, and locks are no longer used at all

pulsar-broker/src/main/java/org/apache/pulsar/utils/SimpleCache.java
poorbarcode: I have modified it to the original plan: the reader of a namespace will always only be executed in one thread, and locks are no longer used at all

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/TableView.java
poorbarcode: I have modified it to the original plan: the reader of a namespace will always only be executed in one thread, and locks are no longer used at all

poorbarcode:
@lhotari @shibd 

I have modified it to the original plan: the reader of a namespace will always only be executed in one thread, and locks are no longer used at all.

You can review again at your convenience

codelipenghui (approved):

shibd (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
```diff
@@ -173,4 +177,14 @@ public void close() throws Exception {
...
 
+    public TableView<T> getTableView(ScheduledExecutorService scheduledExecutor) {
+        TableView<T> tableView = tableViewThreadLocal.get();
+        if (tableView == null) {
+            tableView = new TableView<>(this::createReader,
+                    pulsarClient.getConfiguration().getOperationTimeoutMs(), scheduledExecutor);
+            tableViewThreadLocal.set(tableView);
+        }
+        return tableView;
+    }
```
lhotari, line 188: Using a ThreadLocal for this purpose seems brittle. Is there really a need to have independent TableView instances for different threads? If there isn't, this doesn't seem to be the correct solution.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
```diff
@@ -41,13 +41,19 @@ public class SystemTopicTxnBufferSnapshotService<T> {
...
     private final ConcurrentHashMap<NamespaceName, ReferenceCountedWriter<T>> refCountedWriterMap;
-    @Getter
-    private final TableView<T> tableView;
+
+    /** SystemTopicTxnBufferSnapshotService is created only three, see also
+     *  {@link TransactionBufferSnapshotServiceFactory}. At the same time, each object can only
+     * be fixed threads access, see also {@link PulsarService#transactionSnapshotRecoverExecutorProvider}.
+     * So the un-static ThreadLocal is safe.
+     */
+    private final ThreadLocal<TableView<T>> tableViewThreadLocal = new ThreadLocal<>();
```
lhotari, line 56: How is the TableView instance going to be closed/shutdown if this is the solution?

poorbarcode: > How is the TableView instance going to be closed/shutdown if this is the solution?

It does not need to be closed; it binds to SystemTopicTxnBufferSnapshotService, and `SystemTopicTxnBufferSnapshotService` will only be closed when the Pulsar broker shuts down.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
poorbarcode: > Using a ThreadLocal for this purpose seems brittle. Is there really a need to have independent TableView 

TableView only used in `pulsar-transaction-snapshot-recover` threads; it is safety enough

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
lhotari: It should be closed because we have tests running in the same JVM. Resource leaks such as this one should be prevented.

lhotari (dismissed):
Please fix the resource leak that exists in `tableViewThreadLocal`. Fixing it is important for our CI. Test runs become slow, flaky, or fail (contribute to OOME) when resources are leaked.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
poorbarcode: > It should be closed because we have tests running in the same JVM. Resource leaks such as this one should be prevented.

No, it will not cause leaks, for example:

- Background
  - TbaleView ThreadLocal was bound to `SystemTopicTxnBufferSnapshotService`
  - All tests that use  `SystemTopicTxnBufferSnapshotService` will use the same thread pool `pulsar -> transactionSnapshotRecoverExecutorProvider`
 

- Env
  - CPU: 3 
  - `pulsar-transaction-snapshot-recover threads`: 3 // which is the same as CPU cores.

- Test 1
  -  created 3 TableView for each pulsar-transaction-snapshot-recover thread
  - finished test without closing 3 TableView

- Test 2
  - It will reuse the 3 TableViews that the first test created

poorbarcode:
> Please fix the resource leak that exists in tableViewThreadLocal. Fixing it is important for our CI. Test runs become slow, flaky, or fail (contribute to OOME) when resources are leaked.

I have already responded to your concern here. Please take a look first https://github.com/apache/pulsar/pull/25293#discussion_r2975853856

lhotari:
> > Please fix the resource leak that exists in tableViewThreadLocal. Fixing it is important for our CI. Test runs become slow, flaky, or fail (contribute to OOME) when resources are leaked.
> 
> I have already responded to your concern here. Please take a look first [#25293 (comment)](https://github.com/apache/pulsar/pull/25293#discussion_r2975853856)

@poorbarcode I'll fix that issue separately.

poorbarcode:
@lhotari Could you point to the exact test method?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/SystemTopicTxnBufferSnapshotService.java
lhotari: Oh yes, you are right, it's not a static field.

lhotari:
> @lhotari Could you point to the exact test method?

I took a closer look and in this case, it happens to be fine since TableView and SimpleCache don't keep state by starting their own threads. The allocated resources such as the Reader will get closed when the client closes. This is fine.

PR #25296, opened 2026-03-07
[fix][offload] Close all resources in BlobStoreBackedReadHandleImplV2.closeAsync

iamsanjaymalakar:
### Motivation
`BlobStoreBackedReadHandleImplV2#closeAsync()` closes indices and dataStreams sequentially inside a single try/catch. If closing one resource throws, the method exits early and the remaining resources are never closed. This can leak file descriptors and leave resources open when offload read handles are closed.

### Modifications
- Update `closeAsync()` to attempt closing all `OffloadIndexBlockV2` instances and all `DataInputStreams`.
- Collect the first `IOException` and attach subsequent failures via suppressed exceptions.
- Complete the returned `CompletableFuture` exceptionally if any close operation fails; otherwise mark the handle Closed and complete normally.

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
+++ b/tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
@@ -132,34 +132,39 @@ public class BlobStoreBackedReadHandleImplV2 implements ReadHandle {
 
         CompletableFuture<Void> promise = closeFuture.get();
         executor.execute(() -> {
-            IOException first = null;
-            for (OffloadIndexBlockV2 indexBlock : indices) {
-                try {
-                    indexBlock.close();
-                } catch (IOException e) {
-                    if (first == null) {
-                        first = e;
-                    } else {
-                        first.addSuppressed(e);
+            try {
+                IOException first = null;
+                for (OffloadIndexBlockV2 indexBlock : indices) {
+                    try {
+                        indexBlock.close();
+                    } catch (IOException e) {
+                        if (first == null) {
+                            first = e;
+                        } else {
+                            first.addSuppressed(e);
+                        }
                     }
                 }
-            }
-            for (DataInputStream dataStream : dataStreams) {
-                try {
-                    dataStream.close();
-                } catch (IOException e) {
-                    if (first == null) {
-                        first = e;
-                    } else {
-                        first.addSuppressed(e);
+                for (DataInputStream dataStream : dataStreams) {
+                    try {
+                        dataStream.close();
+                    } catch (IOException e) {
+                        if (first == null) {
+                            first = e;
+                        } else {
+                            first.addSuppressed(e);
+                        }
                     }
                 }
-            }
-            if (first != null) {
-                promise.completeExceptionally(e);
-            } else {
                 state = State.Closed;
-                promise.complete(null);
+                if (first != null) {
+                    promise.completeExceptionally(first);
+                } else {
+                    promise.complete(null);
+                }
+            } catch (Throwable t) {
+                state = State.Closed;
+                promise.completeExceptionally(t);
             }
         });
         return promise;


tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
```diff
@@ -132,17 +132,34 @@ public CompletableFuture<Void> closeAsync() {
...
+                    if (first == null) {
+                        first = e;
+                    } else {
+                        first.addSuppressed(e);
+                    }
                 }
+            }
+            if (first != null) {
+                promise.completeExceptionally(e);
+            } else {
```
codelipenghui: Bug: `e` is not in scope here — it only exists inside the `catch` blocks. Should be `first`. This will not compile.

tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
```diff
@@ -132,17 +132,34 @@ public CompletableFuture<Void> closeAsync() {
...
                 state = State.Closed;
                 promise.complete(null);
```
codelipenghui: `state` should be set to `Closed` unconditionally (move it before the `if/else`), not only on the success path. Once all resources have had `close()` attempted, the handle is effectively closed. Leaving it as `Opened` on failure means `readAsync()` won't reject calls, but the underlying resources are already closed, causing confusing secondary errors. Also, `closeFuture` is already set via `compareAndSet`, so retrying `closeAsync()` returns the same failed future — the handle becomes permanently stuck.

tiered-storage/jcloud/src/main/java/org/apache/bookkeeper/mledger/offload/jcloud/impl/BlobStoreBackedReadHandleImplV2.java
```diff
@@ -132,17 +132,34 @@ public CompletableFuture<Void> closeAsync() {
 
         CompletableFuture<Void> promise = closeFuture.get();
         executor.execute(() -> {
-            try {
-                for (OffloadIndexBlockV2 indexBlock : indices) {
+            IOException first = null;
+            for (OffloadIndexBlockV2 indexBlock : indices) {
```
codelipenghui: The individual `try/catch` blocks only catch `IOException`. If any `close()` throws a `RuntimeException`, it propagates out of the lambda, gets swallowed by the `ExecutorService`, and `promise` is never completed — callers hang forever on `closeAsync().get()`. Consider wrapping the entire lambda body in `try { ... } catch (Throwable t) { state = State.Closed; promise.completeExceptionally(t); }`.

iamsanjaymalakar:
Thanks for the reviews! I addressed the feedback by making `closeAsync()` close all resources without early exit and ensuring the returned future is always completed (even on runtime exceptions). I didn’t add a test because resource-leak/file-lock tests tend to be nondeterministic and flaky across OS/CI environments.

lhotari:
Please run "Personal CI" to validate such changes on your side. You can run Pulsar CI in your personal fork by following these instructions: https://pulsar.apache.org/contribute/personal-ci/. It's free.
The benefit is that you have full control of CI and you won't need to wait for maintainers to approve the CI workflow run to get CI feedback.

lhotari:
One detail about BlobStoreBackedReadHandleImplV2 is that it's not used in Pulsar's production code.
The 
`org.apache.bookkeeper.mledger.LedgerOffloader#readOffloaded(long, org.apache.bookkeeper.mledger.proto.MLDataFormats.OffloadContext, java.util.Map<java.lang.String,java.lang.String>)` method is not called from production code which would use BlobStoreBackedReadHandleImplV2.
This something that I noticed while fixing a flaky tests a while ago in #24331.
["PIP 76: Streaming Offload"](https://github.com/apache/pulsar/wiki/PIP-76:-Streaming-Offload) hasn't been fully implemented.

Issue #25235, opened 2026-02-09
Add --state filter flag to pulsar-admin functions list


david-streamlio:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Motivation

IMHO, it's kind of annoying that pulsar-admin functions list doesn't distinguish between running functions and stopped ones. If you have a namespace with a mix of active and stopped functions, there's no way to quickly see which ones are actually running without checking each one individually.
This bash one-liner solves that, but it shouldn't be necessary:

```
for function in $(pulsar-admin functions list --namespace $NAMESPACE --tenant $TENANT); do
  echo $function
  pulsar-admin functions status --namespace $NAMESPACE --tenant $TENANT --name $function | fgrep running
done
```

This is slow and noisy — it makes N+1 API calls just to answer "what's currently running?" It would be much more useful if functions list supported filtering by runtime state directly.

### Solution

Add optional filter flags to pulsar-admin functions list:

--state flag
Filter functions by their running state.

```
# Show only running functions
pulsar-admin functions list --tenant public --namespace default --state RUNNING

# Show only stopped functions
pulsar-admin functions list --tenant public --namespace default --state STOPPED
```
Accepted values: RUNNING, STOPPED (and potentially FAILED, UNKNOWN to cover all states returned by functions status).

--long / -l flag (optional, nice-to-have)
Show extended output with status inline, similar to ls -l:

```
NAME                        STATE      INSTANCES
add-offset-function         RUNNING    1/1
topic-filter-key            STOPPED    0/1
my-dedup-function           RUNNING    3/3
```

### Benefit

Eliminates the N+1 API call pattern shown above
Makes it practical to monitor function health from scripts and CI/CD pipelines
Consistent with other pulsar-admin subcommands that support filtering (e.g., topics list)
Reduces operational friction for namespaces with many functions

### Compatibility
This is a purely additive change. Without any flags, functions list behaves exactly as it does today.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

onceMisery:
Could you assign it to me?

onceMisery:
@david-streamlio  Hello！I have some implementation details that I need your help to confirm.

### Server-side Filtering
Add a state query parameter to the Worker REST endpoint. The server performs filtering and returns the result. This involves a single API call and delivers optimal performance, but requires changes to the API interface, Worker implementation, and REST endpoint.

### Client-side Filtering
[...]

PR #25299, opened 2026-03-09
[pip] PIP-459: Batch Status Summary and Filtered Listing for Pulsar Functions

onceMisery:
- Main Issue: [#25235](https://github.com/apache/pulsar/issues/25235)
 
### Motivation
See [pip/pip-459.md](https://github.com/onceMisery/pulsar/blob/pip/issues/20235-pulsar-admin/pip/pip-459.md) for the full proposal.

### Modifications
Added PIP-459 proposal document.

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

[matching-pr](https://github.com/onceMisery/pulsar/pull/3)

### DIFF
--- a/pip/pip-459.md
+++ b/pip/pip-459.md
@@ -29,6 +29,7 @@ This PIP proposes a batch status summary API for Pulsar Functions and integrates
 - Add admin client support for the new summary API, including fallback to legacy workers that do not implement it.
 - Add CLI support for long-format listing and state-based filtering using the batch summary API.
 - Add pagination support for namespace-level function summaries.
+- Add an optional request parameter to include detailed per-function runtime metrics in the summary response.
 - Improve worker-side summary generation latency by using controlled parallelism.
 - Add a worker configuration knob to cap summary-query parallelism.
 - Add a worker metric to observe summary-query execution time.
@@ -48,7 +49,7 @@ This proposal adds a new REST endpoint:
 
 `GET /admin/v3/functions/{tenant}/{namespace}/status/summary`
 
-The endpoint returns a list of `FunctionStatusSummary` objects. Each object contains:
+The endpoint returns a list of `FunctionStatusSummary` objects. By default, each object contains the lightweight fields needed for namespace-level listing:
 
 - `name`
 - `state`: `RUNNING`, `STOPPED`, `PARTIAL`,  `UNKNOWN`
@@ -56,6 +57,8 @@ The endpoint returns a list of `FunctionStatusSummary` objects. Each object cont
 - `numRunning`
 - `error`
 - `errorType`
+
+The endpoint also accepts `detailed=true` to include additional per-function runtime metrics in each response element, such as message-processing counters, exception summaries, and last-invocation timing. When `detailed=false` (the default), those extra fields are omitted to keep the endpoint lightweight for large namespace listings.
 
 The server remains a generic summary provider. It does not apply state filtering. The CLI consumes the summary list and performs presentation concerns locally, such as `--state` filtering and `--long` formatting. This separation keeps the endpoint reusable for other clients and prevents coupling the REST contract to one CLI presentation format.
 
@@ -93,6 +96,16 @@ public class FunctionStatusSummary {
     private int numRunning;
     private String error;
     private ErrorType errorType;
+
+    // Present only when the request includes detailed=true.
+    private Long numReceived;
+    private Long numSuccessfullyProcessed;
+    private Long numUserExceptions;
+    private List<ExceptionInformation> latestUserExceptions;
+    private Long numSystemExceptions;
+    private List<ExceptionInformation> latestSystemExceptions;
+    private Double averageLatency;
+    private Long lastInvocationTime;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pip/pip-459.md
```diff
@@ -0,0 +1,329 @@
...
+### REST endpoint
+
+The worker exposes a new endpoint:
+
+`GET /admin/v3/functions/{tenant}/{namespace}/status/summary`
+
+The endpoint accepts two optional query parameters:
+
+- `limit`: maximum number of functions to return; must be greater than `0` when present
+- `continuationToken`: exclusive cursor based on function name in lexicographical order
```
lhotari: How would this `continuationToken` work and what would be the format? The Pulsar Functions Admin API is stateless and subsequent calls could go to different Function Worker instances in a cluster with multiple workers.

pip/pip-459.md
```diff
@@ -0,0 +1,329 @@
...
+Another alternative was overloading the existing list endpoint to optionally return enriched status objects. That was rejected because one endpoint returning different response types based on query flags complicates client typing, documentation, and compatibility.
+
+# General Notes
+
+This PIP intentionally solves the namespace-level operator workflow first. The same pattern may later be extended to sources and sinks, but that work is kept separate to avoid broadening the review scope.
+
+The worker-side implementation still computes summaries by querying per-function status and then aggregating results. Controlled parallelism improves latency, but it does not fundamentally change the cost model. Future work may explore more direct or cached aggregation paths if very large namespaces make the endpoint hot.
+
+# Links
+
```
lhotari, line 372: Add the discussion link here, https://lists.apache.org/thread/k792434hgm237qmtbn4fsd62hdzdt0h7

```suggestion
* Mailing List discussion thread: https://lists.apache.org/thread/k792434hgm237qmtbn4fsd62hdzdt0h7
* Mailing List voting thread: TBD
```

onceMisery: Thanks for raising this concern!

  You're absolutely right that the Admin API is stateless and requests can hit
  different workers. My design addresses this through deterministic, name-based
  cursor pagination:

  **How it works:**
  - continuationToken is simply the last function name from the previous page
  - Each worker independently sorts all function names lexicographically
  - The token acts as an exclusive lower bound to find the next page
  - No server-side state is required; the token is self-contained

  This approach works consistently across workers because the sorting is
  deterministic and metadata is replicated.

  **Trade-off:**
  If functions are created/deleted during pagination, clients may see duplicates
  or miss entries. This is a common limitation of stateless cursor-based pagination
  in distributed systems (similar to Kubernetes list pagination or DynamoDB queries).

  **One improvement I should make:**
  I also agree the API should return nextContinuationToken explicitly rather than only a bare list, so I can update the response shape to make the paging contract clear.

```json
  {
    "summaries": [...],
    "nextContinuationToken": "func-xyz"  // or null
  }
```

  Would this design work for you? I can update the PR accordingly.

onceMisery:
Thanks for raising this concern!
  You're absolutely right that the Admin API is stateless and requests can hit
  different workers. My design addresses this through deterministic, name-based
  cursor pagination:
 How it works:
 continuationToken is simply the last function name from the previous page
  Each worker independently sorts all function names lexicographically
The token acts as an exclusive lower bound to find the next page
No server-side state is required; the token is self-contained


  This approach works consistently across workers because the sorting is
  deterministic and metadata is replicated.


  Trade-off:
  If functions are created/deleted during pagination, clients may see duplicates
  or miss entries. This is a common limitation of stateless cursor-based pagination
  in distributed systems (similar to Kubernetes list pagination or DynamoDB queries).


 One improvement I should make:
  Currently the API returns List<FunctionStatusSummary>. I should wrap this in a
  response object with nextContinuationToken (null when no more data), so clients
  can easily detect the end of pagination:
  {
    "summaries": [...],
    "nextContinuationToken": "func-xyz" // or null
  }
 The name continuationToken might be ambiguous. What do you think of startAfter?  
 for example:  
  public List<FunctionStatusSummary> getFunctionsWithStatus(
      String tenant,
      String namespace,
      Integer limit,
      String startAfter) 


  Would this design work for you? I can update the PR accordingly.

pip/pip-459.md
```diff
@@ -0,0 +1,331 @@
...
+        NETWORK_ERROR,
+        INTERNAL_ERROR
+    }
+
+    private String name;
+    private SummaryState state;
+    private int numInstances;
+    private int numRunning;
+    private String error;
+    private ErrorType errorType;
```
shibd, line 98: hi, @onceMisery Thanks for PIP, Have we considered including a bit more information in this response, for example:

```
- `receivedTotal`
- `processedSuccessfullyTotal`
- `systemExceptionsTotal`
- `userExceptionsTotal`
- `avgProcessLatency`
- `userMetrics`
```

These values are already aggregate, and they could give operators a more direct view of a function's actual health instead of relying only on `RUNNING / STOPPED / PARTIAL / UNKNOWN`.

Another option would be to keep the default response lightweight, but add a query parameter to control the level of detail returned by the REST API. That would let us preserve the current "summary" use case while still supporting a more diagnostic view when needed.

pip/pip-459.md
shibd: Hi @onceMisery,
https://github.com/apache/pulsar/pull/25299#issuecomment-4024172832

This design makes sense to me. It can reduce the number of requests needed to fetch function stats. The only trade-off is that each worker needs to first list the functions under the namespace and sort them lexicographically.

1. Should we rename this parameter to something more explicit, such as startAfterFunctionName?

2. We should also clearly document the pagination behavior in the API, similar to how you described it in your comment.

pip/pip-459.md
onceMisery: I think this is a great suggestion. @shibd

onceMisery:
@shibd @lhotari 
 I fully agree with this suggestion: 
1. The continuationToken should be renamed to startAfterFunctionName.
2. add a query parameter to control the level of detail returned by the REST API. That would let us preserve the current "summary" use case while still supporting a more diagnostic view when needed.

shibd (approved):

onceMisery:
Hi @lhotari  @codelipenghui 

Just a gentle ping on this PIP. I started the discussion thread on the mailing list a while ago (link: https://lists.apache.org/thread/k792434hgm237qmtbn4fsd62hdzdt0h7), but haven't received any recent feedback.

Could you kindly take a look when you have a moment? 
Your insights and suggestions would be highly appreciated. Thanks!

PR #23620, opened 2024-11-21
[feat][broker] Prevent auto-creation of topics using legacy cluster-based naming scheme

zjxxzjwang:
### Motivation

Topic name rule

- old version: {tenant}/{cluster}/{namespace}/{topic}

- new version: {tenant}/{namespace}/{topic}

If the name of an automatically created topic contains a "/", the topic will be resolved into a four-segment format (containing a layer of cluster directories). As a result, the automatically created topic does not conform to the latest topic naming rules

### Modifications

No allowto automatically create topic names that contain slashes

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #25304, opened 2026-03-09
[fix] PIP-457: Remove support for V1 topic names and V1 Admin API (implementation)

merlimat:
This commit implements PIP-457 by removing all V1 topic name support and V1 Admin API endpoints from Apache Pulsar.

Topic naming changes:
- Remove V1 topic name format (persistent://tenant/cluster/namespace/topic) Only V2 format is now supported (persistent://tenant/namespace/topic)
- Remove V1 namespace name format (tenant/cluster/namespace) Only V2 format is now supported (tenant/namespace)
- TopicName and NamespaceName constructors now reject V1 format names
- fromPersistenceNamingEncoding() still handles 5-part managed ledger names for backward compatibility with existing stored data

V1 Admin API removal:
- Delete entire org.apache.pulsar.broker.admin.v1 package (Namespaces, PersistentTopics, NonPersistentTopics, Properties, Brokers, BrokerStats, Clusters, Functions, ResourceQuotas, SchemasResource)
- Delete V1 topic lookup endpoint (broker.lookup.v1.TopicLookup)
- Delete V1 WebSocket stats endpoint (websocket.admin.v1)
- Remove V1 servlet registrations from PulsarWebResource, ProxyServiceStarter, WebSocketServiceStarter
- Remove /admin/v1/ path prefix handling from all client admin impls (BrokerStatsImpl, BrokersImpl, LookupImpl, NamespacesImpl, NonPersistentTopicsImpl, ResourceQuotasImpl, SchemasImpl, TopicPoliciesImpl, TopicsImpl)
- Remove V1 cluster-aware namespace handling from NamespacesBase

Deprecated Properties API removal:
- Delete Properties.java interface (V1 term for Tenants)
- Remove Properties compat bridge methods from TenantsImpl
- Remove properties() accessor from PulsarAdmin/PulsarAdminImpl

TopicVersion enum removal:
- Delete TopicVersion.java (V1/V2 enum)
- Remove TopicVersion parameter from healthcheck API (Brokers interface, BrokersImpl, BrokersBase, CmdBrokers)
- Remove topicVersion HTTP query parameter from healthcheck requests
- Simplify PulsarService.runHealthCheck() and BrokerService

V1 heartbeat and SLA namespace cleanup:
- Remove V1 heartbeat namespace format (pulsar/cluster/brokerId) Rename V2 format methods to be the standard (pulsar/brokerId)
- Update SLA namespace to V2 format (sla-monitor/brokerId)
- Simplify HealthChecker to use single heartbeat topic
- Remove V1 heartbeat registration from ServiceUnitStateTableViewBase

Property to tenant terminology rename:
- Rename parameter/variable names from "property" to "tenant" in AdminResource, TransactionsBase, and 7 V2 Namespaces endpoints
- Fix @PathParam("property") to @PathParam("tenant") in V2 Namespaces
- Rename in test utilities and test variables

Configuration cleanup:
- Remove allowAutoTopicCreationType config (V1 only)
- Remove V1 auto-topic-creation settings from broker.conf/standalone.conf

Test cleanup:
- Delete V1-specific test suites (V1AdminApiTest, V1AdminApi2Test, V1ProducerConsumerTest, V1ProducerConsumerBase, V1ProxyAuthenticationTest)
- Remove V1 test cases from TopicNameTest, NamespaceNameTest
- Update all remaining tests to use V2 format exclusively
- Remove TopicVersion test parameters from AdminApiHealthCheckTest
- Update SLAMonitoringTest for 2-part SLA namespace format

Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTestBase.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTestBase.java
@@ -290,8 +290,10 @@ public abstract class ReplicatorTestBase extends TestRetrySupport {
         // Remove r4 from existing namespace replication clusters before shrinking the tenant,
         // since canUpdateCluster validation prevents removing a cluster that namespaces still reference.
         Set<String> targetClusters = Sets.newHashSet("r1", "r2", "r3");
-        for (String ns : admin1.namespaces().getNamespaces("pulsar")) {
-            admin1.namespaces().setNamespaceReplicationClusters(ns, targetClusters, false);
+        if (admin1.tenants().getTenants().contains("pulsar")) {
+            for (String ns : admin1.namespaces().getNamespaces("pulsar")) {
+                admin1.namespaces().setNamespaceReplicationClusters(ns, targetClusters, false);
+            }
         }
         updateTenantInfo("pulsar",
                 new TenantInfoImpl(Sets.newHashSet("appid1", "appid2", "appid3"),


pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -133,40 +126,20 @@ private TopicName(String completeTopicName) {
...
-            // The rest of the name can be in different forms:
-            // new:    tenant/namespace/<localName>
-            // legacy: tenant/cluster/namespace/<localName>
-            // Examples of localName:
-            // 1. some, name, xyz
-            // 2. xyz-123, feeder-2
-
-
-            parts = Splitter.on("/").limit(4).splitToList(rest);
+            // Expected format: tenant/namespace/<localName>
```
codelipenghui, line 135: `limit(3)` silently reinterprets V1 topic names instead of rejecting them.

For `persistent://tenant/cluster/namespace/topic`, this splits into `["tenant", "cluster", "namespace/topic"]` and is accepted as a valid V2 topic with tenant="tenant", namespace="cluster", localName="namespace/topic".

Users with old V1 topic names would **silently produce/consume on the wrong namespace**. Unlike `NamespaceName` which rejects 3-part names with a clear error, `TopicName` silently accepts.

Suggestion: keep `limit(4)`, detect `parts.size() == 4`, and throw:
```java
} else if (parts.size() == 4) {
    throw new IllegalArgumentException(
        "V1 topic names (with cluster component) are no longer supported. "
        + "Please use the V2 format: '<domain>://tenant/namespace/topic'. Got: " + completeTopicName);
}
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/metrics/AbstractMetrics.java
```diff
@@ -173,7 +173,9 @@ protected String parseNamespaceFromLedgerName(String ledgerName) {
         Matcher m = V2_LEDGER_NAME_PATTERN.matcher(ledgerName);
 
         if (m.matches()) {
-            return m.group(1);
+            // Ledger name format: tenant/namespace/domain/topic
+            // Extract only tenant/namespace (groups 2 and 3), excluding the domain.
+            return m.group(2) + "/" + m.group(3);
```
codelipenghui: For V2 ledger names this is a bugfix (old code incorrectly included domain in namespace label). But for V1 ledger names still in storage (`tenant/cluster/namespace/domain/topic`), this returns `tenant/cluster` instead of the correct namespace.

Also, even for V2, this changes the metrics `namespace` dimension from `tenant/namespace/domain` to `tenant/namespace`, which will break existing dashboards and alerts.

Should be documented as a known breaking change in release notes.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/TenantsBase.java
```diff
@@ -179,11 +177,8 @@ public void updateTenant(@Suspended final AsyncResponse asyncResponse,
                     if (!tenantAdmin.isPresent()) {
                         throw new RestException(Status.NOT_FOUND, "Tenant " + tenant + " not found");
                     }
-                    TenantInfo oldTenantAdmin = tenantAdmin.get();
-                    Set<String> newClusters = new HashSet<>(newTenantAdmin.getAllowedClusters());
-                    return canUpdateCluster(tenant, oldTenantAdmin.getAllowedClusters(), newClusters);
+                    return tenantResources().updateTenantAsync(tenant, old -> newTenantAdmin);
```
codelipenghui: The `canUpdateCluster()` validation was removed. Previously, updating a tenant checked that removed clusters had no active namespaces. Without this check, users can remove clusters from a tenant's allowed list while active namespaces still exist, potentially orphaning them.

While V1 cluster hierarchy is gone, the safety check itself (preventing orphaned namespaces) is still valuable for V2.

merlimat: That's a good catch. I think the check was broken as in that it only really worked for v1 namespaces...

pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/metrics/AbstractMetrics.java
merlimat: > But for V1 ledger names still in storage (tenant/cluster/namespace/domain/topic), this returns tenant/cluster instead of the correct namespace.

I think it's fine to assume there no more v1 topics at this point.

> Also, even for V2, this changes the metrics namespace dimension from tenant/namespace/domain to tenant/namespace, which will break existing dashboards and alerts.

Ok. I'll keep the same "wrong" label for v2. We can decide to change it separately.

merlimat: done

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
merlimat: 👍 done

pulsar-websocket/src/main/java/org/apache/pulsar/websocket/AbstractWebSocketHandler.java
```diff
@@ -271,28 +263,15 @@ protected void extractTopicName(HttpServletRequest request) {
...
-        int endPosition = isConsumer ? parts.size() - 1 : parts.size();
-        StringBuilder topicName = new StringBuilder(parts.get(startPosition));
-        while (++startPosition < endPosition) {
-            if (StringUtils.isEmpty(parts.get(startPosition))) {
-               continue;
-            }
-            topicName.append("/").append(parts.get(startPosition));
-        }
-        final String name = Codec.decode(topicName.toString());
+        final String name = Codec.decode(parts.get(7));
```
shibd, line 274: This now only decodes `parts.get(7)`, so a valid V2 WebSocket topic like `/ws/v2/consumer/persistent/tenant/ns/some/topic/with/slashes/sub` gets truncated to `some`. Even if we decide to reject `/` in local names, this should fail explicitly rather than silently parsing the wrong topic.

shibd: 1. I was thinking that our topic name convention should support /. So for a URL like:
```
/ws/v2/consumer/persistent/tenant/ns/some/topic/with/slashes/sub
                                     ^------> ^
```
we should probably parse the topic name as `some/topic`. 


2. Maybe we need to reject it when a  topic name include `/`

What do you think?

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
shibd: So, with this limitation, it means we cannot allow `/` in topic names. 

Would this be a breaking change? Currently, users are still able to produce and consume messages using topic names that contain `/`:  `persistent://tenant/cluster/namespace/topic  (topic name is: namespace/topic )`

Of course, we could restrict topic names from containing `/`.  Even if it is a breaking change, I think it would be meaningful in the long run.

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/NamespacesImpl.java
```diff
@@ -90,7 +88,7 @@ public List<String> getNamespaces(String tenant, String cluster) throws PulsarAd
     }
 
     public CompletableFuture<List<String>> getNamespacesAsync(String tenant, String cluster) {
```
shibd: Do we need to remove this method?

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
merlimat: @shibd 

> So, with this limitation, it means we cannot allow / in topic names.

V2 topic never allowed `/` in the local topic names. In all the cases these were interpreted by the parser as V1 topics (therefore the cluster/namespace was messed up in confusing ways). 

This was done to not break compatibility in the v1->v2 coexistence transition, since v1 topic literally had no restrictions.

pulsar-websocket/src/main/java/org/apache/pulsar/websocket/AbstractWebSocketHandler.java
merlimat: As mentioned in the other comment, slashes were never allowed in v2 topics. 

In this case we should definitely reject the consumer

pulsar-client-admin/src/main/java/org/apache/pulsar/client/admin/internal/NamespacesImpl.java
merlimat: You're right. This would call the `GET /admin/v2/{tenant}/{cluster}` which returns the policies of a non-existing namespace.

merlimat: Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewBase.java
```diff
@@ -50,17 +50,12 @@ protected void init(PulsarService pulsar) throws MetadataStoreException {
         // Add heartbeat and SLA monitor namespace bundle.
         NamespaceName heartbeatNamespace =
                 NamespaceService.getHeartbeatNamespace(brokerId, pulsar.getConfiguration());
-        NamespaceName heartbeatNamespaceV2 = NamespaceService
-                .getHeartbeatNamespaceV2(brokerId, pulsar.getConfiguration());
```
lhotari: Is this related to PIP-457?

lhotari (approved):
LGTM

pulsar-websocket/src/main/java/org/apache/pulsar/websocket/AbstractWebSocketHandler.java
lhotari: What has happened in the past is that if a user accidentially used a topic name with a slash, it became a v1 topic name accidentially. In v1 topics, the missing cluster, tenant and namespace get created automatically when they are missing. #23620 was added to prevent this from happening.

shibd: Thanks for explaining, I see.

PR #24991, opened 2025-11-16
[feat][monitor]  PIP-447 for Customizable Prometheus Labels for Topic Metrics

coderzc:
PIP: https://github.com/apache/pulsar/blob/master/pip/pip-447.md

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #25306, opened 2026-03-10
[feat][monitor] Add custom topic metric labels to OpenTelemetry metrics attributes

coderzc:
Main Issue: N/A

PIP: https://github.com/apache/pulsar/blob/master/pip/pip-447.md

### Motivation

This PR is part 2 of PIP-447.

Part 1 for Prometheus labels has already been merged in https://github.com/apache/pulsar/pull/24991. This follow-up extends the same custom topic metric label support to OpenTelemetry topic metrics so both metrics pipelines expose the same topic property based labels.

### Modifications

- add custom topic metric labels to OpenTelemetry topic metrics for persistent topics
- extend `PersistentTopicAttributes` to build OpenTelemetry attributes with custom labels
- update `OpenTelemetryTopicStats` to attach custom labels across topic metrics
- add broker tests for custom labels in OpenTelemetry metrics
- fix the renamed config setter in the test after rebasing on latest `master`

### Verifying this change

- Ran `mvn clean install -DskipTests`
- Added `OpenTelemetryCustomLabelsTest` to verify custom labels are exposed in OpenTelemetry metrics and omitted when the feature is disabled

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR in forked repository: https://github.com/coderzc/pulsar/pull/51

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/OpenTelemetryTopicStats.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/OpenTelemetryTopicStats.java
@@ -415,73 +415,94 @@ public class OpenTelemetryTopicStats implements AutoCloseable {
             var persistentTopicMetrics = persistentTopic.getPersistentTopicMetrics();
 
             var persistentTopicAttributes = persistentTopic.getTopicAttributes();
+            // For persistent topics, get custom attributes once and reuse
+            var customLabels = persistentTopicAttributes.getCustomAttributes();
+            var attributesWithCustomLabels = persistentTopicAttributes
+                .buildAttributesWithCustomLabels(attributes, customLabels);
             var managedLedger = persistentTopic.getManagedLedger();
             var managedLedgerStats = persistentTopic.getManagedLedger().getStats();
-            storageCounter.record(managedLedgerStats.getStoredMessagesSize(), attributes);
-            storageLogicalCounter.record(managedLedgerStats.getStoredMessagesLogicalSize(), attributes);
-            storageBacklogCounter.record(managedLedger.getEstimatedBacklogSize(), attributes);
-            storageOffloadedCounter.record(managedLedger.getOffloadedSize(), attributes);
-            storageInCounter.record(managedLedgerStats.getReadEntriesSucceededTotal(), attributes);
-            storageOutCounter.record(managedLedgerStats.getAddEntrySucceedTotal(), attributes);
+            storageCounter.record(managedLedgerStats.getStoredMessagesSize(), attributesWithCustomLabels);
+            storageLogicalCounter.record(managedLedgerStats.getStoredMessagesLogicalSize(), attributesWithCustomLabels);
+            storageBacklogCounter.record(managedLedger.getEstimatedBacklogSize(), attributesWithCustomLabels);
+            storageOffloadedCounter.record(managedLedger.getOffloadedSize(), attributesWithCustomLabels);
+            storageInCounter.record(managedLedgerStats.getReadEntriesSucceededTotal(), attributesWithCustomLabels);
+            storageOutCounter.record(managedLedgerStats.getAddEntrySucceedTotal(), attributesWithCustomLabels);
 
             backlogQuotaLimitSize.record(
-                    topic.getBacklogQuota(BacklogQuota.BacklogQuotaType.destination_storage).getLimitSize(),
-                    attributes);
+                topic.getBacklogQuota(BacklogQuota.BacklogQuotaType.destination_storage).getLimitSize(),
+                attributesWithCustomLabels);
             backlogQuotaLimitTime.record(
-                    topic.getBacklogQuota(BacklogQuota.BacklogQuotaType.message_age).getLimitTime(),
-                    attributes);
-            backlogQuotaAge.record(topic.getBestEffortOldestUnacknowledgedMessageAgeSeconds(), attributes);
+                topic.getBacklogQuota(BacklogQuota.BacklogQuotaType.message_age).getLimitTime(),
+                attributesWithCustomLabels);
+            backlogQuotaAge.record(topic.getBestEffortOldestUnacknowledgedMessageAgeSeconds(),
+                attributesWithCustomLabels);
             var backlogQuotaMetrics = persistentTopicMetrics.getBacklogQuotaMetrics();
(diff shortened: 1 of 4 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/OpenTelemetryTopicStats.java
```diff
@@ -415,73 +415,94 @@ private void recordMetricsForTopic(Topic topic) {
             var persistentTopicMetrics = persistentTopic.getPersistentTopicMetrics();
 
             var persistentTopicAttributes = persistentTopic.getTopicAttributes();
+            // For persistent topics, get custom attributes once and reuse
+            var customLabels = persistentTopicAttributes.getCustomAttributes();
+            var attributesWithCustomLabels = persistentTopicAttributes
```
shibd: Custom labels are only merged into the persistent-only metrics below. The AbstractTopic topic counters above (`subscription`, `producer`, `consumer`, `message`, `bytes`, `publish rate limit`) still record with `attributes`, so this change does not actually expose custom labels across all OpenTelemetry topic metrics.

pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/OpenTelemetryCustomLabelsTest.java
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.pulsar.opentelemetry.OpenTelemetryAttributes;
+import org.awaitility.Awaitility;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+@Test(groups = "broker")
+public class OpenTelemetryCustomLabelsTest extends BrokerTestBase {
+
+    private static final Set<String> ALLOWED_CUSTOM_METRIC_LABEL_KEYS = Set.of("__SLA_TIER", "APP_OWNER");
```
shibd: The allow-list key here is `__SLA_TIER`, but the test later writes `SLA_TIER`. `getCustomMetricLabelsMap()` filters by exact property key before lowercasing for OpenTelemetry output, so this label is not actually whitelisted.

pulsar-broker/src/test/java/org/apache/pulsar/broker/stats/OpenTelemetryCustomLabelsTest.java
```diff
@@ -0,0 +1,226 @@
...
+            .put(OpenTelemetryAttributes.PULSAR_DOMAIN, "persistent")
+            .put(OpenTelemetryAttributes.PULSAR_TENANT, "prop")
+            .put(OpenTelemetryAttributes.PULSAR_NAMESPACE, "prop/ns-abc")
+            .put(OpenTelemetryAttributes.PULSAR_TOPIC, topic2)
+            .put("sla_tier", "platinum")
+            .put("app_owner", "team-b")
+            .build();
+
+        // Verify metrics contain custom labels for topic1
+        assertMetricLongSumValue(metrics, OpenTelemetryTopicStats.MESSAGE_IN_COUNTER, attributesTopic1, 5);
```
shibd, line 155: This test is asserting `MESSAGE_IN_COUNTER` / `PRODUCER_COUNTER`, but the patch only changes the persistent-topic metrics in the lower block (`storage*`, backlog, transaction, compaction, delayed subscription). Even if this passes, it does not verify the metrics that were actually modified here.

pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/OpenTelemetryTopicStats.java
shibd: https://github.com/apache/pulsar/pull/25306/changes#diff-e3d8386fc3f9292780771f4eae19d9297637d374244f9c4b3f89b6c0e442e2e3R399-R412

pulsar-broker/src/main/java/org/apache/pulsar/broker/stats/OpenTelemetryTopicStats.java
```diff
@@ -415,73 +415,94 @@ private void recordMetricsForTopic(Topic topic) {
...
+                .buildAttributesWithCustomLabels(attributes, customLabels);
             var managedLedger = persistentTopic.getManagedLedger();
             var managedLedgerStats = persistentTopic.getManagedLedger().getStats();
-            storageCounter.record(managedLedgerStats.getStoredMessagesSize(), attributes);
-            storageLogicalCounter.record(managedLedgerStats.getStoredMessagesLogicalSize(), attributes);
-            storageBacklogCounter.record(managedLedger.getEstimatedBacklogSize(), attributes);
-            storageOffloadedCounter.record(managedLedger.getOffloadedSize(), attributes);
-            storageInCounter.record(managedLedgerStats.getReadEntriesSucceededTotal(), attributes);
-            storageOutCounter.record(managedLedgerStats.getAddEntrySucceedTotal(), attributes);
+            storageCounter.record(managedLedgerStats.getStoredMessagesSize(), attributesWithCustomLabels);
```
shibd: Could we push this concern down into `PersistentTopicAttributes` instead of threading `attributesWithCustomLabels` through each recorder here? `TopicAttributes` already exposes `getCommonAttributes()`, so overriding that in `PersistentTopicAttributes` to merge in `getCustomAttributes()` would let both the shared topic metrics and the persistent-only metrics pick up the custom labels automatically. That would also reduce the amount of mechanical churn in `OpenTelemetryTopicStats` and make it harder to miss one metric when new recorders are added later.

coderzc: Thanks, I agree the merge logic should be centralized in `PersistentTopicAttributes`.

My main concern is overriding `getCommonAttributes()`: today it is a cheap accessor over pre-built static attributes, while custom labels are dynamic and require reading topic properties. Hiding that work in a common getter makes the behavior less obvious.

I kept `getCommonAttributes()` unchanged, but moved the custom-label merge closer to `PersistentTopicAttributes` by introducing an explicit per-scrape attribute snapshot there. This keeps the dynamic work out of the common getter while also removing most of the repetitive merging from `OpenTelemetryTopicStats`.

shibd (approved):

PR #25311, opened 2026-03-11
[improve][test] Introduce shared Pulsar cluster for faster integration tests

merlimat:
## Motivation

Many broker integration tests start a full Pulsar service (broker + bookie + metadata store) per test class — or even per test method. This is extremely slow and resource-intensive. For example, `NullValueTest` previously took **~31 seconds**, with almost all of that time spent on service startup and shutdown rather than the actual test logic.

## Modifications

Introduces two new test infrastructure classes:

### `SharedPulsarCluster`
A JVM-wide singleton that manages a lightweight Pulsar cluster:
- **Single bookie** with `DbLedgerStorage`, no journal sync/writes, minimal RocksDB cache, and `UnpooledHeap` allocator for low memory footprint
- **Single broker** with in-memory metadata store (no ZooKeeper dependency)
- **JVM shutdown hook** that closes everything and deletes all temporary bookie data directories
- Default `test-cluster` and `test-tenant` pre-created at startup

### `SharedPulsarBaseTest`
Base class for tests that use the shared cluster:
- `@BeforeClass` initializes `admin` and `pulsarClient` from the singleton (once per class)
- `@BeforeMethod` creates a unique namespace per test method for full isolation
- `@AfterMethod` force-deletes the namespace (cascading to all topics)
- `newTopicName()` helper generates unique topic names within the test namespace

### Converted test
- **NullValueTest** — first proof-of-concept conversion, replacing `BrokerTestBase` with `SharedPulsarBaseTest`

## Test Timing

With the shared cluster, `NullValueTest` runs in **~4 seconds** (including ~3s singleton init on first use), compared to **~31 seconds** when it started its own service.

## Verifying this change

All 8 test invocations pass (4 test methods × 2 data provider params):
```
mvn test -pl pulsar-broker -Dtest=NullValueSharedTest -DfailIfNoTests=false
```

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SharedPulsarBaseTest.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Creates a unique namespace for the current test method. The namespace is automatically
+     * cleaned up in {@link #cleanupSharedTest()}.
+     */
+    @BeforeMethod(alwaysRun = true)
+    public void setupSharedTest() throws Exception {
+        String nsName = "test-" + UUID.randomUUID().toString().substring(0, 8);
+        String ns = SharedPulsarCluster.TENANT_NAME + "/" + nsName;
+        namespace = ns;
+        admin.namespaces().createNamespace(ns, Set.of(SharedPulsarCluster.CLUSTER_NAME));
```
lhotari, line 139: It would be useful to move this logic as a method directly to SharedPulsarCluster class. The benefit of that would be to reuse SharedPulsarCluster later in Junit Jupiter tests when we start doing that for Pulsar 5.0 implementation. This can also be done later.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SharedPulsarBaseTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public void cleanupSharedTest() throws Exception {
+        String ns = namespace;
+        if (ns != null) {
+            namespace = null;
+            try {
+                admin.namespaces().deleteNamespace(ns, true);
+                log.info("Deleted test namespace: {}", ns);
+            } catch (Exception e) {
+                log.warn("Failed to delete namespace {}: {}", ns, e.getMessage());
+            }
```
lhotari, line 156: Similar comment as above

PR #25312, opened 2026-03-11
[fix][broker]system topic was created with different partitions acrossing clusters after enabled namespace-level replication

poorbarcode:
### Motivation

The situation that replication does not work as expected.

- cluster: `c1` 
  - namespace: `public/default`

 The system topic `__change_events` has already been created as a non-partitioned topic; in other words, the topic named `__change_events` has no partition.

- cluster: `c2`, has the following configurations
  - `allowAutoTopicCreationType=partitioned`
  - `defaultNumPartitions=1`
 
These two configurations mean that the new topic will be created as a partitioned topic and contain `1` partition. 

If we enable replication for the namespace between two clusters, the following issue will happen

- The cluster `c2` will create a partitioned topic `__change_events` , and it will have a partition like this `__change_events-partition-0`
- The cluster `c1` will start replication for the non-partitioned topic `__change_events`, which triggers a non-partitioned topic creation on the `c2` cluster.
- `c2` will also start a replication to `c1` since it has a partitioned topic which has a partition named `__change_events-partition-0`, it triggers a partitioned creation on `c1` cluster.

Eventually, both clusters will hold both partitioned topics and non-partitioned topics as follows, but they have the same name.
- `admin/partitioned-toipic`: `{partitions: 1}`
- `managed-ledger/persistent`: `[__change_events-partition-0, __change_events]`

Because this is a system topic, it will lead to fragmentation and cause serious problems

### Modifications

- Create a topic with the same type as the remote side, if the topic has already been created on the remote side.
- The change only affects if the following conditions match
    - Enable replication
    - The topic creation was triggered by `consumer/producer`'s starting
    - The remote side has already created the topic
    - `broker.conf -> isCreateTopicToRemoteClusterForReplication` is `true`
    - enabled topic auto-creation

This PR will not cause such an infinite loop: the local cluster queries the remote cluster for the topic type and partition count, and on the remote side, the request is handled by handlePartitionMetadataRequest. When determining the auto-creation policy, the remote cluster would otherwise query the local cluster again. However, during the initial query ("the local cluster queries the remote cluster for the topic type and partition count"), the request is already limited to existing topics only, so the remote cluster will not execute the logic for fetching the auto-creation policy.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalZKTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalZKTest.java
@@ -799,7 +799,7 @@ public class OneWayReplicatorUsingGlobalZKTest extends OneWayReplicatorTest {
         Awaitility.await().untilAsserted(() -> {
             PersistentTopic persistentTopic1 = (PersistentTopic) broker1.getTopic(tp, false).join().get();
             assertTrue(persistentTopic1.getReplicators().isEmpty());
-            PersistentTopic persistentTopic2 = (PersistentTopic) broker1.getTopic(tp, false).join().get();
+            PersistentTopic persistentTopic2 = (PersistentTopic) broker2.getTopic(tp, false).join().get();
             assertTrue(persistentTopic2.getReplicators().isEmpty());
         });
         admin1.topics().delete(tp, false);


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -3458,6 +3478,78 @@ private CompletableFuture<PartitionedTopicMetadata> createDefaultPartitionedTopi
...
+                    if (t != null) {
+                        Throwable actEx = FutureUtil.unwrapCompletionException(t);
+                        if (actEx instanceof PulsarClientException.NotFoundException
+                            | actEx instanceof PulsarClientException.TopicDoesNotExistException
+                            | actEx instanceof PulsarAdminException.NotFoundException) {
+                            future.complete(TopicExistsInfo.newTopicNotExists());
+                        } else {
+                            FutureUtil.completeAfter(future,
+                                findRemoteTopicMetadataForAutoCreation(topicName, remoteClusters, index + 1, actEx));
+                        }
```
coderzc, line 3534: Add `return null;`

poorbarcode: Fixed, Thanks

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalZKTest.java
```diff
@@ -669,6 +676,136 @@ public void testSetClustersAndAllowedClusters() throws Exception {
...
+        });
+
+        // cleanup.
+        admin1.topics().setReplicationClusters(tp, Arrays.asList(cluster1));
+        admin2.topics().setReplicationClusters(tp, Arrays.asList(cluster2));
+        Awaitility.await().untilAsserted(() -> {
+            PersistentTopic persistentTopic1 = (PersistentTopic) broker1.getTopic(tp, false).join().get();
+            assertTrue(persistentTopic1.getReplicators().isEmpty());
+            PersistentTopic persistentTopic2 = (PersistentTopic) broker1.getTopic(tp, false).join().get();
+            assertTrue(persistentTopic2.getReplicators().isEmpty());
```
Demogorgon314, line 803:
```suggestion
            PersistentTopic persistentTopic1 = (PersistentTopic) broker1.getTopic(tp, false).join().get();
            assertTrue(persistentTopic1.getReplicators().isEmpty());
            PersistentTopic persistentTopic2 = (PersistentTopic) broker2.getTopic(tp, false).join().get();
            assertTrue(persistentTopic2.getReplicators().isEmpty());
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -3458,6 +3478,79 @@ private CompletableFuture<PartitionedTopicMetadata> createDefaultPartitionedTopi
...
+                }
+                PulsarClient client = getReplicationClient(remoteCluster, clusterData);
+                CompletableFuture<TopicExistsInfo> future = new CompletableFuture<>();
+                client.getPartitionsForTopic(topicName.toString(), false).handle((topics, t) -> {
+                    if (t != null) {
+                        Throwable actEx = FutureUtil.unwrapCompletionException(t);
+                        if (actEx instanceof PulsarClientException.NotFoundException
+                            | actEx instanceof PulsarClientException.TopicDoesNotExistException
+                            | actEx instanceof PulsarAdminException.NotFoundException) {
+                            future.complete(TopicExistsInfo.newTopicNotExists());
```
Demogorgon314, line 3530: If there is 3+ clusters, this branch returns TopicExistsInfo.newTopicNotExists() as soon as the first remote cluster says the topic is missing, the later clusters are never consulted.

coderzc: +1

poorbarcode: Yes, take the first one as the standard. Since the second cluster has already been compared with the first cluster when creating the topic, we can assume that the old cluster is consistent

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalZKTest.java
poorbarcode: Modified

PR #25321, opened 2026-03-14
[improve][offload] PIP-461: Add queued latency metrics for offloader executors

BewareMyPower:
### Motivation

`BlobStoreManagedLedgerOffloader` uses `OrderedScheduler` instances for both offload/write work and read work, but those executors currently do not expose an offloader-specific queued-latency metric.

The jcloud offloader runs blocking blob-store operations on these executors:
- `scheduler` handles synchronous offload and delete work.
- `readExecutor` handles blocking offloaded-ledger open/read/close work.

When these tasks queue up behind other blocking work, it is hard to tell from the existing offloader metrics whether the executors themselves are saturated.

### Modifications

This change adds two new offloader metrics:
- `brk_ledgeroffloader_offload_executor_queue_latency`
- `brk_ledgeroffloader_read_offload_executor_queue_latency`

The implementation:
- Extends `LedgerOffloaderStats` and `LedgerOffloaderStatsImpl` with the two queue-latency summaries.
- Adds a small wrapper utility to record executor queue wait time when a task actually starts running.
- Instruments the jcloud offloader paths that submit blocking work to `scheduler` and `readExecutor`.
- Instruments the offloaded read-handle implementations so queued latency is also tracked for read and close tasks after the handle is opened.
- Extends existing tests to assert the new metrics are recorded.

The queue-latency metrics use the same topic/namespace label behavior as the existing ledger offloader metrics.

### Verifying this change

- `./mvnw -pl managed-ledger -Dmaven.test.skip=true install`
- `./mvnw -pl tiered-storage/jcloud -Dtest=BlobStoreManagedLedgerOffloaderTest,BlobStoreBackedReadHandleImplTest -Dsurefire.failIfNoSpecifiedTests=false test`

Notes:
- The jcloud targeted test run passed locally: 26 tests, 0 failures.
- The updated `pulsar-broker` test class was not run locally because this checkout currently has unrelated compile failures in `pulsar-broker` / `pulsar-metadata`.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR in forked repository: N/A

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloaderStats.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/LedgerOffloaderStats.java
@@ -48,6 +48,10 @@ public interface LedgerOffloaderStats extends AutoCloseable {
 
     void recordReadOffloadDataLatency(String topic, long latency, TimeUnit unit);
 
+    void recordOffloadExecutorQueueLatency(String topic, long latency, TimeUnit unit);
+
+    void recordReadOffloadExecutorQueueLatency(String topic, long latency, TimeUnit unit);
+
     void recordDeleteOffloadOps(String topic, boolean succeed);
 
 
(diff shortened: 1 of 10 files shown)


BewareMyPower:
I will write a proposal for it since it adds new metrics

PR #25322, opened 2026-03-14
[improve][pip] PIP-461: Add queued latency metrics for offloader executors

BewareMyPower:
### Motivation

The default read threads for tiered storage is 2, when there are many concurrent read operations, many will be queued for long but there is no metric for it, so it's hard to tune the threads.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [x] The metrics

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:

### DIFF
--- a/pip/pip-461.md
+++ b/pip/pip-461.md
@@ -227,5 +227,5 @@ queueing problem exists today.
 
 ## Links
 
-* Mailing List discussion thread:
+* Mailing List discussion thread: https://lists.apache.org/thread/l3hyysy61yxwj3rcszqtghtf2y49kdpm
 * Mailing List voting thread:


pip/pip-461.md
```diff
@@ -0,0 +1,231 @@
...
+
+1. capture `System.nanoTime()` at submission time
+2. when the wrapped task begins execution, compute `System.nanoTime() - enqueueTime`
+3. record the value into the corresponding offloader queue metric
+4. run the original task body
+
+The helper is used in these paths:
+
+- `BlobStoreManagedLedgerOffloader.offload(...)`
+- `BlobStoreManagedLedgerOffloader.streamingOffload(...)`
```
shibd, line 100: Maybe we can leave `streamingOffload()` out of scope for now, since this path does not seem to be wired into the current production offload flow yet. If you would still like to include it here, could we also note that there is an intentional delayed resubmission in `BlobStoreManagedLedgerOffloader.streamingOffloadLoop()` and subtract that delay from the queue-latency metric?

```java
scheduler.chooseThread(segmentInfo)
        .schedule(() -> {
            streamingOffloadLoop(partId, dataObjectLength);
        }, 100, TimeUnit.MILLISECONDS);
```

Otherwise the metric may end up including this built-in 100ms delay, instead of only the actual executor queue time.

pip/pip-461.md
```diff
@@ -0,0 +1,231 @@
...
+     executor queue before starting execution
+   - Attributes:
+     - `namespace`
+     - `topic` when topic-level offloader metrics are enabled
+   - Unit: microseconds
+
+2. `brk_ledgeroffloader_read_offload_executor_queue_latency`
+   - Description: Time that a blocking tiered-storage read task spends waiting in the offloader read executor queue
+     before starting execution
+   - Attributes:
```
shibd, line 145: Maybe it could be helpful to add one more attribute here, something like `operation` (or `caller`). These executors seem to handle more than just one kind of work, so separating values such as `offload`, `delete`, `open`, `read`, and `close` might make the queue-latency metrics easier to interpret in practice. For example, it would be easier to tell whether the latency is coming from read traffic itself or from extra work scheduled on the same executor.

PR #25346, opened 2026-03-18
[fix][broker] Fix concurrency bug in BucketDelayedDeliveryTracker

merlimat:
## Summary

- Fix a concurrency bug in `BucketDelayedDeliveryTracker` where `containsMessage()` and `nextDeliveryTime()` used a `StampedLock` optimistic read pattern, but no write method ever acquired the `StampedLock` write lock.
- This meant the optimistic read validation **always succeeded**, leaving these read methods effectively unsynchronized while `addMessage()` concurrently modified shared state (`lastMutableBucket`, `immutableBuckets`).
- Replace the broken `StampedLock` pattern with `synchronized`, consistent with all write methods that already use `synchronized`.

The symptom for this issue was a flaky test: 

```
  Error:  Failures: 
  Error:  org.apache.pulsar.broker.delayed.bucket.BucketDelayedDeliveryTrackerThreadSafetyTest.testConcurrentContainsMessageWithWrites
  [INFO]   Run 1: PASS
  Error:    Run 2: BucketDelayedDeliveryTrackerThreadSafetyTest.testConcurrentContainsMessageWithWrites:176 No exceptions should occur during concurrent operations expected [0] but found [1]
```

## Documentation

- [x] `doc-not-needed`
(Your PR doesn't need any doc update)

_No response_

### Tip

Add the labels `ready-to-test` and `area/test` to trigger the CI.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/BucketDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/bucket/BucketDelayedDeliveryTracker.java
@@ -43,7 +43,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.locks.StampedLock;
 import java.util.stream.Collectors;
 import javax.annotation.concurrent.ThreadSafe;
 import lombok.Getter;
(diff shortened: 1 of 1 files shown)


Denovo1998:
@merlimat 
Some background information about this matter is in the Bug section of https://github.com/apache/pulsar/issues/24600.

Directly changing back to synchronized does not seem to be a good solution, which may lead to poor performance, especially the call frequency of the read path of delayed messages is very high.

Maybe you can help promote https://github.com/apache/pulsar/pull/24739.

merlimat:
@Denovo1998 that seems to be quite a more complex change (not saying it's wrong, though it's hard to grasp all the changes and their implications from the PR). Probably @coderzc will have more context to review that. 

For the time being, I think removing the race condition (and the exception being thrown from the map or list) is definitely worth the cost of the synchronization.

coderzc (approved):

coderzc:
> @Denovo1998 that seems to be quite a more complex change (not saying it's wrong, though it's hard to grasp all the changes and their implications from the PR). Probably @coderzc will have more context to review that.
> 
> 

+1

Denovo1998:
+1

Denovo1998 (approved):

PR #25349, opened 2026-03-18
[fix][broker] Missing assignment of the corresponding InFlightTask in PersistentReplicator.readEntriesFailed

gosonzhang:
### Motivation

The situation that replication does not work as expected.

Fixed an implementation bug:

When PersistentReplicator replicates an entry, it constructs an InFlightTask object to track the replication process. However, in the readEntriesFailed() function, the failure result is not filled into the corresponding InFlightTask object. This caused the readMoreEntries() function to continuously report the "Not scheduling read due to pending read or no permits." error after encountering a failure in the readEntriesFailed() function, blocking replica replication and causing geo replication stuck.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

https://github.com/gosonzhang/pulsar/pull/1

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
@@ -508,6 +508,9 @@ public abstract class PersistentReplicator extends AbstractReplicator
             return;
         }
 
+        InFlightTask inFlightTask = (InFlightTask) ctx;
+        inFlightTask.setEntries(Collections.emptyList());
+
         // Reduce read batch size to avoid flooding bookies with retries
         readBatchSize = topic.getBrokerService().pulsar().getConfiguration().getDispatcherMinReadBatchSize();
 
@@ -988,5 +991,10 @@ public abstract class PersistentReplicator extends AbstractReplicator
     @VisibleForTesting
     String getReplicatorId() {
         return  replicatorId;
+    }
+
+    @VisibleForTesting
+    public void incrementWaitForCursorRewindingRefCnf() {
+        waitForCursorRewindingRefCnf++;
     }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
@@ -992,9 +992,4 @@ public abstract class PersistentReplicator extends AbstractReplicator
     String getReplicatorId() {
         return  replicatorId;
     }
-
-    @VisibleForTesting
-    public void incrementWaitForCursorRewindingRefCnf() {
-        waitForCursorRewindingRefCnf++;
-    }
 }
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
@@ -642,12 +642,8 @@ public class ReplicatorTest extends ReplicatorTestBase {
     @Test(timeOut = 30000)
     public void testReplicatorClearBacklog() throws Exception {
 
-        // This test is to verify that reset cursor fails on global topic
-        SortedSet<String> testDests = new TreeSet<>();
-
         final TopicName dest = TopicName
                 .get(BrokerTestUtil.newUniqueName("persistent://pulsar/ns/clearBacklogTopic"));
-        testDests.add(dest.toString());
 
         @Cleanup
         MessageProducer producer1 = new MessageProducer(url1, dest);
@@ -655,12 +651,31 @@ public class ReplicatorTest extends ReplicatorTestBase {
         @Cleanup
         MessageConsumer consumer1 = new MessageConsumer(url3, dest);
 
-        // Produce from cluster1 and consume from the rest
+        // Produce from cluster1 to trigger topic and replicator creation
         producer1.produce(2);
+
         PersistentTopic topic = (PersistentTopic) pulsar1.getBrokerService().getTopicReference(dest.toString()).get();
-        PersistentReplicator replicator = (PersistentReplicator) spy(
-                topic.getReplicators().get(topic.getReplicators().keySet().stream().toList().get(0)));
-        replicator.incrementWaitForCursorRewindingRefCnf();
+        PersistentReplicator replicator = (PersistentReplicator) topic.getReplicators()
+                .get(topic.getReplicators().keySet().stream().toList().get(0));
+
+        // Wait until the first batch of messages is fully replicated (backlog = 0),
+        // so that disconnect() won't be rejected due to non-zero backlog
+        Awaitility.await().untilAsserted(() -> {
+            assertEquals(replicator.getNumberOfEntriesInBacklog(), 0);
+        });
+
+        // Disconnect replicator to stop geo-replication, so new messages will form backlog
+        pauseReplicator(replicator);
+
+        // Produce more messages while replication is paused ¡ª these will accumulate as backlog
+        producer1.produce(2);
+
+        // wait for backlog to accumulate
+        Awaitility.await().untilAsserted(() -> {
+            assertEquals(replicator.getNumberOfEntriesInBacklog(), 2);
+        });
+
+        // Clear the backlog
         replicator.clearBacklog().get();
         Thread.sleep(100);
         replicator.updateRates(); // for code-coverage
@@ -672,12 +687,8 @@ public class ReplicatorTest extends ReplicatorTestBase {
     @Test(timeOut = 30000)
     public void testReplicatorExpireMsgAsync() throws Exception {
 
-        // This test is to verify that reset cursor fails on global topic
-        SortedSet<String> testDests = new TreeSet<>();
-
         final TopicName dest = TopicName
                 .get(BrokerTestUtil.newUniqueName("persistent://pulsar/ns/clearBacklogTopic"));
-        testDests.add(dest.toString());
 
         @Cleanup
         MessageProducer producer1 = new MessageProducer(url1, dest);
@@ -685,12 +696,31 @@ public class ReplicatorTest extends ReplicatorTestBase {
         @Cleanup
         MessageConsumer consumer1 = new MessageConsumer(url3, dest);
 
-        // Produce from cluster1 and consume from the rest
+        // Produce from cluster1 to trigger topic and replicator creation
         producer1.produce(2);
+
         PersistentTopic topic = (PersistentTopic) pulsar1.getBrokerService().getTopicReference(dest.toString()).get();
-        PersistentReplicator replicator = (PersistentReplicator) spy(
-                topic.getReplicators().get(topic.getReplicators().keySet().stream().toList().get(0)));
-        replicator.incrementWaitForCursorRewindingRefCnf();
+        PersistentReplicator replicator = (PersistentReplicator) topic.getReplicators()
+                .get(topic.getReplicators().keySet().stream().toList().get(0));
+
+        // Wait until the first batch of messages is fully replicated (backlog = 0),
+        // so that disconnect() won't be rejected due to non-zero backlog
+        Awaitility.await().untilAsserted(() -> {
+            assertEquals(replicator.getNumberOfEntriesInBacklog(), 0);
+        });
+
+        // Disconnect replicator to stop geo-replication, so new messages will form backlog
+        pauseReplicator(replicator);
+
+        // Produce more messages while replication is paused ¡ª these will accumulate as backlog
+        producer1.produce(2);
+
+        // wait for backlog to accumulate
+        Awaitility.await().untilAsserted(() -> {
+            assertEquals(replicator.getNumberOfEntriesInBacklog(), 2);
+        });
+
+        // Clear the backlog
         replicator.clearBacklog().get();
         Thread.sleep(100);
         replicator.updateRates(); // for code-coverage
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
@@ -44,8 +44,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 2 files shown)


lhotari (changes requested):
Please add a test case to reproduce the issue and to validate that the problem gets fixed.

lhotari:
@poorbarcode What is the intended way to handle the InFlightTask when the reading fails?

gosonzhang:
@lhotari ok

lhotari:
@gosonzhang I noticed that you added a heavily mocked test case.

I'd recommend reproducing the issue with a test that uses the MockedPulsarServiceBaseTest. Tests using MockedPulsarServiceBaseTest are more like integration tests and there isn't a need for heavy mocking in the test code. This results in more representative tests for reproducing issues.

To inject read errors, you can use this type of approach:
https://github.com/apache/pulsar/blob/b6cfecce5f3a1eecbf6f5df81cb835fbbfe35980/pulsar-broker/src/test/java/org/apache/pulsar/client/api/KeySharedSubscriptionBrokerCacheTest.java#L198-L204

lhotari:
Please merge master branch or rebase your changes to get fixes to flaky tests. A lot of fixes were merged recently after many tests became flaky after some test infrastructure changes.

gosonzhang:
@lhotari Okay, let me see.

gosonzhang:
@lhotari Every CI attempt fails, even for parts that haven't been modified. I'm unsure what's causing this.

Regarding the bug, @poorbarcode, from my analysis, it's quite clear and involves multiple versions. The test results confirm it's indeed the issue.

For the test cases:
1. after analysis, placing them in `ReplicatorTest.testReplicatorWithFailedAck()` is more appropriate.
2. I also modified `testReplicatorExpireMsgAsync()` and `testReplicatorClearBacklog()`, primarily removing the call to `replicator.readEntriesFailed()`. Since `InflightTask` is private, it can't be used in these cases, and passing null doesn't meet expectations. These cases aim to stop replication upon failure, so I add test settings to the `waitForCursorRewindingRefCnf` value prevents it from getting stuck in continuous replication checks.

lhotari:
> Every CI attempt fails, even for parts that haven't been modified. I'm unsure what's causing this.

@gosonzhang please rebase or merge latest origin/master.

gosonzhang:
> > Every CI attempt fails, even for parts that haven't been modified. I'm unsure what's causing this.
> 
> @gosonzhang please rebase or merge latest origin/master.

@lhotari I'm really stuck on this.

I've run the test cases for the broker-replication group multiple times locally without any errors[1]; and the test case that weren't affected by the changes shouldn't be throwing any errors either[2]:

1. image

2. image

lhotari:
> @lhotari I'm really stuck on this.
> 
> I've run the test cases for the broker-replication group multiple times locally without any errors[1]; and the test case that weren't affected by the changes shouldn't be throwing any errors either[2]:

@gosonzhang There are many flaky tests in Pulsar. Many issues are the result of bad test design. Another class of problems are the resource leaks from previously run tests. Resource leaks could include memory leaks or thread leaks. The impact of resource leaks is that the test runs very slowly and when the test design isn't great, it causes the test to fail due to slow execution. In some cases, the flaky tests are real production code bugs.

Recently we have made a change to migrate many tests to use org.apache.pulsar.broker.service.SharedPulsarCluster. It seems that there's at least a memory leak since it uses an in-memory metadata store. For very long test runs, it might be accumulating a significant amount used heap memory, resulting in frequent full GCs when running tests.

Claude Code with Claude 4.6 Opus/Sonnet is really good in fixing flaky tests and we are currently doing great progress in addressing the high amount of flakes in the Pulsar code base. At the same time when some optimizations are made, it brings up new flakes when the test execution dynamics changes.

If you run into a flaky test that blocks your progress, you can disable the test in your PR or move it to the flaky group so that it doesn't block PR merging. Please report any flakes that aren't already reported in https://github.com/apache/pulsar/issues?q=is%3Aissue%20state%3Aopen%20flaky

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -989,4 +992,9 @@ protected boolean hasPendingRead() {
     String getReplicatorId() {
         return  replicatorId;
     }
+
+    @VisibleForTesting
+    public void incrementWaitForCursorRewindingRefCnf() {
+        waitForCursorRewindingRefCnf++;
+    }
```
lhotari: Why is it necessary to change internal state of the implementation in the test? Is there a way to prevent this?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
```diff
@@ -1686,14 +1686,28 @@ public void testReplicatorWithFailedAck() throws Exception {
...
+                log.info("asyncReadEntriesOrWait will be failed");
+                callback.readEntriesFailed(new ManagedLedgerException("Mocked read failure"), ctx);
+                return null;
+            } else {
+                log.info("asyncReadEntriesOrWait will proceed normally");
+                return invocation.callRealMethod();
+            }
+        }).when(spyCursor).asyncReadEntriesOrWait(Mockito.anyInt(), Mockito.anyLong(),
+                Mockito.any(AsyncCallbacks.ReadEntriesCallback.class), Mockito.any(), Mockito.any(Position.class));
+
```
lhotari, line 1733: Is it possible to add a new test case? These tests might not be the best possible example of good test style. If you are able to create a new test class with a different style and minimal mocking, that would be preferred (such as using `pulsarTestContext.getMockBookKeeper().setReadHandleInterceptor(...)` to inject failures/delays).

lhotari:
> For the test cases:
> 
> 1. after analysis, placing them in `ReplicatorTest.testReplicatorWithFailedAck()` is more appropriate.

why is is useful to add to an existing test case? That seems to make the test case even harder to understand. Pulsar's tests aren't always good examples how to continue to create new tests so feel free to break out of the existing patterns and add new test classes when the base class limits the options.

> 2. I also modified `testReplicatorExpireMsgAsync()` and `testReplicatorClearBacklog()`, primarily removing the call to `replicator.readEntriesFailed()`. Since `InflightTask` is private, it can't be used in these cases, and passing null doesn't meet expectations. These cases aim to stop replication upon failure, so I add test settings to the `waitForCursorRewindingRefCnf` value prevents it from getting stuck in continuous replication checks.

The `waitForCursorRewindingRefCnf++` in tests doesn't seem to be a great idea since it modifies internal state. It brings the concern whether the test is sufficiently representative.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
gosonzhang: The functions `readEntriesFailed()` and `readEntriesComplete()` work like a car engine; they will keep running in a loop unless an exception occurs or the engine runs out of fuel.

The original test case aimed to execute `readEntriesFailed()` to stop the process, and then test whether the `ClearBacklog()` and `ExpireMsgAsync()` functions worked correctly. Based on the consideration of equivalent replacement, an exception flag is added to stop the copying process, thus halting the copying operation.

gosonzhang:
> > @lhotari I'm really stuck on this.
> > I've run the test cases for the broker-replication group multiple times locally without any errors[1]; and the test case that weren't affected by the changes shouldn't be throwing any errors either[2]:
> 
> @gosonzhang There are many flaky tests in Pulsar. Many issues are the result of bad test design. Another class of problems are the resource leaks from previously run tests. Resource leaks could include memory leaks or thread leaks. The impact of resource leaks is that the test runs very slowly and when the test design isn't great, it causes the test to fail due to slow execution. In some cases, the flaky tests are real production code bugs.
> 
> Recently we have made a change to migrate many tests to use org.apache.pulsar.broker.service.SharedPulsarCluster. It seems that there's at least a memory leak since it uses an in-memory metadata store. For very long test runs, it might be accumulating a significant amount used heap memory, resulting in frequent full GCs when running tests.
> 
> Claude Code with Claude 4.6 Opus/Sonnet is really good in fixing flaky tests and we are currently doing great progress in addressing the high amount of flakes in the Pulsar code base. At the same time when some optimizations are made, it brings up new flakes when the test execution dynamics changes.
> 
> If you run into a flaky test that blocks your progress, you can disable the test in your PR or move it to the flaky group so that it doesn't block PR merging. Please report any flakes that aren't already reported in https://github.com/apache/pulsar/issues?q=is%3Aissue%20state%3Aopen%20flaky

It is passed already in my repository and the  failure isn't caused by this PR changes.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
lhotari: I understand that it's not easy to reproduce concurrency issues. However, the PR description doesn't mention anything about a race condition. Based on the PR title and description it seems that it's only about missing assignment of the corresponding InFlightTask in PersistentReplicator.readEntriesFailed.
If some more conditions are involved, please clarify it in the description.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
gosonzhang: In this test case, mocking is mandatory, a precise triggering of the repair scenario is needed to determine if the problem has been resolved; otherwise, the system will continue running normally without triggering the case.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
gosonzhang: Like `readEntriesComplete()`, it executes in a single thread, so there are no race condition.

Adding `incrementWaitForCursorRewindingRefCnf()` is primarily to achieve the desired effect: calling `readEntriesFailed()` before modification will stop the copying process.

gosonzhang:
> > For the test cases:
> > 
> > 1. after analysis, placing them in `ReplicatorTest.testReplicatorWithFailedAck()` is more appropriate.
> 
> why is is useful to add to an existing test case? That seems to make the test case even harder to understand. Pulsar's tests aren't always good examples how to continue to create new tests so feel free to break out of the existing patterns and add new test classes when the base class limits the options.
> 
> > 2. I also modified `testReplicatorExpireMsgAsync()` and `testReplicatorClearBacklog()`, primarily removing the call to `replicator.readEntriesFailed()`. Since `InflightTask` is private, it can't be used in these cases, and passing null doesn't meet expectations. These cases aim to stop replication upon failure, so I add test settings to the `waitForCursorRewindingRefCnf` value prevents it from getting stuck in continuous replication checks.
> 
> The `waitForCursorRewindingRefCnf++` in tests doesn't seem to be a great idea since it modifies internal state. It brings the concern whether the test is sufficiently representative.

 What suggestions do you have?

The initial idea was to keep the execution logic consistent before and after the test case modification. Personally, I also think that constructing a test API to execute the `waitForCursorRewindingRefCnf++` operation is not a good approach, but I haven't found a better one.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
lhotari: what do you mean with "a precise triggering of the repair scenario is needed to determine if the problem has been resolved" exactly? What is the "precise triggering of the repair scenario"? What is required for triggering it?

If one of the challenges is to check logs in the test to validate behavior, it's possible to use org.apache.pulsar.utils.TestLogAppender for capturing logs in a test case.

lhotari:
> What suggestions do you have?
> 
> The initial idea was to keep the execution logic consistent before and after the test case modification. Personally, I also think that constructing a test API to execute the `waitForCursorRewindingRefCnf++` operation is not a good approach, but I haven't found a better one.

I don't know yet until I understand exactly what the test is currently doing and how the problem is triggered in production code. After understanding that it's easier to recommend option. I asked some questions in another comment.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
gosonzhang: I need the test cases to call readEntriesFailed() to ensure that the changes are executed and to confirm that the tests are running as expected.

lhotari: > I need the test cases to call readEntriesFailed() to ensure that the changes are executed and to confirm that the tests are running as expected.

That doesn't yet explain what triggers the problem in actual production code. With `pulsarTestContext.getMockBookKeeper().setReadHandleInterceptor(...)` you can inject failures that would result in `readEntriesFailed` to be called. It's possible to then validate whether you can reproduce the issue by checking the logs for "Not scheduling read due to pending read or no permits." with `TestLogAppender`. This would be a true integration test that reproduces the actual problem. I would assume that Claude Code with Claude Opus 4.6 could achieve this just by providing the PR as context (instructed to fetch all PR details including comments) to create a representative test to reproduce the issue before creating a fix.

gosonzhang: ok

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
gosonzhang: Let me try a different approach based on the goals of the test cases.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ReplicatorTest.java
gosonzhang: I have reviewed your suggestion, but I still believe the current modification is more appropriate.

Although I have used tools to construct a getMockBookKeeper scenario to simulate exceptions, and kept the Pulsar dependencies unmocked, the changes to the ReplicatorTestBasefile are particularly extensive. From the perspective of addressing the issue, I think that absolutely avoiding mocking within Pulsar's logic deviates from the original intent of the test case.

This modification primarily addresses the issue where the PersistentReplicator.readEntriesFailed()implementation did not assign a value to its corresponding InFlightTask. This omission could affect the proper execution of geo-replication.

A Bookie exception is just one of many scenarios that could lead to the invocation of readEntriesFailed(), as long as any test case can simulate the execution of readEntriesFailed()and confirm that the modification is correct, that should be sufficient.

gosonzhang:
> > What suggestions do you have?
> > The initial idea was to keep the execution logic consistent before and after the test case modification. Personally, I also think that constructing a test API to execute the `waitForCursorRewindingRefCnf++` operation is not a good approach, but I haven't found a better one.
> 
> I don't know yet until I understand exactly what the test is currently doing and how the problem is triggered in production code. After understanding that it's easier to recommend option. I asked some questions in another comment.

The testReplicatorExpireMsgAsync() and testReplicatorClearBacklog() methods related to waitForCursorRewindingRefCnf have been modified.

lhotari:
It looks like the problem is handled by #25767 which was recently merged. I'll close this PR. Please reopen if it's a different issue that this addresses.

PR #13951, opened 2022-01-25
[OAuth2] Enable configurable preemptive token refresh in Java Client

michaeljmarshall:
### Motivation

This is a draft PR for a PIP I plan to submit. I still need to add tests and documentation for this feature.

The core feature that I am adding is the ability to refresh the OAuth2 token in the background without affecting the current token. This will improve stability for OAuth2 users, especially if/when their IDP is unavailable.

### Modifications

* Add a `ScheduledThreadPoolExecutor` to the `AuthenticationOAuth2` and schedule refresh token tasks to preemptively refresh a token.

### Verifying this change

I will add tests before removing the draft status of this PR.

### Does this pull request potentially affect one of the following parts:

This PR changes the default behavior of the `AuthenticationOAuth2` class and it also adds a new configuration value for end users. I plan to discuss the change on the mailing list.

### Documentation

- [x] `doc-required` 

I will add documentation once the feature design is solidified.

PR #25363, opened 2026-03-19
[improve][client] Enable configurable preemptive OAuth2 token refresh

lhotari:
Continuation of #13951 (originally authored by @michaeljmarshall). Rebased onto master, merge conflicts resolved, and all reviewer feedback addressed.

### Motivation

In some client use cases it is helpful to start refreshing the OAuth2 token in the background before it expires. This reduces latency spikes caused by blocking token-fetch calls and improves resilience when the Identity Provider is temporarily unavailable.

A key use case is reducing coupling to OAuth server availability. When the feature is enabled, the client continuously attempts to refresh the token in the background using exponential backoff, while continuing to serve requests with the existing valid token. This means a transient OAuth server outage does not immediately affect the Pulsar client — the client tolerates the outage for as long as the current token remains valid, which can be up to `(1 - earlyTokenRefreshPercent) * expires_in` seconds after the first refresh failure.

### Modifications

**Core feature (`AuthenticationOAuth2`)**
* Add optional preemptive background token refresh, scheduled at `earlyTokenRefreshPercent * expires_in` milliseconds after a successful fetch.
* Default value is `1` (100 %), which **disables** the feature and preserves existing behaviour.
* On refresh failure the scheduler retries with exponential backoff; a final synchronous attempt is made on the next `getAuthData()` call if the token has since expired.
* Replace the per-instance `ScheduledThreadPoolExecutor` (raised as a reviewer concern) with a **shared static `INTERNAL_SCHEDULER`** that uses `io.netty.util.concurrent.DefaultThreadFactory` (daemon threads, named `oauth2-token-refresher`) and allows the core thread to time out after 10 s of idle time — so no OS thread is held when no instances are actively refreshing.
* Callers may supply their own `ScheduledExecutorService`; this class never shuts it down.
* `close()` cancels the pending refresh task via the scheduler thread to avoid a race where a running refresh re-schedules itself; neither the shared nor the caller-supplied scheduler is shut down.
* Add `earlyTokenRefreshPercent` JSON configuration parameter, enabling the feature for broker-side / reflection-created instances. Accepts a **decimal** (`"0.8"` → 0.8) or an **integer percentage** (`"80"` → 0.8, `"100"` → 1.0 = disabled).

**`AuthenticationFactoryOAuth2`** (updated)
* Remove class-level `@Deprecated`; keep `@Deprecated` only on the individual `clientCredentials(...)` static methods.
* Extend `ClientCredentialsBuilder` with two new methods that cover the early refresh use case, making it the single builder for all client-credentials configurations:
  * `earlyTokenRefreshPercent(double)` — fractional value in (0, 1) to enable preemptive refresh, ≥ 1 to disable (default).
  * `scheduler(ScheduledExecutorService)` — optional external scheduler; if omitted the shared internal daemon scheduler is used automatically when early refresh is enabled.

**`AuthenticationOAuth2StandardAuthzServer`** (updated)
* Overrides `parseAuthParameters()` to inject the RFC 8414 well-known metadata path, keeping the `configure()` refactor transparent to subclasses.

### Verifying this change

New and updated unit tests in `AuthenticationOAuth2Test`:
* Token caching and expiry-driven refresh (no early refresh).
* Early refresh schedules background calls before expiry.
* External scheduler is used when provided and is **not** shut down on `close()`.
* `configure()` via default constructor: decimal and integer `earlyTokenRefreshPercent` values, disabled (`100`), zero and non-numeric values throw `IllegalArgumentException`.
* `parseEarlyRefreshPercent` unit tests for all branches.
* Background scheduler actually triggers `authenticate()` when enabled via `configure()`.

### Does this pull request potentially affect one of the following parts?

- **Default behaviour is unchanged** — `earlyTokenRefreshPercent` defaults to `1` (disabled).
- Adds new configuration surface for OAuth2 users wanting proactive token management.
- Thread-resource concern (one thread per client) addressed by shared daemon executor that releases the thread when idle.

### Documentation

- [x] `doc` — `earlyTokenRefreshPercent` parameter and early-refresh behaviour documented in javadoc on `AuthenticationOAuth2` and `ClientCredentialsBuilder`.

Closes #13951


merlimat (approved):

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2.java
```diff
@@ -96,21 +228,111 @@ public void start() throws PulsarClientException {
...
+    private void handleSuccessfulTokenRefresh() {
+        if (scheduler != null) {
+            scheduler.execute(() -> {
+                if (earlyTokenRefreshPercent < 1) {
+                    backoff = buildBackoff(cachedToken.latest.getExpiresIn());
+                    long expiresInMillis = TimeUnit.SECONDS.toMillis(cachedToken.latest.getExpiresIn());
+                    scheduleRefresh((long) (expiresInMillis * earlyTokenRefreshPercent));
+                }
+            });
+        }
```
nodece:
```suggestion
        if (scheduler != null && earlyTokenRefreshPercent < 1) {
            scheduler.execute(() -> {
                  backoff = buildBackoff(cachedToken.latest.getExpiresIn());
                  long expiresInMillis = TimeUnit.SECONDS.toMillis(cachedToken.latest.getExpiresIn());
                  scheduleRefresh((long) (expiresInMillis * earlyTokenRefreshPercent));
            });
        }
```

nodece (approved):

PR #25367, opened 2026-03-20
[improve][common] Optimize TopicName.get() to reduce lock contention on cache lookup

liangyepianzhou:
### Motivation

`TopicName.get()` previously used `ConcurrentHashMap.computeIfAbsent()` to populate the topic-name cache. Although `computeIfAbsent` is atomic, it holds the internal bin-lock for the entire duration of the mapping function, which includes the non-trivial `TopicName` construction (string splitting, validation, etc.).

Under high-concurrency workloads where many threads simultaneously encounter the same uncached topic name, this causes unnecessary lock contention and can degrade throughput.

### Modifications

Replace `computeIfAbsent` with an explicit two-step pattern:

1. **Fast path**: call `cache.get(topic)` first — a single volatile read with no locking — and return immediately on a cache hit (steady-state case).
2. **Slow path** (cache miss): construct `TopicName` *outside* the lock, then use `cache.putIfAbsent()` to insert. If two threads race on the same key, one wins the `putIfAbsent` and the other's instance is discarded; this is safe because `TopicName` is immutable.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository:


liangyepianzhou:
image

**Background**

While load-testing Pulsar with a single topic containing **1,000,000 partitions**, we observed that `TopicName.get()` was consuming a disproportionate amount of CPU, showing up prominently in flame graphs.

**Root Cause Analysis**

Two compounding issues were identified:

1. **`topicNameCacheMaxCapacity` too small** — With 1M partitions, the default cache capacity is inevitably exceeded, triggering repeated `clearIfReachedMaxCapacity` calls and causing **cache stampedes** where all partitions miss simultaneously.

2. **`new TopicName()` executed inside the lock** — The current implementation uses `computeIfAbsent(key, TopicName::new)`, which holds the `ConcurrentHashMap` bin-lock for the entire duration of object construction. Under concurrent cache misses, threads serialize on the same lock, degrading throughput significantly.

**Fix**

Replace `computeIfAbsent` with a lock-free `get` + out-of-lock construction + `putIfAbsent` pattern:

```java
// Before
return cache.computeIfAbsent(topic, TopicName::new);

// After
TopicName cached = cache.get(topic);
if (cached != null) return cached;
TopicName created = new TopicName(topic);          // constructed outside lock
TopicName existing = cache.putIfAbsent(topic, created);
return existing != null ? existing : created;
```

Under concurrent misses on the same key, each thread constructs its own instance independently; `putIfAbsent` elects the winner and the losers are simply GC'd — eliminating bin-lock contention entirely.

**Benchmark Results (JMH · SingleShotTime · 8 threads · 1M partitions · full cold cache)**

| Benchmark | Avg (ms) | Median (ms) | p99 (ms) | Min (ms) |
|-----------|----------|-------------|----------|----------|
| `miss_A` — `computeIfAbsent` | 427.1 | 395.3 | 644.5 | 316.5 |
| `miss_B` — `get + putIfAbsent` | **287.1** | **278.9** | **368.0** | **236.1** |
| **Improvement** | **~1.49×** | **~1.42×** | **~1.75×** | **~1.34×** |

The tail latency improvement (**1.75×** at p99) is especially significant: `computeIfAbsent` causes severe jitter under lock contention (316–644 ms range), while `get+putIfAbsent` stays stable (236–368 ms).

image


- Raw results: [miss_B_1M_result.json](https://github.com/user-attachments/files/26133809/miss_B_1M_result.json)
- Benchmark source:  [TopicNameCacheBenchmark.java](https://github.com/user-attachments/files/26133887/TopicNameCacheBenchmark.java)

merlimat (commented):
Change looks good. 

can you try to compare to just call `cache .put()` instead of `putIfAbsent()`?

It would be good if you could add the jhm code to the microbenchmarks module

lhotari:
Related PR from the past #24457 (change was rejected so I closed it) with a lot of interesting comments.
One of the comments came from Caffeine author: https://github.com/apache/pulsar/pull/24457#issuecomment-3007063192.
Besides performance another detail to consider is the duplication of java.lang.String instances in memory.

lhotari (commented):
This doesn't address one of the key problems which is that the cache is cleared if the size exceeds the max size setting (`topicNameCacheMaxCapacity`):

https://github.com/apache/pulsar/blob/2e3e78d992e873e358d16fa875f79cb607019284/pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java#L54-L62

We'd be better off by switching to use Caffeine again. The Caffeine author commented in https://github.com/apache/pulsar/pull/24457#issuecomment-3007063192 that the bottleneck in the earlier version has been addressed. Since we have upgraded to Java 17 for the client, we can use the newer Caffeine version.

lhotari:
The flamegraph in https://github.com/apache/pulsar/pull/25367#issuecomment-4095828178 looks like it was from an old version of Pulsar that uses Guava Cache (before #23052 <3.0.6, <3.3.1).
@liangyepianzhou What version of Pulsar are you load testing?

lhotari:
> The tail latency improvement (**1.75×** at p99) is especially significant: `computeIfAbsent` causes severe jitter under lock contention (316–644 ms range), while `get+putIfAbsent` stays stable (236–368 ms).

The benchmark seems to measure the miss of 1M cache entries. That's not a very realistic scenario that there would be such amount of misses at once.

I think it would be more useful to have a solution that ensures that the memory use of the cache is bounded.
That is something that I have added quite recently to AbstractMetadataStore children cache:
https://github.com/apache/pulsar/blob/6e577f0bc5a7c3184fdb75d6afdd54a56578aa26/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/AbstractMetadataStore.java#L123-L141
in https://github.com/apache/pulsar/pull/24868
The max size is limited to 20% of max heapsize.

For TopicName cache there could be a bytesize limit too. The StringInterner solution would be useful at least for namespace and tenant Strings to ensure that those don't cause heap duplication if the namespace or tenant caches expire or overflow.

lhotari:
> * Benchmark source:  [TopicNameCacheBenchmark.java](https://github.com/user-attachments/files/26133887/TopicNameCacheBenchmark.java)

The benchmark doesn't use actual code which would be part of the Pulsar code base.
I believe that this is a more representative way: https://github.com/lhotari/pulsar/blob/lh-fix-topicname-memory-leak/microbench/src/main/java/org/apache/pulsar/common/naming/TopicNameBenchmark.java
That was done in #25367 which wasn't merged.

liangyepianzhou:
> The flamegraph in [#25367 (comment)](https://github.com/apache/pulsar/pull/25367#issuecomment-4095828178) looks like it was from an old version of Pulsar that uses Guava Cache (before #23052 <3.0.6, <3.3.1). @liangyepianzhou What version of Pulsar are you load testing?

The **flamegraph** comes from 3.0.5, I change version to 3.0.13 and make `topicNameCacheMaxCapacity ` = -1. CPU usage dropped from 90% to 9%, but the percentage of `TopicName.get()` in the flame graph is still very high.

>The benchmark seems to measure the miss of 1M cache entries. That's not a very realistic scenario that there would be such amount of misses at once.

In my scenario, one topic has 1 million partitions, so there are 1 million cache misses during the initial startup. Topic loading is currently very slow, and I suspect this may be one of the reasons.

liangyepianzhou:
> The flamegraph in [#25367 (comment)](https://github.com/apache/pulsar/pull/25367#issuecomment-4095828178) looks like it was from an old version of Pulsar that uses Guava Cache (before #23052 <3.0.6, <3.3.1). @liangyepianzhou What version of Pulsar are you load testing?

Thanks for pointing out the historical context and Ben Manes' insights!

Just to clarify the scope of this PR: the `get + putIfAbsent` change is intentionally a minimal, targeted fix to reduce bin-lock contention during cache misses, independent of the broader cache design questions (capacity bounding, Caffeine version, soft references, etc.).

That said, I'd like to confirm whether this small optimization is still worthwhile on its own:

 1. Even with a properly bounded cache, concurrent cache misses (e.g. during a cold start or after a cache clear) will still happen. In those cases, `putIfAbsent` avoids holding the bin-lock during `TopicName` construction, which seems beneficial regardless of the eviction strategy.
 2. The fast path (`cache.get()` first) is a strict improvement for the steady-state hit case with zero synchronization overhead.
 
So the question is: **do you see any reason this low-risk change should be blocked on the larger cache redesign?** Happy to fold it into a bigger PR if that's preferred, but wanted to check if it can stand alone first.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
```diff
@@ -77,12 +77,32 @@ public static TopicName get(String domain, String tenant, String namespace, Stri
...
+     * while constructing a new TopicName object. The construction (string splitting / parsing) is
+     * pure CPU work and can be done outside the lock. In the typical steady-state (cache hit) this
+     * method does a single volatile read via {@code get()} and returns immediately with no
+     * synchronization overhead.
+     *
+     * <p>In the cache-miss case, two threads racing on the same key may both construct a
+     * {@code TopicName} instance, but only one wins the {@code putIfAbsent} and the loser's
+     * instance is simply discarded. This is safe because {@code TopicName} is immutable and
+     * construction is cheap compared to the lock-contention / context-switch cost of
+     * {@code computeIfAbsent}.
```
lhotari: these details could be in code comments since the public API user shouldn't need to know about the implementation details.

lhotari (approved):
LGTM, great solution @liangyepianzhou. Just one comment about documenting implementation details in javadoc.

pulsar-common/src/main/java/org/apache/pulsar/common/naming/TopicName.java
liangyepianzhou: Good suggestion. fixed

liangyepianzhou:
> Change looks good.
> 
> can you try to compare to just call `cache .put()` instead of `putIfAbsent()`?
> 
> It would be good if you could add the jhm code to the microbenchmarks module

Thanks for the suggestion. Fixed.

BewareMyPower:
I think we should not depend much from the cache itself. The `TopicName` construction is low-efficient and used nearly everywhere. Could you try my previous patch here (https://github.com/apache/pulsar/pull/24463) to see if there is any improvement? If so, I can resolve the conflicts again

lhotari:
> I think we should not depend much from the cache itself. The `TopicName` construction is low-efficient and used nearly everywhere. Could you try my previous patch here (#24463) to see if there is any improvement? If so, I can resolve the conflicts again

@BewareMyPower Could you rebase #24463? I think we can handle that optimization separately.

On the caching: one benefit worth preserving is that the TopicName and NamespaceName caches reduce java.lang.String instance duplication, which is valuable beyond just avoiding the construction cost itself. So I'd lean toward keeping the caches rather than removing them.

liangyepianzhou:
> I think we should not depend much from the cache itself. The `TopicName` construction is low-efficient and used nearly everywhere. Could you try my previous patch here (#24463) to see if there is any improvement? If so, I can resolve the conflicts again

 @BewareMyPower  After applying your optimization, the benchmark results are as follows: jmh2 uses `computeIfAbsent`, jmh1 uses `put`.

Benchmark                           Mode  Cnt  Score   Error  Units
TopicNameGetBenchmark.coldStartGet    ss   50  5.600 ± 1.722  us/op
[1]+  Done                    java -jar ./microbench-*-benchmarks.jar TopicNameGetBenchmark 2>&1 > jmh1.log

Benchmark                           Mode  Cnt  Score   Error  Units
TopicNameGetBenchmark.coldStartGet    ss   50  6.446 ± 2.168  us/op
[1]+  Done                    java -jar ./microbench-*-benchmarks.jar TopicNameGetBenchmark 2>&1 > jmh2.log

BewareMyPower:
so the result is now with this PR (replace `computeIfAbsent` with `put`), it's
- 1.5x of the master branch (427 / 287)
- 0.87x of the master branch + my improvement in #24463 (5.600 / 6.446)

~~Then I believe both PRs are valuable, I will rebase my PR soon, for this PR, just go ahead to fix the CI~~

BewareMyPower:
I'm rebasing my PR currently. But it's a bit confusing from your test result. It seems that switching to `put` would be slower with my improvement. Is there anything wrong I understood?

liangyepianzhou:
> I'm rebasing my PR currently. But it's a bit confusing from your test result. It seems that switching to `put` would be slower with my improvement. Is there anything wrong I understood?

 Are you testing the cold start / cache-miss scenario?

My understanding is that the main latency comes from the cache-miss startup phase, where the optimization is most noticeable. You can use my JMH program to test it.

BewareMyPower:
No. I didn't run any test for now. I'm just analyzing your test report here: https://github.com/apache/pulsar/pull/25367#issuecomment-4286350318

> mh2 uses computeIfAbsent, jmh1 uses put.

- jmh1: 5.600 ± 1.722 us/op
- jmh2: 6.446 ± 2.168 us/op

**Oh my bad, I misunderstood the unit (`us/op`). I've thought it's ops/us**

BewareMyPower (approved):

PR #25386, opened 2026-03-23
[fix][client] Fixed stuck issues caused by an exception encountered in the receiveIndividualMessagesFromBatch()

gosonzhang:
### Motivation

Fixed three issues with `receiveIndividualMessagesFromBatch()` parsing failures:

1. On the consumer side, corrupted  batch messages that fail to parse based on formats formed by [a] or other conditions should be discarded to avoid consumption stucks.

2. When parsing a batch of messages and an exception is thrown in the middle, the permits (minus 1) of the remaining unprocessed messages are all leaked, causing consumption stucks.

3. When batch index ack is enabled, messages being processed will be confirmed in advance in the event of a parsing exception.

[a]. https://github.com/apache/pulsar/pull/24061

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

[none](https://github.com/gosonzhang/pulsar/pull/3)

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1786,6 +1786,7 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
 
         SingleMessageMetadata singleMessageMetadata = new SingleMessageMetadata();
         int skippedMessages = 0;
+        int processedMessages = 0;
         try {
             for (int i = 0; i < batchSize; ++i) {
                 final MessageImpl<T> message = newSingleMessage(i, batchSize, brokerEntryMetadata, msgMetadata,
@@ -1815,13 +1816,14 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
                     continue;
                 }
                 executeNotifyCallback(message);
+                processedMessages++;
             }
             if (ackBitSet != null) {
                 ackBitSet.recycle();
             }
-        } catch (IllegalStateException e) {
-            log.warn("[{}] [{}] unable to obtain message in batch", subscription, consumerName, e);
-            discardCorruptedMessage(messageId, cnx, ValidationError.BatchDeSerializeError);
+        } catch (IllegalStateException | IllegalArgumentException e) {
+            discardCorruptedBatchMessage(messageId, cnx,
+                    (batchSize - skippedMessages - processedMessages), ValidationError.BatchDeSerializeError);
         }
 
         if (deadLetterPolicy != null && possibleSendToDeadLetterTopicMessages != null) {
@@ -2151,6 +2153,14 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
         log.error("[{}][{}] Discarding corrupted message at {}:{}", topic, subscription, messageId.getLedgerId(),
                 messageId.getEntryId());
         discardMessage(messageId, currentCnx, validationError, 1);
+    }
+
+    private void discardCorruptedBatchMessage(MessageIdData messageId, ClientCnx currentCnx,
+                                              int unreadMessages, ValidationError validationError) {
+        log.error("[{}] [{}] Discarding corrupted batch message at {}:{}, unread count={}, exception={}",
+                subscription, consumerName, messageId.getLedgerId(), messageId.getEntryId(),
+                unreadMessages, validationError);
+        discardMessage(messageId, currentCnx, validationError, unreadMessages);
     }
 
     private void discardMessage(MessageIdData messageId, ClientCnx currentCnx, ValidationError validationError,
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
@@ -1822,8 +1822,9 @@ public class ConsumerImpl<T> extends ConsumerBase<T> implements ConnectionHandle
                 ackBitSet.recycle();
             }
         } catch (IllegalStateException | IllegalArgumentException e) {
-            discardCorruptedBatchMessage(messageId, cnx,
-                    (batchSize - skippedMessages - processedMessages), ValidationError.BatchDeSerializeError);
+            // For IllegalArgumentException see PR: https://github.com/apache/pulsar/pull/24061
+            discardCorruptedBatchMessage(messageId, cnx, batchSize,
+                    skippedMessages, processedMessages, ValidationError.BatchDeSerializeError);
         }
 
         if (deadLetterPolicy != null && possibleSendToDeadLetterTopicMessages != null) {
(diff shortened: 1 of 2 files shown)


pulsar-client/src/test/java/org/apache/pulsar/client/impl/BatchMessageDecodeFailureTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+        consumer.receiveIndividualMessagesFromBatch(brokerEntryMetadata, metadata, 0, null,
+                batchBuffer, new MessageIdData().setLedgerId(2000).setEntryId(2),
+                mockCnx, DEFAULT_CONSUMER_EPOCH, false);
+
+        // The first valid message should be enqueued via executeNotifyCallback,
+        // which uses internalPinnedExecutor.execute(). Since it's async,
+        // we check the stats synchronously — discardCorruptedBatchMessage
+        // is called inline in the catch block.
+        verify(statsRecorder, times(1)).incrementNumReceiveFailed();
```
lhotari: since the execution is async, Mockito's `timeout` should be used. `timeout(2000).times(1)` or just `timeout(2000)`.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/BatchMessageDecodeFailureTest.java
```diff
@@ -0,0 +1,220 @@
...
+        consumer.receiveIndividualMessagesFromBatch(brokerEntryMetadata, metadata, 0, null,
+                truncatedBuffer, new MessageIdData().setLedgerId(4000).setEntryId(4),
+                mockCnx, DEFAULT_CONSUMER_EPOCH, false);
+
+        // No messages should be enqueued
+        assertEquals(consumer.numMessagesInQueue(), 0,
+                "Truncated payload should not produce any messages in queue");
+
+        // The receive-failed counter should have been incremented
+        verify(statsRecorder, times(1)).incrementNumReceiveFailed();
```
lhotari: use `timeout(2000)`

pulsar-client/src/test/java/org/apache/pulsar/client/impl/BatchMessageDecodeFailureTest.java
```diff
@@ -0,0 +1,220 @@
...
+        consumer.receiveIndividualMessagesFromBatch(brokerEntryMetadata, metadata, 0, null,
+                corruptedPayload, new MessageIdData().setLedgerId(1000).setEntryId(1),
+                mockCnx, DEFAULT_CONSUMER_EPOCH, false);
+
+        // No valid messages should be enqueued
+        assertEquals(consumer.numMessagesInQueue(), 0,
+                "Corrupted batch should not produce any messages in queue");
+
+        // Verify discardCorruptedBatchMessage incremented the failed counter
+        verify(statsRecorder, times(1)).incrementNumReceiveFailed();
```
lhotari: use `timeout(2000)`

pulsar-client/src/test/java/org/apache/pulsar/client/impl/BatchMessageDecodeFailureTest.java
gosonzhang: ok

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1815,13 +1816,14 @@ void receiveIndividualMessagesFromBatch(BrokerEntryMetadata brokerEntryMetadata,
...
                 executeNotifyCallback(message);
+                processedMessages++;
             }
             if (ackBitSet != null) {
                 ackBitSet.recycle();
             }
-        } catch (IllegalStateException e) {
-            log.warn("[{}] [{}] unable to obtain message in batch", subscription, consumerName, e);
-            discardCorruptedMessage(messageId, cnx, ValidationError.BatchDeSerializeError);
+        } catch (IllegalStateException | IllegalArgumentException e) {
```
lhotari, line 1824: What scenario cause `IllegalArgumentException`? It could be helpful to add a comment in the test case.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1815,13 +1816,14 @@ void receiveIndividualMessagesFromBatch(BrokerEntryMetadata brokerEntryMetadata,
...
                 }
                 executeNotifyCallback(message);
+                processedMessages++;
             }
             if (ackBitSet != null) {
                 ackBitSet.recycle();
             }
-        } catch (IllegalStateException e) {
-            log.warn("[{}] [{}] unable to obtain message in batch", subscription, consumerName, e);
-            discardCorruptedMessage(messageId, cnx, ValidationError.BatchDeSerializeError);
```
lhotari, line 1824: this method comes dead code after this change. it's worth removing if it's the case.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1815,13 +1816,14 @@ void receiveIndividualMessagesFromBatch(BrokerEntryMetadata brokerEntryMetadata,
...
+        } catch (IllegalStateException | IllegalArgumentException e) {
+            discardCorruptedBatchMessage(messageId, cnx,
+                    (batchSize - skippedMessages - processedMessages), ValidationError.BatchDeSerializeError);
```
lhotari: There's also another existing bug: the acknowledgements of messages already in processing are getting ignored, assuming that batch index acknowledgement is used. It might be useful to also fix that issue while fixing problems in this area. 
I'd assume that the fix would be to pass an ack set where the remaining indices are acknowledged (when batch index ack is enabled).

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
gosonzhang: It needs to be removed

In the original code, a similar error would be printed whenever the `discardCorruptedMessage` function is entered. So, merging them and keeping only one log output is sufficient.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
gosonzhang: Yes, that's right, let me see

gosonzhang:
@lhotari Could you please take another look? I've made the modifications according to your suggestions.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -2153,6 +2156,38 @@ private void discardCorruptedMessage(MessageIdData messageId, ClientCnx currentC
...
+                ackBitSet.clear(i);
+            }
+        }
+        ByteBuf cmd = Commands.newAck(consumerId, messageId.getLedgerId(), messageId.getEntryId(),
+                ackBitSet, AckType.Individual, validationError, Collections.emptyMap(), -1);
+        currentCnx.ctx().writeAndFlush(cmd, currentCnx.ctx().voidPromise());
+        if (ackBitSet != null) {
+            ackBitSet.recycle();
+        }
+        increaseAvailablePermits(currentCnx, batchSize - corruptedStartIndex);
```
Denovo1998, line 2187: It is not advisable to directly use `batchSize - (skipped + processed)` to replenish permits. The `receiveIndividualMessagesFromBatch()` method will filter out already-acknowledged batch indices at the beginning based on the inbound ackSet; these indices have not consumed broker permits. If this is a partial batch redelivery, the current algorithm would include the undelivered indices, potentially leading to permit over-credit.

It might be better to pass in the current inbound ackBitSet, count only the bits after the failing index that are still deliverable, and construct a new ack set based on that.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
```diff
@@ -1815,13 +1816,15 @@ void receiveIndividualMessagesFromBatch(BrokerEntryMetadata brokerEntryMetadata,
...
                 executeNotifyCallback(message);
+                processedMessages++;
             }
             if (ackBitSet != null) {
                 ackBitSet.recycle();
             }
-        } catch (IllegalStateException e) {
-            log.warn("[{}] [{}] unable to obtain message in batch", subscription, consumerName, e);
-            discardCorruptedMessage(messageId, cnx, ValidationError.BatchDeSerializeError);
+        } catch (IllegalStateException | IllegalArgumentException e) {
```
Denovo1998, line 1824: Commands.deSerializeSingleMessageInBatch() may also throw IndexOutOfBoundsException for truncated payload (triggered by readUnsignedInt() or retainedSlice()), whereas newSingleMessage() currently does not wrap it into IllegalStateException. As a result, encountering a corrupted batch with truncation or length mismatch can still bypass discardCorruptedBatchMessage, and permit leak or stuck issues would persist.

Denovo1998 (changes requested):

PR #25398, opened 2026-03-24
[cleanup] PIP-463: Migrate build system from Maven to Gradle

merlimat:
## Summary

Complete migration of Apache Pulsar's build system from Maven to Gradle, as proposed in [PIP-463](https://github.com/apache/pulsar/blob/master/pip/pip-463.md).

### What this PR does:
- Adds Gradle build scripts (`build.gradle.kts`, `settings.gradle.kts`, `gradle/libs.versions.toml`) for all modules
- Converts all CI workflows from Maven to Gradle
- Removes all Maven files (`pom.xml`, `mvnw`, `.mvn/`)
- Adds enforced platform module (`pulsar-dependencies`) for consistent dependency versions
- Adds binary license check and Trivy container scan to CI

### Verified equivalence with Maven:
- All CI test suites passing (unit, integration, system, shade tests on Java 17/21/24)
- Server/shell distribution tarballs produce the same JARs
- Shaded client JARs contain the same relocated classes
- Docker images build and work correctly
- NAR connector packages are identical

## Test plan
- [x] All unit tests passing
- [x] All integration tests passing  
- [x] All system tests passing (IO, Functions, Schema, Tiered Storage, Connectors)
- [x] Shade tests on Java 17/21/24
- [x] Binary license check passing
- [x] Docker images building correctly
- [x] Server distribution tarball verified against Maven output

--------

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/distribution/shell/build.gradle.kts
+++ b/distribution/shell/build.gradle.kts
@@ -34,7 +34,6 @@ val distLib by configurations.creating {
     exclude(group = "net.java.dev.jna", module = "jna")
     exclude(group = "net.java.dev.jna", module = "jna-platform")
     exclude(group = "io.netty", module = "netty-transport-native-kqueue")
-    exclude(group = "com.github.ben-manes.caffeine", module = "caffeine")
     exclude(group = "io.prometheus", module = "simpleclient_caffeine")
 }
 dependencies {


lhotari:
Please update the build instructions in [README.md](https://github.com/merlimat/pulsar/blob/gradle/README.md#build)

.github/workflows/ci-gradle-cache-update.yaml
lhotari, line 1: this file could be renamed to ci-gradle-cache-update.yaml

.github/workflows/codeql.yaml
```diff
@@ -72,15 +61,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
```
lhotari: pin to hash

.github/workflows/pulsar-ci-flaky.yaml
```diff
@@ -197,34 +196,25 @@ jobs:
...
         uses: actions/setup-java@v4
         with:
           distribution: ${{ env.JDK_DISTRIBUTION }}
           java-version: ${{ env.CI_JDK_MAJOR_VERSION }}
 
-      - name: Build core-modules
-        run: |
-          mvn -B -T 1C -ntp -Pcore-modules,-main clean install -DskipTests -Dlicense.skip=true -Drat.skip=true
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
```
lhotari: pin to hash

.github/workflows/pulsar-ci.yaml
```diff
@@ -422,15 +348,18 @@ jobs:
...
-      - name: Upload Surefire reports
+      - name: Upload test reports
         uses: actions/upload-artifact@v4
         if: ${{ !success() || env.TRACE_TEST_RESOURCE_CLEANUP != 'off' }}
         with:
-          name: Unit-${{ matrix.group }}-surefire-reports
-          path: surefire-reports
+          name: Unit-${{ matrix.group }}-test-reports
+          path: |
+            surefire-reports
```
lhotari: remove surefire-reports

merlimat:
> Please update the build instructions in [README.md](https://github.com/merlimat/pulsar/blob/gradle/README.md#build)

Yes, there are several followup cleanup tasks. One other is the `bin/pulsar` and other scripts that build the classpath. I'll address each separately.

distribution/shell/build.gradle.kts
```diff
@@ -0,0 +1,274 @@
...
+                    if (mappedName.startsWith("bouncy-castle-bc-")) {
+                        mappedName = mappedName + "-pkg"
+                    }
+                    "${mappedName}.${file.extension}"
+                }
+                else -> file.name
+            }
+            file.name to newName
+        }
+    }
```
lhotari: This doesn't look like the correct approach. Projects where the Maven artifact name differs from the directory name should be configured directly in the build, rather than relying on external mapping.

For example, the artifact name for `pulsar-client` should be set to
`pulsar-client-original` at the source, so no separate mapping is needed.

Rename the project in settings.gradle.kts:

// Option 1: Include then rename
include(":pulsar-client")
project(":pulsar-client").name = "pulsar-client-original"

// Option 2: Include with the desired name, point to the existing directory
include(":pulsar-client-original")
project(":pulsar-client-original").projectDir = file("pulsar-client")

With either approach, the project name becomes "pulsar-client-original",
so the default artifactId and JAR filename will match automatically.

Note that project references use the project name, not the directory name.
After renaming, other subprojects would reference it as:

dependencies {
  implementation(project(":pulsar-client-original"))
}

The approach makes the maven artifactId correct, `pulsar-client-original`. 

There's also an option to keep the project name and directory name as pulsar-client and configure base archivesName to `pulsar-client-original` and set MavenPublication's `artifactId` to `pulsar-client-original`, but I wouldn't recommend that approach.

distribution/shell/build.gradle.kts
```diff
@@ -0,0 +1,274 @@
...
+            "com.google.errorprone:error_prone_annotations" -> useVersion(errorproneVersion)
+            "com.google.j2objc:j2objc-annotations" -> useVersion(j2objcAnnotationsVersion)
+            "io.netty:netty-tcnative-boringssl-static",
+            "io.netty:netty-tcnative-classes" -> useVersion(nettyTcnativeVersion)
+            "org.yaml:snakeyaml" -> useVersion(snakeyamlVersion)
+            "org.javassist:javassist" -> useVersion(javassistVersion)
+            "org.bouncycastle:bcprov-jdk18on" -> useVersion(bouncycastleBcprovVersion)
+            "org.jline:jline" -> useVersion(jline3Version)
+        }
+    }
```
lhotari: I think that these dependencies should be in the pulsar-dependencies platform that is enforced in all projects, including the distribution project modules.

managed-ledger/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+    implementation(libs.commons.lang3)
+
+    testImplementation(project(":testmocks"))
+    testImplementation(libs.dropwizardmetrics.core)
+    testImplementation(libs.snappy.java)
+    testImplementation(libs.awaitility)
+    testImplementation(libs.opentelemetry.sdk.testing)
+    testImplementation(libs.opentelemetry.sdk.extension.autoconfigure)
+    testImplementation(libs.zookeeper) { artifact { classifier = "tests" } }
+    testImplementation(files(rootProject.tasks.named("filteredBkServerTestJar").map { (it as Jar).archiveFile }))
```
lhotari: This is a hack. It reaches across project boundaries to grab a task output by name, bypassing Gradle's dependency model. Problems:

1. No automatic task dependency — Gradle doesn't know this input comes from that task, so it may not be built yet (unless wired manually)
2. Breaks project isolation — directly referencing rootProject.tasks couples projects tightly
3. Opaque to Gradle — the dependency isn't visible to build caching, up-to-date checks, or dependency resolution

The recommended Gradle approach is to use cross-project artifact sharing via configurations. The producing project would expose the JAR as an outgoing variant:

// In the project that produces the filtered JAR                                                                                                                                                                                                                                                                                                                         
val filteredTestJar by configurations.creating {
isCanBeConsumed = true
isCanBeResolved = false                                                                                                                                                                                                                                                                                                                                              
}

artifacts {                                                                                                                                                                                                                                                                                                                                                            
add("filteredTestJar", tasks.named<Jar>("filteredBkServerTestJar"))
}

Then the consuming project depends on it normally:

// In managed-ledger
dependencies {
testImplementation(project(path = ":", configuration = "filteredTestJar"))                                                                                                                                                                                                                                                                                           
}

distribution/io/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+
+    // Collect NAR files from each connector project's build/libs directory
+    connectorNarProjects.forEach { projectPath ->
+        val narDir = project(projectPath).layout.buildDirectory.dir("libs")
+        from(narDir) {
+            into(".")
+            include("*.nar")
+        }
+        dependsOn("${projectPath}:nar")
+    }
```
lhotari: The recommended approach is to resolve NARs through dependency resolution, just like JARs. If the NAR plugin publishes an outgoing variant (or artifact), the consuming project should depend on it as a normal dependency and let Gradle resolve the files:

val connectorNars by configurations.creating {                                                                                                                                                                                                                                                                                                                           
isCanBeResolved = true
isCanBeConsumed = false
// Request NAR artifacts specifically                                                                                                                                                                                                                                                                                                                                
attributes {
attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "nar")                                                                                                                                                                                                                                                                                                 
}           
}

dependencies {
connectorNarProjects.forEach { projectPath ->
connectorNars(project(projectPath))                                                                                                                                                                                                                                                                                                                              
}
}

// Then in the task:
from(connectorNars) {                                                                                                                                                                                                                                                                                                                                                    
into(".")   
}

This way Gradle handles task ordering, caching, and parallel safety automatically — no dependsOn or directory path assumptions needed.

distribution/offloaders/build.gradle.kts
```diff
@@ -0,0 +1,64 @@
...
+    from(jcloudNarFiles) {
+        into("${baseDir}/offloaders")
+        include("*.nar")
+    }
+    from(fileSystemNarFiles) {
+        into("${baseDir}/offloaders")
+        include("*.nar")
+    }
+    dependsOn(":tiered-storage:tiered-storage-jcloud:nar")
+    dependsOn(":tiered-storage:tiered-storage-file-system:nar")
```
lhotari: similar solution as mentioned in comment to pulsar-io-distribution build file about nars should be used so that Gradle's features would work properly.

distribution/server/build.gradle.kts
```diff
@@ -0,0 +1,341 @@
...
+
+val serverDistTar by tasks.registering(Tar::class) {
+    // Map of Gradle project names to Maven artifact names for renaming.
+    // Only exact matches are applied (the name after the key must start with a version digit).
+    val projectNameMapping = mapOf(
+        "pulsar-client" to "pulsar-client-original",
+        "pulsar-client-admin" to "pulsar-client-admin-original",
+        "pulsar-functions-local-runner" to "pulsar-functions-local-runner-original",
+        "bc" to "bouncy-castle-bc",
+    )
```
lhotari: similar comment as om shell distribution's build file to remove the need for any mapping like this.

distribution/server/build.gradle.kts
```diff
@@ -0,0 +1,341 @@
...
+        exclude("**/animal-sniffer-annotations-*.jar")
+        // Exclude .nar files (BookKeeper NAR artifacts — we add .jar variants explicitly)
+        exclude("**/*.nar")
+        // Exclude the non-classified netty-transport-native-epoll jar
+        // (only the platform-classified variants like -linux-x86_64 belong in the dist)
+        exclude {
+            val n = it.file.name
+            n.startsWith("netty-transport-native-epoll-") && n.endsWith(".jar") &&
+                !n.contains("-linux-") && !n.contains("-osx-") && !n.contains("-windows-")
+        }
```
lhotari: exclusions should be handled in the `distList` configuration

distribution/server/build.gradle.kts
```diff
@@ -0,0 +1,341 @@
...
+                    if (mappedName.startsWith("bouncy-castle-bc-")) {
+                        mappedName = mappedName + "-pkg"
+                    }
+                    "org.apache.pulsar-${mappedName}.${ext}"
+                }
+                else -> file.name
+            }
+            file.name to newName
+        }
+    }
```
lhotari, line 270: the pulsar-client -> pulsar-client-original type of mappings become obsolete after setting the project names as explained in shell distribution's build file review comment.

distribution/server/build.gradle.kts
```diff
@@ -0,0 +1,341 @@
...
+    from(project(":pulsar-functions:pulsar-functions-runtime-all").tasks.named("jar")) {
+        into("${baseDir}/instances")
+        rename(".*", "java-instance.jar")
+    }
+
+    // Java examples JAR
+    from(project(":pulsar-functions:pulsar-functions-api-examples").tasks.named("jar")) {
+        into("${baseDir}/examples")
+        rename(".*", "api-examples.jar")
+    }
```
lhotari, line 302: This could be improve, but it might not be a show stopper for Gradle's configuration cache. A better approach would be to use a configuration and then map the file name here to the desired one.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.0-bin.zip
```
lhotari: 9.4.1 is latest

pulsar-bom/pom.xml
```diff
@@ -1,685 +0,0 @@
...
-  <packaging>pom</packaging>
-  <parent>
-    <groupId>org.apache</groupId>
-    <artifactId>apache</artifactId>
-    <version>35</version>
-    <relativePath/>
-  </parent>
-
-  <groupId>org.apache.pulsar</groupId>
-  <artifactId>pulsar-bom</artifactId>
```
lhotari, line 35: we continue to need pulsar-bom. Gradle can also publish BOMs.

pulsar-bom can avoid issues for example where pulsar-client-api and pulsar-client / pulsar-client-original are not aligned. That issue came up once in a while in the past.

We recommend using pulsar-bom in the Java client docs: https://pulsar.apache.org/docs/4.1.x/client-libraries-java-setup/#pulsar-bom-maven

pulsar-build/run_unit_group_gradle.sh
```diff
@@ -0,0 +1,254 @@
...
+    if [ $use_fail_fast -eq 1 ]; then
+      failfast_args="-DtestFailFast=true"
+    else
+      failfast_args="-DtestFailFast=false"
+      # When fail-fast is off, use --continue so Gradle runs all test tasks
+      # even if one module's tests fail (similar to Maven's --fail-at-end).
+      continue_args="--continue"
+    fi
+    echo "::group::Run tests for " "$@"
+    ./gradlew --no-configuration-cache $continue_args "$@" $failfast_args "${COMMANDLINE_ARGS[@]}"
```
lhotari, line 51: why is `--no-configuration-cache` needed?

build.gradle.kts
```diff
@@ -0,0 +1,330 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+buildscript {
+    // The license plugin pulls in plexus-utils:2.0.6 which conflicts with
+    // the Shadow plugin's plexus-utils:4.0.2 (missing 4-arg matchPath method).
+    // Force the newer version to avoid NoSuchMethodError at shading time.
+    configurations.classpath {
+        resolutionStrategy.force("org.codehaus.plexus:plexus-utils:4.0.2")
```
lhotari, line 25: is it possible to add this version to libs.versions.toml?

build.gradle.kts
```diff
@@ -0,0 +1,330 @@
...
+val pulsarVersion = catalog.findVersion("pulsar").get().requiredVersion
+
+apply(from = "gradle/code-quality.gradle.kts")
+
+allprojects {
+    group = "org.apache.pulsar"
+    version = pulsarVersion
+}
+
+subprojects {
```
lhotari, line 44: allprojects/subprojects is no longer recommended as the best practice for large gradle builds. This is explained in "Sharing Build Logic using buildSrc",https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html .
More specifically in the "Avoid cross-project configuration using subprojects and allprojects" section.

I agree that it's easier to get started with allprojects/subproject strategy so this is fine for now.

settings.gradle.kts
```diff
@@ -0,0 +1,361 @@
...
+            ipAddresses { it.map { _ -> "0.0.0.0" } }
+        }
+    }
+}
+
+dependencyResolutionManagement {
+    @Suppress("UnstableApiUsage")
+    repositories {
+        mavenCentral()
+        maven { url = uri("https://packages.confluent.io/maven/") }
```
lhotari: there should be an additional filter so that this repository is only used for relevant dependencies. Something like this:
  maven {
        url = uri("https://packages.confluent.io/maven/")
        content {
            includeGroupByRegex("io\\.confluent\\..*")
        }
}

pulsar-broker/build.gradle.kts
```diff
@@ -0,0 +1,187 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+plugins {
+    id("com.google.protobuf") version "0.9.6"
```
lhotari: Add to libs.versions.toml

pulsar-broker/build.gradle.kts
```diff
@@ -0,0 +1,187 @@
...
+    compileOnly(libs.swagger.annotations)
+    compileOnly(libs.swagger.core)
+    compileOnly(libs.jsr305)
+
+    testImplementation(project(":testmocks"))
+    testImplementation(project(path = ":pulsar-broker-common", configuration = "testJar"))
+    testImplementation(project(path = ":managed-ledger", configuration = "testJar"))
+    testImplementation(project(path = ":pulsar-metadata", configuration = "testJar"))
+    testImplementation(project(path = ":pulsar-package-management:pulsar-package-core", configuration = "testJar"))
+    testImplementation(files(rootProject.tasks.named("filteredBkServerTestJar").map { (it as Jar).archiveFile }))
```
lhotari: use the solution mentioned earlier

.github/workflows/codeql.yaml
merlimat: done

.github/workflows/pulsar-ci-flaky.yaml
merlimat: done

distribution/shell/build.gradle.kts
merlimat: done

.github/workflows/ci-gradle-cache-update.yaml
merlimat: done

managed-ledger/build.gradle.kts
merlimat: done

distribution/io/build.gradle.kts
merlimat: Fixed the NAR plugin to register them correctly

gradle/wrapper/gradle-wrapper.properties
merlimat: done

pulsar-bom/pom.xml
merlimat: done

build.gradle.kts
merlimat: fixed

settings.gradle.kts
merlimat: done

pulsar-broker/build.gradle.kts
merlimat: done

pulsar-broker/build.gradle.kts
merlimat: done

pulsar-build/run_unit_group_gradle.sh
merlimat: Added comments (basically because of shadow plugin)

distribution/server/build.gradle.kts
merlimat: Done

distribution/offloaders/build.gradle.kts
merlimat: done

.github/workflows/pulsar-ci.yaml
merlimat: done

distribution/shell/build.gradle.kts
merlimat: done

distribution/server/build.gradle.kts
merlimat: done

distribution/server/build.gradle.kts
merlimat: done

tests/docker-images/latest-version-image/build.gradle.kts
```diff
@@ -0,0 +1,99 @@
...
+// Docker image module — no Java compilation needed
+tasks.named("compileJava") { enabled = false }
+tasks.named("compileTestJava") { enabled = false }
+tasks.named("jar") { enabled = false }
+
+val pulsarVersion = project.version.toString()
+val dockerOrganization = findProperty("docker.organization") as String? ?: "apachepulsar"
+val dockerTag = findProperty("docker.tag") as String? ?: "latest"
+val dockerPlatforms = findProperty("docker.platforms") as String? ?: ""
+val golangImage = findProperty("docker.golang.image") as String? ?: "golang:1.24-alpine"
```
lhotari: findProperty isn't compatible with configuration cache. Should use providers API such as:
    val dockerOrganization = providers.gradleProperty("docker.organization").getOrElse("apachepulsar")                                                                                                                                                                                                                                                                       
  val dockerTag = providers.gradleProperty("docker.tag").getOrElse("latest")                                                                                                                                                                                                                                                                                               
  val dockerPlatforms = providers.gradleProperty("docker.platforms").getOrElse("")
  val golangImage = providers.gradleProperty("docker.golang.image").getOrElse("golang:1.24-alpine")

tests/integration/build.gradle.kts
```diff
@@ -0,0 +1,144 @@
...
+ * under the License.
+ */
+
+// Resolve lz4-java capability conflict: at.yawk.lz4:lz4-java (used by Pulsar) and
+// org.lz4:lz4-java (used by kafka-clients) both provide the org.lz4:lz4-java capability.
+configurations.all {
+    resolutionStrategy.capabilitiesResolution.withCapability("org.lz4:lz4-java") {
+        select("at.yawk.lz4:lz4-java:0")
+    }
+}
```
lhotari: would this be beneficial for all projects?

tests/integration/build.gradle.kts
```diff
@@ -0,0 +1,144 @@
...
+    testImplementation(libs.jackson.dataformat.yaml)
+    testImplementation(libs.opensearch.rest.high.level.client)
+    testImplementation(libs.elasticsearch.java)
+    testImplementation(libs.testcontainers.elasticsearch)
+    testImplementation(libs.rabbitmq.amqp.client)
+    testImplementation(libs.joda.time)
+    testImplementation(libs.awaitility)
+    testImplementation(libs.restassured)
+    testImplementation(libs.testcontainers.localstack)
+    testImplementation("software.amazon.kinesis:amazon-kinesis-client:3.1.2")
```
lhotari: add version to libs.versions.toml

tests/pulsar-client-all-shade-test/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Use the shadow JAR from pulsar-client-all which contains relocated netty/jackson/etc classes.
+val shadedAllJar = project(":pulsar-client-all").tasks.named<Jar>("shadowJar")
+
+dependencies {
+    // Shadow JAR for compilation (provides relocated classes like org.apache.pulsar.shade.io.netty)
+    testImplementation(files(shadedAllJar.map { it.archiveFile }))
```
lhotari: The modern approach is to consume the shadow JAR as a project dependency with a specific configuration/variant. The producing project should publish the shadow JAR through an outgoing configuration:
// In :pulsar-client-all build.gradle.kts                                                                                                                                                                                                                                                                                                                                
val shadowElements by configurations.creating {                                                                                                                                                                                                                                                                                                                        
isCanBeConsumed = true                                                                                                                                                                                                                                                                                                                                               
isCanBeResolved = false
attributes {                                                                                                                                                                                                                                                                                                                                                         
attribute(Usage.USAGE_ATTRIBUTE, objects.named(Usage.JAVA_RUNTIME))                                                                                                                                                                                                                                                                                            
attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling.SHADOWED))                                                                                                                                                                                                                                                                                         
}                                                                                                                                                                                                                                                                                                                                                                    
outgoing.artifact(tasks.named<ShadowJar>("shadowJar"))                                                                                                                                                                                                                                                                                                               
}

// In the consuming project
dependencies {                                                                                                                                                                                                                                                                                                                                                           
testImplementation(project(path = ":pulsar-client-all", configuration = "shadowElements"))                                                                                                                                                                                                                                                                         
}

pulsar-client/build.gradle.kts
```diff
@@ -0,0 +1,106 @@
...
+    testImplementation(libs.opentelemetry.sdk)
+    testImplementation(libs.opentelemetry.sdk.testing)
+    testImplementation(libs.joda.time)
+    testImplementation(project(":pulsar-client-messagecrypto-bc"))
+    testImplementation(project(":pulsar-functions:pulsar-functions-proto"))
+}
+
+protobuf {
+    protoc {
+        val protocVersion = findProperty("protobufVersion") ?: libs.versions.protobuf.get()
```
lhotari: replace usage of findProperty with Providers API:
val protocVersion = providers.gradleProperty("protobufVersion")
.orElse(libs.versions.protobuf)
.get()

pulsar-client-all/build.gradle.kts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    relocate("org.asynchttpclient", "$shadePrefix.org.asynchttpclient")
+    relocate("org.checkerframework", "$shadePrefix.org.checkerframework")
+    relocate("org.codehaus.jackson", "$shadePrefix.org.codehaus.jackson")
+    relocate("org.eclipse.jetty", "$shadePrefix.org.eclipse")
+    relocate("org.glassfish", "$shadePrefix.org.glassfish")
+    relocate("org.jvnet", "$shadePrefix.org.jvnet")
+    relocate("org.objenesis", "$shadePrefix.org.objenesis")
+    relocate("org.reactivestreams", "$shadePrefix.org.reactivestreams")
+    relocate("org.roaringbitmap", "$shadePrefix.org.roaringbitmap")
```
lhotari, line 175: btw. we should probably check if the shadow plugin also handles relocating multi-release classes. Maven's shade plugin required a hack to achieve that.

pulsar-client-all/build.gradle.kts
```diff
@@ -0,0 +1,187 @@
...
+    relocate("com.spotify.futures", "$shadePrefix.com.spotify.futures")
+    relocate("com.squareup", "$shadePrefix.com.squareup")
+    relocate("com.sun.activation", "$shadePrefix.com.sun.activation")
+    relocate("com.thoughtworks.paranamer", "$shadePrefix.com.thoughtworks.paranamer")
+    relocate("com.typesafe", "$shadePrefix.com.typesafe")
+    relocate("com.yahoo", "$shadePrefix.com.yahoo")
+    relocate("com.github.benmanes", "$shadePrefix.com.github.benmanes")
+    relocate("io.airlift", "$shadePrefix.io.airlift")
+    relocate("io.grpc", "$shadePrefix.io.grpc")
+    relocate("io.netty", "$shadePrefix.io.netty")
```
lhotari, line 134: We should also relocate Netty's native libraries.

pulsar-client-shaded/build.gradle.kts
```diff
@@ -0,0 +1,166 @@
...
+        exclude(group = "it.unimi.dsi", module = "fastutil")
+    }
+    implementation(project(":pulsar-client-dependencies-minimized"))
+    implementation(project(":pulsar-client-messagecrypto-bc"))
+}
+
+val shadePrefix = "org.apache.pulsar.shade"
+
+tasks.shadowJar {
+    dependsOn(":pulsar-client-dependencies-minimized:shadowJar")
```
lhotari: this should be removed since the project dependencies already include the project and it's properly configured to expose the shadow jar

pulsar-common/build.gradle.kts
```diff
@@ -0,0 +1,135 @@
...
+        isIgnoreExitValue = true
+    }.standardOutput.asText.map { it.trim() }
+    val gitUserName = providers.exec {
+        commandLine("git", "config", "user.name")
+        isIgnoreExitValue = true
+    }.standardOutput.asText.map { it.trim() }
+    val buildHost = providers.exec { commandLine("hostname") }
+        .standardOutput.asText.map { it.trim() }
+    val buildTime = providers.exec { commandLine("date", "-u", "+%Y-%m-%dT%H:%M:%SZ") }
+        .standardOutput.asText.map { it.trim() }
```
lhotari: minor issue: these would fail when building on windows. replacing with pure Java would be better.

something like:
val buildHost = providers.provider { java.net.InetAddress.getLocalHost().hostName }                                                                                                                                                                                                                                                                                      
val buildTime = providers.provider {                                                                                                                                                                                                                                                                                                                                   
java.time.ZonedDateTime.now(java.time.ZoneOffset.UTC)                                                                                                                                                                                                                                                                                                                
.format(java.time.format.DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'"))

pulsar-bom/build.gradle.kts
```diff
@@ -0,0 +1,106 @@
...
+// Allow the platform to depend on other projects
+javaPlatform {
+    allowDependencies()
+}
+
+dependencies {
+    constraints {
+        // Client API
+        api(project(":pulsar-client-api"))
+        api(project(":pulsar-client-admin-api"))
```
lhotari, line 37: Before a release there will be a need to have proper configuration for projects that get published to maven. 

After that is setup, something like this could be performed:
rootProject.subprojects {                                                                                                                                                                                                                                                                                                                                                            
pluginManager.withPlugin("maven-publish") {                                                                                                                                                                                                                                                                                                                          
dependencies {                                                                                                                                                                                                                                                                                                                                                   
constraints {                                                                                                                                                                                                                                                                                                                                                
api(project(path))
}                                                                                                                                                                                                                                                                                                                                                            
}
}                                                                                                                                                                                                                                                                                                                                                                    
}

settings.gradle.kts
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// ──────────────────────────────────────────────────────────────────────────────
+// Docker modules (enabled with -Pdocker)
+// ──────────────────────────────────────────────────────────────────────────────
+
+// Also auto-enable when running docker-related tasks (e.g., ./gradlew docker)
+val dockerRequested = settings.extra.has("docker") ||
+    gradle.startParameter.taskNames.any { it.contains("docker", ignoreCase = true) }
+if (dockerRequested) {
```
lhotari, line 322: the conditionals would most likely need to use Providers API so that these are compatible with configuration cache. We could resolve configuration cache compatibility in a later PR if it slows down things at this point.

It seems very maven like to handle any conditionals in the settings.gradle.kts and isn't generally recommended for gradle. One of the problems is that in the IDE, you wouldn't get Gradle integration when the projects unless you set the flags. 

The proper solution would be to define the dependencies in the gradle build. This would also ensure that when integration tests are run, the docker image would only get built again if there were no code changes. 

Perhaps Claude finds a solution to resolve this where we could get rid of all conditionals and the "core modules" solution completely.

Since we are getting rid of Pulsar IO connectors in the project, that would also simplify the build and we could get rid of the "core modules" part and properly implementing the build to work with Pulsar IO connectors as part of it.

pulsar-functions/worker/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
...
+// NAR/JAR files needed by tests (mirrors Maven's maven-dependency-plugin config).
+// Use task path strings so we don't force cross-project configuration at config time.
+tasks.withType<Test> {
+    dependsOn(
+        ":pulsar-functions:pulsar-functions-api-examples:jar",
+        ":pulsar-functions:pulsar-functions-api-examples-builtin:nar",
+        ":pulsar-io:pulsar-io-data-generator:nar",
+        ":pulsar-io:pulsar-io-netty:nar",
+        ":pulsar-io:pulsar-io-cassandra:nar",
+    )
```
lhotari: The recommended approach would be to use dependency configuration also for these cases. However, it's not a show stopper.

tests/docker-images/latest-version-image/build.gradle.kts
merlimat: done — replaced `findProperty()` with `providers.gradleProperty()` in all Docker build scripts

tests/integration/build.gradle.kts
merlimat: done — added `amazon-kinesis-client-v3` to libs.versions.toml

distribution/shell/build.gradle.kts
merlimat: The `distLib` configuration already extends from `implementation` which inherits the enforced platform constraints. The dependencies declared here are the same ones Maven lists in the distribution POM — they're application-level deps, not version constraints.

distribution/server/build.gradle.kts
merlimat: Acknowledged — this is simplified now that projects were renamed. The remaining mapping is minimal (bouncy-castle-bc → pkg classifier). Will revisit if needed.

pulsar-client/build.gradle.kts
merlimat: done — replaced all `findProperty()` with `providers.gradleProperty()` across all build scripts

pulsar-client-all/build.gradle.kts
merlimat: done — added `filesMatching("META-INF/native/**")` to rename netty native libs with `org_apache_pulsar_shade_` prefix in all 3 shaded JARs

pulsar-common/build.gradle.kts
merlimat: done — replaced `hostname` and `date` commands with `InetAddress.getLocalHost().hostName` and `ZonedDateTime.now()`

settings.gradle.kts
merlimat: The Providers API isn't available in `settings.gradle.kts` the same way. The `settings.extra.has()` and `gradle.startParameter.taskNames` are the standard approaches in settings scripts. Settings are always re-evaluated (not cached by the configuration cache).

pulsar-functions/worker/build.gradle.kts
merlimat: done — replaced cross-project task references with dependency configurations (`testNars`, `testExamplesJar`) that resolve NAR/JAR files through Gradle's dependency model

pulsar-io/kinesis/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+    testImplementation(libs.awaitility)
+    testImplementation(libs.jsonassert)
+}
+
+// The KPL shaded project's default jar is disabled (shadow JAR replaces it).
+// Gradle can't infer the task dependency from runtimeClasspath alone, so
+// declare it explicitly for the NAR task.
+tasks.named("nar") {
+    dependsOn(":pulsar-io:pulsar-io-kinesis-kpl-shaded:shadowJar")
+}
```
lhotari: The modern approach is to consume the shadow JAR as a project dependency with a specific configuration/variant. The producing project should publish the shadow JAR through an outgoing configuration:

// In :pulsar-client-all build.gradle.kts
val shadowElements by configurations.creating {
isCanBeConsumed = true
isCanBeResolved = false
attributes {
attribute(Usage.USAGE_ATTRIBUTE, objects.named(Usage.JAVA_RUNTIME))
attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling.SHADOWED))
}
outgoing.artifact(tasks.named<ShadowJar>("shadowJar"))
}

// In the consuming project
dependencies {
testImplementation(project(path = ":pulsar-client-all", configuration = "shadowElements"))
}

pulsar-client-shaded/build.gradle.kts
merlimat: done — removed all 3 `dependsOn` for `pulsar-client-dependencies-minimized:shadowJar`. Verified the shadow JAR is automatically triggered through the project dependency chain.

pulsar-client-all/build.gradle.kts
merlimat: Verified — the Shadow plugin handles multi-release JARs correctly. Classes under `META-INF/versions/` are properly relocated (e.g., `META-INF/versions/11/org/apache/pulsar/shade/com/fasterxml/jackson/...`). Output matches Maven's shade plugin.

tests/integration/build.gradle.kts
merlimat: done — moved lz4-java capability resolution to root `build.gradle.kts` (applied to all subprojects). Removed the duplicated blocks from kafka, kafka-connect-adaptor, debezium/core, and integration test modules.

pulsar-bom/build.gradle.kts
merlimat: Yes, ASF release will need another pass of changes for sure

tests/pulsar-client-all-shade-test/build.gradle.kts
merlimat: done — added `shadowElements` consumable configuration (shared via root `build.gradle.kts` for all Shadow plugin modules). Shade tests now use `project(path = "...", configuration = "shadowElements")` instead of `files(shadowJar)`.

merlimat: Done

tests/integration/build.gradle.kts
merlimat: Yes, moved it in top build file.

tests/pulsar-client-all-shade-test/build.gradle.kts
merlimat: Also fixed kinesis — now uses `implementation(project(path = "...", configuration = "shadowElements"))` instead of `dependsOn` hack. The shadow JAR is automatically resolved and bundled in the NAR.

pulsar-io/kinesis/build.gradle.kts
merlimat: done

merlimat:
Yes, this is just the starting point, that will allow development not to be interrupted. There are still remaining tasks needed and a lot of improvement opportunities.

settings.gradle.kts
lhotari: A added a comment about this in https://github.com/apache/pulsar/pull/25398#pullrequestreview-4013949246, we can mark this conversation resolved and handle this later. The development experience in the IDE won't be great when there are conditionals. That's one of the reasons why they should be removed.

lhotari: The way to resolve the core-modules vs. full-build would be to create a separate distribution for pulsar-all. For building everything, one would either build the pulsar-all docker image or the full-build distribution with connectors and offloaders.

merlimat: `pulsar-all` will go away as part of PIP-465, so it will simplify a bit the things. 

I think we'll be able to remove core-module profile then (maybe just keep a profile to skip the shaded jars modules).

lhotari:
> Yes, this is just the starting point, that will allow development not to be interrupted. There are still remaining tasks needed and a lot of improvement opportunities.

Yes, let's get this merged and start iterating. 

There's a great Gradle cache that would be useful for pull requests. With the solution only changed modules would required to be compiled, checked and so on. It's coming from an Apache Calcite maintainer and the action is already is already approved. https://github.com/marketplace/actions/gradle-cache#sample-integrations

settings.gradle.kts
lhotari: > maybe just keep a profile to skip the shaded jars modules

That shouldn't be necessary. unless shaded tests are run, there shouldn't be a need to build them unless there's some other dependency. This seems to apply for the other use cases too (docker, integration tests, test groups in bash scripts). Besides task dependencies it's possible to use [Lifecycle tasks](https://docs.gradle.org/current/userguide/organizing_tasks.html#sec:lifecycle_tasks) to aggregate dependencies. The way to use it for grouping is to add a lifecycle task (a task with just a name and dependencies) to every project with the same name, it covers most use cases.

pulsar-bom/build.gradle.kts
lhotari: I can start working on the release part after this PR has been merged.

PR #25437, opened 2026-03-30
[fix][broker]Producer with AUTO_PRODUCE schema failed to reconnect, which caused by schema incompatible

poorbarcode:
### Motivation

**Background**
- Under schema-validation-enforced settings
- Create a producer with `AUTO_PRODUCE` schema
  - The producer started.
- Registers a new schema to the topic by Admin
- The producer reconnects
  - **(Highlight)Issue**:  the producer could reconnect with stale `AutoProduceBytesSchema (BYTES)`, which then became incompatible with the schema already registered on the cluster.

You can reproduce the issue with the new test `testReplicationReconnectWithSchemaValidationEnforced`

### Modifications

The goal is to make replication recover automatically: when an `IncompatibleSchemaException` occurs, refresh the topic schema and retry producer creation so the replication pipeline can resume.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
@@ -86,6 +86,7 @@ import org.apache.pulsar.client.impl.metrics.InstrumentProvider;
 import org.apache.pulsar.client.impl.metrics.LatencyHistogram;
 import org.apache.pulsar.client.impl.metrics.Unit;
 import org.apache.pulsar.client.impl.metrics.UpDownCounter;
+import org.apache.pulsar.client.impl.schema.AutoProduceBytesSchema;
 import org.apache.pulsar.client.impl.schema.JSONSchema;
 import org.apache.pulsar.client.impl.schema.SchemaUtils;
 import org.apache.pulsar.client.impl.transaction.TransactionImpl;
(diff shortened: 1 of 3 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java
```diff
@@ -2021,6 +2022,16 @@ public CompletableFuture<Void> connectionOpened(final ClientCnx cnx) {
...
 
+            if (cause instanceof PulsarClientException.IncompatibleSchemaException
+                    && schema instanceof AutoProduceBytesSchema autoProduceBytesSchema
+                    && !autoProduceBytesSchema.hasUserProvidedSchema()) {
+                client.reloadSchemaForAutoProduceProducer(topic, autoProduceBytesSchema)
+                    .whenComplete((__, throwable) -> {
+                        future.completeExceptionally(cause);
+                    });
+                return null;
+            }
```
codelipenghui, line 2033: Will this impact the producer with auto produce schema created by user? As I understand only replicator can refresh the schema, but user will specify the schema. Will it cause infinite loop with IncompatibleSchemaException?

poorbarcode: > Will this impact the producer with auto produce schema created by user? 

It does not matter.
- The schema AutoProduceBytesSchema was only used for registering the producer.
- The producer created by the user will register schemas when publishing messages, see also https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L2521-L2523

>  As I understand only replicator can refresh the schema, but user will specify the schema. Will it cause infinite loop with IncompatibleSchemaException?

Since the producer with `AUTO_PRODUCE` schema will use the schema that was got from the server-side, it will not get the error `IncompatibleSchemaException` anymore

codelipenghui (approved):

PR #21648, opened 2023-11-30
[improve][pip] PIP-321  Introduce `allowed-cluster` at the namespace level

liangyepianzhou:
### Motivation
Because the configurations of clusters at the namespace level are unclear, some flexible topic policies can not work as expected, e.g. geo-replication at the topic level. see more information about the bug at https://github.com/apache/pulsar/pull/21564.
### Modifications
Use `replication-clusters`, `allowed-clusters`, and `topic-policy-synchronized-clusters` to replace a single `replication-clusters` originally in the namespace policy.
See more explanation in the proposal.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Matching PR in forked repository

PR in forked repository:

PR #25443, opened 2026-03-31
[fix][broker]The partitions of the topic was wrongly created even though this cluster was not allowed to access it

poorbarcode:
### Motivation

Env:
- `cluster-1` and `cluster-2` share configuration metadata store
- create a partitioned topic `public/default/tp1`
  - The ZK node `/admin/partitioned-topics/{topic}` will be created on the shared configuration metadata store.
  - The topic will enable a binary way replication between `cluster-1` and `cluster-2`
- Remove `cluster-2` from the topic-level `replication policies`
  - The partition `public/default/tp1-partition-0` will be deleted automatically.

Issue: call `pulsar-admin topics create-missed-partitions <topic>`, the partition `public/default/tp1-partition-0` was loaded up again.

### Modifications

Fix the issue

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

PR in forked repository: x

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -550,7 +550,7 @@ public class PersistentTopicsBase extends AdminResource {
     }
 
     protected void internalCreateMissedPartitions(AsyncResponse asyncResponse) {
-        pulsar().getBrokerService().isAllowedCurrentClusterAccess(topicName).thenAccept(allowed -> {
+        pulsar().getBrokerService().isCurrentClusterAllowed(topicName).thenAccept(allowed -> {
             if (!allowed) {
                 resumeAsyncResponseExceptionally(asyncResponse,
                     new RestException(Status.BAD_REQUEST, String.format("Topic [%s] is not allowed to be loaded"
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -550,31 +550,39 @@ public class PersistentTopicsBase extends AdminResource {
     }
 
     protected void internalCreateMissedPartitions(AsyncResponse asyncResponse) {
-        getPartitionedTopicMetadataAsync(topicName, false, false).thenAccept(metadata -> {
-            if (metadata != null && metadata.partitions > 0) {
-                validateNamespaceOperationAsync(topicName.getNamespaceObject(),
-                        NamespaceOperation.CREATE_TOPIC)
-                .thenCompose(__ -> tryCreatePartitionsAsync(metadata.partitions)).thenAccept(v -> {
-                    asyncResponse.resume(Response.noContent().build());
-                }).exceptionally(e -> {
-                    log.error()
-                            .attr("topic", topicName)
-                            .log("Failed to create partitions for topic");
-                    resumeAsyncResponseExceptionally(asyncResponse, e);
-                    return null;
-                });
-            } else {
-                throw new RestException(Status.NOT_FOUND, String.format("Topic %s does not exist", topicName));
+        pulsar().getBrokerService().isAllowedCurrentClusterAccess(topicName).thenAccept(allowed -> {
+            if (!allowed) {
+                resumeAsyncResponseExceptionally(asyncResponse,
+                    new RestException(Status.BAD_REQUEST, String.format("Topic [%s] is not allowed to be loaded"
+                        + " up on the current cluster, please recheck replication polices.", topicName)));
             }
-        }).exceptionally(ex -> {
-            // If the exception is not redirect exception we need to log it.
-            if (!isRedirectException(ex)) {
-                log.error()
+            getPartitionedTopicMetadataAsync(topicName, false, false).thenAccept(metadata -> {
+                if (metadata != null && metadata.partitions > 0) {
+                    validateNamespaceOperationAsync(topicName.getNamespaceObject(),
+                        NamespaceOperation.CREATE_TOPIC)
+                        .thenCompose(__ -> tryCreatePartitionsAsync(metadata.partitions)).thenAccept(v -> {
+                            asyncResponse.resume(Response.noContent().build());
+                        }).exceptionally(e -> {
+                            log.error()
+                                    .attr("topic", topicName)
+                                    .log("Failed to create partitions for topic");
+                            resumeAsyncResponseExceptionally(asyncResponse, e);
+                            return null;
+                        });
+                } else {
+                    resumeAsyncResponseExceptionally(asyncResponse,
+                            new RestException(Status.NOT_FOUND, String.format("Topic %s does not exist", topicName)));
+                }
+            }).exceptionally(ex -> {
+                // If the exception is not redirect exception we need to log it.
+                if (!isRedirectException(ex)) {
+                    log.error()
                         .attr("topic", topicName)
                         .log("Failed to create partitions for topic");
-            }
-            resumeAsyncResponseExceptionally(asyncResponse, ex);
-            return null;
+                }
+                resumeAsyncResponseExceptionally(asyncResponse, ex);
+                return null;
+            });
         });
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
@@ -47,6 +47,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import javax.ws.rs.WebApplicationException;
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -2010,29 +2013,82 @@ private CompletableFuture<Void> checkTopicAlreadyMigrated(TopicName topicName) {
...
+            }
+            // No settings for replication/allowed_clusters.
+            if (nsPolicies.isEmpty()) {
+                return true;
+            }
+            // Namespace level settings.
+            return nsPolicies.get().replication_clusters.isEmpty()
+                    || nsPolicies.get().replication_clusters.contains(cluster);
+        });
+    }
```
codelipenghui: Don't we already have a mechanism for retrieving the applied policies for a topic? We should use it to avoid writing duplicate code.

poorbarcode: > Don't we already have a mechanism for retrieving the applied policies for a topic? We should use it to avoid writing duplicate code.

No, we don't have such a method. There are currently two methods

- Persistent topic has a `HierarchyTopicPolicies` of merged policies, and this result is relied upon for inspection.
- Check when the namespace modifies replication/allowed-cluster, see also `BrokerService.isCurrentClusterAllowed(Namespace)`

The newly added method has maximised the reuse of code (by merging the namespace check method `BrokerService.isCurrentClusterAllowed(Namespace`).

BTW, the method newly added will become the method you mentioned: `a mechanism for retrieving the applied policies for a topic`

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -550,31 +550,39 @@ private CompletableFuture<Set<String>> getReplicationClusters() {
     }
 
     protected void internalCreateMissedPartitions(AsyncResponse asyncResponse) {
-        getPartitionedTopicMetadataAsync(topicName, false, false).thenAccept(metadata -> {
-            if (metadata != null && metadata.partitions > 0) {
-                validateNamespaceOperationAsync(topicName.getNamespaceObject(),
+        pulsar().getBrokerService().isCurrentClusterAllowed(topicName).thenAccept(allowed -> {
+            if (!allowed) {
+                resumeAsyncResponseExceptionally(asyncResponse,
```
Denovo1998, line 565: Here resumeAsyncResponseExceptionally(...) does not stop the subsequent chain, so even if allowed=false, the subsequent getPartitionedTopicMetadataAsync(...)->tryCreatePartitionsAsync(...) will still continue to execute. The client may receive a failure, but the partition will still be recreated.

Change to a single thenCompose chain, and directly return a failed future when not allowed?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
```diff
@@ -550,31 +550,39 @@ private CompletableFuture<Set<String>> getReplicationClusters() {
     }
 
     protected void internalCreateMissedPartitions(AsyncResponse asyncResponse) {
-        getPartitionedTopicMetadataAsync(topicName, false, false).thenAccept(metadata -> {
-            if (metadata != null && metadata.partitions > 0) {
-                validateNamespaceOperationAsync(topicName.getNamespaceObject(),
+        pulsar().getBrokerService().isCurrentClusterAllowed(topicName).thenAccept(allowed -> {
```
Denovo1998, line 563: The original implementation had a unified exceptionally(...) at the outermost layer to ensure that AsyncResponse would definitely be resumed; now that the outermost layer has been replaced with isCurrentClusterAllowed(...).thenAccept(...), no error handling has been attached outside anymore. If an exception is thrown during the process of reading the namespace/topic policy, this suspended request will be hung indefinitely until the timeout. Here, the unified exceptionally/whenComplete at the outermost layer should be retained.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalZKTest.java
```diff
@@ -149,6 +149,13 @@ public void testDeleteRemoteTopicByGlobalPolicy() throws Exception {
         });
         waitReplicatorStopped(subTopic, pulsar1, pulsar2, true);
 
+        try {
+            admin2.topics().createMissedPartitions(topicName);
```
Denovo1998, line 153: This test currently only asserts that the admin call throws an exception, but does not assert that the partition still does not exist after the failure. Since "request rejected" and "subsequent continue tryCreatePartitionsAsync" are two separate paths in the implementation, relying solely on the exception message cannot prevent false positives where "the response fails but the side effect has already occurred". It is recommended to add another assertion to confirm that topicName-partition-0 has not been recreated, or that the corresponding replicator remains stopped.

Denovo1998 (dismissed):

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorUsingGlobalZKTest.java
poorbarcode: Fixed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
poorbarcode: Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/PersistentTopicsBase.java
poorbarcode: Thanks, it is really a bug. Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -4136,6 +4153,38 @@ public void setPulsarChannelInitializerFactory(PulsarChannelInitializer.Factory
         this.pulsarChannelInitFactory = factory;
     }
 
+    /**
+     * @return Triple [namespace policies, global topic policies, topic policies].
```
codelipenghui: Stale copy-paste from `getCombinedTopicPolicies` — this method returns `CompletableFuture<Boolean>`, not a `Triple`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -4136,6 +4153,38 @@ public void setPulsarChannelInitializerFactory(PulsarChannelInitializer.Factory
...
+     */
+    public CompletableFuture<Boolean> isCurrentClusterAllowed(@NonNull TopicName topicName) {
+        final String cluster = getPulsar().getConfig().getClusterName();
+        return getCombinedTopicPolicies(topicName).thenApply(triple -> {
+            Optional<TopicPolicies> topicP = triple.getRight();
+            Optional<TopicPolicies> globalTopicP = triple.getMiddle();
+            Optional<Policies> nsPolicies = triple.getLeft();
+            // Disabled a cluster for a namespace manually.
+            if (nsPolicies.isPresent() && !isCurrentClusterAllowed(topicName.getNamespaceObject(), nsPolicies.get())) {
```
codelipenghui, line 4166: Possible regression: `isCurrentClusterAllowed(NamespaceName, Policies)` returns false when namespace `replication_clusters` excludes the current cluster (with empty `allowed_clusters`). But topic-level `replicationClusters` is supposed to override namespace-level defaults — e.g. ns `replication_clusters=[c1]`, topic `replicationClusters=[c1,c2]` should be allowed on `c2`, but this short-circuit returns false.

Suggest only short-circuiting on the `allowed_clusters` hard gate (PIP-321), then deferring to topic-level checks before falling back to ns `replication_clusters`.

codelipenghui (commented):
Looks good to me, left a few comments.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
poorbarcode: > Possible regression: isCurrentClusterAllowed(NamespaceName, Policies) returns false when namespace replication_clusters excludes the current cluster (with empty allowed_clusters). But topic-level replicationClusters is supposed to override namespace-level defaults — e.g. ns replication_clusters=[c1], topic replicationClusters=[c1,c2] should be allowed on c2, but this short-circuit returns false.

This situation does not exist. Let's review all possible use cases:

1. Both clusters use a shared metadata store
  a. The topic will never be allowed to be accessed by the current cluster, even if the topic-level replication policy contains `{current cluster}`.
  b. The [PIP-321 Introduce `allowed-cluster` at the namespace level](https://github.com/apache/pulsar/pull/21648/changes#diff-01b14c55e65a5acb0c710b4eb0d45e039f43a2d5f54ec4740ee64f069ef6abc6R29-R30) introduces concept `namespace level allowed_clusters` to let clusters that using shared metadata store can enable topic level replication.
2. Both clusters use their own metadata store
  a. All namespace-level `replication_clutsers` will contain the current cluster, and namespace-level `replication cluster` can not be set to a empty value. 
  b. No one will remove `current cluster` from namespace-level `replication_clusters`, which is meaningless

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
poorbarcode: Good pointer, I fixed it

Denovo1998 (approved):

Denovo1998:
Sorry. I forgot this.

PR #25457, opened 2026-04-01
[improve][build] Add Maven publishing conventions for ASF release

lhotari:
### Motivation

The Pulsar build was migrated from Maven to Gradle, but Maven publishing (deploying artifacts to Maven repositories) was not yet implemented. This PR adds the publishing infrastructure so artifacts land at the same Maven coordinates as before (`org.apache.pulsar:*`), with proper POM metadata, GPG signing, sources/javadoc JARs, and Gradle module metadata.

#### Parent POM (`build.gradle.kts`)

The root project publishes `org.apache.pulsar:pulsar` as a POM-only parent artifact containing all shared ASF metadata (license, SCM, organization, mailing lists, developers, issue management). Child module POMs reference this via `<parent>`, inheriting metadata instead of embedding it — matching the Maven build's parent POM structure.

#### Convention plugins (`build-logic/`)

**`pulsar.public-java-library-conventions`** (new) — combined plugin for published Java library modules that applies both `pulsar.java-conventions` and `pulsar.publish-conventions`. Establishes a clear convention boundary: modules using this plugin are public libraries published to Maven, while internal-only modules use just `pulsar.java-conventions`. Includes dependency validation ensuring public modules don't depend on unpublished projects.

**`pulsar.publish-conventions`** (new) — core publishing plugin configuring `maven-publish` and `signing`:

- **java-library modules**: publishes JAR + sources JAR + javadoc JAR + POM + Gradle module metadata (`.module`), with version mapping that inlines resolved dependency versions into the POM
- **java-platform modules** (`pulsar-bom`, `pulsar-dependencies`): publishes POM + Gradle module metadata only (no JAR), with `<packaging>pom</packaging>` and `<dependencyManagement>` entries
- **Shadow/shaded modules** (`pulsar-client-shaded`, `pulsar-client-admin-shaded`, `pulsar-client-all`, etc.): publishes the shadow JAR as the primary artifact
- **Child POM `<parent>` injection**: every child POM gets a `<parent>` reference to the root parent POM via `withXml`
- **POM cleanup**: removes `<scope>compile</scope>` (Maven default) and `<dependencyManagement>` from non-platform POMs
- **GPG signing**: configurable via Gradle properties (`useGpgCmd`, `signing.gnupg.keyName`, `signing.gnupg.homeDir`), disabled by default for local development
- **Local deploy repository**: `build/local-deploy-repo` via `publishAllPublicationsToLocalDeployRepository` task
- **Configuration cache compatible**

**`pulsar.nar-conventions`** (updated) — now applies `pulsar.publish-conventions` and configures NAR-specific publishing:

- Publishes only the `.nar` artifact with `<packaging>nar</packaging>`
- POM has no `<dependencies>` section — NAR bundles everything
- No sources/javadoc JARs (NAR is self-contained)
- Unpublished example NAR modules (`java-examples-builtin`, `client-tools-customcommand-example`) disable publishing explicitly

#### Module changes

- **52 published java-library modules**: use `id("pulsar.public-java-library-conventions")`
- **4 published NAR modules**: use `id("pulsar.java-conventions")` + `id("pulsar.nar-conventions")` — publishing comes from `nar-conventions`
- **2 platform modules** (`pulsar-bom`, `pulsar-dependencies`): use `id("pulsar.publish-conventions")` directly
- **Unpublished modules** (`buildtools`, `tests/*`, `microbench`, `jclouds-shaded`, etc.): use `id("pulsar.java-conventions")`

**Modules changed to internal (no longer published)**:
- `jclouds-shaded` — only referenced by `tiered-storage/jcloud` (a NAR module that bundles it)

#### Project coordinates (`gradle.properties`)

Moved `group` and `version` from the version catalog / `pulsar.java-conventions` to `gradle.properties` — the standard Gradle pattern. Gradle auto-applies these to all projects.

#### Other

- `gradle.properties`: added `systemProp.org.gradle.internal.publish.checksums.insecure=true` for ASF Nexus compatibility
- `gradle/setup-test-gpg.sh`: helper script to create a temporary GPG keyring for testing signing without touching `~/.gnupg`

### Verifying this change

- Ran `./gradlew publishAllPublicationsToLocalDeployRepository` — all 58 modules published successfully (52 java-library + 4 NAR + 2 platform)
- Plus the root parent POM (`org.apache.pulsar:pulsar`)
- Verified parent POM contains all ASF metadata (license, SCM, organization, developers, etc.)
- Verified child POMs have `<parent>` reference and no inline metadata
- Verified `pulsar-bom` produces only `.pom` + `.module` files with `<dependencyManagement>` entries
- Verified `pulsar-client-shaded` publishes 24MB shadow JAR as primary artifact
- Verified NAR modules publish `.nar` with `<packaging>nar</packaging>` and **no `<dependencies>` section**
- Verified `jclouds-shaded` is NOT published
- Verified example NAR modules (`java-examples-builtin`, `client-tools-customcommand-example`) are NOT published
- Verified GPG signing produces `.asc` files when configured with a test key
- Verified Gradle module metadata (`.module` files) published alongside all POMs
- Verified configuration cache compatibility
- Verified unpublished dependency validation: temporarily adding `implementation(project(":buildtools"))` to a published module correctly fails with:
  ```
  Published module ':pulsar-client-api' depends on unpublished projects:
    - implementation -> :buildtools (not published)
  Either publish the dependency or move it to a test/compileOnly scope.
  ```

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [x] `doc-not-needed`

### DIFF
--- a/bouncy-castle/bc/build.gradle.kts
+++ b/bouncy-castle/bc/build.gradle.kts
@@ -19,6 +19,7 @@
 
 plugins {
     id("pulsar.java-conventions")
+    id("pulsar.publish-conventions")
 }
 
 dependencies {
--- a/bouncy-castle/bc/build.gradle.kts
+++ b/bouncy-castle/bc/build.gradle.kts
@@ -18,8 +18,7 @@
  */
 
 plugins {
-    id("pulsar.java-conventions")
-    id("pulsar.publish-conventions")
+    id("pulsar.public-java-library-conventions")
 }
 
 dependencies {
--- a/build-logic/conventions/src/main/kotlin/pulsar.publish-conventions.gradle.kts
+++ b/build-logic/conventions/src/main/kotlin/pulsar.publish-conventions.gradle.kts
@@ -1,0 +1,219 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Convention plugin for publishing Pulsar modules to Maven repositories.
+// Configures maven-publish, GPG signing, POM metadata, sources/javadoc JARs,
+// and a local deploy repository for testing.
+
+plugins {
+    `maven-publish`
+    signing
+}
+
+// --- java-library projects: JAR + sources + javadoc ---
+pluginManager.withPlugin("java-library") {
+    val sourceSets = the<SourceSetContainer>()
+
+    // Match Maven's javadoc configuration: no doclint, don't fail on errors
+    tasks.withType<Javadoc>().configureEach {
+        (options as StandardJavadocDocletOptions).apply {
+            addStringOption("Xdoclint:none", "-quiet")
+        }
+        isFailOnError = false
+    }
+
+    val sourcesJar by tasks.registering(Jar::class) {
+        archiveClassifier.set("sources")
+        from(sourceSets["main"].allJava)
+    }
+
+    val javadocJar by tasks.registering(Jar::class) {
+        archiveClassifier.set("javadoc")
+        from(tasks.named(JavaPlugin.JAVADOC_TASK_NAME))
+    }
+
+    // NAR modules disable the jar task and produce .nar files instead.
+    // Detect the NAR plugin and configure the publication accordingly.
+    val isNarModule = plugins.hasPlugin("io.github.merlimat.nar")
+
+    if (isNarModule) {
+        // NAR modules: publish the .nar artifact with packaging=nar
+        publishing {
+            publications {
+                create<MavenPublication>("maven") {
+                    artifact(tasks.named("nar"))
+                    artifact(sourcesJar)
+                    artifact(javadocJar)
+                    // Set packaging to "nar" in POM
+                    pom.packaging = "nar"
+                }
+            }
+        }
+    } else {
+        // Standard java-library modules: publish from components["java"]
+        publishing {
+            publications {
+                create<MavenPublication>("maven") {
+                    from(components["java"])
+                    artifact(sourcesJar)
+                    artifact(javadocJar)
+
+                    versionMapping {
+                        usage(Usage.JAVA_RUNTIME) {
+                            fromResolutionResult()
+                        }
+                        usage(Usage.JAVA_API) {
+                            fromResolutionOf("runtimeClasspath")
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+// --- java-platform projects (BOM, dependencies): POM-only, no JAR ---
+pluginManager.withPlugin("java-platform") {
+    publishing {
+        publications {
+            create<MavenPublication>("maven") {
+                from(components["javaPlatform"])
+            }
+        }
+    }
+}
+
+// --- Common POM metadata for all publications ---
+run {
+    // Capture values in a local scope so withXml closures don't capture the script object
+    // (which would break configuration cache serialization)
+    val projectName = project.name
+    val projectDescription = project.description
+    val archivesNameValue = the<BasePluginExtension>().archivesName.get()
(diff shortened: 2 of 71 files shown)


bouncy-castle/bc/build.gradle.kts
```diff
@@ -19,6 +19,7 @@
 
 plugins {
     id("pulsar.java-conventions")
+    id("pulsar.publish-conventions")
```
merlimat: Is this going to be used to select which modules are going to be published? 

I think there are quite a few modules that really shouldn't be published as they're not part of client apis/sdks at all :)

build-logic/conventions/src/main/kotlin/pulsar.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,219 @@
...
+        publications {
+            withType<MavenPublication>().configureEach {
+                artifactId = archivesNameValue
+
+                pom {
+                    name.set("Apache Pulsar :: $projectName")
+                    description.set(projectDescription ?: "Apache Pulsar :: $projectName")
+                    url.set("https://pulsar.apache.org")
+                    inceptionYear.set("2017")
+
```
nodece: Please add developer information.
```
developers {
    developer {
        organization.set("Apache Pulsar developers")
        organizationUrl.set("https://pulsar.apache.org/")
    }
}
```

build-logic/conventions/src/main/kotlin/pulsar.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,219 @@
...
+    val isPlatformProject = plugins.hasPlugin("java-platform")
+    val localDeployRepoDir = rootProject.layout.buildDirectory.dir("local-deploy-repo")
+
+    publishing {
+        publications {
+            withType<MavenPublication>().configureEach {
+                artifactId = archivesNameValue
+
+                pom {
+                    name.set("Apache Pulsar :: $projectName")
```
nodece:
```suggestion
                    name.set("$projectName")
```

bouncy-castle/bc/build.gradle.kts
lhotari: I believe that this is the recommended approach in Gradle builds. We won't be adding new modules often, so it's possible to decide at that point whether a module should be published or not. For example, testing related modules shouldn't be published by default. If java-conventions publishes by default, it will be forgotten. 
For publishing, there should be an extra check that ensures that project references in a published project are satisfied. That prevents mistakes where resolving a dependency from maven central wouldn't be possible.

build-logic/conventions/src/main/kotlin/pulsar.publish-conventions.gradle.kts
lhotari: This is following the current convention in branch-4.2 maven pom.xml files.

build-logic/conventions/src/main/kotlin/pulsar.publish-conventions.gradle.kts
lhotari: Good point. 

I also noticed that all of this metadata should be in a parent pom instead of duplicating the information.

lhotari: Fixed

bouncy-castle/bc/build.gradle.kts
lhotari: To simplify, I'll add a separate plugin "pulsar.public-java-library-conventions" which will include java-conventions and publish-conventions so that there wouldn't be a need to have 2 separate plugins for public libraries that need to be published. All projects that are public cannot reference internal projects in the compile/runtime maven scopes. That could be checked at publishing time in a better way when there's a explicit convention for public library projects.

nodece (approved):

jclouds-shaded/build.gradle.kts
```diff
@@ -18,7 +18,7 @@
  */
 
 plugins {
-    id("pulsar.java-conventions")
+    id("pulsar.public-java-library-conventions")
```
merlimat: Should this just use `java-conventions`? It doesn't need to be published on Maven central (as several others).

bouncy-castle/bc/build.gradle.kts
lhotari: This has been addressed. @merlimat, PTAL

jclouds-shaded/build.gradle.kts
lhotari: I'll revisit that. We might still want to publish just the tiered-storage nar files to Maven central for easier access, however all dependencies can be dropped from the pom.xml when it's a nar project.

lhotari: @merlimat jclouds-shaded won't get published anymore. I've revisited the way how nar files are published. All dependencies will be removed from pom.xml files in nar projects and just the nar file will get published.

lhotari: PTAL @merlimat

merlimat (approved):

PR #25461, opened 2026-04-03
[improve][broker]Part-3 of PIP-433:  always allow replicator to register a new compatible schema

poorbarcode:
### Motivation

This is part-3 of PIP-433, see the goal 1 in [PIP-433](https://github.com/apache/pulsar/blob/master/pip/pip-433.md)

### Modifications

 always allow the replicator to register a new compatible schema, users can turn off this function

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -2754,13 +2754,18 @@ public abstract class NamespacesBase extends AdminResource {
                 "schemaValidationEnforced");
     }
 
-    protected void internalSetIsAllowAutoUpdateSchema(boolean isAllowAutoUpdateSchema) {
+    protected void internalSetIsAllowAutoUpdateSchema(boolean isAllowAutoUpdateSchema,
+                                                      Boolean isAllowAutoUpdateSchemaWithReplicator) {
         validateNamespacePolicyOperation(namespaceName, PolicyName.SCHEMA_COMPATIBILITY_STRATEGY,
                 PolicyOperation.WRITE);
         validatePoliciesReadOnlyAccess();
 
         mutatePolicy((policies) -> {
                     policies.is_allow_auto_update_schema = isAllowAutoUpdateSchema;
+                    if (isAllowAutoUpdateSchemaWithReplicator != null) {
+                        policies.is_allow_auto_update_schema_with_replicator =
+                                isAllowAutoUpdateSchemaWithReplicator;
+                    }
                     return policies;
                 }, (policies) -> policies.is_allow_auto_update_schema,
                 "isAllowAutoUpdateSchema");
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -1808,8 +1808,10 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                     }
 
                     disableTcpNoDelayIfNeeded(topicName.toString(), producerName);
-
-                    CompletableFuture<SchemaVersion> schemaVersionFuture = tryAddSchema(topic, schema);
+                    boolean isReplicatorProducer = Producer.isRemoteOrShadow(producerName,
+                            getBrokerService().getPulsar().getConfig().getReplicatorPrefix());
+                    CompletableFuture<SchemaVersion> schemaVersionFuture = tryAddSchema(topic, schema,
+                            isReplicatorProducer);
 
                     schemaVersionFuture.exceptionallyAsync(exception -> {
                         if (producerFuture.completeExceptionally(exception)) {
@@ -2934,7 +2936,13 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         service.getTopicIfExists(topicName).thenAccept(topicOpt -> {
             if (topicOpt.isPresent()) {
                 Topic topic = topicOpt.get();
-                CompletableFuture<SchemaVersion> schemaVersionFuture = tryAddSchema(topic, schema);
+                boolean isReplicatorProducer = false;
+                if (commandGetOrCreateSchema.hasProducerName()) {
+                    isReplicatorProducer = Producer.isRemoteOrShadow(commandGetOrCreateSchema.getProducerName(),
+                            getBrokerService().getPulsar().getConfig().getReplicatorPrefix());
+                }
+                CompletableFuture<SchemaVersion> schemaVersionFuture =
+                        tryAddSchema(topic, schema, isReplicatorProducer);
                 schemaVersionFuture.exceptionally(ex -> {
                     ServerError errorCode = BrokerServiceException.getClientErrorCode(ex);
                     String message = ex.getMessage();
@@ -3419,9 +3427,10 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         });
     }
 
-    private CompletableFuture<SchemaVersion> tryAddSchema(Topic topic, SchemaData schema) {
+    private CompletableFuture<SchemaVersion> tryAddSchema(Topic topic, SchemaData schema,
+                                                          boolean isReplicatorProducer) {
         if (schema != null) {
-            return topic.addSchema(schema);
+            return topic.addSchema(schema, isReplicatorProducer);
         } else {
             return topic.hasSchema().thenCompose((hasSchema) -> {
                 log.debug()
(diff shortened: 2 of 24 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -2748,7 +2750,13 @@ protected void handleGetOrCreateSchema(CommandGetOrCreateSchema commandGetOrCrea
...
             if (topicOpt.isPresent()) {
                 Topic topic = topicOpt.get();
-                CompletableFuture<SchemaVersion> schemaVersionFuture = tryAddSchema(topic, schema);
+                boolean isReplicatorProducer = false;
+                if (commandGetOrCreateSchema.hasProducerName()) {
+                    isReplicatorProducer = Producer.isRemoteOrShadow(commandGetOrCreateSchema.getProducerName(),
+                            getBrokerService().getPulsar().getConfig().getReplicatorPrefix());
+                }
+                CompletableFuture<SchemaVersion> schemaVersionFuture =
+                        tryAddSchema(topic, schema, isReplicatorProducer);
```
codelipenghui, line 2759: Will it be a security issue that user can just set producer name to pulsar.repl.* to get rid of the schema upload control? if user disabled is_allow_auto_update_schema

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdNamespaces.java
```diff
@@ -2163,14 +2163,20 @@ private class SetIsAllowAutoUpdateSchema extends CliCommand {
         @Option(names = { "--disable", "-d" }, description = "Disable schema validation enforced")
         private boolean disable = false;
 
+        @Option(names = { "--disable-for-replicator" },
+                description = "By default, brokers always allow replicator to register new compatible schemas even"
+                + " when auto updates are disabled, if you want to disable replicators to register compatible schemas,"
+                + " please set it to true")
+        private boolean disableForReplicator;
+
```
codelipenghui, line 2171: --disable-for-replicator is a primitive boolean defaulting to false, so every set-is-allow-auto-update-schema call sends allowAutoUpdateSchemaWithReplicator=true unless the flag is supplied. A user who ran ... --disable --disable-for-replicator and later runs ... --enable silently re-enables replicator updates — different from the REST API's null-preserving semantics.

codelipenghui: Please also check the Java API

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/policies/data/Policies.java
```diff
@@ -218,6 +222,8 @@ public boolean equals(Object obj) {
                     && schema_validation_enforced == other.schema_validation_enforced
                     && schema_compatibility_strategy == other.schema_compatibility_strategy
                     && is_allow_auto_update_schema == other.is_allow_auto_update_schema
+                    && is_allow_auto_update_schema_with_replicator
+                        == other.is_allow_auto_update_schema_with_replicator
```
codelipenghui, line 226: Should use Objects.equals()?

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdNamespaces.java
```diff
@@ -2163,14 +2163,20 @@ private class SetIsAllowAutoUpdateSchema extends CliCommand {
         @Option(names = { "--disable", "-d" }, description = "Disable schema validation enforced")
         private boolean disable = false;
 
+        @Option(names = { "--disable-for-replicator" },
+                description = "By default, brokers always allow replicator to register new compatible schemas even"
+                + " when auto updates are disabled, if you want to disable replicators to register compatible schemas,"
+                + " please set it to true")
+        private boolean disableForReplicator;
```
codelipenghui, line 2170: PIP says CLI flag is --enable-for-replicator, implementation uses --disable-for-replicator.

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -2571,13 +2571,18 @@ protected void internalSetSchemaValidationEnforced(boolean schemaValidationEnfor
...
 
         mutatePolicy((policies) -> {
                     policies.is_allow_auto_update_schema = isAllowAutoUpdateSchema;
+                    if (isAllowAutoUpdateSchemaWithReplicator != null) {
+                        policies.is_allow_auto_update_schema_with_replicator =
+                                isAllowAutoUpdateSchemaWithReplicator;
+                    }
                     return policies;
                 }, (policies) -> policies.is_allow_auto_update_schema,
                 "isAllowAutoUpdateSchema");
```
codelipenghui, line 2588: Missed isAllowAutoUpdateSchemaWithReplicator?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
poorbarcode: Yes, it will.

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdNamespaces.java
poorbarcode: Fixed

pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/policies/data/Policies.java
poorbarcode: Good suggestion

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
poorbarcode: It has been modified at the line-2583

pulsar-client-tools/src/main/java/org/apache/pulsar/admin/cli/CmdNamespaces.java
poorbarcode: Fixed

Denovo1998:
@poorbarcode @codelipenghui 
Seems like there are some issues?

When geo-replicator creates a remote producer in the broker, it indeed carries the replicator's producer name and uses Schema.AUTO_PRODUCE_BYTES().

But the key point of AUTO_PRODUCE_BYTES is: when establishing a producer connection, it doesn't send the "specific schema of this message" all at once. The actual specific schema registration happens later during the message sending process, when encountering a new schema, it then triggers GET_OR_CREATE_SCHEMA. This call chain looks like this locally:

1. When replicator acts as a producer to establish a connection, the `ProducerImpl.connectionOpened(...)` method sends the PRODUCER command, which includes the producerName. But later, when it actually needs to register a specific schema, it follows the path `tryRegisterSchema(...) -> getOrCreateSchemaAsync(...)`.

2. The method finally calls Commands' newGetOrCreateSchema(requestId, topic, schemaInfo).

3. This command builder only wrote requestId/topic/schema, without writing producerName.

4. So the CommandGetOrCreateSchema sent from the sender to the broker always has the field producerName absent.

pulsar-common/src/main/java/org/apache/pulsar/common/protocol/Commands.java
```diff
@@ -1361,10 +1361,12 @@ public static ByteBuf newGetSchemaResponseError(long requestId, ServerError erro
         return serializeWithSize(newGetSchemaResponseErrorCommand(requestId, error, errorMessage));
     }
 
-    public static ByteBuf newGetOrCreateSchema(long requestId, String topic, SchemaInfo schemaInfo) {
+    public static ByteBuf newGetOrCreateSchema(long requestId, String topic, String producerName,
+                                               SchemaInfo schemaInfo) {
         BaseCommand cmd = localCmd(Type.GET_OR_CREATE_SCHEMA);
         Schema schema = cmd.setGetOrCreateSchema()
                 .setRequestId(requestId)
+                .setProducerName(producerName)
```
poorbarcode, line 1369: @Denovo1998 

>  Seems like there are some issues? https://github.com/apache/pulsar/pull/25461#issuecomment-4296557751

Yes, I forgot to modify the client-side code. I added the change now

poorbarcode: I will also add a test for the case then

Technoboy- (approved):

PR #25455, opened 2026-04-01
[feat] [pip] PIP-466: New Java Client API (V5) with Scalable Topic Support

merlimat:
## Summary

Introduce PIP-466 proposing a new Java client API (`pulsar-client-api-v5` / `pulsar-client-v5`) designed for scalable topics ([PIP-460](https://github.com/apache/pulsar/blob/master/pip/pip-460.md)).

Key points:
- Three purpose-built consumer types: StreamConsumer (ordered), QueueConsumer (parallel), CheckpointConsumer (framework connectors)
- Opaque topic model — partition/segment details hidden from applications
- Works with scalable topics, partitioned topics, and non-partitioned topics
- Existing `pulsar-client-api` remains unchanged
- Same API model will be adopted by non-Java SDKs

## Motivation

See [pip/pip-466.md](https://github.com/apache/pulsar/blob/master/pip/pip-466.md) for the full proposal.

----- 

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #25489, opened 2026-04-08
[feat] PIP-466: Add V5 client API for scalable topics

merlimat:
## Summary

Implementation of [PIP-466: New Java Client API (V5)](https://github.com/apache/pulsar/pull/25455) for scalable topic support.

- Introduces the `pulsar-client-api-v5` module with type-safe interfaces for scalable topics
- **Producer**: `Producer` / `ProducerBuilder` with key-based routing, flow control, and async variants
- **StreamConsumer**: exclusive segment assignment with key-ordered delivery, controller-coordinated rebalancing
- **QueueConsumer**: shared dispatch across all segments, no controller required
- **CheckpointConsumer**: externalized position tracking via serializable `Checkpoint` snapshots for stream processing frameworks (Flink, Beam)
- **PulsarClient / PulsarClientBuilder**: entry point with ServiceLoader-based provider discovery
- Async counterparts (`AsyncProducer`, `AsyncStreamConsumer`, `AsyncQueueConsumer`, `AsyncCheckpointConsumer`) for non-blocking usage
- Schema, auth, config, and internal provider packages
- Comprehensive usage examples in `Examples.java`

Related PIPs:
- [PIP-460: Scalable Topics](https://github.com/apache/pulsar/blob/master/pip/pip-460.md)
- [PIP-466: New Java Client API (V5)](https://github.com/apache/pulsar/pull/25455)

## Test plan
- [x] Module compiles as part of full Gradle build
- [ ] Integration tests with V5 client implementation (separate PR)


pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/async/AsyncQueueConsumer.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param messageId the ID of the message to acknowledge
+     */
+    void acknowledge(MessageId messageId);
+
+    /**
+     * Acknowledge a single message.
+     *
+     * @param message the message to acknowledge
+     */
+    void acknowledge(Message<T> message);
```
lhotari, line 56: Is there a need to have support for async acknowledgements? With the current Pulsar client, this is mainly useful with ack-receipts which can be useful in some rare use cases. The reason why this has come up in the past is that some other message broker support a mode where acknowledgements are acknowledged. I haven't used this myself, but I've heard users requesting this. Without async support, ack receipts skill the performance.

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/async/AsyncMessageBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface AsyncMessageBuilder<T> extends MessageMetadata<T, AsyncMessageBuilder<T>> {
+
+    /**
+     * Send the message asynchronously.
+     *
+     * @return a {@link CompletableFuture} that completes with the {@link MessageId} assigned
+     *         to the published message by the broker
+     */
+    CompletableFuture<MessageId> send();
```
lhotari, line 41: Just a random idea: Not sure if it's useful to make an interface design to prepare for LazyLog support where the producer wouldn't receive back the message id, but just an acknowledgement that the sending succeeded.

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * @param authPluginClassName the fully qualified class name of the authentication plugin
+     * @param authParams          the authentication parameters as key-value pairs
+     * @return an {@link Authentication} instance created from the specified plugin
+     * @throws PulsarClientException if the plugin class cannot be loaded or instantiated
+     */
+    public static Authentication create(String authPluginClassName, Map<String, String> authParams)
+            throws PulsarClientException {
+        return PulsarClientProvider.get().createAuthentication(authPluginClassName, authParams);
+    }
```
lhotari, line 92: These should be async methods since creating the authentication could involve IO. There's https://github.com/apache/pulsar/blob/master/pip/pip-97.md about this.

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/schema/Schema.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /**
+     * Get a Protobuf schema for a generated message class.
+     *
+     * @param <T>   the Protobuf message type
+     * @param clazz the Protobuf generated message class
+     * @return a {@link Schema} that encodes and decodes the Protobuf message
+     */
+    static <T extends com.google.protobuf.GeneratedMessageV3> Schema<T> protobuf(Class<T> clazz) {
```
lhotari: Should use `com.google.protobuf.Message` instead of `com.google.protobuf.GeneratedMessageV3`:
```suggestion
    static <T extends com.google.protobuf.Message> Schema<T> protobuf(Class<T> clazz) {
```

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/PulsarClient.java
```diff
@@ -0,0 +1,108 @@
...
+    <T> CheckpointConsumerBuilder<T> newCheckpointConsumer(Schema<T> schema);
+
+    // --- Transactions ---
+
+    /**
+     * Create a new transaction.
+     *
+     * @return a new {@link Transaction} in the {@link Transaction.State#OPEN} state
+     */
+    Transaction newTransaction();
```
lhotari: Does creating a transaction contain IO operations? if it does, shouldn't this return `CompletableFuture<Transaction>`?

lhotari: Would it be useful to return a `TransactionBuilder` instead, just in case creating a transaction contains configurable parameters later?

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/PulsarClient.java
```diff
@@ -0,0 +1,108 @@
...
+
+    // --- Lifecycle ---
+
+    /**
+     * Close the client and release all resources, waiting for pending operations to complete.
+     *
+     * @throws PulsarClientException if an error occurs while closing the client
+     */
+    @Override
+    void close() throws PulsarClientException;
```
lhotari, line 114: Would it be useful to also add closeAsync?

lhotari (commented):
I added a few review comments

lhotari (approved):
LGTM, minor comments. feel free to mark the comments resolved and proceed with merging after checking.

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/async/AsyncQueueConsumer.java
lhotari: discussed. No need for async ack at this point. We can add later if really needed.

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/auth/AuthenticationFactory.java
lhotari: Will be covered in a separate PIP

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/schema/Schema.java
merlimat: Fixed

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/PulsarClient.java
merlimat: Fixed. I didn't add a builder since the only option is the txn timeout which makes more sense to keep at client configuration (for consistency with all the other options).

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/PulsarClient.java
merlimat: Added

PR #25502, opened 2026-04-09
[fix][broker] Unthrottle producers immediately when publish rate limiting is disabled

thetumbled:
### Motivation

When publish rate limiting is disabled (for example by clearing the publish rate policy so update(null) is called), token buckets are nulled out immediately, but producers that were already throttled could remain blocked until a previously scheduled unthrottle task ran. That delayed task may have been scheduled with a very long wait time when the token bucket was deeply depleted, so disabling the limit in configuration did not restore traffic quickly. This change triggers the existing lock-free unthrottle path right away after disablement when an executor was already cached from a prior throttle on this limiter.

### Modifications

- In PublishRateLimiterImpl, cache the ScheduledExecutorService used when scheduling unthrottleQueuedProducers after the first throttling event on this limiter instance.
- When update(PublishRate) is called with a null PublishRate (rate limiting disabled), call scheduleUnthrottling(cachedExecutor, 0) so queued producers are processed without waiting on the old delayed task.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
@@ -43,6 +43,11 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
 
     private final AtomicInteger throttledProducersCount = new AtomicInteger(0);
     private final AtomicBoolean processingQueuedProducers = new AtomicBoolean(false);
+
+    /**
+     * Executor used for the last {@link #scheduleUnthrottling} from this limiter (set when throttling starts).
+     * Used to schedule an immediate follow-up run after publish-rate limits change.
+     */
     private volatile ScheduledExecutorService lastUnthrottleExecutor;
     private final Consumer<Producer> throttleAction;
     private final Consumer<Producer> unthrottleAction;
@@ -172,10 +177,8 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
     private void scheduleImmediateUnthrottling() {
         ScheduledExecutorService executor = lastUnthrottleExecutor;
         if (executor != null) {
-            // Wake the existing unthrottle path without waiting on an old delay.
             scheduleUnthrottling(executor, 0L);
         }
-        // If executor is null, no throttle has happened yet on this limiter
     }
 
     public void update(PublishRate maxPublishRate) {
@@ -199,6 +202,8 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
         } else {
             tokenBucketOnByte = null;
         }
+        // After any bucket rebuild, wake unthrottling:
+        // old scheduled delay may be invalid and cause unnecessary wait time for producers to be unthrottled.
         scheduleImmediateUnthrottling();
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
@@ -43,6 +43,7 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
 
     private final AtomicInteger throttledProducersCount = new AtomicInteger(0);
     private final AtomicBoolean processingQueuedProducers = new AtomicBoolean(false);
+    private volatile ScheduledExecutorService lastUnthrottleExecutor;
     private final Consumer<Producer> throttleAction;
     private final Consumer<Producer> unthrottleAction;
 
@@ -88,6 +89,7 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
         // this is to avoid scheduling unthrottling multiple times for concurrent producers
         if (throttledProducersCount.incrementAndGet() == 1) {
             ScheduledExecutorService executor = producer.getCnx().getBrokerService().executor().next();
+            lastUnthrottleExecutor = executor;
             scheduleUnthrottling(executor, calculateThrottlingDurationNanos());
         }
     }
@@ -173,6 +175,12 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
         } else {
             tokenBucketOnMessage = null;
             tokenBucketOnByte = null;
+            ScheduledExecutorService executor = lastUnthrottleExecutor;
+            if (executor != null) {
+                // Wake the existing unthrottle path without waiting on an old delay.
+                scheduleUnthrottling(executor, 0L);
+            }
+            // If executor is null, no throttle has happened yet on this limiter
         }
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
@@ -169,7 +169,7 @@ public class PublishRateLimiterImpl implements PublishRateLimiter {
         update(maxPublishRate);
     }
 
-    private void unthrottleImmediately() {
+    private void scheduleImmediateUnthrottling() {
         ScheduledExecutorService executor = lastUnthrottleExecutor;
         if (executor != null) {
             // Wake the existing unthrottle path without waiting on an old delay.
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
```diff
@@ -88,6 +89,7 @@ private void scheduleDecrementThrottleCount(Producer producer) {
         // this is to avoid scheduling unthrottling multiple times for concurrent producers
         if (throttledProducersCount.incrementAndGet() == 1) {
             ScheduledExecutorService executor = producer.getCnx().getBrokerService().executor().next();
```
nodece, line 96: You should provide (and reuse) a dedicated `Executor` when creating the rate limiter, for example:

```java
private final Executor executor;

public PublishRateLimiterImpl(MonotonicClock monotonicClock,
                              Consumer<Producer> throttleAction,
                              Consumer<Producer> unthrottleAction,
                              Executor executor) {
    this.monotonicClock = monotonicClock;
    this.throttleAction = throttleAction;
    this.unthrottleAction = unthrottleAction;
    this.executor = executor;
}
```

This ensures that all scheduling (e.g., unthrottling tasks) runs on a controlled and consistent execution context, rather than relying on ad-hoc or shared threads.

When publish rate limiting is disabled, you can immediately release any throttled producers by invoking:

```java
scheduleUnthrottling(executor, 0L);
```

Using a delay of `0L` guarantees that the unthrottling task is executed asynchronously but without delay, ensuring producers are promptly resumed without blocking the caller thread.

thetumbled: WDYT. @lhotari

liangyepianzhou:
<del>
The scenario is valid, but the fix needs further evaluation. The root cause is that disabling rate limiting should **directly signal** throttled producers to be released, rather than waiting for the next scheduled cycle — i.e., the `disable` semantic itself should trigger an immediate (synchronous or async) unthrottle.

The current fix, via `update(null)`, appears to work around the root cause by adding an operational workaround rather than addressing the underlying code defect. This may fail to resolve the fundamental issue, and worse, increases cognitive overhead for users who need to understand why an extra `update(null)` call is necessary.

</del>

Sorry, the code version I was looking at doesn't seem to be the latest.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
liangyepianzhou, line 185:
```suggestion
        if (maxPublishRate != null) {
        if (PublishRate.normalize(maxPublishRate) != null) {
```

liangyepianzhou: Based on the latest version of the code, this change appears to have a correctness issue. In the path where rate limiting is disabled, the value passed to the following method will never be null:

`org.apache.pulsar.broker.service.PublishRateLimiter#update(org.apache.pulsar.common.policies.data.PublishRate)`

Otherwise, the following method would throw an NPE:

`org.apache.pulsar.broker.service.AbstractTopic#updatePublishRateLimiter`.

Here it should fall back to the broker-level configuration, where the publishRate is never null.

lhotari:
> That delayed task may have been scheduled with a very long wait time when the token bucket was deeply depleted,

@thetumbled  did the problem happen also with the fix in #25262 and #25269?

thetumbled:
> > That delayed task may have been scheduled with a very long wait time when the token bucket was deeply depleted,
> 
> @thetumbled did the problem happen also with the fix in #25262 and #25269?

We haven't encountered this problem in our production environment yet. Since the bucket value is allowed to be negative, I am concerned that the producer may get stuck for a long time. We need a way to disable the throttling feature as quickly as possible.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
thetumbled: Good point. The latest commit handle this situation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
```diff
@@ -167,12 +169,22 @@ public void update(Policies policies, String clusterName) {
         update(maxPublishRate);
     }
 
+    private void unthrottleImmediately() {
```
lhotari: rename to `scheduleImmediateUnthrottling` so that it better conveys the meaning. In updateTokenBuckets, this method should be called whenever the rates change.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
```diff
@@ -189,6 +201,9 @@ protected void updateTokenBuckets(long publishThrottlingRateInMsg, long publishT
         } else {
             tokenBucketOnByte = null;
         }
+        if (tokenBucketOnByte == null && tokenBucketOnMessage == null) {
```
lhotari: remove the condition since either limit might be impacting the throttling. Let's say if it's throttled on bytes and just that limit gets increased, it impacts the throttling decision. `scheduleUnthrottling` will handle it correctly.
The call to `scheduleImmediateUnthrottling` (new name) should be made in every case. The modification in `update` method also relies on this.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
```diff
@@ -167,12 +169,22 @@ public void update(Policies policies, String clusterName) {
...
+        // If executor is null, no throttle has happened yet on this limiter
+    }
+
     public void update(PublishRate maxPublishRate) {
         if (maxPublishRate != null) {
             updateTokenBuckets(maxPublishRate.publishThrottlingRateInMsg, maxPublishRate.publishThrottlingRateInByte);
         } else {
             tokenBucketOnMessage = null;
             tokenBucketOnByte = null;
+            unthrottleImmediately();
```
lhotari: this should be changed to `updateTokenBuckets(0, 0);` to avoid code duplication.

lhotari (commented):
LGTM, a few comments to address.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
lhotari: The current solution picks any executor from the broker service since ordered execution isn't needed for this implementation. `unthrottleQueuedProducers` will return immediately if there's another thread processing the throttled producers. To avoid race conditions, it will schedule one more execution after processing. The purpose of this is to avoid synchronization on hot paths and also to avoid a single thread becoming a bottleneck.

lhotari: Therefore I think that the solution in the PR is fine. Perhaps a comment would clarify it.

liangyepianzhou:
Maybe a test should be added to verify that the fix actually works. Otherwise, there's no way to fully confirm whether issues similar to https://github.com/apache/pulsar/pull/25502#discussion_r3063424848 still exist — the scope of manual review is always limited.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PublishRateLimiterImpl.java
thetumbled: good point, I have updated the implementation.

thetumbled:
> Maybe a test should be added to verify that the fix actually works. Otherwise, there's no way to fully confirm whether issues similar to [#25502 (comment)](https://github.com/apache/pulsar/pull/25502#discussion_r3063424848) still exist — the scope of manual review is always limited.

I will add one later.

lhotari (approved):

thetumbled:
Tests have been added. You can comment out the call to `scheduleImmediateUnthrottling();` to verify the original behavior, and the test will output an error. @liangyepianzhou 
image

PR #25503, opened 2026-04-10
[improve][broker] update broker error message to include tenant name

pgier:
### Motivation

This is just to improve the log message to include the tenant name when there the tenant is not found

### Modifications

Just updated the log

### Verifying this change

PR for CI: https://github.com/pgier/pulsar/pull/13

No, it does not affect any of these.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -142,7 +142,7 @@ public abstract class NamespacesBase extends AdminResource {
                 .thenCompose(__ -> tenantResources().tenantExistsAsync(tenant))
                 .thenCompose(existed -> {
                     if (!existed) {
-                        throw new RestException(Status.NOT_FOUND, "Tenant not found: " + tenant);
+                        throw new RestException(Status.NOT_FOUND, "Tenant not found");
                     }
                     return tenantResources().getListOfNamespacesAsync(tenant);
                 });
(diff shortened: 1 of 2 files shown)


dlg99 (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -142,7 +142,7 @@ protected CompletableFuture<List<String>> internalGetTenantNamespaces(String ten
                 .thenCompose(__ -> tenantResources().tenantExistsAsync(tenant))
                 .thenCompose(existed -> {
                     if (!existed) {
-                        throw new RestException(Status.NOT_FOUND, "Tenant not found");
+                        throw new RestException(Status.NOT_FOUND, "Tenant not found: " + tenant);
```
merlimat: The tenant name is already included in the path of the request. I don't think it needs to be in the response too.

pgier: The issue is that the log on the server doesn't show which tenant has the issue.  I've updated the code so that it only prints the tenant name in the server log, WDYT?

merlimat (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
merlimat: 👍 This makes sense. 

My main concern for including in the response was that we have so many of these return error messages, and we should be consistent in what we return :)

PR #25506, opened 2026-04-10
[improve] PIP-467: Convert pulsar-common module logging from SLF4J to slog

merlimat:
## Summary
- Convert all 42 source and test files in `pulsar-common` from SLF4J to slog structured logging
- Replace `@Slf4j` with `@CustomLog` and convert log statements to slog fluent API with structured attributes
- Use `.exception(e)` for Throwable objects and lazy attrs where appropriate
- Remove Logger parameter from internal `FileUtils` methods (use class-level logger instead)

## Test plan
- [x] `compileJava` and `compileTestJava` pass
- [x] `checkstyleMain` and `checkstyleTest` pass
- [ ] CI passes

### Motivation

PIP-467

### Modifications

Convert pulsar-common module from SLF4J to slog structured logging.

> [!NOTE]
> Please label this PR with `ready-to-test`

### DIFF
--- a/distribution/shell/src/assemble/LICENSE.bin.txt
+++ b/distribution/shell/src/assemble/LICENSE.bin.txt
@@ -391,6 +391,8 @@ The Apache Software License, Version 2.0
     - opentelemetry-api-incubator-1.56.0-alpha.jar
     - opentelemetry-common-1.56.0.jar
     - opentelemetry-context-1.56.0.jar
+  * Slog
+    - slog-0.9.5.jar
 
  * BookKeeper
     - bookkeeper-common-allocator-4.17.3.jar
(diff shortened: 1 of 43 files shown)


lhotari:
btw. Should we also add the `.git-blame-ignore-revs` file for the large refactorings? It would have to be added after merging to master branch since the commit hash gets created at that point. 
Locally it would require `git config blame.ignoreRevsFile .git-blame-ignore-revs`. 
GitHub supports it by default: [docs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view), [announcement](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view)

lhotari (approved):
LGTM

merlimat:
> btw. Should we also add the .git-blame-ignore-revs file for the large refactorings? It would have to be added after merging to master branch since the commit hash gets created at that point.

For as pure formatting change sure, though here there could potentially be bugs deriving from this commit, so it might be good to not hide them.

PR #25509, opened 2026-04-11
[fix][test] Fix flaky ExtensibleLoadManagerImpl client reconnection tests: PulsarClientException$AlreadyClosedException: Client already closed

void-ptr974:
The tests were failing intermittently with:
    PulsarClientException$AlreadyClosedException: Client already closed : state = Closed

  Root cause: 4 PulsarClient instances were shared across all test methods
  via @BeforeClass. Once any client entered the Closed state (e.g. due to
  a previous test timeout or incomplete cleanup of Mockito-spied lookup
  services), all subsequent tests using that client would fail.

  Fix: each test method now creates and closes its own PulsarClient
  instances, eliminating shared mutable state between tests.


void-ptr974:
https://github.com/apache/pulsar/actions/runs/24268154576/job/70867532721?pr=25507

merlimat (approved):

PR #18701, opened 2022-12-01
[fix][broker] Fix duplicated schemas creation

codelipenghui:
Fixes: #18292
Another solution of #18293

### Motivation

Fix the duplicated schemas creation due to race conditions.
The broker will get all the schemas to check whether or not the given schema exists.
And then put the schema into the schema storage once the schema doesn't exist.
But If there are parallel requests with the same schema. 
The Schema registry will put duplicated same schemas into the
schema storage.

For each schema update, we need to update the schema locator.
The PR uses the Zookeeper data version control to ensure
the schema locator will not be updated based on outdated data.
And will retry to check if schema exists and compatibility while
encountering bad version exceptions.

And the created Ledger with failed locator updating will be removed
to avoid the Ledger leak issue.

### Modifications

Add a new put schema API, which can accept a function to check if the
schema exists or is compatible.

### Verifying this change

A new test was added for creating producers with schema in parallel

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #25514, opened 2026-04-13
[fix][broker] Clean up orphan ledger on concurrent initial schema creation in BookkeeperSchemaStorage

### LINKED ISSUE #18292: [Bug] Too much schemas ledgers are created when multi producer start concurrently
### Search before asking

### Version

2.8.1

### Minimal reproduce step

1. create a new topic
2. start up multi producers with the same Schema
3. after producers are created, check the schemas by `bin/pulsar-admin schemas get -a tenant/namespace/topic`
4. you'll see many versions of schemas

All producers with the same schema will only create one ledger to persistent schema entry.

### What did you see instead?

Many duplicated ledgers for schema are created. like:

```SHELL
$ bin/pulsar-admin schemas get t_schema/ns_schema/t13 -a
{
  "name" : "t13",
  "schema" : "",
  "type" : "STRING",
  "timestamp" : 1666852962358,
  "properties" : { },
  "schemaDefinition" : ""
}
{
  "name" : "t13",
  "schema" : "",
  "type" : "STRING",
  "timestamp" : 1666852962365,
  "properties" : { },
  "schemaDefinition" : ""
}
{
  "name" : "t13",
  "schema" : "",
  "type" : "STRING",
  "timestamp" : 1666852962369,
  "properties" : { },
  "schemaDefinition" : ""
}
{
  "name" : "t13",
  "schema" : "",
  "type" : "STRING",
  "timestamp" : 1666852962365,
  "properties" : { },
  "schemaDefinition" : ""
}
{
  "name" : "t13",
  "schema" : "",
  "type" : "STRING",
  "timestamp" : 1666852962368,
  "properties" : { },
  "schemaDefinition" : ""
}
{
  "name" : "t13",
  "schema" : "",
  "type" : "STRING",
  "timestamp" : 1666852962368,
  "properties" : { },
  "schemaDefinition" : ""
}


```

### Anything else?

This problem increases many read and writes requests to zookeeper, which may make the pulsar cluster in an unstable state.
As we saw, there are more than 80 thousand ledgers are created when starting 1600 producers at the same time(in one minutes).

```shell
$grep 'Creating producer' pulsar-broker.log | grep '19:43:' | wc -l
1632

>>
19:43:26.994 [pulsar-io-4-102] INFO org.apache.pulsar.broker.service.ServerCnx - [][persistent://checker/checker/checker-partition-53] Creating producer. producerId=53
19:43:26.994 [pulsar-io-4-102] INFO org.apache.pulsar.broker.service.ServerCnx - [][persistent://checker/checker/checker-partition-41] Creating producer. producerId=41
19:43:26.994 [pulsar-io-4-102] INFO org.apache.pulsar.broker.service.ServerCnx - [][persistent://checker/checker/checker-partition-39] Creating producer. producerId=39
19:43:26.994 [pulsar-io-4-102] INFO org.apache.pulsar.broker.service.ServerCnx - [][persistent://checker/checker/checker-partition-34] Creating producer. producerId=34
19:43:26.994 [pulsar-io-4-102] INFO org.apache.pulsar.broker.service.ServerCnx - [][persistent://checker/checker/checker-partition-33] Creating producer. producerId=33
19:43:26.995 [pulsar-io-4-102] INFO org.apache.pulsar.broker.service.ServerCnx - [][persistent://checker/checker/checker-partition-29] Creating producer. producerId=29


$ grep 'LedgerCreateOp' pulsar-broker.log| grep '19:43:' | wc -l
87112

>>
19:43:59.986 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524473
19:43:59.986 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524474
19:43:59.986 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524475
19:43:59.986 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524476
19:43:59.986 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524479
19:43:59.986 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524480
19:43:59.993 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524556
19:43:59.993 [main-EventThread] INFO  org.apache.bookkeeper.client.LedgerCreateOp - Ensemble: [xxx] for ledger: 524558
```

This means the broker create 87111 ledgers unexpectedly, these ledgers (also including the znode for ledger and schema) will keep existing until we delete schema forcefully. 

geniusjoe:
Fixes #18292
Related #18701

### Motivation

When multiple requests concurrently create a schema for a brand-new topic (i.e., the schema locator z-node does not yet exist), each request first creates a BookKeeper ledger via `addNewSchemaEntryToStore`, then attempts to create the schema locator z-node via CAS (`createSchemaLocator` with `expectedVersion = -1L`). 

Only one request succeeds; the others fail with BadVersionException (because ZK MetadataStore translates NODEEXISTS to BadVersionException when expectedVersion == -1). However, the ledgers created by the failed requests were never cleaned up, resulting in orphan ledgers (dirty data) in BookKeeper.

Note that the existing `updateSchemaLocator` method already has cleanup logic for this scenario (deleting the orphan ledger when CAS fails with `BadVersionException`) in #18701, but the `createNewSchema` method — which handles the **initial** schema creation path — was missing this cleanup.

### Modifications

- **`BookkeeperSchemaStorage.createNewSchema`**: Added a `whenComplete` callback after `createSchemaLocator`. When the CAS operation fails due to `AlreadyExistsException` or `BadVersionException`, the orphan BookKeeper ledger is asynchronously deleted, consistent with the existing cleanup logic in `updateSchemaLocator`. The method is also refactored into three clearly commented steps for better readability.

- **`PulsarMockLedgerHandle`**: Added a new constructor that accepts `Map<String, byte[]> customMetadata` and passes it to `LedgerMetadataBuilder.withCustomMetadata()`, so that mock ledgers can retain custom metadata set during creation.

- **`PulsarMockBookKeeper`**: Updated `asyncCreateLedger` to forward the `properties` parameter to the new `PulsarMockLedgerHandle` constructor, enabling tests to inspect ledger custom metadata.

- **`SchemaTest`**: Added `testConcurrentCreateSchemaNoOrphanLedger` test that verifies orphan ledgers are cleaned up when 16 producers concurrently create schema on a brand-new topic. The test inspects surviving BK ledgers via `PulsarMockBookKeeper.getLedgerMap()` and asserts that only 1 ledger with matching `pulsar/schemaId` custom metadata exists.

### Verifying this change

- Added `testConcurrentCreateSchemaNoOrphanLedger` in `SchemaTest` that concurrently creates 16 producers with the same AVRO schema on a brand-new topic, then verifies:
  1. Only 1 schema version exists via `admin.schemas().getAllSchemas()`
  2. Only 1 surviving BK ledger has `customMetadata["pulsar/schemaId"]` matching the topic's schema name (orphan ledgers from failed concurrent creations were deleted)

- [ ] The public API
- [x] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
@@ -1380,6 +1380,68 @@ public class SchemaTest extends MockedPulsarServiceBaseTest {
         producers2.clear();
     }
 
+    /**
+     * Test that when multiple producers concurrently create schema for a brand-new topic,
+     * the orphan BookKeeper ledgers created by the losing requests are properly cleaned up.
+     */
+    @Test
+    public void testConcurrentCreateSchemaNoOrphanLedger() throws Exception {
+        final String namespace = "test-namespace-" + randomName(16);
+        String ns = PUBLIC_TENANT + "/" + namespace;
+        admin.namespaces().createNamespace(ns, Sets.newHashSet(CLUSTER_NAME));
+
+        final String topic = getTopicName(ns, "testConcurrentCreateSchemaNoOrphanLedger");
+        final String schemaName = TopicName.get(topic).getSchemaName();
+
+        org.apache.bookkeeper.client.PulsarMockBookKeeper mockBk =
+                (org.apache.bookkeeper.client.PulsarMockBookKeeper) pulsar.getBookKeeperClient();
+
+        // Concurrently create producers with the same schema on a brand-new topic
+        int concurrency = 16;
+        @Cleanup("shutdownNow")
+        ExecutorService executor = Executors.newFixedThreadPool(concurrency);
+        List<CompletableFuture<Producer<Schemas.PersonOne>>> producers = new ArrayList<>(concurrency);
+        CountDownLatch latch = new CountDownLatch(concurrency);
+        for (int i = 0; i < concurrency; i++) {
+            executor.execute(() -> {
+                producers.add(pulsarClient.newProducer(Schema.AVRO(Schemas.PersonOne.class))
+                        .topic(topic).createAsync());
+                latch.countDown();
+            });
+        }
+        latch.await();
+        FutureUtil.waitForAll(producers).join();
+
+        // Verify only 1 schema version exists
+        assertEquals(admin.schemas().getAllSchemas(topic).size(), 1);
+
+        // Count surviving BK ledgers whose customMetadata "pulsar/schemaId" matches this topic's schemaName.
+        // If orphan ledgers were not cleaned up, there would be more than 1.
+        int schemaLedgerCount = 0;
+        for (org.apache.bookkeeper.client.PulsarMockLedgerHandle lh
+                : mockBk.getLedgerMap().values()) {
+            Map<String, byte[]> metadata = lh.getLedgerMetadata().getCustomMetadata();
+            byte[] schemaIdBytes = metadata.get("pulsar/schemaId");
+            if (schemaIdBytes != null
+                    && schemaName.equals(
+                            new String(schemaIdBytes, java.nio.charset.StandardCharsets.UTF_8))) {
+                schemaLedgerCount++;
+            }
+        }
+        assertEquals(schemaLedgerCount, 1,
+                "Expected exactly 1 schema ledger for the topic, but found "
+                        + schemaLedgerCount + ". Orphan ledgers were not cleaned up.");
+
+        // Cleanup
+        producers.forEach(p -> {
+            try {
+                p.join().close();
+            } catch (Exception ignore) {
+            }
+        });
+        producers.clear();
+    }
+
     @EqualsAndHashCode
     static class User implements Serializable {
         private String name;
--- a/pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
@@ -55,6 +55,7 @@ import lombok.EqualsAndHashCode;
 import org.apache.avro.Schema.Parser;
 import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.client.BookKeeper;
+import org.apache.bookkeeper.mledger.impl.LedgerMetadataUtils;
 import org.apache.pulsar.broker.BrokerTestUtil;
 import org.apache.pulsar.broker.auth.MockedPulsarServiceBaseTest;
 import org.apache.pulsar.broker.service.schema.BookkeeperSchemaStorage;
@@ -1400,13 +1401,17 @@ public class SchemaTest extends MockedPulsarServiceBaseTest {
         int concurrency = 16;
         @Cleanup("shutdownNow")
         ExecutorService executor = Executors.newFixedThreadPool(concurrency);
-        List<CompletableFuture<Producer<Schemas.PersonOne>>> producers = new ArrayList<>(concurrency);
+        List<CompletableFuture<Producer<Schemas.PersonOne>>> producers =
+                Collections.synchronizedList(new ArrayList<>(concurrency));
         CountDownLatch latch = new CountDownLatch(concurrency);
         for (int i = 0; i < concurrency; i++) {
             executor.execute(() -> {
-                producers.add(pulsarClient.newProducer(Schema.AVRO(Schemas.PersonOne.class))
-                        .topic(topic).createAsync());
-                latch.countDown();
+                try {
+                    producers.add(pulsarClient.newProducer(Schema.AVRO(Schemas.PersonOne.class))
+                            .topic(topic).createAsync());
+                } finally {
+                    latch.countDown();
+                }
             });
         }
         latch.await();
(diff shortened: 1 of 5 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
```diff
@@ -1376,6 +1376,68 @@ public void testCreateSchemaInParallel() throws Exception {
...
+        // Verify only 1 schema version exists
+        assertEquals(admin.schemas().getAllSchemas(topic).size(), 1);
+
+        // Count surviving BK ledgers whose customMetadata "pulsar/schemaId" matches this topic's schemaName.
+        // If orphan ledgers were not cleaned up, there would be more than 1.
+        int schemaLedgerCount = 0;
+        for (org.apache.bookkeeper.client.PulsarMockLedgerHandle lh
+                : mockBk.getLedgerMap().values()) {
+            Map<String, byte[]> metadata = lh.getLedgerMetadata().getCustomMetadata();
+            byte[] schemaIdBytes = metadata.get("pulsar/schemaId");
```
codelipenghui: Replace with LedgerMetadataUtils.METADATA_PROPERTY_SCHEMAID

pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
```diff
@@ -1376,6 +1376,68 @@ public void testCreateSchemaInParallel() throws Exception {
...
+        final String schemaName = TopicName.get(topic).getSchemaName();
+
+        org.apache.bookkeeper.client.PulsarMockBookKeeper mockBk =
+                (org.apache.bookkeeper.client.PulsarMockBookKeeper) pulsar.getBookKeeperClient();
+
+        // Concurrently create producers with the same schema on a brand-new topic
+        int concurrency = 16;
+        @Cleanup("shutdownNow")
+        ExecutorService executor = Executors.newFixedThreadPool(concurrency);
+        List<CompletableFuture<Producer<Schemas.PersonOne>>> producers = new ArrayList<>(concurrency);
```
codelipenghui: Use Collections.synchronizedList(...)

pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
```diff
@@ -1376,6 +1376,68 @@ public void testCreateSchemaInParallel() throws Exception {
...
+        CountDownLatch latch = new CountDownLatch(concurrency);
+        for (int i = 0; i < concurrency; i++) {
+            executor.execute(() -> {
+                producers.add(pulsarClient.newProducer(Schema.AVRO(Schemas.PersonOne.class))
+                        .topic(topic).createAsync());
+                latch.countDown();
```
codelipenghui: Need to execute `latch.countDown();` in finally block to avoid any exceptions cause the test stuck.

pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
geniusjoe: Fixed

pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
geniusjoe: Fixed

pulsar-broker/src/test/java/org/apache/pulsar/schema/SchemaTest.java
geniusjoe: Fixed

codelipenghui (approved):

dao-jun (approved):
LGTM

hanmz (approved):
lgtm

Denovo1998:
@geniusjoe @dao-jun @codelipenghui @hanmz 

Seems like there are some issues?

```
Request A and Request B create a schema for a new topic simultaneously.

T1: A finds the locator does not exist.

T2: B also finds the locator does not exist.

T3: A creates BK ledger L1 and writes the schema entry.

T4: B creates BK ledger L2 and writes the schema entry.

T5: A's createSchemaLocator is successful, and the locator points to L1.

T6: B's createSchemaLocator fails, receiving a BadVersionException / AlreadyExistsException.

T7: B's whenComplete calls asyncDeleteLedger(L2).

T8: asyncDeleteLedger only asynchronously commits deletion and returns immediately.

T9: B's createNewSchema future still completes with a BadVersionException.

T10: The upper BookkeeperSchemaStorage#put catches the BadVersionException and immediately retries.

T11: The retry re-reads the locator, finds that the schema already exists, and the request succeeds and returns.

T12: But the deletion callback of L2 may not have finished executing.
```

`whenComplete` only waits for this callback to return, but the callback starts a callback-based `bookKeeper.asyncDeleteLedger(...)` and returns before the ledger deletion is confirmed. As a result, the outer `put(...)` can observe the original `BadVersionException` / `AlreadyExistsException`, retry immediately, and complete the schema creation before the orphan ledger is actually removed. This makes the new `getLedgerMap()` assertion timing-sensitive, and also weakens the cleanup guarantee exposed to callers. Could we wrap `asyncDeleteLedger` in a `CompletableFuture` and compose it before rethrowing the original CAS exception so the existing retry happens after the delete callback completes?

geniusjoe:
@Denovo1998 
Thank you very much for your review. I initially considered removing the redundant schema ledger as an asynchronous task to avoid blocking the main schema creation logic. 

However, the current approach could indeed introduce race condition issues in unit tests. I think your proposed solution makes sense. If you have time, you could go ahead and implement it as you suggested. Perhaps the cleanup semantics for both `createNewSchema` and `updateSchemaLocator` need to be updated consistently.

PR #25515, opened 2026-04-14
[improve] Add PULSAR_LOG_FORMAT env var for easy JSON logging

merlimat:
### Motivation

Pulsar already includes a JSON-capable layout (`JsonTemplateLayout` with the ECS template) in `conf/log4j2.yaml`, but enabling it requires knowing internal Log4j2 system properties. This is undiscoverable for users who want JSON logs in containerized environments (k8s, Docker) where log aggregators like ELK, Loki, or Datadog parse JSON.

This PR adds a single, discoverable environment variable — `PULSAR_LOG_FORMAT` — that switches both console and file output to the JSON (ECS) layout.

### Modifications

- **`conf/log4j2.yaml`**: introduces a new `pulsar.log.format` property (default `text`, or `json`) and uses Log4j2's `SystemPropertyArbiter` to select the layout at config-load time. The arbiter is applied in three places:
  - The `Console` appender
  - The top-level `RollingFile` appender
  - The per-function `RollingFile` nested inside the `RoutingAppender`

  Arbiters are evaluated once at configuration load, so only one layout is included — there is a single appender instance and no duplicate file handles, competing rollover policies, or risk of mixed text + JSON output.

- **`bin/pulsar`**: defaults `PULSAR_LOG_FORMAT` to `text` and forwards it as `-Dpulsar.log.format=...`. No appender-name switching.

- **`conf/pulsar_env.sh`**: adds a documented commented-out example for `PULSAR_LOG_FORMAT`.

- **slog 0.9.5 → 0.9.7**: the previous slog version triggered a NPE in `JsonTemplateLayout` because `MutableLogEvent.clear()` left the context stack null and Log4j2's `ThreadContextStackResolver.isResolvable()` does not null-check it. slog 0.9.7 sets the context stack to `EMPTY_STACK` after clearing the event. License files for both the server and shell distributions are updated.

#### Usage

```bash
# Foreground (k8s, Docker stdout)
PULSAR_LOG_FORMAT=json bin/pulsar standalone

# Daemon mode (rolling JSON files)
PULSAR_LOG_FORMAT=json bin/pulsar-daemon start broker
```

Without `PULSAR_LOG_FORMAT`, behavior is unchanged.

### Verifying this change

This change is a trivial rework / code cleanup without any test coverage — it only wires existing Log4j2 appenders behind a `SystemPropertyArbiter` and bumps a dependency.

Manually verified:
- `PULSAR_LOG_FORMAT=json bin/pulsar standalone` produces one ECS-formatted JSON object per log line on stdout.
- Default behavior (no env var) is unchanged — `PatternLayout` output is identical to master.
- The slog 0.9.7 bump resolves the NPE that previously surfaced when JSON logging was enabled.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

Bumps the `slog` dependency from 0.9.5 to 0.9.7 (fixes the NPE in `JsonTemplateLayout`). The new `PULSAR_LOG_FORMAT` env var is opt-in and does not change any default behavior.


dao-jun (approved):

conf/log4j2.yaml
```diff
@@ -66,27 +66,48 @@ Configuration:
...
+            size: 1 GB
+        # Delete file older than 30days
+        DefaultRolloverStrategy:
+            Delete:
+              basePath: ${sys:pulsar.log.dir}
+              maxDepth: 2
+              IfFileName:
+                glob: "${sys:pulsar.log.file}*log.gz"
+              IfLastModified:
+                age: 30d
```
lhotari: Claude suggested using a different type of solution:

Log4j2 2.19+ introduced Arbiter — compile-time conditional blocks evaluated once at configuration load. SystemPropertyArbiter checks a system property and includes/excludes config sections accordingly:

```yaml
name: RollingFile
    fileName: "${sys:pulsar.log.dir}/${sys:pulsar.log.file}"
    filePattern: "${sys:pulsar.log.dir}/${sys:pulsar.log.file}-%d{MM-dd-yyyy}-%i.log.gz"
    immediateFlush: ${sys:pulsar.log.immediateFlush}
    SystemPropertyArbiter:
      - propertyName: pulsar.log.format
        propertyValue: json
        JsonTemplateLayout:
          eventTemplateUri: ${sys:pulsar.log.console.json.template}
      - propertyName: pulsar.log.format
        propertyValue: text
        PatternLayout:
          Pattern: "%d{ISO8601_OFFSET_DATE_TIME_HHMM} [%t] %-5level %logger{36} - %msg %X%n"
```

The concern about the solution was this:

Both RollingFile and RollingFileJson appenders point to the same fileName and filePattern. Log4j2
  initializes all appenders defined in the config at startup, regardless of whether they're referenced by any logger. This means
  both appenders will attempt to open the same file. Log4j2 uses a FileManager keyed by file path, so both appenders will share the
  same underlying FileManager instance — they won't conflict on file locks. However, there are subtle concerns:

  1. Both appenders are initialized — even the unused one opens the file and registers its rollover policies. The unused
  TimeBasedTriggeringPolicy and SizeBasedTriggeringPolicy will still be active on the shared FileManager, potentially triggering
  unnecessary rollovers.
  2. Mixed format risk — if someone misconfigures and both appenders end up active (e.g., one via RoutingAppender default and
  another via pulsar.log.appender), the same file would contain mixed text + JSON lines.

lhotari: The SystemPropertyArbiter solution is also provided as an example in the docs: https://logging.apache.org/log4j/2.x/manual/configuration.html#arbiters . file example: https://github.com/apache/logging-log4j2/blob/rel/2.25.3/src/site/antora/modules/ROOT/examples/manual/configuration/arbiters.yaml#L22-L29

conf/log4j2.yaml
lhotari: for consistency, it would be useful to choose json format using the SystemPropertyArbiter solution also in this case.

lhotari: this should also be changed to use the SystemPropertyArbiter solution

conf/log4j2.yaml
```diff
@@ -40,6 +40,8 @@ Configuration:
         value: "info"
       - name: "pulsar.routing.appender.default"
         value: "Console"
+      - name: "pulsar.log.format"
+        value: "text"
       - name: "pulsar.log.console.json.template"
```
lhotari: could we also rename this to pulsar.log.json.template and also update the environment variable name in `bin/pulsar`? There could be backwards compatibility in the `bin/pulsar` script for the environment variable name.

lhotari (approved):

PR #25516, opened 2026-04-14
[feat] PIP-468: Scalable Topic Controller

merlimat:
## Summary

Adds PIP-468, a sub-PIP of [PIP-460: Scalable Topics](https://github.com/apache/pulsar/blob/master/pip/pip-460.md), describing the broker-side **Scalable Topic Controller** — the component responsible for segment lifecycle management, split/merge coordination, consumer assignment, leader election, and DAG watch sessions for scalable topics.

The document covers:

- **Architecture** — the four-layer design (Admin API → ScalableTopicService → ScalableTopicController → ScalableTopicResources) and the three broker roles (any broker for DAG watch sessions, controller leader for layout mutations and consumer assignment, segment-owning brokers for produce/consume).
- **Split and merge protocols** — the step-by-step create-first / update-metadata-last ordering that guarantees no messages are lost and all subscription cursors exist before producers are redirected, plus cross-broker coordination via the Segments admin API.
- **Consumer assignment** — StreamConsumer / CheckpointConsumer receive coordinated assignments from the controller leader; QueueConsumer attaches to all segments with no controller-side assignment. Includes a detailed **Consumer Session Lifecycle** section clarifying what is persisted (the `ConsumerRegistration`) vs. what is in-memory on the leader (the keep-alive state and grace-period timer), and how controller leader failover restarts grace-period timers from zero.
- **Leader election, failover, and metadata store schema** — including the new `SubscriptionMetadata` and `ConsumerRegistration` paths.
- **Public-facing changes** — new binary protocol commands (`CommandScalableTopicLookup` / `Update` / `Close`), new Admin REST endpoints for scalable topics and segments, new `ScalableTopics` admin client interface, and the `scalableTopicEnabled` broker configuration.

This PIP only changes documentation under `pip/`. Implementation will land in subsequent PRs against the parent PIP-460 tracking issue.

## Test plan
- [ ] Review the document on the PIP mailing list thread (TBD link to be added once thread is open).

### DIFF
--- a/pip/pip-468.md
+++ b/pip/pip-468.md
@@ -1,0 +1,615 @@
+# PIP-468: Scalable Topic Controller
+
+*Sub-PIP of [PIP-460: Scalable Topics](pip-460.md)*
+
+## Motivation
+
+[PIP-460](pip-460.md) introduces scalable topics as a new topic type in Pulsar, built on a DAG of range segments that can be split and merged to scale independently of the initial configuration. PIP-460 defines the overall vision but defers the detailed design of each component to dedicated sub-PIPs.
+
+This PIP specifies the **Scalable Topic Controller** — the broker-side component responsible for:
+
+1. **Segment lifecycle management** — creating, terminating, and deleting the underlying segment topics that back a scalable topic.
+2. **Segment layout coordination** — executing split and merge operations with correct ordering guarantees so that no messages are lost and all subscription cursors exist before producers are redirected.
+3. **Consumer assignment** — coordinating which consumers receive messages from which segments for a given subscription.
+4. **Leader election** — ensuring exactly one broker acts as the controller for each scalable topic, with automatic failover.
+5. **DAG watch sessions** — pushing topology updates to connected clients (producers and consumers) when the segment layout changes.
+
+Without a well-defined controller, split and merge operations would be unsafe: producers could write to segments that have no subscription cursors, consumers could miss messages during topology changes, and concurrent layout modifications could corrupt the segment DAG.
+
+---
+
+## Design
+
+### Architecture Overview
+
+The controller subsystem consists of four layers:
+
+```
+┌──────────────────────────────────────────────────────────┐
+│                    Admin API Layer                        │
+│         ScalableTopics REST  +  Segments REST             │
+└──────────────┬───────────────────────┬───────────────────┘
+               │                       │
+┌──────────────▼───────────────────────▼───────────────────┐
+│                  ScalableTopicService                     │
+│         Per-broker singleton; manages controllers         │
+└──────────────────────────┬───────────────────────────────┘
+                           │
+┌──────────────────────────▼───────────────────────────────┐
+│              ScalableTopicController (per topic)          │
+│   Leader-elected; owns layout, split/merge, consumers     │
+│                                                           │
+│   ┌─────────────────┐  ┌───────────────────────────────┐ │
+│   │  SegmentLayout   │  │  SubscriptionCoordinator (×N) │ │
+│   │  (immutable DAG) │  │  per-subscription assignments │ │
+│   └─────────────────┘  └───────────────────────────────┘ │
+└──────────────────────────┬───────────────────────────────┘
+                           │
+┌──────────────────────────▼───────────────────────────────┐
+│              ScalableTopicResources                       │
+│         Metadata store access (read/write/watch)          │
+└──────────────────────────────────────────────────────────┘
+```
+
+### Broker Roles and Request Routing
+
+A scalable topic involves three distinct broker roles. Understanding which broker handles which operation is critical to the design.
+
+#### Any broker — DAG watch sessions
+
+A **DAG watch session** (scalable topic lookup) can be served by **any broker** in the cluster. The session reads segment metadata from the shared metadata store and registers a watch for changes. No state is held on the broker beyond the watch registration. This means producers and consumers can connect to any broker for topic discovery — the same way regular topic lookups work today.
+
+When the metadata changes (due to a split or merge performed by the controller leader), the metadata store notification fires on whichever broker is serving the watch session, and that broker pushes the updated DAG to the client.
+
+#### Controller leader — layout mutations and consumer assignment
+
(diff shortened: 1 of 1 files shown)


pip/pip-468.md
```diff
@@ -0,0 +1,615 @@
...
+
+Segment topics are managed exclusively through a dedicated REST API at `/admin/v2/segments`. This is a deliberate design choice: segment topics use the `segment://` domain and must not be confused with regular `persistent://` topics.
+
+### Endpoints
+
+| Method | Path | Description |
+|--------|------|-------------|
+| `PUT` | `/{tenant}/{namespace}/{topic}/{descriptor}` | Create a segment topic. Optional request body with subscription names to pre-create. |
+| `POST` | `/{tenant}/{namespace}/{topic}/{descriptor}/terminate` | Terminate (seal) a segment topic. |
+| `DELETE` | `/{tenant}/{namespace}/{topic}/{descriptor}` | Delete a segment topic. |
```
lhotari, line 341: are these APIs intended to be used by other external client than other brokers? if external clients aren't supposed to use these APIs, we should consider creating an internal API endpoint on a separate port which would be internal only to the cluster.
When the internal API is separated, it wouldn't end up in the public OpenAPI specs and there wouldn't be a risk that it gets misused.

pip/pip-468.md
```diff
@@ -0,0 +1,615 @@
...
+| `nextSegmentId` | `long` | Counter for assigning IDs to new segments |
+| `allSegments` | `Map<Long, SegmentInfo>` | Complete DAG: all segments (active + sealed) |
+| `activeSegments` | `Map<Long, SegmentInfo>` | Filtered view: only `ACTIVE` segments |
+
+**Key operations:**
+
+- `findActiveSegment(hash)` — returns the active segment whose hash range contains the given hash value. Used by producers for message routing.
+- `splitSegment(segmentId)` — validates the segment is active, computes the midpoint of its hash range, creates two child `SegmentInfo` records covering `[start, mid]` and `[mid+1, end]`, seals the parent, and returns a new layout with incremented epoch.
+- `mergeSegments(id1, id2)` — validates both segments are active and adjacent (the end of one equals `start - 1` of the other), creates a merged `SegmentInfo` covering the combined range, seals both parents, and returns a new layout.
+- `getLineage(segmentId)` — returns the full ancestor + descendant chain for a segment, used for DAG traversal during catch-up reads.
```
lhotari, line 155: In a catch up read, there's a need to take the subscription into account. Where would this be handled?

lhotari (commented):
some initial comments

pip/pip-468.md
lhotari: discussed. we can keep these available on the current API endpoint.

pip/pip-468.md
lhotari: discussed. this is covered in future pips.

lhotari (approved):
LGTM

liangyepianzhou:
I have a question: if my ScalableTopic has a large number of sub-partitions — say, 1 million — would there be a situation where every broker needs to perform calculations across all 1 million sub-partitions of that ScalableTopic just to determine which sub-partitions belong to it? I'm a bit worried about scalability issues caused by this kind of metadata computation.

merlimat:
> I have a question: if my ScalableTopic has a large number of sub-partitions — say, 1 million — would there be a situation where every broker needs to perform calculations across all 1 million sub-partitions of that ScalableTopic just to determine which sub-partitions belong to it? I'm a bit worried about scalability issues caused by this kind of metadata computation.

The number of "segments" is dynamically adjusted based on several factors (eg: current/historic traffic, consumer parallelisms), though there would never be a reason to have such a large number of segments. Most likely scenario is for segments to be at most few 100s, and in some exceptional cases to touch 1000s.

The reverse can be true though: eg. having millions of scalable topics (with most of the having few or just one segment)

liangyepianzhou:
> > I have a question: if my ScalableTopic has a large number of sub-partitions — say, 1 million — would there be a situation where every broker needs to perform calculations across all 1 million sub-partitions of that ScalableTopic just to determine which sub-partitions belong to it? I'm a bit worried about scalability issues caused by this kind of metadata computation.
> 
> The number of "segments" is dynamically adjusted based on several factors (eg: current/historic traffic, consumer parallelisms), though there would never be a reason to have such a large number of segments. Most likely scenario is for segments to be at most few 100s, and in some exceptional cases to touch 1000s.
> 
> The reverse can be true though: eg. having millions of scalable topics (with most of the having few or just one segment)

 In scenarios where users require strict ordering and very slow consumption speeds, and considering that k-shared mode may cause out-of-order delivery upon consumer disconnection — or in cases where a proxy is inserted between the broker and client, making k-shared unusable — the only option is to increase the number of partitions to boost consumption concurrency. This could result in a single ScalableTopic having millions of sub-partitions.

In such scenarios, can the new design avoid the scalability issues caused by metadata computation?

merlimat:
> > > I have a question: if my ScalableTopic has a large number of sub-partitions — say, 1 million — would there be a situation where every broker needs to perform calculations across all 1 million sub-partitions of that ScalableTopic just to determine which sub-partitions belong to it? I'm a bit worried about scalability issues caused by this kind of metadata computation.
> > 
> > 
> > The number of "segments" is dynamically adjusted based on several factors (eg: current/historic traffic, consumer parallelisms), though there would never be a reason to have such a large number of segments. Most likely scenario is for segments to be at most few 100s, and in some exceptional cases to touch 1000s.
> > The reverse can be true though: eg. having millions of scalable topics (with most of the having few or just one segment)
> 
> In scenarios where users require strict ordering and very slow consumption speeds, and considering that k-shared mode may cause out-of-order delivery upon consumer disconnection — or in cases where a proxy is inserted between the broker and client, making k-shared unusable — the only option is to increase the number of partitions to boost consumption concurrency. This could result in a single ScalableTopic having millions of sub-partitions.
> 
> In such scenarios, can the new design avoid the scalability issues caused by metadata computation?

Scaling up the number of segments (or partitions for what it matters) for this scenario is not an efficient way of solving the problem. 

Scalable topics are going to address these scenarios with the same mechanism used today for KeyShared subscriptions.

PR #25518, opened 2026-04-14
[improve][broker] Use full bundle name for namespace bundle destination affinity in ModularLoadManagerImpl

Demogorgon314:
### Motivation

When unloading a namespace bundle with `destinationBroker` under `ModularLoadManagerImpl`, the affinity key was stored by `bundleRange` only.

Different namespaces can share the same bundle range, especially for single-bundle namespaces such as `0x00000000_0xffffffff`. In that case, the destination affinity can collide across namespaces, and the next lookup may
consume the affinity for the wrong bundle.

As a result, `unload --destinationBroker` may appear ineffective or unstable when transferring multiple bundles that share the same range.

### Modifications

- Store namespace bundle destination affinity by full bundle name instead of bare `bundleRange`.
- Consume the destination affinity in `ModularLoadManagerWrapper#getLeastLoaded` by full bundle name as well.
- Add an admin-side test to verify the affinity key passed to the load manager is the full bundle name.
- Add a load manager test to verify affinity lookup is resolved by full bundle name.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -1395,9 +1395,15 @@ public abstract class NamespacesBase extends AdminResource {
         if (StringUtils.isBlank(destinationBroker)) {
             return CompletableFuture.completedFuture(null);
         }
-        String bundleName = pulsar().getNamespaceService().getNamespaceBundleFactory()
-                .getBundle(namespaceName.toString(), bundleRange)
-                .toString();
+        final String bundleName;
+        try {
+            bundleName = pulsar().getNamespaceService().getNamespaceBundleFactory()
+                    .getBundle(namespaceName.toString(), bundleRange)
+                    .toString();
+        } catch (RuntimeException e) {
+            return FutureUtil.failedFuture(e);
+        }
+
         return pulsar().getLoadManager().get().getAvailableBrokersAsync()
                 .thenCompose(brokers -> {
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
```diff
@@ -1395,6 +1395,15 @@ public CompletableFuture<Void> setNamespaceBundleAffinityAsync(String bundleRang
...
             return CompletableFuture.completedFuture(null);
         }
+        final String bundleName;
+        try {
+            bundleName = pulsar().getNamespaceService().getNamespaceBundleFactory()
+                    .getBundle(namespaceName.toString(), bundleRange)
+                    .toString();
+        } catch (RuntimeException e) {
+            return FutureUtil.failedFuture(e);
+        }
```
BewareMyPower: No need to catch unchecked exception here.

This method is called in `thenCompose`, any uncaught exception will be reflected in the final returned future.

BewareMyPower (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
Demogorgon314: But if some one called `setNamespaceBundleAffinityAsync` on out of `thenCompose` it will cause unexpect behavior.

merlimat (approved):

PR #20026, opened 2023-04-06
[improve][broker] Prevent range conflicts with Key Shared sticky consumers when TCP/IP connections get orphaned

lhotari:
### Motivation

In Kubernetes upgrades, it's possible to get into a state where a TCP/IP connection seems to be alive on the broker. The connection is orphaned and therefore consuming resources and causing resource conflicts with producers and consumers. 
This PR intends to target the issue with consumers that are using the Key shared subscription type with sticky hash ranges. The prevents the issue of "Range conflict with consumer ..." which happens when the previous connection from the client is still active on the broker while the client is reconnecting on a new connection. 

### Modifications

- make `Dispatch.addConsumer` and `StickyKeyConsumerSelector.addConsumer` asynchronous
- add connectionLivenessCheckTimeoutMillis setting that defaults to 5000ms
- when a conflicting consumer is found, first do an active check with Pulsar's Ping command to see if the connection is alive.
- resume the attempt to add the consumer after the check completes

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

PR #15382, opened 2022-04-29
[Broker/Client] Close connection if a ping message cannot be sent

lhotari:
### Motivation

The keep alive implementation used in the Broker and the Java client ignores any failures that happen while writing the Ping message.
In Netty 4+, it's necessary to add a listener to detect failures when writing to outbound channels. 
The connection should be closed if a ping message cannot be sent. The future is notified once the OS level TCP/IP stack accepts the message for delivery.

This change will help detect the situation where writing to the outbound channel returns EBADF errno from the OS level and the connection isn't yet completely closed.
EBADF errno maps to java.nio.channels.ClosedChannelException which has this kind of explanation:
> Checked exception thrown when an attempt is made to invoke or complete an I/O operation upon channel that is closed, or at least closed to that operation. That this exception is thrown does not necessarily imply that the channel is completely closed. A socket channel whose write half has been shut down, for example, may still be open for reading.

### Modifications

Add a listener that closes the connection if sending the Ping message fails.

PR #25520, opened 2026-04-14
[improve][broker] Close connection when close consumer write fails

### LINKED ISSUE #25146: [Bug] Partial partition consumers silent after topic fencing
### Search before reporting

### Read release policy

### User environment

- Observed version: Pulsar 2.10.1 (unsupported)
- Likely affected: All Pulsar versions, including 3.x, 4.x

### Issue Description

When a topic is fenced, some partition consumers may silently stop receiving messages, while others continue working normally.

It only occurs under specific timing and failure conditions, which makes it hard to detect and debug:

- The broker sometimes fails to deliver the close command to certain consumers (e.g. due to transient network issues, broker overload, or connection state races).
- Consumers rely entirely on broker-initiated notifications and do not actively verify fencing state after subscription is established.

Once this happens, the affected partition consumer enters a permanent silent failure state.

Symptoms
- ✅ TCP connection remains alive
- ✅ No exceptions or error logs
- ❌ Some partitions stop receiving messages indefinitely

Meanwhile, other partitions of the same consumer continue to work normally, which further hides the issue.

### Error messages

```text

```

### Reproducing the issue

1. Create a partitioned topic and subscribe a consumer.
2. Trigger topic fencing via topic unload.
3. Publish messages to the topic.
4. Observe that some partitions stop receiving messages while others continue normally, with no errors or logs.

This issue is not 100% reproducible.

### Additional information

This topic is partitioned, which has one producer and one consumer.

- partition-2 stats post unload
   
Image

- partition-2 internal-stats post unload
  
Image

- grafana
  
Image

Around 13:00, an observation at the partition level of a specific topic revealed a discrepancy in data flow. While partition-2 had active producers with a non-zero production rate, the consumption rate for that specific partition was 0.

However, when observing at the topic level, the overall consumption rate remained non-zero. This indicates that while the topic appears "healthy" in aggregate, the partition-2 is in a silent or stalled state.

nodece:
Fix #25146

### Motivation

When the broker asks a client to close a consumer, if the `CloseConsumer` command does not reach the client, the client may keep the consumer alive locally while the broker has already removed it, leading to a silent inconsistent state.

### Modifications

- update broker-side consumer close handling to close the whole connection when sending `CloseConsumer` fails
- keep the existing protocol behavior for older clients that already fall back to closing the connection
- add a broker test covering the failed `CloseConsumer` write path

### Verifying this change

- Added a `ServerCnxTest` case for failed `CloseConsumer` writes to ensure the connection is closed

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -3405,14 +3405,25 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         if (getRemoteEndpointProtocolVersion() >= v5.getValue()) {
             assignedBrokerLookupData.ifPresentOrElse(lookup -> {
                         LookupData lookupData = getLookupData(lookup);
-                        writeAndFlush(Commands.newCloseConsumer(consumerId, -1L,
+                        writeCloseConsumerAndCloseConnectionOnFailure(Commands.newCloseConsumer(consumerId, -1L,
                                 lookupData.getBrokerUrl(),
-                                lookupData.getBrokerUrlTls()));
+                                lookupData.getBrokerUrlTls()), consumerId);
                     },
-                    () -> writeAndFlush(Commands.newCloseConsumer(consumerId, -1L, null, null)));
+                    () -> writeCloseConsumerAndCloseConnectionOnFailure(
+                            Commands.newCloseConsumer(consumerId, -1L, null, null), consumerId));
         } else {
             close();
         }
+    }
+
+    private void writeCloseConsumerAndCloseConnectionOnFailure(ByteBuf cmd, long consumerId) {
+        ctx.writeAndFlush(cmd).addListener(future -> {
+            if (!future.isSuccess()) {
+                log.warn("[{}] Forcing connection to close since cannot send close consumer command for consumer {}",
+                        remoteAddress, consumerId, future.cause());
+                close();
+            }
+        });
     }
 
     /**
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ServerCnxTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/ServerCnxTest.java
@@ -45,10 +45,14 @@ import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertTrue;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.DefaultChannelId;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
 import io.vertx.core.impl.ConcurrentHashSet;
 import java.io.Closeable;
 import java.io.IOException;
@@ -2486,6 +2490,35 @@ public class ServerCnxTest {
         assertTrue(channel.isActive());
 
         channel.finish();
+    }
+
+    @Test
+    public void testCloseConsumerClosesConnectionWhenWriteFails() throws Exception {
+        resetChannel();
+        setChannelConnected();
+
+        var ctx = mock(ChannelHandlerContext.class);
+        var writeFuture = mock(ChannelFuture.class);
+        var writeFailure = new RuntimeException("close consumer write failed");
+        when(ctx.writeAndFlush(any())).thenReturn(writeFuture);
+        when(writeFuture.isSuccess()).thenReturn(false);
+        when(writeFuture.cause()).thenReturn(writeFailure);
+        when(writeFuture.addListener(any())).thenAnswer(invocation -> {
+            GenericFutureListener<Future<? super Void>>  listener = invocation.getArgument(0);
+            listener.operationComplete(writeFuture);
+            return writeFuture;
+        });
+        serverCnx.setCtx(ctx);
+
+        var consumer = mock(Consumer.class);
+        when(consumer.consumerId()).thenReturn(1L);
+
+        serverCnx.setRemoteEndpointProtocolVersion(ProtocolVersion.v12.getValue());
+
+        serverCnx.closeConsumer(consumer, Optional.empty());
+
+        verify(ctx).writeAndFlush(any());
+        verify(ctx).close();
     }
 
     @Test(timeOut = 30000)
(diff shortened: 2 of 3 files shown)


merlimat (approved):

lhotari (approved):

lhotari:
> When the broker asks a client to close a consumer, if the `CloseConsumer` command does not reach the client, the client may keep the consumer alive locally while the broker has already removed it, leading to a silent inconsistent state.

The change looks ok, but I have doubts whether there could be a scenario where the command doesn't reach the client and there's no other error / exception in the logs. Any unhandled exception would result in the connection being closed by [this code in exceptionCaught](https://github.com/apache/pulsar/blob/daa5f8db6ce4befeda0b71b6e1d4ae73f4d56c11/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L535-L553).

lhotari:
> > When the broker asks a client to close a consumer, if the `CloseConsumer` command does not reach the client, the client may keep the consumer alive locally while the broker has already removed it, leading to a silent inconsistent state.
> 
> The change looks ok, but I have doubts whether there could be a scenario where the command doesn't reach the client and there's no other error / exception in the logs. Any unhandled exception would result in the connection being closed by [this code in exceptionCaught](https://github.com/apache/pulsar/blob/daa5f8db6ce4befeda0b71b6e1d4ae73f4d56c11/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L535-L553).

Sorry for the noise. Yes, this PR helps. https://github.com/apache/pulsar/pull/15382 was a similar change to Ping/Pong.

lhotari:
>  the client may keep the consumer alive locally while the broker has already removed it

Btw. there are also other cases where the consumer might be alive on the client side (until the next Ping command) which are caused by different reason than what is handled by this PR. They happen especially in Kubernetes environments. One of the unresolved issues in Kubernetes is https://github.com/kubernetes/kubernetes/issues/104098 (this was an issue that Michael was digging into a few years ago) which is challenging. IIRC, the main scenario was rebooting a k8s node. That resulted in problems in many Pulsar clients and stale connections. The problem at that time with k8s node reboots was mitigated with https://github.com/apache/pulsar/pull/20026. The previous improvement was  https://github.com/apache/pulsar/pull/15382, but that has the latency of the keep alive interval.

A long list of related issues in Kubernetes (list compiled with Claude):
> When a Kubernetes node is rebooted or a pod is rescheduled, TCP connections can become stale because kube-proxy doesn't always clean up conntrack entries for TCP connections. When a node becomes NotReady, kube-proxy can delete conntrack entries for UDP, but doesn't delete conntrack entries for TCP [GitHub](https://github.com/kubernetes/kubernetes/issues/104098) ([kubernetes/kubernetes#104098](https://github.com/kubernetes/kubernetes/issues/104098)). After a node reboot, stale conntrack entries remain, and the external traffic can arrive at the node before kube-proxy has set up the NAT entries [GitHub](https://github.com/kubernetes/kubernetes/issues/118814) ([kubernetes/kubernetes#101607](https://github.com/kubernetes/kubernetes/issues/101607)). There's also [kubernetes/kubernetes#118814](https://github.com/kubernetes/kubernetes/issues/118814) which covers the same issue with IPVS mode. Similarly, when a pod backing a service is restarted and gets a new IP, all existing TCP connections can just stall [GitHub](https://github.com/kubernetes/kubernetes/issues/124290) instead of receiving a RST ([kubernetes/kubernetes#124290](https://github.com/kubernetes/kubernetes/issues/124290)). And [kubernetes/kubernetes#100698](https://github.com/kubernetes/kubernetes/issues/100698) documents how established TCP sessions to a leader pod don't get disconnected when the service endpoints change — the conntrack rules for those sessions are not cleared, which is a serious problem for leader/follower setups with long-lived connections [GitHub](https://github.com/kubernetes/kubernetes/issues/100698).

Many of the issues are simply not easy to resolve as explained in the discussion that Michael had on the issue (for example [comment](https://github.com/kubernetes/kubernetes/issues/104098#issuecomment-1518521697) and the following one are at the gist of the problem). There's also a reference https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/ in the issue comments.

lhotari:
Some additional details from Claude:
> Cilium's Socket-LB, when used as a kube-proxy replacement, mitigates many of the stale connection and conntrack problems that affect kube-proxy in both iptables and IPVS modes. Because Socket-LB performs service-to-backend translation at the socket layer (at connect() time for TCP) rather than per-packet NAT, it avoids creating conntrack entries for service translation entirely. This eliminates the class of problems where stale conntrack entries cause connections to be black-holed or forwarded to dead backends after node reboots, pod restarts, or endpoint changes.
Additionally, because Cilium tracks which sockets are connected to which backends in BPF maps, it can proactively destroy sockets connected to removed backends — using either the kernel's SOCK_DESTROY API (via netlink, requiring CONFIG_INET_DIAG_DESTROY) or the newer bpf_sock_destroy() kfunc (kernel 6.5+, being adopted in Cilium 1.17+). This is something neither kube-proxy iptables mode nor IPVS mode can do, since they operate at the packet level and have no visibility into application sockets.
That said, Socket-LB doesn't cover all traffic paths — north-south traffic (external to cluster) and some scenarios like hostNamespaceOnly mode still use eBPF-based per-packet NAT at the tc/XDP layer, which is closer to what kube-proxy does (though still using BPF maps rather than iptables/IPVS). The conntrack-related issues are most effectively mitigated for east-west pod-to-service traffic where Socket-LB is fully in play.

It seems that there's a recommendation to replace kube-proxy with Cilium. This option seems to be available also on all major cloud provider managed k8s services (EKS, GKE and AKS). The default option is kube-proxy in most cases.

Claude's comments:

> eBPF-based kube-proxy replacements like Cilium (and Calico in eBPF mode) are available on all major managed Kubernetes services — EKS, GKE, and AKS. However, the level of integration and support varies:
> 
> - **GKE** offers "GKE Dataplane V2" which is Cilium under the hood, provided as a first-party Google option. When you select Dataplane V2 at cluster creation, you get Cilium with kube-proxy replacement managed by Google.
> 
> - **AKS** offers "Azure CNI Powered by Cilium" as a first-party managed option. When you create a cluster with `--network-dataplane cilium`, AKS deploys Cilium as the data plane and does not use kube-proxy at all. AKS manages the Cilium configuration — customers who need more control can use AKS BYOCNI and install Cilium manually instead. AKS also supports a BYOCNI mode for fully self-managed Cilium installations.
> 
> - **EKS** supports Cilium either as a full CNI replacement (replacing the default AWS VPC CNI) or in "chaining" mode alongside the VPC CNI. Replacing kube-proxy with Cilium on EKS is well-documented and supported, but it's a bring-your-own setup — AWS doesn't manage Cilium for you.
> 
> The default on EKS remains kube-proxy with the AWS VPC CNI. On GKE and AKS, Cilium-based dataplanes are available as first-party options at cluster creation time, though kube-proxy-based configurations are still the default unless explicitly selected otherwise.
> 
> It's worth noting that "recommendation" may be too strong — it's more accurate to say that eBPF-based kube-proxy replacement is increasingly recognized as a best practice for clusters at scale or clusters that need better connection lifecycle management. For smaller clusters, kube-proxy works fine and is simpler to operate.
> 
> Cilium isn't the only option either — Calico's eBPF mode also provides connect-time load balancing and can replace kube-proxy. The broader trend is toward eBPF-based alternatives for service networking, with Cilium being the most prominent.

nodece:
@lhotari Thanks for your share, this is useful for me!

PR #25521, opened 2026-04-14
[fix][sec] Upgrade log4j to 2.25.4 to address CVE-2026-34477, CVE-2026-34478, CVE-2026-34480, CVE-2026-34481

merlimat:
### Motivation

log4j 2.25.3 is affected by four CVEs fixed in 2.25.4:

- **CVE-2026-34477** — `verifyHostName` attribute in `<Ssl>` configuration was silently ignored in all versions since 2.12.0. TLS connections from SMTP, Socket, and Syslog appenders configured via the nested `<Ssl>` element were not verifying hostnames, enabling MITM attacks. The fix for CVE-2025-68161 was incomplete and only addressed the `log4j2.sslVerifyHostName` system property path.
- **CVE-2026-34478** — `Rfc5424Layout` silently renamed `newLineEscape` and `useTlsMessageFormat` attributes in 2.21.0, causing CRLF injection for TCP framing users and a silent TLS-to-plain-TCP downgrade for RFC 5425 users.
- **CVE-2026-34480** — `XmlLayout` did not sanitize characters forbidden by XML 1.0, producing invalid XML when log messages or MDC values contained them. Depending on the StAX implementation, this either emits malformed XML or throws during the logging call and drops the event.
- **CVE-2026-34481** — `JsonTemplateLayout` produced invalid JSON (NaN / Infinity / -Infinity) for `MapMessage` entries containing non-finite floats, which downstream log processors reject per RFC 8259.

None of these affect Pulsar users with the default log4j2.yaml configuration, but shipping library versions with known vulnerabilities trips code scanning and downstream security audits.

### Modifications

- Bump `log4j2` version in `gradle/libs.versions.toml` from 2.25.3 to 2.25.4
- Update server and shell distribution LICENSE.bin.txt to reflect the new jar versions


utafrali (approved):
Clean, minimal security patch that bumps log4j2 from 2.25.3 to 2.25.4 and keeps the LICENSE.bin.txt files in sync. The changes are correct, well-motivated, and follow the project's standard dependency management pattern.

dao-jun (approved):

lhotari (approved):

PR #25525, opened 2026-04-15
[improve][client] Best-effort retry for individual/batch-index acks on send failure when ackReceiptEnabled=false

nodece:
### Motivation

When `ackReceiptEnabled=false`, grouped consumer acks are sent in the background and local write failures could drop pending ack state. This increases the risk of ack holes that can block progress in cases like `Key_Shared` consumption.

### Modifications

- Retain pending grouped acks for best-effort retry when send fails and `ackReceiptEnabled=false` — applies to **individual** and **batch-index** acknowledgements only
- Skip cumulative ack restore on send failure — cumulative acks cover unbounded ranges and restoring them after a partial send could re-ack already-processed messages, so retry is intentionally excluded
- Keep `ackReceiptEnabled=true` on explicit future/ack-receipt failure semantics instead of internal pending restore
- Add coverage for both no-receipt retry behavior and ack-receipt failure behavior

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java
@@ -1080,29 +1080,25 @@ public class ClientCnx extends PulsarHandler {
     }
 
     private <T> void sendRequestAndHandleTimeout(ByteBuf requestMessage, long requestId,
-                                                                 RequestType requestType, boolean flush,
-                                                                 TimedCompletableFuture<T> future) {
+                                                 RequestType requestType, boolean flush,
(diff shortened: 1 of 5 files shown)


dao-jun (approved):
lgtm

Denovo1998:
The final patch no longer restores cumulative grouped acknowledgements on local write failure, making the behavior a best-effort retry for only individual and batch-index acknowledgements. If this change is intentional, could we update the PR title and body accordingly? Otherwise, this branch still loses the retry opportunity for grouped cumulative acknowledgements when `ackReceiptEnabled=false`.

Denovo1998 (approved):

nodece:
@Denovo1998 Both the title and body have been updated. We only retry for individual ack when `ackReceiptEnabled=false`.

PR #25538, opened 2026-04-16
[improve][client] Implement tls_client_auth for AuthenticationOAuth2

izumo27:
### Motivation

Currently, the Client Credentials Flow in AuthenticationOAuth2 supports only authentication using `client_secret`.
This PR adds `tls_client_auth` using a certificate.
https://datatracker.ietf.org/doc/rfc8705/

### Modifications
Added `TlsClientAuthFlow` class for the flow using a certificate.
- The `client_id` is treated as an optional parameter for users.
  - Although `client_id` is required by the RFC, it may not be necessary in cases such as Athenz. Allowing it to be optional as a client input provides greater flexibility. As long as the pulsar client includes `client_id` internally, it remains compliant with the RFC.
  - https://github.com/apache/pulsar/pull/25538/changes#diff-979a00edf42b23b8e525310f703a0ac1203538ed742f907bfd658af53f8b49f2R109
- ~Since a certificate is only required when obtaining a token, a different HTTP client is used than the one used for retrieving metadata.~
  - ~https://github.com/apache/pulsar/pull/25538/changes#diff-979a00edf42b23b8e525310f703a0ac1203538ed742f907bfd658af53f8b49f2R141-R146~

### Verifying this change

  - https://github.com/izumo27/pulsar/pull/7

- Added tests for tls_client_auth
- Modify existing tests for cilent_sercre

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
@@ -21,9 +21,10 @@ package org.apache.pulsar.client.impl.auth.oauth2;
 import java.net.URL;
 import java.time.Duration;
 import java.util.concurrent.ScheduledExecutorService;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.pulsar.client.api.Authentication;
-import org.apache.pulsar.client.impl.auth.oauth2.protocol.TokenEndpointAuthMethod;
 import org.apache.pulsar.client.impl.auth.oauth2.protocol.DefaultMetadataResolver;
+import org.apache.pulsar.client.impl.auth.oauth2.protocol.TokenEndpointAuthMethod;
 
 /**
  * Factory class that allows to create {@link Authentication} instances
@@ -111,6 +112,18 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
+         * Optional token endpoint auth method.
+         * Defaults to {@code client_secret_post}.
+         *
+         * @param tokenEndpointAuthMethod the token endpoint auth method
+         * @return the builder
+         */
+        public ClientCredentialsBuilder tokenEndpointAuthMethod(TokenEndpointAuthMethod tokenEndpointAuthMethod) {
+            this.tokenEndpointAuthMethod = tokenEndpointAuthMethod;
+            return this;
+        }
+
+        /**
          * Required issuer URL.
          *
          * @param issuerUrl the issuer URL
@@ -122,7 +135,8 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
-         * Required (when using client_secret) credentials URL.
+         * Required credentials URL.
+         * Only used by {@code client_secret_post}
          *
          * @param credentialsUrl the credentials URL
          * @return the builder
@@ -133,7 +147,8 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
-         * Required (when using tls_client_auth) path to the file for a client certificate.
+         * Optional path to the file for a client certificate.
+         * Required when {@code tokenEndpointAuthMethod} is {@code tls_client_auth}
          *
          * @param tlsCertFile the path to the file for a client certificate
          * @return the builder
@@ -144,7 +159,8 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
-         * Required (when using tls_client_auth) path to the file for a client private key.
+         * Optional path to the file for a client private key.
+         * Required when {@code tokenEndpointAuthMethod} is {@code tls_client_auth}
          *
          * @param tlsKeyFile the path to the file for a client private key
          * @return the builder
@@ -156,7 +172,7 @@ public final class AuthenticationFactoryOAuth2 {
 
         /**
          * Optional client identifier issued by the authorization server.
-         * This parameter can only be set when tls_client_auth.
+         * Only used by {@code tls_client_auth}.
          * Defaults to {@code pulsar-client} when not provided.
          *
          * @param clientId the client identifier
@@ -164,18 +180,6 @@ public final class AuthenticationFactoryOAuth2 {
          */
         public ClientCredentialsBuilder clientId(String clientId) {
             this.clientId = clientId;
-            return this;
-        }
-
-        /**
-         * Optional token endpoint auth method.
-         * Defaults to {@code client_secret_post}.
-         *
-         * @param tokenEndpointAuthMethod the token endpoint auth method
-         * @return the builder
-         */
-        public ClientCredentialsBuilder tokenEndpointAuthMethod(TokenEndpointAuthMethod tokenEndpointAuthMethod) {
-            this.tokenEndpointAuthMethod = tokenEndpointAuthMethod;
             return this;
         }
 
@@ -250,8 +254,7 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
-         * Optional Certificate refresh interval in seconds
-         * This parameter can only be set when tls_client_auth.
+         * Optional certificate refresh interval.
          *
          * @param autoCertRefreshDuration the Certificate refresh interval
          * @return the builder
@@ -311,9 +314,15 @@ public final class AuthenticationFactoryOAuth2 {
                         .connectTimeout(connectTimeout)
                         .readTimeout(readTimeout)
                         .trustCertsFilePath(trustCertsFilePath)
+                        .certFile(tlsCertFile)
+                        .keyFile(tlsKeyFile)
+                        .autoCertRefreshDuration(autoCertRefreshDuration)
                         .wellKnownMetadataPath(wellKnownMetadataPath)
                         .build();
             } else {
+                if (StringUtils.isBlank(tlsCertFile) || StringUtils.isBlank(tlsKeyFile)) {
+                    throw new IllegalArgumentException("Required configuration parameters: tlsCertFile, tlsKeyFile");
+                }
                 flow = TlsClientAuthFlow.builder()
                         .issuerUrl(issuerUrl)
                         .clientId(clientId)
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
@@ -21,6 +21,7 @@ package org.apache.pulsar.client.impl.auth.oauth2;
 import java.net.URL;
 import java.time.Duration;
 import java.util.concurrent.ScheduledExecutorService;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.pulsar.client.api.Authentication;
 import org.apache.pulsar.client.impl.auth.oauth2.protocol.DefaultMetadataResolver;
 
@@ -92,12 +93,16 @@ public final class AuthenticationFactoryOAuth2 {
 
         private URL issuerUrl;
         private URL credentialsUrl;
+        private String clientId;
+        private String tlsCertFile;
+        private String tlsKeyFile;
         private String audience;
         private String scope;
         private Duration connectTimeout;
         private Duration readTimeout;
         private String trustCertsFilePath;
         private String wellKnownMetadataPath;
+        private Duration autoCertRefreshDuration;
         private double earlyTokenRefreshPercent = AuthenticationOAuth2.EARLY_TOKEN_REFRESH_PERCENT_DEFAULT;
         private ScheduledExecutorService scheduler;
 
@@ -116,13 +121,48 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
-         * Required credentials URL.
+         * Required (when using client_secret) credentials URL.
          *
          * @param credentialsUrl the credentials URL
          * @return the builder
          */
         public ClientCredentialsBuilder credentialsUrl(URL credentialsUrl) {
             this.credentialsUrl = credentialsUrl;
+            return this;
+        }
+
+        /**
+         * Required (when using tls_client_auth) path to the file for a client certificate.
+         *
+         * @param tlsCertFile the path to the file for a client certificate
+         * @return the builder
+         */
+        public ClientCredentialsBuilder tlsCertFile(String tlsCertFile) {
+            this.tlsCertFile = tlsCertFile;
+            return this;
+        }
+
+        /**
+         * Required (when using tls_client_auth) path to the file for a client private key.
+         *
+         * @param tlsKeyFile the path to the file for a client private key
+         * @return the builder
+         */
+        public ClientCredentialsBuilder tlsKeyFile(String tlsKeyFile) {
+            this.tlsKeyFile = tlsKeyFile;
+            return this;
+        }
+
+        /**
+         * Optional client identifier issued by the authorization server.
+         * This parameter can only be set when tls_client_auth.
+         * Defaults to {@code pulsar-client} when not provided.
+         *
+         * @param clientId the client identifier
+         * @return the builder
+         */
+        public ClientCredentialsBuilder clientId(String clientId) {
+            this.clientId = clientId;
             return this;
         }
 
@@ -197,6 +237,18 @@ public final class AuthenticationFactoryOAuth2 {
         }
 
         /**
+         * Optional Certificate refresh interval in seconds
+         * This parameter can only be set when tls_client_auth.
+         *
+         * @param autoCertRefreshDuration the Certificate refresh interval
+         * @return the builder
+         */
+        public ClientCredentialsBuilder autoCertRefreshDuration(Duration autoCertRefreshDuration) {
+            this.autoCertRefreshDuration = autoCertRefreshDuration;
+            return this;
+        }
+
+        /**
          * The fraction of the token's {@code expires_in} time at which the client starts attempting
          * a background refresh. Must be greater than 0. Values &ge; 1 disable early refresh (the default).
          *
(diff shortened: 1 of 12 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2.java
```diff
@@ -158,7 +158,14 @@ public void configure(String encodedAuthParamString) {
         Map<String, String> params = parseAuthParameters(encodedAuthParamString);
         String type = params.getOrDefault(CONFIG_PARAM_TYPE, TYPE_CLIENT_CREDENTIALS);
         if (TYPE_CLIENT_CREDENTIALS.equals(type)) {
-            this.flow = ClientCredentialsFlow.fromParameters(params);
+            if (params.containsKey(TlsClientAuthFlow.CONFIG_PARAM_CERT_FILE)
+                && params.containsKey(TlsClientAuthFlow.CONFIG_PARAM_KEY_FILE)
+               && StringUtils.isNotBlank(params.get(TlsClientAuthFlow.CONFIG_PARAM_CERT_FILE))
+               && StringUtils.isNotBlank(params.get(TlsClientAuthFlow.CONFIG_PARAM_KEY_FILE))) {
```
lhotari, line 164: Would it make sense to add a separate parameter `auth_method` (or `token_endpoint_auth_method`) which takes `tls_client_auth` or `client_secret_basic` (default) for this purpose? It makes it possible to also add validation if a required setting for `tls_client_auth` is missing. 
Having a `auth_method` parameter would also make it easier to add more methods later.
It looks like there's also `self_signed_tls_client_auth`.

Is it valid to provide a client certificate although `client_secret_basic` is used?
That could be useful if there's a reverse proxy that validates that client's identity, but the Authorization Server uses `client_secret_basic` for authentication.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
```diff
@@ -196,6 +236,18 @@ public ClientCredentialsBuilder wellKnownMetadataPath(String wellKnownMetadataPa
             return this;
         }
 
+        /**
+         * Optional Certificate refresh interval in seconds
+         * This parameter can only be set when tls_client_auth.
```
lhotari, line 241: This feature would be useful also for ClientCredentialsFlow, I'll add some other comments about unifying this.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlow.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Override
+    public void initialize() throws PulsarClientException {
+        super.initialize();
+        assert this.metadata != null;
+
+        URL tokenUrl = this.metadata.getTokenEndpoint();
+        TlsHttpClientContext tlsHttpClientContext = createTlsHttpClientContext(tokenUrl.getHost());
+        this.tokenHttpClient = tlsHttpClientContext.httpClient();
+        this.sslFactory = tlsHttpClientContext.sslFactory();
```
lhotari, line 144: Make the current defaultHttpClient in FlowBase protected and rename it to createHttpClient. Override the method and create the http client there. This avoids having 2 separate httpClient instances.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlow.java
```diff
@@ -0,0 +1,244 @@
...
+            confBuilder.setCookieStore(null);
+            confBuilder.setUseProxyProperties(true);
+            confBuilder.setFollowRedirect(true);
+            confBuilder.setConnectTimeout(
+                    getParameterDurationToMillis(
+                            CONFIG_PARAM_CONNECT_TIMEOUT, connectTimeout, DEFAULT_CONNECT_TIMEOUT));
+            confBuilder.setReadTimeout(
+                    getParameterDurationToMillis(CONFIG_PARAM_READ_TIMEOUT, readTimeout, DEFAULT_READ_TIMEOUT));
+            confBuilder.setUserAgent(String.format("Pulsar-Java-v%s", PulsarVersion.getVersion()));
+            SslEngineFactory sslEngineFactory = new PulsarHttpAsyncSslEngineFactory(sslFactory, hostname);
```
lhotari, line 188: In FlowBase, create another protected method which creates a DefaultAsyncHttpClientConfig.Builder which could then be used in the overridden createHttpClient where the HttpClient instance should be created.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlow.java
```diff
@@ -0,0 +1,244 @@
...
+                .build();
+    }
+
+    @Override
+    public void initialize() throws PulsarClientException {
+        super.initialize();
+        assert this.metadata != null;
+
+        URL tokenUrl = this.metadata.getTokenEndpoint();
+        TlsHttpClientContext tlsHttpClientContext = createTlsHttpClientContext(tokenUrl.getHost());
```
lhotari, line 142: It's not necessary to pass the hostname when creating PulsarHttpAsyncSslEngineFactory. Therefore it should be possible to use the httpclient as described in the comment about renaming defaultHttpClient to createHttpClient in FlowBase and possibly overriding it in this class. (unless providing a client certificate isn't desired when discovering the metadata, that would be unexpected).

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlow.java
```diff
@@ -0,0 +1,244 @@
...
+
+        if (sslFactory != null && autoCertRefreshSeconds > 0) {
+            scheduler = Executors.newSingleThreadScheduledExecutor(new ExecutorProvider
+                            .ExtendedThreadFactory("tls-cert-refresher"));
+            scheduler.scheduleAtFixedRate(this::refreshSslContext,
+                autoCertRefreshSeconds, autoCertRefreshSeconds, TimeUnit.SECONDS);
+
+            log.info("Scheduled TLS certificate refresh every {} seconds", autoCertRefreshSeconds);
+        }
+
```
lhotari, line 156: This feature would be useful also at FlowBase level

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/TokenClient.java
```diff
@@ -53,13 +53,16 @@ public void close() throws Exception {
...
-    String buildClientCredentialsBody(ClientCredentialsExchangeRequest req) {
+    String buildClientCredentialsBody(ClientCredentialsExchangeRequest req, boolean includeSecret) {
         Map<String, String> bodyMap = new TreeMap<>();
         bodyMap.put("grant_type", "client_credentials");
         bodyMap.put("client_id", req.getClientId());
-        bodyMap.put("client_secret", req.getClientSecret());
+        if (includeSecret) {
+            bodyMap.put("client_secret", req.getClientSecret());
+        }
         // Only set audience and scope if they are non-empty.
```
lhotari, line 66: Instead of using `includeSecret`, Create an `enum` which is called `TokenEndpointAuthMethod`.  The `ClientCredentialsExchangeRequest` should have a field `TokenEndpointAuthMethod authMethod` which carries the token endpoint auth method.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/TokenClient.java
```diff
@@ -82,8 +85,31 @@ String buildClientCredentialsBody(ClientCredentialsExchangeRequest req) {
...
+     * @param req the client credentials request details for TLS client auth
+     * @return a token result
+     * @throws TokenExchangeException
+     */
+    public TokenResult exchangeTlsClientAuth(ClientCredentialsExchangeRequest req)
+            throws TokenExchangeException, IOException {
+        String body = buildClientCredentialsBody(req, false);
+        return executeTokenRequest(body);
+    }
+
```
lhotari, line 104: No need to add this after adding `TokenEndpointAuthMethod authMethod` to `ClientCredentialsExchangeRequest`

lhotari (commented):
Thanks for the contribution! I added some review comments.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/TokenClient.java
izumo27: I added `TokenEndpointAuthMethod` for `client_secret_post` and `tls_client_auth`

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlow.java
izumo27: I modified FlowBase to create the SslFactory.
Since there is no longer any difference between the HTTP clients in `FlowBase` and `TlsClientAuthFlow`, I did not override it in `TlsClientAuthFlow`.

lhotari (approved):

izumo27:
@lhotari PTAL, thanks.

lhotari:
> @lhotari PTAL, thanks.

@izumo27 Please check the copilot review comments and comment on them and mark them resolved after checking or addressing them.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
```diff
@@ -91,7 +127,40 @@ private AsyncHttpClient defaultHttpClient(Duration readTimeout, Duration connect
         return new DefaultAsyncHttpClient(confBuilder.build());
     }
 
+    protected void scheduleSslContextRefreshIfEnabled(long refreshSeconds) {
+        if (sslFactory == null || refreshSeconds <= 0 || sslRefreshScheduler != null) {
+            return;
+        }
+        sslRefreshScheduler = Executors.newSingleThreadScheduledExecutor(
+                new ExecutorProvider.ExtendedThreadFactory("oauth2-tls-cert-refresher"));
```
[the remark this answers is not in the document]
izumo27: When the client shuts down, the flow is closed, so there is no need to create daemon threads.
https://github.com/apache/pulsar/blob/v4.2.0/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationOAuth2.java#L316

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
```diff
@@ -47,28 +55,38 @@ abstract class FlowBase implements Flow {
...
 
     protected FlowBase(URL issuerUrl, Duration connectTimeout, Duration readTimeout, String trustCertsFilePath,
+                       String certFile, String keyFile, Duration autoCertRefreshDuration,
                        String wellKnownMetadataPath) {
         this.issuerUrl = issuerUrl;
-        this.httpClient = defaultHttpClient(readTimeout, connectTimeout, trustCertsFilePath);
+        this.httpClient = defaultHttpClient(readTimeout, connectTimeout, trustCertsFilePath, certFile, keyFile);
+        long autoCertRefreshSeconds = getParameterDurationToSeconds(CONFIG_PARAM_AUTO_CERT_REFRESH_DURATION,
+                autoCertRefreshDuration, DEFAULT_AUTO_CERT_REFRESH_DURATION);
+        scheduleSslContextRefreshIfEnabled(autoCertRefreshSeconds);
```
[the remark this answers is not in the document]
izumo27: Since certificates have an expiration date, it is useful for users that this is enabled by default.
It is also enabled by default in the client configuration.
https://github.com/apache/pulsar/blob/v4.2.0/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java#L273

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/TokenEndpointAuthMethod.java
```diff
@@ -0,0 +1,43 @@
...
+
+    TokenEndpointAuthMethod(String value) {
+        this.value = value;
+    }
+
+    public String value() {
+        return value;
+    }
+
+    public static TokenEndpointAuthMethod fromValue(String value) {
```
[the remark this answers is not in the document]
izumo27: There is no issue since the default value is set.
https://github.com/apache/pulsar/pull/25538/changes#diff-9de24c0f495b7d47fe56fab498509a72d0901d0d391b9d36c814818926fe6d5aR163-R165

pulsar-client/src/test/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlowTest.java
```diff
@@ -0,0 +1,44 @@
...
+        params.put("tlsCertFile", "/path/to/cert.pem");
+        params.put("tlsKeyFile", "/path/to/key.pem");
+        params.put("issuerUrl", "http://localhost");
+        params.put("scope", "http://localhost");
+        OAuth2MockHttpClient.withMockedSslFactory(() -> {
+            TlsClientAuthFlow flow = TlsClientAuthFlow.fromParameters(params);
+            Field clientIdField = flow.getClass().getDeclaredField("clientId");
+            clientIdField.setAccessible(true);
+            assertEquals((String) clientIdField.get(flow), "pulsar-client");
+            flow.close();
```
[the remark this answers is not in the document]
izumo27: Using reflection is not ideal, but verifying `clientId` is important.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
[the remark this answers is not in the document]
lhotari: I think it's better to create a daemon thread since there shouldn't be a reason to have a non-daemon thread in this case.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/protocol/ClientCredentialsExchangeRequest.java
```diff
@@ -42,4 +42,7 @@ public class ClientCredentialsExchangeRequest {
 
     @JsonProperty("scope")
     private String scope;
+
+    @JsonProperty("token_endpoint_auth_method")
+    private TokenEndpointAuthMethod authMethod;
```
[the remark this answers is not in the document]
izumo27: Set a default value when `authMethod` is null.
https://github.com/apache/pulsar/pull/25538/changes#diff-2fae4d00ed0258a32db65773afbde92d185cc2b41ef2601b888614d65cd19a62R59-R61

pulsar-client/src/test/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlowTest.java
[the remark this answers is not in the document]
lhotari: It's better to replace reflection by adding a package private method which is annotated with @VisibleForTesting. Since the test is in the same package as the class, it can access the method.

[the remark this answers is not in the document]
lhotari: FYI https://lists.apache.org/thread/7gr04sqmzyttx4ln6ydtp3qv0xgo1o6m

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
```diff
@@ -47,28 +55,38 @@ abstract class FlowBase implements Flow {
...
     public static final String CONFIG_PARAM_READ_TIMEOUT = "readTimeout";
     public static final String CONFIG_PARAM_TRUST_CERTS_FILE_PATH = "trustCertsFilePath";
+    public static final String CONFIG_PARAM_CERT_FILE = "tlsCertFile";
+    public static final String CONFIG_PARAM_TLS_KEY_FILE = "tlsKeyFile";
+    public static final String CONFIG_PARAM_AUTO_CERT_REFRESH_DURATION = "autoCertRefreshDuration";
     public static final String CONFIG_PARAM_WELL_KNOWN_METADATA_PATH = "wellKnownMetadataPath";
 
     protected static final Duration DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(10);
     protected static final Duration DEFAULT_READ_TIMEOUT = Duration.ofSeconds(30);
+    protected static final Duration DEFAULT_AUTO_CERT_REFRESH_DURATION = Duration.ofSeconds(300);
```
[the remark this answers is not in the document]
izumo27: Same as https://github.com/apache/pulsar/pull/25538#discussion_r3135480178

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
```diff
@@ -92,18 +94,35 @@ public static class ClientCredentialsBuilder {
...
 
+        /**
+         * Optional token endpoint auth method.
+         * Defaults to {@code client_secret_post}.
+         *
+         * @param tokenEndpointAuthMethod the token endpoint auth method
+         * @return the builder
+         */
+        public ClientCredentialsBuilder tokenEndpointAuthMethod(TokenEndpointAuthMethod tokenEndpointAuthMethod) {
+            this.tokenEndpointAuthMethod = tokenEndpointAuthMethod;
```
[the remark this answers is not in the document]
izumo27: If the authMethod is not supported, `IllegalArgumentException` is thrown.
https://github.com/apache/pulsar/pull/25538/changes#diff-4f723fd0ecb3c2991751d5391084a4f8a6a6aa65e5ca35a28864c382ae5603b6R340

izumo27:
@lhotari I checked the copilot comments and resolved them.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/AuthenticationFactoryOAuth2.java
```diff
@@ -236,16 +304,41 @@ public ClientCredentialsBuilder scheduler(ScheduledExecutorService scheduler) {
...
-                    .connectTimeout(connectTimeout)
-                    .readTimeout(readTimeout)
-                    .trustCertsFilePath(trustCertsFilePath)
-                    .wellKnownMetadataPath(wellKnownMetadataPath)
-                    .build();
+            Flow flow;
+            if (tokenEndpointAuthMethod == TokenEndpointAuthMethod.CLIENT_SECRET_POST) {
+                flow = ClientCredentialsFlow.builder()
+                        .issuerUrl(issuerUrl)
+                        .privateKey(credentialsUrl == null ? null : credentialsUrl.toExternalForm())
```
lhotari, line 311: it looks like `.clientId(clientId)` has been missing for ClientCredentialsFlow. It would be useful to add it since this PR touches these lines.

lhotari: It's odd that it's missing also from the ClientCredentialsFlow constructor. We can omit it from this PR.

lhotari: btw. this is how the ClientCredentialsFlow class handles it: https://github.com/apache/pulsar/blob/fae3df958f74f2afb0ba398b69d94fd3e6358c4f/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/ClientCredentialsFlow.java#L156 . It's part of the privateKey parameter.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2/FlowBase.java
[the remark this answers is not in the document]
izumo27: I've fixed to use a daemon thread.

pulsar-client/src/test/java/org/apache/pulsar/client/impl/auth/oauth2/TlsClientAuthFlowTest.java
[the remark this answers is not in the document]
izumo27: I've stopped using reflection.

lhotari (approved):
LGTM, good work @izumo27

izumo27:
Rebased to master

lhotari (approved):

PR #25548, opened 2026-04-17
[improve][ml] Warn and emit metric when cursor ack state exceeds persist limits

### LINKED ISSUE #25540: Add WARN log when cursor ack holes exceed managedLedgerMaxUnackedRangesToPersist
## Summary

When the number of individually deleted message ranges (ack holes) in a subscription cursor exceeds `managedLedgerMaxUnackedRangesToPersist` (default 10,000), the excess ack ranges are silently dropped during cursor persistence. No log message is emitted at any level.

On broker restart, the cursor is recovered from BookKeeper with only the persisted ranges. Acknowledged messages whose ack information was truncated are redelivered as if never acknowledged.

The `broker.conf` comment describes this behavior:

> "After the max number of ranges is reached, the information will only be tracked in memory and messages will be redelivered in case of crashes."

However, the operator has no warning that this threshold has been crossed. The only way to detect it today is to actively monitor `totalNonContiguousDeletedMessagesRange` via the Admin API or Prometheus and compare it manually to the configured limit.

## Impact

When the threshold is silently exceeded and a broker restart occurs, already-acknowledged messages are replayed in bulk. Depending on the application, this can:

- saturate consumers with a sudden spike of messages that were already processed
- cause unexpected side effects if message processing is not idempotent
- make the situation progressively worse, since replayed messages generate new ack holes after being re-processed

Without a log at the threshold crossing, operators have no way to anticipate this before a restart triggers it.

## Proposal

Add a rate-limited `log.warn` when the threshold is crossed. For example:

```
WARN - Cursor {cursor} on ledger {ledger}: individually deleted message ranges ({actual}) 
       exceeds managedLedgerMaxUnackedRangesToPersist ({limit}). 
       Ack state beyond this limit is not persisted and will be lost on broker restart.
```

Suggested locations in `ManagedCursorImpl.java`:
- `buildIndividualDeletedMessageRanges()` — where the truncation happens
- Or on state transition of `isCursorDataFullyPersistable()` from true to false

The log should be rate-limited to avoid spam when the threshold is permanently exceeded.

Happy to submit a PR if a maintainer agrees this is worth adding.

## Related

- PIP-299: `dispatcherPauseOnAckStatePersistentEnabled` — pauses dispatching at the limit, but also operates silently (no log at the transition)

ng-galien:
Fixes #25540

### Motivation

When a cursor persists more ack ranges than `managedLedgerMaxUnackedRangesToPersist` or more batch deleted indexes than `managedLedgerMaxBatchDeletedIndexToPersist`, the excess is silently truncated. On broker restart those acks are lost and messages are redelivered. Today there is no signal when this happens — operators have to monitor `totalNonContiguousDeletedMessagesRange` manually. The issue discussion asks for a WARN log (with tuning advice) and a **cursor-level** metric.

### Modifications

**Commit 1 — warn and emit metric on truncation:**

- WARN log emitted once per crossing (edge-detected) in both `buildIndividualDeletedMessageRanges()` and `buildBatchEntryDeletionIndexInfoList()`, with tuning advice covering the two limits, `managedLedgerPersistIndividualAckAsLongArray`, and `managedCursorInfoCompressionType`.
- Two OTel counters tagged with the cursor's standard attributes (pulsar.managed_ledger.name, pulsar.managed_ledger.cursor.name, pulsar.namespace)
- Signals documented next to the settings in `broker.conf`.

**Commit 2 — fix pre-existing off-by-one in the ranges cap:**

`buildIndividualDeletedMessageRanges` used to persist `maxRanges + 1` entries because the `forEach` callback added before testing `rangeList.size() <= maxRanges`. Regression introduced in #3819 when `stream().limit(N)` was dropped. Without this fix the new WARN/counter fire spuriously when `totalRanges == maxRanges + 1`. Fixed by switching to a check-before-add pattern (symmetric with `buildBatchEntryDeletionIndexInfoList`) with a `MutableBoolean` truncated flag.

### Verifying this change

- `ManagedCursorTest.testPersistUnackedRangesTruncatedCounter`
- `ManagedCursorTest.testPersistBatchDeletedIndexesTruncatedCounter`

- [x] The metrics — adds two cursor-level OTel counters (also documented in `broker.conf`).

### DIFF
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -234,6 +234,10 @@ public class ManagedCursorImpl implements ManagedCursor {
     // active state cache in ManagedCursor. It should be in sync with the state in activeCursors in ManagedLedger.
     private volatile boolean isActive = false;
 
+    // Emit the truncation WARN logs exactly once per crossing.
+    private final AtomicBoolean lastCursorDataFullyPersistable = new AtomicBoolean(true);
+    private final AtomicBoolean lastBatchDeletedIndexFullyPersistable = new AtomicBoolean(true);
+
     // This is a lock used to update the registration state of the cursor in the managed ledger.
     private final Object registerToWaitingCursorsLock = new Object();
     // This is used to track if the cursor is registered in the managed ledger's waitingCursors queue
@@ -3375,6 +3379,7 @@ public class ManagedCursorImpl implements ManagedCursor {
 
             AtomicInteger acksSerializedSize = new AtomicInteger(0);
             List<MessageRange> rangeList = new ArrayList<>();
+            final int maxRanges = getConfig().getMaxUnackedRangesToPersist();
 
             individualDeletedMessages.forEachRawRange((lowerKey, lowerValue, upperKey, upperValue) -> {
                 MessageRange messageRange = new MessageRange();
@@ -3388,11 +3393,32 @@ public class ManagedCursorImpl implements ManagedCursor {
                 acksSerializedSize.addAndGet(messageRange.getSerializedSize());
                 rangeList.add(messageRange);
 
-                return rangeList.size() <= getConfig().getMaxUnackedRangesToPersist();
+                return rangeList.size() <= maxRanges;
             });
 
             this.individualDeletedMessagesSerializedSize = acksSerializedSize.get();
             individualDeletedMessages.resetDirtyKeys();
+
+            if (rangeList.size() > maxRanges) {
+                ledger.getFactory().getOpenTelemetryManagedCursorStats().incrementPersistOverflowRanges();
+                if (lastCursorDataFullyPersistable.compareAndSet(true, false)) {
+                    int totalRanges = individualDeletedMessages.size();
+                    log.warn()
+                        .attr("totalRanges", totalRanges)
+                        .attr("maxRanges", maxRanges)
+                        .attr("truncated", totalRanges - rangeList.size())
+                        .log("Individually deleted message ranges exceed"
+                            + " managedLedgerMaxUnackedRangesToPersist."
+                            + " Acknowledged messages beyond this limit are not persisted"
+                            + " and will be replayed on broker restart."
+                            + " Consider raising managedLedgerMaxUnackedRangesToPersist,"
+                            + " verifying managedLedgerPersistIndividualAckAsLongArray=true (the default),"
+                            + " and setting managedCursorInfoCompressionType=LZ4 to reduce the persisted size.");
+                }
+            } else {
+                lastCursorDataFullyPersistable.compareAndSet(false, true);
+            }
+
             return rangeList;
         } finally {
             lock.writeLock().unlock();
@@ -3407,7 +3433,8 @@ public class ManagedCursorImpl implements ManagedCursor {
             }
             List<BatchedEntryDeletionIndexInfo> result = new ArrayList<>();
             final var iterator = batchDeletedIndexes.entrySet().iterator();
-            while (iterator.hasNext() && result.size() < getConfig().getMaxBatchDeletedIndexToPersist()) {
+            int maxIndexes = getConfig().getMaxBatchDeletedIndexToPersist();
+            while (iterator.hasNext() && result.size() < maxIndexes) {
                 final var entry = iterator.next();
                 BatchedEntryDeletionIndexInfo batchDeletedIndexInfo = new BatchedEntryDeletionIndexInfo();
                 batchDeletedIndexInfo.setPosition()
@@ -3419,6 +3446,26 @@ public class ManagedCursorImpl implements ManagedCursor {
                 }
                 result.add(batchDeletedIndexInfo);
             }
+
+            if (iterator.hasNext()) {
+                ledger.getFactory().getOpenTelemetryManagedCursorStats().incrementPersistOverflowBatchIndexes();
+                if (lastBatchDeletedIndexFullyPersistable.compareAndSet(true, false)) {
+                    int totalIndexes = batchDeletedIndexes.size();
+                    log.warn()
+                        .attr("totalIndexes", totalIndexes)
+                        .attr("maxIndexes", maxIndexes)
+                        .attr("truncated", totalIndexes - result.size())
+                        .log("Batch deleted indexes exceed"
+                            + " managedLedgerMaxBatchDeletedIndexToPersist."
+                            + " Partially acknowledged batch messages beyond this limit are not persisted"
+                            + " and will be replayed on broker restart."
+                            + " Consider raising managedLedgerMaxBatchDeletedIndexToPersist"
+                            + " and setting managedCursorInfoCompressionType=LZ4 to reduce the persisted size.");
+                }
+            } else {
+                lastBatchDeletedIndexFullyPersistable.compareAndSet(false, true);
+            }
+
             return result;
         } finally {
             lock.readLock().unlock();
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3406,7 +3406,7 @@ public class ManagedCursorImpl implements ManagedCursor {
 
             if (truncated.booleanValue()) {
                 ledger.getFactory().getOpenTelemetryManagedCursorStats()
-                        .incrementPersistUnackedRangesTruncated(ledger.getName(), name);
+                        .incrementPersistUnackedRangesTruncated(this);
                 if (lastCursorDataFullyPersistable.compareAndSet(true, false)) {
                     int totalRanges = individualDeletedMessages.size();
                     log.warn()
@@ -3455,7 +3455,7 @@ public class ManagedCursorImpl implements ManagedCursor {
 
             if (iterator.hasNext()) {
                 ledger.getFactory().getOpenTelemetryManagedCursorStats()
-                        .incrementPersistBatchDeletedIndexesTruncated(ledger.getName(), name);
+                        .incrementPersistBatchDeletedIndexesTruncated(this);
                 if (lastBatchDeletedIndexFullyPersistable.compareAndSet(true, false)) {
                     int totalIndexes = batchDeletedIndexes.size();
                     log.warn()
--- a/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
+++ b/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
@@ -3380,8 +3380,13 @@ public class ManagedCursorImpl implements ManagedCursor {
             AtomicInteger acksSerializedSize = new AtomicInteger(0);
             List<MessageRange> rangeList = new ArrayList<>();
             final int maxRanges = getConfig().getMaxUnackedRangesToPersist();
+            final MutableBoolean overflowed = new MutableBoolean(false);
 
             individualDeletedMessages.forEachRawRange((lowerKey, lowerValue, upperKey, upperValue) -> {
+                if (rangeList.size() >= maxRanges) {
+                    overflowed.setTrue();
+                    return false;
+                }
                 MessageRange messageRange = new MessageRange();
                 messageRange.setLowerEndpoint()
                         .setLedgerId(lowerKey)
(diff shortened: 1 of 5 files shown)


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3388,11 +3398,32 @@ private List<MessageRange> buildIndividualDeletedMessageRanges() {
...
             });
 
             this.individualDeletedMessagesSerializedSize = acksSerializedSize.get();
             individualDeletedMessages.resetDirtyKeys();
+
+            if (overflowed.booleanValue()) {
+                ledger.getFactory().getOpenTelemetryManagedCursorStats().incrementPersistOverflowRanges();
+                if (lastCursorDataFullyPersistable.compareAndSet(true, false)) {
+                    int totalRanges = individualDeletedMessages.size();
+                    log.warn()
```
lhotari, line 3412: add attributes for the managed ledger name and cursor name. "managedLedger" and "cursor" as attribute names.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3419,6 +3451,26 @@ private List<BatchedEntryDeletionIndexInfo> buildBatchEntryDeletionIndexInfoList
                 }
                 result.add(batchDeletedIndexInfo);
             }
+
+            if (iterator.hasNext()) {
+                ledger.getFactory().getOpenTelemetryManagedCursorStats().incrementPersistOverflowBatchIndexes();
+                if (lastBatchDeletedIndexFullyPersistable.compareAndSet(true, false)) {
+                    int totalIndexes = batchDeletedIndexes.size();
+                    log.warn()
+                        .attr("totalIndexes", totalIndexes)
```
lhotari, line 3462: add attributes for the managed ledger name and cursor name. "managedLedger" and "cursor" as attribute names.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -55,6 +56,16 @@ public class OpenTelemetryManagedCursorStats implements AutoCloseable {
     public static final String INCOMING_BYTE_COUNTER = "pulsar.broker.managed_ledger.cursor.incoming.size";
     private final ObservableLongMeasurement incomingByteCounter;
 
+    // Broker-level counters incremented when cursor persistence silently truncates ack state.
+    // See managedLedgerMaxUnackedRangesToPersist and managedLedgerMaxBatchDeletedIndexToPersist.
+    public static final String PERSIST_OVERFLOW_RANGES_COUNTER =
+            "pulsar.broker.managed_ledger.cursor.persist.overflow.range.count";
```
lhotari:
```suggestion
    public static final String PERSIST_UNACKED_RANGES_TRUNCATED =
            "pulsar.broker.managed_ledger.cursor.persist.unacked_ranges.truncated";
```
Changing the concept from "overflow" to persisting unacked ranges being truncated would improve the clarity of the counter.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -55,6 +56,16 @@ public class OpenTelemetryManagedCursorStats implements AutoCloseable {
...
+    private final LongCounter persistOverflowRangesCounter;
+
+    public static final String PERSIST_OVERFLOW_BATCH_INDEXES_COUNTER =
+            "pulsar.broker.managed_ledger.cursor.persist.overflow.batch.index.count";
```
lhotari:
```suggestion
    public static final String PERSIST_BATCH_DELETED_INDEXES_TRUNCATED =
            "pulsar.broker.managed_ledger.cursor.persist.batch_deleted_indexes.truncated";
```
Similar motivation for the change here, improving the clarity.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -55,6 +56,16 @@ public class OpenTelemetryManagedCursorStats implements AutoCloseable {
     public static final String INCOMING_BYTE_COUNTER = "pulsar.broker.managed_ledger.cursor.incoming.size";
     private final ObservableLongMeasurement incomingByteCounter;
 
+    // Broker-level counters incremented when cursor persistence silently truncates ack state.
+    // See managedLedgerMaxUnackedRangesToPersist and managedLedgerMaxBatchDeletedIndexToPersist.
+    public static final String PERSIST_OVERFLOW_RANGES_COUNTER =
+            "pulsar.broker.managed_ledger.cursor.persist.overflow.range.count";
+    private final LongCounter persistOverflowRangesCounter;
```
lhotari: make the field name match the "truncated" concept.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -96,6 +107,22 @@ public OpenTelemetryManagedCursorStats(OpenTelemetry openTelemetry, ManagedLedge
                 .setDescription("The total amount of data read from the ledger.")
                 .buildObserver();
 
+        persistOverflowRangesCounter = meter
+                .counterBuilder(PERSIST_OVERFLOW_RANGES_COUNTER)
+                .setUnit("{overflow}")
```
lhotari:
```suggestion
                .setUnit("{truncation}")
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -96,6 +107,22 @@ public OpenTelemetryManagedCursorStats(OpenTelemetry openTelemetry, ManagedLedge
...
+                .setDescription("The number of times a cursor exceeded"
+                        + " managedLedgerMaxUnackedRangesToPersist, causing ack state to be truncated"
+                        + " at persistence. Ack state beyond the limit is lost on broker restart.")
+                .build();
+
+        persistOverflowBatchIndexesCounter = meter
+                .counterBuilder(PERSIST_OVERFLOW_BATCH_INDEXES_COUNTER)
+                .setUnit("{overflow}")
```
lhotari:
```suggestion
                .setUnit("{truncation}")
```

lhotari (commented):
Good work @ng-galien! Some comments on naming. In the metric names, I think it's clearer to describe the effect, "persisted unacked ranges being truncated" or "persisted batch deleted indexes being truncated" — rather than using "overflow". The former tells operators directly what happened (state was dropped on persist), whereas "overflow" is a more abstract term that requires them to infer the consequence.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
ng-galien: Ledger and cursor name are un the log constructor, is it sufficient?

lhotari: yes, I missed that.

lhotari (approved):
LGTM, good work @ng-galien

ng-galien:
Hi @lhotari, thanks for your review. Semantic is aligned with truncate and telemetry is more precise as you suggest.
I've just noticed the logger already have ledger and cursor name at constructor level.

ng-galien:
@lhotari `managed-ledger:checkstyleTest` and and `managed-ledger:test` are now green on side, my bad.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -115,6 +144,18 @@ public void close() {
...
+    public void incrementPersistUnackedRangesTruncated(String managedLedger, String cursor) {
+        persistUnackedRangesTruncated.add(1, Attributes.of(
+                AttributeKey.stringKey("managedLedger"), managedLedger,
+                AttributeKey.stringKey("cursor"), cursor));
+    }
+
+    public void incrementPersistBatchDeletedIndexesTruncated(String managedLedger, String cursor) {
+        persistBatchDeletedIndexesTruncated.add(1, Attributes.of(
+                AttributeKey.stringKey("managedLedger"), managedLedger,
+                AttributeKey.stringKey("cursor"), cursor));
```
lhotari: for attributes, pass `managedCursorAttributes.getAttributes()` from ManagedCursorImpl

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -115,6 +144,18 @@ public void close() {
         batchCallback.close();
     }
 
+    public void incrementPersistUnackedRangesTruncated(String managedLedger, String cursor) {
+        persistUnackedRangesTruncated.add(1, Attributes.of(
+                AttributeKey.stringKey("managedLedger"), managedLedger,
+                AttributeKey.stringKey("cursor"), cursor));
+    }
```
lhotari: The attributes are already available:

```suggestion
    public void incrementPersistUnackedRangesTruncated(ManagedCursor cursor) {
        persistUnackedRangesTruncated.add(1, cursor.getManagedCursorAttributes().getAttributes());
    }
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
```diff
@@ -115,6 +144,18 @@ public void close() {
...
+
+    public void incrementPersistBatchDeletedIndexesTruncated(String managedLedger, String cursor) {
+        persistBatchDeletedIndexesTruncated.add(1, Attributes.of(
+                AttributeKey.stringKey("managedLedger"), managedLedger,
+                AttributeKey.stringKey("cursor"), cursor));
+    }
```
lhotari: The attributes are already available:

```suggestion
    public void incrementPersistBatchDeletedIndexesTruncated(ManagedCursor cursor) {
        persistBatchDeletedIndexesTruncated.add(1, cursor.getManagedCursorAttributes().getAttributes());
    }
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3388,11 +3398,33 @@ private List<MessageRange> buildIndividualDeletedMessageRanges() {
...
-                return rangeList.size() <= getConfig().getMaxUnackedRangesToPersist();
+                return true;
             });
 
             this.individualDeletedMessagesSerializedSize = acksSerializedSize.get();
             individualDeletedMessages.resetDirtyKeys();
+
+            if (truncated.booleanValue()) {
+                ledger.getFactory().getOpenTelemetryManagedCursorStats()
+                        .incrementPersistUnackedRangesTruncated(ledger.getName(), name);
```
lhotari:
```suggestion
                        .incrementPersistUnackedRangesTruncated(this);
```

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
```diff
@@ -3419,6 +3452,27 @@ private List<BatchedEntryDeletionIndexInfo> buildBatchEntryDeletionIndexInfoList
                 }
                 result.add(batchDeletedIndexInfo);
             }
+
+            if (iterator.hasNext()) {
+                ledger.getFactory().getOpenTelemetryManagedCursorStats()
+                        .incrementPersistBatchDeletedIndexesTruncated(ledger.getName(), name);
```
lhotari:
```suggestion
                        .incrementPersistBatchDeletedIndexesTruncated(this);
```

lhotari (commented):
A few comments about details that Claude Code spotted.

In addition 2 comments about the description:
  - reconcile the PR description (it claims broker‑level / no cardinality growth, but the code emits per‑cursor labels),
  - update the Modifications section to use the final metric names.

Since the counter will only be emitted in Otel when the threshold has been crossed, it's fine to increase cardinality. In Prometheus this would be different.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
ng-galien: done

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
ng-galien: done

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpenTelemetryManagedCursorStats.java
ng-galien: done

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java
ng-galien: done

lhotari (approved):

PR #25551, opened 2026-04-20
[fix][broker]Namespaces can be created with may empty replication_clusters policy

poorbarcode:
### Motivation

Before https://github.com/apache/pulsar/pull/24860, the namespace-level policy: `replication_clusters` will never be empty; after https://github.com/apache/pulsar/pull/24860, it will be empty if a namespace is created with the following request

```
curl -x PUT http://localhost:50025/admin/v2/namespaces/{tenant}/{namespace}` -D `{"allowed_clusters": ["{current_cluster}"]}`
```

This will lead to some unexpected problems, such as the namespace not being able to be deleted normally and getting an error `Failed to delete namespace {namespace}
java.util.concurrent.CompletionException: org.apache.pulsar.broker.web.RestException: Cannot delete the global namespace {namespace}. There are still more than one replication cluster configured.`

### Modifications

- Recover the default behaviour that was defined before https://github.com/apache/pulsar/pull/24860
- Improve the error message

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/impl/NamespacesBase.java
@@ -491,7 +491,7 @@ public abstract class NamespacesBase extends AdminResource {
                                 // There are still more than one clusters configured for the global namespace
                                 throw new RestException(Status.PRECONDITION_FAILED,
                                     "Cannot delete the global namespace " + nsName + ". There are still more than "
-                                    + "one replication clusters configured.");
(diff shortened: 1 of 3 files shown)


lhotari:
The test "org.apache.pulsar.broker.admin.AdminApi2Test > testCreateNamespaceWithEmptyReplicationClustersByHttp" fails.

Technoboy- (approved):

lhotari (approved):

PR #25552, opened 2026-04-20
[fix][broker] Use explicit auth ops for namespace properties endpoints

mattisonchao:
### Motivation

Namespace `property` / `properties` endpoints still authorize through `validateAdminAccessForTenantAsync`, so they do not map to explicit namespace operations. This makes it harder to introduce operation-specific authorization for namespace property reads, updates, and deletes while still keeping compatibility with the legacy tenant-admin authorization path.

### Modifications

- Added `GET_PROPERTIES`, `UPDATE_PROPERTIES`, and `DELETE_PROPERTIES` to `NamespaceOperation`
- Updated `PulsarAuthorizationProvider` to recognize the new namespace property operations
- Added a compatibility helper that accepts either tenant-admin authorization or namespace-operation authorization
- Switched namespace `property` / `properties` endpoints to use explicit read, update, and delete operations
- Extended `NamespaceAuthZTest` to verify the expected operation is checked for each endpoint

### Verifying this change

- `./gradlew --no-configuration-cache :pulsar-broker:test --tests org.apache.pulsar.broker.admin.NamespaceAuthZTest --rerun-tasks`

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [x] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

PR in forked repository: https://github.com/mattisonchao/pulsar/pull/3


pulsar-common/src/main/java/org/apache/pulsar/common/policies/data/NamespaceOperation.java
```diff
@@ -34,6 +34,9 @@ public enum NamespaceOperation {
     GET_PERMISSION,
     GRANT_PERMISSION,
     REVOKE_PERMISSION,
+    GET_PROPERTIES,
+    UPDATE_PROPERTIES,
+    DELETE_PROPERTIES,
```
lhotari, line 39: Just wondering if adding new enums breaks existing 3rd party AuthorizationProviders which many Pulsar users use currently in production?

shibd (approved):

PR #25554, opened 2026-04-21
[fix][broker] Fix pending publish state leak when transfer-time writes fail on PersistentTopic

dao-jun:
### Motivation

  During bundle transfer, Pulsar intentionally keeps producer connections alive in the releasing phase and avoids triggering client reconnection until the later close step defined by PIP-307. However, if an in-flight publish reaches PersistentTopic.addFailed() while the topic is already transferring, the broker currently
  returns early without completing the publish callback.

  That leaves broker-side publish state stuck:

  - pendingWriteOps is not decremented
  - producer pendingPublishAcks is not decremented

  This creates a real leak in the transfer path and can block proper producer/topic cleanup, while the wire protocol semantics during transfer should remain unchanged.

### Modifications

  - Update PersistentTopic.addFailed() to complete the broker-side publish callback even when the topic is in transferring state.
  - Use TopicClosedException in that transfer-time failure path so the broker releases internal publish bookkeeping without sending a SendError to the client.
  - Decrement pendingWriteOps in that path to avoid stuck write state on the source broker.
  - Add a focused regression test covering an in-flight publish failure during transfer and asserting that:
      - no SendError is sent to the client
      - pendingWriteOps returns to zero
      - producer pendingPublishAcks returns to zero

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -758,19 +758,21 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
 
     @Override
     public synchronized void addFailed(ManagedLedgerException exception, Object ctx) {
+        PublishContext callback = (PublishContext) ctx;
         /* If the topic is being transferred(in the Releasing bundle state),
-         we don't want to forcefully close topic here.
-         Instead, we will rely on the service unit state channel's bundle(topic) transfer protocol.
-         At the end of the transfer protocol, at Owned state, the source broker should close the topic properly.
+         we don't want to forcefully close the topic here.
+         Instead, complete the broker-side publish callback and let the transfer protocol finish the topic close on the
+         source broker.
          */
         if (transferring) {
             log.debug()
                     .exception(exception)
                     .log("Failed to persist msg in store while transferring");
+            callback.completed(new TopicClosedException(exception), -1, -1);
+            decrementPendingWriteOpsAndCheck();
             return;
         }
 
-        PublishContext callback = (PublishContext) ctx;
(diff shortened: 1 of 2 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -758,19 +758,21 @@ public void addComplete(Position pos, ByteBuf entryData, Object ctx) {
...
+         we don't want to forcefully close the topic here.
+         Instead, complete the broker-side publish callback and let the transfer protocol finish the topic close on the
+         source broker.
          */
         if (transferring) {
             log.debug()
                     .exception(exception)
                     .log("Failed to persist msg in store while transferring");
+            callback.completed(new TopicClosedException(exception), -1, -1);
+            decrementPendingWriteOpsAndCheck();
```
merlimat, line 772: Not sending an error was intentional. 
The idea is to "freeze" the producer until the new broker is ready. if we return error here, the client will close the connection and reconnect to old broker again (then it needs to wait backoff time too...)

dao-jun: Thanks for your explain, now I understand it. 👍

merlimat: It would probably be better to have a comment there though :)

dao-jun: > It would probably be better to have a comment there though :)

Yes, I'd like to do it

PR #25555, opened 2026-04-21
[improve][broker] Cache publish metadata and avoid repeated parsing in the broker send path

dao-jun:
### Motivation

  ServerCnx.handleSend sits on the hot publish path, but the same MessageMetadata can be parsed multiple times for a single message. In the persistent-topic flow, producer-side encryption validation, delayed-delivery checks, and deduplication/chunk handling may all reparse the same payload. That adds avoidable CPU overhead
  to a latency-sensitive path.

  This change reduces that overhead by parsing metadata once per publish attempt and reusing it across the broker publish flow. It also keeps the behavior safe by invalidating the cached metadata after broker interceptors run, so any interceptor-side buffer mutation does not reuse stale data.

### Modifications

  - Make send debug logging in ServerCnx lazy, so metadata is not parsed when debug logging is disabled.
  - Create MessagePublishContext before producer-side publish validation starts.
  - Cache MessageMetadata in MessagePublishContext and reuse it within the same publish attempt.
  - Reuse cached metadata in:
      - producer encryption-required validation
      - persistent topic maximum delivery delay validation
      - deduplication paths that previously reparsed message metadata
  - Invalidate the cached metadata after BrokerInterceptor.onMessagePublish(...) to preserve correctness if an interceptor mutates the payload.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
```diff
@@ -223,14 +229,19 @@ public void publishMessage(long producerId, long lowestSequenceId, long highestS
...
+                messagePublishContext)) {
+            publishMessageToTopic(headersAndPayload, messagePublishContext);
+        } else {
+            messagePublishContext.recycle();
         }
     }
 
     public boolean checkAndStartPublish(long producerId, long sequenceId, ByteBuf headersAndPayload, int batchSize,
-                                        Position position) {
+                                        Position position, PublishContext publishContext) {
```
lhotari, line 242: this changes the signature of a public method. Technically it's a breaking change, but could 3rd party plugins override the Producer class and is changing the method signatures a concern?
If it's not a concern, I'd suggest omitting all parameters that are part of PublishContext and getting them from the PublishContext instead of passing them as parameters to the method. In general, the number of parameters to a method should be minimized.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -4978,13 +4978,11 @@ public Optional<TopicName> getShadowSourceTopic() {
         return Optional.ofNullable(shadowSourceTopic);
     }
 
-    protected boolean isExceedMaximumDeliveryDelay(ByteBuf headersAndPayload) {
+    protected boolean isExceedMaximumDeliveryDelay(ByteBuf headersAndPayload, PublishContext publishContext) {
```
lhotari, line 4980: since this techincally breaks backwards compatibility for 3rd party implementors of PersistentTopic, is it possible that some 3rd party implementations override this method? perhaps not very likely?
(3rd party implementations are possible with org.apache.pulsar.broker.service.TopicFactory)

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
lhotari: It's very unlikely that this breaks anything so it should be fine to change the method signature.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
lhotari: It's very unlikely that this breaks anything so it should be fine to change the method signature.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
lhotari: Adding headersAndPayload to the PublishContext would break a lot of things so keeping that out of PublishContext makes sense. Just omitting the parameters that are already in PublishContext would be useful here.

lhotari (commented):
Yes, this is a very useful change. I've also been thinking of this when I added a similar optimization for the dispatcher side.
A few review comments.

dao-jun:
@lhotari Thanks for your review, the most of pulsar downstream projects(kop, rop, mop...) are not maintained anymore now, maybe some private forks are active, but it's not our duty to maintain their downward compatibility.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
```diff
@@ -292,26 +298,9 @@ private boolean isSupportsReplDedupByLidAndEid() {
...
-    private void publishMessageToTopic(ByteBuf headersAndPayload, long lowestSequenceId, long highestSequenceId,
-                                       int batchSize, boolean isChunked, boolean isMarker, Position position) {
-        MessagePublishContext messagePublishContext = MessagePublishContext.get(this, lowestSequenceId,
-                highestSequenceId, headersAndPayload.readableBytes(), batchSize,
-                isChunked, System.nanoTime(), isMarker, position, isSupportsReplDedupByLidAndEid());
+    private void publishMessageToTopic(ByteBuf headersAndPayload, MessagePublishContext messagePublishContext) {
         if (brokerInterceptor != null) {
-            brokerInterceptor
-                    .onMessagePublish(this, headersAndPayload, messagePublishContext);
+            brokerInterceptor.onMessagePublish(this, headersAndPayload, messagePublishContext);
```
Denovo1998, line 303: The PR body says the cached metadata is invalidated after BrokerInterceptor.onMessagePublish(...), but I don't see that in the final diff.

This changes behavior when metadata has already been cached before the interceptor runs. For example, on an encryption-required topic, checkAndStartPublish(...) calls publishContext.getMessageMetadata(...). If an interceptor then mutates the message metadata in headersAndPayload, PersistentTopic.isExceedMaximumDeliveryDelay(...) and MessageDeduplication will reuse the pre-interceptor cached metadata. In current master those later checks reparse headersAndPayload after the interceptor, so they observe the post-interceptor metadata.

Could we add a MessagePublishContext#clearMessageMetadata()/invalidateMessageMetadata() and call it immediately after BrokerInterceptor.onMessagePublish(...) in both the normal publish path and publishTxnMessage?

---

If multiple interceptors are executed in series and the first interceptor modifies metadata, and the second interceptor reads metadata through publishContext.getMessageMetadata(...), then clearing the cache only after the entire BrokerInterceptors returns cannot solve the visibility within the interceptor chain?

Denovo1998 (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Producer.java
```diff
@@ -208,8 +208,13 @@ public boolean isSuccessorTo(Producer other) {
...
             boolean isChunked, boolean isMarker, Position position) {
-        if (checkAndStartPublish(producerId, sequenceId, headersAndPayload, batchSize, position)) {
-            publishMessageToTopic(headersAndPayload, sequenceId, batchSize, isChunked, isMarker, position);
+        MessagePublishContext messagePublishContext =
+                MessagePublishContext.get(this, sequenceId, headersAndPayload.readableBytes(),
+                        batchSize, isChunked, System.nanoTime(), isMarker, position, isSupportsReplDedupByLidAndEid());
+        if (checkAndStartPublish(producerId, sequenceId, headersAndPayload, position, messagePublishContext)) {
+            publishMessageToTopic(headersAndPayload, messagePublishContext);
+        } else {
+            messagePublishContext.recycle();
```
Denovo1998, line 217: A related lifecycle optimization: after topic.publishMessage(...) / topic.publishTxnMessage(...) returns, the cached MessageMetadata appears no longer needed by the current publish flow. PersistentTopic consumes it synchronously for max-delivery-delay validation and deduplication before asyncAddEntry/transaction append. The later async completion path and messageProduced interceptor don't receive headersAndPayload, so keeping the parsed MessageMetadata attached to MessagePublishContext until recycle extends its lifetime to the full in-flight write duration.

If we add clearMessageMetadata() for the interceptor correctness issue above, it would be useful to also clear it after the topic publish call returns, preferably in a finally block, so high-throughput or high-latency writes don't retain parsed metadata objects longer than necessary.

Denovo1998 (changes requested):

PR #25556, opened 2026-04-21
[fix][doc] Correct loadConf example key in PulsarAdminBuilder

### LINKED ISSUE #24093: [Doc] Typo in PulsarAdminBuilder
### Search before asking

### What issue do you find in Pulsar docs?

https://github.com/apache/pulsar/blob/3e6f7deb0afec2675aac74750f7526557f43c9e7/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/PulsarAdminBuilder.java#L47 

it gives the example of using "serviceHttpUrl" as a key, however the correct key should be serviceUrl since PulsarAdminBuilder uses [ClientConfigurationData](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java) class. 

### What is your suggestion?

Update the comment to use the correct key 

### Any reference?

https://github.com/apache/pulsar/blob/3e6f7deb0afec2675aac74750f7526557f43c9e7/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/PulsarAdminBuilder.java#L47 

singhvishalkr:
Fixes #24093

### Motivation

`PulsarAdminBuilder.loadConf` documents an example that uses `serviceHttpUrl` as the config map key:

```java
Map<String, Object> config = new HashMap<>();
config.put("serviceHttpUrl", "http://localhost:6650");
```

However, `loadConf` resolves its keys against `ClientConfigurationData`, whose field is declared as `serviceUrl`:

https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/conf/ClientConfigurationData.java#L60-L65

```java
@ApiModelProperty(
        name = "serviceUrl",
        value = "Pulsar cluster HTTP URL to connect to a broker."
)
private String serviceUrl;
```

Copy-pasting the current example produces a silently ignored configuration entry and no admin service URL is set, which is a frustrating first-contact experience for anyone reading the API docs.

### Modifications

- `pulsar-client-admin-api/.../PulsarAdminBuilder.java`: change the `loadConf` example key from `serviceHttpUrl` to `serviceUrl`.

### Verifying this change

This change is a trivial Javadoc-only fix without any test coverage.

None of the listed areas are affected. This is a comment-only change inside a Javadoc example.

### Documentation

- [x] `doc-not-needed`

No user-facing documentation outside of this Javadoc example needs updating; the admin configuration reference already uses the correct `serviceUrl` key.

### DIFF
--- a/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/PulsarAdminBuilder.java
+++ b/pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/PulsarAdminBuilder.java
@@ -45,7 +45,7 @@ public interface PulsarAdminBuilder {
      * <pre>
      * {@code
      * Map<String, Object> config = new HashMap<>();
-     * config.put("serviceHttpUrl", "http://localhost:6650");
+     * config.put("serviceUrl", "pulsar://localhost:6650");
      *
      * PulsarAdminBuilder builder = ...;
      * builder = builder.loadConf(config);


dao-jun:
good catch

pulsar-client-admin-api/src/main/java/org/apache/pulsar/client/admin/PulsarAdminBuilder.java
```diff
@@ -45,7 +45,7 @@ public interface PulsarAdminBuilder {
      * <pre>
      * {@code
      * Map<String, Object> config = new HashMap<>();
-     * config.put("serviceHttpUrl", "http://localhost:6650");
+     * config.put("serviceUrl", "http://localhost:6650");
```
merlimat:
```suggestion
     * config.put("serviceUrl", "pulsar://localhost:6650");
```

singhvishalkr:
Thanks @merlimat — you're right, the URL scheme should be `pulsar://` to match the (corrected) key. Applied your suggestion in `7836370` and force-pushed; commit message updated to credit the scheme change to the review. Thanks @dao-jun for the 👍 on the original report.

dao-jun:
Just to clarify, `config.put("serviceUrl", "http://localhost:xxxx");` is also valid.
image

singhvishalkr:
Thanks for the clarification @dao-jun — you're right, `http://localhost:xxxx` is also accepted. `ClientConfigurationData#serviceUrl` takes both the native binary scheme (`pulsar://host:6650`, `pulsar+ssl://host:6651`) and the HTTP lookup scheme (`http://host:8080`, `https://host:8443`), so both spellings are perfectly valid values for the `serviceUrl` key.

I kept the example in this PR as `pulsar://localhost:6650` to match @merlimat's inline suggestion and because that pairs naturally with the default broker port shown in the snippet. Happy to extend it to something like

```java
// Any scheme accepted by PulsarAdmin is fine; e.g.
//   config.put("serviceUrl", "pulsar://localhost:6650");
//   config.put("serviceUrl", "http://localhost:8080");
```

if you think documenting both schemes in the Javadoc would be clearer for users landing on `loadConf`. Otherwise this PR is scoped purely to fixing the `servicUrl` → `serviceUrl` typo and I'd rather keep the doc change minimal — LMK which you prefer and I'll push a follow-up in the same PR.

dao-jun (approved):

dao-jun:
> Thanks for the clarification @dao-jun — you're right, `http://localhost:xxxx` is also accepted. `ClientConfigurationData#serviceUrl` takes both the native binary scheme (`pulsar://host:6650`, `pulsar+ssl://host:6651`) and the HTTP lookup scheme (`http://host:8080`, `https://host:8443`), so both spellings are perfectly valid values for the `serviceUrl` key.
> 
> I kept the example in this PR as `pulsar://localhost:6650` to match @merlimat's inline suggestion and because that pairs naturally with the default broker port shown in the snippet. Happy to extend it to something like
> 
> ```java
> // Any scheme accepted by PulsarAdmin is fine; e.g.
> //   config.put("serviceUrl", "pulsar://localhost:6650");
> //   config.put("serviceUrl", "http://localhost:8080");
> ```
> 
> if you think documenting both schemes in the Javadoc would be clearer for users landing on `loadConf`. Otherwise this PR is scoped purely to fixing the `servicUrl` → `serviceUrl` typo and I'd rather keep the doc change minimal — LMK which you prefer and I'll push a follow-up in the same PR.

No need to keep them both, this is just an example.

PR #25559, opened 2026-04-21
[feat] PIP-468: Add ScalableTopicController and broker infrastructure

merlimat:
## Summary

First PR in the PIP-468 series introducing scalable topics. This PR lands
the core broker-side infrastructure:

- `ScalableTopicController` — per-topic coordinator with leader election,
  consumer session management, and layout tracking.
- `SegmentLayout` — pure data/math for the hash-range segment DAG.
- `SubscriptionCoordinator` — in-memory consumer-assignment tracking
  per subscription.
- `ScalableTopicService` — per-broker service managing controller
  lifecycle (get/release controllers, lifecycle hooks on `BrokerService`).
- Metadata-store resources: `ScalableTopicResources`,
  `ScalableTopicMetadata`, `ConsumerRegistration`, `SubscriptionMetadata`,
  `SubscriptionType`.
- Unit tests for `SegmentLayout` and `SubscriptionCoordinator`.

## Tests

Only the self-contained pure-data components have unit tests in this PR
(`SegmentLayoutTest`, `SubscriptionCoordinatorTest`).

End-to-end tests of `ScalableTopicController` and `ScalableTopicService`
are deferred to subsequent PRs because their setup depends on code that
lands later:
- The tests mock `PulsarAdmin.scalableTopics()`, a type added by the
  admin-API PR.
- `splitSegment`/`mergeSegments` live with the admin layer (they invoke
  `admin.scalableTopics().createSegmentAsync` / `terminateSegmentAsync`)
  and their tests move with them.

## Test plan

- [x] `:pulsar-broker:compileJava` / `:pulsar-broker:compileTestJava`
      succeed on this commit in isolation.
- [x] `./gradlew :pulsar-broker:test --tests "org.apache.pulsar.broker.service.scalable.*"`
      passes (`SegmentLayoutTest`, `SubscriptionCoordinatorTest`).


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ConsumerSession.java
```diff
@@ -0,0 +1,142 @@
...
+        cancelGraceTimer();
+        this.graceTimer = timer;
+    }
+
+    public synchronized void cancelGraceTimer() {
+        if (graceTimer != null) {
+            graceTimer.cancel(false);
+            graceTimer = null;
+        }
+    }
```
lhotari: It seems that it would be cleaner to encapsulate the grace timer implementation details into this class. For example, when creating the session, the SubscriptionCoordinator could provide a callback function which gets called when the grace period expires.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicService.java
```diff
@@ -0,0 +1,218 @@
...
+        controllers.values().forEach(controller -> {
+            try {
+                controller.close().join();
+            } catch (Exception e) {
+                log.warn("Error closing controller for topic {}", controller.getTopicName(), e);
+            }
+        });
+        controllers.clear();
+    }
+
```
lhotari, line 81: Later on we might want to perform this asynchronously in parallel with a configured concurrency level to speed up broker shutdown in phases which don't conflict with each other. I think synchronous close is fine for now.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicService.java
```diff
@@ -0,0 +1,218 @@
...
+        log.info("Leader state change for scalable topic {}: {}", topic, state);
+        if (state == LeaderElectionState.NoLeader) {
+            // Try to re-elect
+            ScalableTopicController controller = controllers.get(topic.toString());
+            if (controller != null) {
+                controller.initialize().exceptionally(ex -> {
+                    log.warn("Failed to re-elect for topic {}", topic, ex);
+                    return null;
+                });
+            }
```
lhotari, line 187: One question came into mind here (not necessary to handle in this PR at all):

The concern I have about this is a potential thundering herd issue in the current LeaderElectionImpl code where there's handling for BadVersionException: https://github.com/apache/pulsar/blob/9855b18ce96f05c49961cbc390ea53337658ca1a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/coordination/impl/LeaderElectionImpl.java#L238-L240
There's a comment "There was a conflict between 2 participants trying to become leaders at same time. Retry to fetch info on new leader."

Does this problem exist? This might apply to Zookeeper since the code in LeaderElectionImpl doesn't use Zookeeper's `sync` which might be required in this case to prevent that stale state isn't returned. Does Oxia have similar visibility issues as Zookeeper (which requires using `sync` in certain cases)?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/SubscriptionCoordinator.java
```diff
@@ -0,0 +1,339 @@
...
+                    computeAssignment(currentLayout, sessions.values());
+            ConsumerAssignment assignment = current.get(existing);
+            if (assignment != null) {
+                existing.sendAssignmentUpdate(assignment);
+            }
+            return CompletableFuture.completedFuture(current);
+        }
+
+        // Fresh registration — persist first, then install in-memory and rebalance.
+        ConsumerSession session = new ConsumerSession(consumerName, consumerId, cnx);
```
lhotari: I added a separate comment in ConsumerSession about encapsulating the timer into ConsumerSession class. It would be possible to pass the timer scheduler and the callback for expiration when the instance is constructed.

I'm not exactly sure if it makes sense. Maybe not. :)

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/SubscriptionCoordinator.java
```diff
@@ -0,0 +1,339 @@
...
+            return;
+        }
+        session.markDisconnected();
+        log.info().attr("consumer", consumerName)
+                .attr("gracePeriodSeconds", gracePeriod.toSeconds())
+                .log("Consumer disconnected; starting grace period");
+        session.setGraceTimer(scheduler.schedule(
+                () -> evictExpiredConsumer(consumerName),
+                gracePeriod.toMillis(), TimeUnit.MILLISECONDS));
+    }
```
lhotari: if grace timer gets encapsulated, this could be `session.startGraceTimer()`

lhotari (approved):
LGTM. Good work!
just a comment about possible encapsulation of the grace timer in ConsumerSession (might not make any sense) which you can ignore. I made an unrelated question about possible thundering heard in leader election (since there's a comment about BadVersionException in the current implementation). Feel free to mark the comments as resolved and merge the PR.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ConsumerSession.java
merlimat: Good point. Fixing it

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicService.java
merlimat: There is still the possibility for bad-version when there are 2 clients trying to become leader at same time. Ultimately it's a conditional write (create-if-not-exists) and only 1 will succeed.

PR #25564, opened 2026-04-22
[feat] PIP-468: Add scalable topic protocol commands and connection handling

merlimat:
> **Stacked on top of #25559.** This PR is opened as a draft and will be
> rebased on master once #25559 lands. Reviewers: the new code in this
> PR is the single commit titled "PIP-468: Add scalable topic protocol
> commands and connection handling"; everything else comes from #25559.

## Summary

Second PR in the PIP-468 series. Introduces the wire protocol for the
scalable-topic client/broker interaction:

- New protocol commands on `PulsarApi.proto`:
  `CommandScalableTopicLookup`, `CommandScalableTopicUpdate`,
  `CommandScalableTopicClose`, `CommandScalableTopicSubscribe`,
  `CommandScalableTopicSubscribeResponse`,
  `CommandScalableTopicAssignmentUpdate`.
- `Commands` factory methods for building each message (including the
  error variants for lookup/subscribe).
- `PulsarDecoder` dispatch entries for all six new message types.
- Broker-side `DagWatchSession` — resolves per-segment broker
  addresses, watches the metadata store for DAG changes, and pushes
  `ScalableTopicUpdate` frames on the wire. Tied to the transport
  connection — closed when the connection drops.
- `ServerCnx` handlers for `ScalableTopicLookup` and
  `ScalableTopicClose`.
- Client-side `DagWatchSession` interface (callback surface for the V5
  client that lands later).
- `ClientCnx` handler for `ScalableTopicUpdate` that dispatches to the
  registered session.
- `ConsumerSession.sendAssignmentUpdate` wired through to the new
  command sender (previously a placeholder no-op).

## Tests

Three new unit-test classes:

- `CommandsScalableTopicTest` (8 tests) — every new factory method
  roundtrips through the full wire frame; uses
  `BaseCommand.materialize()` so the buffer can be safely released
  before assertions.
- `ConsumerSessionTest` (15 tests) — covers identity/equality,
  attach/markDisconnected, grace-timer lifecycle, all three
  `sendAssignmentUpdate` branches (connected, disconnected, null
  command sender), and the `toProto` static helper.
- `DagWatchSessionTest` (12 tests) — session identity, idempotent
  close, `start()` failure on missing topic, metadata-store listener
  registration, notification filtering (wrong path / Deleted /
  post-close), `pushUpdate` serializing the correct DAG to the
  connection, and post-close/no-broker-map edge cases.

The heavy `NamespaceService` lookup path inside `start()` /
`buildResponse` is deferred to broker integration tests — mocking the
full namespace stack here is more noise than signal.

## Test plan

- [x] `./gradlew :pulsar-common:test --tests "org.apache.pulsar.common.protocol.CommandsScalableTopicTest"`
      — 8/8 pass.
- [x] `./gradlew :pulsar-broker:test --tests "org.apache.pulsar.broker.service.scalable.*"`
      — 61/61 pass (5 test classes, including the two existing ones
      from the previous PR).
- [x] `./gradlew :pulsar-broker:checkstyleMain :pulsar-broker:checkstyleTest :pulsar-common:checkstyleMain :pulsar-common:checkstyleTest`
      — all clean.

### DIFF
--- a/pulsar-common/src/main/proto/PulsarApi.proto
+++ b/pulsar-common/src/main/proto/PulsarApi.proto
@@ -834,6 +834,111 @@ message CommandWatchTopicListClose {
     required uint64 watcher_id     = 2;
 }
 
+/// --- Scalable topic commands ---
+
+enum SegmentState {
+    ACTIVE = 0;
+    SEALED = 1;
+}
+
+message SegmentInfoProto {
+    required uint64 segment_id       = 1;
+    required uint32 hash_start       = 2;
+    required uint32 hash_end         = 3;
+    required SegmentState state      = 4;
+    repeated uint64 parent_ids       = 5;
+    repeated uint64 child_ids        = 6;
+    required uint64 created_at_epoch = 7;
+    optional uint64 sealed_at_epoch  = 8;
+}
+
+message SegmentBrokerAddress {
+    required uint64 segment_id     = 1;
+    required string broker_url     = 2;
+    optional string broker_url_tls = 3;
+}
+
+message ScalableTopicDAG {
+    required uint64 epoch                           = 1;
+    repeated SegmentInfoProto segments              = 2;
+    repeated SegmentBrokerAddress segment_brokers   = 3;
+    optional string controller_broker_url           = 4;
+    optional string controller_broker_url_tls       = 5;
+}
+
+// Client -> Broker: Request scalable topic metadata and initiate watch session
+message CommandScalableTopicLookup {
+    required uint64 session_id = 1;   // Client-assigned session ID
+    required string topic      = 2;   // e.g. "topic://tenant/ns/my-topic"
+}
+
+// Broker -> Client: Used for BOTH initial response and subsequent pushed updates
+message CommandScalableTopicUpdate {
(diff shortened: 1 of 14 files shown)


pulsar-common/src/main/proto/PulsarApi.proto
```diff
@@ -1070,6 +1175,14 @@ message BaseCommand {
         WATCH_TOPIC_LIST_CLOSE = 67;
 
         TOPIC_MIGRATED = 68;
+
+        SCALABLE_TOPIC_LOOKUP  = 70;
```
lhotari, line 1179: Is it useful to also add ProtocolVersion enum or a FeatureFlag for the new scalable topics protocol?

lhotari (approved):
LGTM, one minor question

pulsar-common/src/main/proto/PulsarApi.proto
merlimat: That's a good point. It should be a feature flag, so the client can provide a meaningful error message to users. I took a note and will implement in a follow up.

PR #25570, opened 2026-04-22
[fix][broker] Replace Java serialization with JSON in package metadata

lhotari:
### Motivation

`PackageMetadataUtil.fromBytes` used `org.apache.commons.lang3.SerializationUtils.deserialize`, a thin wrapper over `ObjectInputStream.readObject`. `readObject` instantiates arbitrary classes from the stream **before** the post-read `instanceof PackageMetadata` check runs, so any gadget chain on the classpath would fire before validation. This was the only production Java-deserialization surface in Pulsar (`SerializationUtils.deserialize` had exactly one hit; the remaining `ObjectInputStream`/`readObject` references are `BitSetRecyclable`'s in-memory recycler and the BouncyCastle PEM reader, neither a Java-serialization surface).

#### Severity — not externally reachable

The deserialization path is **not reachable from external attackers**. The REST entry points (`pulsar-broker/.../admin/v3/Packages.java`) accept JSON via Jackson from authenticated admins; Pulsar then re-serializes with Java serialization to store in BookKeeper / filesystem. The Java-serialized blob only exists at the storage layer, written by Pulsar itself from already-validated JSON. Exploiting the gadget chain requires either write access to the backing package storage, or a separate future bug that feeds raw bytes into `fromBytes`.

This PR is best understood as **defense in depth plus eliminating an unnecessary dangerous primitive**, not a live RCE fix.

### Modifications

**Write path** — JSON via the existing `ObjectMapperFactory` (same factory used across the codebase for admin/REST serialization).

**Read path** — auto-detects format per entry:
- Leading `{` → JSON (always safe).
- Java stream magic `0xAC 0xED` → legacy deserialization, gated by config and wrapped in a strict JEP 290 `ObjectInputFilter`. The filter allowlist is narrow: `PackageMetadata`, `HashMap`/`LinkedHashMap`/`Map`/`Map.Entry`/`AbstractMap`, `String`/`Number`/`Long`/`Boolean`, `Object`, plus resource limits (`maxdepth=5`, `maxrefs=100`, `maxbytes=1MB`, `maxarray=1024`) and `!*` denying everything else. Class descriptors are filtered *before* instantiation, so gadget chains fail immediately.
- Anything else → `MetadataFormatException` with a message naming the config flag to flip.

**Two new broker config flags (both default `true`, under `CATEGORY_PACKAGES_MANAGEMENT`)**:
- `packagesManagementJsonSerializationEnabled` — write format. Rollback switch; legacy format will be removed in a future release.
- `packagesManagementAllowLegacyJavaSerialization` — read acceptance. The filter is always applied when the legacy path runs. This default is scheduled to flip to `false` in a future release.

The two flags allow a phased rollout: defaults ship the fix active; operators can flip either independently (rollback writes, or tighten reads after migration).

On broker startup the new ctor in `PackagesManagementImpl` emits `WARN`s describing the active combination (migration window, rollback mode, or the misconfigured "new writes use Java but reads reject it" state).

### Verifying this change

- `./gradlew :pulsar-package-management:pulsar-package-core:test --tests 'PackageMetadataSerdeTest' --tests 'PackagesManagementImplTest'` — all pass.
- `./gradlew :pulsar-broker-common:compileJava :pulsar-broker:compileJava` — passes.
- Static audit: `rg 'SerializationUtils\.(de)?serialize|ObjectInputStream' pulsar-package-management/ --glob '*.java'` → production hits are confined to `PackageMetadataUtil.java` (both paths gated by flags); remaining hits are the filter-rejection test fixture.

New tests in `PackageMetadataSerdeTest`:
- `testJsonRoundTrip` — confirms bytes start with `{`.
- `testLegacyRoundTrip` — confirms bytes start with `0xAC 0xED`.
- `testJsonReadableWhenLegacyDisabled` — JSON always works.
- `testLegacyRejectedWhenLegacyDisabled` — rejection names the config flag.
- `testFilterRejectsUnsafeClass` — crafts a valid Java stream with a disallowed class (`ArrayList`); filter rejects **before** instantiation, not via the post-read `instanceof`.
- `testUnknownFormatRejected`, `testEmptyRejected`.

- [x] The public API — `PackageMetadataUtil.fromBytes(byte[])` and `toBytes(PackageMetadata)` remain as `@Deprecated` overloads delegating to the new two-arg forms; source-compatible for external callers.
- [x] The default values of configurations — two new configs added (both default `true`; one scheduled to flip to `false` in a future release as documented in its `@FieldContext` doc).

### Documentation

- [x] `doc-not-needed`

(The config docs in `ServiceConfiguration.java` carry the operator-facing migration notes; a separate release-notes entry will cover the scheduled default flip.)

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -4065,6 +4065,24 @@ public class ServiceConfiguration implements PulsarConfiguration {
     )
     private String packagesManagementLedgerRootPath = "/ledgers";
 
+    @FieldContext(
+        category = CATEGORY_PACKAGES_MANAGEMENT,
+        doc = "Whether new package metadata writes use JSON (safe) or the legacy Java serialization format. "
+            + "Defaults to true. Set to false only as a temporary rollback path; the legacy format will be "
+            + "removed in a future release."
+    )
+    private boolean packagesManagementJsonSerializationEnabled = true;
+
+    @FieldContext(
+        category = CATEGORY_PACKAGES_MANAGEMENT,
+        doc = "Whether to accept reading legacy Java-serialized package metadata written by older brokers. "
+            + "A strict ObjectInputFilter allowlist is applied unconditionally whenever the legacy path "
+            + "runs. Defaults to true for upgrade compatibility; disable once all existing packages have "
+            + "been re-uploaded or rewritten via updateMeta. This default is scheduled to flip to false in "
+            + "a future release."
+    )
+    private boolean packagesManagementAllowLegacyJavaSerialization = true;
+
     /* packages management service configurations (end) */
 
     @FieldContext(
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java
@@ -2074,7 +2074,9 @@ public class PulsarService implements AutoCloseable, ShutdownService {
 
     private void startPackagesManagementService() throws IOException {
         // TODO: using provider to initialize the packages management service.
-        this.packagesManagement = new PackagesManagementImpl();
+        this.packagesManagement = new PackagesManagementImpl(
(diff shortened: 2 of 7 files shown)


nodece (approved):
Makes sense from a security perspective.

dao-jun (approved):

PR #25565, opened 2026-04-22
[feat] PIP-468: Add scalable topics and segments admin APIs with CLI

merlimat:
> **Stacked on top of #25564** (which is stacked on #25559). Opened as a
> draft; will be rebased on master once the upstream PRs land. Reviewers:
> the new code in this PR is the single commit titled "PIP-468: Add
> scalable topics and segments admin APIs with CLI" — everything else
> comes from #25559 and #25564.

## Summary

Third PR in the PIP-468 series. Wires up the admin/management surface
for scalable topics:

### REST APIs (broker)

- `ScalableTopics` REST resource — create / list / get-metadata / get-stats /
  delete / create-subscription / delete-subscription / split / merge.
  Includes `redirectToControllerLeaderIfNeeded` for write operations so
  non-leader brokers automatically 307 to the leader.
- `Segments` REST resource — cross-broker segment-topic
  create / terminate / delete with ownership-based routing.

### Client-admin API

- `ScalableTopics` interface on `PulsarAdmin` with sync and async
  counterparts for every operation.
- `ScalableTopicsImpl` — REST client wiring.
- Data types under `org.apache.pulsar.common.policies.data`:
  `ScalableTopicMetadata` (plus nested `SegmentInfo` and `HashRange`),
  `ScalableTopicStats` (plus nested `SegmentStats` and `SubscriptionStats`
  records), `ScalableSubscriptionType` enum.

### CLI

- `CmdScalableTopics` `pulsar-admin` subcommand with `list`, `create`,
  `get-metadata`, `stats`, `delete`, `split-segment`, `merge-segments`.

### Broker integration

- `ScalableTopicController` gains `createSubscription`,
  `deleteSubscription`, and `getStats` — exposing admin-visible
  operations on top of the infrastructure added in #25559.
- `ScalableTopicService` gains matching per-topic delegations +
  admin-layer entry points (`createScalableTopic`, `deleteScalableTopic`,
  `splitSegment`, `mergeSegments`, etc.) used by the REST handlers.
- Small tweak to `DagWatchSession` to pick up the reviewed changes from
  #25564.

### New test classes (24 tests)

- `ScalableTopicStatsTest` (10 tests) — construction (no-args /
  `@Builder` / all-args), `@Builder.Default` fresh-per-instance
  behavior, equals/hashCode, and the nested `SegmentStats` /
  `SubscriptionStats` records.
- `ScalableTopicMetadataTest` (11 tests) — same construction patterns,
[...]

PR #25573, opened 2026-04-24
[feat] PIP-468: Implement V5 client with scalable topic producer and consumer types

merlimat:
> **Stacked on top of #25565** (which is on top of #25564, which is on top of
> #25559). Opened as a draft; will be rebased on master once the upstream PRs
> land. Reviewers: the new code in this PR is the single commit titled
> "PIP-468: Implement V5 client with scalable topic producer and consumer
> types" — everything else comes from #25559 / #25564 / #25565.

## Summary

Fourth (and final for this series) PR in PIP-468. Adds the
`pulsar-client-v5` module that implements the V5 client API (PIP-466) for
scalable topics.

### Core pieces

- **`PulsarClientV5`** + **`PulsarClientBuilderV5`** — wraps the existing
  v4 `PulsarClientImpl` and adds the scalable-topic routing layer on top.
- **`PulsarClientProviderV5`** — ServiceLoader SPI implementation for the
  V5 API.
- **`ScalableTopicProducer`** — key-hashing segment routing, round-robin
  for unkeyed messages, per-segment underlying v4 producers.
- **`ScalableStreamConsumer`** — ordered, cumulative-ack consumer that
  receives segment assignments from the controller leader and drives v4
  per-segment consumers.
- **`ScalableQueueConsumer`** — controller-bypassing consumer that
  attaches to every segment and drains them with the receive-loop fix
  from the earlier iterations.
- **`ScalableCheckpointConsumer`** — unmanaged connector-style consumer
  that can resume from a caller-supplied `Checkpoint` across all
  segments.
- **`DagWatchClient`** — client-side handle for the DAG watch session;
  implements the `DagWatchSession` callback from the protocol PR.
- **`MessageIdV5`** — V5 message-id wrapping a v4 id with a per-segment
  position vector for cumulative ack.
- **`CheckpointV5`** — caller-durable checkpoint including
  `EARLIEST` / `LATEST` sentinels and timestamp variants.
- **`ClientSegmentLayout`** — client-side view of the DAG built from a
  `ScalableTopicDAG` protobuf.
- **`SegmentRouter`** — 16-bit Murmur3 key-hash routing and round-robin
  for empty keys.
- **`AuthenticationAdapter`**, **`CryptoKeyReaderAdapter`**,
  **`SchemaAdapter`** — bridge V5 API types to the underlying v4
  implementations.
- **`V5Utils`** — small helpers (e.g. `asScalableTopicName` which rewraps
  a parsed name with the `topic://` domain).

### Supporting v4 change

- `PulsarClientImpl.createSegmentProducerAsync` — internal entry point
  for the V5 client to create producers on `segment://` topics,
  bypassing the scalable-domain guard that's meant for user-facing
  producer creation.

### New unit test classes (48 tests)

- **`V5UtilsTest`** (5 tests) — `asScalableTopicName` on bare name /
  short form / persistent:// / topic:// inputs; also pins the
  caching behavior for the topic:// fast path.
- **`SegmentRouterTest`** (10 tests) — key-routing determinism and
  correctness (hash lands inside the picked segment's range),
  single-segment, empty-segments error, uncovered-hash error,
  round-robin cycling + wrap, hash 16-bit bound, string/bytes agree.
- **`MessageIdV5Test`** (15 tests) — constructor + null-rejection,
  immutable position vector, equals/hashCode (position vector is
  diagnostic, not identity), compareTo by segment then v4 id, foreign-
  type rejection, byte-array roundtrip with and without position
  vector, malformed input handling, EARLIEST/LATEST sentinels,
  `toString`.
- **`ClientSegmentLayoutTest`** (7 tests) — `fromProto` filters active
  segments, sorts by hash-range start, builds segment topic names,
  collects broker URLs, propagates the controller URL (plus TLS
  variant), and the returned collections are immutable.
- **`CheckpointV5Test`** (11 tests) — regular checkpoint roundtrip
  with multi-segment positions, empty positions, EARLIEST/LATEST
  sentinel roundtrip (sentinel identity preserved, 1-byte compact
  wire), timestamp variant roundtrip + wire shape, null / empty /
  unknown-type input rejection, immutable positions map.

### Already exercised by prior PRs in this stack

- `PulsarClientImplTest.createSegmentProducerAsync*` — on #25565 (the
  small v4 change we added here is also exercised there).
- Scalable tests for the broker side — 93 tests in #25565 / #25564 /
  #25559.
- Admin-API data-class tests — 24 tests in #25565.
- Protocol-command roundtrip tests — 8 tests in #25564.

## Test plan

- [x] `./gradlew :pulsar-client-v5:test --tests
      "org.apache.pulsar.client.impl.v5.*"` — 48/48 new tests pass.
- [x] Full compile of all impacted modules
      (`pulsar-client-v5`, `pulsar-client-original`, `pulsar-broker`,
      `pulsar-common`, `pulsar-client-admin-api`).
- [x] Checkstyle clean on all modules touched.

### DIFF
--- a/pulsar-client-v5/src/main/java/org/apache/pulsar/client/impl/v5/MessageIdV5.java
+++ b/pulsar-client-v5/src/main/java/org/apache/pulsar/client/impl/v5/MessageIdV5.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.client.impl.v5;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.pulsar.client.api.v5.MessageId;
+
+/**
+ * V5 MessageId implementation that wraps a v4 MessageId and includes a position
+ * vector across all segments.
+ *
+ * <p>The position vector captures the latest delivered message ID per segment at the
+ * moment this message was dequeued. This enables correct cumulative acknowledgment:
+ * when the application acknowledges this message, all segments are advanced to the
+ * positions recorded in the vector, not just the segment this message came from.
+ *
+ * <p>For non-cumulative consumers (QueueConsumer) and readers (CheckpointConsumer),
+ * only the single segment ID and v4 message ID are needed; the position vector is
+ * empty.
+ */
+public final class MessageIdV5 implements MessageId {
+
+    static final long NO_SEGMENT = -1;
+
+    static final MessageIdV5 EARLIEST = new MessageIdV5(
+            org.apache.pulsar.client.api.MessageId.earliest, NO_SEGMENT, Map.of());
+    static final MessageIdV5 LATEST = new MessageIdV5(
+            org.apache.pulsar.client.api.MessageId.latest, NO_SEGMENT, Map.of());
+
+    private final org.apache.pulsar.client.api.MessageId v4MessageId;
+    private final long segmentId;
+
+    /**
+     * Position vector: snapshot of the latest delivered message ID per segment,
+     * taken at the moment this message was delivered to the application.
+     * Used by StreamConsumer for cumulative ack across all segments.
+     */
+    private final Map<Long, org.apache.pulsar.client.api.MessageId> positionVector;
+
+    /**
+     * Create a MessageIdV5 with a position vector for cumulative ack support.
+     */
+    public MessageIdV5(org.apache.pulsar.client.api.MessageId v4MessageId,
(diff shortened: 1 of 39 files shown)


pulsar-client-v5/src/main/java/org/apache/pulsar/client/impl/v5/MessageIdV5.java
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * V5 MessageId implementation that wraps a v4 MessageId and includes a position
+ * vector across all segments.
+ *
+ * <p>The position vector captures the latest delivered message ID per segment at the
+ * moment this message was dequeued. This enables correct cumulative acknowledgment:
+ * when the application acknowledges this message, all segments are advanced to the
+ * positions recorded in the vector, not just the segment this message came from.
+ *
```
lhotari, line 35: It would be useful to clarify this since it leaves a lot of questions about how this could work.

Does a StreamConsumer allow any parallelism?

Is this message id a client side state or does the message id get assigned already on the broker? Is the state mutable or immutable? Most likely this id isn't stable. Just wondering if the concept of a message id is correct if the id changes. It seems more like an acknowledgement token. Is that right?

We already hold additional mutable state in the BatchMessageIdImpl.ackSet field. Is this approach something similar where the state is piggy backed on the message id on the client side?

UPDATE: I found the answer in ScalableStreamConsumer (it's client side state). I'll leave these questions just in case it helps to clarify the javadoc here for future readers. :)

lhotari (approved):
LGTM as a starting point. Just one comment about possible improvement to explain the MessageIdV5 in the javadoc.

pulsar-client-v5/src/main/java/org/apache/pulsar/client/impl/v5/MessageIdV5.java
merlimat: The state it's only on the client side, used for cumulative acks. 

Since you have a single `StreamingConsumer` object, potentially reading from multiple segments, when you do a cumulativeAck on `m1` we need to ack all the messages that the application has seen, from all segments, so this map is just reflecting the last message received from each segment before this particular message.

PR #25574, opened 2026-04-24
[feat][broker] PIP-470: Close inactive topics without deleting their data

codelipenghui:
PIP: #TBD (see `pip/pip-470.md` in this PR)

### Motivation

Deployments with very large numbers of mostly-idle topics suffer broker memory pressure and per-topic metrics cardinality, even when they explicitly do **not** want the topic data deleted. Today the only mechanisms are:

1. `brokerDeleteInactiveTopicsEnabled` — detects inactivity the right way, but deletes the data. Not acceptable when data must be retained.
2. Manual `pulsar-admin topics unload` driven by an external cron — reimplements the broker's inactivity detection outside the broker.

This PR adds a first-class broker option that reuses the existing inactivity detection and **closes (unloads)** the inactive topic instead of deleting it. The topic's BookKeeper ledgers, subscriptions, cursors and policies are all preserved; only the in-memory `PersistentTopic`/`NonPersistentTopic` and its broker-cache entry are released. The next producer/consumer reconnect transparently reloads the topic.

### Modifications

- **`ServiceConfiguration`**: new dynamic config `brokerCloseInactiveTopicsEnabled` (default `false`).
- **`PulsarService.start()`**: fail-fast validation — `brokerDeleteInactiveTopicsEnabled` and `brokerCloseInactiveTopicsEnabled` are mutually exclusive.
- **`BrokerService.startInactivityMonitor()`**: schedule `checkGC` when **either** flag is on.
- **`AbstractTopic.isCloseWhileInactive()`**: broker-level helper reading the new flag.
- **`PersistentTopic.checkGC()`** / **`NonPersistentTopic.checkGC()`**: when close mode is active (and delete is not), after the same inactivity and replication-producer checks as today, call `close(true, false)`. That path already disconnects clients, closes the managed ledger without deleting ledgers, and removes the topic from the broker cache via `disposeTopic` → `removeTopicFromCache`. The retention-window guard is bypassed in the close branch because it exists only to prevent data loss, which is moot when nothing is deleted.
- Partitioned-topic metadata cleanup is not run in close mode (nothing is deleted).
- Added **`pip/pip-470.md`** describing motivation, design, and interactions.

Detection inputs are unchanged and reused: `brokerDeleteInactiveTopicsMode`, `brokerDeleteInactiveTopicsFrequencySeconds`, `brokerDeleteInactiveTopicsMaxInactiveDurationSeconds`.

### Verifying this change

- **New** `InactiveTopicCloseTest`:
  - `testCloseInactiveTopicKeepsDataAndEvictsFromCache` — produces a message, drops the subscription, waits for the inactivity window, asserts the topic is evicted from the broker cache (`getTopicReference` empty) but still listed in metadata, and a fresh consumer on a new subscription reads the original message from earliest — proving data is preserved.
  - `testActiveTopicIsNotClosed` — topic with an active subscription must not be closed.
  - `testMutualExclusionWithDeleteInactive` — broker startup fails when both flags are enabled.
- **Existing** `InactiveTopicDeleteTest` (20 tests) — all pass; delete-path behavior is unchanged.

- [x] The default values of configurations — a new configuration `brokerCloseInactiveTopicsEnabled` is introduced with default `false`; no existing configuration default is changed.

### Documentation

- [x] `doc-required` — when this feature merges, the inactive-topic-management docs should be updated to describe the close-without-delete mode alongside the existing delete mode.

PR in forked repository: https://github.com/codelipenghui/incubator-pulsar/tree/penghui/jolly-nightingale-e2e980


pip/pip-470.md
```diff
@@ -0,0 +1,301 @@
...
+`brokerDeleteInactiveTopicsEnabled` is true. With this PIP, the monitor also
+runs when `brokerCloseInactiveTopicsEnabled` is true.
+
+`PersistentTopic.checkGC()` and `NonPersistentTopic.checkGC()` gain a branch:
+if close mode is enabled (and delete mode is not), after the same
+isActive / duration / replication-producer checks pass, the topic calls its
+own `close(disconnectClients=true, closeWithoutWaitingClientDisconnect=false)`
+path. That path already:
+
+- disconnects producers and consumers,
```
poorbarcode: Since the inactivation only happens when the topic is inactive, which confirms no producer registered, the inactivation will not disconnect producers.

pip/pip-470.md
```diff
@@ -0,0 +1,301 @@
...
+(unload) instead of a delete when the topic is determined to be inactive.
+
+# Goals
+
+## In Scope
+
+- A new broker configuration `brokerCloseInactiveTopicsEnabled` (default
+  `false`) that, when enabled, causes the inactivity monitor to **close**
+  (unload) topics determined to be inactive rather than delete them.
+- Mutual exclusion with `brokerDeleteInactiveTopicsEnabled`: only one of the
```
poorbarcode, line 72: We'd better add a new limitation: the new feature only `brokerDeleteInactiveTopicsMode` is `delete_when_no_subscriptions`. Otherwise, the active consumer will load up the topic again after an inactivation, which leads to frequent invalid unloading and loading up

poorbarcode: Updated here https://github.com/apache/pulsar/pull/25574/changes/26ba4eec4291caf31428d101f7dd00459e794f82#diff-5edf14cc6f25857d0cfdd26b2d3b3141230ecfb0dfa95aebf7583fd76ede4c4bR3620-R3630

poorbarcode (approved):

PR #25576, opened 2026-04-24
[fix][client] Fix bug Improper Neutralization in TransactionMetaStoreHandler (CWE-74)

### LINKED ISSUE #25324: [Bug] Improper Neutralization in TransactionMetaStoreHandler (CWE-74)
### Search before reporting

### Read release policy

### User environment

The pulsar version is the newest version in the master branch

### 1. Description
A potential injection vulnerability exists in the `TransactionMetaStoreHandler.toStringSubscriptionList` method. The method constructs a string representation of a subscription list by directly formatting `topic` and `subscription` fields using `java.lang.String.format` without any escaping or neutralization of special characters (e.g., spaces, newlines, or delimiters).

### 2. Vulnerable Code Snippet
In `TransactionMetaStoreHandler.java`, the fields are concatenated into a single string:

```java
// File: pulsar-client/.../TransactionMetaStoreHandler.java
private String toStringSubscriptionList(List<Subscription> list) {
    // ... logic for null/empty ...
    StringBuilder builder = new StringBuilder("[");
    for (Subscription subscription : list) {
        // VULNERABILITY: Raw strings are formatted without escaping
        builder.append(String.format("%s %s", subscription.getTopic(), subscription.getSubscription()));
    }
    return builder.append("]").toString();
}
```

This string is then used to create a `description` for a transaction operation:

```java
// Line 210 in addSubscriptionToTxn
String description = String.format("Add subscription %s to TXN %s", 
    toStringSubscriptionList(subscriptionList), String.valueOf(txnID));
```

### 3. Attack Scenario
Because `topic` and `subscription` names can often be influenced by external clients in Apache Pulsar:
1. **Log Injection**: An attacker could provide a subscription name containing newline characters (`\n`) and fake log entries (e.g., `\n[INFO] Transaction 123 committed successfully`). If this `description` is logged, it can deceive administrators.
2. **Structural Ambiguity**: If a topic name contains a space, the resulting `[Topic Subscription]` string becomes ambiguous, potentially misleading downstream components or monitoring tools that parse this description.

### 4. Suggested Fix
Implement a defensive "Escape" or "Neutralization" strategy. Special characters in the components should be sanitized or the entire list should be serialized using a standard, safe format (like JSON) or a custom escaper.

```java
// Suggested Fix using simple character replacement or a utility
builder.append(String.format("[%s : %s]", 
    sanitize(subscription.getTopic()), 
    sanitize(subscription.getSubscription())));
```

### 5. Risk Assessment
- **CWE-74**: Improper Neutralization of Special Elements in Output.
- **Confidence Score**: 7/10
- **Severity**: Low/Medium (Primarily affects auditing, logging, and monitoring integrity).

### Error messages

```text

```

### Reproducing the issue

See Issue Description

### Additional information

See Issue Description

PavelZeger:
Fixes https://github.com/apache/pulsar/issues/25324

## Motivation

`TransactionMetaStoreHandler.toStringSubscriptionList()` builds a string from user-controlled `topic` and `subscription` fields without sanitizing them. This violates CWE-74 (Injection) — malicious input containing special characters such as newlines (`\n`), tabs (`\t`), or brackets could forge log entries or corrupt exception messages that surface to callers.

## Modifications

- Applied `StringEscapeUtils.escapeJava()` from Apache Commons Text (already a transitive dependency) to sanitize `topic` and `subscription` fields in `toStringSubscriptionList()`
- Replaced manual `StringBuilder` concatenation with `StringJoiner` for cleaner output formatting

## Verifying this change

- Added `testAddSubscriptionToTxnDescriptionSanitized` to verify that non-printable characters (`\n`, `\t`) in topic/subscription fields are escaped in the output
- Added `testAddSubscriptionToTxnDescriptionBracketInjection` to verify that bracket injection attempts in topic fields do not corrupt the output structure

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-client/build.gradle.kts
+++ b/pulsar-client/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
         exclude(group = "io.netty")
     }
     implementation(libs.commons.lang3)
+    implementation(libs.commons.text)
     implementation(libs.asynchttpclient)
     implementation(libs.netty.reactive.streams)
     implementation(libs.slog)
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TransactionMetaStoreHandler.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TransactionMetaStoreHandler.java
@@ -18,6 +18,7 @@
  */
 package org.apache.pulsar.client.impl;
 
+import static org.apache.commons.text.StringEscapeUtils.escapeJava;
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.buffer.ByteBuf;
 import io.netty.util.Recycler;
@@ -29,6 +30,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.List;
+import java.util.StringJoiner;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
@@ -393,7 +395,7 @@ public class TransactionMetaStoreHandler extends HandlerState
     }
 
     public CompletableFuture<Void> addSubscriptionToTxn(TxnID txnID, List<Subscription> subscriptionList) {
-            log.debug().attr("subscription", subscriptionList).attr("txn", txnID).log("Add subscription to txn.");
+        log.debug().attr("subscription", subscriptionList).attr("txn", txnID).log("Add subscription to txn.");
 
         CompletableFuture<Void> callback = new CompletableFuture<>();
         if (!canSendRequest(callback)) {
@@ -403,7 +405,7 @@ public class TransactionMetaStoreHandler extends HandlerState
         ByteBuf cmd = Commands.newAddSubscriptionToTxn(
                 requestId, txnID.getLeastSigBits(), txnID.getMostSigBits(), subscriptionList);
         String description = String.format("Add subscription %s to TXN %s", toStringSubscriptionList(subscriptionList),
-                String.valueOf(txnID));
+                txnID);
         OpForVoidCallBack op = OpForVoidCallBack.create(cmd, callback, client, description, cnx());
         internalPinnedExecutor.execute(() -> {
             pendingRequests.put(requestId, op);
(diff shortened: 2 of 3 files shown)


lhotari (commented):
The changes are fine, but #25324 doesn't make much sense. The TransactionMetaStoreHandler is a client side class. If an attacker would use invalid input on the client side, they would get the invalid log output in their own logs. 

In Pulsar, we haven't clearly defined the security model. However, Pulsar clients are assumed to be trusted. There are multiple attack vectors in the Pulsar binary protocol for example related to DoS attacks. Sending invalid input to the broker could result in invalid log output.

@merlimat has recently migrated the Pulsar code base to use https://github.com/merlimat/slog for logging. One efficient way to handle logging issues would be to add a feature to slog which escapes java.lang.String (or Object.toString()) attributes. This way there wouldn't be a need to add other type of escaping to shield from log injection attacks.

lhotari:
I created https://github.com/merlimat/slog/issues/4 as a feature request to slog.

PavelZeger:
> I created [merlimat/slog#4](https://github.com/merlimat/slog/issues/4) as a feature request to slog.

Should I close this PR if so,@lhotari?

lhotari:
> Should I close this PR if so,@lhotari?

@PavelZeger Yes, I think this should be closed.

It looks like log4j2 already supports [`%enc{%msg}` and `%enc{%X}`](https://logging.apache.org/log4j/2.3.x/manual/layouts.html#:~:text=enc%7Bpattern%7D%0Aencode%7Bpattern%7D) to prevent such attacks.

One option would be to add a value `text-escaped` to `pulsar.log.format` so that someone who is concerned about log forging could enable it by setting the environment variable `PULSAR_LOG_FORMAT=text-escaped`.

In the default log configuration, the handling of `text-escaped` would be added here:
https://github.com/apache/pulsar/blob/ae2b8c76cec9dfc240fcbf6ae323f42d0ccbcc28/conf/log4j2.yaml#L78-L86

According to Claude, Log4j's PatternLayout will output the exception after the log line unless the pattern explicitly contains `%ex` or `%throwable`. Since exception messages could also contain input, the exception would have to be escaped as well if there's a need to prevent any log forging issues.

Something like this would encode the log message, the MDC context and the exception stacktrace (including exception messages):
`%d{ISO8601_OFFSET_DATE_TIME_HHMM} [%t] %-5level %logger{36} - %enc{%msg} %enc{%X} %enc{%throwable}%n`

The `%enc` feature in Log4j should be a sufficient way to handle [OWASP Log Injection/Forging](https://owasp.org/www-community/attacks/Log_Injection) and [CWE-117: Improper Output Neutralization for Logs](https://cwe.mitre.org/data/definitions/117.html). Since it comes with the tradeoff that log readability suffers, it shouldn't be the default. For Pulsar 5.0, we can tackle this by adding the escaping feature to slog.

Would you like to create a separate PR for the `text-escaped` format? It would be useful to test it whether it works as expected.

In addition, it would be helpful to document this somewhere in the Pulsar documentation, which is maintained at https://github.com/apache/pulsar-site. The challenge might be finding a suitable location in the existing documentation.

PR #4683, opened 2019-07-07
Cleanup tests in the presto module

vzhikserg:
### Modifications

Cleanup in the presto module which includes the following changes:
* Use the preferred way of the schema's creation. 
* Predicates and Functions were converted to lambda
* Add static import statements for Assert to improve the readability of the tests

PR #15908, opened 2022-06-02
Enable TCP/IP keepalive for all ZK client connections in all components started with bin/pulsar

lhotari:
### Motivation

This is a follow-up on #12982 which enabled TCP/IP keepalive for ZK server side.
This PR enables TCP/IP keepalive for all ZK client connections in all components

- Enable TCP keepAlive flag on the sockets used for client connections
  - see https://issues.apache.org/jira/browse/ZOOKEEPER-3712 for details
    - "Some broken network infrastructure may lose the FIN packet that is sent from closing client.
        These never closed client sockets cause OS resource leak. Enabling this option terminates
        these zombie sockets by idle check. ... Currently this option is only available
        when default NIOServerCnxnFactory is used."

### Modifications

add `-Dzookeeper.clientTcpKeepAlive=true` to `OPTS` in `bin/pulsar` script so that the setting is passed to all components that are started using `bin/pulsar` script.

PR #15818, opened 2022-05-27
Configure DLog Bookie, Pulsar, and Admin clients via pass through config

michaeljmarshall:
### Motivation

The Pulsar broker, proxy, websocket proxy, and function worker each start several clients, including the Pulsar Client, the Pulsar Admin Client, and the Bookkeeper Client. Until now, there hasn't been an easy way to pass arbitrary configuration to these clients, which can cause problems. A recent example is the client memory limit that we fixed here: https://github.com/apache/pulsar/pull/15752. I propose that we enable pass through configuration for these clients, so that users can override any default configuration.

Enable pass through configuration support for the bookkeeper clients created by `DistributedLog` and `BookKeeperPackagesStorage`. This support is already present for the managed ledger bookkeeper client, which takes all `bookkeeper_` prefixed configs. In order to simplify configuration, I reused the `bookkeeper_` prefix since it's reasonable to assume that these configs will be the same.

Enable pass through configuration support for the broker clients using the `brokerClient_` prefix. Note that setting this once affects all clients. I think this is ideal, but it could lead to some confusion. A good secondary feature might be to add another prefix that targets specific clients, like the replication client.

Initially implemented as https://github.com/apache/pulsar/pull/15763, but instead of submitting many PRs, I am just submitting one larger one.

### Modifications

* Update all Broker, Proxy, WebSocket Proxy, and Function Worker clients so they can take arbitrary config. The Pulsar Client and the Pulsar Admin Client take configuration starting with `brokerClient_`. The Bookkeeper Client and the DLog Bookeeper Client take configuration starting with `bookkeeper_`.
* Create `PropertiesUtils` class to reduce code duplication for mapping prefixed properties to the `Map<String, Object>`
* Add `PulsarAdminBuilder#loadConf` method to the interface to match the same method in the Pulsar Client builder.
* Update the `BookKeeperPackagesStorageConfiguration` to expose the underlying `properties` object.
* Add documentation to the `broker.conf`, the `proxy.conf`, the `websocket.conf` and the `functions_worker.yml` files.

### Verifying this change

I added several tests and also manually verified that the configuration overrides applied.

### Alternative Implementation

I had initially wanted to run the override logic at the end of all other configuration. However, it doesn't work correctly because of the issue highlighted here https://github.com/apache/pulsar/issues/8509. It might be helpful to change the implementation of the `loadConf` logic so that it doesn't serialize secrets as `****`.

### Does this pull request potentially affect one of the following parts:

[...]

PR #25580, opened 2026-04-27
[feat][broker]:Support broker configuration for BookKeeper client TCP keep-alive options

wolfstudy:
Fixes #xyz

### Motivation

Apache BookKeeper [PR #4683](https://github.com/apache/bookkeeper/pull/4683) introduced TCP
keep-alive configuration options for the BookKeeper client, namely `tcpKeepIdle`,
`tcpKeepIntvl` and `tcpKeepCnt`. This change has been released in BookKeeper `4.17.3`.

Now that the BookKeeper client exposes these TCP keep-alive tuning knobs, the Pulsar Broker should also expose matching
configuration entries, so that operators can tune the TCP keep-alive behavior of the broker's
BookKeeper client without having to patch code or rely on OS-wide defaults.

This is particularly useful in the following scenarios:
- Detecting and recovering from half-open connections (e.g. Bookie host crash, NAT/LB silently
  dropping idle connections) faster than the OS default (which is typically 2 hours on Linux).
- Environments where network devices aggressively close idle TCP connections, causing the
  broker's BookKeeper client to unexpectedly hit broken connections under low traffic.
- Fine-grained tuning of keep-alive probes to balance between fast failure detection and
  extra network overhead.

### Modifications

Added three new broker configuration entries under `ServiceConfiguration`, and wired them
through `BookKeeperClientFactoryImpl` into the underlying BookKeeper `ClientConfiguration`:

| Broker config                 | BookKeeper client setter | Default | Semantics                                                                 |
|-------------------------------|--------------------------|---------|---------------------------------------------------------------------------|
| `tcpKeepAliveTimeSeconds`     | `setTcpKeepIdle`         | `0`     | Idle time before the first keep-alive probe is sent (seconds).            |
| `tcpKeepAliveIntervalSeconds` | `setTcpKeepIntvl`        | `0`     | Interval between subsequent keep-alive probes (seconds).                  |
| `tcpKeepAliveProbeCount`      | `setTcpKeepCnt`          | `0`     | Number of unacknowledged probes before the connection is considered dead. |

Behavior details:
- The default value `0` means "do not override" — the broker will fall back to the BookKeeper
  client default (`-1`), which in turn falls back to the OS-level TCP keep-alive settings.
- A value is only applied to the BookKeeper `ClientConfiguration` when it is strictly greater
  than `0`, so existing deployments are fully backward compatible.

### Verifying this change

- Added `BookKeeperClientFactoryImplTest#testSetTcpKeepAliveConfiguration`, which asserts:
  - When none of the new configs are set, the resulting BookKeeper `ClientConfiguration`
    returns the BookKeeper defaults (`getTcpKeepIdle() == -1`, `getTcpKeepIntvl() == -1`,
    `getTcpKeepCnt() == -1`).
  - When the new configs are set (e.g. `60 / 10 / 5`), the corresponding values are correctly
    forwarded to `ClientConfiguration` (`getTcpKeepIdle() == 60`, `getTcpKeepIntvl() == 10`,
    `getTcpKeepCnt() == 5`).

- [x] The public API — added three new broker configuration entries
  (`tcpKeepAliveTimeSeconds`, `tcpKeepAliveIntervalSeconds`, `tcpKeepAliveProbeCount`).
  These are additive and default to `0` (no behavior change unless explicitly configured),
  so existing deployments are not affected.

### Documentation

- [x] `doc-not-needed`
  The new fields carry inline `@FieldContext` documentation that is automatically surfaced
  in the generated broker reference documentation.

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -357,6 +357,31 @@ public class ServiceConfiguration implements PulsarConfiguration {
             + "(0 to disable limiting)")
     private int maxHttpServerConnections = 2048;
 
+    @FieldContext(
+            category = CATEGORY_SERVER,
+            doc = "TCP keepalive time in seconds. This is the duration between the last data packet sent "
+                    + "(simple ACKs are not considered data) and the first keepalive probe. "
+                    + "Default is 7200 seconds (2 hours). "
+                    + "If set to 0, the system default value will be used."
+    )
+    private int tcpKeepAliveTimeSeconds = 0;
+
+    @FieldContext(
+            category = CATEGORY_SERVER,
+            doc = "TCP keepalive interval in seconds. This is the duration between subsequent keepalive probes, "
+                    + "if no acknowledgement is received from the peer. Default is 75 seconds. "
+                    + "If set to 0, the system default value will be used."
+    )
+    private int tcpKeepAliveIntervalSeconds = 0;
+
+    @FieldContext(
+            category = CATEGORY_SERVER,
+            doc = "TCP keepalive probe count. This is the number of unacknowledged probes to send before considering "
+                    + "the connection dead and notifying the application layer. Default is 9 probes. "
+                    + "If set to 0, the system default value will be used."
+    )
+    private int tcpKeepAliveProbeCount = 0;
+
     @FieldContext(category = CATEGORY_SERVER, doc =
             "Gzip compression is enabled by default. Specific paths can be excluded from compression.\n"
                     + "There are 2 syntaxes supported, Servlet url-pattern based, and Regex based.\n"
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -357,31 +357,6 @@ public class ServiceConfiguration implements PulsarConfiguration {
             + "(0 to disable limiting)")
     private int maxHttpServerConnections = 2048;
 
-    @FieldContext(
-            category = CATEGORY_SERVER,
-            doc = "TCP keepalive time in seconds. This is the duration between the last data packet sent "
-                    + "(simple ACKs are not considered data) and the first keepalive probe. "
-                    + "Default is 7200 seconds (2 hours). "
-                    + "If set to 0, the system default value will be used."
-    )
-    private int tcpKeepAliveTimeSeconds = 0;
-
-    @FieldContext(
-            category = CATEGORY_SERVER,
-            doc = "TCP keepalive interval in seconds. This is the duration between subsequent keepalive probes, "
-                    + "if no acknowledgement is received from the peer. Default is 75 seconds. "
-                    + "If set to 0, the system default value will be used."
-    )
-    private int tcpKeepAliveIntervalSeconds = 0;
-
-    @FieldContext(
-            category = CATEGORY_SERVER,
-            doc = "TCP keepalive probe count. This is the number of unacknowledged probes to send before considering "
-                    + "the connection dead and notifying the application layer. Default is 9 probes. "
-                    + "If set to 0, the system default value will be used."
-    )
-    private int tcpKeepAliveProbeCount = 0;
-
     @FieldContext(category = CATEGORY_SERVER, doc =
             "Gzip compression is enabled by default. Specific paths can be excluded from compression.\n"
                     + "There are 2 syntaxes supported, Servlet url-pattern based, and Regex based.\n"
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1256,6 +1256,28 @@ managedLedgerDefaultAckQuorum=2
 # the BookkeeperPackagesStorage bookkeeper client), except the distributed log bookkeeper client.
 # The dlog bookkeeper client is configured in the functions worker configuration file.
 
+# TCP keep-alive settings for the BookKeeper client connections from the broker to bookies.
+# These options were introduced in BookKeeper 4.17.3 (see apache/bookkeeper#4683) and are
+# forwarded to the BookKeeper client via the generic "bookkeeper_" prefix mechanism above.
+# A value of -1 (BookKeeper default) defers to the operating system's TCP keep-alive settings.
+# Uncomment and adjust to enable finer-grained keep-alive probing from the broker.
+#
+# Note: TCP keep-alive is a two-sided mechanism. Tuning these on the broker (BookKeeper
+# client) side only affects connections initiated by the broker. To complete the picture
+# you should also tune the bookie node's OS-level keep-alive settings via sysctl, e.g.
+# net.ipv4.tcp_keepalive_time / tcp_keepalive_intvl / tcp_keepalive_probes, since the
+# BookKeeper server (bookie) currently does not expose application-level keep-alive knobs.
+
+# Idle time (in seconds) before the first TCP keep-alive probe is sent on an otherwise idle
+# BookKeeper client connection.
+#bookkeeper_tcpKeepIdle=300
+# Interval (in seconds) between subsequent TCP keep-alive probes when no acknowledgement is
+# received from the bookie.
+#bookkeeper_tcpKeepIntvl=60
+# Number of unacknowledged TCP keep-alive probes before the connection is considered dead
+# and closed.
+#bookkeeper_tcpKeepCnt=5
+
 # How frequently to flush the cursor positions that were accumulated due to rate limiting. (seconds).
 # Default is 60 seconds
 managedLedgerCursorPositionFlushSeconds=60
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/BookKeeperClientFactoryImpl.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/BookKeeperClientFactoryImpl.java
@@ -138,15 +138,6 @@ public class BookKeeperClientFactoryImpl implements BookKeeperClientFactory {
         bkConf.setDiskWeightBasedPlacementEnabled(conf.isBookkeeperDiskWeightBasedPlacementEnabled());
         bkConf.setMetadataServiceUri(conf.getBookkeeperMetadataStoreUrl());
         bkConf.setLimitStatsLogging(conf.isBookkeeperClientLimitStatsLogging());
-        if (conf.getTcpKeepAliveTimeSeconds() > 0) {
-            bkConf.setTcpKeepIdle(conf.getTcpKeepAliveTimeSeconds());
-        }
-        if (conf.getTcpKeepAliveIntervalSeconds() > 0) {
-            bkConf.setTcpKeepIntvl(conf.getTcpKeepAliveIntervalSeconds());
-        }
-        if (conf.getTcpKeepAliveProbeCount() > 0) {
-            bkConf.setTcpKeepCnt(conf.getTcpKeepAliveProbeCount());
-        }
 
         if (!conf.isBookkeeperMetadataStoreSeparated()) {
             // If we're connecting to the same metadata service, with same config, then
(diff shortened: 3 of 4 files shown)


liudezhi2098 (approved):

hanmz (approved):
lgtm

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -357,6 +357,31 @@ public class ServiceConfiguration implements PulsarConfiguration {
...
+    )
+    private int tcpKeepAliveIntervalSeconds = 0;
+
+    @FieldContext(
+            category = CATEGORY_SERVER,
+            doc = "TCP keepalive probe count. This is the number of unacknowledged probes to send before considering "
+                    + "the connection dead and notifying the application layer. Default is 9 probes. "
+                    + "If set to 0, the system default value will be used."
+    )
+    private int tcpKeepAliveProbeCount = 0;
```
lhotari: These setting are specific to Pulsar. 
There's already a solution to configure BookKeeper client options by prefixing the options with `bookkeeper_` in `broker.conf`. For Pulsar Helm chart, that would be `PULSAR_PREFIX_bookkeeper_`.

https://github.com/apache/pulsar/blob/33fe7559b4d1a2cfb9d35756d38f7bad72b54309/conf/broker.conf#L1253-L1257

For example, this would be the way to configure BookKeeper client TCP keep-alive options for Pulsar Broker:
```
"PULSAR_PREFIX_bookkeeper_tcpKeepIdle": "300"
"PULSAR_PREFIX_bookkeeper_tcpKeepIntvl": "60"
"PULSAR_PREFIX_bookkeeper_tcpKeepCnt": "5"
```

lhotari (changes requested):
The settings in ServiceConfiguration are specific to Pulsar. 
There's already a solution to configure BookKeeper client options by prefixing the options with `bookkeeper_` in `broker.conf`. For Pulsar Helm chart, that would be `PULSAR_PREFIX_bookkeeper_`.

https://github.com/apache/pulsar/blob/33fe7559b4d1a2cfb9d35756d38f7bad72b54309/conf/broker.conf#L1253-L1257

For example, this would be the way to configure BookKeeper client TCP keep-alive options for Pulsar Broker:
```
"PULSAR_PREFIX_bookkeeper_tcpKeepIdle": "300"
"PULSAR_PREFIX_bookkeeper_tcpKeepIntvl": "60"
"PULSAR_PREFIX_bookkeeper_tcpKeepCnt": "5"
```

It could be reasonable to set these as defaults (or commented out lines) in broker.conf (without `PULSAR_PREFIX_`) and document the settings. Could you please modify this PR to handle it this way?

lhotari:
Sidenote: there's been a proposal to add keepalive to Pulsar components, but that was rejected in #14841. For broker connections, we rely on Pulsar's application level keep-alive Ping/Pong commands. The gap in Ping/Pong was addressed in #15382.

lhotari:
> Now that the BookKeeper client exposes these TCP keep-alive tuning knobs, the Pulsar Broker should also expose matching
> configuration entries, so that operators can tune the TCP keep-alive behavior of the broker's
> BookKeeper client without having to patch code or rely on OS-wide defaults.

To solve the issues where connections stall, it's usually necessary to configure TCP keep-alive on both sides: the client and the server. Keep-alive is enabled by default for BookKeeper client and server, but they do rely on OS defaults. 

It will anyways be necessary to tune OS defaults since the BookKeeper server doesn't have a way to configure the keep-alive parameters. In cloud managed k8s nodes, the settings have reasonable values by default at least in GCP.

```
# sysctl -a |grep keepalive
net.ipv4.tcp_keepalive_intvl = 60
net.ipv4.tcp_keepalive_probes = 5
net.ipv4.tcp_keepalive_time = 300
```

wolfstudy:
> Sidenote: there's been a proposal to add keepalive to Pulsar components, but that was rejected in #14841. For broker connections, we rely on Pulsar's application level keep-alive Ping/Pong commands. The gap in Ping/Pong was addressed in #15382.


Thanks @lhotari reply, the two PRs mentioned here address a different issue than the one this current PR aims to resolve.
The primary objective of the current PR is to resolve connection interruptions that occur in physical machine deployment environments. These interruptions stem from the fact that Bookkeeper and Broker instances are deployed on separate physical nodes, and—due to intervening firewall restrictions—rely on the operating system's default keep-alive settings, which have an excessively long timeout duration.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
wolfstudy: Indeed, a similar mechanism exists in `BookKeeperClientFactoryImpl.java`; the specific code is as follows:
```
PropertiesUtils.filterAndMapProperties(conf.getProperties(), "bookkeeper_")
        .forEach((key, value) -> {
            bkConf.setProperty(key, value);
        });
```

This mechanism appears to be a simpler implementation; I will remove the corresponding configuration options.

lhotari:
> > Sidenote: there's been a proposal to add keepalive to Pulsar components, but that was rejected in #14841. For broker connections, we rely on Pulsar's application level keep-alive Ping/Pong commands. The gap in Ping/Pong was addressed in #15382.
> 
> Thanks @lhotari reply, the two PRs mentioned here address a different issue than the one this current PR aims to resolve. The primary objective of the current PR is to resolve connection interruptions that occur in physical machine deployment environments. These interruptions stem from the fact that Bookkeeper and Broker instances are deployed on separate physical nodes, and—due to intervening firewall restrictions—rely on the operating system's default keep-alive settings, which have an excessively long timeout duration.

Yes, that was a sidenote. Please check what I provided in the example. It's already possible to tune the keep-alive for BookKeeper client. In addition, you will need to anyways tune BookKeeper server separately to complete the solution (the BookKeeper server side can only be tuned with OS level parameters).

Configuring BookKeeper client options in a Pulsar broker k8s deployment:
```
"PULSAR_PREFIX_bookkeeper_tcpKeepIdle": "300"
"PULSAR_PREFIX_bookkeeper_tcpKeepIntvl": "60"
"PULSAR_PREFIX_bookkeeper_tcpKeepCnt": "5"
```

The `bookkeeper_` prefix solution was added in #9232 and documented in #15818. Please read my comments in the review.

wolfstudy:
> > Now that the BookKeeper client exposes these TCP keep-alive tuning knobs, the Pulsar Broker should also expose matching
> > configuration entries, so that operators can tune the TCP keep-alive behavior of the broker's
> > BookKeeper client without having to patch code or rely on OS-wide defaults.
> 
> To solve the issues where connections stall, it's usually necessary to configure TCP keep-alive on both sides: the client and the server. Keep-alive is enabled by default for BookKeeper client and server, but they do rely on OS defaults.
> 
> It will anyways be necessary to tune OS defaults since the BookKeeper server doesn't have a way to configure the keep-alive parameters. In cloud managed k8s nodes, the settings have reasonable values by default at least in GCP.
> 
> ```
> # sysctl -a |grep keepalive
> net.ipv4.tcp_keepalive_intvl = 60
> net.ipv4.tcp_keepalive_probes = 5
> net.ipv4.tcp_keepalive_time = 300
> ```

Thanks for the context! A quick clarification on a couple of points:
1. BookKeeper server now exposes keep-alive tuning. Since BookKeeper 4.17.3 (via apache/bookkeeper#4683), both the BK client and the bookie server expose tcpKeepIdle, tcpKeepIntvl, and tcpKeepCnt as configuration options. Pulsar already pulls in BookKeeper 4.17.3, so tuning can be done at the application level on both ends — we no longer have to rely on OS defaults for the bookie server side.
2. Relying on OS defaults is not always viable. The typical Linux kernel defaults are tcp_keepalive_time=7200s (2h), tcp_keepalive_intvl=75s, tcp_keepalive_probes=9, which means a broken connection can go undetected for ~2h 11min. That's too long for a messaging system where broker↔bookie liveness matters. While GCP GKE ships with saner defaults (time=300, intvl=60, probes=5), this is not guaranteed across other environments:
- EKS / AKS / on-prem clusters often keep the 7200s default.
- Tuning net.ipv4.tcp_keepalive_* via sysctl requires privileged pods or node-level DaemonSets, which many operators either cannot or don't want to deploy.
- OS-level settings are global and affect every TCP socket on the node, whereas the BookKeeper-level settings are scoped to BK connections only.
3. What this PR actually does. It doesn't add new Pulsar-specific config; it just documents how to forward these three BK client options through the existing bookkeeper_ prefix mechanism in broker.conf, so operators have a discoverable, Pulsar-idiomatic way to tune broker↔bookie keep-alive without needing node-level privileges or changing OS defaults. The defaults remain -1 (fall back to OS), so behavior is unchanged unless explicitly opted in.
So the motivation isn't "OS defaults aren't enough"; it's "give operators a per-application, documented knob that works uniformly across environments, now that BookKeeper 4.17.3 makes this possible."

wolfstudy:
> In addition, you will need to anyways tune BookKeeper server separately to complete the solution (the BookKeeper server side can only be tuned with OS level parameters).

Yes, the client side is already fully tunable today via the existing bookkeeper_ prefix mechanism and
the bookie side has to be tuned at the OS level anyway.

The primary goal of this PR is to address precisely the kind of logic you described.

However, it must be said that perhaps making the `SO_KEEPALIVE` setting on the Bookie Server side configurable as well would constitute a more complete implementation.

lhotari (approved):
LGTM

lhotari:
> 2\. Relying on OS defaults is not always viable. The typical Linux kernel defaults are tcp_keepalive_time=7200s (2h), tcp_keepalive_intvl=75s, tcp_keepalive_probes=9, which means a broken connection can go undetected for ~2h 11min. That's too long for a messaging system where broker↔bookie liveness matters. While GCP GKE ships with saner defaults (time=300, intvl=60, probes=5), this is not guaranteed across other environments:
> 
> * EKS / AKS / on-prem clusters often keep the 7200s default.
> * Tuning net.ipv4.tcp_keepalive_* via sysctl requires privileged pods or node-level DaemonSets, which many operators either cannot or don't want to deploy.
> * OS-level settings are global and affect every TCP socket on the node, whereas the BookKeeper-level settings are scoped to BK connections only.

I think that we should update Pulsar documentation to recommend to adjust the TCP Keepalive OS defaults for Pulsar deployments. There shouldn't be a reason why they shouldn't be adjusted to values what GCP uses by default (time=300, intvl=60, probes=5). Besides BookKeeper, Zookeeper needs TCP keepalive settings. We do enable TCP keepalive for Zookeeper, but there isn't a way to tune the settings. To fully close the gap, there's also a need to cover ZooKeeper client and server besides the BookKeeper client and server.

wolfstudy:
> > 2. Relying on OS defaults is not always viable. The typical Linux kernel defaults are tcp_keepalive_time=7200s (2h), tcp_keepalive_intvl=75s, tcp_keepalive_probes=9, which means a broken connection can go undetected for ~2h 11min. That's too long for a messaging system where broker↔bookie liveness matters. While GCP GKE ships with saner defaults (time=300, intvl=60, probes=5), this is not guaranteed across other environments:
> > 
> > * EKS / AKS / on-prem clusters often keep the 7200s default.
> > * Tuning net.ipv4.tcp_keepalive_* via sysctl requires privileged pods or node-level DaemonSets, which many operators either cannot or don't want to deploy.
> > * OS-level settings are global and affect every TCP socket on the node, whereas the BookKeeper-level settings are scoped to BK connections only.
> 
> I think that we should update Pulsar documentation to recommend to adjust the TCP Keepalive OS defaults for Pulsar deployments. There shouldn't be a reason why they shouldn't be adjusted to values what GCP uses by default (time=300, intvl=60, probes=5). Besides BookKeeper, Zookeeper needs TCP keepalive settings. We do enable TCP keepalive for Zookeeper, but there isn't a way to tune the settings. To fully close the gap, there's also a need to cover ZooKeeper client and server besides the BookKeeper client and server.

Good ideas, I will continue to push these matters forward following this PR.

lhotari:
> Besides BookKeeper, Zookeeper needs TCP keepalive settings. We do enable TCP keepalive for Zookeeper, but there isn't a way to tune the settings. To fully close the gap, there's also a need to cover ZooKeeper client and server besides the BookKeeper client and server.

regarding Zookeeper, there are significant gaps in setting TCP keepalive for Zookeeper TCP connections. I added a comment to https://github.com/apache/pulsar/pull/15908#issuecomment-4334026899.

PR #25581, opened 2026-04-27
[fix][broker] Decrement unacked counter when removeAllUpTo removes pending acks

nodece:
### Motivation                                                                                                                                
                                                                                                                                                 
When the cursor's mark-delete position advances (e.g., via message expiry), `readMoreEntries` calls `PendingAcksMap.removeAllUpTo` to remove entries from the consumer's pendingAcks without decrementing the consumer's `unackedMessages` counter, causing it to be permanently inflated.                                                                             
                                                                                                                                              
This is especially problematic with batch index level ack enabled: partial acks leave residual unacked counts in the counter, and after `removeAllUpTo` removes the entry from pendingAcks, subsequent individual batch acks can no longer find the entry to decrement, so the counter never reaches zero.                                                                       
                                                                                                                                                 
### Modifications                                                                                                                             
                                                                                                                                              
- `PendingAcksMap.removeAllUpTo` accepts a callback invoked per removed entry                                                                 
- `Consumer. removePendingAcksUpToPositionAndDecrementUnacked` computes the remaining unacked count per entry (respecting batch index level ack) and decrements the counter                                                                                                                                 
- `PersistentDispatcherMultipleConsumers.readMoreEntries` uses the new method instead of directly calling `removeAllUpTo`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.concurrent.atomic.LongAdder;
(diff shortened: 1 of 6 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -1121,6 +1121,34 @@ public PendingAcksMap getPendingAcks() {
...
+     *
+     * @param markDeleteLedgerId the ledger ID up to which to remove pending acks
+     * @param markDeleteEntryId the entry ID up to which to remove pending acks
+     */
+    public void removePendingAcksUpToAndCountUnacked(long markDeleteLedgerId, long markDeleteEntryId) {
+        if (pendingAcks == null) {
+            return;
+        }
+
+        int[] totalUnacked = {0};
```
lhotari: In other locations, we use MutableInt for such purpose. Using an array is slightly confusing.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -1121,6 +1121,34 @@ public PendingAcksMap getPendingAcks() {
...
+     * counter by the remaining unacked count for each removed entry.
+     *
+     * <p>This is used when the cursor's mark-delete position advances past entries that are still in the consumer's
+     * pending acks. The remaining unacked count accounts for batch index level acknowledgments — only the truly
+     * unacked batch indexes are decremented.
+     *
+     * @param markDeleteLedgerId the ledger ID up to which to remove pending acks
+     * @param markDeleteEntryId the entry ID up to which to remove pending acks
+     */
+    public void removePendingAcksUpToAndCountUnacked(long markDeleteLedgerId, long markDeleteEntryId) {
```
lhotari: instead of "removePendingAcksUpToAndCountUnacked", this could be called "removePendingAcksUpToPositionAndDecrementUnacked" so that the name clearly expresses the intent.

lhotari (approved):
LGTM, just 2 minor comments. Good work @nodece

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
```diff
@@ -1132,19 +1133,20 @@ public PendingAcksMap getPendingAcks() {
...
      * @param markDeleteEntryId the entry ID up to which to remove pending acks
      */
-    public void removePendingAcksUpToAndCountUnacked(long markDeleteLedgerId, long markDeleteEntryId) {
+    public void removePendingAcksUpToPositionAndDecrementUnacked(long markDeleteLedgerId, long markDeleteEntryId) {
         if (pendingAcks == null) {
             return;
         }
 
-        int[] totalUnacked = {0};
+        AtomicInteger atomicTotalUnacked = new AtomicInteger(0);
```
lhotari: `org.apache.commons.lang3.mutable.MutableInt` would be sufficient in this use case since there's no asynchronous or concurrent handling of the pending acks.

lhotari (approved):

PR #25582, opened 2026-04-27
[fix][cli] Load client.conf timeouts in CLI tools

### LINKED ISSUE #22808: [Bug] [cli] Pulsar-client cli doesn't support timeout values given in apache pulsar reference website
### Search before asking

### Read release policy

### Version

OS: Linux, macOS
Java: 17
Pulsar Client and Broker: 3.2.2

### Minimal reproduce step

- Update the `conf/client.conf` file with `lookupTimeoutMs`, `operationTimeoutMs` and other timeout values described [here](https://pulsar.apache.org/reference/#/3.2.x/client/client-configuration-client)
- Run `pulsar-perf` or `pulsar-client` and observe the logs.
- We get this info log from `org.apache.pulsar.client.impl.ConsumerStatsRecorderImpl` where the timeout values doesn't change with the configuration.

> `"operationTimeoutMs":30000,"lookupTimeoutMs":30000`

The timeout values getting picked up from the `conf/client.conf`.

### What did you see instead?

The timeout values are always default values.

### Anything else?

This works fine with Java client. The pulsar-cli tools have this issue.

Praveenkumar76:
Fixes [apache#22808](https://github.com/apache/pulsar/issues/22808)

### Motivation

Pulsar CLI tools (`pulsar-client` and `pulsar-perf`) are not picking up timeout configurations from `conf/client.conf`.

Even when `lookupTimeoutMs` and `operationTimeoutMs` are explicitly set, the CLI continues to use default values:

- `"operationTimeoutMs": 30000`
- `"lookupTimeoutMs": 30000`

This is because these properties are not being properly loaded and applied to the internal `ClientBuilder`.

As a result, user-defined timeout configurations are ignored at runtime, unlike the Java client where the same settings work as expected.

### Modifications

- Loaded `operationTimeoutMs` and `lookupTimeoutMs` explicitly from `Properties`
- Parsed and applied them to:
  - `clientBuilder.operationTimeout(...)`
  - `clientBuilder.lookupTimeout(...)`
- Added basic validation (`isNotBlank`, `trim`) before parsing
- Handled invalid values using `try-catch (NumberFormatException)`
- Logged a warning via `commander.getErr().println()` and fallback to defaults when parsing fails

### Verifying this change

- Added `testTimeoutConfigurationLoading` in `PulsarClientToolTest`
- Injected custom timeout values (55000, 65000) via `Properties`
- Verified that `ClientConfigurationData` reflects the configured values instead of defaults

- [ ] The public API  
- [ ] The schema  
- [ ] The threading model  
- [ ] The binary protocol  
- [ ] The REST endpoints  
- [x] The admin CLI options  
- [ ] The metrics  

### DIFF
--- a/conf/client.conf
+++ b/conf/client.conf
@@ -92,8 +92,11 @@ webserviceTlsProvider=
 proxyServiceUrl=
 
 #Proxy protocol to select type of routing at proxy
-proxyProtocol=
+#proxyProtocol=
 
 # Pulsar Admin Custom Commands
 #customCommandFactoriesDirectory=commandFactories
 #customCommandFactories=
+
+operationTimeoutMs=5000
+lookupTimeoutMs=3000
--- a/conf/client.conf
+++ b/conf/client.conf
@@ -102,4 +102,4 @@ proxyServiceUrl=
 operationTimeoutMs=30000
 
 # Timeout for lookup operations
-lookupTimeoutMs=30000
+lookupTimeoutMs=-1
--- a/conf/client.conf
+++ b/conf/client.conf
@@ -98,5 +98,8 @@ proxyServiceUrl=
 #customCommandFactoriesDirectory=commandFactories
 #customCommandFactories=
 
-operationTimeoutMs=5000
-lookupTimeoutMs=3000
+# Timeout for client operations
+operationTimeoutMs=30000
+
+# Timeout for lookup operations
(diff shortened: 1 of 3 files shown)


nodece:
I think https://github.com/apache/pulsar/pull/25245 can be used instead of this pr.

PR #25586, opened 2026-04-27
[feat] PIP-468: V5 client end-to-end smoke test + segment-bypass for internal subscribe/read

merlimat:
## Summary

First end-to-end test for the PIP-468 V5 client, plus the small plumbing
fix it surfaced.

### Plumbing fix — segment-domain bypass for V5 internals

The V4 client rejects \`segment://\` topics for end-user subscribe / read
paths (added in #25573 so the segment domain stays internal to the V5
SDK). But the V5 \`ScalableQueueConsumer\` / \`ScalableStreamConsumer\` /
\`ScalableCheckpointConsumer\` themselves build per-segment v4 consumers
and readers and were tripping over the same check via the public
\`ConsumerBuilder\` / \`ReaderBuilder\` paths.

Two new internal-only entry points on \`PulsarClientImpl\` mirror the
existing \`createSegmentProducerAsync\` escape hatch:

- \`subscribeSegmentAsync(ConsumerConfigurationData, Schema)\`
- \`createSegmentReaderAsync(ReaderConfigurationData, Schema)\`

Both validate the config (single topic, valid topic name, sub /
\`startMessageId\` set), then skip the scalable-domain check and dispatch
to the existing single-topic codepaths. The three V5 internal classes
build a config object directly and invoke the bypass instead of going
through the v4 builder.

### End-to-end test infra

- \`pulsar-broker\` test scope picks up \`pulsar-client-v5\` /
  \`pulsar-client-api-v5\` so V5 end-to-end tests can live next to the
  other client integration tests.
- New \`V5ClientBaseTest\` extending \`SharedPulsarBaseTest\`:
  - \`newV5Client()\` — V5 PulsarClient against the shared in-memory
    cluster.
  - \`newScalableTopic(int)\` — creates a \`topic://\` scalable topic in
    the per-test namespace via the admin API.
  - \`track(...)\` registers producers / consumers / clients for ordered
    cleanup in \`@AfterMethod\`.
- New \`V5SmokeTest\` exercises the full happy path on a 1-segment
  scalable topic:
  admin \`createScalableTopic\` → \`DagWatchClient\` session → per-segment
  producer create → wire send → per-segment consumer attach → receive
  → ack.

This is the start of a phased E2E test suite. Subsequent PRs will add
per-consumer-type basic tests (Stream / Queue / Checkpoint), schema
matrix, producer / consumer feature coverage, and (separately, on a
multi-broker base class) controller leader-election / redirect /
rebalance scenarios.

## Test plan

- [x] \`./gradlew :pulsar-broker:test --tests
      "org.apache.pulsar.client.api.v5.V5SmokeTest"\` passes (1/1).
- [x] \`./gradlew :pulsar-broker:checkstyleTest
      :pulsar-client-v5:checkstyleMain
      :pulsar-client-original:checkstyleMain\` clean.
- [x] Pre-existing V5 unit tests (48 tests) and broker scalable tests
      still pass.

### DIFF
--- a/pulsar-broker/build.gradle.kts
+++ b/pulsar-broker/build.gradle.kts
@@ -111,6 +111,8 @@ dependencies {
     testImplementation(project(path = ":pulsar-package-management:pulsar-package-core", configuration = "testJar"))
     testImplementation(libs.bookkeeper.common) { artifact { classifier = "tests" } }
     testImplementation(libs.zookeeper) { artifact { classifier = "tests" } }
+    testImplementation(project(":pulsar-client-v5"))
+    testImplementation(project(":pulsar-client-api-v5"))
     testImplementation(project(":pulsar-functions:pulsar-functions-local-runner-original"))
     testImplementation(project(":pulsar-functions:pulsar-functions-api-examples"))
     testImplementation(project(":pulsar-io:pulsar-io-batch-discovery-triggerers"))
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5ClientBaseTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5ClientBaseTest.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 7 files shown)


lhotari (approved):
LGTM

PR #25589, opened 2026-04-28
[fix][broker] Fix race in pending acks removal in redeliverUnacknowledgedMessages

nodece:
### Motivation

`Consumer.redeliverUnacknowledgedMessages` has race conditions with the expire path (`readMoreEntries → removeAllUpTo`), which can corrupt the `unackedMessages` counter:

* The epoch-based variant resets the counter to 0, while expire may concurrently decrement it to a negative value.
* The list-based variant performs `get` and `remove` separately, allowing expire to remove entries in between and causing incorrect redelivery counting.

### Modifications

* Make redelivery operations atomic with respect to expire:

  * Add `PendingAcksMap.forEachAndClear()` to atomically iterate and clear all entries under a single lock.
  * Add `PendingAcksMap.removeAndGet()` to atomically get and remove entries.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Consumer.java
@@ -1155,15 +1155,12 @@ public class Consumer {
     }
 
     public void redeliverUnacknowledgedMessages(long consumerEpoch) {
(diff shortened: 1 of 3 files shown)


lhotari (approved):
LGTM. Good catch! Just thinking that the usage of "expire" is slightly misleading in the title since it could be confused with "message expiration". An alternative title could be "Fix race in pending acks removal in redeliverUnacknowledgedMessages"

PR #10667, opened 2021-05-21
Fix issues in advanceNonDurableCursors

jerrypeng:
### Motivation

There are a couple of issues in advanceNonDurableCursors

1. Durable cursors unnecessarily run through this logic
2. For Non-durable and durable cursors that are caught up reading the topic.  The following warning message can be printed

```
message: [public/default/persistent/test] Failed to mark delete while trimming data ledgers: Invalid mark deleted position
```

This happens because highestPositionToDelete calculated when the reader/consumer is caught up is going to be higher than the last confirmed for the of the managed ledger.  There is also no point in trying to advance the mark delete position if the reader/consumer is caught up.

PR #25592, opened 2026-04-28
[fix][broker] Move pending acks cleanup to selected mark-delete callbacks

nodece:
### Motivation

Pending-ack cleanup was previously executed in the `readMoreEntries` dispatch path. This introduced unnecessary overhead in a hot path and also risked concurrent access to the same pending-ack entries from both dispatch and ack threads.
﻿
Cleanup is only required when messages are removed through mark-delete advancing operations such as expiry, skip, or clear-backlog. Ack-triggered mark-delete operations are much more frequent and should avoid carrying additional cleanup work.

### Modifications

* Introduced atomic remaining-unacked tracking for both non-batch and batch messages.

  * `PendingAcksMap` now stores and updates the actual `remainingUnacked` count instead of relying on batch size–based accounting.
  * Added atomic helper methods such as `updateRemainingUnacked` and `removePendingAckAndGet` to simplify and centralize unacked-message accounting.
  * Removed legacy logic that recalculated unacked counts from cursor state or batch size.

* Refactored pending-ack cleanup to leverage the existing `markDeletePositionMoveForward` hook.

  * Removed pending-ack cleanup from the `readMoreEntries` dispatch path.
  * Cleanup of pending acks and redelivery state is now triggered only after mark-delete advancement operations such as:

    * message expiry
    * skip entries
    * clear backlog
  * Ack-triggered paths do not invoke `markDeletePositionMoveForward`; they continue using `afterAckMessages` to avoid additional overhead on the high-frequency ack path.

* Updated pending-ack iteration and removal paths to consistently use the maintained `remainingUnacked` count directly, improving correctness and concurrency handling for partial batch acknowledgments.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Dispatcher.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Dispatcher.java
@@ -131,6 +131,15 @@ public interface Dispatcher {
         //No-op
     }
 
+    /**
+     * This hook is invoked after cursor mark-delete operations triggered by
+     * message removal flows such as expiry, skip, or clear backlog, but not for
+     * regular ack-driven mark-delete operations due to their higher frequency.
+     *
+     * <p>Since the cursor ack set may no longer be available after mark-delete,
+     * the cleanup logic relies on the remaining unacked count stored in
+     * {@code PendingAcksMap} entries.
+     */
     default void markDeletePositionMoveForward() {
         // No-op
     }
@@ -140,23 +149,6 @@ public interface Dispatcher {
      */
     default boolean checkAndUnblockIfStuck() {
         return false;
-    }
-
-    /**
-     * Prune pending-ack entries up to the specified position and updates the
-     * consumer unacked-message counters accordingly.
-     *
-     * <p>This hook is invoked after the cursor mark-delete operation completes
-     * (for example, during message expiry, skip, or clear backlog). Since the
-     * cursor ack set may no longer be available after mark-delete, the counter
-     * adjustment relies on the remaining unacked count stored in the
-     * {@code PendingAcksMap} entries.
-     *
-     * @param ledgerId the ledger ID of the inclusive upper bound position
-     * @param entryId the entry ID of the inclusive upper bound position
-     */
-    default void prunePendingAcksUpToPosition(long ledgerId, long entryId) {
-        // No-op by default
     }
 
     /**
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Dispatcher.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Dispatcher.java
@@ -143,6 +143,23 @@ public interface Dispatcher {
     }
 
     /**
+     * Prune pending-ack entries up to the specified position and updates the
+     * consumer unacked-message counters accordingly.
+     *
+     * <p>This hook is invoked after the cursor mark-delete operation completes
+     * (for example, during message expiry, skip, or clear backlog). Since the
+     * cursor ack set may no longer be available after mark-delete, the counter
+     * adjustment relies on the remaining unacked count stored in the
+     * {@code PendingAcksMap} entries.
+     *
+     * @param ledgerId the ledger ID of the inclusive upper bound position
+     * @param entryId the entry ID of the inclusive upper bound position
+     */
+    default void prunePendingAcksUpToPosition(long ledgerId, long entryId) {
+        // No-op by default
+    }
+
+    /**
      * A callback hook after acknowledge messages.
      * @param exOfDeletion the ex of {@link org.apache.bookkeeper.mledger.ManagedCursor#asyncDelete},
      *              {@link ManagedCursor#asyncClearBacklog} or {@link ManagedCursor#asyncSkipEntries)}.
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -589,10 +589,14 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
     }
 
     @Override
-    public void prunePendingAcksUpToPosition(long ledgerId, long entryId) {
-        redeliveryMessages.removeAllUpTo(ledgerId, entryId);
-        for (Consumer consumer : consumerList) {
-            consumer.removePendingAcksUpToPositionAndDecrementUnacked(ledgerId, entryId);
+    public void markDeletePositionMoveForward() {
+        Position markDeletePosition = cursor.getMarkDeletedPosition();
+        if (markDeletePosition != null) {
+            redeliveryMessages.removeAllUpTo(markDeletePosition.getLedgerId(), markDeletePosition.getEntryId());
+            for (Consumer consumer : consumerList) {
+                consumer.removePendingAcksUpToPositionAndDecrementUnacked(
+                        markDeletePosition.getLedgerId(), markDeletePosition.getEntryId());
+            }
         }
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -145,7 +145,6 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
     protected enum ReadType {
         Normal, Replay
     }
-    private Position lastMarkDeletePositionBeforeReadMoreEntries;
     private volatile long readMoreEntriesCallCount;
 
     public PersistentDispatcherMultipleConsumers(PersistentTopic topic, ManagedCursor cursor,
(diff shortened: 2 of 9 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java
```diff
@@ -211,9 +212,18 @@ public void markDeleteComplete(Object ctx) {
...
+                Dispatcher dispatcher = subscription.getDispatcher();
+                if (dispatcher != null) {
+                    Position mdPos = cursor.getMarkDeletedPosition();
+                    if (mdPos != null) {
+                        dispatcher.prunePendingAcksUpToPosition(mdPos.getLedgerId(), mdPos.getEntryId());
+                    }
+                    // If the subscription is a Key_Shared subscription, we should to trigger message dispatch.
+                    if (subscription.getType() == SubType.Key_Shared) {
+                        dispatcher.markDeletePositionMoveForward();
+                    }
```
lhotari: instead of adding a separate `prunePendingAcksUpToPosition`, leave the pruning as an implementation of the dispatcher. Modify the existing `markDeletePositionMoveForward` so that it contains the markDeletedPosition argument. In the implementation, the pruning can be handled for Key_Shared.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/Dispatcher.java
```diff
@@ -142,6 +142,23 @@ default boolean checkAndUnblockIfStuck() {
...
+     * cursor ack set may no longer be available after mark-delete, the counter
+     * adjustment relies on the remaining unacked count stored in the
+     * {@code PendingAcksMap} entries.
+     *
+     * @param ledgerId the ledger ID of the inclusive upper bound position
+     * @param entryId the entry ID of the inclusive upper bound position
+     */
+    default void prunePendingAcksUpToPosition(long ledgerId, long entryId) {
+        // No-op by default
+    }
```
lhotari: instead of adding a separate `prunePendingAcksUpToPosition`, leave the pruning as an implementation of the dispatcher. Modify the existing `markDeletePositionMoveForward` so that it contains the markDeletedPosition argument. In the implementation, the pruning can be handled for Key_Shared.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -362,17 +361,6 @@ public synchronized void readMoreEntries() {
         // increment the counter for readMoreEntries calls, to track the number of times readMoreEntries is called
         readMoreEntriesCallCount++;
 
-        // remove possible expired messages from redelivery tracker and pending acks
-        Position markDeletePosition = cursor.getMarkDeletedPosition();
-        if (lastMarkDeletePositionBeforeReadMoreEntries != markDeletePosition) {
```
lhotari, line 367: Another detail is that trimming (based on retention) doesn't currently trigger the callbacks at all. Trimming is handled by the org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl#internalTrimLedgers method which calls advanceCursorsIfNecessary (btw. The javadoc of the advanceCursorsIfNecessary method is misleading. For some reason, the javadoc added in https://github.com/apache/pulsar/pull/10667/changes hasn't made it to the code).

Trimming based on retention should also be handled before we can remove this code.

lhotari: Even the current code doesn't currently handle trimming correctly since trimming should trigger state cleanup proactively. However, it would cleanup state correctly. (In some cases that's sufficient, in some cases it's not)

lhotari: There should be a callback for ManagedCursor for deletions that happen by retention trimming.

nodece: I think retention trimming is unrelated here.

Messages only become eligible for retention trimming after they have already been removed by ack, expiry, skip, or clear-backlog operations. Those are already the paths where pending-acks cleanup happens now.

lhotari: @nodece you're right that retention trimming isn't the path I should have flagged. I read through the code and made an analysis with Claude Code. There's a remaining gap in BacklogQuotaManager handling.

Assisted analysis of `internalTrimLedgers` by Claude Code:

- The slowest-cursor heap that drives `cursors.getSlowestCursorPosition()` only contains **durable** cursors. [`ManagedLedgerImpl#addCursor`](https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L744-L753) passes `positionForOrdering = null` for non-durable cursors, and [`ManagedCursorContainerImpl#add`](https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorContainerImpl.java#L85-L103) skips adding to the heap when position is null.
- So the slowest durable cursor's mark-delete position *defines* the trim boundary in [`internalTrimLedgers`](https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3008-L3044), and `ledgersToDelete` only contains ledgers strictly before it (with the boundary case when the slowest is on the last entry of a ledger, allowing that ledger to be deleted).
- In [`advanceCursorsIfNecessary`](https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3241-L3275), the strict-greater check `highestPositionToDelete > cursor.getMarkDeletedPosition()` is therefore always false for any durable cursor (their mark-delete is by definition ≥ the slowest durable cursor's, which already bounds `highestPositionToDelete`). Only **non-durable** cursors — Reader/compaction-style — can fall behind and be advanced here. The misleading rename of the method from `advanceNonDurableCursors` plus the missing javadoc from #10667 is what threw me off.

So retention trimming doesn't move durable subscription mark-delete forward, and the PR doesn't need to handle it. My review comment was wrong on that point.

---

Assisted analysis of BacklogQuotaManager:

That said, while verifying this I noticed a related gap on the **`BacklogQuotaManager`** side that I think the PR does miss. The new `Dispatcher#markDeletePositionMoveForward` hook is wired in three places:

| Source | Location | Triggering API |
|---|---|---|
| Subscription clear-backlog | [`PersistentSubscription#clearBacklog`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L780) | `cursor.asyncClearBacklog(...)` |
| Subscription skip-messages | [`PersistentSubscription#skipMessages`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L823) | `cursor.asyncSkipEntries(...)` |
| Expiry monitor | [`PersistentMessageExpiryMonitor#findEntryComplete`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java#L235) | `cursor.asyncMarkDelete(...)` |

But [`BacklogQuotaManager`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java) advances durable subscription cursors directly through the `ManagedCursor` API, bypassing the `PersistentSubscription` wrappers:

| BQM path | Advancement call | Hook fires? |
|---|---|---|
| [`dropBacklogForSizeLimit`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java#L190) | `slowestConsumer.skipEntries(messagesToSkip, IndividualDeletedEntries.Include)` — directly on `ManagedCursor`, bypassing `PersistentSubscription#skipMessages` | ❌ No |
| [`dropBacklogForTimeLimit` precise=true](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java#L232) | `subscription.getExpiryMonitor().expireMessages(target)` | ✅ Yes (via expiry monitor) |
| [`dropBacklogForTimeLimit` precise=false](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java#L255-L265) | `slowestConsumer.markDelete(nextPosition)` — directly on `ManagedCursor`, no subscription-level callback | ❌ No |

So two of the three quota-eviction paths advance the durable subscription mark-delete without firing the new cleanup hook. Pre-PR these were silently cleaned up by the per-dispatch loop in `readMoreEntries` that this PR removes, so it's a regression introduced here. Triggers under `BacklogQuota.RetentionPolicy.consumer_backlog_eviction`. Side-effects when it does trigger:

- `Consumer.pendingAcks` accumulates entries for already-mark-deleted positions until the consumer disconnects.
- `Consumer.unackedMessages` counter drifts (entries the consumer "still owes" but the broker has already discarded).
- `redeliveryMessages` may retain stale entries.

Two ways to close it:

1. Have `BacklogQuotaManager` route through `PersistentSubscription#skipMessages` (and an analogous subscription-level "markDelete forward" method) so the existing callbacks fire, or
2. Have `BacklogQuotaManager` invoke `dispatcher.markDeletePositionMoveForward()` directly after each eviction step.

Option 1 is cleaner architecturally; option 2 is a smaller diff. Either is fine with me — happy to take this in a follow-up if you'd prefer to keep this PR scoped to its current changes.

nodece: Good catch, this is a different path. We can improve this after merging this PR.

lhotari: > Good catch, this is a different path. We can improve this after merging this PR.

@nodece would you also be able to handle the remaining BacklogQuotaManager part?

nodece: @lhotari Yes, I will continue to improve this part.

lhotari: I created #25613 to track this.

lhotari (approved):
LGTM, follow up issue #25613

PR #25595, opened 2026-04-28
[fix][test] Fix flaky ManagedLedgerTest.testInvalidateReadHandleWhenConsumed

merlimat:
### Motivation

`ManagedLedgerTest.testInvalidateReadHandleWhenConsumed` is flaky on CI with `expected [3] but found [2]` on `assertEquals(ledger.ledgerCache.size(), 3)`.

With `maxEntriesPerLedger=1` and 3 entries, the test expects 4 ledgers (3 closed + 1 current empty). When the read happens before the last ledger roll completes, the final entry is read directly from `currentLedger` instead of via `ledgerCache`, so `ledgerCache.size()` is 2 instead of the expected 3.

Same root cause and fix as #25495 (which fixed the sibling `testInvalidateReadHandleWhenDeleteLedger`).

### Modifications

Wait for `ledger.ledgers.size() == 4` before clearing the entry cache and reading entries.

### Verifying this change

This change is already covered by existing tests.


lhotari (approved):
LGTM

PR #25596, opened 2026-04-28
[fix][test] Fix flaky ExtensibleLoadManagerImplTest.testLoadBalancerServiceUnitTableViewSyncer

merlimat:
### Motivation

`testLoadBalancerServiceUnitTableViewSyncer` is flaky with `expected [true] but found [false]` after the 30s wait for `primaryLoadManager.getServiceUnitStateTableViewSyncer().isActive()`.

The `loadManagerExecutor` is a single-threaded scheduled executor. After `makeSecondaryAsLeader()` followed immediately by `makePrimaryAsLeader()`, the new leader's `playLeader()` is queued behind any still-running `playFollower()` from the prior demotion. `syncer.start()` itself opens both table views and runs `syncExistingItems` + `syncTailItems`. Under CI load this serial chain can exceed 30s.

### Modifications

Increase the timeout from 30s to 60s for the syncer-state checks after both leader transitions, and apply the same to the symmetric pair in the cleanup section after the dynamic config is removed.

### Verifying this change

Locally verified with 3 successful back-to-back runs of the test.


lhotari (approved):
LGTM

PR #25610, opened 2026-04-28
[fix] Client-v5: Retry producer send on AlreadyClosed across segment seal

merlimat:
## Summary

When a scalable-topic segment is sealed (split or merge), the per-segment v4 producer cached inside `ScalableTopicProducer` can be closed by the v4 layer before the V5 client's DAG watch delivers the new layout. A send issued in that window currently surfaces as a bare `PulsarClientException` wrapping `AlreadyClosedException`, even though the operation should be retryable.

Treat `AlreadyClosedException` the same as `TopicTerminatedException` in both the sync and async send paths: drop the stale per-segment producer entry, back off briefly so the DAG watch can deliver the new revision, and re-route on the next attempt. From the application's perspective, sends issued during a layout transition succeed transparently.

`V5ProducerSplitRetryTest` exercises this exactly: produce, split, immediately produce again with no explicit wait for layout convergence, drain via the consumer — every message arrives.

## Test plan
- [x] `./gradlew :pulsar-broker:test --tests \"org.apache.pulsar.client.api.v5.V5ProducerSplitRetryTest\"`

## Matching PR
- [ ] No matching PR.

### area/client

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5ProducerSplitRetryTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5ProducerSplitRetryTest.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM

PR #25624, opened 2026-04-30
[fix][broker] Close pending acks cleanup gap in BacklogQuotaManager

### LINKED ISSUE #25613: Address remaining gap in pending acks cleanup for BacklogQuotaManager
PR #25592 causes a regression in pending acks cleanup which is essential for at least Key_Shared subscriptions. There's a need to cover the gap in pending acks cleanup for BacklogQuotaManager.

---

Assisted analysis of BacklogQuotaManager:

That said, while verifying this I noticed a related gap on the **`BacklogQuotaManager`** side that I think the PR does miss. The new `Dispatcher#markDeletePositionMoveForward` hook is wired in three places:

| Source | Location | Triggering API |
|---|---|---|
| Subscription clear-backlog | [`PersistentSubscription#clearBacklog`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L780) | `cursor.asyncClearBacklog(...)` |
| Subscription skip-messages | [`PersistentSubscription#skipMessages`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentSubscription.java#L823) | `cursor.asyncSkipEntries(...)` |
| Expiry monitor | [`PersistentMessageExpiryMonitor#findEntryComplete`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentMessageExpiryMonitor.java#L235) | `cursor.asyncMarkDelete(...)` |

But [`BacklogQuotaManager`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java) advances durable subscription cursors directly through the `ManagedCursor` API, bypassing the `PersistentSubscription` wrappers:

| BQM path | Advancement call | Hook fires? |
|---|---|---|
| [`dropBacklogForSizeLimit`](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java#L190) | `slowestConsumer.skipEntries(messagesToSkip, IndividualDeletedEntries.Include)` — directly on `ManagedCursor`, bypassing `PersistentSubscription#skipMessages` | ❌ No |
| [`dropBacklogForTimeLimit` precise=true](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java#L232) | `subscription.getExpiryMonitor().expireMessages(target)` | ✅ Yes (via expiry monitor) |
| [`dropBacklogForTimeLimit` precise=false](https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java#L255-L265) | `slowestConsumer.markDelete(nextPosition)` — directly on `ManagedCursor`, no subscription-level callback | ❌ No |

So two of the three quota-eviction paths advance the durable subscription mark-delete without firing the new cleanup hook. Pre-PR these were silently cleaned up by the per-dispatch loop in `readMoreEntries` that this PR removes, so it's a regression introduced here. Triggers under `BacklogQuota.RetentionPolicy.consumer_backlog_eviction`. Side-effects when it does trigger:

- `Consumer.pendingAcks` accumulates entries for already-mark-deleted positions until the consumer disconnects.
- `Consumer.unackedMessages` counter drifts (entries the consumer "still owes" but the broker has already discarded).
- `redeliveryMessages` may retain stale entries.

Two ways to close it:

1. Have `BacklogQuotaManager` route through `PersistentSubscription#skipMessages` (and an analogous subscription-level "markDelete forward" method) so the existing callbacks fire, or
2. Have `BacklogQuotaManager` invoke `dispatcher.markDeletePositionMoveForward()` directly after each eviction step.

Option 1 is cleaner architecturally; option 2 is a smaller diff. Either is fine with me — happy to take this in a follow-up if you'd prefer to keep this PR scoped to its current changes.

_Originally posted by @lhotari in https://github.com/apache/pulsar/pull/25592#discussion_r3159016462_

nodece:
Fix #25613

### Motivation

PR #25592 added `Dispatcher#markDeletePositionMoveForward()` to clean up pending acks and redelivery state when cursors advance via subscription-level APIs (skip, expiry, clear-backlog).

However, `BacklogQuotaManager` bypasses these APIs and directly updates `ManagedCursor`, so the hook is not triggered in two eviction paths:

* Size eviction: `ManagedCursor.skipEntries`
* Time eviction (non-precise): `ManagedCursor.markDelete`

This leads to stale state in `Key_Shared` subscriptions:

* `pendingAcks` not cleaned up
* `unackedMessages` drift

### Modifications

- BacklogQuotaManager
  - Add helper to trigger `Dispatcher#markDeletePositionMoveForward`
  - Invoke after cursor advancement in:
     - size eviction
     - time eviction (non-precise)
   - Extract eviction math into shared methods:
     - `computeEvictionTarget`
     - `computeEntriesToEvict`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java
@@ -19,6 +19,7 @@
 package org.apache.pulsar.broker.service;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
+import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -33,12 +34,14 @@ import org.apache.bookkeeper.mledger.PositionFactory;
 import org.apache.bookkeeper.mledger.proto.ManagedLedgerInfo;
 import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.resources.NamespaceResources;
+import org.apache.pulsar.broker.service.persistent.PersistentSubscription;
 import org.apache.pulsar.broker.service.persistent.PersistentTopic;
 import org.apache.pulsar.broker.service.persistent.PersistentTopicMetrics.BacklogQuotaMetrics;
 import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.policies.data.BacklogQuota;
 import org.apache.pulsar.common.policies.data.BacklogQuota.BacklogQuotaType;
 import org.apache.pulsar.common.policies.data.impl.BacklogQuotaImpl;
+import org.apache.pulsar.common.util.Codec;
 import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.metadata.api.MetadataStoreException;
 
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM

PR #25628, opened 2026-04-30
[fix][sec] Prevent path traversal in PackageName toRestPath

### LINKED ISSUE #25323: [Bug] Potential Path Traversal / Injection in PackageName.toRestPath via String.format
### Search before reporting

### Read release policy

### User environment

The pulsar version is the newest version in the master branch

### 1. Description
A security vulnerability has been identified in the `PackageName.toRestPath()` method. The current implementation uses `String.format` to build a path string without sanitizing the individual fields.

### 2. Vulnerable Code Snippet
In `PackageName.java`, the code is implemented as follows:

```java
public String toRestPath() {
    // The fields (tenant, namespace, etc.) are concatenated without escaping
    return String.format("%s/%s/%s/%s/%s", type, tenant, namespace, name, version);
}
```

### 3. Attack Scenario (PoC)
An attacker can provide a malicious `packageName` to trigger path traversal. For example:
- **Input:** `public/tenant-a/../../system-tenant/ns/pkg@v1`
- **Generated Path:** `public/tenant-a/../../system-tenant/ns/pkg/v1`
- **Resolved Path:** `public/system-tenant/ns/pkg/v1` (Accessing unintended tenant data)

### 4. Suggested Fix
To remediate this, use `URLEncoder` to escape each component before formatting the string. This ensures that characters like `/` or `..` are treated as literal data rather than path instructions.

```java
import java.net.URLEncoder;
import java.nio.charset.StandardCharsets;

public String toRestPath() {
    return String.format("%s/%s/%s/%s/%s", 
        type, 
        URLEncoder.encode(tenant, StandardCharsets.UTF_8),
        URLEncoder.encode(namespace, StandardCharsets.UTF_8),
        URLEncoder.encode(name, StandardCharsets.UTF_8),
        URLEncoder.encode(version, StandardCharsets.UTF_8));
}
```

### 5. Impact
- **CWE-74**: Improper Neutralization of Special Elements in Output.
- **CWE-22**: Improper Limitation of a Pathname to a Restricted Directory ('Path Traversal').
- **Severity**: High (Potential unauthorized access/deletion of package data).

### Error messages

```text

```

### Reproducing the issue

See Attack Scenario (PoC) in Issue Description

### Additional information

See Attack Scenario (PoC) in Issue Description

Praveenkumar76:
Fixes #25323

### Motivation

A potential path traversal vulnerability (CWE-22) exists in `PackageName.toRestPath()`.

The method builds REST path segments using package fields such as tenant, namespace, name, and version. These values were previously concatenated directly into the generated path without URL encoding.

Although current constructor validation blocks malformed input (like extra slashes) in normal flows, relying only on upstream validation violates the principle of Defense in Depth. If object construction is bypassed through reflection, deserialization, future code changes, or alternate call paths, malicious values containing traversal sequences such as `../` could be propagated directly into generated REST paths.

This change applies defense-in-depth by ensuring `toRestPath()` safely encodes path components according to RFC 3986 before constructing the final path.

### Modifications

- Updated `PackageName.toRestPath()` to safely encode:
  - `tenant`
  - `namespace`
  - `name`
  - `version`

- Used Guava's `UrlEscapers.urlPathSegmentEscaper().escape(...)` to construct the components. This correctly follows RFC 3986 for URL paths (converting spaces to `%20` and slashes to `%2F`) and avoids the `+` space-encoding issue caused by standard `URLEncoder`, without requiring new external HTTP dependencies.

- Added a reflection-based unit test (`testPathTraversalBypassConstructor`) to explicitly verify traversal payloads are encoded safely even when constructor validation is bypassed.

### Verifying this change

- Ran package management module tests successfully.
- Added a test that injects traversal-like values (for example `../../`) and verifies the output path is safely encoded.
- Confirmed normal package name inputs continue to behave correctly.

Example verification command:

```bash
./gradlew :pulsar-package-management:pulsar-package-core:test --tests "PackageNameTest"
```

*Does this pull request potentially affect one of the following parts:*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java
+++ b/pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java
@@ -18,6 +18,7 @@
  */
 package org.apache.pulsar.packages.management.core.common;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.google.common.cache.CacheBuilder;
@@ -108,6 +109,18 @@ public class PackageName {
             String.format("%s://%s/%s/%s@%s", type.toString(), tenant, namespace, name, version);
     }
 
+    @VisibleForTesting
+    PackageName(PackageType type, String tenant, String namespace, String name, String version) {
+        this.type = type;
+        this.tenant = tenant;
+        this.namespace = namespace;
+        this.name = name;
+        this.version = version;
+        this.completeName = String.format("%s/%s/%s", tenant, namespace, name);
+        this.completePackageName =
+                String.format("%s://%s/%s/%s@%s", type.toString(), tenant, namespace, name, version);
+    }
+
     public PackageType getPkgType() {
         return this.type;
     }
@@ -137,13 +150,14 @@ public class PackageName {
     }
 
     public String toRestPath() {
-        // Use Guava's urlPathSegmentEscaper to safely encode each segment and prevents Path Traversal (CWE-22)
+        // Use Guava's URL path segment escaper to safely encode each segment and prevent path traversal (CWE-22).
+        var escaper = UrlEscapers.urlPathSegmentEscaper();
         return String.format("%s/%s/%s/%s/%s",
                 type.toString(),
-                UrlEscapers.urlPathSegmentEscaper().escape(tenant),
-                UrlEscapers.urlPathSegmentEscaper().escape(namespace),
-                UrlEscapers.urlPathSegmentEscaper().escape(name),
-                UrlEscapers.urlPathSegmentEscaper().escape(version));
+                escaper.escape(tenant),
+                escaper.escape(namespace),
+                escaper.escape(name),
+                escaper.escape(version));
     }
 
     @Override
--- a/pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java
+++ b/pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java
@@ -23,6 +23,8 @@ import com.google.common.base.Strings;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
@@ -136,7 +138,12 @@ public class PackageName {
     }
 
     public String toRestPath() {
-        return String.format("%s/%s/%s/%s/%s", type, tenant, namespace, name, version);
+        return String.format("%s/%s/%s/%s/%s",
+                type,
+                URLEncoder.encode(tenant, StandardCharsets.UTF_8),
+                URLEncoder.encode(namespace, StandardCharsets.UTF_8),
+                URLEncoder.encode(name, StandardCharsets.UTF_8),
+                URLEncoder.encode(version, StandardCharsets.UTF_8));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


nodece (approved):

merlimat (approved):

pulsar-package-management/core/src/test/java/org/apache/pulsar/packages/management/core/common/PackageNameTest.java
```diff
@@ -116,4 +116,22 @@ public void testPackageNameErrors() {
...
     }
+
+    @Test
+    public void testPathTraversalBypassConstructor() throws Exception {
+        // Create a normal, valid package to bypass the splitter
+        PackageName packageName = PackageName.get("function://tenant-a/ns/name@v1");
+
+        java.lang.reflect.Field tenantField = PackageName.class.getDeclaredField("tenant");
+        tenantField.setAccessible(true);
+        tenantField.set(packageName, "tenant-a/../../system-tenant");
```
lhotari: Reflection should be avoided in tests, please check https://lists.apache.org/thread/7gr04sqmzyttx4ln6ydtp3qv0xgo1o6m.
Instead, package private methods should be used annotationed with `@VisibleForTesting`.

It seems that it's not possible to exploit the "vulnerability" since PackageName.get already handles validation?

pulsar-package-management/core/src/main/java/org/apache/pulsar/packages/management/core/common/PackageName.java
```diff
@@ -136,7 +137,13 @@ public String toString() {
...
 
     public String toRestPath() {
-        return String.format("%s/%s/%s/%s/%s", type, tenant, namespace, name, version);
+        // Use Guava's urlPathSegmentEscaper to safely encode each segment and prevents Path Traversal (CWE-22)
+        return String.format("%s/%s/%s/%s/%s",
+                type.toString(),
+                UrlEscapers.urlPathSegmentEscaper().escape(tenant),
+                UrlEscapers.urlPathSegmentEscaper().escape(namespace),
+                UrlEscapers.urlPathSegmentEscaper().escape(name),
+                UrlEscapers.urlPathSegmentEscaper().escape(version));
```
lhotari: Since `UrlEscapers.urlPathSegmentEscaper()` is repeated, it should be placed in a variable, for example 
`var escaper = UrlEscapers.urlPathSegmentEscaper()` and then used for each invocation of `escape`.

Praveenkumar76: Yes, it violates the DRY principle. I've updated the code to extract `UrlEscapers.urlPathSegmentEscaper()` into a local variable and reused it for the escape calls to avoid redundant invocations.

pulsar-package-management/core/src/test/java/org/apache/pulsar/packages/management/core/common/PackageNameTest.java
Praveenkumar76: Thanks for the reference
I’ve removed reflection from the tests and added a package-private constructor annotated with @VisibleForTesting. Regarding exploitability - agreed that PackageName.get currently blocks this, but this change ensures defense in depth by making toRestPath() safe regardless of how the object is created.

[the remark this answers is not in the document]
Praveenkumar76: yes, this can leak state via the cache. I’ve updated the test to avoid mutating cached instances and instead construct a non-cached instance for isolation.

lhotari (approved):
LGTM

lhotari:
Thanks for the contribution @Praveenkumar76

lhotari:
> chore: trigger CI to retry flaky broker test

To retry CI after it fails due to flaky tests, add a comment `/pulsarbot rerun` instead.

PR #25631, opened 2026-04-30
[feat]PIP-468: Wire transactions into V5 client

merlimat:
## Summary

Wires transactions into the V5 client and lets transactional consumer flows actually run on scalable topics.

- **V5 client wiring** (commit 1): `PulsarClientBuilder.transactionPolicy(...)` enables `enableTransaction` and stores the per-txn timeout; `PulsarClient.newTransaction[Async]()` and a `Transaction` / `AsyncTransaction` API that wraps the v4 transaction. Producer's `MessageBuilder.transaction(txn)` and consumer's `acknowledge(MessageId, Transaction)` route the unwrapped v4 transaction down to the segment producers / cursors. Stream consumer's cumulative ack iterates the position vector and forwards the txn to each segment.
- **Broker pending-ack name fix** (commit 2): `MLPendingAckStore.getTransactionPendingAckStoreSuffix()` was string-concatenating the origin topic with `-<sub>__transaction_pending_ack`. For a `segment://t/n/x/0000-ffff-0` topic that produced an unparseable segment-topic name (`>3` dash-parts in the descriptor); the resulting `IllegalArgumentException` fired synchronously inside `PersistentSubscription`'s ctor (under `openCursorComplete`) and triggered the cursor open's `exceptionally` path on a future that was already removed — NPE in `ManagedLedgerImpl`, subscribe never returned. Map segment topics to a flat persistent pending-ack name in the same namespace, encoding the segment descriptor into the local name.
- **Test infra**: `SharedPulsarCluster` now boots with the transaction coordinator enabled and pre-creates the `pulsar/system` namespace + transaction-coordinator partitioned topic so V5 (and v4) transactional tests can run on the shared broker. Other tests pay no meaningful cost — the coordinator only does work when a client opts in.
- **Coverage**: new `V5TransactionTest` exercises abort state transition, pre/post-commit visibility, aborted messages discarded, multi-topic atomic commit, and transactional consume-and-produce.

## Known limitation

Two further `V5TransactionTest` cases (`testCommitSpansSplit`, `testCommitSpansMerge`) are included but `@Test(enabled = false)` — they document a broker gap: the per-segment `TransactionBuffer` model doesn't compose with scalable-topic layout changes. After split/merge, the parent segments are sealed, but the txn coordinator's `END_TXN_ON_PARTITION` is sent to every segment that received writes; the sealed parent never produces its commit/abort marker and the commit RPC times out (~30s). The fix is structural — transaction tracking needs to live at the scalable-topic level rather than per-segment, with a single decision record that segments consult on read. That's a follow-up; the disabled tests are the regression guard for it.

## Test plan

- [x] `./gradlew :pulsar-broker:test --tests "org.apache.pulsar.client.api.v5.V5TransactionTest"` — 5 enabled tests pass; 2 disabled
- [x] Existing V5 client tests (smoke, queue, stream, checkpoint, split/merge, ...) still pass

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/pendingack/impl/MLPendingAckStore.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/pendingack/impl/MLPendingAckStore.java
@@ -523,7 +523,18 @@ public class MLPendingAckStore implements PendingAckStore {
     }
 
     public static String getTransactionPendingAckStoreSuffix(String originTopicName, String subName) {
-        return TopicName.get(originTopicName) + "-" + subName + SystemTopicNames.PENDING_ACK_STORE_SUFFIX;
+        TopicName origin = TopicName.get(originTopicName);
+        // Segment topics ("segment://tenant/ns/topic/<hexStart>-<hexEnd>-<segmentId>") cannot
+        // host a derived pending-ack topic in the segment domain — the descriptor parser would
+        // reject any name with extra dashes appended. Map to a flat persistent topic in the same
+        // namespace, encoding the segment descriptor into the local name.
+        if (origin.isSegment()) {
+            return String.format("persistent://%s/%s/%s-%s-%s%s",
+                    origin.getTenant(), origin.getNamespacePortion(),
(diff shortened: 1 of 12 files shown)


lhotari (approved):
LGTM

PR #25632, opened 2026-04-30
[feat][broker] PIP-468: Filter scalable topics by property via secondary index

merlimat:
## Summary

Adds the ability to look up scalable topics in a namespace by an exact
`(propertyKey, propertyValue)` pair, backed by the `MetadataStore`
secondary-index API. Native-index stores (Oxia) serve the lookup directly
in O(matches); other stores transparently fall back to a children scan.

### MetadataCache

- New overloads on `MetadataCache`:
    - `create(path, value, indexExtractor)`
    - `readModifyUpdate(path, modifyFn, indexExtractor)`
  
  where `indexExtractor` is `Function<T, Map<String, String>>` returning the
  `indexName -> secondaryKey` map associated with the record.
- `MetadataCacheImpl` overrides both: when an extractor is provided and the
  underlying store is `MetadataStoreExtended`, the write goes through
  `storeExtended.put(path, bytes, version, options, indexes)`. Stores
  without index support transparently fall back to the plain `put`.

### ScalableTopicResources

- Index name = `topic-prop-` + property key, secondary key = property value.
  Each entry of the topic's `properties` map becomes one index entry.
- `createScalableTopicAsync` / `updateScalableTopicAsync` thread the
  extractor through on every write so updates refresh the index.
- New `findScalableTopicsByPropertyAsync(namespace, propertyKey, propertyValue)`
  calls `store.findByIndex` with scan prefix `/topics/<namespace>`. The
  fallback predicate re-deserialises the record and re-checks the property.

### Admin API

- `GET /admin/v2/scalable/{tenant}/{ns}` accepts optional `?propertyKey` +
  `?propertyValue` query params. When both are set the listing is served by
  the index; otherwise the existing children-listing path runs unchanged.
- `ScalableTopics` admin client gains `listScalableTopicsByProperty` (sync +
  async).
- `pulsar-admin scalable-topics list <ns> --property key=value` filters the
  CLI listing the same way.

### Scope

- Scalable topics only — partitioned and non-partitioned managed-ledger
  topics are not touched.
- No backward compat / reindex: only newly-created or updated scalable
  topics are queryable.

## Test plan

- `MetadataCacheSecondaryIndexTest` (pulsar-metadata) — runs the new cache
  overloads against every store implementation (Memory, RocksDB, ZooKeeper,
  MockZooKeeper, **Oxia**). Asserts index queryability, refresh on update,
  and empty-extractor behaviour.
- `ScalableTopicPropertyIndexTest` (pulsar-broker) — focused unit coverage
  for `findScalableTopicsByPropertyAsync`: exact/multi match, namespace
  scoping, empty results, update refresh, and topics with no properties not
  being matched.
- `ScalableTopicsListByPropertyTest` (pulsar-broker) — end-to-end through
  `PulsarAdmin` against a real shared broker.

- area/broker
- area/client-tools

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/ScalableTopicResources.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/ScalableTopicResources.java
@@ -18,7 +18,11 @@
  */
 package org.apache.pulsar.broker.resources;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
@@ -28,6 +32,7 @@ import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.util.Codec;
 import org.apache.pulsar.common.util.FutureUtil;
+import org.apache.pulsar.common.util.ObjectMapperFactory;
 import org.apache.pulsar.metadata.api.MetadataCache;
 import org.apache.pulsar.metadata.api.MetadataStore;
 import org.apache.pulsar.metadata.api.MetadataStoreException;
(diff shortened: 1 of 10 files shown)


lhotari (approved):
LGTM

PR #25633, opened 2026-04-30
[feat] PIP-468: Multi-broker shared cluster + V5 cross-broker tests

merlimat:
## Summary

Add a JVM-wide multi-broker test cluster mirroring `SharedPulsarCluster`, plus
V5 cross-broker coverage for scalable topics. Three brokers share a single
bookie + a real Oxia metadata store (the in-memory store collapses ephemeral
sessions across instances, so leader election doesn't work correctly across
multiple brokers in the same JVM).

### Infrastructure

- `SharedMultiBrokerPulsarCluster` — JVM-wide singleton with `NUM_BROKERS = 3`,
  lazy start, shutdown-hook teardown. Backed by `OxiaContainer` (testcontainers).
  Exposes per-broker `PulsarService` / `PulsarAdmin` / `PulsarClient` lists plus
  broker-0 conveniences.
- `SharedMultiBrokerPulsarBaseTest` — v4 base wrapping the cluster with a fresh
  namespace per test method.
- `V5MultiBrokerClientBaseTest` — V5 layer on top, with one V5 `PulsarClient`
  per broker.

### Tests (`V5MultiBrokerScalableTopicTest`)

- `testProduceConsumeAcrossBrokers` — produce on broker 0's V5 client, consume
  on the last broker's; lookups must route to segment owners.
- `testControllerLeadershipDistributesAcrossBrokers` — cycle which broker
  materializes each new topic's controller; assert leadership lands on every
  broker.
- `testStreamConsumerControllerCoordinationAcrossBrokers` — alice + bob attached
  via two non-leader brokers must reach the controller via the lookup →
  controller-URL path; segments split across the group.
- `testCheckpointConsumerControllerCoordinationAcrossBrokers` — same with
  `consumerGroup(...)`.
- `testSplitFromNonLeaderBrokerRedirectsToOwner` — split request to a non-leader
  broker is 307-redirected to the controller leader.
- `testMergeFromNonLeaderBrokerRedirectsToOwner` — same for merge.

### Broker fixes (uncovered while building the tests)

- `DagWatchSession.buildDagProto` was setting `controllerBrokerUrl` on the
  response object but never propagating it into the `ScalableTopicDAG` proto.
  Without this V5 clients always saw a null controller URL and fell back to
  their configured service URL — on multi-broker that's rarely the controller
  leader, so consumer subscribe lands on the wrong broker. Single-broker tests
  didn't catch it because the fallback URL was always the controller.
- `ScalableTopicController` previously discovered subscriptions for split / merge
  by querying segment topics through `admin.topics().getSubscriptionsAsync(
  segment://...)`, which the v4 Topics REST endpoint doesn't route. Replaced
  with `resources.listSubscriptionsAsync(topicName)` — the scalable-topic
  metadata is the single source of truth for subscriptions; segments are storage
  layout. Single-broker tests didn't catch it because the segment topic was
  always locally cached and short-circuited the admin path.

- area/test
- area/broker

### DIFF
--- a/pulsar-broker/build.gradle.kts
+++ b/pulsar-broker/build.gradle.kts
@@ -132,6 +132,7 @@ dependencies {
     testImplementation(libs.jetty.ee8.proxy)
     testImplementation(libs.jetty.websocket.jetty.client)
     testImplementation(libs.opentelemetry.sdk.testing)
+    testImplementation(libs.oxia.testcontainers)
     testRuntimeOnly(libs.avro.protobuf) {
         exclude(group = "com.google.protobuf")
     }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/DagWatchSession.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/DagWatchSession.java
@@ -169,6 +169,16 @@ public class DagWatchSession {
             }
         }
 
+        // Propagate the controller-broker URL so V5 clients can connect to the right broker
+        // for scalable-topic subscribe. Without this the client falls back to its configured
+        // service URL, which on a multi-broker cluster is rarely the controller leader.
+        if (response.controllerBrokerUrl() != null) {
+            dag.setControllerBrokerUrl(response.controllerBrokerUrl());
+        }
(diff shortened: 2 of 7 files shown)


lhotari (approved):
LGTM

PR #25637, opened 2026-04-30
[feat] PIP-468: Test schema evolution coverage on scalable topics

merlimat:
## Summary

Schemas on a scalable topic are scoped at the parent topic, not per-segment:
`TopicName.getSchemaName()` returns `tenant/ns/encodedLocalName`, and the
segment-domain parser sets `localName = parent` (descriptor is held in separate
`segmentRange` / `segmentId` fields), so every segment of the same scalable
topic shares one schema id — `t/ns/parent`. The schema registry therefore holds
a single version history per scalable topic, and split / merge children inherit
it transparently.

This PR adds tests pinning that contract.

## Tests (`V5SchemaEvolutionTest`)

- **`testSingleSchemaIdAcrossSegments`** — produce v1 on a single-segment topic,
  split into two children, produce v1 again across them.
  `admin.schemas().getAllSchemas` returns exactly one entry. If schemas were
  per-segment we'd see one entry per segment.
- **`testCompatibleSchemaUpgradeAcrossSplit`** — produce v1, split, produce a
  BACKWARD-compatible v2 across the new children. The compatibility check sees
  the pre-split v1 history and accepts v2; the registry ends up with both
  versions; the consumer reads every message (v1 records decode with `j == null`,
  v2 with `j` set).
- **`testIncompatibleSchemaAfterSplitFails`** — same setup but the post-split
  producer attempts a schema with a required-not-nullable extra field
  (BACKWARD-incompatible). The send is rejected, proving the broker still
  consults the pre-split history.

- area/test

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5SchemaEvolutionTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5SchemaEvolutionTest.java
@@ -1,0 +1,278 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


lhotari (approved):
LGTM

PR #25639, opened 2026-04-30
[feat] PIP-468: Filter scalable topic listing by multiple properties (AND)

merlimat:
## Summary

Follow-up to #25632. Replaces the single `(propertyKey, propertyValue)`
filter on the scalable-topic listing API with a `Map<String, String>` of
filters that all must match (AND semantics). Empty map degenerates to the
unfiltered listing.

### Resources

- `ScalableTopicResources.findScalableTopicsByPropertiesAsync(ns, Map<String, String>)`
  picks any single filter to drive the index lookup (native stores narrow
  on it; fallback stores scan), then post-filters with an "all properties
  match" predicate. The post-filter runs both inside `findByIndex` (for
  fallback stores) and in the result-mapping stage, because native-index
  implementations don't apply `fallbackFilter` — on the fallback path the
  re-check is a no-op.

### Admin API

- `GET /admin/v2/scalable/{tenant}/{ns}` now takes repeatable
  `?property=key=value` query params instead of paired `propertyKey` /
  `propertyValue`. Malformed entries (no `=`, empty key, empty value) fail
  with 412.
- `ScalableTopics` admin client renames the API to
  `listScalableTopicsByProperties(ns, Map<String, String>)`. The client
  serialises every entry as a separate `?property=k=v` parameter.
- `pulsar-admin scalable-topics list -p k1=v1 -p k2=v2 ...` accepts
  multiple `--property` options; same key=value parsing as the matching
  `create` subcommand.

## Test plan

- Existing `ScalableTopicPropertyIndexTest` updated to use the new
  signature; gains an AND-semantics case (`owner=alice AND team=platform`
  narrows to a single record even though either filter alone matches more)
  and an empty-filter case (returns full namespace listing).
- `ScalableTopicsListByPropertyTest` end-to-end test exercises the same
  multi-filter combo through `PulsarAdmin`.

- area/broker
- area/client-tools

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/ScalableTopicResources.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/ScalableTopicResources.java
@@ -118,32 +118,57 @@ public class ScalableTopicResources extends BaseResources<ScalableTopicMetadata>
     }
 
     /**
-     * List scalable topics in a namespace whose {@code properties} map contains the given
-     * key/value pair. On stores with native secondary index support (Oxia) this is served
-     * by the index registered at create/update time; otherwise it falls back to a children
-     * scan + per-record property check.
+     * List scalable topics in a namespace whose {@code properties} map contains every
+     * key/value pair in {@code propertyFilters} (AND semantics).
      *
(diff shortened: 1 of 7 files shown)


lhotari (approved):
LGTM

PR #25640, opened 2026-05-01
[fix][test] Fix flaky testGetExcludedBookiesWithIsolationGroups

merlimat:
## Summary

`IsolatedBookieEnsemblePlacementPolicy.initialize()` triggers a fire-and-forget cache load on `bookieMappingCache`. The test calls `getExcludedBookiesWithIsolationGroups` immediately after, sees `cachedRackConfiguration == null`, and gets back an empty set instead of the expected 2 excluded bookies.

Apply the same Awaitility wait pattern used in [#25473](https://github.com/apache/pulsar/pull/25473) for `testBookieInfoChange`: after `onClusterChanged()`, wait for the cache to be populated before asserting.

## Test plan

- [x] Five consecutive local runs pass
- [ ] CI is green


lhotari (approved):
LGTM

PR #25641, opened 2026-05-01
[fix][test] Make NamespacesTest.cleanupAfterMethod tolerant of transient infra failures

merlimat:
## Summary

The `@AfterMethod` cleanup calls `admin.namespaces().getNamespaces(tenant)`. Under load (long-running test method, GC pause) the broker's ZK session can expire and that call returns a 500. The actual test method had already passed at that point, so masking the result with a cleanup exception isn't useful.

Wrap the cleanup in try/catch + warn-log. Subsequent test methods either recover (ZK reconnects) or fail-fast with a clearer error than a session-expired stack trace from a teardown hook.

## Test plan

- [x] `testCreateNamespaces` passes locally with the change applied
- [ ] CI is green


lhotari (approved):
LGTM

PR #25611, opened 2026-04-28
[test][client-v5] V5 layout-dynamics coverage + consumer fixes (DAG replay, late-ack)

merlimat:
## Summary

End-to-end coverage for the scalable-topic-specific behaviors of the V5 client (split / merge / key routing / cumulative ack / DAG following), plus two consumer-side fixes the tests surfaced.

### Tests

- **V5KeyRoutingTest** — same key always lands on the same segment (per-key order preservation), and a varied key set on a 4-segment topic flows end-to-end.
- **V5SegmentSplitTest** — admin-triggered split with producer + consumer running through it; messages produced before and after are delivered exactly once.
- **V5SegmentMergeTest** — same shape, with a merge.
- **V5DAGFollowingTest** —
  - Existing subscription drains across a chain of splits.
  - Brand-new EARLIEST subscription replays the sealed parent (newly possible with the QueueConsumer fix below).
- **V5CumulativeAckTest** — three StreamConsumer cumulative-ack scenarios:
  - Single ack on the last received id → empty backlog on re-attach.
  - Mid-stream ack → re-attach replays only the unacked tail.
  - Cross-split ack → ack parent fully, split, ack children, re-attach sees no backlog.
- **V5QueueConsumerMultiSegmentAckTest** — QueueConsumer counterparts:
  - Individual acks on every message of a multi-segment topic → re-attach sees no backlog.
  - Partial acks → re-attach replays exactly the unacked tail.

### V5 client fixes

- **`ClientSegmentLayout`** — also expose `sealedSegments()` from the DAG.
- **`ScalableQueueConsumer`** — subscribe to *all* segments (active + sealed) so a fresh subscription replays sealed-parent backlog before transitioning onto the children. Defer the per-segment v4 consumer close from layout-change time to `TopicTerminated` time so acks for messages pulled before the seal still flow.
- **`ScalableStreamConsumer`** — keep controller-driven (active-only) subscription, but apply the same close-on-`TopicTerminated` deferral.

Layout-dynamics waits use `Awaitility.await()` for polling, with a comment explaining why the wait is needed (V5 client's DAG watch is async even though the admin call is synchronous server-side).

## Test plan
- [x] `./gradlew :pulsar-broker:test --tests \"org.apache.pulsar.client.api.v5.V5KeyRoutingTest\"`
- [x] `./gradlew :pulsar-broker:test --tests \"org.apache.pulsar.client.api.v5.V5SegmentSplitTest\"`
[...]

PR #25642, opened 2026-05-01
[fix][broker] PIP-468: StreamConsumer / CheckpointConsumer DAG-replay with parent-drain ordering

merlimat:
## Summary

The Stream and Checkpoint consumer subscriptions are controller-driven: their
segment assignment comes from `SubscriptionCoordinator`. Previously
`computeAssignment` handed out only active segments, so a fresh EARLIEST consumer
joining after a split couldn't read messages produced before the split — those
live on the now-sealed parent and were silently orphaned. The QueueConsumer fix
in #25611 already extended the queue flow to active+sealed; this PR brings
StreamConsumer / CheckpointConsumer in line and adds a per-subscription parent-drain
ordering guarantee on top.

## Behavior

- Sealed segments are always included in the assignment. The client drains them
  via the existing per-segment v4 receive loop / Reader; an already-drained
  segment yields `TopicTerminated` immediately and the v4 consumer/Reader closes.

- An active child of a split / merge is held back until *every* parent in the
  DAG has been drained for this subscription. Without this we'd hand a consumer
  the child of a just-split segment immediately, which breaks per-key ordering
  against unread messages still sitting in the parent. Initial active segments
  (no parents) are unaffected.

- Parent-drain ordering only applies to **STREAM** consumers — they're the ones
  that need per-key ordering across a split. **CHECKPOINT** consumers track
  position client-side via `Checkpoint` and never create per-segment subscription
  cursors (they read via Readers), so the ordering machinery would block their
  children indefinitely. **QUEUE** consumers are shared and accept out-of-order
  delivery by design. The coordinator skips the ordering filter for both.

## Implementation

- New `SegmentDrainChecker` interface — async predicate for "is segment X drained
  for subscription S".

- `SubscriptionCoordinator` gets the checker + an exponential `Backoff` (2s
  initial, 15min max). Tracks `drainedSegmentIds` in-memory, runs a self-rescheduling
  poll that queries every undrained sealed segment and rebalances when state
  changes. The backoff resets on every progress event (drain detected, layout
  change, fresh consumer registration); the poller stops entirely once all
  sealed segments are drained.

- `ScalableTopicController.isSegmentDrained` calls the new admin endpoint
  `GET /segments/{tenant}/{ns}/{topic}/{descriptor}/subscription/{sub}/backlog`,
  which the v4 admin routes to the segment topic's owner — works whether the
  controller and the segment colocate or not. 404 (topic / subscription not yet
  loaded) is treated as "not drained, retry on the next poll".

- `ScalableConsumerType` plumbed from `ServerCnx` through `ScalableTopicService`
  → `ScalableTopicController` → `SubscriptionCoordinator`. The controller
  passes a real drain checker only for STREAM; for CHECKPOINT / QUEUE the
  coordinator gets a `null` checker and skips parent-drain entirely.

- `ScalableTopicController.close` stops every coordinator's drain poller.

### `V5StreamConsumerDagReplayTest` (new)

- `testEarliestSubscribePostSplitReadsSealedBacklog` — fresh EARLIEST consumer
  reads pre-split sealed backlog AND post-split active.
- `testEarliestSubscribeAfterTwoSplitsReadsAllBacklog` — two-deep DAG; fresh
  consumer reads every message across both sealed generations.
- `testSealedBacklogNotRedeliveredAfterFirstConsumerDrains` — second consumer
  joining after the first drained sees nothing (cursor at end on the sealed
  segment).

### `V5CheckpointConsumerDagReplayTest` (new)

Mirror of the Stream test for both unmanaged and managed (consumerGroup) paths,
single-split and two-splits scenarios. Four tests total.

### `SubscriptionCoordinatorTest.testActiveChildrenBlockedUntilParentDrained` (new)

Controllable mock checker; asserts children of segment 0 are excluded until
segment 0 is marked drained, then included after the next poll.

- area/broker
- area/test

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/Segments.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/Segments.java
@@ -28,6 +28,7 @@ import java.util.concurrent.CompletableFuture;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.Encoded;
+import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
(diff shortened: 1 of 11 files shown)


lhotari:
**Code review (Claude Code)**

Posting three correctness concerns surfaced by a local review with Claude Code.

### 1. [BUG] Backoff reset is ineffective if a poll is already scheduled

In `SubscriptionCoordinator` (`registerConsumer`, `onLayoutChange`, `markSegmentsDrained`):

```java
drainBackoff.reset();
ensureDrainPollerRunning();
```

`ensureDrainPollerRunning` is a no-op when `drainPollTask` is still scheduled and not yet cancelled/done. If the poller had backed off to the 15-minute cap during a quiet period, a fresh EARLIEST consumer joining (or a layout change, or an observed drain) will still wait up to 15 minutes for the next drain check — directly contradicting the comment (*"kick off drain checks at the shortest delay"*) and the docstring on `DEFAULT_DRAIN_INITIAL_DELAY` (*"a freshly subscribed EARLIEST consumer doesn't wait"*). Fix: cancel the existing scheduled task before re-arming, or have these progress events explicitly reschedule with the reset delay.

### 2. [BUG] `close()` can race with an in-flight poll and leak a scheduled task

Also in `SubscriptionCoordinator`: `pollDrainStatus` nullifies `drainPollTask` at the top of its `synchronized` block, then the async `whenComplete` callback calls `ensureDrainPollerRunning()` which schedules a new task. If `close()` runs in that window, it sees `drainPollTask == null` and does nothing; the subsequent re-arm then schedules a task that fires after the controller has gone away. There is no `closed` flag checked in `ensureDrainPollerRunning` or in the `whenComplete` rearm path. Fix: add a `closed` boolean (set in `close()`) and short-circuit `ensureDrainPollerRunning` / the `whenComplete` rearm on it.

### 3. [BUG] Restore path defaults every recovered subscription to STREAM, deadlocking CHECKPOINT/QUEUE on leader handoff

In `ScalableTopicController.createCoordinator(String)`:

```java
private SubscriptionCoordinator createCoordinator(String subscription) {
    return createCoordinator(subscription, ScalableConsumerType.STREAM);
}
```

`restoreSubscription` calls this — meaning after any controller restart / leader change, every restored subscription gets a coordinator with parent-drain enforcement, regardless of the consumer type that originally created it. The Javadoc on `registerConsumer` confirms the policy is then locked: *"The coordinator's setting is fixed at first registration; subsequent registers with a different type still work but won't change the ordering policy."*

For a CHECKPOINT subscription on a topic with sealed parents, the parent will never report drained (CHECKPOINT doesn't create per-segment cursors — explicitly noted in the PR body), so children remain blocked indefinitely after a restart. The *"conservative default"* comment is inverted — STREAM is the *blocking* mode, so it's the least-conservative default in this context. Fix: persist the subscription type in metadata alongside the subscription so it can be restored faithfully, or use a `null` checker (no enforcement) on the restore path until a real consumer re-registers and reveals the type.

lhotari (approved):
LGTM, however Claude Code highlighted some correctness issues (please check the comment to see if these are valid).

merlimat:
@lhotari thanks for the review — addressed all three in 300b3dc3377:

1. **Backoff reset on a pending task** — added `resetAndRearmDrainPoll()` that cancels the in-flight `drainPollTask` before resetting and re-arming. Used everywhere a progress event happens (register, layout change, observed drain).

2. **`close()` race with in-flight poll** — added a `closed` flag set by `close()`, checked in both `ensureDrainPollerRunning` and `pollDrainStatus` so the `whenComplete` rearm path becomes a no-op once the coordinator has been closed.

3. **Restore-path STREAM default** — restore now creates the coordinator with `null` drain checker (no enforcement). On first register-after-restore, `ScalableTopicController` installs a real checker via the new `SubscriptionCoordinator.installDrainChecker(...)` if and only if the consumer type is STREAM. CHECKPOINT / QUEUE subscriptions stay un-blocked. Persisting the type in metadata is the proper long-term fix and noted in the code comment as a follow-up.

Four new unit tests in `SubscriptionCoordinatorTest` covering each fix:
- `testFreshRegisterCancelsBackedOffPollAndRearmsImmediately`
- `testCloseRaceWithInFlightPollDoesNotLeakRearm`
- `testRestoredCoordinatorStartsWithoutParentDrainOrdering`
- `testInstallDrainCheckerAfterRestoreEnablesOrdering`

All tests pass; checkstyle clean.

PR #25644, opened 2026-05-01
[fix][broker] ConcurrentLongHashMap throw ArrayIndexOutOfBoundsException

### LINKED ISSUE #25643: [Bug] Master Code ConcurrentLongHashMap throw ArrayIndexOutOfBoundsException
### Search before reporting

### Read release policy

### User environment

master code 

### Issue Description

concurrent jmh test access map concurrently the exception throws 

### Error messages

```text

```

### Reproducing the issue

see jmh test 

### Additional information

_No response_

void-ptr974:
Fixes #25643

## Bug

  `ConcurrentLongHashMap.Section` publishes `keys`, `values`, `capacity` as three separate `volatile` fields. During `rehash` a reader can observe a torn snapshot — e.g. new `keys` (length 8)
   with old `values` (length 4) — then compute `bucket = signSafeMod(hash, values.length)` ∈ [0,8) and index `keys[bucket]` against length 4, throwing `ArrayIndexOutOfBoundsException` at
  `Section.get`. The exception fires *before* the optimistic-read `validate(stamp)` check, so the existing fallback can't recover.

  ## Reproduce in ≈ 2 minutes

  ```sh
  # Revert just the production file to upstream master, keeping this PR's tests + benchmark:
  git checkout origin/master -- pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java

  ./gradlew :microbench:shadowJar

  java -jar microbench/build/libs/microbench-*-benchmarks.jar \
      "ConcurrentLongHashMapBenchmark.concurrentExpandShrink" \
      -wi 1 -i 1 -w 3s -r 3s -f 1 -p implementation=clhm
  ```

  `ArrayIndexOutOfBoundsException` fires within the first warmup iteration. 10/10 reproductions across 10 fresh JVMs on the unpatched code, 0/10 after this PR.

  ## Fix

  Bundle `keys`, `values`, `capacity` into one immutable inner class published via a single volatile reference:

  ```java
  private static final class Table<V> {
      final long[] keys;
      final V[] values;
      final int capacity;
  }
  private volatile Table<V> table;
  ```

  Final-field semantics plus one volatile-acquire-release pair give the reader a consistent triple from one read. The partial-publish window is impossible by construction.

  ## Tests added

  `pulsar-common/src/test/.../ConcurrentLongHashMapTest`:

  - `testNoLostGetAfterPublish` — writer puts `k` then publishes `k` via a volatile counter; every reader observing the counter must see a non-null value for `k`.
  - `testCorrectnessAgainstConcurrentHashMap` — 8 threads × 50K random ops mirrored to `ConcurrentHashMap`; per-call return values and final state must match.
  - `testConcurrentMultiWriterExpandShrink` — 8 writers × 8 readers on a single section with `autoShrink`; no torn `(k,v)` pair is allowed to surface.
  - `testForEachDuringWrites` — `forEach` during concurrent puts/removes must not throw or expose internal sentinels.
  - `testConcurrentExpandAndShrinkAndGet` — strengthened existing test.

  ```sh
  ./gradlew :pulsar-common:test --tests ConcurrentLongHashMapTest
  ```

  20/20 pass in 2.3s.

  A JMH benchmark (`ConcurrentLongHashMapBenchmark` in the `microbench` module) is included so this race can be regression-tested on every JDK / hardware change.

  - [ ] The public API
  - [ ] The schema
  - [ ] The threading model
  - [ ] The binary protocol
  - [ ] The REST endpoints
  - [ ] The admin CLI options
  - [ ] The metrics

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java
@@ -288,11 +288,25 @@ public class ConcurrentLongHashMap<V> {
         void accept(long key, V value);
     }
 
-    // A section is a portion of the hash map that is covered by a single
+    // A section is a portion of the hash map that is covered by a single lock. The keys, values
+    // and capacity arrays are bundled into an immutable Table snapshot so that readers always see
+    // a consistent (key, value, length) triple, eliminating the partial-publish race that the
+    // previous design had to paper over with Math.min(keys.length, values.length).
     @SuppressWarnings("serial")
     private static final class Section<V> extends StampedLock {
-        private volatile long[] keys;
-        private volatile V[] values;
+        private static final class Table<V> {
+            final long[] keys;
+            final V[] values;
+            final int capacity;
+
+            Table(long[] keys, V[] values, int capacity) {
+                this.keys = keys;
+                this.values = values;
+                this.capacity = capacity;
+            }
+        }
+
+        private volatile Table<V> table;
 
         private volatile int capacity;
         private final int initCapacity;
@@ -314,8 +328,7 @@ public class ConcurrentLongHashMap<V> {
                 float expandFactor, float shrinkFactor) {
             this.capacity = alignToPowerOfTwo(capacity);
             this.initCapacity = this.capacity;
-            this.keys = new long[this.capacity];
-            this.values = (V[]) new Object[this.capacity];
+            this.table = new Table<>(new long[this.capacity], (V[]) new Object[this.capacity], this.capacity);
             this.size = 0;
             this.usedBuckets = 0;
             this.autoShrink = autoShrink;
@@ -331,11 +344,10 @@ public class ConcurrentLongHashMap<V> {
             long stamp = tryOptimisticRead();
             boolean acquiredLock = false;
 
-            // add local variable here, so OutOfBound won't happen
-            long[] keys = this.keys;
-            V[] values = this.values;
-            // calculate table.length as capacity to avoid rehash changing capacity
-            int bucket = signSafeMod(keyHash, values.length);
+            Table<V> table = this.table;
+            long[] keys = table.keys;
+            V[] values = table.values;
+            int bucket = signSafeMod(keyHash, table.capacity);
 
             try {
                 while (true) {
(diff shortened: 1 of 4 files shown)


pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java
```diff
@@ -288,11 +289,27 @@ public interface EntryProcessor<V> {
...
-    // A section is a portion of the hash map that is covered by a single
+    // A section is a portion of the hash map that is covered by a single lock. The keys, values
+    // and capacity arrays are bundled into an immutable Table snapshot so that readers always see
+    // a consistent (key, value, length) triple, eliminating the partial-publish race that the
+    // previous design had to paper over with Math.min(keys.length, values.length).
     @SuppressWarnings("serial")
     private static final class Section<V> extends StampedLock {
-        private volatile long[] keys;
-        private volatile V[] values;
+        private static final class Table<V> {
```
merlimat: This could be converted to a Java record

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java
```diff
@@ -288,11 +289,27 @@ public interface EntryProcessor<V> {
...
+                this.values = values;
+                this.capacity = capacity;
+            }
+        }
+
+        // Section is Serializable only by inheritance from StampedLock; never actually serialized.
+        @SuppressFBWarnings("SE_BAD_FIELD")
+        private volatile Table<V> table;
 
         private volatile int capacity;
```
merlimat: shouldn't this get removed now?

merlimat (commented):
Great find! 

I just left a couple of comments

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java
void-ptr974: Done — converted to a record. All field accesses now use the generated accessors.

pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/ConcurrentLongHashMap.java
void-ptr974: Done — removed. Outer `capacity()` now reads `s.table.capacity()`.

void-ptr974:
Thanks @merlimat — both addressed in the latest commit. PTAL.

merlimat (approved):

PR #25650, opened 2026-05-01
[fix][broker] PIP-468: Fix listener leak in DagWatchSession

merlimat:
## Summary

Every `DagWatchSession.start()` called
`resources.getStore().registerListener(this::onNotification)` directly. `close()`
only flipped a closed flag because `MetadataStore` exposes no
`unregisterListener` — so each closed session left a stale listener registered
for the broker's lifetime. Both a memory leak and a per-notification dispatch
tax that grew linearly with the total number of sessions ever opened.

The fix mirrors the `TopicResources` / `TopicListener` pattern already used
elsewhere in the codebase: register a single shared store-level listener at
construction and dispatch through a per-listener registry.

## Changes

- `ScalableTopicResources` gains a single metadata-store listener at
  construction time and a `Map<MetadataPathListener, String>` registry keyed
  by an exact metadata path. `registerPathListener` / `deregisterPathListener`
  manage entries; the constructor fan-out dispatches each notification to
  listeners whose registered path matches. Listener exceptions are caught and
  logged so one broken subscriber can't poison the dispatch loop.
- `DagWatchSession` implements `ScalableTopicResources.MetadataPathListener`.
  `start()` registers via the registry, `close()` deregisters cleanly. The
  per-event path-equality check stays in `onNotification` as a defensive
  no-op against any future registry-level bug.

## Test plan

- New `ScalableTopicListenerRegistryTest` covers register / deregister
  lifecycle, exact-path dispatch, multiple listeners on the same path, and
  isolation across listeners (one throwing must not interrupt others).
- `DagWatchSessionTest`'s old `testStartRegistersMetadataStoreListener`
  rewritten as `testStartRegistersWithResources` (verifies the registry call)
  and `testCloseDeregistersPathListener` (asserts the leak is gone).
- Full `org.apache.pulsar.broker.service.scalable.*` and
  `org.apache.pulsar.broker.resources.*` test packages pass.

- area/broker

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/ScalableTopicResources.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/resources/ScalableTopicResources.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import lombok.CustomLog;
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM

PR #25657, opened 2026-05-03
PIP-468: V5 policy options use builder pattern

merlimat:
## Summary

Convert the V5 client policy options from records (with public canonical constructors) into final classes whose only construction paths are an explicit `Builder` and a small set of named factory methods. This keeps the public surface to deliberate, documented combinations and lets the classes evolve without breaking source compatibility every time a new field is added.

Refactored: `BackoffPolicy`, `BatchingPolicy`, `ChunkingPolicy`, `CompressionPolicy`, `ConnectionPolicy`, `DeadLetterPolicy`, `EncryptionPolicy`, `ProcessingTimeoutPolicy`, `TlsPolicy`, `TransactionPolicy`.

Each now has:
- Private constructor.
- private final fields with record-style accessors (`foo()` — backward compatible with existing callers).
- A static `builder()` plus a fluent inner `Builder` with javadoc on each knob.
- Small, opinionated factories where there's an obvious dominant case (e.g. `CompressionPolicy.disabled()`, `TlsPolicy.ofInsecure()`, `BatchingPolicy.ofDisabled()`, `BackoffPolicy.fixed`/`exponential(initial, max)`, `ProcessingTimeoutPolicy.of(timeout)`, `EncryptionPolicy.forProducer`/`forConsumer`).
- Lombok `@EqualsAndHashCode` + `@ToString` to preserve value-type semantics.

Updated callers — `Examples.java`, `V5DeadLetterPolicyTest`, `V5TransactionTest`, `V5ProducerBatchingTest` — to use the builder/factory entry points.

## Test plan

- [x] `pulsar-client-api-v5` and `pulsar-broker` checkstyle clean.
- [x] V5 affected tests pass: `V5DeadLetterPolicyTest`, `V5TransactionTest`, `V5ProducerBatchingTest`, `V5ProcessingTimeoutTest`.

### DIFF
--- a/pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/config/BackoffPolicy.java
+++ b/pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/config/BackoffPolicy.java
@@ -20,27 +20,55 @@ package org.apache.pulsar.client.api.v5.config;
 
 import java.time.Duration;
 import java.util.Objects;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
 
 /**
  * Backoff configuration for broker reconnection attempts.
  *
  * <p>The delay for attempt {@code n} is {@code min(initialInterval * multiplier^(n-1), maxInterval)}.
  *
- * @param initialInterval the delay before the first reconnection attempt
- * @param maxInterval     the maximum delay between reconnection attempts
- * @param multiplier      the multiplier applied after each attempt
+ * <p>Use {@link #fixed(Duration, Duration)} or {@link #exponential(Duration, Duration)} for
+ * the common cases, or {@link #builder()} to configure all knobs explicitly.
  */
-public record BackoffPolicy(
-        Duration initialInterval,
-        Duration maxInterval,
-        double multiplier
-) {
-    public BackoffPolicy {
+@EqualsAndHashCode
+@ToString
+public final class BackoffPolicy {
+
+    private final Duration initialInterval;
+    private final Duration maxInterval;
+    private final double multiplier;
+
+    private BackoffPolicy(Duration initialInterval, Duration maxInterval, double multiplier) {
         Objects.requireNonNull(initialInterval, "initialInterval must not be null");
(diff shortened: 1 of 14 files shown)


pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/config/BackoffPolicy.java
```diff
@@ -20,27 +20,55 @@
...
  */
-public record BackoffPolicy(
-        Duration initialInterval,
-        Duration maxInterval,
-        double multiplier
-) {
-    public BackoffPolicy {
+@EqualsAndHashCode
+@ToString
+public final class BackoffPolicy {
```
lhotari, line 36: Is there a way to configure random jitter for the `BackoffPolicy`?

lhotari (approved):
LGTM, one unrelated question about `BackoffPolicy`.

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/config/BackoffPolicy.java
merlimat: Not really: 

https://github.com/apache/pulsar/blob/77b05364f70811dc0dba13281fba68ac5a619478/pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java#L133-L138

It's hardcoded to 10% and, worse, it's not applied the first time, so all first attempts are coming together

merlimat: eg: `        return Duration.ofMillis(Math.max(initialMillis, currentMillis));`

but `initialMillis` is already the minimum and the first time `currentMillis` is like `initialMillis - ~10%`

lhotari: That's fine for now, perhaps we take a look later if there's a need to have jitter on all attempts. It could be useful.

PR #25658, opened 2026-05-03
[feat] PIP-468: record wall-clock created/sealed timestamps on segments

merlimat:
## Summary

Add `createdAtMs` / `sealedAtMs` to `SegmentInfo` (and the `SegmentInfoProto` wire format) alongside the existing `createdAtEpoch` / `sealedAtEpoch` — those are DAG generation numbers, not clocks, and can't drive retention or timestamp-based seek.

The new fields are populated at the canonical lifecycle points:

- `ScalableTopicController#createInitialMetadata` records a single `nowMs` shared by all initial segments.
- `SegmentLayout#splitSegment` and `mergeSegments` take an explicit `nowMs` and thread it into the parents' seal time and the children's create time. `ScalableTopicController` captures `final long nowMs` once and passes it to both the local preview and the CAS-retried metadata transformer, so retries don't drift the timestamp between writers.
- `DagWatchSession` writes both fields into the wire format.

This is the prerequisite for upcoming work on sealed-segment retention GC and V5 seek-by-timestamp; **no behavior change in this PR**.

## Test plan

- [x] New unit coverage in `SegmentLayoutTest` (`testSplitRecordsWallClockTimestamps`, `testMergeRecordsWallClockTimestamps`) and proto-roundtrip assertions in `DagWatchSessionTest`.
- [x] All existing scalable tests pass: `SegmentLayoutTest` (17), `SubscriptionCoordinatorTest` (16), `DagWatchSessionTest` (14), `ScalableTopicControllerTest` (24), `V5SmokeTest`, `V5SegmentSplitTest`.
- [x] Checkstyle clean (`pulsar-common`, `pulsar-broker`, `pulsar-client-v5`).

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
@@ -285,8 +285,13 @@ public class ScalableTopicController {
     public CompletableFuture<SegmentLayout> splitSegment(long segmentId) {
         checkLeader();
 
+        // Single timestamp shared by the local preview and the CAS-retried metadata update,
+        // so the children's createdAtMs and the parent's sealedAtMs always agree even if the
+        // CAS retries due to concurrent writers.
+        final long nowMs = System.currentTimeMillis();
+
         // Compute the new layout locally to derive child segment info
-        SegmentLayout newLayout = currentLayout.splitSegment(segmentId);
+        SegmentLayout newLayout = currentLayout.splitSegment(segmentId, nowMs);
         SegmentInfo child1 = newLayout.getAllSegments().get(newLayout.getNextSegmentId() - 2);
         SegmentInfo child2 = newLayout.getAllSegments().get(newLayout.getNextSegmentId() - 1);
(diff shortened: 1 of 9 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -285,8 +285,13 @@ public CompletableFuture<SegmentLayout> getLayout() {
     public CompletableFuture<SegmentLayout> splitSegment(long segmentId) {
         checkLeader();
 
+        // Single timestamp shared by the local preview and the CAS-retried metadata update,
+        // so the children's createdAtMs and the parent's sealedAtMs always agree even if the
+        // CAS retries due to concurrent writers.
+        final long nowMs = System.currentTimeMillis();
```
lhotari, line 291: In some of the existing code locations, PulsarService's `java.time.Clock clock` field is used for timestamps, however it's not consistently applied in the code base. Would it be useful to use `java.time.Clock` for timestamps?

lhotari (approved):
LGTM, just one question

PR #25668, opened 2026-05-05
[feat] PIP-468: sealed-segment retention GC for scalable topics

merlimat:
## Summary

After a split / merge the parent segment is sealed and accepts no further writes; eventually its data ages out. There was no mechanism to actually delete it. This PR adds:

### 1. Single owner of segment-topic deletion

The v4 inactive-topic GC would otherwise race the controller for whole-topic deletion of segment-backing topics. `PersistentTopic#checkGC` now early-returns when the topic is in the `segment://` domain — the controller is the sole lifecycle owner.

### 2. Periodic GC tick on the `ScalableTopicController` leader

Each tick:

- Resolves the effective retention from **topic-policies → namespace policy → broker default**. Negative ⇒ keep forever, tick is a no-op.
- Picks sealed segments where `(now - sealedAtMs) >= retentionMs`.
- For each candidate, polls every existing subscription's backlog on that segment via the existing `/segments/.../backlog` admin endpoint. All-zero ⇒ prunable.
- CAS-prunes the layout (re-validating against the latest layout to handle a concurrent prune by a former leader gracefully), reloads, notifies subscriptions, then deletes the backing topic via `admin.topics().deleteAsync(force=true)`.
- Layout-prune is the point of no return; backing-topic delete is best-effort and retried on subsequent ticks.

The clock is injectable (`java.time.Clock`) so tests can fast-forward past retention deterministically. `splitSegment` / `mergeSegments` now read the wall-clock through the same `Clock` so that test ticks compute consistent ages.

## Test plan

- [x] `testGcTickPrunesDrainedSealedSegmentPastRetention` — split, tick within retention (no prune), advance past retention, tick again, assert pruned + delete called.
- [x] `testGcTickRespectsKeepForeverRetention` — negative retention leaves the segment in place even after a year of clock advance.
- [x] Existing scalable + V5 suites green: `org.apache.pulsar.broker.service.scalable.*` (76 tests), `V5SmokeTest`, `V5SegmentSplitTest`, `V5DeadLetterPolicyTest`.
- [x] Checkstyle clean (`pulsar-broker` main + test).

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
@@ -748,29 +748,103 @@ public class ScalableTopicController {
 
     /**
      * For each candidate sealed segment, check that every existing subscription has
-     * drained it (backlog == 0); prune the ones that pass. Failures on individual
-     * segments are logged and skipped — the next tick retries.
+     * drained it (backlog == 0); prune the ones that pass. The drain checks fan out
+     * concurrently, but the resulting layout mutation is coalesced into a <em>single</em>
+     * CAS write so multiple eligible segments don't compete on the same metadata znode.
+     *
+     * <p><b>Subscription-type behaviour.</b> The drain check is the per-segment backlog
+     * admin endpoint — Pulsar's standard cursor-position view, which works the same way
+     * for STREAM (Exclusive) and QUEUE (Shared) subscriptions: a sealed segment with
+     * cursor at the end reports backlog 0. For CHECKPOINT subscriptions there is no
+     * broker-side cursor, the endpoint returns {@code NotFoundException}, and
+     * {@code isSegmentDrained} reports {@code false} — the segment is treated as
+     * "still in use" and never pruned while a CHECKPOINT subscription is registered.
+     *
+     * <p><b>Parent-vs-child ordering.</b> Sealed segments form a DAG; pruning is allowed
+     * in any order because the active leaves always cover the full hash range, and the
+     * managed-ledger storage of each segment is independent. {@link SegmentLayout#pruneSegment}
+     * rewrites the parent/child edges, so consumers using the post-prune layout see the
+     * pruned segment as "no longer present" — equivalent to "drained" for parent-drain
+     * ordering.
      */
     private CompletableFuture<Void> pruneEligibleAsync(List<SegmentInfo> candidates) {
         return resources.listSubscriptionsAsync(topicName)
                 .thenCompose(subs -> {
-                    List<CompletableFuture<Void>> perSegment = new ArrayList<>();
+                    // Fan out drain checks; collect the survivors.
+                    List<CompletableFuture<SegmentInfo>> filtered = new ArrayList<>();
                     for (SegmentInfo seg : candidates) {
-                        perSegment.add(prunable(seg, subs)
-                                .thenCompose(prunable -> prunable
-                                        ? pruneSegmentAsync(seg)
-                                        : CompletableFuture.completedFuture(null))
+                        filtered.add(prunable(seg, subs)
+                                .thenApply(ok -> ok ? seg : null)
                                 .exceptionally(ex -> {
                                     log.warn().attr("segmentId", seg.segmentId())
                                             .exceptionMessage(ex)
-                                            .log("GC: failed to evaluate / prune segment;"
+                                            .log("GC: failed to evaluate prunability;"
                                                     + " will retry on next tick");
                                     return null;
                                 }));
                     }
-                    return CompletableFuture.allOf(
-                            perSegment.toArray(CompletableFuture[]::new));
-                });
+                    return CompletableFuture.allOf(filtered.toArray(CompletableFuture[]::new))
+                            .thenApply(__ -> {
+                                List<SegmentInfo> drained = new ArrayList<>();
+                                for (var f : filtered) {
+                                    SegmentInfo s = f.join();
+                                    if (s != null) {
+                                        drained.add(s);
+                                    }
+                                }
+                                return drained;
+                            });
+                })
+                .thenCompose(this::pruneAllAsync);
+    }
+
+    /**
+     * Coalesce all drained-and-eligible segments into a single layout-mutation CAS,
+     * then fan out the per-segment backing-topic deletes. This is the path that
+     * actually mutates state. Re-validates leadership before the CAS — drain checks
+     * can take seconds, leadership may have flipped in the meantime, and we don't
+     * want a deposed leader writing layout updates.
+     */
+    private CompletableFuture<Void> pruneAllAsync(List<SegmentInfo> drained) {
+        if (drained.isEmpty()) {
+            return CompletableFuture.completedFuture(null);
+        }
+        if (!isLeader() || closed) {
+            return CompletableFuture.completedFuture(null);
+        }
+        for (SegmentInfo s : drained) {
+            log.info().attr("segmentId", s.segmentId())
+                    .attr("sealedAtMs", s.sealedAtMs())
+                    .log("GC: pruning sealed segment past retention");
+        }
+        return resources.updateScalableTopicAsync(topicName, md -> {
+            SegmentLayout latest = SegmentLayout.fromMetadata(md);
+            SegmentLayout updated = latest;
+            for (SegmentInfo s : drained) {
+                // Re-validate per segment: another writer (or a previous failed
+                // tick of this same loop) may have already pruned it.
+                if (updated.getAllSegments().containsKey(s.segmentId())) {
+                    updated = updated.pruneSegment(s.segmentId());
+                }
+            }
+            return updated == latest ? md : updated.toMetadata(md.getProperties());
+        }).thenCompose(__ -> resources.getScalableTopicMetadataAsync(topicName, true))
+          .thenCompose(optMd -> {
+              currentLayout = SegmentLayout.fromMetadata(optMd.orElseThrow());
+              return notifySubscriptions(currentLayout);
+          })
+          .thenCompose(__ -> {
+              CompletableFuture<?>[] deletes = drained.stream()
+                      .map(this::deleteSegmentBackingTopic)
+                      .toArray(CompletableFuture[]::new);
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -619,10 +700,225 @@ public CompletableFuture<org.apache.pulsar.common.policies.data.ScalableTopicSta
...
+                        }
+                        log.warn().attr("segment", name).exceptionMessage(cause)
+                                .log("GC: failed to delete backing segment topic;"
+                                        + " will retry on next tick");
+                        return null;
+                    });
+        } catch (PulsarServerException e) {
+            return CompletableFuture.failedFuture(e);
+        }
+    }
```
lhotari, line 897: Shouldn't the existing `ScalableTopicService.deleteUnderlyingSegmentTopic` method be used instead?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -619,10 +700,225 @@ public CompletableFuture<org.apache.pulsar.common.policies.data.ScalableTopicSta
...
+        return resources.updateScalableTopicAsync(topicName, md -> {
+            SegmentLayout latest = SegmentLayout.fromMetadata(md);
+            // Re-validate against the latest layout: another writer may have already
+            // pruned this segment, or replaced it via a follow-up split / merge. If
+            // it's gone, leave the metadata untouched (no-op CAS).
+            if (!latest.getAllSegments().containsKey(segmentId)) {
+                return md;
+            }
+            SegmentLayout updated = latest.pruneSegment(segmentId);
+            return updated.toMetadata(md.getProperties());
```
lhotari: `pruneEligibleAsync` issues `pruneSegmentAsync` for every drained candidate concurrently, and each one calls `resources.updateScalableTopicAsync(topicName, …)` — i.e., a CAS on the single topic-metadata node. With more than one or two eligible segments per tick, each CAS conflicts with its siblings, forcing optimistic-retry storms. Coalescing all eligible segments into a single layout-mutation lambda would be both faster and simpler. The post-CAS `getScalableTopicMetadataAsync(true)` and `notifySubscriptions(currentLayout)` are also issued N times for what is effectively one layout transition.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -619,10 +700,225 @@ public CompletableFuture<org.apache.pulsar.common.policies.data.ScalableTopicSta
...
+     * segments are logged and skipped — the next tick retries.
+     */
+    private CompletableFuture<Void> pruneEligibleAsync(List<SegmentInfo> candidates) {
+        return resources.listSubscriptionsAsync(topicName)
+                .thenCompose(subs -> {
+                    List<CompletableFuture<Void>> perSegment = new ArrayList<>();
+                    for (SegmentInfo seg : candidates) {
+                        perSegment.add(prunable(seg, subs)
+                                .thenCompose(prunable -> prunable
+                                        ? pruneSegmentAsync(seg)
```
lhotari: `prunable` is both a method name and a lambda parameter name. it would be better to rename either one for clarity

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -619,10 +700,225 @@ public CompletableFuture<org.apache.pulsar.common.policies.data.ScalableTopicSta
...
+
+        return resources.updateScalableTopicAsync(topicName, md -> {
+            SegmentLayout latest = SegmentLayout.fromMetadata(md);
+            // Re-validate against the latest layout: another writer may have already
+            // pruned this segment, or replaced it via a follow-up split / merge. If
+            // it's gone, leave the metadata untouched (no-op CAS).
+            if (!latest.getAllSegments().containsKey(segmentId)) {
+                return md;
+            }
+            SegmentLayout updated = latest.pruneSegment(segmentId);
```
lhotari: would it be useful to check that the current broker is still the leader before this is performed?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -619,10 +700,225 @@ public CompletableFuture<org.apache.pulsar.common.policies.data.ScalableTopicSta
...
+                                                    + " will retry on next tick");
+                                    return null;
+                                }));
+                    }
+                    return CompletableFuture.allOf(
+                            perSegment.toArray(CompletableFuture[]::new));
+                });
+    }
+
+    private CompletableFuture<Boolean> prunable(SegmentInfo seg, List<String> subs) {
```
lhotari, line 850: Does this currently handle queue consumers too? Is it possible that a child segment gets pruned before one of it's parent segments?  
Does checkpoint consumer bring up special cases? Is it possible that a child segment gets pruned while a checkpoint consumer is actively consuming one of the parent segments?

lhotari (commented):
I added some review comments.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: Done — using `scalableTopics().deleteSegmentAsync(toSegmentPersistentName(seg), true)`, the same primitive `ScalableTopicService.deleteUnderlyingSegmentTopic` uses. As a side-benefit, this drops the hand-built `persistent://` URL (which was missing the segment descriptor and would have been buggy at delete time).

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: Good catch — fixed in 3b1dc1a. Drain checks still fan out concurrently, but the layout mutation is now a single CAS that prunes every eligible segment in one shot, followed by a single metadata reload + `notifySubscriptions(...)` and parallel backing-topic deletes. No more siblings competing on the same znode.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: Renamed: the inner lambda variable is now `ok` (and the surrounding code was rewritten as part of the coalescing fix). The method name `prunable` stays.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: Added: re-check `isLeader()` (and `closed`) at the top of `pruneAllAsync` just before the CAS, since drain checks can take seconds and leadership may flip meanwhile. The CAS itself would still protect the metadata, but bailing early avoids a deposed leader writing layout updates and racing the new one on backing-topic deletes.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: Added a docstring on `pruneEligibleAsync` covering all three cases:

- **STREAM / QUEUE**: the drain check is the per-segment backlog admin endpoint, which works the same way for Exclusive (STREAM) and Shared (QUEUE) cursors — both report 0 backlog when the cursor is at the end of a sealed segment.
- **CHECKPOINT**: no broker-side cursor exists, so the backlog endpoint returns `NotFoundException`. `isSegmentDrained` (the existing helper) maps that to `false`, so a segment with any registered CHECKPOINT subscription is treated as "still in use" and never pruned. Conservative; safe default.
- **Parent-vs-child ordering**: a child can be pruned before its parent. Sealed segments form a DAG, and `pruneSegment` rewrites parent/child edges; the active leaves always cover the full hash range. Consumers using the post-prune layout see the missing segment as "no longer present" — equivalent to "drained" for STREAM parent-drain ordering.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: Done

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: done

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
merlimat: > Does this currently handle queue consumers too? Is it possible that a child segment gets pruned before one of it's parent segments?

Yes, since it checks the subscription, which is the same between queuing and streaming.

> Is it possible that a child segment gets pruned while a checkpoint consumer is actively consuming one of the parent segments?

Good point. Yes, the segment can definitely be deleting while the checkpoint consumer is reading, since it doesn't imply any retention. The client should be able to recover and skip the now deleted segment. I'll add in a separate PR since it imply more changes in client side.

lhotari (approved):
LGTM

PR #25669, opened 2026-05-05
[improve][client] PIP-468: Make Backoff jitter configurable

merlimat:
### Motivation

The existing `Backoff` class always subtracts up to 10% jitter from the next delay, never adds, and skips jitter on the first invocation entirely (the result is clamped back up to `initialDelay`). When many clients restart together, the first reconnect attempt fires at the exact configured initial delay — defeating the point of jitter. The jitter percentage was also hard-coded.

### Modifications

- `Backoff` takes a configurable `jitterPercent` (default 10).
- Jitter is applied symmetrically as `±jitterPercent/2`: the underlying delay is multiplied by a uniform random factor in `[1 − jitterPercent/200, 1 + jitterPercent/200)`.
- Jitter is applied on every call, including the first. The previous `Math.max(initial, …)` clamp is removed; the returned value may now be slightly below `initialDelay` or slightly above `maxBackoff`.
- `BackoffPolicy` (v5 client API) gains a `jitterPercent` field with a `withJitter(double)` helper. Existing factory methods seed the default jitter so callers are unaffected.

The new `BackoffPolicy.jitterPercent` is currently API-shape only: the connection-backoff path is still TODO'd in `PulsarClientBuilderV5`, and the redelivery-backoff path goes through `MultiplierRedeliveryBackoff` which has no jitter support. Wiring is left for a follow-up.

### Verifying this change

- New unit tests in `BackoffTest`:
  - `jitterIsAppliedSymmetricallyOnFirstCall` — across 500 resets verifies values both above and below the base, all within `[base · 0.9, base · 1.1]` for `jitterPercent=20`.
  - `jitterPercentZeroDisablesJitter` — verifies the disable path returns the exact base.
  - `negativeJitterIsRejected` — validation.
- Existing tests updated to set `jitterPercent(0)` for deterministic exponential-progression and mandatory-stop checks.

- [x] The public API

`org.apache.pulsar.common.util.Backoff` and `org.apache.pulsar.client.api.v5.config.BackoffPolicy` gain new optional fields/builder methods. Existing callers are unaffected (defaults preserve the 10% magnitude, though the direction now spans `±5%` instead of `−10%`).


pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/config/BackoffPolicy.java
```diff
@@ -20,136 +20,84 @@
...
-    }
-
-    /**
-     * @return the multiplier applied after each attempt
-     */
-    public double multiplier() {
-        return multiplier;
+        if (jitterPercent < 0) {
+            throw new IllegalArgumentException("jitterPercent must be >= 0");
+        }
```
lhotari: Should we limit jitterPercent to 100?

pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java
```diff
@@ -205,6 +213,24 @@ public Builder mandatoryStop(Duration mandatoryStop) {
...
+         * jitter.
+         *
+         * @param jitterPercent the jitter percentage, must be {@code >= 0}
+         * @return this builder
+         * @throws IllegalArgumentException if {@code jitterPercent} is negative
+         */
+        public Builder jitterPercent(double jitterPercent) {
+            if (jitterPercent < 0) {
+                throw new IllegalArgumentException("jitterPercent must be >= 0");
+            }
```
lhotari: Should we limit jitterPercent to 100?

pulsar-client-api-v5/src/main/java/org/apache/pulsar/client/api/v5/config/BackoffPolicy.java
merlimat: In theory it doesn't need to get capped at 100%, since you could fluctuate with >100%, though in practice it wouldn't make much sense for the backoff purpose.. let me add the check.

lhotari (approved):

PR #25679, opened 2026-05-05
[fix][test] Fix flaky OneWayReplicatorDeduplicationTest.testDeduplication

### LINKED ISSUE #25141: Flaky-test: OneWayReplicatorDeduplicationTest.testDeduplication
### Search before reporting

### Example failure

Adding Thread.sleep in the MessageDeduplication.takeSnapshot method can improve the trigger probability.

### Exception stacktrace

```text
org.awaitility.core.ConditionTimeoutException: Assertion condition expected [true] but found [false] within 10 seconds.

	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:985)
	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:769)
	at org.apache.pulsar.broker.service.OneWayReplicatorDeduplicationTest.testDeduplication(OneWayReplicatorDeduplicationTest.java:521)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:328)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at --- Async.Stack.Trace --- (captured by IntelliJ IDEA debugger)
	at java.base/java.util.concurrent.FutureTask.<init>(FutureTask.java:134)
	at java.base/java.util.concurrent.AbstractExecutorService.newTaskFor(AbstractExecutorService.java:113)
	at java.base/java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:147)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeWithTimeoutWithNewExecutor(MethodInvocationHelper.java:385)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeWithTimeout(MethodInvocationHelper.java:303)
	at org.testng.internal.invokers.TestInvoker.invokeMethod(TestInvoker.java:686)
	at org.testng.internal.invokers.TestInvoker.invokeTestMethod(TestInvoker.java:221)
	at org.testng.internal.invokers.MethodRunner.runInSequence(MethodRunner.java:50)
	at org.testng.internal.invokers.TestInvoker$MethodInvocationAgent.invoke(TestInvoker.java:969)
	at org.testng.internal.invokers.TestInvoker.invokeTestMethods(TestInvoker.java:194)
	at org.testng.internal.invokers.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:148)
	at org.testng.internal.invokers.TestMethodWorker.run(TestMethodWorker.java:128)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at org.testng.TestRunner.privateRun(TestRunner.java:829)
	at org.testng.TestRunner.run(TestRunner.java:602)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:437)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:431)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:391)
	at org.testng.SuiteRunner.run(SuiteRunner.java:330)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:95)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1256)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1176)
	at org.testng.TestNG.runSuites(TestNG.java:1099)
	at org.testng.TestNG.run(TestNG.java:1067)
	at com.intellij.rt.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:65)
	at com.intellij.rt.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:105)
Caused by: java.lang.AssertionError: expected [true] but found [false]
	at org.apache.pulsar.broker.service.OneWayReplicatorDeduplicationTest.lambda$testDeduplication$3(OneWayReplicatorDeduplicationTest.java:529)
	at org.awaitility.core.AssertionCondition.lambda$new$0(AssertionCondition.java:53)
	at org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:248)
	at org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:235)
```

Praveenkumar76:
Fixes [apache#25141](https://github.com/apache/pulsar/issues/25141)

### Motivation

`OneWayReplicatorDeduplicationTest.testDeduplication` is flaky because `MessageDeduplication.takeSnapshot` is designed to drop concurrent snapshot requests if an existing snapshot operation is already in progress. The test relies on specific positions being persisted, but if a manual trigger is dropped due to this intentional "lossy" design, the test fails with a `ConditionTimeoutException` while waiting for the state to update.

### Modifications

- Modified `OneWayReplicatorDeduplicationTest` to set `brokerDeduplicationSnapshotIntervalSeconds` to 1 second in the test configuration.

- By increasing the frequency of the background snapshot monitor, the system is guaranteed to capture and persist the deduplication state frequently enough to satisfy the test's assertions, even if a specific manual trigger request is skipped due to a race condition.

### Verifying this change

- Verified that the flaky test now runs consistently without failures.
- Confirmed that the 1-second background interval provides a sufficient "safety net" to persist the state when concurrent manual triggers are dropped.

- `org.apache.pulsar.broker.service.OneWayReplicatorDeduplicationTest.testDeduplication`

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorDeduplicationTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorDeduplicationTest.java
@@ -114,6 +114,7 @@ public class OneWayReplicatorDeduplicationTest extends OneWayReplicatorTestBase 
                                      LocalBookkeeperEnsemble bookkeeperEnsemble, ZookeeperServerTest brokerConfigZk) {
         super.setConfigDefaults(config, clusterName, bookkeeperEnsemble, brokerConfigZk);
         // For check whether deduplication snapshot has done.
+        config.setBrokerDeduplicationSnapshotIntervalSeconds(1);
         config.setBrokerDeduplicationEntriesInterval(10);
         config.setReplicationStartAt("earliest");
         // To cover more cases, write more than one ledger.


lhotari:
> The root cause is a race condition in `MessageDeduplication.takeSnapshot`. When multiple snapshot requests occur concurrently, the current implementation uses a `compareAndSet` guard to allow only one active snapshot. If another request arrives while a snapshot is already in progress, the method immediately returns a completed future, effectively dropping the new request.

The title of the PR should reflect that this is a production code change, not a fix to a flaky test.

lhotari (changes requested):
I don't think that MessageDuplication needs to be changed. The test itself is invalid and should be fixed instead. Here's the explanation of MessageDuplication, answer from DeepWiki:

> # `MessageDeduplication.takeSnapshot(Position)` — Role and Correctness
> 
> ## Role of the method
> 
> The private `takeSnapshot(Position)` in `org.apache.pulsar.broker.service.persistent.MessageDeduplication` persists the current deduplication state to a `ManagedCursor`. The snapshot captures the highest sequence IDs persisted per producer (the `highestSequencedPersisted` map) and marks the cursor at the supplied `Position`, which corresponds to the last confirmed entry in the `ManagedLedger`. This is what allows deduplication state to be recovered correctly after a broker restart or topic unload/reload.
> 
> ## When it's invoked
> 
> - Periodically, after `snapshotInterval` entries have been persisted (counter-driven, from `recordMessagePersistedNormal` / `recordMessagePersistedRepl`).
> - After replaying the cursor on a deduplication status check, if enough entries were processed.
> - After purging inactive producers, to persist the trimmed state.
> - From a scheduled task driven by `brokerDeduplicationSnapshotFrequencyInSeconds` / `brokerDeduplicationSnapshotIntervalSeconds`, via the public `takeSnapshot()` wrapper.
> - From `BacklogQuotaManager` when the deduplication cursor is the slowest consumer and needs to be advanced.
> 
> ## Correctness requirements when one is already in progress
> 
> Concurrency is gated by an `AtomicBoolean snapshotTaking`. If a snapshot is requested while another is in flight, the new request is **dropped (not queued)** and a warning is logged — there is no waiting and no replacement. The contract relies on two things to stay correct under that drop:
> 
> 1. The snapshot reads `highestSequencedPersisted` at invocation time and pairs it with the supplied `Position` (last confirmed entry), so each snapshot is internally consistent.
> 2. The underlying `ManagedCursor.markDelete` is monotonic — it never moves the cursor backward — so even though concurrent requests are skipped, missing one is safe: a later snapshot will advance the cursor to a position at least as far as the dropped one would have. Combined with the periodic/scheduled triggers, the cursor is guaranteed to keep moving forward.
> 
> The practical implication: callers must not assume their requested snapshot actually ran. The mechanism is designed to be lossy-but-monotonic — fine for periodic persistence, but not something to rely on for "snapshot exactly at this position right now" semantics.
> 
> ---
> 
> *Source: [DeepWiki query on apache/pulsar](https://deepwiki.com/search/in-orgapachepulsarbrokerservic_24d75d62-23c2-4a1f-a625-7d433ff820e3)*

One possible way to address the flaky test problem is to configure `brokerDeduplicationSnapshotIntervalSeconds` to a low value in the test instead of relying on `brokerDeduplicationEntriesInterval`. This shouldn't change the intention of the test at all.

Praveenkumar76:
Thanks for the detailed explanation, @lhotari.  

My original change aimed to resolve the `ConditionTimeoutException` by ensuring the test never missed a snapshot signal, but I now understand that the 'lossy-but-monotonic' design is intentional for production performance.  

I agree that fixing the test configuration is the better approach here, I will:

Revert the changes to MessageDeduplication.java.  

Update the test to use a lower brokerDeduplicationSnapshotIntervalSeconds as suggested to ensure frequent snapshots during verification.  

Update the PR title/description to reflect a test stabilization change.

Pushing the update shortly.

lhotari (approved):
LGTM

lhotari:
Thanks for the contribution @Praveenkumar76

PR #25681, opened 2026-05-05
[fix][broker] Correct two race conditions in the tracker code and logic bug in InMemoryDelayedDeliveryTracker that failed with NoSuchElementException

### LINKED ISSUE #25617: [Bug] NoSuchElementException in InMemoryDelayedDeliveryTracker.nextDeliveryTime broke message dispatching until broker was restarted
### Search before reporting

### Read release policy

### User environment

Broker Version
```
$ bin/pulsar version
Current version of pulsar is: 4.1.2
Git Revision 5f587c1204d13ec6b657c7eef472f8427368a9f6
Git Branch branch-4.1
Built by Lari Hotari <lhotari@apache.org> on Laris-MBP-2.lan at 2025-11-11T23:36:57+0200
```
Broker OS:
```
$ uname -a
Linux pulsar-broker-0 6.6.113+ #1 SMP Wed Jan 28 03:26:59 UTC 2026 aarch64 GNU/Linux
```
Broker Java:
```
$ java -version
openjdk version "21.0.9" 2025-10-21 LTS
OpenJDK Runtime Environment Corretto-21.0.9.11.1 (build 21.0.9+11-LTS)
OpenJDK 64-Bit Server VM Corretto-21.0.9.11.1 (build 21.0.9+11-LTS, mixed mode)
```

Client Library:
https://github.com/streamnative/pulsar-rs
```
pulsar = { version = "6.7.2", features = [
    "compression",
    "tokio-runtime",
], default-features = false }
```

Client OS Version - Same as Broker OS

### Issue Description

We have a production broker and bookie cluster of 6 nodes each. While scaling up a new feature of our service that wrote high volume messages to a topic with a deliver_at_time mostly set for the upcoming minute, we saw this exposed pulsar metric increase over time:

`sum by (topic)(pulsar_subscription_back_log_no_delayed{namespace=~"pulsar-$environment", subscription="delivery-subscription"})`

Image

This was despite the fact that our consumers were consuming from that topic actively and we had plenty of headroom to consume more messages from our consumers.

After an investigation, we restarted our consumer pool to no effect, and then restarted the brokers (a rolling restart). We then saw a _massive_ spike from 42.7 thousand to to 2.64 _million_ entries on that topic, which then started to be processed by our consumer group.

Image

Investigating the messages now being sent, we found messages from more than a week ago, including some that we have evidence were previously processed and acked successfully.

Thankfully we had added a stale message detection to our consumer, so this was only a data loss event and not something worse (sending messages from a week ago).

Investigating the logs, we found 165 instances of java.util.NoSuchElementException coming from Long2ObjectAVLTreeMap/InMemoryDelayedDeliveryTracker (see error message attached), with some of those correlating pretty closely to increases in the "stuck" backlog or high volume writes to pulsar. 

There are three bugs here from my perspective:

- Why did the broker cluster stop dispatching messages that had passed their deliver_at_time for this topic until the broker was restarted?
- Why did restarting the broker cause multiple order of magnitude more messages to be (re)? send to our consumers?
- Why did pulsar have uncaught exceptions that appear to be the source of this bug only reported by an _info_ level log? Why did the broker not fail a health check or otherwise notify us via metrics?

This was a very serious bug for us, causing non-trivial data loss as our messages have a QoS timing window for prompt delivery that we missed. 

If an engineer has not manually noticed that a reported metric was increasing over time, this could have gone on for much longer than a week. If we had not had a stale message detector, we could have reprocessed and resent messages we sent multiple days ago (that our standard pulsar idempotency detector would have missed). The fact that this issue was manually discovered by a lucky glance as a dashboard and not due to failing health checks or other standard metrics also has us concerned about the suitability for our high performance workload. 

### Error messages

```text
java.util.NoSuchElementException: null
	at it.unimi.dsi.fastutil.longs.Long2ObjectAVLTreeMap.firstLongKey(Long2ObjectAVLTreeMap.java:934) ~[it.unimi.dsi-fastutil-8.5.16.jar:?]
	at org.apache.pulsar.broker.delayed.InMemoryDelayedDeliveryTracker.nextDeliveryTime(InMemoryDelayedDeliveryTracker.java:273) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.delayed.AbstractDelayedDeliveryTracker.updateTimer(AbstractDelayedDeliveryTracker.java:99) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.delayed.InMemoryDelayedDeliveryTracker.getScheduledMessages(InMemoryDelayedDeliveryTracker.java:229) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.getMessagesToReplayNow(PersistentDispatcherMultipleConsumers.java:1327) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.readMoreEntries(PersistentDispatcherMultipleConsumers.java:385) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.lambda$readMoreEntriesAsync$4(PersistentDispatcherMultipleConsumers.java:324) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:173) ~[io.netty-netty-common-4.1.128.Final.jar:4.1.128.Final]
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:166) ~[io.netty-netty-common-4.1.128.Final.jar:4.1.128.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:472) ~[io.netty-netty-common-4.1.128.Final.jar:4.1.128.Final]
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:405) ~[io.netty-netty-transport-classes-epoll-4.1.128.Final.jar:4.1.128.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:998) ~[io.netty-netty-common-4.1.128.Final.jar:4.1.128.Final]
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[io.netty-netty-common-4.1.128.Final.jar:4.1.128.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty-netty-common-4.1.128.Final.jar:4.1.128.Final]
```

### Reproducing the issue

~We are attempting to reproduce this in a synthetic testing environments and will update the issue if we can come up with one.~

~Our average write and process speed is on the order of 1.5k / second on roughly 20 topics. This feature would write a much smaller amount (order 10/second) with short but intense spikes of 4k / second or more with deliver_after_time that are very similar or identical times set.~

See https://github.com/chamons/pulsar-scheduled-exception-repro

### Additional information

I found https://github.com/apache/pulsar/issues/25174 which has some of the same hallmarks:

- A broker exception
- That seems to cause scheduled messages to not be dispatched
- That is resolved by restarting the broker

However, the stack trace is different. 

chamons:
Fixes https://github.com/apache/pulsar/issues/25617

### Motivation

While running a production system under load, we found a traffic pattern that could cause uncaught exceptions in the brokers of our pulsar cluster. If you have a large spike of messages with a matching deliver_at_time, it is possible to trigger java.util.NoSuchElementException and other exceptions:

```
	at it.unimi.dsi.fastutil.longs.Long2ObjectAVLTreeMap.firstLongKey(Long2ObjectAVLTreeMap.java:934) ~[it.unimi.dsi-fastutil-8.5.16.jar:?]
	at org.apache.pulsar.broker.delayed.InMemoryDelayedDeliveryTracker.nextDeliveryTime(InMemoryDelayedDeliveryTracker.java:273) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.delayed.AbstractDelayedDeliveryTracker.updateTimer(AbstractDelayedDeliveryTracker.java:99) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.delayed.InMemoryDelayedDeliveryTracker.getScheduledMessages(InMemoryDelayedDeliveryTracker.java:229) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.getMessagesToReplayNow(PersistentDispatcherMultipleConsumers.java:1327) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.readMoreEntries(PersistentDispatcherMultipleConsumers.java:385) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
	at org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumers.lambda$readMoreEntriesAsync$4(PersistentDispatcherMultipleConsumers.java:324) ~[org.apache.pulsar-pulsar-broker-4.1.2.jar:4.1.2]
```

This is a significant issue, as exceptions here were uncaught and put the broker into an invalid state that prevented some delayed messages from being delivered until restart.

Using our external repro as a guide, we were able to isolate two race conditions and a logic bug that triggered exceptions that matched these. 

With these three fixes, we were able to run our sample case for 30 minutes straight without exceptions (before hand it would trigger within 3-5 minutes).

### Modifications

Posting each commit's description here for simplicity:

```
[fix][broker] Correct multiple race conditions in PersistentDispatcherMultipleConsumers
- https://github.com/apache/pulsar/issues/25617

DelayedDeliveryTracker is not thread safe, and any access to it must be done
holding the object lock. There were five cases I found, four of them I could
correct just by adding synchronized to the method declaration. In one of them
I used a manual scope since it was only a subset of the method.

This is a significant issue, as exceptions here were uncaught and put
the broker into an invalid state that prevented some delayed messages
from being delivered until restart.

The included unit test failed 100% of the time when run locally without the fix.
```

```
[fix][broker] Fix multiple race conditions in InMemoryDelayedDeliveryTracker

While reviewing the code for more issues to explain the crashes we were still seeing,
I noticed that InMemoryDelayedDeliveryTracker contains two major state fields:

- delayedMessageMap
- delayedMessagesCount

These are kept in sync in various methods, but those methods were _not_ correctly
synchronized, which meant it was trivial for us to get them out of sync
and then crash peeking at the map.

Added a unit test that triggers 100% of the time before the fix.
```

```
[fix][broker] Fix crash when duplicate ledger/entry messages come into InMemoryDelayedDeliveryTracker

When the same ledger/entry messages come into InMemoryDelayedDeliveryTracker, we can desync the
delayedMessagesCount/delayedMessageMap count such that there are less items in the map than the count

Instead, we track if the bitmap doesn't already contain the item, and only increment delayedMessagesCount if not.
```

### Verifying this change

I have a clean [CI here](https://github.com/chamons/pulsar/actions/runs/25519871246?pr=2)

image

The included unit tests fail 100% of the time when run locally without the fix, and we were able to run https://github.com/chamons/pulsar-scheduled-exception-repro without issues with the fixes included.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -127,16 +127,24 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
             messagesHaveFixedDelay = false;
             return false;
         }
-            log.debug()
-                    .attr("ledgerId", ledgerId)
-                    .attr("entryId", entryId)
-                    .attr("deliveryInMs", () -> deliverAt - clock.millis())
-                    .log("Add message");
-                long timestamp = trimLowerBit(deliverAt, timestampPrecisionBitCnt);
-        delayedMessageMap.computeIfAbsent(timestamp, k -> new TreeMap<>())
-                .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap())
-                .add(entryId);
-        delayedMessagesCount.incrementAndGet();
+
+        log.debug()
+                .attr("ledgerId", ledgerId)
+                .attr("entryId", entryId)
+                .attr("deliveryInMs", () -> deliverAt - clock.millis())
+                .log("Add message");
+        long timestamp = trimLowerBit(deliverAt, timestampPrecisionBitCnt);
+
+        Roaring64Bitmap bitmap = delayedMessageMap.computeIfAbsent(timestamp, k -> new TreeMap<>())
+            .computeIfAbsent(ledgerId, k -> new Roaring64Bitmap());
+        // Roaring64Bitmap does not store duplicates, so track if it a new element
+        // so we can keep delayedMessagesCount in sync
+        boolean isNew = !bitmap.contains(entryId);
+
+        if (isNew) {
+            bitmap.add(entryId);
+            delayedMessagesCount.incrementAndGet();
+        }
 
         updateTimer();
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -122,7 +122,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     }
 
     @Override
-    public synchronized boolean addMessage(long ledgerId, long entryId, long deliverAt) {
+    public boolean addMessage(long ledgerId, long entryId, long deliverAt) {
         if (deliverAt < 0 || deliverAt <= getCutoffTime()) {
             messagesHaveFixedDelay = false;
             return false;
@@ -169,7 +169,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
      * Return true if there's at least a message that is scheduled to be delivered already.
      */
     @Override
-    public synchronized boolean hasMessageAvailable() {
+    public boolean hasMessageAvailable() {
         boolean hasMessageAvailable = !delayedMessageMap.isEmpty()
                 && delayedMessageMap.firstKey() <= getCutoffTime();
         if (!hasMessageAvailable) {
@@ -182,7 +182,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
      * Get a set of position of messages that have already reached.
      */
     @Override
-    public synchronized NavigableSet<Position> getScheduledMessages(int maxMessages) {
+    public NavigableSet<Position> getScheduledMessages(int maxMessages) {
         int n = maxMessages;
         NavigableSet<Position> positions = new TreeSet<>();
         long cutoffTime = getCutoffTime();
@@ -245,14 +245,14 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     }
 
     @Override
-    public synchronized CompletableFuture<Void> clear() {
+    public CompletableFuture<Void> clear() {
         this.delayedMessageMap.clear();
         this.delayedMessagesCount.set(0);
         return CompletableFuture.completedFuture(null);
     }
 
     @Override
-    public synchronized long getNumberOfDelayedMessages() {
+    public long getNumberOfDelayedMessages() {
         return delayedMessagesCount.get();
     }
 
@@ -263,14 +263,14 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
      * @return the memory usage of the buffer
      */
     @Override
-    public synchronized long getBufferMemoryUsage() {
+    public long getBufferMemoryUsage() {
         return delayedMessageMap.values().stream().mapToLong(
                 ledgerMap -> ledgerMap.values().stream().mapToLong(
                         Roaring64Bitmap::getLongSizeInBytes).sum()).sum();
     }
 
     @Override
-    public synchronized void close() {
+    public void close() {
         super.close();
     }
 
@@ -283,7 +283,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
                 && !hasMessageAvailable();
     }
 
-    protected synchronized long nextDeliveryTime() {
+    protected long nextDeliveryTime() {
         return delayedMessageMap.firstKey();
     }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
@@ -122,7 +122,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     }
 
     @Override
-    public boolean addMessage(long ledgerId, long entryId, long deliverAt) {
+    public synchronized boolean addMessage(long ledgerId, long entryId, long deliverAt) {
         if (deliverAt < 0 || deliverAt <= getCutoffTime()) {
             messagesHaveFixedDelay = false;
             return false;
@@ -161,7 +161,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
      * Return true if there's at least a message that is scheduled to be delivered already.
      */
     @Override
-    public boolean hasMessageAvailable() {
+    public synchronized boolean hasMessageAvailable() {
         boolean hasMessageAvailable = !delayedMessageMap.isEmpty()
                 && delayedMessageMap.firstKey() <= getCutoffTime();
         if (!hasMessageAvailable) {
@@ -174,7 +174,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
      * Get a set of position of messages that have already reached.
      */
     @Override
-    public NavigableSet<Position> getScheduledMessages(int maxMessages) {
+    public synchronized NavigableSet<Position> getScheduledMessages(int maxMessages) {
         int n = maxMessages;
         NavigableSet<Position> positions = new TreeSet<>();
         long cutoffTime = getCutoffTime();
@@ -237,14 +237,14 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
     }
 
     @Override
-    public CompletableFuture<Void> clear() {
+    public synchronized CompletableFuture<Void> clear() {
         this.delayedMessageMap.clear();
         this.delayedMessagesCount.set(0);
         return CompletableFuture.completedFuture(null);
     }
 
     @Override
-    public long getNumberOfDelayedMessages() {
+    public synchronized long getNumberOfDelayedMessages() {
         return delayedMessagesCount.get();
     }
 
@@ -255,14 +255,14 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
      * @return the memory usage of the buffer
      */
     @Override
-    public long getBufferMemoryUsage() {
+    public synchronized long getBufferMemoryUsage() {
         return delayedMessageMap.values().stream().mapToLong(
                 ledgerMap -> ledgerMap.values().stream().mapToLong(
                         Roaring64Bitmap::getLongSizeInBytes).sum()).sum();
     }
 
     @Override
-    public void close() {
+    public synchronized void close() {
         super.close();
     }
 
@@ -275,7 +275,7 @@ public class InMemoryDelayedDeliveryTracker extends AbstractDelayedDeliveryTrack
                 && !hasMessageAvailable();
     }
 
-    protected long nextDeliveryTime() {
+    protected synchronized long nextDeliveryTime() {
         return delayedMessageMap.firstKey();
     }
 }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -427,7 +427,7 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
         }
     }
 
-    protected Predicate<Position> createReadEntriesSkipConditionForNormalRead() {
+    protected synchronized Predicate<Position> createReadEntriesSkipConditionForNormalRead() {
         Predicate<Position> skipCondition = null;
         // Filter out and skip read delayed messages exist in DelayedDeliveryTracker
         if (delayedDeliveryTracker.isPresent()) {
@@ -1378,7 +1378,7 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
     }
 
     @Override
-    public long getNumberOfDelayedMessages() {
+    public synchronized long getNumberOfDelayedMessages() {
         return delayedDeliveryTracker.map(DelayedDeliveryTracker::getNumberOfDelayedMessages).orElse(0L);
     }
 
@@ -1389,7 +1389,9 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
         }
 
         if (delayedDeliveryTracker.isPresent()) {
-            return this.delayedDeliveryTracker.get().clear();
+            synchronized (this) {
+                return this.delayedDeliveryTracker.get().clear();
+            }
         } else {
             DelayedDeliveryTrackerFactory delayedDeliveryTrackerFactory =
                     topic.getBrokerService().getDelayedDeliveryTrackerFactory();
@@ -1464,11 +1466,11 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
     }
 
 
-    public long getDelayedTrackerMemoryUsage() {
+    public synchronized long getDelayedTrackerMemoryUsage() {
         return delayedDeliveryTracker.map(DelayedDeliveryTracker::getBufferMemoryUsage).orElse(0L);
     }
 
-    public Map<String, TopicMetricBean> getBucketDelayedIndexStats() {
+    public synchronized Map<String, TopicMetricBean> getBucketDelayedIndexStats() {
         if (delayedDeliveryTracker.isEmpty()) {
             return Collections.emptyMap();
         }
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
@@ -1383,15 +1383,13 @@ public class PersistentDispatcherMultipleConsumers extends AbstractPersistentDis
     }
 
     @Override
-    public CompletableFuture<Void> clearDelayedMessages() {
+    public synchronized CompletableFuture<Void> clearDelayedMessages() {
         if (!topic.isDelayedDeliveryEnabled()) {
             return CompletableFuture.completedFuture(null);
         }
 
         if (delayedDeliveryTracker.isPresent()) {
-            synchronized (this) {
-                return this.delayedDeliveryTracker.get().clear();
-            }
+            return this.delayedDeliveryTracker.get().clear();
         } else {
             DelayedDeliveryTrackerFactory delayedDeliveryTrackerFactory =
                     topic.getBrokerService().getDelayedDeliveryTrackerFactory();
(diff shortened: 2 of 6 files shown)


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumersTest.java
```diff
@@ -152,4 +160,96 @@ public void testSkipReadEntriesFromCloseCursor() throws Exception {
         // Verify: the topic can be deleted successfully.
         admin.topics().delete(topicName, false);
     }
+
+    @Test
+    public void testRaceConditionInTrackDelayedDelivery() throws Exception {
```
chamons, line 163: This is based on pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/bucket/BucketDelayedDeliveryTrackerThreadSafetyTest.java

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -1379,7 +1379,9 @@ protected synchronized boolean shouldPauseDeliveryForDelayTracker() {
 
     @Override
     public long getNumberOfDelayedMessages() {
-        return delayedDeliveryTracker.map(DelayedDeliveryTracker::getNumberOfDelayedMessages).orElse(0L);
+        synchronized (this) {
```
merlimat: nit: this could be `synchronized` on the method itself

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumersTest.java
```diff
@@ -152,4 +160,96 @@ public void testSkipReadEntriesFromCloseCursor() throws Exception {
...
+        long deliverAt = System.currentTimeMillis() + 5000;
+
+        MessageMetadata messageMetadata = new MessageMetadata()
+            .setSequenceId(1)
+            .setProducerName("testProducer")
+            .setPartitionKeyB64Encoded(false)
+            .setPublishTime(System.currentTimeMillis())
+            .setDeliverAtTime(deliverAt);
+
+        ExecutorService executorService = Executors.newFixedThreadPool(32);
```
merlimat, line 202: These are many threads for a unit test :) 

The executor needs to get shutdown. Use `@Cleanup("shutdown")`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
chamons: Can you add synchronized on a @override method? I thought the signature had to match?

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumersTest.java
chamons: Good catch. I borrowed this from BucketDelayedDeliveryTrackerThreadSafetyTest but they have a cleanup method I missed.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
chamons: TIL

chamons:
@merlimat - We found that this is only a partial fix for https://github.com/apache/pulsar/issues/25617, as there is at least one other set of race conditions that can trigger it.

Would you prefer a single larger PR with all of the included or should I fix this up and land piecemeal?

pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
```diff
@@ -274,4 +279,111 @@ public void testDelaySequence(InMemoryDelayedDeliveryTracker tracker) throws Exc
...
+    @Test(dataProvider = "delayedTracker")
+    public void testAddMultipleMessagesSameWindow(InMemoryDelayedDeliveryTracker tracker) throws Exception {
+        tracker.addMessage(1, 1, 50);
+        tracker.addMessage(1, 1, 50);
+        tracker.addMessage(1, 1, 50);
+
+        clockTime.set(60);
+
+        tracker.getScheduledMessages(10);
+    }
```
mattday9: I don't see any assertions. It's just executing and flow through.

pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
```diff
@@ -274,4 +279,111 @@ public void testDelaySequence(InMemoryDelayedDeliveryTracker tracker) throws Exc
...
+                try {
+                    startLatch.await();
+                    for (int j = 0; j < operationsPerThread; j++) {
+                        tracker.clear();
+                        Thread.sleep(1);
+                    }
+                } catch (Exception e) {
+                    errors.incrementAndGet();
+                    firstException.compareAndSet(null, e);
+                    e.printStackTrace();
```
mattday9: nit: Typically shows a yellow line in the IDE.

pulsar-broker/src/test/java/org/apache/pulsar/broker/delayed/InMemoryDeliveryTrackerTest.java
```diff
@@ -274,4 +279,111 @@ public void testDelaySequence(InMemoryDelayedDeliveryTracker tracker) throws Exc
...
+                } finally {
+                    doneLatch.countDown();
+                }
+            });
+        }
+
+        startLatch.countDown();
+        assertTrue(doneLatch.await(30, TimeUnit.SECONDS), "Test should complete within 30 seconds");
+
+        if (errors.get() > 0) {
```
mattday9: == 1 ?

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
```diff
@@ -122,21 +122,29 @@ private static long trimLowerBit(long timestamp, int bits) {
     }
 
     @Override
-    public boolean addMessage(long ledgerId, long entryId, long deliverAt) {
+    public synchronized boolean addMessage(long ledgerId, long entryId, long deliverAt) {
```
lhotari: To keep the performance characteristics of the previous implementation, it would be necessary to use
`synchronized (dispatcher) { ... }` blocks instead of making the methods `synchronized`. Java runtime will make that a no-op when the thread already holds the dispatcher's lock which is the current intended design.

A future improvement could be to make this pattern handled by a base class, which would extend `AbstractDelayedDeliveryTracker`, so that the public methods would marked final and the implementing class would implement a protected method with a different name, for example `internalAddMessage`. This could be handled in another PR so that the scope of this PR doesn't go beyond the bug fix.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
lhotari, line 1: The changes to make methods `synchronized` in this class are fine.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
```diff
@@ -1389,7 +1389,9 @@ public CompletableFuture<Void> clearDelayedMessages() {
         }
 
         if (delayedDeliveryTracker.isPresent()) {
-            return this.delayedDeliveryTracker.get().clear();
+            synchronized (this) {
+                return this.delayedDeliveryTracker.get().clear();
+            }
```
lhotari: for consistency, it could be useful to make the method itself `synchronized`. That would also cover the `BucketDelayedDeliveryTrackerFactory` case.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
lhotari: Similar fixes should be made also to `org.apache.pulsar.broker.service.persistent.PersistentDispatcherMultipleConsumersClassic` class, which is the legacy implementation (before PIP-379, can be activated with `subscriptionKeySharedUseClassicPersistentImplementation=true` and `subscriptionSharedUseClassicPersistentImplementation=true`).

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
lhotari: I looked again and I didn't find execution paths which would by pass the synchronized locks of the dispatcher (included in this PR). Have you checked if the bugs reproduce if these methods aren't synchronized? If that's the case, it we could remove the `synchronized` handling from this PR and create a separate PR to prevent future regressions with the `AbstractDelayedDeliveryTracker` refactoring.

btw. Comment about the existing code (not this PR): One detail that annoys me is the separate handling for `BucketDelayedDeliveryTracker` implementation in the dispatcher. It's not great design. The `DelayedDeliveryTracker` interface should have been extended to cover the additional integration methods instead of coupling directly to the implementation which is the whole point of using an interface (which is in place to allow third-part implementations by configuring `dispatchRateLimiterFactoryClassName`)

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
chamons: I will apply those changes this morning along with a duplicate copy of the test.

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
chamons: If I [remove the synchronized in that file](https://gist.github.com/chamons/954c0d442ecee83817868e4c1d838704)

My new test (org.apache.pulsar.broker.delayed.InMemoryDeliveryTrackerTest) fails with [multiple different exceptions](https://gist.github.com/chamons/2bf7e73f9b72901c3db9958e7940b775). 

```
    java.lang.AssertionError: No exceptions should occur during concurrent operations expected [0] but found [5]
```

The test itself seemed reasonable.

I might have added too many, I went with the heuristic of "delayedMessageMap and delayedMessagesCount must be kept in sync, so any method that uses them should hold the lock". Close is the most obvious case where that might be overkill, I'm unsure.

lhotari: The test is expected to fail since the existing design in assumes that the dispatcher synchronizes the access. One way to address this is expand the scope of this PR to handle the refactoring which takes care of synchronization in `AbstractDelayedDeliveryTracker` so that the subclasses like `InMemoryDeliveryTracker` wouldn't need to worry about the synchronization concerns. (for example making addMessage final in `AbstractDelayedDeliveryTracker` and delegating to an abstract method `internalAddMessage` which is executed inside a `synchronized(dispatcher) { .... }` block. Another approach is to remove this test from this PR and add it to a separate PR which refactors `AbstractDelayedDeliveryTracker` as described.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentDispatcherMultipleConsumers.java
chamons: Done!

pulsar-broker/src/main/java/org/apache/pulsar/broker/delayed/InMemoryDelayedDeliveryTracker.java
chamons: I will test reverting InMemoryDelayedDeliveryTracker against our [stress test sample](https://github.com/chamons/pulsar-scheduled-exception-repro/) and see if it still works and report back.

chamons: We were able to confirm that removing this section of the PR does not impact the fix. I have removed it.

chamons:
I have resolved all current PR review comments and am ready for a re-review.

lhotari (approved):
LGTM, thanks for the important bug fix and contribution @chamons! Good job!

chamons:
Thanks @lhotari - Do we have any idea when this fix might make it into a public release? Could it make it into a 4.x branch? 

We'd like to make some plans, and knowing what timetable to expect would be really helpful.

merlimat:
Yes, it's a bug fix and it will be backported into 4.0.x (prob 4.0.11) release

lhotari:
@chamons I have cherry-picked in this to branch-4.0, branch-4.1 and branch-4.2 . Since the implementations in each branch is slightly different, the change has been backported. Please note that 4.1.x is no longer supported, so the release won't be made on that branch although the fix is present on the branch. It's recommended to upgrade to 4.2.2 once it comes out within about a month. In the mean time, you can build your own release. The Pulsar release is performed using these steps: https://pulsar.apache.org/contribute/release-process/#build-release-artifacts . The ["Build release artifacts"](https://pulsar.apache.org/contribute/release-process/#build-release-artifacts) and ["Stage docker images"](https://pulsar.apache.org/contribute/release-process/#release-pulsar-30-and-later) are the only required steps in this case.

chamons:
@lhotari - I wanted to check in and see if we had a better idea when an official release for 4.2.x is coming out?

I have a local handmade image we're using for testing, but as one can imagine we'd like to use an official release image if possible.

lhotari:
> @lhotari - I wanted to check in and see if we had a better idea when an official release for 4.2.x is coming out?
> 
> I have a local handmade image we're using for testing, but as one can imagine we'd like to use an official release image if possible.

I'll propose starting the 4.0.11 and 4.2.2 releases this week. Decisions about releases are made on the [dev mailing list](https://pulsar.apache.org/contact/#mailing-lists). After the first release candidate has been published, it takes about 1 week to complete the release.

lhotari:
FYI, there's another fix including a performance improvement in #26012, will be part of 4.2.3/4.0.12 releases in about 2 months.

PR #25682, opened 2026-05-05
[feat] PIP-468: V5 end-to-end encryption API redesign

merlimat:
## Summary

Reshape the V5 end-to-end encryption surface so first-time setup is straightforward and misconfiguration is harder to express:

- **Role-specific async SPI**: split the dual-role `CryptoKeyReader` into `PublicKeyProvider` (producer side) and `PrivateKeyProvider` (consumer side). Both return `CompletableFuture<EncryptionKey>` so KMS-backed implementations don't have to block. `CryptoKeyProvider` extends both for in-process use.
- **Per-side failure actions**: split `CryptoFailureAction` into `ProducerCryptoFailureAction` (`FAIL`, `SEND_UNENCRYPTED`) and `ConsumerCryptoFailureAction` (`FAIL`, `DISCARD`, `CONSUME`). Compile-time prevents producer-side DISCARD/CONSUME misconfigurations.
- **Per-side policy types**: split `EncryptionPolicy` into `ProducerEncryptionPolicy` and `ConsumerEncryptionPolicy`. The consumer policy accepts a null `privateKeyProvider` when the failure action is `DISCARD` or `CONSUME` — matching v4's "no key reader, just receive raw bytes" mode.
- **Cleaner value type**: `EncryptionKeyInfo` → `EncryptionKey` (final class, private ctor + `of(...)` factories). Avoids records' broken `byte[]` equality.
- **Batteries-included default**: new `PemFileKeyProvider` for the common load-from-disk case. Eliminates v4's friction of having to implement the SPI from scratch just to encrypt one message.
- **Internal adapter**: `CryptoKeyReaderAdapter` exposes role-specific `forProducer` / `forConsumer` factories that bridge to v4's combined `CryptoKeyReader`.

## Test plan

- [x] `PemFileKeyProviderTest` (5 unit tests) — registered keys, missing key, missing file, dual-role usage.
- [x] `EncryptionPolicyTest` (9 unit tests) — builder validation, defaults, optional consumer provider for DISCARD/CONSUME.
- [x] `V5EncryptionTest` (4 end-to-end tests):
  - Producer encrypts → consumer decrypts on a single-segment scalable topic.
  - 30 messages across a 3-segment scalable topic, all decrypt correctly.
  - `CONSUME` without a provider delivers raw encrypted bytes.
  - `DISCARD` without a provider silently drops the message.
- [x] `pulsar-client-api-v5`, `pulsar-client-v5`, `pulsar-broker` checkstyle clean.


lhotari (approved):
LGTM

PR #25687, opened 2026-05-05
[feat] PIP-468: DagWatchClient auto-reconnects on broker disconnect

merlimat:
## Summary

Replace the TODO in `DagWatchClient.connectionClosed()` that left long-lived producers / consumers stranded on a stale layout after a transient broker disconnect. The DAG watch is the consumer / producer's only path to learning about split / merge events; without reconnect, a single network blip silently drops the client off future layout updates and never surfaces an error to the application.

The new logic mirrors the proven pattern used by `ScalableTopicsWatcher` (namespace-watcher session) and `ScalableConsumerClient` (controller session):

- `start()` (initial create) keeps its current behavior — the future is failed on disconnect so `producer.create()` / `consumer.subscribe()` fails fast when the broker is unreachable.
- After the initial layout has arrived, `connectionClosed()` schedules a reconnect with exponential backoff (100 ms → 30 s). The reconnect path calls `getConnection` again, re-registers the same `sessionId` on the fresh `ClientCnx`, and re-issues `ScalableTopicLookup`; the broker re-pushes the current layout via `onUpdate`, which resets the backoff.
- The connect / reconnect logic is shared in a single `attach(cnx)` method so both paths use identical wiring.

Adds a `forceCloseConnectionForTesting()` hook (mirroring the one on `ScalableConsumerClient`) for cross-module integration tests.

## Test plan

- [x] `V5DagWatchAutoReconnectTest` (3 integration tests):
  - Producer's DAG channel force-closed mid-life — sends after the close still succeed.
  - Consumer's DAG channel force-closed mid-life — receives after the close still succeed.
  - Reflective check that the `cnx` field on `DagWatchClient` gets re-populated within the backoff window.
- [x] Full V5 broker test suite green (128/128).
- [x] `pulsar-client-v5` and `pulsar-broker` checkstyle clean.


lhotari (approved):
LGTM

PR #25547, opened 2026-04-17
[feat][pip] PIP-469: Legacy-aware topic policies backend routing and metadata-store topic policies

BewareMyPower:

PR #25707, opened 2026-05-07
[feat][broker] PIP-469: Legacy-aware topic policies backend routing and metadata-store topic policies

BewareMyPower:
pip: #25547 

This PR reuses `TopicPoliciesTest` to test most functionality of the new metadata store based implementation, so it also introduces improvements on `TopicPoliciesTest`, which requires too much time to run before (100+ tests, where each test calls `internalSetup` and `internalCleanup`).


pulsar-broker/src/test/java/org/apache/pulsar/broker/service/TopicPolicyTestUtils.java
```diff
@@ -72,6 +72,13 @@ public static TopicPolicies getGlobalTopicPolicies(TopicPoliciesService topicPol
     public static Optional<TopicPolicies> getTopicPoliciesBypassCache(TopicPoliciesService topicPoliciesService,
                                                                       TopicName topicName, boolean isGlobal)
             throws Exception {
+        if (topicPoliciesService instanceof LegacyAwareTopicPoliciesService legacyService) {
+            TopicPoliciesService resolved = legacyService.resolveService(topicName.getNamespaceObject()).get();
+            return getTopicPoliciesBypassCache(resolved, topicName, isGlobal);
+        }
+        if (topicPoliciesService instanceof MetadataStoreTopicPoliciesService metadataStoreService) {
+            return metadataStoreService.getTopicPoliciesDirectFromStore(topicName, isGlobal).get();
+        }
```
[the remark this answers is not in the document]
BewareMyPower: The test should be responsible to set a timeout when calling `getTopicPoliciesBypassCache`,

BewareMyPower:
We might need to consider whether to change the `registerListener` API because we cannot register listener for both `LegacyAwareTopicPoliciesService` and `SystemTopicBasedTopicPoliciesService`.

dao-jun (approved):
LGTM, just 2 minor comments


  1. Metadata-store listeners are never unregistered on close() (resource leak)

  MetadataStoreTopicPoliciesService.start() line 354-355:

  localStore.registerListener(notification -> handleNotification(notification, false));
  configurationStore.registerListener(notification -> handleNotification(notification, true));

  The return value of registerListener (a registration handle) is discarded. close() only clears the local listeners map and invalidates caches, but never unregisters these metadata-store listeners. After close(), the service can still receive and process notifications (the closed flag guards most paths, but the listener
  lambdas themselves are still held by the metadata store). Consider capturing and closing the registration handles in close().

  2. Path structure deviates from PIP-469 design document

  The PIP design doc specifies:
  - Global: /admin/topic-policies/{tenant}/{namespace}/{domain}/{encodedTopic}
  - Local: /admin/local-policies/topic-policies/{tenant}/{namespace}/{domain}/{encodedTopic}

  The implementation uses:
  - Global: /admin/topic-policies/global/{tenant}/{namespace}/{domain}/{encodedTopic}
  - Local: /admin/topic-policies/local/{tenant}/{namespace}/{domain}/{encodedTopic}

  The code's approach is arguably cleaner (both scopes under one root), but the discrepancy from the approved design should be explicitly called out and either the PIP or the code should be updated for consistency.

BewareMyPower:
Let me address these two comments before merging

BewareMyPower:
@dao-jun I've updated the document for why the implementation adopts a different path.

> The return value of registerListener (a registration handle) is discarded. close() only clears the local listeners map and invalidates caches, but never unregisters these metadata-store listeners. After close(), the service can still receive and process notifications (the closed flag guards most paths, but the listener
lambdas themselves are still held by the metadata store). Consider capturing and closing the registration handles in close().

Regarding the previous comment, it seems to be wrong. Firstly, `registerListener`  does not return any handle, see https://github.com/apache/pulsar/blob/8f9f5b49d631e235e86d79e48a63722e74db4413/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStore.java#L249

Secondly, it follows the similar pattern of https://github.com/apache/pulsar/blob/8f9f5b49d631e235e86d79e48a63722e74db4413/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java#L381-L384

Though this PR has registered two listeners with a parameter that indicates if the policy is global so that we don't need to check if the notification path starts with `GLOBAL_POLICIES_ROOT` or `LOCAL_POLICIES_ROOT`.

Could you double check the review comments if they are generated from LLM?

shibd (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
```diff
@@ -0,0 +1,282 @@
...
+        });
+    }
+
+    private static TopicPolicies updatePolicies(Optional<TopicPolicies> currentPolicies,
+                                                boolean isGlobalPolicy,
+                                                Consumer<TopicPolicies> policyUpdater) {
+        TopicPolicies policies = currentPolicies.map(TopicPolicies::clone).orElseGet(TopicPolicies::new);
+        policies.setIsGlobal(isGlobalPolicy);
+        policyUpdater.accept(policies);
+        policies.setIsGlobal(isGlobalPolicy);
```
nodece:
```suggestion
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
```diff
@@ -0,0 +1,282 @@
...
+        if (topicListeners == null) {
+            return;
+        }
+        for (TopicPolicyListener listener : topicListeners) {
+            try {
+                listener.onUpdate(policies == null ? null : policies.clone());
+            } catch (Throwable error) {
+                log.error().attr("topic", topicName).exception(error).log("Call topic policy listener error");
+            }
+        }
```
nodece, line 237: Using `computeIfPresent` here is slightly safer for concurrent access since it avoids the race window between `get()` and iteration.

```suggestion
        listeners.computeIfPresent(topicName, (__, topicListeners) -> {
            for (TopicPolicyListener listener : topicListeners) {
                try {
                    listener.onUpdate(policies == null ? null : policies.clone());
                } catch (Throwable error) {
                    log.error().attr("topic", topicName).exception(error).log("Call topic policy listener error");
                }
            }
            return topicListeners;
        });
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
```diff
@@ -0,0 +1,282 @@
...
+        TopicPolicies cloned = policies.clone();
+        cloned.setIsGlobal(isGlobalPolicy);
+        return cloned;
+    }
+
+    @VisibleForTesting
+    public CompletableFuture<Optional<TopicPolicies>> getTopicPoliciesDirectFromStore(TopicName topicName,
+                                                                                      boolean isGlobal) {
+        TopicName partitionedTopicName = normalizeTopicName(topicName);
+        String path = pathFor(partitionedTopicName, isGlobal);
```
nodece: `pathFor` has been handled for the `topicName`.

```suggestion
        String path = pathFor(topicName, isGlobal);
```

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
```diff
@@ -0,0 +1,282 @@
...
+            return CompletableFuture.completedFuture(Optional.empty());
+        }
+        boolean global = type == GetType.GLOBAL_ONLY;
+        return cache(global).get(pathFor(partitionedTopicName, global))
+                .thenApply(policies -> policies.map(policy -> cloneWithScope(policy, global)));
+    }
+
+    @Override
+    public boolean registerListener(TopicName topicName, TopicPolicyListener listener) {
+        listeners.compute(normalizeTopicName(topicName), (__, topicListeners) -> {
```
nodece, line 140: Please add the close check.

nodece (changes requested):

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
BewareMyPower: Is there really an issue? The `SystemTopicBasedTopicPoliciesService` does not check close state as well. This method is only called when initializing a `PersistentTopic`, while the topic policies service is closed during broker shutdown.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
BewareMyPower: It will call `onUpdate` with the node lock held for the concurrent hash map, see the document of `ConcurrentHashMap#computeIfPresent`:

```
     * this method if the key is present, else not at all.  Some
     * attempted update operations on this map by other threads may be
     * blocked while computation is in progress, so the computation
     * should be short and simple.
```

This method is called by the future callback of `MetadataCache#get`, which can be executed in the current thread of metadata event thread. `topicListeners` is null or a `CopyOnWriteArrayList`, it's safe to iterate over it.

The race might look like:
1. A topic is created but the initialization is not done
2. `notifyListeners` is called for a modification on topic policies
3. `registerListener` is called, but the notification is missed.

Is there a real issue? If yes, we should modify the topic creation logic to disallow topic policies modification before a topic initialization is done.

The design is similar to `SystemTopicBasedTopicPoliciesService`, so I think it should not be a blocker. It's just a rare case that should not cause serious issues.

nodece: I think the more realistic race is during `close()` rather than concurrent modification of the listener list itself.

For example:

1. `MetadataStoreTopicPoliciesService` is closing
2. A topic is still asynchronously initializing
3. `registerListener()` is invoked after the service is logically closed
4. The listener remains registered forever and may never be cleaned up

This can potentially lead to a resource leak.

I agree `CopyOnWriteArrayList` iteration itself is safe, and `computeIfPresent` may unnecessarily hold the CHM node lock while invoking callbacks.

But from a design perspective, Pulsar has many async initialization flows, so it is still important to carefully handle lifecycle races like `close()` vs late listener registration.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java
nodece: I agree this is probably fine in most normal shutdown flows.

My concern is more about lifecycle completeness in async scenarios rather than an immediate functional bug.

In most cases this may not cause visible issues, but under repeated startup/shutdown, test environments, or delayed async callbacks, it could leave stale listeners or create subtle lifecycle inconsistencies.

nodece (approved):
LGTM

BewareMyPower:
@nodece Let me merge it first because this topic is out of the scope of the topic policies service itself. I agree that your comments [1] and [2] make sense, but it's more like an issue that should be resolved by structured concurrency, which needs careful design.

It would be better add a common abstraction for listeners and skip all of them after closing the topic policies service.

[1] https://github.com/apache/pulsar/pull/25707#discussion_r3258091432
[2] https://github.com/apache/pulsar/pull/25707#discussion_r3258114750

BewareMyPower:
I tried to write a test to reproduce and then found that even if a listener is registered due to the race, there still might be a real issue.

```java
    private static final CountDownLatch realCloseLatch = new CountDownLatch(1);
    private static final CountDownLatch closeLatch = new CountDownLatch(1);

    public static class MockTopicPoliciesService extends MetadataStoreTopicPoliciesService {

        public void close() {
            super.close();
            realCloseLatch.countDown();
            try {
                if (!closeLatch.await(10, TimeUnit.SECONDS)) {
                    log.info().log("Failed to close TopicPoliciesService within the timeout");
                }
                log.info().log("Interrupted when closing TopicPoliciesService");
            }
        }
    }
```

The test can be easily written based on the mocked implementation above.

`MetadataStoreTopicPoliciesService#close` is a very simple operation that changes `closed` to `true`. Even if a different broker receives a topic policies update request and updates the policies, `handleNotification` will skip calling `notifyListeners` in https://github.com/apache/pulsar/blob/8652efa4d60b791a5f1ee4e52f7ffda6ebbbb256/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/MetadataStoreTopicPoliciesService.java#L195-L200

Therefore, basically these zombie listeners won't be notified.

In another case, if the CAS of the `closed` flag happened before listeners were notified, where asynchronous operations had been started, these listeners would be cleared and would not be notified anymore.

PR #25710, opened 2026-05-07
[improve][metadata] Add Option-set API to MetadataStore

merlimat:
## Motivation

The metadata-store API exposes per-operation flags through several disjoint mechanisms that have grown over time:
- `EnumSet<CreateOption>` for `Ephemeral` / `Sequential`
- a separate `Map<String, String> secondaryIndexes` parameter for index hints
- nothing today for routing hints (PartitionKey) needed by sharded backends like Oxia

Each new flag has been added as an explicit method argument or a new overload, which does not scale. PIP-471 (metadata-driven transactions) needs PartitionKey routing on most read/write methods, which would otherwise require yet another argument across the surface.

## Modifications

Introduce a single options bag based on a sealed `Option` interface and make `Set<Option>` the canonical form across the metadata API.

**API surface** (`MetadataStore`, `MetadataStoreExtended`, `MetadataCache`):
- `Option` sealed interface with subtypes `Ephemeral`, `Sequential`, `SecondaryIndex(name, key)`, `PartitionKey(key)`
- `OptionsHelper` for extracting typed values (`isEphemeral`, `isSequential`, `secondaryIndexes`, `partitionKey`)
- `Set<Option>` form is the **abstract canonical** method on every read/write entry point (`get`, `getChildren`, `getChildrenFromStore`, `exists`, `put`, `delete`, `findByIndex`)
- The legacy no-opts and `EnumSet<CreateOption>` / `Map<String,String>` overloads are now defaults that translate **into** the canonical `Set<Option>` form — old callers continue to work unchanged

**Implementation wiring:**
- `AbstractMetadataStore.storePut` hook signature flipped to `(path, data, version, Set<Option>)` — single method, the EnumSet+Map overload is gone
- `LocalMemoryMetadataStore`, `RocksdbMetadataStore`, `OxiaMetadataStore` (two overloads collapsed into one), and `OpPut` / `AbstractBatchedMetadataStore` (used by `ZKMetadataStore`) all read options via `OptionsHelper`
- Wrappers (`DualMetadataStore`, `FaultInjectionMetadataStore`, `DualMetadataCache`) override the canonical `Set<Option>` methods; legacy callers route through the interface default forwarders

`MetadataEvent` (sync replication payload) still carries `Set<CreateOption>` — `AbstractMetadataStore` bridges to/from `Set<Option>` at the event handler boundary; `SecondaryIndex`/`PartitionKey` are not propagated through the legacy event payload, which matches existing behavior.

## Verifying this change

This change is already covered by existing tests since it preserves behavior of all current call sites — only the canonical override point moved.

- [x] The public API (`MetadataStore` / `MetadataStoreExtended` / `MetadataCache`) — strictly additive: new abstract `Set<Option>` methods are added; legacy forms become defaults that forward, so existing callers and external implementations are unaffected.

PR in forked repository: https://github.com/merlimat/pulsar/pull/new/mmerli/metadata-options


lhotari (approved):
LGTM

PR #25712, opened 2026-05-07
[cleanup] PIP-468: Remove deprecated registerConsumer overloads

merlimat:
## Summary

`ScalableTopicController` and `ScalableTopicService` each carried a `registerConsumer` overload that defaulted `ScalableConsumerType` to `STREAM` "for backward compatibility". The only production caller (`ServerCnx`) already passes the explicit type, so the overloads served only stale test code and were leaving deprecation warnings on every broker test compile.

Removed both overloads. Updated their only callers — broker tests (`ScalableTopicControllerTest`, `ScalableTopicServiceTest`) — to pass `ScalableConsumerType.STREAM` explicitly. No production behavior change.

## Test plan

- [x] `pulsar-broker` compile + checkstyle clean.
- [x] `ScalableTopicControllerTest` 31/31 pass.
- [x] `ScalableTopicServiceTest` 16/16 pass.
- [x] No more `[deprecation] registerConsumer` warnings in the broker test compile output.


lhotari (approved):
LGTM

PR #25709, opened 2026-05-07
[fix][broker] include segment descriptor in underlying topic name

Dream95:
### Motivation

The underlying persistent topic name for a scalable topic segment should include the segment descriptor, for example `persistent://tenant/ns/my-topic/0000-ffff-0`.

### Modifications

Include the segment descriptor when converting a scalable segment topic name to its underlying `persistent://` topic name.
The generated name now preserves the segment suffix, changing from `persistent://tenant/ns/my-topic` to `persistent://tenant/ns/my-topic/0000-ffff-0`.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

### Matching PR in forked repository

PR in forked repository: https://github.com/Dream95/pulsar/pull/8

PR #25717, opened 2026-05-07
[feat] PIP-468: segment-aware admin endpoints for cursor lifecycle

merlimat:
## Summary

Supersedes #25709, which had the right intent but produced an invalid `persistent://t/n/parent/{descriptor}` URL that `TopicName` parses as a legacy V1 topic name and rejects synchronously — making the admin call hard-fail before any HTTP request goes out.

`createSubscription` / `deleteSubscription` on a scalable topic was silently broken before either change. `ScalableTopicController.createSubscriptionOnSegment` built a `persistent://t/n/parent` URL (no descriptor) and handed it to the generic `topics().createSubscriptionAsync` admin API. That URL addresses the parent scalable topic, which has no managed-ledger backing, so the admin call hit `Subscription Busy` / `NotFound` and the `.exceptionally` block swallowed it. Per-segment cursors were never pre-created. Lazy cursor creation by the V5 client on first consumer subscribe masked the bug in the happy path; admin `createSubscription` had no observable effect beyond writing the `SubscriptionMetadata` to the metadata store. The same shape of bug existed in `ScalableTopics.deleteSegmentTopics` (whole-scalable-topic teardown).

The clean fix is segment-specific admin endpoints, all the way down — no `persistent://` conversion.

### New REST endpoints (`Segments.java`, super-user only, segment-domain only)
- `PUT /segments/{tenant}/{namespace}/{topic}/{descriptor}/subscription/{subscription}` — create a cursor at earliest position
- `DELETE /segments/{tenant}/{namespace}/{topic}/{descriptor}/subscription/{subscription}` — delete the cursor

Same `validateSuperUserAccessAsync` → `validateTopicOwnershipAsync` → `getOrCreateTopic` (or `getTopicIfExists` for delete) pattern as the existing `/segments/.../subscription/{sub}/backlog`, `/seek`, `/skip-all`.

### New admin client methods (`ScalableTopics`)
- `createSegmentSubscriptionAsync(segmentTopic, subscription)`
- `deleteSegmentSubscriptionAsync(segmentTopic, subscription)`

Both take a `segment://...` URI directly. `ScalableTopicsImpl` builds the REST path from `TopicName` parts — same pattern as the existing `getSegmentSubscriptionBacklogAsync` / `seekSegmentSubscriptionAsync` / `clearSegmentSubscriptionBacklogAsync` helpers.

### Caller updates
- `ScalableTopicController.createSubscriptionOnSegment` / `deleteSubscriptionOnSegment` now call `scalableTopics().createSegmentSubscriptionAsync` / `deleteSegmentSubscriptionAsync`. Removed the broken `toSegmentUnderlyingPersistentName` helper.
- `ScalableTopics.deleteSegmentTopics` (whole-scalable-topic teardown) now uses `admin.scalableTopics().deleteSegmentAsync(segmentTopicName, force)` — the existing segment-aware admin endpoint already used for split-segment teardown — instead of fabricating a `persistent://` URL.

## Test plan

- [x] `V5ScalableSubscriptionAdminTest` (new, 1 test) — behavioral round trip: create scalable topic → `admin.scalableTopics().createSubscription` → produce 30 messages with no consumer attached → subscribe and assert all 30 are received. **Fails on master**, **passes here**.
- [x] `ScalableTopicControllerTest` 31/31 — updated mocks: `topics.createSubscriptionAsync` / `deleteSubscriptionAsync` → `scalableTopics.createSegmentSubscriptionAsync` / `deleteSegmentSubscriptionAsync`. Verifies the controller calls the new admin path.
- [x] `ScalableTopicServiceTest` 16/16 — unchanged, no regressions.
- [x] Full V5 broker test suite green (153/153).
- [x] `pulsar-client-admin-api`, `pulsar-client-admin-original`, `pulsar-broker` checkstyle clean.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
@@ -57,7 +57,6 @@ import org.apache.pulsar.common.naming.TopicDomain;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.policies.data.NamespaceOperation;
 import org.apache.pulsar.common.policies.data.TopicOperation;
-import org.apache.pulsar.common.scalable.SegmentInfo;
 import org.apache.pulsar.common.scalable.SegmentTopicName;
 import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.metadata.api.MetadataStoreException;
@@ -675,8 +674,9 @@ public class ScalableTopics extends AdminResource {
     }
 
     /**
-     * Best-effort delete underlying persistent topics for all segments.
-     * Uses the internal admin client which handles cross-broker routing.
+     * Best-effort delete the underlying topic for every segment in the DAG. Uses the
+     * segment-aware admin endpoint, which routes to the segment-owning broker via the
+     * standard bundle-ownership lookup.
      */
     private CompletableFuture<Void> deleteSegmentTopics(TopicName parentTopic,
                                                          ScalableTopicMetadata metadata,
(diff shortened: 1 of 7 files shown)


lhotari (approved):
LGTM

PR #25719, opened 2026-05-07
[fix][broker] PIP-468: skip deleted segments in checkpoint consumer

merlimat:
## Summary

Segment topics must only be created by the controller's explicit-create path — never auto-created on client connect — so a producer or consumer racing a controller-driven delete (post-prune retention GC, force-delete) can't silently re-create the topic and mask the deletion. Plus, when a checkpoint consumer does observe a missing segment topic, it must silently skip it instead of failing.

- `BrokerService.isAllowAutoTopicCreationAsync` now returns `false` for `TopicDomain.segment`.
- `ScalableTopics.createScalableTopic` materializes the initial segment(s) up front via the admin client (so multi-broker bundle ownership is honored) instead of relying on auto-create-on-publish.
- `Segments.createSegment` switched to `getTopic(name, true)` for explicit creation now that the auto-create policy forbids segments.
- `Segments.deleteSegment` actually propagates `?force=true` to `deleteForcefully()` (was being silently dropped).
- `ScalableTopicController.initialize`, after winning leadership, idempotently re-creates any active-segment backing topics that are missing — recovering from a `createScalableTopic` that committed metadata but failed to materialize all initial segments, or an out-of-band force-delete of an active segment. Sealed segments are intentionally not healed: a missing sealed topic means data was retention-pruned, and the V5 checkpoint consumer skips it.
- `ScalableCheckpointConsumer` swallows `TopicDoesNotExistException` / `NotFoundException` at reader-create and during the read loop, dropping the segment from `segmentReaders` + `lastReceivedPositions` so the consumer keeps delivering from segments that still exist.

## Test plan

- [x] New `V5CheckpointConsumerSkipDeletedSegmentTest`: 2-segment topic, force-delete one, open consumer → must subscribe to survivor and deliver its full backlog. Verified the new `Segment backing topic deleted (retention expired); skipping` log path actually fires.
- [x] New `ScalableTopicControllerTest.testInitializeRecreatesMissingActiveSegments` + `testInitializeDoesNotCreateSegmentsWhenNotLeader` cover the recovery path.
- [x] Existing `ScalableTopicControllerTest.testSplitSegment` / `testMergeSegments` updated for the new init-time `createSegmentAsync` calls (clearInvocations).
- [x] All `org.apache.pulsar.client.api.v5.*` and `org.apache.pulsar.broker.service.scalable.*` tests pass.
- [x] Checkstyle clean.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
@@ -174,9 +174,8 @@ public class ScalableTopics extends AdminResource {
                     Map<String, String> props = properties != null ? properties : Map.of();
                     ScalableTopicMetadata metadata = ScalableTopicController.createInitialMetadata(
                             numInitialSegments, props);
-                    // Segment persistent topics are auto-created on demand when clients connect,
-                    // so we only need to store the metadata here.
-                    return resources().createScalableTopicAsync(tn, metadata);
+                    return resources().createScalableTopicAsync(tn, metadata)
+                            .thenCompose(ignored -> createInitialSegmentTopicsAsync(tn, metadata));
                 })
                 .thenAccept(__ -> {
                     log.info().attr("clientAppId", clientAppId()).attr("topic", tn)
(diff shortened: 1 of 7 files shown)


lhotari (approved):
LGTM

PR #25723, opened 2026-05-08
[improve] Wire Set<Option> through backend hooks + Oxia PartitionKey

merlimat:
## Motivation

Follow-up to #25710 (the additive Set<Option> API). That PR landed the public-API surface and the canonical `Set<Option>` form on every read/write method, but the backend-side hooks on `AbstractMetadataStore` (`storeGet`, `storeDelete`, `getChildrenFromStore`, `existsFromStore`, `storeScanChildren`, `storeFindByIndex`) still took only path/version — opts were dropped at the `AbstractMetadataStore` boundary. That meant `Option.PartitionKey`, the routing hint sharded backends like Oxia need, never reached the client.

This PR plumbs `Set<Option>` end-to-end and wires `OxiaMetadataStore` to consume `Option.PartitionKey` natively.

## Modifications

**Backend hooks**: `AbstractMetadataStore.storeGet`, `storeDelete`, `getChildrenFromStore`, `existsFromStore`, `storeScanChildren`, and `storeFindByIndex` all take `Set<Option>`. Backends override the new signatures; opts ignored where not used (LocalMemory, RocksDB).

**ZK batching layer**: `OpGet`, `OpDelete`, `OpGetChildren` each carry the opts so the batch flusher has them available; `AbstractBatchedMetadataStore` plumbs opts into the Op constructors.

**Oxia**: `OxiaMetadataStore` extracts `PartitionKey` via `OptionsHelper.partitionKey` and attaches the matching client option (`GetOption.PartitionKey` / `PutOption.PartitionKey` / `DeleteOption.PartitionKey` / `ListOption.PartitionKey` / `RangeScanOption.PartitionKey`) to every Oxia call. The two `storePut` overloads collapse into one.

**API completeness**: `MetadataStore.deleteRecursive` and `deleteIfExists` got `Set<Option>` overloads to round out the surface. `AbstractMetadataStore.deleteRecursive` threads opts through every recursive child + the final `deleteIfExists` so PartitionKey routing applies to the whole subtree.

**Wrappers** (`DualMetadataStore`, `FaultInjectionMetadataStore`): updated to the new signatures.

## Verifying this change

New `OxiaPartitionKeyTest` runs against a 3-shard Oxia cluster and verifies put/get/delete/getChildren round-trips with `Option.PartitionKey`.

The test deliberately asserts only the round-trip — asserting that a record routed to shard(K) is unreachable from shard(J) is statistically flaky on 3 shards because half the (K,J) pairs collide. A stronger correctness assertion would need access to Oxia internals.

- [x] The public `MetadataStore` API — strictly additive: new abstract `Set<Option>` overloads added for `deleteRecursive` and `deleteIfExists`, legacy no-opts forms become defaults that forward. Existing callers and external implementations are unaffected.
- [x] The backend-side hooks on `AbstractMetadataStore` — these are protected, used only by the in-tree backends, and all in-tree backends are updated in this PR.

PR in forked repository: https://github.com/merlimat/pulsar/pull/new/mmerli/metadata-partition-key

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStore.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/api/MetadataStore.java
@@ -198,8 +198,8 @@ public interface MetadataStore extends AutoCloseable {
         return delete(path, expectedVersion, Set.of());
     }
 
-    default CompletableFuture<Void> deleteIfExists(String path, Optional<Long> expectedVersion) {
-        return delete(path, expectedVersion)
+    default CompletableFuture<Void> deleteIfExists(String path, Optional<Long> expectedVersion, Set<Option> opts) {
+        return delete(path, expectedVersion, opts)
                 .exceptionally(e -> {
                     if (e.getCause() instanceof NotFoundException) {
                         LOG.info().attr("path", path).log("Path not found while deleting (this is not a problem)");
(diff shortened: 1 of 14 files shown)


lhotari (approved):
LGTM

PR #25727, opened 2026-05-09
[fix][broker] URL-encode sub-name in Txn pending-ack topic 

shibd:
### Motivation

When a subscription name contains `/` (e.g. a Pulsar Function's fully-qualified name like `tenant/namespace/function`), opening the transaction pending-ack store fails with:

**Root cause:** `MLPendingAckStore.getTransactionPendingAckStoreSuffix()` appends the raw subscription name to the topic name string. When the result is passed to `TopicName.get()`, the extra `/` characters produce four path segments, which `TopicName` misidentifies as a deprecated V1 (cluster-scoped) name and throws synchronously.

### Changes

- **`MLPendingAckStore.getTransactionPendingAckStoreSuffix()`**: URL-encode the subscription name with `Codec.encode()` before appending it. `TopicName` always `Codec.decode()`s the local-name on parse and re-encodes it on output (`getPersistenceNamingEncoding()`), so there is no double-encoding and the managed-ledger path is correct.
- **`PersistentTopic.unsubscribe()`**: Remove the pre-emptive `Codec.encode()` call — encoding is now done centrally in `getTransactionPendingAckStoreSuffix()`.
- **Tests**: Two new assertions — one verifying slash-containing subscription names work correctly, one as a regression guard for normal names.

### Verifying this change
All tests passed

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
@@ -58,6 +58,7 @@ import org.apache.pulsar.common.naming.TopicDomain;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.policies.data.NamespaceOperation;
 import org.apache.pulsar.common.policies.data.TopicOperation;
+import org.apache.pulsar.common.scalable.SegmentInfo;
 import org.apache.pulsar.common.scalable.SegmentTopicName;
 import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.metadata.api.MetadataStoreException;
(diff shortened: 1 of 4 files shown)


poorbarcode (approved):

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
```diff
@@ -58,6 +58,7 @@
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.policies.data.NamespaceOperation;
 import org.apache.pulsar.common.policies.data.TopicOperation;
+import org.apache.pulsar.common.scalable.SegmentInfo;
```
BewareMyPower, line 61:
```suggestion
```

shibd: This was caused because the previous PR did not import this package. Otherwise, the build would not have compiled successfully.
https://github.com/apache/pulsar/pull/25719

lhotari:
does 4.0.x need this fix too?

pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
BewareMyPower: What confused me is that the previous PR passed the CI validation. Is it caused by a concurrent commits?

PR #25754, opened 2026-05-12
[improve][broker] PIP-473: transaction metadata data layer

merlimat:
## Summary

Foundation data layer for [PIP-473](https://github.com/apache/pulsar/issues/...) (metadata-driven transactions). Introduces typed records + path/index constants + a `TxnMetadataStore` façade over `MetadataStore`, ready for the v5 TC to plug into in a follow-up.

- New records under `pulsar-broker/.../transaction/metadata/`:
  - `TxnHeader` — header at `/txn/<txnId>`, the CAS linearization point for `endTxn`
  - `TxnOp` — op-log entry written under `/txn-op/<txnId>` with `SequenceKeysDeltas`
  - `TxnEvent` — per-segment / per-(segment, sub) notification event
  - `Versioned<T>` — `(value, version)` pair returned from reads so callers can CAS
  - `TxnState`, `TxnOpKind` — Jackson-friendly enums (serialized by name)
- `TxnPaths` — path templates, index names, key formatters
- `TxnMetadataStore` façade — header CRUD with CAS, op append with kind-specific secondary indexes, deadline / final-state range queries (via `scanByIndex`), and segment / subscription event publish + subscribe (via `subscribeSequence`)
- Bundled metadata-layer fix: the `__seq_counter__` sidecar that backs `SequenceKeysDeltas` on non-native stores was leaking through `scanChildren` and the `scanByIndex` fallback. Filter it out in `AbstractMetadataStore`'s defaults plus the Memory and RocksDB native overrides. Oxia uses native sequence keys (no sidecar) and is unaffected.

No runtime path consumes this yet — the v5 TC PR will wire it up.

## Test plan

- [x] `pulsar-broker:test --tests TxnMetadataStoreTest` — 7 tests on Memory: header CAS lifecycle, write/ack op append + index scan, deadline range scan (filters terminal + out-of-range), final-state range scan, segment + subscription event publish + subscribe
- [x] `pulsar-metadata:test --tests MetadataStoreScanChildrenTest` — added `skipsSequenceCounterSidecars` across ZooKeeper / Memory / RocksDB / Oxia / MockZooKeeper
- [x] `pulsar-metadata:test --tests SequenceKeysTest` + `MetadataStoreSecondaryIndexTest` — pre-existing suites still green after the abstract-layer change
- [x] Checkstyle clean on `pulsar-broker` + `pulsar-metadata`

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnMetadataStore.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnMetadataStore.java
@@ -117,7 +117,8 @@ public class TxnMetadataStore {
      */
     public CompletableFuture<Stat> appendOp(String txnId, TxnOp op) {
         Option.SecondaryIndex idx = switch (op.getKind()) {
-            case WRITE -> new Option.SecondaryIndex(TxnPaths.IDX_WRITES_BY_SEGMENT, op.getSegment());
+            case WRITE -> new Option.SecondaryIndex(TxnPaths.IDX_WRITES_BY_SEGMENT,
+                    TxnPaths.segmentKey(op.getSegment()));
             case ACK -> new Option.SecondaryIndex(TxnPaths.IDX_ACKS_BY_SEGMENT_SUBSCRIPTION,
                     TxnPaths.ackIndexKey(op.getSegment(), op.getSubscription()));
         };
@@ -129,8 +130,9 @@ public class TxnMetadataStore {
 
     /** Stream all write ops targeting {@code segment}. */
     public CompletableFuture<Void> listWritesBySegment(String segment, ScanConsumer consumer) {
+        String key = TxnPaths.segmentKey(segment);
         return store.scanByIndex(TxnPaths.TXN_OP_PREFIX, TxnPaths.IDX_WRITES_BY_SEGMENT,
-                segment, segment,
+                key, key,
                 gr -> {
                     TxnOp op = fromJson(gr.getValue(), TxnOp.class);
                     return op.getKind() == TxnOpKind.WRITE && segment.equals(op.getSegment());
@@ -207,7 +209,7 @@ public class TxnMetadataStore {
 
     /** Append a per-segment notification event. */
     public CompletableFuture<Stat> publishSegmentEvent(String segment, TxnEvent event) {
-        Set<Option> opts = Set.of(new Option.PartitionKey(segment), APPEND_DELTAS);
+        Set<Option> opts = Set.of(new Option.PartitionKey(TxnPaths.segmentKey(segment)), APPEND_DELTAS);
         return store.put(TxnPaths.segmentEventsParent(segment), toJson(event), Optional.empty(), opts);
     }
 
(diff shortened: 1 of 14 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnMetadataStore.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Stream terminal transactions in {@code state} whose finalization time falls in
+     * {@code [fromMsInclusive, toMsInclusive]}. Pass {@code null} on either bound for unbounded.
+     */
+    public CompletableFuture<Void> listFinalizedByStateAndTimeRange(TxnState state,
+                                                                    Long fromMsInclusive, Long toMsInclusive,
+                                                                    ScanConsumer consumer) {
+        String from = fromMsInclusive == null ? state.name() + ":" : TxnPaths.finalStateKey(state, fromMsInclusive);
+        String to = toMsInclusive == null
+                ? state.name() + ":" + "9".repeat(TxnPaths.LONG_KEY_WIDTH)
```
lhotari: use constant?

lhotari:
Claude Code review findings to check:
```
  1. [INTENT MISMATCH] Raw segment URIs are embedded into MetadataStore paths without encoding — TxnPaths.segmentEventsParent / subscriptionEventsParent in
  pulsar-broker/.../transaction/metadata/TxnPaths.java:90-102
    - PIP-473 segment names take the form segment://tenant/ns/topic/<hexStart>-<hexEnd>-<segmentId> (see existing comment at MLPendingAckStore.java:534). Concatenated into a metadata path,
  that produces e.g. /txn-segment-events/segment://tenant/ns/topic/... — which contains // (empty path component) and :.
    - ZooKeeper's PathUtils.validatePath rejects this; Oxia, while not slash-validating, would interpret the embedded / as path levels and break the sidecar / sequence-key semantics.
    - The existing convention in transaction code (MLPendingAckStore.java:533, TopicTransactionBuffer.java:868) is Codec.encode(...). This data layer breaks that convention without saying so.
    - Recommended: Either encode segment (and subscription) in TxnPaths helpers, or add a @throws/precondition comment documenting that callers must pre-encode. Currently the tests use
  "segment://A" only because Memory happens to tolerate it — that's a misleading green light for the v5 TC PR.
  2. [BUG] isSequenceCounterChild filter uses a permissive endsWith predicate — pulsar-metadata/.../AbstractMetadataStore.java:608-612, 651-655, 776-782
    - A user record whose final path segment ends with the literal __seq_counter__ would be silently filtered from scanChildren / scanByIndex. Low likelihood in practice, but easy to harden.
    - Since sequenceCounterPath(prefix) = prefix + "__seq_counter__" is always a sibling of the prefix at the same parent level, the sidecar name is always <prefix-leaf>__seq_counter__. A
  stricter check would require the sidecar to be co-located with a known prefix, or at minimum scope by an internal marker char not allowed in user keys.
    - Acceptable as-is if you're confident no user key will collide; worth a one-line comment justifying the looseness.
```

lhotari (approved):
LGTM, just check the local Claude Code review findings I added a comment

PR #25766, opened 2026-05-13
[BUG] Fix stats in PulsarZooKeeperClient multi.

zhaizhibo:
### Motivation

In `PulsarZooKeeperClient.multi()`, the stats were incorrectly using `createStats` instead of `multiStats`.

### Modifications

Changed `createStats` to `multiStats` in the `multi()` method.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
@@ -100,7 +100,9 @@ public class ZKMetadataStore extends AbstractBatchedMetadataStore
             this.rootPath = new ConnectStringParser(zkConnectString).getChrootPath();
 
             isZkManaged = true;
-            StatsLogger statsLogger = metadataStoreConfig.getStatsProvider().getStatsLogger("ZKMetadataStore");
+            StatsLogger statsLogger = metadataStoreConfig.getStatsProvider()
+                    .getStatsLogger("ZkMetadataStore")
+                    .scopeLabel("storeName", metadataStoreConfig.getMetadataStoreName());
             zkc = PulsarZooKeeperClient.newBuilder().connectString(zkConnectString)
                     .connectRetryPolicy(new BoundExponentialBackoffRetryPolicy(100, 60_000, Integer.MAX_VALUE))
                     .allowReadOnlyMode(metadataStoreConfig.isAllowReadOnlyOperations())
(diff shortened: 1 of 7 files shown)


merlimat (commented):
Looks good. please add a test to verify we're passing the right stats object.

zhaizhibo:
@merlimat Thank you for your review. The test has been added. Additionally, I noticed an issue where the broker zk client’s statsLogger was not set. I have added a configuration (bookkeeperClientExposeStatsToPrometheus) to expose the broker zk client’s metrics, with the default value set to false.

dao-jun (approved):

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
```diff
@@ -30,6 +30,7 @@
 import java.util.stream.Collectors;
 import lombok.CustomLog;
 import lombok.SneakyThrows;
+import org.apache.bookkeeper.stats.StatsLogger;
```
merlimat, line 33: That's BookKeeper stats provider. We shouldn't be using it here

merlimat (changes requested):

pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/ZKMetadataStore.java
zhaizhibo: @merlimat But we had used `org.apache.bookkeeper.stats.OpStatsLogger` and `org.apache.bookkeeper.stats.StatsLogger` in `PulsarZooKeeperClient`, and the builder of PulsarZooKeeperClient is also `org.apache.bookkeeper.stats.StatsLogger`. Should we also replace all references to this type?

void-ptr974:
I think we should separate the bug fix from the metrics exposure work.

For this PR, the minimal and correct fix is to change async multi() from createStats to multiStats and add a focused test verifying that async multi records into multiStats rather than createStats.

The later changes introduce a new metadata-store metrics exposure path by passing a BookKeeper StatsProvider through MetadataStoreConfig. That seems like a separate design discussion, and it also expands the dependency on the BookKeeper stats provider in the metadata-store layer, which has already been pointed out in review.

If we want to expose metadata-store/ZK operation metrics to Prometheus, I think we should do it in a follow-up PR with a Pulsar-owned metrics abstraction, or have broker-level Prometheus code export metadata-store operation metrics directly. The metric should probably be modeled as operation-level metrics, e.g. operation=multi, provider=zk, store=local/configuration, instead of encoding everything into the metric name.

Also, the test should not only assert that zk_multi metric names exist. The metric may exist because the OpStatsLogger was created during initialization. It should assert that async multi actually increments multiStats and does not increment createStats.

Issue #25097, opened 2025-12-19
[Bug] Geo-replication for individual topics stops and does not recover in various scenarios


tarmacmonsterg:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Read release policy

- [x] I understand that [unsupported versions](https://pulsar.apache.org/contribute/release-policy/#supported-versions) don't get bug fixes. I will attempt to reproduce the issue on a supported version of Pulsar client and Pulsar broker.

### User environment

Broker: apachepulsar/pulsar-all: 4.0.7

Helm chart: https://github.com/apache/pulsar-helm-chart/tree/pulsar-4.0.1

### Issue Description

Replication for some topics randomly stops during normal operation, causing backlog to accumulate.
The issue is observed in two main cases:
	1.	A sudden spike in the publish rate to a topic (for example, a steady rate of 5 messages per second followed by a burst of 5,000 messages per second lasting about 5 minutes).
	2.	External infrastructure issues, such as frequent broker restarts or resource-related problems (for example, high iowait on BookKeeper).

The issue occurs at the topic level. For example, we have both partitioned and non-partitioned topics. The problem is observed with non-partitioned topics, and there were also cases where replication got stuck for a single partition of a partitioned topic.

The only way to restore replication is to disable replication for the namespace and then re-enable it.

### Error messages

```text
No any error or warn messages
```

### Reproducing the issue

Can't reproduce using standalone cluster

### Additional information

https://apache-pulsar.slack.com/archives/C078TGY9R29/p1764167528705519

2 Clusters:
* 5 Bookies
* 4 Brokers
* 3 Proxies

Replication: 2/2/2

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
@tarmacmonsterg Which Pulsar clients do you use to produce messages to the topics? Just wondering if you might be facing the issue fixed by #25106.

tarmacmonsterg:
Hello @lhotari .
I with some news about the problem.
We already updated till 4.0.9 and in some scenarious, we had this problem.

The main thing I’ve noticed is that the issue starts immediately after the error:

`Too many requests to the same Bookie while reading Lxxxxxxx Exxxxxxx from bookie: xxxxxxx`

After that, replication for the topic stops.
We observe this issue on both partitioned and non-partitioned topics.

I enabled debug logging, but there’s nothing particularly unusual.
[...]

PR #25767, opened 2026-05-13
[improve][broker] Prevent stale replicator pending reads after termination

Denovo1998:
Fixes #xyz

Main Issue: #xyz

PIP: #xyz 

### Motivation

This is a follow-up to #25625 for the replication read-failure path related to #25097.

#25625 completes a replicator `InFlightTask` when a managed-ledger read fails, so retryable read failures do not leave stale pending-read state behind. This PR improves the state handling for the race where the read failure callback arrives after the replicator has already left `Started`, for example during termination or producer restart. In that case, `readEntriesFailed` should still clear the failed read task so the replicator does not keep stale pending-read state while the object is still alive.

### Modifications

- Complete failed `InFlightTask` contexts before checking whether the replicator is still in the `Started` state.
- Keep the cleanup defensive by only completing contexts that are actually `InFlightTask` instances.
- Add error logs for unexpected failed-read callback states:
  - the callback context is not an `InFlightTask`;
  - the `InFlightTask` already has non-null entries when the failed callback is handled.
- Remove duplicated failed-task completion from the later retry/error branches in `readEntriesFailed`.
- Add a regression test that starts a real replication read, blocks the managed-ledger read failure, terminates the replicator through the normal lifecycle, releases the failure callback, and verifies the pending-read state is cleared.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
@@ -532,11 +532,20 @@ public abstract class PersistentReplicator extends AbstractReplicator
     }
 
     private void completeFailedReadTask(Object ctx) {
-        if (ctx instanceof InFlightTask) {
-            InFlightTask inFlightTask = (InFlightTask) ctx;
-            if (inFlightTask.entries == null) {
-                inFlightTask.setEntries(Collections.emptyList());
-            }
+        if (!(ctx instanceof InFlightTask)) {
+            log.error()
+                    .attr("ctx", ctx)
+                    .log("Unexpected read entries failed context");
+            return;
+        }
+
+        InFlightTask inFlightTask = (InFlightTask) ctx;
+        if (inFlightTask.entries == null) {
+            inFlightTask.setEntries(Collections.emptyList());
+        } else {
+            log.error()
+                    .attr("inFlightTask", inFlightTask)
+                    .log("Unexpected completed in-flight task in read entries failed callback");
         }
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
@@ -494,7 +494,7 @@ public abstract class PersistentReplicator extends AbstractReplicator
 
     @Override
     public void readEntriesFailed(ManagedLedgerException exception, Object ctx) {
-        InFlightTask inFlightTask = (InFlightTask) ctx;
+        completeFailedReadTask(ctx);
         if (state != Started) {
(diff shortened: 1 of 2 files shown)


nodece (approved):

lhotari (approved):
LGTM

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -533,6 +531,15 @@ public void readEntriesFailed(ManagedLedgerException exception, Object ctx) {
         brokerService.executor().schedule(this::readMoreEntries, waitTimeMillis, TimeUnit.MILLISECONDS);
     }
 
+    private void completeFailedReadTask(Object ctx) {
+        if (ctx instanceof InFlightTask) {
```
poorbarcode: If you think the object ctx will not be an instance of `InflightTask` in the future, please add an error log for the logic case, which can quickly expose incorrect modifications.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -533,6 +531,15 @@ public void readEntriesFailed(ManagedLedgerException exception, Object ctx) {
...
+            InFlightTask inFlightTask = (InFlightTask) ctx;
+            if (inFlightTask.entries == null) {
```
poorbarcode: If the value of `inFlightTask.entries` is not null, there must be a bug; you should add an error log in the `else` logic branch

poorbarcode:
Replicator termination only occurs in the two events of "disabling replication" and closing the topic, and these two events will trigger the JVM GC for the replicator object. So this PR only fixes the state of "pending reads" at the moment when "the topic is turned off, and the replicator object has not been GC", and the state "pending reads" at this moment will not cause any problems.

However, this modification is still useful, which makes our project more perfect. I suggest changing the label in the title to "improve". Then in the future, if users discover a replication problem, they won't spend too much time on considering such case

Denovo1998:
> Replicator termination only occurs in the two events of "disabling replication" and closing the topic, and these two events will trigger the JVM GC for the replicator object. So this PR only fixes the state of "pending reads" at the moment when "the topic is turned off, and the replicator object has not been GC", and the state "pending reads" at this moment will not cause any problems.
> 
> However, this modification is still useful, which makes our project more perfect. I suggest changing the label in the title to "improve". Then in the future, if users discover a replication problem, they won't spend too much time on considering such case

@poorbarcode Okay. PTAL again.

poorbarcode (approved):
LGTM

PR #25770, opened 2026-05-14
[fix][proxy] Close channel on connection failure

nodece:
### Motivation

When establishing a direct proxy connection fails, the outbound channel may remain open until it is cleaned up asynchronously. In some cases, this can leave stale connections or unreleased resources behind, especially during rapid reconnects or connection timeouts.

This change ensures the outbound channel is explicitly closed before closing the inbound channel when connection establishment fails, improving resource cleanup consistency and reducing the risk of leaked or half-open channels.

### Modifications

* Explicitly closes the outbound channel (`f.channel()`) when the connection attempt fails before closing the inbound channel.


lhotari (approved):
LGTM

PR #25772, opened 2026-05-14
[improve][broker] PIP-473: MetadataPendingAckStore for segment subscriptions

merlimat:
## Summary

P4 of the [PIP-473](https://github.com/apache/pulsar/blob/master/pip/pip-473.md) metadata-driven transactions stack. Adds the ack-side complement to [#25768](https://github.com/apache/pulsar/pull/25768) — a `PendingAckStore` implementation for subscriptions on `segment://` topics that reads truth from the metadata-store layout.

- `MetadataPendingAckStore`:
  - `appendIndividualAck` / `appendCumulativeAck` write `TxnOp(ACK, segment, sub, ledgerId, entryId, cumulative)` records via the `TxnMetadataStore` façade.
  - `appendCommitMark` / `appendAbortMark` are **no-ops** — in v5 the TC owns the lifecycle, the store consumes its events.
  - Subscription-event listener (`/txn-subscription-events/<seg>:<sub>-*`) uses the same header-truthing reconcile pattern as the P3 TB: re-read headers for every currently-open txn this subscription is involved in, then for terminals call `PendingAckHandleImpl.commitTxn` / `abortTxn` (no SPI change) and delete the corresponding `/txn-op` ack records.
  - Recovery via scan (Option C): subscribe → scan `listAcksBySegmentSubscription` → group by txnId → fetch headers → seed open-txn set → terminal txns discovered mid-scan are processed inline → drain post-recovery reconcile.
- `MetadataPendingAckStoreProvider` + `DispatchingTransactionPendingAckStoreProvider` (routes by `TopicName.isSegment()`, falls through to legacy `MLPendingAckStore` otherwise).

### Supporting changes

- `TxnOp` gains a nullable `Boolean cumulative` field (only set on cumulative acks).
- `TxnMetadataStore.deleteAckOpsForSegmentSubscriptionAndTxn` mirrors the P3 write-op cleanup helper; the two flavors share a private `scanAndDeleteOpsForTxn`.

### Not in this PR (deferred to P5.4)

- **No default-config flip.** `transactionPendingAckStoreProviderClassName` stays `MLPendingAckStoreProvider`. The new store is opt-in until the v5 TC is wired up.
- **MLPendingAckStore segment-workaround stays in place.** Removing it now would break segment subscriptions while the legacy store is still the default. P5.4 will remove it as part of the same PR that flips all three providers.

## Test plan

- [x] `pulsar-broker:test --tests MetadataPendingAckStoreTest` — 6 cases on the metadata store (Memory backend, mocked `PersistentSubscription` + `PendingAckHandleImpl`): individual ack append, cumulative ack append, no-op marks, commit event drives `handle.commitTxn` + ack-op cleanup, abort event drives `handle.abortTxn` + cleanup, recovery rebuilds state + processes terminal txns discovered mid-scan.
- [x] `pulsar-broker:test --tests DispatchingTransactionPendingAckStoreProviderTest` — 2 routing tests.
- [x] P2 + P3 unit tests still green after the `TxnOp.cumulative` addition.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
@@ -152,7 +152,10 @@ public class MetadataTransactionBuffer implements TransactionBuffer {
 
             @Override
             public void onError(Throwable throwable) {
-                // Future propagation handled by listWritesBySegment's returned future.
+                // Recovery still fails loudly via the scan's returned future and the terminal
+                // whenComplete below; logging here captures the cause with segment context.
+                log.warn().attr("segment", segmentName).exception(throwable)
+                        .log("TB recovery scan errored");
             }
 
             @Override
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
@@ -261,7 +261,7 @@ public class MetadataTransactionBuffer implements TransactionBuffer {
 
     private CompletableFuture<Position> recordOp(TxnID txnId, String txnIdKey, Position position) {
         TxnOp op = new TxnOp(TxnOpKind.WRITE, segmentName, null,
-                position.getLedgerId(), position.getEntryId());
+                position.getLedgerId(), position.getEntryId(), null);
         return txnStore.appendOp(txnIdKey, op).thenApply(stat -> {
             synchronized (lock) {
                 TxnEntry entry = txns.get(txnIdKey);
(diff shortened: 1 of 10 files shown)


lhotari (approved):
LGTM

Please check this local Claude Code review finding:

> [BUG] Silent onError in the recovery scan loses txns without surfacing the failure — MetadataPendingAckStore.java:264-285 (the new ScanConsumer() block inside replayAsync).  onError does nothing. If the scanByIndex underneath listAcksBySegmentSubscription emits a per-entry error (e.g. a corrupt JSON record, or a transient store error during streaming), the scan  still completes the outer CompletableFuture as if it had succeeded. The store will then proceed with an incomplete txnIdKeys set: it will neither track those open txns nor clean up their
 lingering op records, and recoveryFuture will complete successfully. There is no way for the operator to know recovery was partial. The same pattern in
TxnMetadataStore.scanAndDeleteOpsForTxn (preserved from P3) at least only causes a deletion miss — here it silently breaks the store's correctness invariant. At minimum log the error; ideally propagate it so recovery fails loud.

PR #25773, opened 2026-05-14
[improve][fn] make built-in connector reload incremental

Dream95:
### Motivation

`reloadBuiltInSources` / `reloadBuiltInSinks` currently reload connectors by rebuilding the full connector map and closing all existing connector instances.  
This causes unnecessary classloader churn and can invalidate previously referenced connector instances even when the archive content has not changed.

### Modifications

- Added incremental reload logic in `ConnectorUtils.reloadConnectors(...)`:
  - Reuse existing `Connector` instances when both archive path and archive MD5 are unchanged.
  - Create new `Connector` instances only for added/changed archives.
  - Return evicted connectors (removed/replaced) so callers can close only those.
- Updated `ConnectorsManager.reloadConnectors(...)` to use the incremental reload result and close only evicted connectors.
- Extended `Connector` to keep archive MD5 (`archiveMd5Hex`) for identity checks during reload.
- Extracted MD5 utility methods into `pulsar-common` `FileUtils` and reused them from both connector reload and `NarUnpacker`.
- Added tests:
  - `ConnectorUtilsReloadTest`
    - unchanged archive is reused and remains usable
    - changed archive is replaced and old connector is closed
    - removed archive is evicted and old connector is closed
  - `ConnectorsManagerReloadConnectorsTest`
    - manager-level reload reuses unchanged connector instance

### Verifying this change

- Added unit tests in `pulsar-functions/utils/src/test/java/org/apache/pulsar/functions/utils/io/ConnectorUtilsReloadTest.java`
- Added runtime-level test in `pulsar-functions/runtime/src/test/java/org/apache/pulsar/functions/worker/ConnectorsManagerReloadConnectorsTest.java`
- CI should cover compilation and execution of the new tests.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

PR in forked repository: https://github.com/Dream95/pulsar/pull/9

### DIFF
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/nar/FileUtils.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/nar/FileUtils.java
@@ -32,7 +32,6 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HexFormat;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import lombok.CustomLog;
@@ -56,7 +55,7 @@ public class FileUtils {
      * @throws IOException
      *             if cannot read file
      */
-    protected static byte[] calculateMd5sum(final File file) throws IOException {
+    public static byte[] calculateMd5sum(final File file) throws IOException {
         try (final FileInputStream inputStream = new FileInputStream(file)) {
             // codeql[java/weak-cryptographic-algorithm] - md5 is sufficient for this use case
             final MessageDigest md5 = MessageDigest.getInstance("md5");
@@ -73,17 +72,6 @@ public class FileUtils {
         } catch (NoSuchAlgorithmException nsae) {
             throw new IllegalArgumentException(nsae);
         }
-    }
-
-    /**
-     * Calculates the MD5 digest of the specified file as hexadecimal.
-     *
-     * @param file file to digest
-     * @return MD5 hex string
-     * @throws IOException if the file cannot be read
-     */
-    public static String calculateMd5Hex(File file) throws IOException {
-        return HexFormat.of().formatHex(calculateMd5sum(file));
     }
 
     public static void ensureDirectoryExistAndCanReadAndWrite(final File dir) throws IOException {
--- a/pulsar-common/src/main/java/org/apache/pulsar/common/nar/FileUtils.java
+++ b/pulsar-common/src/main/java/org/apache/pulsar/common/nar/FileUtils.java
@@ -25,10 +25,14 @@
 package org.apache.pulsar.common.nar;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HexFormat;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import lombok.CustomLog;
@@ -42,6 +46,45 @@ import lombok.CustomLog;
 public class FileUtils {
 
     public static final long MILLIS_BETWEEN_ATTEMPTS = 50L;
+
+    /**
+     * Calculates an md5 sum of the specified file.
+     *
+     * @param file
+     *            to calculate the md5sum of
+     * @return the md5sum bytes
+     * @throws IOException
+     *             if cannot read file
+     */
+    protected static byte[] calculateMd5sum(final File file) throws IOException {
+        try (final FileInputStream inputStream = new FileInputStream(file)) {
+            // codeql[java/weak-cryptographic-algorithm] - md5 is sufficient for this use case
+            final MessageDigest md5 = MessageDigest.getInstance("md5");
+
+            final byte[] buffer = new byte[1024];
+            int read = inputStream.read(buffer);
+
+            while (read > -1) {
+                md5.update(buffer, 0, read);
+                read = inputStream.read(buffer);
+            }
+
+            return md5.digest();
+        } catch (NoSuchAlgorithmException nsae) {
+            throw new IllegalArgumentException(nsae);
+        }
+    }
+
+    /**
+     * Calculates the MD5 digest of the specified file as hexadecimal.
+     *
+     * @param file file to digest
+     * @return MD5 hex string
+     * @throws IOException if the file cannot be read
+     */
+    public static String calculateMd5Hex(File file) throws IOException {
+        return HexFormat.of().formatHex(calculateMd5sum(file));
+    }
 
     public static void ensureDirectoryExistAndCanReadAndWrite(final File dir) throws IOException {
         if (dir.exists() && !dir.isDirectory()) {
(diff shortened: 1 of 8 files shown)


pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/worker/ConnectorsManager.java
```diff
@@ -89,24 +91,34 @@ public Path getSinkArchive(String sinkType) {
     }
 
     public void reloadConnectors(WorkerConfig workerConfig) throws IOException {
-        TreeMap<String, Connector> oldConnectors = connectors;
-        this.connectors = createConnectors(workerConfig);
-        closeConnectors(oldConnectors);
+        boolean enableClassloading = workerConfig.getEnableClassloadingOfBuiltinFiles()
+                || ThreadRuntimeFactory.class.getName().equals(workerConfig.getFunctionRuntimeFactoryClassName());
+        Pair<TreeMap<String, Connector>, List<Connector>> reload = ConnectorUtils.reloadConnectors(
```
lhotari: nit: since Java records are available, it's better to use a custom return type that's a Java record 
such as
```java
record ReloadConnectorsResult(Map<String, Connector> connectors, List<Connector> connectorsToClose) {}
```

lhotari: nit: since Java records are available, it's better to use a custom return type that's a Java record 
such as
```java
record ReloadConnectorsResult(Map<String, Connector> connectors, List<Connector> connectorsToClose) {}
```

pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/io/ConnectorUtils.java
```diff
@@ -186,4 +196,65 @@ public static TreeMap<String, Connector> searchForConnectors(String connectorsDi
...
+     * the caller must {@link Connector#close()} each (typically via {@code ConnectorsManager}).
+     *
+     * @param previous                 connectors from the previous scan (may be empty, never null)
+     * @param connectorsDirectory      same semantics as {@link #searchForConnectors}
+     * @param narExtractionDirectory   same semantics as {@link #searchForConnectors}
+     * @param enableClassloading       same semantics as {@link #searchForConnectors}
+     * @return left: new map keyed by connector name (reused values are identical instances from {@code previous});
+     *         right: connectors the caller should close
+     */
+    public static Pair<TreeMap<String, Connector>, List<Connector>> reloadConnectors(
```
lhotari: nit: since Java records are available, it's better to use a custom return type that's a Java record 
such as
```java
record ReloadConnectorsResult(Map<String, Connector> connectors, List<Connector> connectorsToClose) {}
```

pulsar-common/src/main/java/org/apache/pulsar/common/nar/FileUtils.java
```diff
@@ -43,6 +47,45 @@ public class FileUtils {
...
+     * Calculates the MD5 digest of the specified file as hexadecimal.
+     *
+     * @param file file to digest
+     * @return MD5 hex string
+     * @throws IOException if the file cannot be read
+     */
+    public static String calculateMd5Hex(File file) throws IOException {
+        return HexFormat.of().formatHex(calculateMd5sum(file));
+    }
+
```
lhotari: Move these to ConnectorUtils instead. It's better to keep unnecessary utilities out of pulsar-commons.

lhotari (commented):
good work. some minor comments

lhotari (commented):
good work. some minor comments

dao-jun (approved):
ConnectorsManager.connectors is volatile, and reloadConnectors assigns a new map atomically. The closeConnectors is called on only the evicted set. Between the assignment this.connectors = reload.getLeft() and the close of evicted connectors, there's a brief window where a caller could be using an evicted connector.
  However, this was already the case before the change (old behavior also swapped the map then closed the old one), so this PR doesn't regress thread safety. Worth noting but not a new concern.

lhotari (approved):
LGTM.

The existing code (before this PR) unnecessarily uses TreeMap as parameter, field, or variable types. A general recommendation is to use the minimal Map interface type, usually one of `java.util.Map`, `java.util.SequencedMap`, or `java.util.SortedMap`. In the case of connectors, the use the TreeMap is justified as the implementation to get stable iteration order by key. This is available when `java.util.Map` is the parameter, field, or variable type.

I'd suggest addressing the refactoring to get rid of unnecessary use of TreeMap in parameter, field, and variable types for connector related classes after this current PR has been merged. For example,

In ConnectorsManager
```java
    @Getter
    private volatile Map<String, Connector> connectors;
```

The implementation instance would remain as `TreeMap`, so this change would only be about the types for fields, variables and parameters.


```java
public record ReloadConnectorsResult(Map<String, Connector> connectors, List<Connector> connectorsToClose) {
}
```

This change could also be made part of this PR since it already closes touches the code where `TreeMap` is used and also adds a new `TreeMap` parameter/field in ReloadConnectorsResult.

lhotari:
@Dream95 There's #25782 about the follow up related to TreeMap types. Would you like to take it?

pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/worker/ConnectorsManager.java
```diff
@@ -89,24 +91,34 @@ public Path getSinkArchive(String sinkType) {
     }
 
     public void reloadConnectors(WorkerConfig workerConfig) throws IOException {
-        TreeMap<String, Connector> oldConnectors = connectors;
-        this.connectors = createConnectors(workerConfig);
-        closeConnectors(oldConnectors);
+        boolean enableClassloading = workerConfig.getEnableClassloadingOfBuiltinFiles()
+                || ThreadRuntimeFactory.class.getName().equals(workerConfig.getFunctionRuntimeFactoryClassName());
```
lhotari: This could be extracted into a private method to avoid code duplication since the same logic is also in createConnectors.

pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/io/ConnectorUtils.java
```diff
@@ -186,4 +199,66 @@ public static TreeMap<String, Connector> searchForConnectors(String connectorsDi
...
+                    if (prev != null
+                            && prev.getArchivePath() != null
+                            && archive.equals(prev.getArchivePath())
+                            && md5Hex.equals(prev.getArchiveMd5Hex())) {
+                        next.put(name, prev);
+                    } else {
+                        if (prev != null) {
+                            toClose.add(prev);
+                        }
+                        next.put(name, new Connector(archive, cntDef, narExtractionDirectory, enableClassloading,
```
lhotari, line 255: add logging here about finding a changed connector

Dream95:
@lhotari  Yes, I'll take  separate PR. Besides the connector changes here, FunctionsManager has the same unnecessary TreeMap usage; I'll address both together.

Dream95:
Since reload is admin-driven and usually serialized,  If we need stricter ordering, I’m happy to add synchronization in a follow-up.

lhotari (approved):
LGTM

PR #25774, opened 2026-05-14
[feat][pip] PIP-477: Dynamic Bookie Cluster Switching for Brokers

wolfstudy:
### Motivation

Pulsar brokers are statically bound to a single Bookie cluster via
`bookkeeperMetadataServiceUri` at startup. When the underlying Bookie cluster
hits capacity limits or non-recoverable failures, there is no runtime way to
re-target brokers to another Bookie cluster without restart and full data
migration. This PIP introduces zero-downtime Bookie cluster switching by
attributing each ledger to its owning cluster and routing reads accordingly.


pip/pip-470.md
```diff
@@ -0,0 +1,587 @@
...
+```protobuf
+message ManagedLedgerInfo {
+  message LedgerInfo {
+    required int64 ledgerId = 1;
+    optional int64 entries = 2;
+    optional int64 size = 3;
+    optional int64 timestamp = 4;
+    optional OffloadContext offloadContext = 5;
+    optional string bookieClusterName = 6;   // ★ new (Topic Data Ledger attribution)
+    repeated KeyValue properties = 7;        // ★ new (extension slot)
```
dao-jun: we already have the field `properties` in PIP-404. And maybe we can also use the field to store `bookieClusterName`

pip/pip-470.md
```diff
@@ -0,0 +1,587 @@
+# PIP-470: Dynamic Bookie Cluster Switching for Brokers
```
lhotari:
```suggestion
# PIP-477: Dynamic Bookie Cluster Switching for Brokers
```

lhotari: PIP-470 is taken. PIP-477 is currently available. When the discussion thread is started on dev mailing list, that's how to reserve a PIP number.

lhotari (changes requested):
Thanks @wolfstudy for the proposal.

**My main concern is architectural rather than technical: I'd like this to build on the storage-layer pluggability introduced by [PIP-384](https://github.com/apache/pulsar/blob/master/pip/pip-384.md) rather than adding this complexity to core Pulsar.**

PIP-384 deliberately decoupled `ManagedLedgerStorage` and introduced `ManagedLedgerStorageClass` so that the BookKeeper client is owned by a storage class, and a broker can hold multiple storage classes — each topic resolving to one of them via persistence policy. The framing PIP-384 establishes is "this topic lives on this storage backend." The core essence of PIP-477 — moving writes to a new BK cluster while preserving readability of historical data on the old one — fits naturally as an extension of that framing: **a topic can move between storage classes, with per-ledger attribution determining which storage class serves each historical read**.

Concretely, I think the path forward looks like this:

1. **Implement PIP-477 first as a custom set of implementations of the existing interfaces, outside of core Pulsar.** A custom `ManagedLedgerStorage` can hold a `Map<clusterName, BookkeeperManagedLedgerStorageClass>` and expose the active one — this covers what `DynamicBookKeeperClientFactory` does. The orchestrator state machine, the Broker-ZK registry under `/admin/bookie-clusters/*`, the watcher, and the CLI/REST surface can live in the plugin module. The REST endpoints under `/admin/v2/bookie-clusters/*` can be exposed via [`AdditionalServlet`](https://github.com/apache/pulsar/blob/master/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/web/plugin/servlet/AdditionalServlet.java) without touching core REST classes.

2. **The per-ledger attribution slot already exists in core.** [PIP-404](https://github.com/apache/pulsar/blob/master/pip/pip-404.md) added `repeated KeyValue properties = 6` to `LedgerInfo` and exposed `ManagedLedger#asyncAddLedgerProperty` / `asyncRemoveLedgerProperty` / `asyncGetLedgerProperty` precisely as a generic extension slot for plugins. The `★ new` `repeated KeyValue properties = 7` proposed in §"proto Changes" of PIP-477 is duplicating what's already there — `bookieClusterName` can be encoded as a property in the existing `LedgerInfo.properties` slot today, with no core proto change. `ManagedCursorInfo` similarly already has `cursorProperties` (a `repeated StringProperty` at tag 8) which can carry the same attribution.

3. **The CLI surface is also already pluggable.** [PIP-201](https://github.com/apache/pulsar/blob/master/pip/pip-201.md) added the `CustomCommandFactory` / `CustomCommandGroup` / `CustomCommand` SPI (in `pulsar-client-tools-api`), bundled as a `.nar`, so `pulsar-admin bookie-clusters …` can be shipped entirely in the plugin with no `pulsar-admin` core change.

4. **Where there are real SPI gaps, add narrow hooks — don't inline the feature.** A custom `ManagedLedgerFactory` returned by the storage class already gets you most of the way; the remainder should be small, focused additions, narrower than PIP-477 and useful to anyone writing a custom storage class.

5. **If the plugin proves valuable to other users of Pulsar later, it can be moved into the Pulsar tree** as a separate module shipped alongside the broker, opt-in via configuration. The ~99% of users who never use this feature pay zero cost in core surface area, the orchestrator code can evolve on its own cadence.

This approach also provides a forcing function for a cleaner abstraction. PIP-384 currently assumes "one topic, one storage class for its lifetime." Generalizing to "one topic, one storage class at a time, with attribution-driven routing for previously-written ledgers" is a natural and reusable extension — useful well beyond the BK-cluster-switching scenario (e.g., per-topic storage-tier migration, gradual rollout of an alternative `ManagedLedgerStorage` implementation).

Would you be open to reframing the PIP along these lines?

Happy to discuss further on the dev list or here.

wolfstudy:
> Thanks @wolfstudy for the proposal.
> 
> **My main concern is architectural rather than technical: I'd like this to build on the storage-layer pluggability introduced by [PIP-384](https://github.com/apache/pulsar/blob/master/pip/pip-384.md) rather than adding this complexity to core Pulsar.**
> 
> PIP-384 deliberately decoupled `ManagedLedgerStorage` and introduced `ManagedLedgerStorageClass` so that the BookKeeper client is owned by a storage class, and a broker can hold multiple storage classes — each topic resolving to one of them via persistence policy. The framing PIP-384 establishes is "this topic lives on this storage backend." The core essence of PIP-477 — moving writes to a new BK cluster while preserving readability of historical data on the old one — fits naturally as an extension of that framing: **a topic can move between storage classes, with per-ledger attribution determining which storage class serves each historical read**.
> 
> Concretely, I think the path forward looks like this:
> 
> 1. **Implement PIP-477 first as a custom set of implementations of the existing interfaces, outside of core Pulsar.** A custom `ManagedLedgerStorage` can hold a `Map<clusterName, BookkeeperManagedLedgerStorageClass>` and expose the active one — this covers what `DynamicBookKeeperClientFactory` does. The orchestrator state machine, the Broker-ZK registry under `/admin/bookie-clusters/*`, the watcher, and the CLI/REST surface can live in the plugin module. The REST endpoints under `/admin/v2/bookie-clusters/*` can be exposed via [`AdditionalServlet`](https://github.com/apache/pulsar/blob/master/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/web/plugin/servlet/AdditionalServlet.java) without touching core REST classes.
> 2. **The per-ledger attribution slot already exists in core.** [PIP-404](https://github.com/apache/pulsar/blob/master/pip/pip-404.md) added `repeated KeyValue properties = 6` to `LedgerInfo` and exposed `ManagedLedger#asyncAddLedgerProperty` / `asyncRemoveLedgerProperty` / `asyncGetLedgerProperty` precisely as a generic extension slot for plugins. The `★ new` `repeated KeyValue properties = 7` proposed in §"proto Changes" of PIP-477 is duplicating what's already there — `bookieClusterName` can be encoded as a property in the existing `LedgerInfo.properties` slot today, with no core proto change. `ManagedCursorInfo` similarly already has `cursorProperties` (a `repeated StringProperty` at tag 8) which can carry the same attribution.
> 3. **The CLI surface is also already pluggable.** [PIP-201](https://github.com/apache/pulsar/blob/master/pip/pip-201.md) added the `CustomCommandFactory` / `CustomCommandGroup` / `CustomCommand` SPI (in `pulsar-client-tools-api`), bundled as a `.nar`, so `pulsar-admin bookie-clusters …` can be shipped entirely in the plugin with no `pulsar-admin` core change.
> 4. **Where there are real SPI gaps, add narrow hooks — don't inline the feature.** A custom `ManagedLedgerFactory` returned by the storage class already gets you most of the way; the remainder should be small, focused additions, narrower than PIP-477 and useful to anyone writing a custom storage class.
> 5. **If the plugin proves valuable to other users of Pulsar later, it can be moved into the Pulsar tree** as a separate module shipped alongside the broker, opt-in via configuration. The ~99% of users who never use this feature pay zero cost in core surface area, the orchestrator code can evolve on its own cadence.
> 
> This approach also provides a forcing function for a cleaner abstraction. PIP-384 currently assumes "one topic, one storage class for its lifetime." Generalizing to "one topic, one storage class at a time, with attribution-driven routing for previously-written ledgers" is a natural and reusable extension — useful well beyond the BK-cluster-switching scenario (e.g., per-topic storage-tier migration, gradual rollout of an alternative `ManagedLedgerStorage` implementation).
> 
> Would you be open to reframing the PIP along these lines?
> 
> Happy to discuss further on the dev list or here.

Thanks for the thorough review. I agree to reframe PIP-470 along the PIP-384 extension axis:

- **Drop the new proto fields**. Encode bookieClusterName into LedgerInfo.properties (already added by PIP-404) and ManagedCursorInfo.cursorProperties (already exists). The remaining open question is SchemaStorageFormat.PositionInfo, where no equivalent properties slot exists today — I'll either propose a tiny additive slot or fold attribution into SchemaLocator properties as a follow-up.

- **Ship as a plugin NAR module** (pulsar-bookie-cluster-switching) that provides a custom ManagedLedgerStorage holding a Map<clusterName, BookkeeperManagedLedgerStorageClass>. Active cluster selection, ZK registry under /admin/bookie-clusters/*, watcher, and orchestrator (BUILD/PROMOTE/CLEANUP) all live in the plugin.

- **REST surface** moves to AdditionalServlet; CLI moves to PIP-201 CustomCommandFactory NAR.

- **Narrow core hooks** I still need (these would be useful to any custom storage class, not just BK switching):

   - ManagedLedgerConfig.activeBookKeeperSupplier — per-call active BK resolution on the write path.
   - BookkeeperSchemaStorage.schemaReadBookKeeperResolver — analogous read-path resolution for schema ledgers.
   - An owner-broker-callable hook for cursor CAS under the ManagedCursorImpl lock (or expose it as an internal endpoint via AdditionalServlet).
- **Opt-in via configuration**; the ~99% of users that don't switch BK clusters pay zero cost in core.

I'll update the PIP draft accordingly and post the revised version on the dev list.

wolfstudy:
@lhotari Fixed, PTAL again, thanks

wolfstudy:
@lhotari @merlimat PTAL again, thanks

pip/pip-477.md
```diff
@@ -0,0 +1,991 @@
...
+`RoutingManagedLedgerFactory` wraps the base factory. When `ManagedLedger`
+asks for a new ledger:
+
+```java
+// Pseudocode of the stamping path
+String activeCluster = multiClusterStorage.getActiveClusterName();
+BookKeeper bk = multiClusterStorage.getBookKeeper(activeCluster);    // via S1
+LedgerHandle lh = bk.asyncCreateLedger(...);
+
+// In the same metadata CAS that records the new LedgerInfo:
```
void-ptr974: It looks like the intended direction is to write the attribution in the same metadata update that first records the new `LedgerInfo`. Could we make the required core change for that explicit?

Today `asyncAddLedgerProperty` is an after-the-fact update. To make this atomic, it seems `ManagedLedgerImpl` needs a creation-time way for the storage implementation to provide `LedgerInfo.properties` before calling `store.asyncUpdateLedgerIds(...)`. The cursor and schema paths seem to need the analogous behavior for `ManagedCursorInfo.cursorProperties` and `SchemaStorageFormat.PositionInfo.properties`.

Could the PIP clarify whether this is part of the proposed SPI changes, and how legacy ledgers without attribution are distinguished from newly-created ledgers that should always have attribution?

pip/pip-477.md
```diff
@@ -0,0 +1,991 @@
...
+        // Internally:
+        //   newBk.asyncCreateLedgerAdv(ledgerId, …)        ← SAME ledgerId
+        //   for entryId in [0, lastAddConfirmed]:
+        //     newLh.addEntry(entryId, srcEntry)            ← SAME entryId
+        //   newLh.close()
+```
+
+#### Cursor Ledger BUILD (centralized)
+
+Identical pattern over `/managed-ledgers/<topic>/<cursorName>` znodes, copying
```
void-ptr974, line 865: Could we discuss whether cursor metadata needs to be migrated by copying the existing cursor ledger with the same ledgerId?

If the topic owner can force a cursor metadata checkpoint/rollover after the active BK cluster changes, it could create a new cursor metadata ledger on the target BK, persist the current mark-delete position / deleted ranges / cursor properties, and CAS-update `ManagedCursorInfo.cursorsLedgerId`.

Would that be sufficient, or is there a reason the old cursor ledger needs to be byte-copied with `asyncCreateLedgerAdv`?

pip/pip-477.md
wolfstudy: Good idea, I will clarify this further in the PIP.

pip/pip-477.md
wolfstudy: > Could we discuss whether cursor metadata needs to be migrated by copying the existing cursor ledger with the same ledgerId? If the topic owner can force a cursor metadata checkpoint/rollover after the active BK cluster changes, ... would that be sufficient ... ?

Thanks — this is a sharp observation and you are right that, considered in isolation, a cursor metadata ledger is semantically just a checkpoint: only the latest cursorsLedgerId matters; older cursor ledgers carry no business value once superseded. So technically we could indeed handle cursors with a forced rollover instead of byte-copying with asyncCreateLedgerAdv.

We have considered this and chose to keep the byte-copy + ledgerId-preserving design for cursors. Our reasoning, for the record (and to make the trade-off auditable), is now captured explicitly in the PIP under **"Alternatives Considered → Alt-5"**. The short version:

1. Symmetry with Schema is load-bearing. Schema metadata ledgers are referenced directly by SchemaLocator.indexList[].position.ledgerId/entryId for every historical-version read; there is no equivalent "rollover and forget" path for schema. If we adopted rollover for cursors, BUILD/PROMOTE would be two divergent state machines (one mutating cursorsLedgerId, one immutable on position.ledgerId). With the unified design, both paths are "byte-copy with same ledgerId → CAS only the attribution property", and LedgerCopyUtil is one shared utility.

2. Rollback would no longer be O(1). Today's PROMOTE only flips _pulsar.bookieClusterName; ROLLBACK is the reverse flip — constant-time, no data movement (§"Phase 7"). Under rollover-based migration, cursorsLedgerId itself moves, so ROLLBACK either has to mint another fresh cursor ledger on the old cluster or resurrect a previously-deleted ledgerId. That's more code and more failure modes for an operation we want to keep boring.

3. I9 / I8 invariants weaken. The current design's "PROMOTE changes ONLY the attribution KV" is auditable: a single CAS, one field. Alt-5 turns PROMOTE into a multi-field rewrite of ManagedCursorInfo, which couples the migration path with the markDelete-flush path on owner brokers.

4. Owner-broker tail latency. Forced rollover must run under the ManagedCursorImpl lock, serialized with in-flight markDelete flushes and BK acks, before the CAS can proceed. Performing this concurrently across every cursor on every owner during the switch window has measurable impact on markDelete p99. Today's BUILD runs entirely on the orchestrator broker against dormant copies; owner brokers see only a sub-second PROMOTE RPC.

5. The byte-copy cost is bounded. Cursor metadata ledgers are KB–MB; copying them with bounded concurrency on the orchestrator is cheap and pays for itself by simplifying every other corner of the design.

So we are keeping the current cursor BUILD path. I've added Alt-5 to the PIP so this conversation is preserved and a future PIP can revisit the trade-off if priorities change.

void-ptr974: Thanks, the explanation makes sense to me. I agree that preserving the same ledgerId can make the promotion/rollback state machine much simpler, especially if the goal is to keep existing positions unchanged and only switch the cluster attribution.

I have two related questions that may be worth clarifying in the PIP.

First, if the design relies on copying ledgers with the same ledgerId into the target BK cluster, do we need to explicitly define the safety invariant around ledgerId collision? Since different BK clusters normally have independent ledgerId allocation spaces, the source ledgerId may already exist in the target cluster, or may need to be reserved/imported in a way that prevents future allocation conflicts. It would be helpful to document whether the target BK cluster is expected to be empty/dedicated, or whether the migration flow will explicitly detect/reserve/import these ledgerIds safely.

Second, I wonder whether preserving the same ledgerId is strictly required for all metadata ledgers, especially schema/cursor ledgers. For normal data ledgers, I understand the motivation to keep "(ledgerId, entryId)" stable. But for schema/cursor ledgers, the ledgerId seems more like an internal storage reference. If same-ledgerId import is hard to guarantee, could schema/cursor use a newly allocated target ledgerId and record the mapping in metadata instead?

For example, schema could potentially be modeled as "schema version -> source position / target position", with historical versions backfilled and new versions dual-written during the migration window. Cursor ledgers may also be able to use an explicit migrated target ledger id, as long as recovery/delete/cleanup paths consistently resolve it. This would be more complex than same-ledgerId copy, so I’m not suggesting it must replace the current design. I just wonder whether it is worth documenting as an alternative/trade-off, since it may make the design more applicable to non-empty target BK clusters.

I may be missing some constraints here, so I’d be happy to hear your thoughts.

PR #25792, opened 2026-05-15
[improve][broker] PIP-473: group all txn metadata paths under /txn/

merlimat:
## Summary

P2.5 — path-prefix cleanup for PIP-473 metadata. Moves the four transaction-metadata prefixes shipped in P2 ([#25754](https://github.com/apache/pulsar/pull/25754)) and used by P3 / P4 ([#25768](https://github.com/apache/pulsar/pull/25768), [#25772](https://github.com/apache/pulsar/pull/25772)) under a single `/txn/` root, grouped by purpose:

| Before | After |
|---|---|
| `/txn/<txnId>` | `/txn/id/<txnId>` |
| `/txn-op/...` | `/txn/op/...` |
| `/txn-segment-events/...` | `/txn/segment-events/...` |
| `/txn-subscription-events/...` | `/txn/subscription-events/...` |

Headers move to `/txn/id/<txnId>` so `getChildren("/txn")` returns the four purpose-buckets cleanly, not a mix of txnIds and sibling prefixes. The new `/txn/` root also leaves room for upcoming P3.5 additions like `/txn/segment-state/<segment>/...` (per the visibility-state design note) without further restructuring.

No behaviour change. The new TC and TB/PendingAckStore providers all remain disabled by default — same path layout is used by exactly the same code, just under different names.

## Test plan

- [x] All path construction already goes through `TxnPaths` helpers, so the rename is contained to the constants in `TxnPaths.java` plus the corresponding javadoc/comment text in the records and the participant code. Audit for hardcoded paths in `pulsar-broker/src/main` and `pulsar-broker/src/test` came up clean.
- [x] `pulsar-broker:test --tests TxnMetadataStoreTest` — 7 cases pass.
- [x] `pulsar-broker:test --tests MetadataTransactionBufferTest` — 7 cases pass.
- [x] `pulsar-broker:test --tests MetadataPendingAckStoreTest` — 6 cases pass.
- [x] `pulsar-broker:test --tests DispatchingTransactionBufferProviderTest` — 2 cases pass.
- [x] `pulsar-broker:test --tests DispatchingTransactionPendingAckStoreProviderTest` — 2 cases pass.
- [x] `pulsar-broker:test --tests TxnIdsTest` — 7 cases pass.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
@@ -65,8 +65,8 @@ import org.apache.pulsar.metadata.api.ScanConsumer;
  * <ul>
  *   <li><b>Publish</b> — {@link #appendBufferToTxn} reads the txn header (cache-first), appends the
  *       entry to the managed ledger, then appends a {@link TxnOp} record under
- *       {@code /txn-op/<txnId>-<seq>}. Both must succeed before we ack the publisher.</li>
- *   <li><b>State transitions</b> — driven by {@code /txn-segment-events/<segment>-*} sequence
+ *       {@code /txn/op/<txnId>-<seq>}. Both must succeed before we ack the publisher.</li>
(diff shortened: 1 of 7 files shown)


lhotari (approved):
LGTM

Issue #19061, opened 2022-12-26
Flaky-test: ResourceGroupUsageAggregationTest.testProduceConsumeUsageOnRG


codelipenghui:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234

### Exception stacktrace

```
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 5, Time elapsed: 8.132 s <<< FAILURE! - in org.apache.pulsar.broker.resourcegroup.ResourceGroupUsageAggregationTest
[630](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:631)
  Error:  testProduceConsumeUsageOnRG(org.apache.pulsar.broker.resourcegroup.ResourceGroupUsageAggregationTest)  Time elapsed: 0.185 s  <<< FAILURE!
[631](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:632)
  org.apache.pulsar.client.admin.PulsarAdminException: Resource group does not exist: runProduceConsume
[632](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:633)
  	at org.apache.pulsar.broker.resourcegroup.ResourceGroupService.checkResourceGroupExists(ResourceGroupService.java:417)
[633](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:634)
  	at org.apache.pulsar.broker.resourcegroup.ResourceGroupService.registerTenant(ResourceGroupService.java:196)
[634](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:635)
  	at org.apache.pulsar.broker.resourcegroup.ResourceGroupUsageAggregationTest.testProduceConsumeUsageOnRG(ResourceGroupUsageAggregationTest.java:140)
[635](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:636)
  	at org.apache.pulsar.broker.resourcegroup.ResourceGroupUsageAggregationTest.testProduceConsumeUsageOnRG(ResourceGroupUsageAggregationTest.java:85)
[636](https://github.com/apache/pulsar/actions/runs/3779970439/jobs/6425838234#step:10:637)
[...]

PR #25812, opened 2026-05-19
[improve][broker] Change bookkeeperClientSeparatedIoThreadsEnabled to True by default

zhanglistar:
### Motivation

Apache Pulsar PR #16333 introduced `bookkeeperClientSeparatedIoThreadsEnabled` so the BookKeeper client can use a dedicated IO thread pool instead of sharing the broker's Pulsar `EventLoopGroup`. The upstream default remained `false` for backward compatibility.

In our deployment, sharing IO threads between the broker and the BookKeeper client can cause contention under storage-heavy workloads. Enabling separated IO threads by default isolates BookKeeper network I/O from broker I/O and better matches our production expectations.

### Modifications

- Set `bookkeeperClientSeparatedIoThreadsEnabled=true` in `conf/broker.conf` and `conf/standalone.conf`, and update related comments.
- Change the Java default in `ServiceConfiguration` from `false` to `true`, and update `@FieldContext` documentation.
- Update unit tests to reflect the new default:
  - `ServiceConfigurationTest#testBookKeeperClientIoThreads`
  - `BookKeeperClientFactoryImplTest#testBookKeeperIoThreadsConfiguration`

Operators who want the previous behavior can set:

```properties
bookkeeperClientSeparatedIoThreadsEnabled=false
```

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1112,8 +1112,8 @@ bookkeeperClientNumWorkerThreads=
 bookkeeperClientNumIoThreads=
 
 # Use separated IO threads for BookKeeper client
-# Default is false, which will use Pulsar IO threads
-bookkeeperClientSeparatedIoThreadsEnabled=false
+# Default is true, which will use dedicated BookKeeper IO threads
+bookkeeperClientSeparatedIoThreadsEnabled=true
 
 # Speculative reads are initiated if a read request doesn't complete within a certain time
 # Using a value of 0, is disabling the speculative reads
--- a/conf/standalone.conf
+++ b/conf/standalone.conf
@@ -682,8 +682,8 @@ bookkeeperClientNumWorkerThreads=
 bookkeeperClientNumIoThreads=
 
 # Use separated IO threads for BookKeeper client
-# Default is false, which will use Pulsar IO threads
-bookkeeperClientSeparatedIoThreadsEnabled=false
+# Default is true, which will use dedicated BookKeeper IO threads
+bookkeeperClientSeparatedIoThreadsEnabled=true
 
 # Speculative reads are initiated if a read request doesn't complete within a certain time
 # Using a value of 0, is disabling the speculative reads
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -2201,9 +2201,10 @@ public class ServiceConfiguration implements PulsarConfiguration {
 
     @FieldContext(
             category = CATEGORY_STORAGE_BK,
-            doc = "Use separated IO threads for BookKeeper client. Default is false, which will use Pulsar IO threads"
+            doc = "Use separated IO threads for BookKeeper client. Default is true, which will use dedicated "
+                    + "BookKeeper IO threads"
     )
-    private boolean bookkeeperClientSeparatedIoThreadsEnabled = false;
+    private boolean bookkeeperClientSeparatedIoThreadsEnabled = true;
 
     /**** --- Managed Ledger. --- ****/
     @FieldContext(
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/BookKeeperClientFactoryImplTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/BookKeeperClientFactoryImplTest.java
@@ -312,11 +312,11 @@ public class BookKeeperClientFactoryImplTest {
         EventLoopGroup eventLoopGroup = mock(EventLoopGroup.class);
         BookKeeper.Builder builder = factory.getBookKeeperBuilder(conf, eventLoopGroup,
                 mock(StatsLogger.class), mock(ClientConfiguration.class));
-        assertEquals(FieldUtils.readField(builder, "eventLoopGroup", true), eventLoopGroup);
-        conf.setBookkeeperClientSeparatedIoThreadsEnabled(true);
+        assertNull(FieldUtils.readField(builder, "eventLoopGroup", true));
+        conf.setBookkeeperClientSeparatedIoThreadsEnabled(false);
         builder = factory.getBookKeeperBuilder(conf, eventLoopGroup,
                 mock(StatsLogger.class), mock(ClientConfiguration.class));
-        assertNull(FieldUtils.readField(builder, "eventLoopGroup", true));
+        assertEquals(FieldUtils.readField(builder, "eventLoopGroup", true), eventLoopGroup);
     }
 
     @Test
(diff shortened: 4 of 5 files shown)


zhanglistar:
https://github.com/apache/pulsar/issues/19061 Not related flaky test.

hangc0276 (approved):

lhotari (changes requested):
This shouldn't be required after #26599 and #26620 since they resolve the problems caused by contention (thread or lock contention itself won't disappear). Performance increases 26x when there's a constant load in a test case included in #26620. This change will ship in Pulsar 5.0.0.

PR 26620 improvements in scenario of 500 isolated V4 producer clients/connections, 20 isolated consumer clients on one Key_Shared subscription, shared client resources, random keys, unbatched 128-byte messages.
25x consume improvement for the time that all producers are producing
99.84% lower maximum backlog (11,3M -> 17,6k)

| Metric | Master baseline | PR | Improvement |
|---|---:|---:|---:|
| Steady ingress (produce) | 139,464 msg/s | 152,971 msg/s | +9.7% |
| Steady egress (consume) | 5,738 msg/s | 153,002 msg/s | +2566.6% |
| Whole-run producer throughput | 132,828 msg/s | 132,798 msg/s | -0.02% |
| Whole-run consumer throughput | 76,983 msg/s | 143,732 msg/s | +86.7% |
| Sampled maximum backlog | 11,366,885 messages | 17,693 messages | 99.84% lower |

lhotari:
Reopening this for more consideration. Since there's blocking happening on Bookkeeper IO threads, it will also Netty IO progress when Pulsar IO threads are blocked. I'll check that.

lhotari:
I benchmarked this change with the standalone performance launcher in [`tests/performance`](https://github.com/apache/pulsar/tree/769ce7f51c6f04bbd0b2043fd5bd54b16476df69/tests/performance), on current master (`769ce7f51c6f`) with `bookkeeperClientSeparatedIoThreadsEnabled` set explicitly (`false`, then `true`), so the only difference between variants is this setting. Runs alternated between the variants.

### Setup

Each run starts 1 broker, 3 bookies and the client containers with Testcontainers, on a single 16-core host. The broker, the bookies and the clients all compete for the same CPUs. The workload generator validates every run: complete delivery, per-key (per-device) ordering on every subscription, and duplicates. All 16 runs below passed, with no duplicates and no ordering violations.

**IoT high-rate** ([`iot-telemetry-high-rate.yaml`](https://github.com/apache/pulsar/blob/769ce7f51c6f04bbd0b2043fd5bd54b16476df69/tests/performance/scenarios/iot-telemetry-high-rate.yaml), which extends [`iot-telemetry.yaml`](https://github.com/apache/pulsar/blob/769ce7f51c6f04bbd0b2043fd5bd54b16476df69/tests/performance/scenarios/iot-telemetry.yaml); background in [`iot-telemetry.md`](https://github.com/apache/pulsar/blob/769ce7f51c6f04bbd0b2043fd5bd54b16476df69/tests/performance/iot-telemetry.md)):
- 500 preconnected gateway producers publish as fast as they can to one persistent topic.
- Messages are small keyed telemetry messages: 1M device keys, 64-byte payloads, batching enabled, broker deduplication enabled.
- Five applications each consume through their own 10-consumer Key_Shared subscription.
- 1M warmup messages, then 5M measured messages.
- Default ledger replication (E=2, W=2, A=2).

**Key_Shared 500×20** (the launcher version of [`key-shared-500x20.yaml`](https://github.com/apache/pulsar/blob/769ce7f51c6f04bbd0b2043fd5bd54b16476df69/tests/performance/scenarios/key-shared-500x20.yaml), which on master runs through the legacy pulsar-perf based profiling test):
- The same 500-producer, single-topic IoT workload.
- One 20-consumer Key_Shared subscription.
- Unbatched 128-byte messages.
- Single-copy ledgers (E=W=A=1), like the pulsar-perf version.
- 4M measured messages.

This launcher scenario isn't on master yet; it's [`iot-key-shared-500x20.yaml`](https://github.com/lhotari/pulsar/blob/lh-key-shared-500x20-scenario/tests/performance/scenarios/iot-key-shared-500x20.yaml).

Minor differences from running master as-is:
- My branch adds off-CPU profiling support to the launcher. That only matters for the profiled runs used to check thread pools.
- It uses the Wolfi (glibc) test image by default.
- The two settings were passed as broker config overrides (`brokerEnvs`) in local scenario variants.

### Results

"Delivered" is the measured messages divided by the time until the slowest subscription received its last message. It differs from producer throughput when consumers fall behind.

| Workload | Setting | Producer msg/s | Delivered msg/s | Consumers still draining after producers finished | E2E p99 (one subscription) | Publish p99 |
|---|---|---:|---:|---:|---:|---:|
| IoT high-rate | `false` (current default) | 70.7k, 79.0k, 62.2k | 50.4k, 37.1k, 44.9k | 28–72 s | 6.3–74 s | 2.0–2.9 s |
| | `true` (this PR; 32 BK IO threads = 2 × cores) | 57.4k, 63.4k, 59.9k | 57.1k, 63.0k, 59.4k | 0.5–0.8 s | 2.7–4.8 s | 2.1–3.8 s |
| | `true` + `bookkeeperClientNumIoThreads=8` | 64.3k, 64.3k | 63.7k, 63.9k | 0.5–0.8 s | 2.7–2.9 s | 2.0–2.1 s |
| Key_Shared 500×20 | `false` | 101.5k, 106.1k | 101.0k, 105.5k | 0.2 s | 1.6–1.8 s | 1.2–1.6 s |
| | `true` | 103.2k, 102.5k | 102.8k, 102.1k | 0.2 s | 1.6–1.7 s | 1.3 s |
| | `true` + 8 IO threads | 104.7k, 105.2k | 104.3k, 104.7k | 0.1–0.2 s | 1.7–1.8 s | 1.3 s |

- **IoT high-rate:** with shared IO threads, the producers outrun the five Key_Shared subscriptions and a backlog builds up. The consumers are still draining 28–72 s after the producers finish. With separated BookKeeper IO threads, publishing is a little slower but the consumers keep up. **Delivered throughput goes from about 44k to about 60k msg/s (+35 % on average), and to about 64k msg/s (+45 %) with 8 IO threads, which also gave the tightest tails and nearly identical runs. Judged by producer throughput alone, `true` would have looked like a 19 % regression.**
- **Key_Shared 500×20:** single subscription, single-copy ledgers. No measurable difference; all variants are within the run-to-run spread of about ±2 %.
- **Thread pools:** CPU profiles of a Key_Shared pair confirm the setting takes effect. With `true`, the `bookkeeper-io-*` threads take about 8 % of broker CPU samples, the `pulsar-io-*` share drops by about the same amount, and total broker CPU is about the same.

My guess is that bookie add and read responses no longer queue on the same event loops as the producer and consumer connections. I haven't profiled the IoT workload to confirm that.

### Caveats and a suggestion

These are two or three runs per variant on one shared host, so treat the numbers as indicative. The IoT difference is well outside the spread between runs; the Key_Shared one isn't.

Since `bookkeeperClientNumIoThreads` defaults to 2 × cores, enabling separated IO threads adds that many event loops next to the broker's own IO threads. On this host, 8 threads did at least as well as 32. It may be worth documenting sizing guidance for `bookkeeperClientNumIoThreads` together with the new default, or considering a smaller default when the IO threads are separated.

lhotari:
There are tradeoffs that this change makes, however I believe that this is a better default setting for Pulsar 5.0 since it will allow a more efficient broker cache usage and reduce bookkeeper reads as well when consumers can keep up with producers.

Issue #25815, opened 2026-05-19
Flaky-test: ResendRequestTest.testSharedSingleAckedPartitionedTopic


lhotari:
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

### Example failure

https://github.com/apache/pulsar/actions/runs/26080299303/job/76681740543?pr=25814#step:11:591

### Exception stacktrace

```text
Gradle suite > Gradle test > org.apache.pulsar.broker.service.ResendRequestTest > testSharedSingleAckedPartitionedTopic FAILED
      org.testng.internal.thread.ThreadTimeoutException: Method org.apache.pulsar.broker.service.ResendRequestTest.testSharedSingleAckedPartitionedTopic() didn't finish within the time-out 60000
          at java.base@21.0.11/jdk.internal.misc.Unsafe.park(Native Method)
          at java.base@21.0.11/java.util.concurrent.locks.LockSupport.park(LockSupport.java:371)
          at java.base@21.0.11/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionNode.block(AbstractQueuedSynchronizer.java:519)
          at java.base@21.0.11/java.util.concurrent.ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3780)
          at java.base@21.0.11/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3725)
          at java.base@21.0.11/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:1746)
          at app//org.apache.pulsar.common.util.collections.GrowableArrayBlockingQueue.take(GrowableArrayBlockingQueue.java:200)
          at app//org.apache.pulsar.client.impl.MultiTopicsConsumerImpl.internalReceive(MultiTopicsConsumerImpl.java:388)
          at app//org.apache.pulsar.client.impl.ConsumerBase.receive(ConsumerBase.java:282)
          at app//org.apache.pulsar.broker.service.ResendRequestTest.testSharedSingleAckedPartitionedTopic(ResendRequestTest.java:517)
```

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lhotari:
Buildscan: https://scans.gradle.com/s/mcgcdn6ekltme/tests/task/:pulsar-broker:test/details/org.apache.pulsar.broker.service.ResendRequestTest?top-execution=1

PR #25817, opened 2026-05-19
[fix][broker] Fix non-batched null-value messages not removed during topic compaction

grishaf:
### Motivation
When a non-Java producer (C++, Python, Go) sends a null-value message (tombstone) on a compacted topic, the key is not removed during topic compaction. The tombstone is retained in the compacted view instead of being deleted.
**Root cause:**
`AbstractTwoPhaseCompactor.extractKeyAndSize()` computed payload size using `headersAndPayload.readableBytes()`, which returns the combined size of the serialized MessageMetadata + payload. For null-value messages the payload is empty, but the metadata is always present, so `readableBytes()` is always > 0 (e.g. 32 bytes of metadata). This prevents the tombstone path (`size <= 0 → latestForKey.remove(key)`) from being reached in phase one of compaction.
The batch code path is not affected because it extracts per-message payload sizes from `SingleMessageMetadata.payloadSize`, which correctly returns 0 for null-value messages. The Java client always sets `numMessagesInBatch` in the metadata (even with `enableBatching(false)`), so all Java-produced messages go through the batch path — which is why this bug was never caught by existing tests.
### Modifications
- **`AbstractTwoPhaseCompactor.extractKeyAndSize()`**: Fixed to compute the correct payload-only size by using `Commands.skipMessageMetadata()` to skip past the metadata before reading `readableBytes()`. For compressed messages, `getUncompressedSize()` was already correct.
- **`EventTimeOrderCompactor.extractMessageCompactionData()`**: Refactored to reuse `extractKeyAndSize()` instead of duplicating the (buggy) size calculation.
### Verifying this change
- Added `CompactionTest.testNonBatchedMessageWithNullValue` — writes raw non-batch entries (no `numMessagesInBatch` in metadata, simulating C++/Python producers) directly to the managed ledger, triggers compaction, and verifies tombstoned keys are removed from the compacted view.
| Test                                | Without fix                        | With fix |
|-------------------------------------|------------------------------------|----------|
| testNonBatchedMessageWithNullValue  | FAILED: expected [2] but found [4] | PASSED   |
- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


pulsar-broker/src/main/java/org/apache/pulsar/compaction/AbstractTwoPhaseCompactor.java
```diff
@@ -462,14 +462,22 @@ private CompletableFuture<Void> addToCompactedLedger(LedgerHandle lh, RawMessage
...
     if (msgMetadata.hasPartitionKey()) {
-      int size = headersAndPayload.readableBytes();
+      int payloadSize;
       if (msgMetadata.hasUncompressedSize()) {
-        size = msgMetadata.getUncompressedSize();
+        payloadSize = msgMetadata.getUncompressedSize();
+      } else {
+        ByteBuf headersAndPayload = m.getHeadersAndPayload();
+        Commands.skipMessageMetadata(headersAndPayload);
+        payloadSize = headersAndPayload.readableBytes();
```
lhotari: It's better to call `ByteBuf.duplicate()` so that the original headerAndPayload buffer's readIndex doesn't advance:
```suggestion
        ByteBuf headersAndPayload = m.getHeadersAndPayload().duplicate();
        Commands.skipMessageMetadata(headersAndPayload);
        payloadSize = headersAndPayload.readableBytes();
```

grishaf:
PR failed on a known flaky test
https://github.com/apache/pulsar/issues/25815

lhotari:
> PR failed on a known flaky test #25815

@grishaf btw. in case of flaky tests, you can trigger a retry for the CI build by adding a comment `/pulsarbot rerun` to the PR. This works only after the CI workflow jobs have all finished.

lhotari:
@grishaf Please avoid force pushes to a PR where the review has already started since it makes it harder to track changes after the previous review. Instead, merge the latest `origin/master` to your PR branch.

lhotari:
I had to trigger a full rebuild since "Pulsar CI" currently caches build artifacts for 3 days and if the artifacts expire, there's a need to trigger a full rebuild. Merging origin/master to the PR branch is a recommended approach. Another way to trigger is to push a commit or close and re-open the PR (however that's not a recommended approach for retrying after flaky tests, `/pulsarbot rerun` should be used instead unless artifacts have expired).

lhotari (approved):
LGTM, thanks for the contribution @grishaf

pulsar-broker/src/main/java/org/apache/pulsar/compaction/AbstractTwoPhaseCompactor.java
```diff
@@ -462,14 +462,22 @@ private CompletableFuture<Void> addToCompactedLedger(LedgerHandle lh, RawMessage
...
     if (msgMetadata.hasPartitionKey()) {
-      int size = headersAndPayload.readableBytes();
+      int payloadSize;
       if (msgMetadata.hasUncompressedSize()) {
-        size = msgMetadata.getUncompressedSize();
+        payloadSize = msgMetadata.getUncompressedSize();
+      } else {
+        ByteBuf headersAndPayload = m.getHeadersAndPayload().duplicate();
+        Commands.skipMessageMetadata(headersAndPayload);
+        payloadSize = headersAndPayload.readableBytes();
```
Denovo1998, line 480: Should we check `msgMetadata.hasNullValue() && msgMetadata.isNullValue()` before deriving the size from payload bytes?

The new payload-only calculation fixes plain non-batch tombstones, but it can still keep encrypted tombstones when the producer does not set `uncompressedSize`. Encryption can turn an empty/null payload into non-empty ciphertext, so after `skipMessageMetadata(...)`, `readableBytes()` may still be > 0 and phase one will put the tombstone into `latestForKey` instead of removing the key.

Java producers usually avoid this because `ProducerImpl#updateMessageMetadata` sets `uncompressedSize=0` for null payloads, but this PR is specifically fixing producers whose metadata does not include that size. The explicit `nullValue` flag is the protocol-level tombstone signal, so I think this helper should return size 0 for `nullValue=true` before falling back to `uncompressedSize` or payload bytes.

```java
if (msgMetadata.hasNullValue() && msgMetadata.isNullValue()) {
    payloadSize = 0;
} else if (msgMetadata.hasUncompressedSize()) {
    payloadSize = msgMetadata.getUncompressedSize();
} else {
    ByteBuf headersAndPayload = m.getHeadersAndPayload().duplicate();
    Commands.skipMessageMetadata(headersAndPayload);
    payloadSize = headersAndPayload.readableBytes();
}
```

pulsar-broker/src/test/java/org/apache/pulsar/compaction/CompactionTest.java
```diff
@@ -640,6 +640,80 @@ public void testBatchMessageWithNullValue() throws Exception {
         assertEquals(messages.get(2).getKey(), "key5");
     }
 
+    /**
+     * Write raw non-batch entries directly to the managed ledger, simulating
+     * messages from C++/Python clients that do not set numMessagesInBatch.
+     * Verifies that null-value tombstones remove keys during compaction.
```
Denovo1998: Small wording nit: this comment says the raw entries simulate C++/Python clients because they do not set `numMessagesInBatch`. The Java non-batch send path also does not set `MessageMetadata.numMessagesInBatch`; Java avoids the old bug mainly because `ProducerImpl#updateMessageMetadata` sets `uncompressedSize=0` for null payloads.

Could we reword this as "raw non-batch entries without `uncompressedSize`, as seen with some non-Java clients" so future readers don't infer that `numMessagesInBatch` is what protects Java non-batch messages?

Denovo1998 (changes requested):

Denovo1998 (approved):

BewareMyPower (approved):

PR #25819, opened 2026-05-19
[improve][fn] Avoid gRPC timeout when getting status of a dead process runtime

Dream95:
### Motivation

In process runtime mode,  When the child process had exited, the stub could still be non-null, so the call continued through gRPC getFunctionStatus and could block until the 5-second deadline expired.

﻿Aligned ProcessRuntime.getFunctionStatus() with ThreadRuntime.getFunctionStatus() by checking !isAlive() before invoking gRPC.

### Modifications

- In `ProcessRuntime.getFunctionStatus()`, if `!isAlive()`, complete immediately with `running=false` and populate `failureException` from the stored `deathException` when available, without invoking gRPC.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


lhotari (approved):
LGTM

PR #25821, opened 2026-05-19
[improve][broker] PIP-473 P3.5: durable per-segment visibility state

merlimat:
## Summary

Fixes a latent correctness bug in P3's `MetadataTransactionBuffer` ([#25768](https://github.com/apache/pulsar/pull/25768)): a TB restart (or new `EARLIEST` subscription) on a topic with old committed transactional data filtered those committed messages, because the original `/txn/id/<txnId>` headers had been GC'd and P3's `isTxnAborted` defaulted "unknown → aborted" while cleanup-on-apply also removed the `/txn/op` records. Committed-and-GC'd became indistinguishable from orphan.

This PR adds the durable per-segment visibility state that the legacy snapshot used to provide, relocated to the metadata store and pruned in step with the segment ML's trimming. Combined with a publish-path ordering reversal that eliminates orphans by construction, the design no longer needs the "unknown → aborted" default.

Design note: `.claude/notes/pip-473-visibility-state.md` in the v5-transactions worktree.

### Publish-path ordering — no orphans

`appendBufferToTxn` writes the `/txn/op` record **before** the ML append:

1. Read header (cache-first). Reject if terminal.
2. PUT `/txn/op/<txnId>-<seq>`.
3. ML `asyncAddEntry`.
4. Ack publisher.

A crash between (2) and (3) leaves an op record with no data — TC times out → ABORTED → participant cleanup deletes the record. There is never a ledger entry with no corresponding `/txn/op` record. WRITE op records carry sentinel 0/0 positions; positions live in the TB's in-memory tracking only.

### Durable per-segment state

Under `/txn/segment-state/...`:

| Path | Schema |
|---|---|
| `/txn/segment-state/watermark/<encoded-segment>` | `SegmentWatermark(ledgerId, entryId)` |
| `/txn/segment-state/aborted/<encoded-segment>:<txnId>` | `AbortedTxnRecord(maxLedgerId, maxEntryId)` |

Both carry `partitionKey = segmentKey(segment)`. The aborted records also have a `idx:txn-aborted-by-position` secondary-index entry keyed by `<encoded-segment>:<padded-position>` — backend-agnostic range trimming.

### `isTxnAborted` semantics

Flipped to "below watermark + not in aborted set → visible". The aborted set is hydrated from durable records at recovery and updated on abort-apply. Unknown txns now default visible — this is the bug fix.

### Recovery & apply ordering

Recovery loads the durable watermark + aborted set, then scans `/txn/op` for in-flight open txns at last shutdown. Recovery-discovered opens have unknown positions; `maxReadPosition` pins at the watermark until they resolve. **No segment-replay needed** — the orphan-elimination invariant makes `/txn/op` an authoritative source.

Apply ordering serialises through a `stateTail` `CompletableFuture` chain so concurrent applies don't race the watermark CAS: in-memory state change → write aborted record (if abort) → CAS watermark forward → delete `/txn/op` records.

### Test plan

- [x] `pulsar-broker:test --tests MetadataTransactionBufferTest` — 10 cases, all green. Includes the bug-fix scenarios:
  - `restartAfterCommit_committedDataStillVisible`
  - `restartAfterAbort_abortedTxnStillFiltered`
  - `recoveryDiscoveredOpenTxn_pinsAtWatermark`
  - `publishOrdering_opRecordWrittenBeforeMlAppend`
- [x] `pulsar-broker:test --tests TxnMetadataStoreTest` — 7 cases (P2) still green after the new segment-state helpers.
- [x] `pulsar-broker:test --tests TxnIdsTest` — 7 cases.
- [x] `pulsar-broker:test --tests MetadataPendingAckStoreTest` — 6 cases (P4) unaffected.
- [x] `pulsar-broker:test --tests Dispatching*Test` — 4 routing tests pass.

### Deferred / follow-ups

- **ML trim hook for pruning aborted records.** The `idx:txn-aborted-by-position` index is in place; wiring the actual scan-and-delete to ML trim events is a follow-up. Aborted records accumulate until segment teardown for now.
- **PendingAck-side cursor durability check.** Verify in P4's apply path that `handle.commitTxn` / `abortTxn`'s future completes only after the cursor write is durable (so we never delete `/txn/op` ACK records before the cursor reflects the commit). Separate small verification.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
@@ -41,6 +41,7 @@ import org.apache.pulsar.broker.transaction.buffer.AbortedTxnProcessor;
 import org.apache.pulsar.broker.transaction.buffer.TransactionBuffer;
 import org.apache.pulsar.broker.transaction.buffer.TransactionBufferReader;
 import org.apache.pulsar.broker.transaction.buffer.TransactionMeta;
+import org.apache.pulsar.broker.transaction.metadata.SegmentWatermark;
 import org.apache.pulsar.broker.transaction.metadata.TxnHeader;
 import org.apache.pulsar.broker.transaction.metadata.TxnIds;
 import org.apache.pulsar.broker.transaction.metadata.TxnMetadataStore;
@@ -58,35 +59,35 @@ import org.apache.pulsar.metadata.api.MetadataStoreException;
 import org.apache.pulsar.metadata.api.ScanConsumer;
 
 /**
- * {@link TransactionBuffer} for {@code segment://} topics that reads truth from the metadata-store
- * transaction layout (PIP-473) rather than from a per-topic snapshot log.
+ * {@link TransactionBuffer} for {@code segment://} topics that reads truth from the
+ * metadata-store transaction layout (PIP-473).
  *
- * <p>Lifecycle:
- * <ul>
- *   <li><b>Publish</b> — {@link #appendBufferToTxn} reads the txn header (cache-first), appends the
- *       entry to the managed ledger, then appends a {@link TxnOp} record under
- *       {@code /txn/op/<txnId>-<seq>}. Both must succeed before we ack the publisher.</li>
- *   <li><b>State transitions</b> — driven by {@code /txn/segment-events/<segment>-*} sequence
- *       events. The events are wake-ups; the truth is the header. On each notification we
- *       re-read headers for every currently-open txn and apply the resulting state changes.</li>
- *   <li><b>Recovery</b> (Option C) — scan {@code idx:writes-by-segment} for this segment, group by
- *       {@code txnId}, fetch each header, and seed the in-memory cache. Then subscribe to the
- *       event stream for forward updates.</li>
- * </ul>
+ * <p><b>Publish-path ordering.</b> {@link #appendBufferToTxn} writes the {@code /txn/op} record
+ * <em>before</em> the managed-ledger append. This eliminates the orphan class — a crash between
+ * the two writes leaves either (a) no entry and no op record, or (b) an op record with no entry
+ * (TC times out → ABORTED → cleanup). There's never a ledger entry with no op record. The invariant
+ * <i>every transactional entry in the segment has a corresponding {@code /txn/op} record at the
+ * time of append</i> lets recovery scan {@code /txn/op} authoritatively without segment-replay.
+ *
+ * <p><b>Durable visibility state.</b> The TB persists a per-segment
+ * {@code /txn/segment-state/<segment>/watermark} record (the resolved-below mark) plus one
+ * {@code aborted/<txnId>} record per aborted txn with still-readable data. Together they let
+ * {@link #isTxnAborted} answer correctly for as long as the data is in the segment ML, even after
+ * the original {@code /txn/id/<txnId>} headers have been GC'd. The aborted records carry the txn's
+ * max position via a secondary index so the TB can range-delete them when the segment ML trims its
+ * older data.
+ *
+ * <p><b>Recovery.</b> Load the durable watermark + aborted set; scan {@code /txn/op} for this
+ * segment to discover any txns still open at the time of the previous shutdown (their first
+ * positions aren't known — the in-memory watermark stays pinned at the durable value until they
+ * resolve); subscribe to the segment-event stream.
  *
  * <p><b>TC ordering contract.</b> There is a TOCTOU window between the header authorization read
- * in {@link #appendBufferToTxn} and the managed-ledger append: the TC may flip the header (commit
- * or abort) in between, and the entry still lands. On commit that's harmless — the message is
- * visible. On abort, the subsequent segment-event delivery marks the txn ABORTED in the cache and
- * {@link #isTxnAborted} filters it. This relies on the TC publishing the segment event <em>after</em>
- * the header CAS lands so a participant that lost the race always learns the decision. The legacy
- * {@code TopicTransactionBuffer} has the same window with marker-message ordering.
- *
- * <p><b>In-memory growth.</b> Terminal txns stay in the {@code txns} cache for the segment's
- * lifetime so {@code isTxnAborted} can answer authoritatively for dispatcher reads — evicting a
- * COMMITTED entry would mean the default "unknown → aborted" filter wrongly hides its messages.
- * Long-running segments with high txn turnover will accumulate cached entries. Cache pruning tied
- * to data-ledger trimming / header GC is a P5/P6 concern.
+ * and the {@code /txn/op} write: the TC may flip the header (commit or abort) in between. The op
+ * record still lands. On commit that's harmless. On abort, the subsequent segment-event delivery
+ * marks the txn ABORTED and the per-txn aborted record is written before the op record is deleted,
+ * so {@link #isTxnAborted} filters subsequent reads. This relies on the TC publishing the segment
+ * event <em>after</em> the header CAS.
  */
 @CustomLog
 public class MetadataTransactionBuffer implements TransactionBuffer {
@@ -101,14 +102,34 @@ public class MetadataTransactionBuffer implements TransactionBuffer {
     private volatile AutoCloseable subscription;
     private volatile boolean closed;
 
-    /** Guards {@link #txns} + {@link #maxReadPosition} + {@link #lastDispatchable}. */
+    /** Guards mutable state below. */
     private final Object lock = new Object();
 
-    /** Cached per-txn state, populated by appendBufferToTxn and refreshed by event reconcile. */
+    /** In-memory per-txn state for txns this segment is involved in. */
     private final Map<String, TxnEntry> txns = new HashMap<>();
 
-    private Position maxReadPosition;
+    /** Aborted-txn set hydrated from durable state at recovery, updated on abort-apply. */
+    private final Set<String> abortedTxns = new HashSet<>();
+
+    /** Count of OPEN txns we discovered at recovery whose first position we don't know. */
+    private int recoveryDiscoveredOpen;
+
+    /** Durable watermark, mirrored in memory. May be null on a fresh segment. */
+    private SegmentWatermark watermark;
+    /** Version of the durable watermark record; -1 if it doesn't exist yet. */
+    private long watermarkVersion = -1L;
+
+    /** Latest dispatched position from non-txn publishes — the natural ceiling when no opens pin. */
     private Position lastDispatchable;
+
+    /** Current maxReadPosition; never moves above the watermark while recovery-discovered opens exist. */
+    private Position maxReadPosition;
+
+    /**
+     * Serialised chain for watermark-persist + op-record-cleanup. Each apply enqueues itself on the
+     * tail so we never have two in-flight watermark CASes racing.
+     */
+    private CompletableFuture<Void> stateTail = CompletableFuture.completedFuture(null);
 
     private final LongAdder committedCount = new LongAdder();
     private final LongAdder abortedCount = new LongAdder();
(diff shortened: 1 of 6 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
```diff
@@ -136,46 +157,82 @@ private void recover() {
...
-                    return;
-                }
-                opsByTxn.computeIfAbsent(txnIdKey, k -> new ArrayList<>())
-                        .add(PositionFactory.create(op.getLedgerId(), op.getEntryId()));
-            }
+        // 1. Load durable watermark.
+        CompletableFuture<Void> watermarkLoad = txnStore.getSegmentWatermark(segmentName)
+                .thenAccept(opt -> {
+                    if (opt.isPresent()) {
+                        synchronized (lock) {
```
lhotari, line 164: Just one comment about synchronization.

A further improvement could be to refactor to a model which avoids synchronization. One approach would be to use virtual threads on Java 25 which handles synchronization in a non-blocking way, however that might not be feasible with Netty. Another approach would be to follow "single writer principle" and ensure that mutations happen in a single thread sequentially. A third option would be to use state objects that tie all individual state changes together and use volatile or CAS to ensure that state transitions are atomic without inconsistencies. The current Pulsar code base uses a lot of synchronization and it would be great if we could reduce that since synchronization causes issues with Netty IO threads and thread pools in general.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnPaths.java
```diff
@@ -134,6 +149,85 @@ public static String ackIndexKey(String segment, String subscription) {
...
+     *     encoded prefix scopes scans to one segment; the padded position is lexicographically
+     *     ordered so range scans by trim point work naturally.
+     */
+    public static String abortedByPositionIndexKey(String segment, long ledgerId, long entryId) {
+        return segmentKey(segment) + ":" + longKey(ledgerId) + ":" + longKey(entryId);
+    }
+
+    /** @return the lower bound of the per-segment range in {@link #IDX_TXN_ABORTED_BY_POSITION}. */
+    public static String abortedByPositionSegmentLowerBound(String segment) {
+        return segmentKey(segment) + ":" + longKey(0L) + ":" + longKey(0L);
```
lhotari: the suffix could be a constant

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnPaths.java
```diff
@@ -134,6 +149,85 @@ public static String ackIndexKey(String segment, String subscription) {
...
+    }
+
+    /** @return the upper bound of the per-segment range in {@link #IDX_TXN_ABORTED_BY_POSITION}. */
+    public static String abortedByPositionSegmentUpperBound(String segment) {
+        return segmentKey(segment) + ":" + MAX_LONG_KEY + ":" + MAX_LONG_KEY;
```
lhotari: the suffix could be a constant

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnPaths.java
```diff
@@ -134,6 +149,85 @@ public static String ackIndexKey(String segment, String subscription) {
...
+            return null;
+        }
+        String name = abortedPath.substring(lastSlash + 1);
+        int colon = name.indexOf(':');
+        return colon < 0 ? null : name.substring(0, colon);
+    }
+
     /** @return {@code value} formatted as a zero-padded fixed-width decimal for use as a range-scan index key. */
     public static String longKey(long value) {
         return String.format("%0" + LONG_KEY_WIDTH + "d", value);
```
lhotari: unrelated to this PR. For performance reasons, String.format should be avoided in hot paths.

Claude suggested this type of optimization (assumes non-negative value):
```java
public static String longKey(long value) {
    char[] buf = new char[20];
    long v = value; // assume non-negative for brevity
    for (int i = 19; i >= 0; i--) {
        buf[i] = (char) ('0' + (v % 10));
        v /= 10;
    }
    return new String(buf);
}
```

lhotari:
These are findings from a local Claude Code review. Please check before merging.

## [BUG] Recovery race: aborted txn discovered via leftover `/txn/op` record is visible until `applyTerminalNow` finishes

For a txn whose `/txn/op` records survived the previous broker's crash and whose header is ABORTED (or GC'd → defaulted to ABORTED), `applyHeaderForRecovery` writes `entry.state = ABORTED` into `txns` but does **not** add to `abortedTxns`. Only the eventual `applyTerminalNow` (enqueued on `stateTail`) adds to the set. Meanwhile `recoveryFuture.complete(null)` fires as soon as the per-txn header reads finish, before `stateTail` settles.

`isTxnAborted` consults only `abortedTxns` — not the `txns` map. So in the window between recovery completing and the queued `applyTerminalNow` finishing, `isTxnAborted` returns **false** for these recovered-aborted txns. A dispatcher that's already attached, or a new EARLIEST subscription created in that window, can read above the watermark (recovery clamps `maxReadPosition` only when there are recovery-discovered **OPEN** txns) and expose data that should be filtered.

`restartAfterAbort_abortedTxnStillFiltered` doesn't exercise this path because the test pre-populates the durable aborted record, so `abortedTxns` is hydrated during `scanAbortedTxns` and the race never opens. The hole is the "previous broker crashed mid-apply, durable aborted record never written, only the `/txn/op` record remains" path.

Suggested fix: in `applyHeaderForRecovery`, when `state == TxnState.ABORTED`, immediately `abortedTxns.add(txnIdKey)` under the lock. `applyTerminalNow`'s idempotent `else-if` branch already tolerates this. A regression test for the "op record exists, header GC'd or ABORTED, no durable aborted record" scenario would lock this in.

## [BUG] `persistAbortedRecord` uses current watermark as max-position for recovery-discovered txns — comment claims this is conservative, but the math goes the other way

The trim-driven pruning rule (per the index design in `TxnPaths`) is "prune aborted records whose max-position < trim point." For an aborted record to stick around as long as needed, its stored max position must be ≥ the *highest* position of the txn's data in the segment.

For a recovery-discovered txn the TB doesn't know the real positions, so `persistAbortedRecord` falls back to `watermark.ledgerId()/entryId()`. But the watermark is the *lowest* still-resolved point in the segment — the txn's actual data sits *above* the watermark (that's exactly why recovery pins `maxReadPosition` at the watermark while these txns remain open). Storing `max = watermark` means the trim-pruner will drop the aborted record as soon as the ML trims past the watermark, while the txn's real data is still readable at higher positions. Once the record is gone, a future restart hydrates `abortedTxns` without this entry, and `isTxnAborted` returns false → the aborted entries become visible.

The javadoc says "the worst case is the record sits around longer" — but `max = watermark` is the *earliest* prunable position, not the latest. Suggested fix: for the unknown-position case, use the current `maxReadPosition` (the segment LAC at the time the txn resolves, since the watermark pin is just being released) or `Long.MAX_VALUE` — both keep the record alive until the segment is fully trimmed.

This is latent: trim-driven pruning is explicitly deferred per the PR description, so the buggy max position has no consumer yet. Worth fixing in this PR before the trim wiring lands — otherwise it's a hidden footgun for the next author.

## [QUALITY] `wasRecoveryDiscovered` parameter is dead

`persistAbortedRecord(String txnIdKey, Position lastPos, boolean wasRecoveryDiscovered)` — the third parameter is passed but never referenced in the body. The `lastPos \!= null` check is what actually gates the fallback. Either drop the parameter or make it the explicit gate.

## [QUALITY] Empty `else if` branch with the operative side-effect in the condition

```java
} else if (newState == TxnState.ABORTED && abortedTxns.add(txnIdKey)) {
    // Idempotent path — header re-confirms ABORTED and we hadn't recorded it yet...
}
```

Using `Set.add()` for its side effect inside a boolean condition with an empty body is misleading. Prefer:
```java
} else if (newState == TxnState.ABORTED) {
    abortedTxns.add(txnIdKey);
}
```

## [QUALITY] `persistWatermarkIfAdvanced` doesn't recover from a stale `watermarkVersion` on CAS failure

If `casSegmentWatermark` fails (version mismatch from a split-brain scenario, or any transient error), the `.exceptionally` handler just logs and `watermarkVersion` stays stale in memory. Subsequent retries through `stateTail` CAS with the same stale version and keep failing. For the single-owner segment model this shouldn't happen, but if it does the only recovery is restarting the TB. Consider re-reading the watermark on CAS failure to self-heal.

## [QUALITY] Two consecutive `synchronized (lock)` blocks in `syncMaxReadPositionForNormalPublish`

The non-txn-publish path acquires the lock, updates `lastDispatchable`, releases, then re-acquires only to compose onto `stateTail`. Could be one block.

## [QUALITY] Redundant ternary on `expectedVersion`

In `persistWatermarkIfAdvanced`:
```java
Optional<Long> expected = expectedVersion == -1L ? Optional.of(-1L) : Optional.of(expectedVersion);
```
Both branches yield the same value — simplify to `Optional.of(expectedVersion)`.

## [QUALITY] Test coverage gap: the recovery-rebuilds-OPEN scenario lost a useful assertion

The deleted `recovery_rebuildsOpenTxnStateFromOpRecords` checked `isTxnAborted` for an OPEN recovery-discovered txn (must return false) and validated that op records on a *different* segment weren't mistakenly recovered here. The new `recoveryDiscoveredOpenTxn_pinsAtWatermark` covers `maxReadPosition` pinning but drops both of those assertions. Worth carrying them forward.

## [INTENT MISMATCH] Soft invariant: "messages above the watermark never reach `isTxnAborted`"

The comment in `isTxnAborted` says "The watermark itself caps the dispatcher, so messages above it never reach this check." During normal operation `maxReadPosition` is `previous(min(open firstPositions))` — it caps at the *lowest* open txn's first write, so committed-but-still-pending-cleanup data sits above. The watermark-only cap is only true while recovery-discovered opens exist. The invariant is correct in spirit but the comment's framing is loose. Fine as-is; just don't write code in a follow-up that relies on it being literally true.

lhotari (approved):
Minor comments and some unrelated comments. I did a local Claude Code review which brought up some finding that you might want to check before merging.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnPaths.java
merlimat: Extracted into `MIN_POSITION_SUFFIX`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnPaths.java
merlimat: Extracted into `MAX_POSITION_SUFFIX`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/metadata/TxnPaths.java
merlimat: Done — replaced `String.format` with a direct zero-padded char-buffer fill. Verified it produces identical output across our value range (non-negative ledger/entry ids and epoch millis).

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
merlimat: Agreed the coarse `synchronized(lock)` here isn't ideal. I'd like to keep this PR scoped to the durable visibility state and address the concurrency model as a follow-up. The single-writer approach looks like the best fit since the buffer is already driven from per-topic ordered callbacks — that would let us drop the lock without blocking Netty IO threads. Thanks for laying out the options.

merlimat:
Thanks for the thorough local review — went through all of it. Pushed fixes in 8753e3c.

**Confirmed bugs, both fixed:**

1. **Recovery race** — exactly right. A recovery-discovered ABORTED txn was only added to `abortedTxns` by the queued `applyTerminalNow` on `stateTail`, which can settle after `recoveryFuture` completes, and ABORTED recovery entries aren't watermark-pinned — so `isTxnAborted` could return false in that window. Now hydrate `abortedTxns` synchronously under the lock in `applyHeaderForRecovery`. The race only opens with 2+ terminal txns (the first applies synchronously since `stateTail` is complete; subsequent ones queue behind its in-flight persist), so the regression test forces that with two txns and a hung first `putAbortedTxn`. Verified it fails without the fix.

2. **`persistAbortedRecord` max-position** — agreed, the math goes the wrong way. The txn's data sits *above* the watermark, so keying the prune on the watermark drops the record too early. Switched the unknown-position fallback to the segment LAC (`ledger.getLastConfirmedEntry()`) — a correct conservative upper bound since the data was written in a prior epoch. Regression test asserts the stored position is the LAC, not the watermark, and fails without the fix.

**Quality items:**
- Dropped the dead `wasRecoveryDiscovered` parameter (the `lastPos != null` check is the real gate).
- Hoisted `abortedTxns.add()` out of the empty `else if` condition.
- `persistWatermarkIfAdvanced` now self-heals on CAS `BadVersionException` — re-reads the watermark so the next retry uses a fresh version instead of looping.
- Merged the two consecutive `synchronized` blocks in `syncMaxReadPositionForNormalPublish`.
- Simplified the redundant `expectedVersion` ternary.
- Tightened the `isTxnAborted` invariant comment to describe the actual `maxReadPosition` cap (lowest open txn's first write in steady state; watermark only while recovery-discovered opens remain) rather than implying a watermark-only cap.
- Restored the dropped `isTxnAborted == false` + cross-segment isolation assertions in `recoveryDiscoveredOpenTxn_pinsAtWatermark`.

On the broader synchronization refactor (from the inline comment): I'd like to keep that as a follow-up — the single-writer approach looks like the best fit since the buffer is already driven from per-topic ordered callbacks.

lhotari (approved):

PR #25825, opened 2026-05-19
[fix][bk] Fix NPE in IsolatedBookieEnsemblePlacementPolicy when policy class does not match

Shawyeok:
Fixes NPE in `IsolatedBookieEnsemblePlacementPolicy`

### Motivation

A `NullPointerException` occurs in `getExcludedBookiesWithIsolationGroups` when `getIsolationGroup()` returns a `Pair` whose `left` value is `null`. This happens when the `EnsemblePlacementPolicyConfig` carries a policy class name that does not equal `IsolatedBookieEnsemblePlacementPolicy` — in that case the `if` block that calls `pair.setLeft()` / `pair.setRight()` is skipped entirely, leaving both sides of the `MutablePair` as `null`.

This bug was observed during the upgrade from the 2.8 branch to the 3.0 branch. The same defect may also exist in master.

**Full stack trace from production:**

```
2026-05-18T13:28:40.756Z [BookKeeperClientWorker-OrderedExecutor-35-0] ERROR org.apache.bookkeeper.common.util.SingleThreadExecutor - Error while running task: Cannot invoke "java.util.Set.contains(Object)" because the return value of "org.apache.commons.lang3.tuple.Pair.getLeft()" is null
java.lang.NullPointerException: Cannot invoke "java.util.Set.contains(Object)" because the return value of "org.apache.commons.lang3.tuple.Pair.getLeft()" is null
    at org.apache.pulsar.bookie.rackawareness.IsolatedBookieEnsemblePlacementPolicy.getExcludedBookiesWithIsolationGroups(IsolatedBookieEnsemblePlacementPolicy.java:192)
    at org.apache.pulsar.bookie.rackawareness.IsolatedBookieEnsemblePlacementPolicy.getExcludedBookies(IsolatedBookieEnsemblePlacementPolicy.java:141)
    at org.apache.pulsar.bookie.rackawareness.IsolatedBookieEnsemblePlacementPolicy.replaceBookie(IsolatedBookieEnsemblePlacementPolicy.java:127)
    at org.apache.bookkeeper.client.BookieWatcherImpl.replaceBookie(BookieWatcherImpl.java:316)
    at org.apache.bookkeeper.client.EnsembleUtils.replaceBookiesInEnsemble(EnsembleUtils.java:69)
    at org.apache.bookkeeper.client.ReadOnlyLedgerHandle.handleBookieFailure(ReadOnlyLedgerHandle.java:222)
    at org.apache.bookkeeper.client.PendingAddOp.writeComplete(PendingAddOp.java:353)
    at org.apache.bookkeeper.proto.BookieClientImpl.completeAdd(BookieClientImpl.java:287)
    at org.apache.bookkeeper.proto.BookieClientImpl.access$200(BookieClientImpl.java:79)
    at org.apache.bookkeeper.proto.BookieClientImpl$ChannelReadyForAddEntryCallback.lambda$operationComplete$0(BookieClientImpl.java:405)
    at org.apache.bookkeeper.common.util.OrderedExecutor$TimedRunnable.run(OrderedExecutor.java:203)
    at org.apache.bookkeeper.common.util.SingleThreadExecutor.safeRunTask(SingleThreadExecutor.java:137)
    at org.apache.bookkeeper.common.util.SingleThreadExecutor.run(SingleThreadExecutor.java:107)
    at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
    at java.base/java.lang.Thread.run(Thread.java:840)
```

This error causes topics to time out during loading. Since there is no automatic recovery path, the only way to restore service is by restarting the broker.

### Modifications

Two complementary fixes:

1. **`getIsolationGroup()`**: Ensure not returning pair of null: add compatibility with `ZkIsolatedBookieEnsemblePlacementPolicy` and fallback to defaultIsolationGroups if met a mis-match policy class.

### Verifying this change

- Added `testReplaceBookieWithNonMatchingPolicyClassShouldNotThrowNPE` to reproduce the exact production failure: calling `replaceBookie` with custom metadata whose `EnsemblePlacementPolicyConfig` policy class does not match `IsolatedBookieEnsemblePlacementPolicy`, which previously caused the NPE.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/bookie/rackawareness/IsolatedBookieEnsemblePlacementPolicy.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/bookie/rackawareness/IsolatedBookieEnsemblePlacementPolicy.java
@@ -166,11 +166,12 @@ public class IsolatedBookieEnsemblePlacementPolicy extends RackawareEnsemblePlac
         return Optional.empty();
     }
 
-    private static Pair<Set<String>, Set<String>> getIsolationGroup(
+    @VisibleForTesting
+    Pair<Set<String>, Set<String>> getIsolationGroup(
             EnsemblePlacementPolicyConfig ensemblePlacementPolicyConfig) {
-        MutablePair<Set<String>, Set<String>> pair = new MutablePair<>();
-        String className = IsolatedBookieEnsemblePlacementPolicy.class.getName();
-        if (ensemblePlacementPolicyConfig.getPolicyClass().getName().equals(className)) {
+        // Retain compatibility with ZkIsolatedBookieEnsemblePlacementPolicy
+        if (IsolatedBookieEnsemblePlacementPolicy.class.isAssignableFrom(ensemblePlacementPolicyConfig.getPolicyClass())) {
+            MutablePair<Set<String>, Set<String>> pair = new MutablePair<>(Collections.emptySet(), Collections.emptySet());
             Map<String, Object> properties = ensemblePlacementPolicyConfig.getProperties();
             String primaryIsolationGroupString = ConfigurationStringUtil
                     .castToString(properties.getOrDefault(ISOLATION_BOOKIE_GROUPS, ""));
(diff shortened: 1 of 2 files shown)


pulsar-broker-common/src/main/java/org/apache/pulsar/bookie/rackawareness/IsolatedBookieEnsemblePlacementPolicy.java
```diff
@@ -215,13 +218,7 @@ Set<BookieId> getExcludedBookiesWithIsolationGroups(int ensembleSize,
...
-                Set<String> primaryIsolationGroup = Collections.emptySet();
-                Set<String> secondaryIsolationGroup = Collections.emptySet();
                 Set<BookieId> primaryGroupBookies = new HashSet<>();
-                if (isolationGroups != null) {
-                    primaryIsolationGroup = isolationGroups.getLeft();
-                    secondaryIsolationGroup = isolationGroups.getRight();
-                }
                 for (String group : allGroups) {
                     Set<String> bookiesInGroup = allGroupsBookieMapping.get(group).keySet();
                     if (!primaryIsolationGroup.contains(group)) {
```
void-ptr974, line 232: This is the problematic behavior for the non-matching policyClass case. After this PR, getIsolationGroup() returns empty primary/secondary groups when the metadata policyClass is not
IsolatedBookieEnsemblePlacementPolicy. With primaryIsolationGroup empty, this condition is true for every configured group, so all grouped bookies are added to excludedBookies below. That avoids the NPE but can
still make replaceBookie fail with BKNotEnoughBookiesException. A non-matching policyClass should be ignored or fall back to defaultIsolationGroups, not be treated as an empty isolation group.

lhotari (approved):
LGTM

pulsar-broker-common/src/main/java/org/apache/pulsar/bookie/rackawareness/IsolatedBookieEnsemblePlacementPolicy.java
lhotari: @Shawyeok Please check this. I wonder if there's a test case that would cover the explained scenario?

Shawyeok: @void-ptr974 
Good point! The tricky part is that handling of empty isolation group in `IsolatedBookieEnsemblePlacementPolicy`. Now it as treated as negative for all bookies.

Currently, we can only call the `getExcludedBookiesWithIsolationGroups` method before resolving any possible pair of null scenarios.

Shawyeok: I've made a change of fallback to use defaultIsolationGroups when policy class doesn't match. Feel free to take a look. @void-ptr974 @lhotari

void-ptr974 (approved):
LGTM, thanks for addressing the fallback behavior.

dao-jun (approved):

Shawyeok:
@lhotari @dao-jun 
All tests have passed, feel free to take a look, thanks.

PR #25860, opened 2026-05-23
[fix][broker] Prevent duplicate geo-replicated messages after target topic reload

### LINKED ISSUE #25861: [Bug] Geo-replication V2 dedup watermark can be lost after target topic reload
### Problem

Persistent geo-replication V2 deduplication uses the source topic position as the target-side dedup watermark. For a replicated message, the source replicator adds `__MSG_PROP_REPL_SOURCE_POSITION=<source-ledger-id>:<source-entry-id>`, and the target broker stores the latest replicated source position as:

- `<replicator-producer>_LID`
- `<replicator-producer>_EID`

This state is not normal producer sequence state. It is the target-side checkpoint used to identify whether a replayed source entry has already been persisted.

There are three related issues:

1. The geo V2 watermark is not recovered from replayed target entries.
2. The geo V2 watermark is stored as two separate snapshot keys, but `_LID` and `_EID` must be restored together.
3. The geo V2 watermark can be removed by normal producer inactivity cleanup, even though it is source-position state rather than producer lifecycle state.

### Failure window

1. Source replicator sends messages to the target.
2. Target persists the messages and updates the in-memory geo V2 dedup watermark.
3. The target topic is unloaded before the latest dedup snapshot includes the watermark.
4. The source replicator has not durably advanced its replication cursor yet.
5. Source reconnects and replays the same source entries.
6. Target reloads without the geo watermark and can write duplicates.

### Impact

This can produce duplicate messages on the target cluster during normal recovery/failover paths. It does not require client misuse or corrupted input. Source cursor replay is an expected recovery behavior, so target-side geo dedup state must survive topic reload and producer inactivity cleanup.

### Proposed fix

- Recover geo-replication V2 dedup watermarks while replaying the dedup cursor by reading `replicatedFrom` and `__MSG_PROP_REPL_SOURCE_POSITION` from persisted replicated messages.
- Store geo-replication watermark keys as complete `_LID/_EID` pairs in dedup snapshots.
- Keep geo-replication watermark state during producer inactivity cleanup.

A proposed fix is available in #25860.

void-ptr974:
Fixes #25861

### Motivation

Persistent geo-replication V2 deduplication uses the source topic position as the target-side deduplication watermark. The source replicator stores that position in `__MSG_PROP_REPL_SOURCE_POSITION=<source-ledger-id>:<source-entry-id>` and the target tracks the latest persisted position for each replication producer.

When a target topic is loaded, deduplication state is restored from the `pulsar.dedup` cursor snapshot and entries after the snapshot are replayed. Previously, replay restored normal producer sequence IDs but did not rebuild geo-replication V2 watermarks. This leaves the following failure window:

1. Replicated messages are persisted on the target.
2. The source has not durably advanced its replication cursor, for example because send receipts have not been processed.
3. The target topic is unloaded before a deduplication snapshot covers the latest replicated source positions.
4. The target reloads without those in-memory V2 watermarks.
5. The source reconnects and replays the same entries, which can then be persisted again.

V2 watermarks also need to remain complete and monotonic. A source position is a ledger/entry pair, and replay or delayed persistence callbacks must not replace a newer watermark with an older one. In addition, this replication checkpoint must not be removed by normal producer inactivity cleanup.

Replay must remain compatible with V1 replicated entries. V1 entries have `replicatedFrom` but no valid V2 source-position property, and their deduplication state is the original producer sequence ID.

### Modifications

- Keep geo/shadow replication watermarks in separate maps as immutable `ReplSourcePosition` pairs instead of independently updating `_LID` and `_EID` values in the normal producer maps.
- Restore complete replication watermark pairs from deduplication cursor properties. Incomplete `_LID`/`_EID` pairs are ignored and logged.
- Rebuild V2 watermarks while replaying persisted replicated entries after the last snapshot.
- Preserve V1 compatibility during replay: markers and replicated entries without a valid V2 source position restore the original producer sequence state.
- Reconstruct the correct replication producer key for geo-replicated and shadow-topic entries.
- Merge replayed and persisted replication watermarks by maximum source position so older replay records and delayed persistence callbacks cannot move state backward.
- Retain replication watermark state when inactive producers are purged.
- Restore pushed V2 state from persisted V2 state after a publish failure, matching the existing normal-producer reset behavior.
- Write each replication watermark to snapshots as a complete `_LID`/`_EID` pair. A pair counts as one producer toward `brokerDeduplicationMaxNumberOfProducers`; replication producers are selected before normal producers.
- Add the source-position property to shadow-replicated messages so their V2 watermark can be recovered.

The V2 duplicate decision retains the existing synchronized get/check/update pattern, now using the dedicated pushed-watermark map. Persisted and replayed watermark updates use `ConcurrentHashMap.merge` with a maximum-position function.

### Tests

The change includes coverage for:

- target unload before the latest V2 watermark is snapshotted while source send receipts are withheld;
- persisted V1 replication entries replayed after unload;
- mixed V1, V2, malformed, and marker replay;
- shadow producer-name recovery;
- monotonic replay and persistence callbacks;
- complete snapshot pairs and the global producer limit;
- inactive replication producers;
- V2 `NotDup` / `Unknown` / `Dup` state transitions and publish-failure reset;
- invalid or missing V2 source positions;
- independent watermarks for multiple replication producers.

Local verification:

```bash
./gradlew :pulsar-broker:test --no-configuration-cache \
  --tests org.apache.pulsar.broker.service.persistent.MessageDuplicationTest
./gradlew :pulsar-broker:checkstyleMain --no-configuration-cache
./gradlew :pulsar-broker:checkstyleTest --no-configuration-cache
git diff --check
```

All local checks above passed. The focused geo-replication unload/replay regression test also passed locally.

### Compatibility and scope

This change does not modify the public API, binary protocol, configuration defaults, or snapshot key format. Existing `_LID`/`_EID` cursor properties remain readable. V1 entries without a valid V2 source position continue to recover original producer sequence state.

This PR addresses deduplication recovery within the existing V1/V2 negotiation design. Coordinated mode transitions, rolling-upgrade fencing, and administrative rollback orchestration are outside its scope.

- [ ] The public API
- [ ] The schema
- [x] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


void-ptr974:
all CI pass in my local repo: https://github.com/void-ptr974/pulsar/actions/runs/26407472425

Denovo1998 (approved):

void-ptr974:
Hi @poorbarcode, this PR touches geo-replication V2 dedup watermark recovery and cleanup behavior. Since it is close to the geo-replicator lifecycle work you have been looking at, could you help take a look when you have time? Thanks!

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -356,104 +413,141 @@ private void setContextPropsIfRepl(PublishContext publishContext, ByteBuf header
...
+                    long entryId = Long.parseLong(ledgerIdAndEntryId[1]);
+                    if (ledgerId >= 0 && entryId >= 0) {
+                        return new ReplSourcePosition(ledgerId, entryId);
                     }
-                    long[] positionPair = new long[]{Long.valueOf(ledgerIdAndEntryId[0]).longValue(),
-                            Long.valueOf(ledgerIdAndEntryId[1]).longValue()};
-                    publishContext.setProperty(MSG_PROP_REPL_SOURCE_POSITION, positionPair);
-                    break;
+                } catch (NumberFormatException e) {
+                    // Log below.
```
poorbarcode: Seems logging below equals logging here, why not log here?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -264,10 +299,30 @@ private CompletableFuture<Void> replayCursor(ManagedCursor cursor) {
...
             highestSequencedPersisted.put(k, v);
+            producerRemoved(k);
+        });
+        replLedgerIds.forEach((producerName, ledgerId) -> {
+            Long entryId = replEntryIds.get(producerName);
+            if (entryId != null) {
+                ReplSourcePosition position = new ReplSourcePosition(ledgerId, entryId);
+                highestReplPositionPushed.put(producerName, position);
+                highestReplPositionPersisted.put(producerName, position);
+            }
```
poorbarcode, line 315: It is an unexpected situation that `entryId` does not exist. We need to print an error log

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -276,6 +331,8 @@ private CompletableFuture<Void> replayCursor(ManagedCursor cursor) {
         return replayTask.replay(cursor, (__, buffer) -> {
             final var metadata = Commands.parseMessageMetadata(buffer);
             final var producerName = metadata.getProducerName();
+            // Rebuild replication watermarks from entries written after the last dedup snapshot.
+            recoverReplWatermarkFromMetadata(metadata);
             final var sequenceId = Math.max(metadata.getHighestSequenceId(), metadata.getSequenceId());
```
poorbarcode: Since we already use the new collections `highestReplPositionPushed` and `highestReplPositionPersisted` to trace the deduplication key, we should not put them into `highestSequencedPushed` and `highestSequencedPersisted` anymore.

- `is marker`: it should be traced by `highestSequencedPushed` and `highestSequencedPersisted`
- otherwise: it should be traced by `highestReplPositionPushed` and `highestReplPositionPersisted`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -603,9 +687,17 @@ private CompletableFuture<Void> takeSnapshot(Position position) {
         }
 
         Map<String, Long> snapshot = new TreeMap<>();
+        highestReplPositionPersisted.forEach((producerName, replSourcePosition) -> {
+            snapshot.put(producerName + REPL_LEDGER_ID_SUFFIX, replSourcePosition.ledgerId());
+            snapshot.put(producerName + REPL_ENTRY_ID_SUFFIX, replSourcePosition.entryId());
+        });
+        int[] normalProducerCount = new int[1];
```
poorbarcode: Rather than using `int[1]`, using `int` is better

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -356,104 +413,141 @@ private void setContextPropsIfRepl(PublishContext publishContext, ByteBuf header
...
-                } else {
-                    return MessageDupStatus.Unknown;
-                }
+    private void recoverReplWatermark(String producerName, ReplSourcePosition replSourcePosition) {
+        highestReplPositionPushed.merge(producerName, replSourcePosition, ReplSourcePosition::max);
+        highestReplPositionPersisted.merge(producerName, replSourcePosition, ReplSourcePosition::max);
+    }
+
+    private void recordReplWatermarkPersisted(String producerName, ReplSourcePosition replSourcePosition) {
+        highestReplPositionPersisted.merge(producerName, replSourcePosition, ReplSourcePosition::max);
```
poorbarcode: If sorting is required here, it indicates that there are issues with other codes.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -102,6 +103,28 @@ public MessageDupUnknownException(String topicName, String producerName) {
...
+            }
+            return Long.compare(entryId, other.entryId);
+        }
+
+        static ReplSourcePosition max(ReplSourcePosition first, ReplSourcePosition second) {
+            return first.compareTo(second) >= 0 ? first : second;
+        }
+    }
+
+    private static class ReplDedupCheck {
```
poorbarcode: This class seems meaningless; it was just used in the method `isDuplicateReplV2`, using method internal variables can avoid defining a new class

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -102,6 +103,28 @@ public MessageDupUnknownException(String topicName, String producerName) {
         }
     }
 
+    @VisibleForTesting
+    record ReplSourcePosition(long ledgerId, long entryId) implements Comparable<ReplSourcePosition> {
```
poorbarcode: Using this object enhances the readability and maintainability of the code, but it also increases the GC pressure. I feel that this class is unnecessary

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -356,104 +413,141 @@ private void setContextPropsIfRepl(PublishContext publishContext, ByteBuf header
...
-                if (lastSequenceLIdPersisted != null && lastSequenceEIdPersisted != null
-                    && (replSequenceLId < lastSequenceLIdPersisted.longValue()
-                        || (replSequenceLId == lastSequenceLIdPersisted.longValue()
-                            && replSequenceEId <= lastSequenceEIdPersisted))) {
-                    return MessageDupStatus.Dup;
-                } else {
-                    return MessageDupStatus.Unknown;
-                }
+    private void recoverReplWatermark(String producerName, ReplSourcePosition replSourcePosition) {
+        highestReplPositionPushed.merge(producerName, replSourcePosition, ReplSourcePosition::max);
```
poorbarcode: If sorting is required here, it indicates that there are issues with other codes.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -356,104 +413,141 @@ private void setContextPropsIfRepl(PublishContext publishContext, ByteBuf header
...
+        ReplSourcePosition replSourcePosition = getReplSourcePosition(publishContext);
+        if (replSourcePosition == null) {
+            logInvalidReplSourcePosition(publishContext, "Message cannot determine whether it is duplicated due to "
+                    + "the acquired message props being invalid, prop not in expected format");
+            return MessageDupStatus.Unknown;
+        }
+
+        String producerName = publishContext.getProducerName();
+        ReplDedupCheck check = new ReplDedupCheck();
+        highestReplPositionPushed.compute(producerName, (__, lastPushed) -> {
```
poorbarcode: Almost all the changes were made around the producer's name:

**Before**
- Cached `{producerName} + suffix`
- Stored `{producerName} + suffix`

**After**
- Cached `{producerName}`
- Stored `{producerName} + suffix`

I consider the new solution to be better, which improves the performance.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorDeduplicationTest.java
```diff
@@ -272,6 +273,107 @@ protected void handleSendReceipt(CommandSendReceipt sendReceipt) {
...
+            });
+
+            // Reload target before a dedup snapshot has to cover the latest replicated source positions.
+            admin2.topics().unload(topicName);
+            pulsar2.getBrokerService().getTopic(topicName, false).join();
+
+            // Reconnect the source replicator. It replays from the old replication cursor and target must deduplicate.
+            stuckSendReceipt.set(false);
+            Awaitility.await().atMost(Duration.ofSeconds(30)).untilAsserted(() -> {
+                assertNotNull(replicator.producer);
```
poorbarcode: Since you called `topics unload`, the object `replicator` will be rebuilt; the verification here is meaningless

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorDeduplicationTest.java
```diff
@@ -272,6 +273,107 @@ protected void handleSendReceipt(CommandSendReceipt sendReceipt) {
...
+                ClientCnx clientCnx = replicator.producer.getClientCnx();
+                assertNotNull(clientCnx);
+                replicatorClientCnx.set(clientCnx);
+            });
+            replicatorClientCnx.get().ctx().channel().close();
+
+            Awaitility.await().atMost(Duration.ofSeconds(30)).untilAsserted(() -> {
+                assertEquals(replicator.getCursor().getNumberOfEntriesInBacklog(true), 0);
```
poorbarcode: Since you called `topics unload`, the object replicator will be rebuilt; the verification here is meaningless

poorbarcode (commented):
> The geo V2 watermark is not recovered from replayed target entries
> When deduplication is enabled or a target topic is loaded, the broker restores dedup state from the pulsar.dedup cursor snapshot and then replays entries after that snapshot.
> The existing replay path restores normal producer sequence ids from message metadata, but it does not rebuild geo V2 watermark state from replicated messages. If the target topic is unloaded after replicated messages are persisted but before the latest dedup snapshot includes their source positions, the in-memory geo watermark is lost.
> If the source replicator later reconnects and replays from its replication cursor, the target broker can fail to identify those source entries as duplicates and persist them again.

This is a bug, and was fixed by the current PR

> The geo V2 watermark must be handled atomically
> A geo V2 watermark is a pair: source ledger id and source entry id. Treating the two values as independent mutable states can expose a mixed pair to duplicate checks or snapshots, for example a new ledger id combined with an old entry id.
> Such a mixed pair does not represent a real source position. If it is snapshotted and later recovered, the target can either move the watermark too far forward and drop valid replayed source entries, or move it backward and allow duplicates.

The situation will never happen, and the current PR did not fixed such issue

> Replay recovery must never move the watermark backward
> Source-position replay is expected during reconnects and failovers. When rebuilding the watermark from target entries, older source positions can appear after newer positions in recovery windows. The recovered watermark must therefore keep the maximum source position seen so far instead of blindly overwriting the current value.

The improvement is meaningless; once such an issue occurs, we need to fix the place that makes it "older source positions can appear after newer positions ", rather than ordering when executing `replayCursor`

> The geo V2 watermark can be removed by normal producer inactivity cleanup
> Normal producer dedup state can be purged after producer inactivity. Geo V2 watermark state has a different lifecycle: it is a source-position checkpoint for replication, not the lifecycle state of the replicator producer.
> The source replicator may disconnect, reconnect, or fail over and replay the same source entries from its replication cursor. If the target has purged the geo watermark because the replicator producer was inactive, those replayed source entries can be accepted and written again.

It is a real issue, but the current PR fixed the issue:
- let repl deduplication status will not be limited by the configuration `brokerDeduplicationMaxNumberOfProducers`
- Do not clear the repl deduplication status when purging inactive producers

void-ptr974:
Pushed `e823f0a01e8` with the review updates:

- Split geo/shadow replication watermark recovery from normal producer sequence recovery during replay.
- Added error logging for incomplete geo-replication watermark snapshot pairs.
- Removed the extra `ReplDedupCheck` helper and simplified `isDuplicateReplV2` local state handling.
- Removed the watermark ordering/max merge logic.
- Changed snapshot normal-producer counting to use a plain `int` loop.
- Updated inactive geo-replication cleanup logging to match the actual behavior.
- Updated the unload-before-snapshot test to re-fetch the current source topic/replicator and wait for target deduplication to become `Enabled`.
- Avoided adding new reflection in the new unit tests by using testing accessors.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. The `NumberFormatException` branch now logs immediately and returns, avoiding the duplicate fall-through log.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. Added error logging for incomplete geo-replication watermark snapshot pairs when either `_LID` or `_EID` is missing.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. Replay now restores marker/normal messages into `highestSequenced*` and replicated geo/shadow messages into `highestReplPosition*`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. Changed this to a plain `int` with a normal loop.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. Removed the watermark ordering/max merge path and now records the recovered watermark directly.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. Removed the max/ordering merge logic and now stores the recovered watermark directly.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. Removed `ReplDedupCheck` and replaced it with method-local variables in `isDuplicateReplV2`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Adjusted. The hot path now uses `long[]` for parsed source positions; `ReplSourcePosition` remains only as the map value for the persisted watermark pair.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorDeduplicationTest.java
void-ptr974: Done. The test now re-fetches the current source topic and replicator after reload before checking the connection.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Done. The in-memory watermark cache now uses the base producer name, while snapshot keys keep the `_LID` / `_EID` suffixes.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorDeduplicationTest.java
void-ptr974: Done. The backlog check now re-fetches the current source topic and replicator before asserting.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
```diff
@@ -262,24 +278,54 @@ private CompletableFuture<Void> replayCursor(ManagedCursor cursor) {
...
-            final var producerName = metadata.getProducerName();
-            final var sequenceId = Math.max(metadata.getHighestSequenceId(), metadata.getSequenceId());
-            highestSequencedPushed.put(producerName, sequenceId);
-            highestSequencedPersisted.put(producerName, sequenceId);
-            producerRemoved(producerName);
+            if (metadata.hasMarkerType() || getReplProducerName(metadata) == null) {
+                recoverProducerSequenceIdFromMetadata(metadata);
+            } else {
+                // Rebuild replication watermarks from entries written after the last dedup snapshot.
+                recoverReplWatermarkFromMetadata(metadata);
```
lhotari: **[BUG] Preserve V1 producer sequence state during replay**

Messages from an older source broker have `replicatedFrom` but do not have `__MSG_PROP_REPL_SOURCE_POSITION`. They now enter this branch because `getReplProducerName` returns a name, and `recoverReplWatermarkFromMetadata` then does nothing when the property is absent. This skips the original producer sequence recovery that V1 still needs: `isDuplicateReplV1` uses the original producer name and `recordMessagePersistedNormal` records that producer’s sequence state ([`MessageDeduplication.java:369-382`](https://github.com/apache/pulsar/blob/5ab9479514e4a0cbf1fd379e1e184acf6e903280/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java#L369-L382) and [`MessageDeduplication.java:648-660`](https://github.com/apache/pulsar/blob/5ab9479514e4a0cbf1fd379e1e184acf6e903280/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java#L648-L660)).

If the target persists V1 messages after the last dedup snapshot, unloads, and the older source reconnects and replays them, the recovered normal-producer watermark is missing or stale and those messages are persisted again. Keep the normal sequence recovery for V1 entries, including entries without a V2 position, and add an unload-before-snapshot regression with a V1 source.

lhotari (changes requested):
The V2 watermark recovery leaves a compatibility regression in the replay path for V1 replication. Please preserve recovery of the original producer sequence state for those messages.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java
void-ptr974: Fixed in 7cab7835ed4. Replay now restores a V2 watermark only when a non-marker replicated entry has a valid source-position property; otherwise it falls back to the original producer sequence recovery required by V1. I also added a persisted-entry unload/reload regression in a2c8baee6e1: it writes a V1 replicated entry after the snapshot, unloads and reloads the topic, verifies recovery under the original producer name, and verifies that replay is classified as a duplicate. The mixed V1/V2/invalid replay cases and the additional V2 state-transition coverage are included as well. The full MessageDuplicationTest class and broker checkstyle passed locally. Could you please take another look?

PR #25863, opened 2026-05-23
[improve][broker] PIP-473 P5.1: metadata-driven transaction coordinator (NEW_TXN / END_TXN)

merlimat:
## Summary

Adds the PIP-473 metadata-driven v5 transaction coordinator — the broker-side service that serves `NEW_TXN` / `END_TXN` against the metadata-store transaction layout (P2's `TxnMetadataStore`) instead of the legacy `TransactionMetadataStoreService`.

Gated behind a new `transactionCoordinatorScalableTopicsEnabled` flag (default **off**). When on, `ServerCnx` routes the TC wire commands to `TransactionCoordinatorV5`; when off, behavior is unchanged. The flag is meant to be flipped together with the scalable-topic transaction buffer and pending-ack store providers (a later phase).

### Leader election

Per-partition coordinator, same as the legacy TC: a broker runs the v5 TC for partition `N` iff it owns partition `N` of `transaction_coordinator_assign`. `handleClientConnect` mirrors the legacy ownership check, so the client-side discovery surface is unchanged. (A pure metadata-store election is a later phase; reusing the assign-topic keeps this PR's scope tight.)

### Wire commands

| Command | Behavior |
|---|---|
| `TC_CLIENT_CONNECT` | ownership check for the tcId's assign partition |
| `NEW_TXN` | create `/txn/id/<tcId>_<seq>` header in `OPEN` |
| `END_TXN` | CAS header to `COMMITTED`/`ABORTED`, fan out participant events |
| `ADD_PARTITION_TO_TXN`, `ADD_SUBSCRIPTION_TO_TXN` | no-op |

`ADD_PARTITION` / `ADD_SUBSCRIPTION` are no-ops per the PIP: v5 participants advertise themselves by writing `/txn/op` records when they apply ops, so the pre-registration step is unnecessary.

### txnId generation — no reuse

`leastSigBits` is drawn from a per-tcId monotonic counter at `/txn/tc-seq/<tcId>` (`TcSequence`, CAS-incremented with retry on `BadVersionException`). Monotonic-per-tcId ⟹ txnIds are never reused — the participant-side aborted set is keyed by txnId, and reuse would corrupt it.

### endTransaction fan-out

`END_TXN` CAS-updates the header to the terminal state, then enumerates the txn's participants via the new `idx:ops-by-txn` secondary index on `/txn/op` (so it doesn't scan the whole namespace), and publishes one segment-event per affected segment + one subscription-event per affected `(segment, subscription)` pair. Idempotent on retry — a header already in the requested terminal state short-circuits.

## Test plan

- [x] `pulsar-broker:test --tests TransactionCoordinatorV5Test` — 8 cases: sequential txnId per tcId, commit/abort fan-out, idempotent retry, mismatched-action failure, unknown-txn failure, add* no-ops, per-tcId scoping.
- [x] `pulsar-broker:test --tests TxnMetadataStoreTest` / `MetadataTransactionBufferTest` / `MetadataPendingAckStoreTest` — green after the `idx:ops-by-txn` index addition.
- [x] Checkstyle clean (main + test).

## Deferred / follow-ups

- **Timeout sweep + GC sweep.** This PR is happy-path `newTxn` / `endTxn` only.
- **Pure metadata-store leader election** (replacing the assign-topic reuse).

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
@@ -3348,6 +3348,21 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
             return;
         }
 
+        if (service.getPulsar().getConfig().isTransactionCoordinatorV5Enabled()) {
+            service.pulsar().getTransactionCoordinatorV5().handleClientConnect(tcId)
+                    .whenComplete((__, e) -> {
+                        if (e == null) {
+                            commandSender.sendTcClientConnectResponse(requestId);
+                        } else {
+                            log.error().attr("requestId", requestId).attr("tcId", tcId).exception(e)
+                                    .log("v5 TC client connect failed");
+                            commandSender.sendTcClientConnectResponse(requestId,
+                                    BrokerServiceException.getClientErrorCode(e), e.getMessage());
+                        }
+                    });
+            return;
+        }
+
         TransactionMetadataStoreService transactionMetadataStoreService =
                 service.pulsar().getTransactionMetadataStoreService();
 
@@ -3414,6 +3429,22 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
             return;
         }
 
+        if (service.getPulsar().getConfig().isTransactionCoordinatorV5Enabled()) {
+            final String v5Owner = getPrincipal();
+            service.pulsar().getTransactionCoordinatorV5()
+                    .newTransaction(tcId, command.getTxnTtlSeconds() * 1000L, v5Owner)
+                    .whenComplete((txnId, e) -> {
+                        if (e == null) {
+                            commandSender.sendNewTxnResponse(requestId, txnId, tcId.getId());
+                        } else {
+                            Throwable cause = handleTxnException(e, BaseCommand.Type.NEW_TXN.name(), requestId);
+                            commandSender.sendNewTxnErrorResponse(requestId, tcId.getId(),
+                                    BrokerServiceException.getClientErrorCode(cause), cause.getMessage());
+                        }
+                    });
+            return;
+        }
+
         TransactionMetadataStoreService transactionMetadataStoreService =
                 service.pulsar().getTransactionMetadataStoreService();
         final String owner = getPrincipal();
@@ -3462,6 +3493,14 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
                         .log("Receive add published partition to txn request " + "from with txnId, topic"));
 
         if (!checkTransactionEnableAndSendError(requestId)) {
+            return;
+        }
+
+        if (service.getPulsar().getConfig().isTransactionCoordinatorV5Enabled()) {
+            // v5: TC doesn't need pre-registration — participants advertise themselves by writing
+            // /txn/op records when they actually apply ops.
+            writeAndFlush(Commands.newAddPartitionToTxnResponse(requestId,
+                    txnID.getLeastSigBits(), txnID.getMostSigBits()));
             return;
         }
 
@@ -3522,6 +3561,20 @@ public class ServerCnx extends PulsarHandler implements TransportCnx {
         final TransactionCoordinatorID tcId = TransactionCoordinatorID.get(command.getTxnidMostBits());
 
         if (!checkTransactionEnableAndSendError(requestId)) {
+            return;
+        }
+
+        if (service.getPulsar().getConfig().isTransactionCoordinatorV5Enabled()) {
+            service.pulsar().getTransactionCoordinatorV5().endTransaction(txnID, txnAction)
+                    .whenComplete((__, e) -> {
+                        if (e == null) {
+                            commandSender.sendEndTxnResponse(requestId, txnID, txnAction);
+                        } else {
+                            Throwable cause = handleTxnException(e, BaseCommand.Type.END_TXN.name(), requestId);
+                            commandSender.sendEndTxnErrorResponse(requestId, txnID,
+                                    BrokerServiceException.getClientErrorCode(cause), cause.getMessage());
+                        }
+                    });
             return;
         }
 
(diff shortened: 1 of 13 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
```diff
@@ -0,0 +1,238 @@
...
+                String subscription = packed.substring(nul + 1);
+                publishes[i++] = txnStore.publishSubscriptionEvent(segment, subscription, event);
+            }
+            return CompletableFuture.allOf(publishes);
+        });
+    }
+
+    // ---- Small helper instead of pulling in a third-party concurrent set --------------------
+
+    private static final class ConcurrentHashSet {
```
lhotari: could this be replaced by the usage of `java.util.concurrent.ConcurrentHashMap.newKeySet()`

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
```diff
@@ -0,0 +1,238 @@
...
+        Set<String> writeSegments = ConcurrentHashSet.create();
+        Set<String> ackParticipants = ConcurrentHashSet.create();
+        return txnStore.listOpsByTxn(txnIdKey, new ScanConsumer() {
+            @Override
+            public void onNext(GetResult r) {
+                TxnOp op = TxnMetadataStore.fromJson(r.getValue(), TxnOp.class);
+                if (op.getKind() == TxnOpKind.WRITE) {
+                    writeSegments.add(op.getSegment());
+                } else if (op.getKind() == TxnOpKind.ACK && op.getSubscription() != null) {
+                    ackParticipants.add(op.getSegment() + "\0" + op.getSubscription());
```
lhotari: using a Java record here would be a better solution that using a java.lang.String

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
```diff
@@ -0,0 +1,238 @@
...
+                    writeSegments.size() + ackParticipants.size()];
+            int i = 0;
+            for (String segment : writeSegments) {
+                publishes[i++] = txnStore.publishSegmentEvent(segment, event);
+            }
+            for (String packed : ackParticipants) {
+                int nul = packed.indexOf('\0');
+                String segment = packed.substring(0, nul);
+                String subscription = packed.substring(nul + 1);
+                publishes[i++] = txnStore.publishSubscriptionEvent(segment, subscription, event);
```
lhotari: parsing could be avoided with the use of a Java record

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
```diff
@@ -3525,6 +3564,20 @@ protected void handleEndTxn(CommandEndTxn command) {
             return;
         }
 
+        if (service.getPulsar().getConfig().isTransactionCoordinatorScalableTopicsEnabled()) {
+            service.pulsar().getTransactionCoordinatorV5().endTransaction(txnID, txnAction)
```
lhotari: Claude Code finding:
The legacy handleEndTxn calls verifyTxnOwnership(txnID) (which checks the principal owns the txn, falling back to a superuser check) before finalizing. The v5 early-return path calls
getTransactionCoordinatorV5().endTransaction(txnID, txnAction) with no principal/ownership check at all. The same applies to the ADD_PARTITION/ADD_SUBSCRIPTION no-ops (legacy verifies
ownership there too).

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
```diff
@@ -0,0 +1,238 @@
...
+                return FutureUtil.failedFuture(
+                        new CoordinatorException.InvalidTxnStatusException(
+                                "Transaction " + txnId + " is " + current.getState()
+                                        + ", cannot transition to " + newState));
+            }
+            TxnHeader updated = new TxnHeader(newState, current.getTimeout(),
+                    current.getCreatedAt(), Instant.now());
+            return txnStore.updateHeader(txnIdKey, updated, v.version())
+                    .thenCompose(stat -> fanOutEvents(txnId, txnIdKey, newState));
+        });
```
lhotari, line 194: It seems that the transaction state is updated before publishing events.
The PR description mentions " Idempotent on retry — a header already in the requested terminal state short-circuits".
If `fanOutEvents` fails, there there a way to retry that later? could the transaction get into a state where the transaction has been updated to a terminal state, but events don't get published?

lhotari: there's a comment "participants tolerate missing events via the header re-read in their reconcile path.", however just wondering how would the reconcile path detect that events are missing if header is in terminal state.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
merlimat: Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
merlimat: Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
merlimat: fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java
merlimat: Fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
merlimat: Fixed

lhotari (approved):
LGTM

PR #25868, opened 2026-05-25
[improve][fn] make built-in functions reload incremental

### LINKED ISSUE #25791: [improve][fn] Apply incremental reload to FunctionsManager (mirror connector reload）
### Search before reporting

### Motivation

Built-in connector reload was fixed in #25773 to be incremental: unchanged NARs keep the same Connector instance and only evicted connectors are closed. Built-in function reload in FunctionsManager still uses the old replace-all + close-all pattern, which has the same bug class and should be aligned with the connector implementation.

### Solution

Mirror the incremental reload approach from #25773 

### Alternatives

_No response_

### Anything else?

_No response_

pratt4:
Fixes #25791.

`FunctionsManager.reloadFunctions()` used to recreate every built-in function archive on reload, then close all previous `FunctionArchive` instances. That meant unchanged NARs were still reopened and their existing classloader/package state was discarded.

This mirrors the incremental reload behavior recently added for connectors.

### Changes

- Track function archive MD5 in `FunctionArchive`
- Add `FunctionUtils.reloadFunctions()`
- Reuse existing `FunctionArchive` instances when the archive path and MD5 are unchanged
- Close only changed or removed function archives
- Update `FunctionsManager.reloadFunctions()` to use the incremental reload result
- Add regression tests for unchanged, changed, removed, and missing-directory reload cases

### DIFF
--- a/pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/functions/FunctionUtils.java
+++ b/pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/functions/FunctionUtils.java
@@ -25,12 +25,16 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.HexFormat;
+import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import lombok.CustomLog;
 import lombok.experimental.UtilityClass;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.pulsar.common.functions.FunctionDefinition;
+import org.apache.pulsar.common.nar.FileUtils;
 import org.apache.pulsar.common.nar.NarClassLoader;
 import org.apache.pulsar.common.util.ObjectMapperFactory;
 import org.zeroturnaround.zip.ZipUtil;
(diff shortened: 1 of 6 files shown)


pratt4:
Hi @lhotari @Dream95 
can you please review this pr whenever you get time ??

pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/functions/FunctionUtils.java
```diff
@@ -42,6 +46,17 @@ public class FunctionUtils {
...
 
+    /**
+     * Computes MD5 digest of a file as lower-case hex (for function archive identity on reload).
+     */
+    public static String computeArchiveMd5Hex(Path path) throws IOException {
+        return calculateMd5Hex(path.toAbsolutePath().normalize().toFile());
+    }
+
+    private static String calculateMd5Hex(File file) throws IOException {
+        return HexFormat.of().formatHex(FileUtils.calculateMd5sum(file));
```
Dream95, line 57: Since this mirrors ConnectorUtils.reloadConnectors, extracting a shared helper maybe better

lhotari: The common methods like this could be extracted to `SharedUtils` which would go into `org.apache.pulsar.functions.utils` package. Adding more utilities to `pulsar-common` should be avoided, so this would go into `pulsar-function-utils`.
Refactoring to avoid all existing duplication would require broader changes. That's why it's better to leave it out of the scope of this PR.

lhotari:
I renamed the title of this PR. It's better not to reference issues in the PR title, but instead describe in imperative form what the PR does ([there's a similar best practice for commit messages](https://cbea.ms/git-commit/#imperative)).

pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/functions/FunctionUtils.java
pratt4: @lhotari 
should i keep this PR focused on the functions incremental reload change and leave the shared utility extraction out of scope??

If needed, I can follow up with a separate issue/PR to extract the common helper into `org.apache.pulsar.functions.utils`.

Apart from that, is there any other changes ??

 thanks !

Dream95 (approved):

lhotari (approved):

pulsar-functions/utils/src/main/java/org/apache/pulsar/functions/utils/functions/FunctionUtils.java
lhotari: this PR is fine

PR #25873, opened 2026-05-27
[pip][ml] PIP-480: Add `readEntries` method to ManagedLedger

dao-jun:
### Motivation

Provide a cursorless read path for downstream projects like KoP.

A caller should be able to pass a `Position` and a maximum entry count, and get the entries that are readable now.
The call must not open a `ManagedCursor`, update cursor metadata, or change acknowledgement state.
The downstream protocol remains responsible for its own offsets.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


pip/pip-480.md
```diff
@@ -0,0 +1,135 @@
...
+ * <p>The start position is inclusive when it points to an existing entry. If it points to a non-existing entry,
+ * the read starts from the next valid entry in ledger order. {@link PositionFactory#EARLIEST} starts from the first
+ * available entry, while {@link PositionFactory#LATEST} starts from the position after the current last confirmed
+ * entry. This method does not wait for future writes and will complete with fewer entries, or an empty list, when
+ * there are not enough currently readable entries.
+ *
+ * <p>The returned entries are raw ledger entries and are not filtered by any cursor acknowledgement state. Callers
+ * are responsible for releasing returned entries.
+ */
+CompletableFuture<List<Entry>> asyncReadEntries(Position startPosition, int numberOfEntries);
```
void-ptr974: Could we also define the maxSize semantics here, especially how it interacts with numberOfEntries?

dao-jun: I thought about this, but later I realized that in Pulsar, how many Entries to read from ML is calculated in the upper layer. Therefore, in order to maintain synchronization, this variable was not added

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+
+* It uses `startReadOperationOnLedger` to move a read from a deleted empty ledger to the next available ledger.
+* It uses the managed-ledger `lastConfirmedEntry` as the upper visibility bound for the current ledger.
+* It uses the BookKeeper `lastAddConfirmed` value for closed ledgers.
+* The per-entry position check treats `maxPosition` as inclusive: a read position equal to `maxPosition` is
+  allowed; a read position strictly after `maxPosition` stops the operation. This constraint overrides any
+  ledger-level upper bound — an explicit `maxPosition` before `lastConfirmedEntry` ends the read early.
+* It reads at most the remaining requested count from each ledger.
+* It completes immediately when it reaches the current ledger end. It does not wait for future writes.
+* If a later ledger read fails after some entries have already been collected, the operation completes with the partial
```
nodece, line 124: Could we clarify the failure semantics?

If a ledger read fails after some entries have been collected, should the future complete exceptionally rather than return a partial result? Otherwise, the caller cannot distinguish a partial result from a normal end-of-read.

Also, on a read failure, do we stop the operation or skip the failed ledger and continue reading?

dao-jun: 1.  return partial
2. stop operation

dao-jun: No need to clarify the failure semantics here, it will in the impl PR

nodece: Returning a partial result seems to hide the read failure from the caller. Since the API returns `CompletableFuture<List<Entry>>`, how can the caller distinguish a partial result caused by a ledger read failure from a normal completion that simply reaches the end of available entries?

dao-jun: @nodece 
This is a common pattern in Pulsar.
You can see OpReadEntry.java
image

BatchedReadEntryProcessor in Bookkeeper
image


IO is expensive for Pulsar, so we better not waste it.

nodece: Got.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+`ManagedCursor` abstraction. If KoP opens a cursor only to fetch entries, it creates extra managed-ledger metadata and
+then has to keep the cursor lifecycle aligned with Kafka offsets.
+
+`ManagedLedger` already exposes `asyncReadEntry(Position, ...)` for reading a single entry by position, but it does not
+provide a batch read primitive for this use case. Callers have to repeat the ledger traversal logic themselves: validate
+the start position, cross ledger boundaries, skip empty ledgers, and stop at the current last confirmed entry.
+
+# Motivation
+
+Provide a cursorless read path for downstream projects like KoP.
```
lhotari, line 20: There's currently a reason to have a cursor. For example, a cursor will prevent trimming while the cursor is active. Another important reason for the cursor is the managed ledger cache (a.k.a "broker cache"). 

It is possible to implement the ManagedLedger internally in a way where an external cursor wouldn't be required and it would be handled internally. However, that breaks the original abstraction which ManagedLedger & ManagedCursor provide.

> If KoP opens a cursor only to fetch entries, it creates extra managed-ledger metadata and
then has to keep the cursor lifecycle aligned with Kafka offsets.

For non-durable cursors "extra metadata" is cheap and only in memory. Does KoP use non-durable cursors?

> `ManagedLedger` already exposes `asyncReadEntry(Position, ...)` for reading a single entry by position, but it does not provide a batch read primitive for this use case.

From an abstraction perspective, this is a mistake and already breaks the abstract. I added the `asyncReadEntry` to `ManagedLedger` interface in #23311 since Pulsar already depended directly on `ManagedLedgerImpl` having this method. (#23311 was about having the possibility to implement a 3rd party `ManagedLedger` implementation such as SN Ursa). However, making the abstraction worse isn't a great approach.

For the batch read, there's also another problem that this introduces when there's no "waiting mode". Without a waiting mode, there would be unnecessary polling when cursors are consuming the tail. 
One of the existing problems in the ManagedCursor API is that there's no streaming API. It would be more useful to start a "reactive" streaming read with a possibility to cancel. I believe that such an API would be useful for batching reads. With a batching read, there could be a batch deadline when the cursor is in tailing mode. This would resolve unnecessary tight loops with tailing reads.
This challenge exists also for moving to batch reads from bookkeeper for Pulsar. Instead of waiting for the requested number of entries, the read from ManagedCursor should support returning less entries when no more are available. 

These were the thoughts based on some previous experiences of the bookkeeper batch read and integrating it efficiently to Pulsar. I didn't check all details. 

In summary: I believe that the correct path would be to ensure that the ManagedLedger abstraction is involved in a reasonable direction.
One possible solution would be to have a separate interface for random reads which could be created by calling a method in ManagedLedger. We could also migrate the existing `asyncReadEntry` entry to that and also add a similar batch read method as long as the details about deadlines for batching of tailing reads is addressed so that the usage doesn't result in inefficiencies.
Could you revisit the proposal in that direction? (a separate random reads interface, which also documents the abstraction and the various requirements for implementations if such exist (for example related to retention/trimming while a read is performed))

lhotari (changes requested):

pip/pip-480.md
lhotari: For Pulsar 5.0, it would be fine to break ManagedLedger in this aspect since there aren't many external implementors. 

If there's objections about breaking it, a default random read implementation could delegate to existing asyncReadEntry. For the Pulsar ManagedLedgerImpl, the implementation would be moved and there could be a delegation from asyncReadEntry to "random read"'s asyncReadEntry. This solution would be possible for backporting. I'd prefer skipping this for Pulsar 5.0 so that we could clean up the abstraction for this part.

dao-jun: @lhotari 

KoP's read path uses non-durable ManagedCursor only because ManagedLedger lacks a positional batch-read API. Every fetch pops a per-offset cursor from a cache, calls asyncReadEntries, then immediately fires asyncMarkDelete to undo the cursor's own backlog-holding side effect. 
Cursors are keyed by Kafka offset, churned through three concurrent maps, with explicit handling for races ("A race - same cursor already cached").

> For the batch read, there's also another problem that this introduces when there's no "waiting mode". Without a waiting mode, there would be unnecessary polling when cursors are consuming the tail.

KoP has its own `DelayedFetchPurgatory` which is as same as Kafka, `waiting mode` handled by it.

> There's currently a reason to have a cursor. For example, a cursor will prevent trimming while the cursor is active.

For this part, a NonDurableCursor does not prevent the ManagedLedger from being trimmed; only a DurableCursor does. KoP never creates DurableCursors — its data cleanup follows the same approach as Kafka. 

In summary: for KoP, it only needs a method to read entries from ManagedLedger, cursor is unnecessary.

> One of the existing problems in the ManagedCursor API is that there's no streaming API. It would be more useful to start a "reactive" streaming read with a possibility to cancel. I believe that such an API would be useful for batching reads

I fully agree with this part, but at this stage, I don't get ready to do it. It should be a further improvement, maybe I can do it when I available.

I wish the PIP can be approved, after solving our current problems, we can explore long-term development.

Thanks!

void-ptr974:
I looked a bit more at the KoP read path and wanted to share a possible direction for discussion, without trying to take over the proposal.

My current understanding is that KoP uses `ManagedCursor` mainly as a positional batch-read mechanism rather than for cursor-owned consumption progress. It maps Kafka offsets to ManagedLedger positions, creates non-durable cursors to read with `maxEntries` and `maxBytes`, and then mark-deletes the temporary cursor to avoid backlog side effects. KoP also already has its own delayed fetch mechanism for tail reads.

So I wonder whether PIP-480 could be framed as a general cursorless positional reader, as an alternative to extending `ManagedLedger` directly with another read method.

The split could be:

- `ManagedCursor` remains for subscription-style consumption: ack state, mark-delete, backlog, retention interaction, cursor lifecycle, and read-or-wait.
- A positional reader would be for callers that own progress externally and only need currently readable raw entries by position.

The reader would still preserve the important capabilities needed by KoP and similar callers:

- caller-provided start position
- non-waiting read of currently readable entries
- `maxEntries` limit
- `maxSizeBytes` limit
- optional upper bound such as `maxPosition`
- cross-ledger reads
- raw `Entry` return, with caller-owned release

The main advantage is that these semantics get a dedicated home. `ManagedLedger` would only need a small factory-style method to create the reader, while `ManagedCursor` can keep its current behavior in v1.

Implementation-wise, this can stay close to the current PR by reusing `ManagedLedgerImpl` internals such as ledger-handle lookup, `lastConfirmedEntry` visibility, entry cache reads, read timeout handling, and cross-ledger traversal. The new operation would simply stay detached from cursor state, so it would not update mark-delete, backlog, waiting cursors, or cursor metadata.

This also keeps the implementation scope controlled: dispatcher and cursor behavior do not need to change for v1, and future additions such as deadline, cancellation, streaming, or richer result reasons can evolve on the reader interface without further expanding `ManagedLedger` itself.

One API question worth discussing early is whether the result should be only `List<Entry>`, or whether it should expose enough information to distinguish normal tail/no-data from an unavailable or trimmed start position.

This would keep the KoP use case covered while also making the abstraction useful for broader Pulsar-side positional reads such as admin/debug reads, read-only inspection, recovery/replay, and scan-style tasks.

dao-jun:
@void-ptr974 
KoP is just a small part, my true intention is https://github.com/apache/pulsar/pull/25891
I will redesign the API change when Im available

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+
+Protocol handlers such as KoP do not always want that state. Kafka offsets are maintained outside Pulsar's
+`ManagedCursor` abstraction. If KoP opens a cursor only to fetch entries, it creates extra managed-ledger metadata and
+then has to keep the cursor lifecycle aligned with Kafka offsets.
+
+`ManagedLedger` already exposes `asyncReadEntry(Position, ...)` for reading a single entry by position, but it does not
+provide a batch read primitive for this use case. Callers have to repeat the ledger traversal logic themselves: validate
+the start position, cross ledger boundaries, skip empty ledgers, and stop at the current last confirmed entry.
+
+# Motivation
```
lhotari, line 18: **[DESIGN] the stated motivation is no longer the actual driver, and the scope does not cover it**

This section frames the proposal around KoP, but in the conversation you wrote that "KoP is just a small part, my true intention is https://github.com/apache/pulsar/pull/25891".

#25891 adds a `RandomReader` to the Pulsar client — a new public client API plus protocol buffer changes — while this document states "No broker binary protocol changes" ([`pip-480.md:131`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L131)). Those two things cannot both be true of the same feature, and a PIP is voted on what it says rather than on what the author intends next.

Either is fine, but the document should pick one:

- broaden PIP-480 to the end-to-end reader (managed-ledger API, protocol command, client API), or
- state explicitly that PIP-480 is only the managed-ledger building block and that the client and protocol surface will be proposed separately.

If it is the second, the ML API shape still has to be justified against both callers, not just KoP. Designing it against a single caller that already has its own offset store and its own `DelayedFetchPurgatory` is exactly how a caller's current convenience ends up baked into a `Stable` interface — a Pulsar client reader has neither of those and will need answers to the waiting and error-signalling questions below.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+ * <p>The start position is inclusive when it points to an existing entry. If it points to a non-existing entry,
+ * the read starts from the next valid entry in ledger order. {@link PositionFactory#EARLIEST} starts from the first
+ * available entry, while {@link PositionFactory#LATEST} starts from the position after the current last confirmed
+ * entry. This method does not wait for future writes and will complete with fewer entries, or an empty list, when
+ * there are not enough currently readable entries.
+ *
+ * <p>The returned entries are raw ledger entries and are not filtered by any cursor acknowledgement state. Callers
+ * are responsible for releasing returned entries.
+ */
+CompletableFuture<List<Entry>> asyncReadEntries(Position startPosition, int numberOfEntries, Position maxPosition);
```
lhotari, line 96: **[DESIGN] the read is bounded by entry count but offers no byte-limit parameter**

The existing APIs support optional byte limits alongside the entry count:

- `ManagedCursor.asyncReadEntries(int numberOfEntriesToRead, long maxSizeBytes, ReadEntriesCallback, Object, Position)` — [`ManagedCursor.java:160-170`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java#L160-L170)
- `ManagedCursor.applyMaxSizeCap(int maxEntries, long maxSizeBytes)` — [`ManagedCursor.java:925`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java#L925)
- `ManagedLedgerUtils.readEntries(ManagedCursor, int numberOfEntriesToRead, long maxBytes, Position maxPosition)` — [`ManagedLedgerUtils.java:67-81`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java#L67-L81), which already returns the same `CompletableFuture<List<Entry>>` shape

There are also unbounded overloads ([`ManagedCursor.java:156-157`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java#L156-L157), [`ManagedLedgerUtils.java:62-64`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java#L62-L64)), and `applyMaxSizeCap` estimates an entry count from the byte budget rather than enforcing a strict byte ceiling ([`ManagedCursorImpl.java:4096-4102`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L4096-L4102)). The question is why the new API omits the option altogether.

Entry count on its own does not bound memory use: the total bytes depend on entry sizes. The broker also exposes `dispatcherMaxReadSizeBytes` ([`ServiceConfiguration.java:1343-1348`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java#L1343-L1348)). These entries retain reference-counted buffers until released ([`EntryImpl.java:118-126`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryImpl.java#L118-L126), [`EntryImpl.java:272-285`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryImpl.java#L272-L285)).

`ManagedLedger` is `@InterfaceStability.Stable` ([`ManagedLedger.java:64-65`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java#L64-L65)), which is another reason to put the read contract on `ManagedReader`. I would retain an optional byte limit on `asyncRead`; the builder can still initially expose only `createAsync()`. For the read limit, `long maxSizeBytes` is already a familiar shape — `NO_MAX_SIZE_LIMIT` already exists as an opt-out ([`ManagedLedgerUtils.java:44`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java#L44)).

Related, and worth a sentence in the document either way: is the caller responsible for limiting concurrent reads and their combined memory use? A per-call limit alone does not bound the aggregate. The cursor's `readOperationCompleted` bookkeeping coordinates pending reads with mark-delete flushing ([`ManagedCursorImpl.java:3733-3745`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L3733-L3745)); it is not itself a concurrency limit.

pip/pip-480.md
lhotari, line 124: **[DESIGN] `List<Entry>` collapses three different outcomes that a cursorless caller cannot tell apart**

A short list can mean three different things under this contract, and the caller cannot distinguish them:

1. the read reached the current readable end (normal tail);
2. a later-ledger read failed after some entries were collected, so the batch was truncated and the exception was dropped ([`pip-480.md:124-125`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L124-L125));
3. `startPosition` no longer exists and the read silently advanced to the next valid entry ([`pip-480.md:68`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L68)).

Case 2 matches the cursor path: `OpReadEntry.internalReadEntriesFailed` returns accumulated entries when the batch is nonempty ([`OpReadEntry.java:137-144`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L137-L144)). The cursor advances its read position ([`OpReadEntry.java:104-113`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L104-L113), [`OpReadEntry.java:194-197`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L194-L197)), whereas a cursorless caller must derive the continuation from the entries. Both paths hide the error on that partial-result call. A subsequent read that fails before collecting any entries does report the error ([`OpReadEntry.java:174-190`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L174-L190)), but that does not identify the cause of the earlier short read. The contract should explain how callers distinguish an ordinary short read from this partial-error result, or explicitly state that they cannot.

Case 3 is worse for the caller this PIP names. A Kafka consumer asking for an offset that retention has deleted expects `OFFSET_OUT_OF_RANGE`; here it gets a successful read that quietly starts somewhere later, and unless it compares the first returned position against what it asked for, the gap is invisible.

This is the same question @void-ptr974 asked ("whether it should expose enough information to distinguish normal tail/no-data from an unavailable or trimmed start position") and I don't think the document can go to a vote without answering it. Either return a small result type that carries the outcome, or commit in the contract that the caller MUST compare the first returned position against `startPosition` and that a truncating error is deliberately dropped — but say which.

While you're in this area: `PositionFactory.LATEST` as `startPosition` means "after the last confirmed entry" ([`pip-480.md:66`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L66)), which combined with "a position after the current last confirmed entry returns an empty list" ([`pip-480.md:69`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L69)) makes `asyncReadEntries(LATEST, n)` always return empty. That is defensible, but the same sentinel is interpreted literally as `maxPosition` in the two-argument overload ([`pip-480.md:51-59`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L51-L59)), so it means two different things in one signature. Worth stating outright.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+* It uses the BookKeeper `lastAddConfirmed` value for closed ledgers.
+* The per-entry position check treats `maxPosition` as inclusive: a read position equal to `maxPosition` is
+  allowed; a read position strictly after `maxPosition` stops the operation. This constraint overrides any
+  ledger-level upper bound — an explicit `maxPosition` before `lastConfirmedEntry` ends the read early.
+* It reads at most the remaining requested count from each ledger.
+* It completes immediately when it reaches the current ledger end. It does not wait for future writes.
+* If a later ledger read fails after some entries have already been collected, the operation completes with the partial
+  result. If no entry has been collected, the operation fails.
+
+Returned `Entry` objects follow the existing managed-ledger ownership model. The caller must release them after use.
```
lhotari, line 127: **[DESIGN] cancellation and abandoned-result ownership need an explicit contract**

"The caller must release them after use" ([`pip-480.md:93-94`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L93-L94)) covers the normal success path. It does not explain who owns entries if the caller cancels or abandons the future while a bookie read is outstanding, or if a terminal failure prevents delivery of entries already collected.

These entries use reference-counted buffers ([`EntryImpl.java:118-126`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryImpl.java#L118-L126), [`EntryImpl.java:272-285`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/EntryImpl.java#L272-L285)), so a result that nobody releases can leak. `ManagedCursor.cancelPendingReadRequest` is a narrower precedent: its contract and implementation cancel a scheduled read-or-wait request ([`ManagedCursor.java:333-339`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java#L333-L339), [`ManagedCursorImpl.java:1214-1226`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java#L1214-L1226)). It does not establish cancellation of an outstanding bookie read or cleanup of this proposed future's partial results.

The PIP already says a later-ledger read failure returns the partial list ([`pip-480.md:124-125`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L124-L125)); those entries still belong to the recipient. Please also define what happens if cancellation prevents delivery, or if some other exceptional completion occurs after collection. Does cancellation stop the underlying work, and who releases entries that arrive after it?

A deadline or cancellation hook could make the lifecycle clearer. If cancellation is deliberately unsupported, the contract should say so and require the caller to arrange eventual release even after its request has timed out.

The closeable `ManagedReader` handle I propose gives this lifecycle an explicit owner: API consumers must close it when finished. Please define how `close()` interacts with outstanding reads and entries arriving afterwards, and distinguish closing the handle from releasing entries already delivered to the caller.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+
+* `startPosition == null`
+* `numberOfEntries <= 0`
+* `maxPosition == null`
+
+If the managed ledger is fenced or closed, the returned future fails with `ManagedLedgerFencedException`.
+
+### Read semantics
+
+The implementation adds an internal `OpReadEntries` operation with the following behavior:
```
lhotari, line 114: **[DESIGN] the proposed internal op forks ledger traversal from `OpReadEntry`, which your own open PR is concurrently reworking**

The behaviours listed under this heading overlap existing code: `OpReadEntry` normalizes the initial position ([`OpReadEntry.java:60-79`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L60-L79)) and continues across ledgers in `checkReadCompletion` ([`OpReadEntry.java:199-215`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L199-L215)). The helper `ManagedLedgerImpl.startReadOperationOnLedger` advances past a missing ledger ([`ManagedLedgerImpl.java:2680-2688`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2680-L2688)), and the read path caps current-ledger visibility at the managed ledger's confirmed entry, uses BookKeeper LAC for older ledgers, and applies `maxPosition` ([`ManagedLedgerImpl.java:2343-2357`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2343-L2357)). These are the traversal and visibility rules the proposed operation needs to share or reproduce.

Meanwhile #22560 (yours, still open, last touched 2026-07-27) is restructuring exactly that machinery — its diff touches `OpReadEntry.java`, `ManagedLedgerImpl.java`, `PositionFactory.java` and `MutablePositionImpl.java` to coalesce skip-split read ranges.

Two independent implementations of managed-ledger traversal, landing in the same release, is a maintenance cost we will pay for years, and it is the kind of thing a PIP is the right place to decide. The document should say whether `OpReadEntries` shares `OpReadEntry`'s traversal (with the cursor interaction factored out) or deliberately forks it, and why. If the answer is "share it", the ordering against #22560 is worth a line too.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+### Metadata format
+
+No metadata format changes.
+
+# Backward & Forward Compatibility
+
+This change is backward compatible for persisted data, BookKeeper ledgers, managed-ledger metadata, and Pulsar network
+protocols.
+
+`ManagedLedger` is a limited-private internal API. Implementations outside `ManagedLedgerImpl` must add an implementation
```
lhotari, line 142: **[COMPAT] preserve existing implementations with a default optional reader factory**

"Limited-private internal API" is half of it. `ManagedLedger` is annotated with *both* `@InterfaceAudience.LimitedPrivate` and `@InterfaceStability.Stable` ([`ManagedLedger.java:64-65`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java#L64-L65)), and it is the `Stable` half that carries the source-compatibility promise this change breaks. The document should name it, so that whoever votes knows what is being traded away.

PIP-424 (#24377) proposes a `default` method throwing `UnsupportedOperationException` on the same interface; this PIP currently adds an abstract read method. I previously said a break could be acceptable for 5.0, but I would now prefer to avoid it with `default Optional<ManagedReaderBuilder> managedReader() { return Optional.empty(); }`. Existing implementors can inherit the unsupported result, while those providing the capability return a builder whose `createAsync()` creates the closeable reader. Put `asyncRead` on `ManagedReader`, so the ledger interface gains only the factory and the read abstraction can evolve separately. Please describe that compatibility approach in this section; the abstract-method break need not be a prerequisite.

The template also asks for `## Upgrade`, `## Downgrade / Rollback`, and `## Pulsar Geo-Replication Upgrade & Downgrade/Rollback Considerations` under this heading ([`TEMPLATE.md:142-158`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/TEMPLATE.md#L142-L158)). "Nothing required" is a fine answer for all three, but they should be present.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+of `asyncReadEntries(Position, int, Position)` to compile against the new interface. The two-parameter overload is a
+`default` method and does not require an implementation.
+
+# Alternatives
+
+NONE
```
lhotari, line 148: **[PROCESS] `Alternatives: NONE` while three alternatives are live in this thread, and required template sections are missing**

There are at least three alternatives already on the table, and the mailing list will ask about them:

1. @void-ptr974's factory/reader proposal, developed into the optional `ManagedReaderBuilder` factory and closeable `ManagedReader` described in my review summary. This is my preferred direction: a default unsupported result preserves existing implementations, an initially optionless builder can gain creation options later, and read operations have their own abstraction and lifecycle.
2. Keeping the non-durable-cursor approach and fixing what is actually painful about it (the per-offset cursor cache and the mark-delete dance you described).
3. The `default` delegating to `asyncReadEntry` that I mentioned earlier in this thread, as the non-breaking variant.

Each only needs a couple of sentences plus why it was rejected — that is the whole purpose of the section, and "NONE" will read as "not considered".

Two other template sections are missing entirely from this PIP: `# Security Considerations` ([`TEMPLATE.md:131-137`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/TEMPLATE.md#L131-L137)) and `# Monitoring` ([`TEMPLATE.md:123-128`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/TEMPLATE.md#L123-L128)). For this proposal both are probably short, but Security Considerations is not trivially empty and is worth a real sentence: this API returns raw entries with no cursor and no subscription in the path, so whatever authorization exists happens entirely in the caller. Saying so — and saying that the method is not reachable from any REST or binary-protocol surface as proposed — is the answer, and it becomes materially more important if the scope grows to cover #25891.

Finally, `# Links` ([`pip-480.md:150-153`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/pip/pip-480.md#L150-L153)) has no mailing-list discussion thread. That link needs to exist before a vote can be called.

pip/pip-480.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+Invalid arguments fail the returned future with `IllegalArgumentException`:
+
+* `startPosition == null`
+* `numberOfEntries <= 0`
+* `maxPosition == null`
+
+If the managed ledger is fenced or closed, the returned future fails with `ManagedLedgerFencedException`.
```
lhotari, line 110: **[NIT] make the choice to report closed and fenced identically explicit**

`ManagedLedgerAlreadyClosedException` ([`ManagedLedgerException.java:126-129`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerException.java#L126-L129)) and `ManagedLedgerFencedException` ([`ManagedLedgerException.java:78-90`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedgerException.java#L78-L90)) are distinct existing types. However, the current batch-read path already reports both closed and fenced ledgers as `ManagedLedgerFencedException` ([`ManagedLedgerImpl.java:2083-2087`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L2083-L2087)), so the PIP is consistent with that path.

For this new API, is preserving that combined failure intentional, or should callers be able to distinguish the states? A short rationale would settle the contract before it becomes part of the interface.

lhotari (commented):
Thanks for the detailed KoP walkthrough — it made the cursorless case much easier to evaluate, and I checked the trimming point in the code: you are right, a non-durable cursor does not hold ledger trimming back. `addCursor` passes a null ordering position for non-durable cursors ([`ManagedLedgerImpl.java:744-753`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L744-L753)), so they never enter the heap that `getSlowestCursorPosition` reads ([`ManagedCursorContainerImpl.java:85-99`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorContainerImpl.java#L85-L99), [`ManagedCursorContainerImpl.java:192-199`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorContainerImpl.java#L192-L199)). `internalTrimLedgers` bases the cursor-dependent trimming boundary on durable cursors ([`ManagedLedgerImpl.java:3035-3052`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3035-L3052)). I withdraw that objection.

I'd like `ManagedLedger` to expose a reader factory, with the read operations on a separate, closeable handle. Something like this (the `asyncRead` arguments below follow the current proposal; the remaining questions about read behavior still need to be settled):

```java
interface ManagedLedger {
    default Optional<ManagedReaderBuilder> managedReader() {
        return Optional.empty();
    }
}

interface ManagedReaderBuilder {
    CompletableFuture<ManagedReader> createAsync();
}

interface ManagedReader extends AutoCloseable {
    // return an interface from asyncRead so that it's possible to extend it later
    // if a List<Entry> is returned, it's not possible to return other information without breaking the API
    interface ReadResult {
      List<Entry> entries();
    }

    CompletableFuture<ReadResult> asyncRead(
            Position startPosition, int numberOfEntries, Position maxPosition);

    @Override
    void close();
}
```

An implementation that supports this capability would return a builder; an unsupported implementation would inherit `Optional.empty()`. That lets existing `ManagedLedger` implementors remain unchanged. The builder can initially have no configuration methods — just `createAsync()` — while leaving room to add reader-creation options later without changing the factory signature.

For example, a future builder option could control whether reads populate the entry cache. Inserting a large scan into that cache can waste memory and evict useful entries when few concurrent readers would reuse them. A caller should be able to opt out of cache population for that workload; the option does not need to be part of the initial builder.

The `ManagedReader` is a handle that API consumers must close when finished. `asyncRead` belongs on that handle, and its contract should define what closing means for in-flight reads and resources. The ownership of entries already returned to the caller still needs to be documented separately.

This gives the read API its own abstraction and lifecycle without requiring another abstract read method on `ManagedLedger`. I have made mistakes in this interface myself when making it "generic"; those mistakes are not a reason to make the abstraction worse now. This is the direction I'd like the redesign to take.

The head has not moved since the changes-requested review, so this round is about the design document rather than a delta, and I've kept it to the things I think have to be settled before this can go to a vote on dev@. The short version:

- Since you wrote that "KoP is just a small part, my true intention is #25891", the motivation section no longer describes the actual driver. #25891 adds a public client API and proto changes; this PIP says there are no binary protocol changes.
- The API bounds entries but offers no byte-limit parameter, although the existing cursor and utility batch-read APIs support optional byte limits ([`ManagedCursor.java:160-170`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java#L160-L170), [`ManagedLedgerUtils.java:67-81`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java#L67-L81)).
- `List<Entry>` cannot distinguish "reached the end" from "a read failed and truncated the batch" from "your start position was trimmed and I silently moved forward" — and without a cursor the caller must derive its own next position from the returned entries and decide how to handle a skipped start or a truncated batch. This is the same question @void-ptr974 raised.
- The compatibility section omits that `ManagedLedger` is `@InterfaceStability.Stable`, and `Alternatives` says NONE while there are live alternatives in this very thread.

One thing the document gets right and that I want to call out: it specifies that invalid arguments fail the returned future rather than throwing synchronously, which is the contract we want for anything returning `CompletableFuture`. The semantics section is also more explicit than most PIPs — the gaps below are about what it does not yet commit to, not about sloppiness.

The earlier changes-requested review still stands; I'm not re-blocking, since you've already said you plan to redesign the API. Take this as the checklist I'd want the redesign to answer.

pip/pip-480.md
lhotari: Thanks for taking the time on this — the KoP walkthrough is the detail I was missing, and I went and checked the trimming point in the code rather than arguing from memory.

You are right about trimming, and I withdraw that objection. `addCursor` passes a null ordering position for non-durable cursors ([`ManagedLedgerImpl.java:744-753`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L744-L753)), so they never enter the heap that `getSlowestCursorPosition` reads ([`ManagedCursorContainerImpl.java:85-99`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorContainerImpl.java#L85-L99), [`ManagedCursorContainerImpl.java:192-199`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorContainerImpl.java#L192-L199)). `internalTrimLedgers` bases the cursor-dependent trimming boundary on durable cursors ([`ManagedLedgerImpl.java:3035-3052`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3035-L3052)). A non-durable cursor does not hold ledger trimming back.

There is one thing the cursor still does for you, though, and I think it belongs in the PIP. `getTheSlowestNonDurationReadPosition` walks the non-durable cursors' read positions ([`ManagedLedgerImpl.java:5136-5146`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L5136-L5146)). During trimming, eligible retained ledgers pass through `releaseReadHandleIfNoLongerRead` ([`ManagedLedgerImpl.java:3101-3115`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3101-L3115)), which invalidates a cached handle only if its ledger is below that minimum ([`ManagedLedgerImpl.java:3223-3230`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L3223-L3230)). A non-durable cursor can therefore keep a handle from being released by this particular path; it does not prevent all cache expiry or ledger deletion. The direct read method currently proposed contributes no such position, so this path may release a handle that a cursor would have kept cached, requiring it to be reopened for a later read. That may be acceptable, but the PIP should say whether it intends to preserve this behaviour.

On the waiting mode: your `DelayedFetchPurgatory` point is fair for KoP. It stops being an answer for the scope you have since described. If the real target is #25891 — a `RandomReader` on the Pulsar client — then the client has no purgatory, and "the caller polls" becomes the broker's problem, not the caller's. That is the main reason I'd rather see the API shaped against both callers before it lands, and it is the first of the inline notes I have left on this round.

On streaming: agreed that it can come later. I would put `asyncRead` on a `ManagedReader extends AutoCloseable` now, obtained through `Optional<ManagedReaderBuilder> managedReader()` on `ManagedLedger`. The default factory would return `Optional.empty()` for unsupported implementations, and the builder would initially expose only `CompletableFuture<ManagedReader> createAsync()`. Consumers must close the reader handle when done. That gives later reader options and operations a home without continuing to expand `ManagedLedger`; the review summary sketches the interfaces and a future cache-population option for scans. The PIP should distinguish this handle lifecycle from cursor acknowledgement state and backlog retention.

I've left the rest as inline comments on the document. None of them are large, but together they are what I think stands between this and a vote on dev@ — most of all the scope question, given your note that #25891 is the real target. No rush on my account; ping me when the redesign is ready and I will look promptly.

lhotari:
Thanks for digging into this — the factory-plus-reader split is the direction I would like to take. More concretely, `ManagedLedger` could expose a default `Optional<ManagedReaderBuilder> managedReader()` returning `Optional.empty()`, with supporting implementations supplying a builder. Initially the builder needs only `CompletableFuture<ManagedReader> createAsync()`. `ManagedReader` would extend `AutoCloseable`, require callers to close the handle, and own `asyncRead`. I have sketched this in the review summary, including a possible future builder option to avoid populating the entry cache for large scans. Please include this design in `Alternatives`, which currently says NONE.

Your last question is the one I'd most want answered before this goes to a vote: whether the result should be a bare `List<Entry>` or something that distinguishes a normal tail from an unavailable or trimmed start position. As the PIP stands, a short list can mean three different things — the readable end was reached, a later-ledger read failed and truncated the batch with the exception dropped, or the start position no longer existed and the read silently moved forward. The cursor path advances its own read position ([`OpReadEntry.java:104-113`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L104-L113), [`OpReadEntry.java:194-197`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/OpReadEntry.java#L194-L197)), though it also hides a partial-read error. A cursorless caller must derive its next position from the entries and decide how to handle a skipped start or a truncated batch. I've raised it inline as well.

The `maxSizeBytes` point in your list is worth underlining too — existing cursor and utility APIs support optional byte limits ([`ManagedCursor.java:160-170`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedCursor.java#L160-L170), [`ManagedLedgerUtils.java:67-81`](https://github.com/apache/pulsar/blob/25811b468b1970094ef2118f5fa239d138bda844/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/ManagedLedgerUtils.java#L67-L81)), while the proposed method offers no such parameter. Some existing overloads are unbounded, so this is about retaining the option to limit bytes, not a universal property of batch reads.

lhotari:
Thanks for saying this plainly — it changes how I read the proposal, and it is worth pulling into the document itself.

If #25891 is the real target, then PIP-480 as written does not cover it. #25891 adds a public client API and protocol buffer changes; PIP-480 states "No broker binary protocol changes" and motivates itself entirely on KoP. The mailing list would be voting on a narrower thing than the one you are building, and the client surface would still need its own proposal afterwards.

For the scope of the PIP, either of these works:

- one PIP covering the end-to-end reader — managed-ledger API, protocol command, client API — or
- PIP-480 kept as the managed-ledger building block, saying so explicitly, with the client and protocol surface proposed separately.

What I would avoid is designing the `ManagedLedger` method against KoP alone and then discovering that the client reader needs different semantics. KoP has its own offset store and its own `DelayedFetchPurgatory`; a Pulsar client reader has neither, so the questions about waiting, error signalling and byte bounds land differently for it. Those are the ones I've left inline.

For the managed-ledger API shape, I do have a preference: the optional reader-builder factory described in my review summary, with `asyncRead` on a closeable `ManagedReader`. Please develop @void-ptr974's factory/reader suggestion in `Alternatives` and explain how the design serves both callers.

Take the time you need on the redesign; there is no deadline here from my side.

PR #25875, opened 2026-05-27
[feat][broker] PIP-475: regular-to-scalable topic migration command

merlimat:
## Summary

Third implementation PR for [PIP-475: Regular-to-Scalable Topic Migration](https://github.com/apache/pulsar/blob/master/pip/pip-475.md). Builds on the broker synthetic-layout lookup (#25822) and the V5 SDK support (#25850). Adds the admin command that converts an existing regular (partitioned or non-partitioned) topic into a scalable topic, plus the broker guard that keeps legacy v4 clients off a migrated topic. The v4-guard work originally scoped as a separate PR is folded in here.

### Migration command (admin REST + client + CLI)
- `POST /admin/v2/scalable-topics/{t}/{n}/{topic}/migrate?force=` (requires `produce` permission).
- `ScalableTopics.migrateToScalable(topic, force)` / `...Async`; `pulsar-admin scalable-topics migrate <topic> [--force]`.
- Orchestration: reject if already scalable (409) or topic missing (404); unless `--force`, reject if any legacy v4 connection is attached (409); build the migrated layout; create the new child segment topics; CAS-write the scalable metadata (commit point — connected V5 lookup sessions transition from the synthetic layout to the real DAG via the metadata watch); terminate the old topic(s) so no further v4 writes land — they become the drainable sealed parent segments.

### Migrated layout (`ScalableTopicController.createMigratedMetadata`)
- N sealed legacy parent segments (the old partitions, each spanning the full hash range since v4 mod-N scattered keys everywhere) + N active range-based children with full parent fan-in; `nextSegmentId = 2N`.

### Legacy-vs-V5 connection detection
- New `ScalableTopicConstants.V5_MANAGED_METADATA_KEY` marker; the V5 SDK sets it on its per-segment producers/consumers **only when attaching to a legacy segment** (a wrapped `persistent://` topic). The migration pre-check counts producers/consumers whose metadata lacks the marker as legacy v4 connections, so a topic served solely by upgraded V5 clients passes the check.

### v4 write guard
- Post-migration termination blocks attached/new v4 producers at send. `BrokerService.isAllowAutoTopicCreationAsync` also refuses to auto-create a `persistent://` topic shadowed by a scalable topic (reuses `TopicTerminatedError`, metadata-cache-backed, gated by `scalableTopicsEnabled`) — enforcing "once scalable, always scalable" even after the old ledgers are GC'd.

## Test plan
- [x] `ScalableTopicControllerTest`: `createMigratedMetadata` for partitioned + non-partitioned sources.
- [x] `ScalableTopicMigrationTest` (7 cases): migrate partitioned + non-partitioned, already-scalable 409, missing 404, legacy-connection 409 without force, success with `--force`, and the auto-create shadow guard.
- [x] Full `pulsar-client-v5` unit suite + `SegmentRouterTest` regression.
- [ ] End-to-end migration test (full operator timeline: v4 produce → upgrade to V5 → migrate → per-key ordering / drain / termination / v4 lockout) — follow-up PR.

### DIFF
--- a/pip/pip-475.md
+++ b/pip/pip-475.md
@@ -211,7 +211,7 @@ There is no `TopicMigratedException` exposed to the application by default. Appl
 
 #### 2.1 Admin REST endpoint
 
-`POST /admin/v2/scalable-topics/{tenant}/{namespace}/{topic}/migrate` — requires `produce` permission on the topic (the same permission needed to write to it in the first place). Migration is irreversible but does not destroy data, so the blast radius is bounded by what a write-permissioned user can already do; super-user is not required.
+`POST /admin/v2/scalable-topics/{tenant}/{namespace}/{topic}/migrate` — authorized via a dedicated `TopicOperation.MIGRATE_TO_SCALABLE` operation, which the default authorization provider maps to `PRODUCE` permission on the topic (the same permission needed to write to it in the first place). Exposing it as its own operation lets third-party RBAC implementations restrict migration independently of produce. Migration is irreversible but does not destroy data, so the blast radius is bounded by what a write-permissioned user can already do; super-user is not required.
 
 Query parameter: `force` (boolean, default `false`).
 
--- a/pip/pip-475.md
+++ b/pip/pip-475.md
@@ -77,7 +77,7 @@ The broker responds to the lookup session based on the input form and the topic'
 | Input form | Lookup response |
 |---|---|
 | `topic://t/n/x` | Real DAG layout (or `NotFound` if the scalable topic doesn't exist — same as today). |
-| `persistent://t/n/x` | If scalable metadata exists for the equivalent `topic://t/n/x`: real DAG layout, with the topic identity promoted to `topic://...` for all subsequent operations. Otherwise: a **synthetic layout** that models the regular topic's partitions as **special segments** (see [Special segments](#special-segments) below). |
+| `persistent://t/n/x` | If scalable metadata exists for the equivalent `topic://t/n/x`: real DAG layout, with the topic identity promoted to `topic://...` for all subsequent operations. Otherwise: a **synthetic layout** that models the regular topic's partitions as **legacy segments** (see [Legacy segments](#legacy-segments) below). |
 | `my-topic` (or any short / unscoped form) | Normalize to `persistent://public/default/my-topic` then apply the rule above. |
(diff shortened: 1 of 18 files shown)


pip/pip-475.md
```diff
@@ -190,21 +203,23 @@ V5-specific features that don't apply on a regular topic surface as ordinary "th
...
 
 There is no `TopicMigratedException` exposed to the application by default. Applications that *want* to observe migrations can subscribe to the lookup session's layout-change events directly via a future hook — out of scope for this PIP.
 
 ### 2. Migration command
 
 #### 2.1 Admin REST endpoint
 
-`POST /admin/v2/scalable-topics/{tenant}/{namespace}/{topic}/migrate-to-scalable` — requires `produce` permission on the topic (the same permission needed to write to it in the first place). Migration is irreversible but does not destroy data, so the blast radius is bounded by what a write-permissioned user can already do; super-user is not required.
+`POST /admin/v2/scalable-topics/{tenant}/{namespace}/{topic}/migrate` — requires `produce` permission on the topic (the same permission needed to write to it in the first place). Migration is irreversible but does not destroy data, so the blast radius is bounded by what a write-permissioned user can already do; super-user is not required.
```
lhotari: perhaps we should add a specific org.apache.pulsar.common.policies.data.TopicOperation enum value `MIGRATE_TO_SCALABLE` that maps to `PRODUCE` in the default implementation? This would make it easier to integrate to 3rd party RBAC implementations for restricting the operation when desired.

lhotari (approved):
LGTM, just one comment about adding a specific TopicOperation so that 3rd party RBAC implementations could limit access when needed.

pip/pip-475.md
merlimat: Good idea — done in 9cc08fd. Added a dedicated `TopicOperation.MIGRATE_TO_SCALABLE`; the default `PulsarAuthorizationProvider` maps it to `canProduceAsync` (so out-of-the-box it still requires produce permission, unchanged), and third-party RBAC providers can now restrict migration independently. Updated the migrate endpoint to validate against it and adjusted the PIP wording.

merlimat: Added

merlimat: 👍

PR #25878, opened 2026-05-27
[test][broker] PIP-475: end-to-end migration tests + transition fixes

merlimat:
## Summary

Final implementation PR for [PIP-475: Regular-to-Scalable Topic Migration](https://github.com/apache/pulsar/blob/master/pip/pip-475.md). Builds on the synthetic-layout lookup (#25822), the V5 SDK support (#25850), and the migration command (#25875). Adds an end-to-end test of the full operator timeline against a live broker, and fixes three real bugs the test surfaced that prevented a connected V5 client from transitioning across the migration boundary.

### Tests (`V5MigrationEndToEndTest`)
- **produce-through-migration:** a V5 producer publishes via the synthetic layout (mod-N routing to the legacy segments); the topic is migrated while only that (marked) V5 producer is attached, so the pre-check passes without `--force`; the producer transparently follows the layout-change push to the real DAG and range-routes new messages to the active children; a V5 queue consumer then drains *every* pre- and post-migration message.
- **v4 lockout:** after migration the old topic is terminated, so a legacy v4 producer can no longer write to it.

### Fixes surfaced by the E2E test
- **`DagWatchSession` (the transition gap):** compute `segment://` URIs from the canonical `topic://` name, not the session's raw input. A session opened with a `persistent://` name (synthetic layout) previously threw *"Parent topic must have domain 'topic'"* inside `buildResponse` once the topic was migrated, so the real DAG was never pushed and connected clients never transitioned.
- **Migration pre-check:** inspect per-partition stats instead of the aggregate. Aggregated partitioned stats merge publishers by name into fresh stat objects that drop per-connection metadata, which hid the V5-managed marker and made every V5 connection look like a legacy v4 one (so a fully-V5 topic was wrongly rejected).
- **`ScalableTopicProducer`:** the send-retry now also covers per-segment producer *creation* (not just `send()`), since a migration terminates the old partition between routing and creation; detect the "segment gone" condition by unwrapping causes (type or message) and give the DAG-watch layout update a larger budget to arrive.

## Test plan
- [x] `V5MigrationEndToEndTest` (2): produce-through-migration, v4 lockout.
- [x] Regression: `ScalableTopicMigrationTest` 7/7, `DagWatchSessionTest` 18/18, `ScalableTopicControllerTest` 35/35, `V5RegularTopicInteropTest` 4/4, full `pulsar-client-v5` unit suite.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/ScalableTopics.java
@@ -339,12 +339,22 @@ public class ScalableTopics extends AdminResource {
         } catch (Exception e) {
             return CompletableFuture.failedFuture(e);
         }
-        CompletableFuture<? extends TopicStats> statsFuture =
-                partitions > 0
-                        ? admin.topics().getPartitionedStatsAsync(persistentBase.toString(), false)
-                        : admin.topics().getStatsAsync(persistentBase.toString());
-        return statsFuture.thenAccept(stats -> {
-            long legacy = countLegacyConnections(stats);
+        // For a partitioned topic, inspect per-partition stats rather than the aggregate:
+        // aggregation merges publishers by producer name into fresh stat objects that drop
+        // per-connection metadata, which would hide the V5-managed marker and make every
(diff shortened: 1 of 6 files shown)


lhotari (approved):
LGTM

PR #25879, opened 2026-05-27
[improve][proxy] Support scalable topics through the proxy

merlimat:
### Motivation

Scalable topics (`topic://`) did not work through the Pulsar proxy. The DAG-watch lookup and per-segment produce/consume paths already ride the proxy's transparent mode via ordinary `topic://` / `segment://` lookups, but the scalable-topic control connections that aren't tied to a specific broker had no path through the proxy:

- the consumer's controller subscribe dialed the controller's advertised address directly, bypassing the proxy;
- the namespace watcher sent its command over the proxy's lookup-mode connection, where it was silently dropped.

### Modifications

- **Proxy**: an empty `proxyToBrokerUrl` in `CommandConnect` is now treated as a request to pair the connection to any broker the proxy selects (configured broker URL, else discovery), bridged through the existing `DirectProxyHandler`. The non-empty path is unchanged.
- **Client**: `PulsarClientImpl.getAnyBrokerProxyConnection()` opens a dedicated connection that sends the empty `proxyToBrokerUrl` sentinel, using a distinct connection-pool key so it cannot collide with a normal service-URL connection.
- **V5 client**: the consumer's controller subscribe uses the any-broker proxy connection when the DAG-watch connection is proxied; the namespace watcher uses it unconditionally (safe because the broker ignores `proxyToBrokerUrl`).

### Verifying this change

- Added `ProxyScalableTopicsTest` (extends `ProxyMultiBrokerBaseTest`): produces and consumes a 2-segment scalable topic end-to-end through the proxy.
- Existing no-proxy V5 tests (`V5SmokeTest`, `V5QueueConsumerBasicTest`, `V5ScalableTopicsWatcherTest`) continue to pass.


lhotari (approved):
LGTM

PR #25880, opened 2026-05-27
[feat][ml] Add stateless RandomReader to ManagedLedger

dao-jun:
### Motivation

Add stateless `RandomReader` to ML.

  Today every read path in managed-ledger goes through a ManagedCursor. A cursor is comparatively heavyweight for several access patterns:

  1. Cursorless reads for downstream consumers. Reading via a cursor requires acknowledging entries — any un-acked read grows the backlog and interferes with retention/trimming. Reading messages scattered across arbitrary positions further requires repeated seek operations. A lightweight, stateless reader (no position, no
  ack, no backlog, no trimming hold) lets downstream projects batch-read entries directly by position, without those side effects.
  2. Cursorless batch reads for Pulsar itself. Internal components that only need to fetch entries at given positions (backfill, repair, offset-based lookups) currently pay the full cursor overhead. A cursorless primitive gives them a cheaper path.
  3. Groundwork for a future RandomReader client. This PR provides the server-side managed-ledger primitive that a planned RandomReader client will build on. That client is intended to support both pure random reads (serving metadata-style lookups) and sequential reads (serving stream-processing frameworks such as Flink).

### Modifications

  - Introduce RandomReader, a cursorless, stateless reader (public final class ... implements AutoCloseable) in the managed-ledger impl package — renamed from the former RandomReaderImpl; the previous RandomReader interface is removed. A reader maintains no read position, performs no ack, contributes no backlog, and does
  not prevent ledger trimming; reads do not wait for future entries, and callers must release every returned Entry. Reads honor count, position (maxPosition, inclusive), and estimated-size limits.
  - Back it with RandomReaders (a per-managed-ledger reader registry) and OpRandomReadEntries (the cross-ledger batch-read operation), both in the impl package. RandomReader instances are created only via RandomReaders#create(boolean populateCache) (package-private constructor).
  - Expose creation through ManagedLedgerImpl#randomReaders() (returning the registry) rather than on the ManagedLedger interface, since RandomReader support is bound to ManagedLedgerImpl's entry cache and is not a pluggable contract. Reader lifecycle is tied to managed-ledger close/fence.
  - Integrate with the entry cache: a populateCache=true reader writes read misses back to the cache and seeds the write tail, and contributes to the add-path expectedReadCount (via OpAddEntry / EntryReadCountHandlerImpl) so cache-eviction weight reflects active random readers; shouldCacheAddedEntry() now also accounts for
  them. A populateCache=false reader serves cache hits without write-back, avoiding cache pollution for truly random, low-frequency access (cache integration touches EntryCache / RangeEntryCacheImpl).
  - Add unit tests in ManagedLedgerTest covering read semantics, position/size boundaries, cross-ledger reads, error/partial results, close/fence lifecycle, and both cache modes.

### Verifying this change

*(or)*

*(or)*

*(example:)*

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


BewareMyPower (commented):
expected read count is not taken carefully.

In `EntryImpl`:

```java
    protected void deallocate() {
        if (decreaseReadCountOnRelease && readCountHandler != null) {
            readCountHandler.markRead();
        }
```

The read count will always be decreased no matter whether the read op is on a `RandomReader`. This could affect the cache eviction strategy

dao-jun:
@BewareMyPower fixed, PTAL

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
```diff
@@ -764,6 +764,29 @@ default ManagedLedgerAttributes getManagedLedgerAttributes() {
...
+    /**
+     * Create a standalone cursorless reader with the requested cache strategy.
+     *
+     * <p>{@code streaming=true} admits read misses to the cache and seeds the write tail for high hit rates.
+     * {@code streaming=false} serves cache hits but neither writes back misses nor seeds the tail, so
+     * low-frequency random reads do not pollute the cache.
+     *
+     * @throws UnsupportedOperationException when the managed-ledger implementation does not support random reads
+     */
+    default RandomReader newRandomReader(boolean streaming) {
```
BewareMyPower: From the perspective of API design, the `streaming` parameter here is confusing. It's eventually converted to the `expectedReadCount` int supplier passed to `ManagedLedgerImpl#asyncReadEntryForRandomReader`.

1. `streaming=true`: returns the count of active streaming readers (created by `newRandomReader(true)` but not closed)
2. `streaming=false`: returns 0

It's a bit confusing to users that when should they set `streaming` to false? For example, in your KoP implementation, should it be set to true or false?

BewareMyPower: Adding a `newRandomReader` method to the `ManagedLedger` interface is also a bad design. It means the downstream implementation is able to override the implementation.

IMO, exposing `RandomReaders#create` to public would be better.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/RandomReaders.java
```diff
@@ -0,0 +1,75 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.mledger.impl;
+
+final class RandomReaders {
+    private final ManagedLedgerImpl ledger;
+    // Lifecycle mutations are synchronized; volatile reads keep add and read paths lock-free.
+    private volatile int activeReaders;
+    private volatile int streamingReaders;
```
BewareMyPower: Why not just use atomic integers here? Even if you have synchronized for `create`, the inconsistency could still happen:

```java
        activeReaders++;
        // size() and streamingCount() will return inconsistency results here
        if (streaming) {
            streamingReaders++;
        }
```

From my understanding, inconsistency should be accepted, using atomic integers could be easy to understand and maintain.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
dao-jun: @BewareMyPower Good point about the API clarity. Let me explain the design intent and see if we can improve the documentation.

  ### The `streaming` parameter semantics:

  The `streaming` parameter controls **cache write-back behavior** for read misses:

  - **`streaming=true`** (default): Read misses are written back to the cache, and this reader contributes to cache tail seeding. This is optimal when you're reading **sequentially** or with **high temporal locality** (e.g., multiple readers consuming the same recent data, catch-up reads, backfill scenarios).

  - **`streaming=false`**: Read misses bypass cache write-back (serve from storage but don't pollute the cache). This is optimal for **truly random, low-frequency access patterns** where caching provides little benefit and would evict more useful entries.

  ### When to use `streaming=false`?

  Use `streaming=false` when:
  1. **Access pattern is truly random** (no temporal or spatial locality)
  2. **Low frequency**: won't re-read the same entries soon
  3. **Large working set**: scanning a huge backlog once
  4. You want to **avoid cache pollution** for other active consumers

  ### For KoP (Kafka-on-Pulsar):

  In KoP's case, it depends on the Kafka consumer pattern:
  - **Typical Kafka consumers** (sequential consumption with consumer groups): use **`streaming=true`** (default) because they exhibit high locality
  - **Admin tools doing random offset reads**: could use `streaming=false`

  WDYT? Would clearer documentation be sufficient, or do you think the API itself needs improvement?

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/RandomReaders.java
dao-jun: You're right about the inconsistency window. Although the inconsistency is confined within the `synchronized` block and the callers (`shouldCacheAddedEntry()`, etc.) can tolerate transient inconsistency, using `AtomicInteger` would make the code clearer and avoid any ambiguity about concurrent semantics. I'll switch to
  `AtomicInteger` for both counters.

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/ManagedLedger.java
BewareMyPower: I mean, the `streaming` argument itself is confusing. Adding detailed API document is good, but it's hard to connect "streaming" to something related to read cache.

BewareMyPower: BTW, WDYT about the comment [here](https://github.com/apache/pulsar/pull/25880#discussion_r3627720677)?

It would be better to provide such APIs:

```java
var reader = RandomReaders.create(ml, streaming);
```

instead of adding a new method to the interface like:

```java
// This behavior could be overwritten and it's hard for a custom ml implementation to implement it properly
var reader = ml.newRandomReader(streaming);
```

dao-jun: > BTW, WDYT about the comment [here](https://github.com/apache/pulsar/pull/25880#discussion_r3627720677)?
> 
> It would be better to provide such APIs:
> 
> ```java
> var reader = RandomReaders.create(ml, streaming);
> ```
> 
> instead of adding a new method to the interface like:
> 
> ```java
> // This behavior could be overwritten and it's hard for a custom ml implementation to implement it properly
> var reader = ml.newRandomReader(streaming);
> ```

@BewareMyPower I don't think this change is appropriate. Because ManagedLedger internally needs to operate on RandomReaders, if we change it this way, ManagedLedger seems to need to expose some other methods.

dao-jun: > I mean, the `streaming` argument itself is confusing. Adding detailed API document is good, but it's hard to connect "streaming" to something related to read cache.

it makes sense, renamed to `populateCache`

BewareMyPower: > ManagedLedger internally needs to operate on RandomReaders

These methods can be package visible. I see only the following methods are used: `closeAll()`, `hasStreamingReaders()`, `size()`, `streamingCount()`.

What other methods must be exposed after my suggested change?

dao-jun: @BewareMyPower The blocker for `RandomReaders.create(ml, populateCache)` as a public
  factory is that RandomReaders is not a stateless factory — it's a per-ManagedLedger
  stateful registry owned by ManagedLedgerImpl, and its counts feed back into the ML's
  own cache decisions:

  - shouldCacheAddedEntry() = cursors.shouldCacheAddedEntry() || randomReaders.hasCachePopulatingReaders()
  - getActiveCachePopulatingRandomReaderCount() feeds OpAddEntry#expectedReadCount
  - closeAll() is bound to the ML close/deactivate lifecycle

  So a public RandomReaders.create(ml, ...) can only be correct if it reaches the ML's
  OWNED registry instance. A detached new RandomReaders would make those counts wrong
  (cache tail seeding silently breaks). Reaching the owned registry requires an
  ml.getRandomReaders()-style accessor, and making it package-visible doesn't help
  external callers like KoP that only hold the ManagedLedger interface — it'd have to
  be public, which is strictly more surface than the single default interface method.

  The interface default method keeps RandomReaders fully encapsulated (package-private
  final), is backward compatible (default throws; ShadowManagedLedgerImpl inherits the
  override), and is polymorphic for callers holding the interface. It also mirrors the
  existing optional-capability default-method pattern on this interface
  (e.g. getLastDispatchablePosition).

  Happy to reconsider if I'm missing a way to reach the per-ML registry without a getter.

BewareMyPower: Okay, then I suggest just removing the public methods from the `ManagedLedger` interface. The `ManagedLedger` interface is pluggable, when you add a new method, it means that you think the method is overridable. However, the behavior of `RandomReaders` is bound to the specific entry cache mechanism of `ManagedLedgerImpl`.

For a `ManagedLedger` implementation that does not inherit from `ManagedLedgerImpl`, we should not allow creating a `RandomReader` on that. The default implementation of throwing an exception is not good because it's hard to know when is `newRandomReader` called.

The `populateCache` parameter is specifically coupled with `ManagedLedgerImpl`'s cache design, but it's unrelated to a custom managed ledger implementation.

Here is an example diff based on your latest commit:  [1.diff.txt](https://github.com/user-attachments/files/30295620/1.diff.txt)

dao-jun: @BewareMyPower addressed. Since there is no more public interface changes, we don't need a PIP. I'll remove the PIP lable

BewareMyPower: Yes, I think so, it's just a new public class added to `impl` package (like `ManagedLedgerUtils` added by me before)

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/RandomReader.java
```diff
@@ -0,0 +1,79 @@
...
+ * <p>A random reader does not maintain a read position, acknowledge entries, contribute to backlog, or prevent ledger
+ * trimming. A read can therefore fail, or start at the next retained ledger, when ledgers are trimmed concurrently.
+ * Reads do not wait for future entries.
+ *
+ * <p>Callers must release every returned {@link Entry}. Completion can run on a BookKeeper, Netty, or managed-ledger
+ * thread; callers that mutate thread-confined state must explicitly select an appropriate executor.
+ */
+@InterfaceAudience.LimitedPrivate
+@InterfaceStability.Evolving
+public interface RandomReader extends Closeable {
```
BewareMyPower: This interface can be moved to `impl` package as well. BTW, do we really need an interface? I'd like to remove `RandomReaderImpl` and move the concrete implementations to `RandomReader` for simplicity. (no need to add many default `read` overloads as well)

dao-jun: resolved

BewareMyPower (approved):

PR #25882, opened 2026-05-27
[fix][client] PIP-475: make async producer survive regular-to-scalable migration

merlimat:
## Motivation

Follow-up to the PIP-475 migration work. #25878 hardened the **sync** producer send path so a connected V5 producer rides through the synthetic→real-DAG transition during a regular-to-scalable migration: when a per-segment producer is (re)created on a partition that the migration just terminated, the send retries onto an active child once the new layout arrives instead of failing.

The **async** path (`producer.async()…send()` → `dispatchSendAttempt` / `appendToDispatchChain`) had the same gap and was missed, so an application using the async producer API and sending across a migration could still see sends fail.

## Modifications

- **Creation-time failures are now retried.** `appendToDispatchChain` previously failed the user future immediately when the per-segment producer failed to create; it now hands the failure to a callback that applies the same retry decision as a send failure.
- **Unified, robust detection.** Both the v4 `sendAsync` failure and the producer-creation failure now go through the shared `isSegmentGoneError` (type or message match), instead of the async path's previous type-only check that missed the message-wrapped exception the creation path surfaces.
- **Full retry budget.** Both paths now use `SEND_RETRY_MAX_ATTEMPTS` / `SEND_RETRY_MAX_BACKOFF_MS` (shared with the sync path) instead of the old 3-attempt (~600 ms) limit.

No wire/API/behavior change for the happy path; this only affects retry behavior on the segment-gone transition.

## Verification

- New `V5MigrationEndToEndTest.testV5AsyncProducerSurvivesMigration`: async sends issued immediately after migration must all complete (none fail across the boundary) and every pre-/post-migration message stays consumable. Passes alongside the existing sync + v4-lockout cases (3/3).
- `pulsar-client-v5` checkstyle/compile and `pulsar-broker` checkstyle clean.

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5MigrationEndToEndTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/client/api/v5/V5MigrationEndToEndTest.java
@@ -97,6 +97,62 @@ public class V5MigrationEndToEndTest extends V5ClientBaseTest {
     }
 
     @Test
+    public void testV5AsyncProducerSurvivesMigration() throws Exception {
+        // Same as the producer-survives-migration case, but driving the *async* producer API
+        // (producer.async()...send()). Async sends issued right after migration must ride
+        // through the synthetic→real-DAG transition — the per-segment producer for a
+        // just-terminated partition fails, and the send must retry onto an active child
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM

PR #25883, opened 2026-05-27
[feat][client] Embed GraalVM native image config

david-streamlio:
Main Issue: https://github.com/quarkusio/quarkus/issues/48776

### Motivation

Applications using `pulsar-client-original` in GraalVM native-image builds currently fail unless the consuming framework (Quarkus, Micronaut, Spring Native, etc.) provides its own reflection and class initialization metadata for Pulsar client internals. This forces every framework to independently reverse-engineer and maintain configuration that tracks Pulsar's internal class layout — and when it drifts, native builds break silently.

A concrete example: Quarkus's Pulsar extension registered `org.apache.pulsar.client.util.WithSNISslEngineFactory` for runtime initialization, but that class was removed in Pulsar 4.x. Because the configuration lived downstream instead of in the client itself, the breakage went unnoticed for a year (https://github.com/quarkusio/quarkus/issues/48776).

Embedding the configuration directly in `pulsar-client-original` following
the [GraalVM embedded configuration convention](https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildConfiguration/#embed-a-configuration-file) means the `native-image` tool picks it up automatically from the classpath, keeping the config in sync with the code that owns the classes.

### Modifications
Added three GraalVM native-image configuration files under
`pulsar-client/src/main/resources/META-INF/native-image/org.apache.pulsar/pulsar-client-original/`:

- **reflect-config.json** — Registers 20 classes for reflective access:
  the three `ConfigurationData` classes (deserialized by Jackson), all six
  `Authentication` implementations (instantiated by name via
  `AuthenticationUtil.create()`), the OAuth2 protocol model classes, schema
  internals (`ProtoBufParsingInfo`, `ProtobufNativeSchemaData`, `KeyValue`),
  `DataURLStreamHandler`, and `SecretsSerializer`.

- **native-image.properties** — Marks eight classes for runtime
  initialization that would otherwise fail during native-image build due to
  eager static initialization of Netty allocators, thread pools, or HTTP
  clients: `PulsarByteBufAllocator`, `Commands`, `Backoff`, `TokenClient`,
  `GenericProtobufNativeSchema`, `ConnectionPool`,
  `ControlledClusterFailover`, and `HttpClient`.

- **resource-config.json** — Includes the async-http-client default
  properties files that are loaded via classloader at runtime.

Added `NativeImageConfigTest` that validates:
- All classes referenced in `reflect-config.json` exist on the classpath
- All classes referenced in `native-image.properties` exist on the classpath
- All `Authentication` implementations shipped with the client are registered
- Both JSON config files are well-formed

This test catches configuration drift early — if a class is renamed, removed, or a new `Authentication` plugin is added without updating the config, the test fails with a descriptive message.

Note: the shaded `pulsar-client` JAR intentionally strips
`META-INF/native-image/**` in its shade plugin filters, which is correct since shading relocates packages and invalidates the original class names. Users building native images should depend on `pulsar-client-original`.

### Verifying this change

- Added `NativeImageConfigTest` that validates all 20 reflection entries and
  8 runtime-initialized classes resolve on the classpath, and that all
  `Authentication` implementations are registered.
- Every class in the configuration was verified against the Pulsar 4.2.1
  source tree.
- The equivalent configuration (derived from Quarkus's
  `SmallRyeReactiveMessagingPulsarProcessor`) was validated by successfully
  compiling six native-image integration tests in the Quarkus
  `integration-tests/reactive-messaging-pulsar` module against Pulsar 4.2.1.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### Documentation
- [x] `doc-not-needed`


pulsar-client/src/main/resources/META-INF/native-image/org.apache.pulsar/pulsar-client-original/resource-config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "resources": {
+    "includes": [
+      {
+        "pattern": "\\Qorg/asynchttpclient/config/ahc-default.properties\\E"
+      },
+      {
+        "pattern": "\\Qorg/asynchttpclient/config/ahc.properties\\E"
+      }
```
lhotari, line 9: Does `pulsar-admin-original` require a similar configuration?

lhotari (commented):
The changes look good so far. However, we'd need an integration test to verify that a Pulsar client application can be compiled to a native image. Besides this, I'd assume that there would have to be a smoke test to cover at least basic functionality (produce/consume). 

Would it be possible to add the integration test to `tests/pulsar-client-native-image` directory? It would be great if the test could be written in TestNG and call the native application with `java.lang.ProcessBuilder` or similar. The native application could have different cli options to perform the actual logic of each test case. I guess there could also be other ways to add test coverage. Perhaps the TestNG class could be directly native compiled without the need for a separate native application, at least junit is supported in the Gradle GraalVM plugin, however TestNG support is most likely missing. We could add Junit Jupiter dependencies to Pulsar alongside TestNG if that makes things easier. There has been plans to start migrating to Junit Jupiter in Pulsar anyways.

GraalVM Gradle plugin docs: https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html

david-streamlio:
Thanks for the review, @lhotari! I've added the integration test coverage you asked for. It follows the approach you suggested exactly.

**New module: `tests/pulsar-client-native-image`**

- **`NativeImageTesterApp`** — a small CLI application (the "native application") with subcommands that perform the actual logic of each test case (e.g. `produce`, `consume`, `produce-consume`). It's the binary that gets compiled to a native image via the `native-maven-plugin`.
- **`NativeImageSmokeTest`** — a TestNG driver that invokes the compiled native binary through `java.lang.ProcessBuilder`, passing the relevant CLI options per case. It first verifies the client compiles to a native image, then runs a basic produce/consume smoke test against a Pulsar broker started with Testcontainers (`PulsarContainer`).
- Wired into `tests/pom.xml` behind a `nativeImageTests` profile so it's opt-in and doesn't slow the default build. Added a nightly + `workflow_dispatch` GitHub Actions workflow (`ci-pulsar-native-image.yaml`) that runs it on a GraalVM community JDK.

I went with the separate-native-application + ProcessBuilder design rather than native-compiling the TestNG class directly, since (as you noted) TestNG native support is missing today. Happy to revisit that toward JUnit Jupiter if/when the migration lands.

I also added native-image metadata for `pulsar-client-admin` (`reflect-config.json` / `resource-config.json` / `native-image.properties`) covering the Jackson-serialized admin model classes and the runtime-initialized `AsyncHttpConnector`, with a static validator test — so admin-based client apps compile too.

Let me know if you'd like any changes to the test structure or CLI surface.

tests/pom.xml
lhotari: The master branch no longer uses maven, so this is obsolete. The new project should be added to top level `settings.gradle.kts`.

tests/pulsar-client-native-image/pom.xml
lhotari: master branch contains a gradle build. Please see `tests/pulsar-client-all-shade-test/build.gradle.kts` as a reference.

In Gradle, you should use GraalVM's Gradle plugin, https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html

.github/workflows/ci-pulsar-native-image.yaml
```diff
@@ -0,0 +1,78 @@
...
+    paths:
+      - '.github/workflows/ci-pulsar-native-image.yaml'
+      - '**/META-INF/native-image/**'
+      - 'tests/pulsar-client-native-image/**'
+
+env:
+  MAVEN_OPTS: -Xss1500k -Xmx1500m -Daether.connector.http.reuseConnections=false -Daether.connector.requestTimeout=60000 -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3 -Dmaven.wagon.http.serviceUnavailableRetryStrategy.class=standard -Dmaven.wagon.rto=60000
+
+jobs:
+  native-image-smoke-test:
```
lhotari: How slow are native image builds? Since Pulsar CI's integration tests run in parallel, it shouldn't be a problem if it's relatively slow. If it takes more than 20 minutes, that's when it starts becoming a problem.

It would be preferred to add a new entry to Pulsar CI's integration test matrix instead of adding a new GitHub Actions workflow:
https://github.com/apache/pulsar/blob/a1b0a4fa15fc6620dabfbc8c83664556f8debd65/.github/workflows/pulsar-ci.yaml#L458-L512

When using the integration test matrix solution, actual commands to run the test group are added to the `pulsar-build/run_integration_group_gradle.sh` file.

lhotari: For speeding up native image builds in CI, Claude suggested using the [quick build mode](https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildOutput/#recommendation-qbm:~:text=QBM%3A%20Use%20Quick%20Build%20Mode%20for%20Faster%20Builds) (`quickBuild`/`-Ob` flag) for native image compilation.

It should be possible to configure the quick build mode in this way with the GraalVM Gradle plugin:
```kotlin
graalvmNative {
    binaries.all {
        quickBuild = true
    }
}
```

pulsar-client/src/main/resources/META-INF/native-image/org.apache.pulsar/pulsar-client-original/resource-config.json
david-streamlio: Yes — `pulsar-client-admin-original` now ships its own native-image metadata (`reflect-config.json`, `resource-config.json`, `native-image.properties`) under `pulsar-client-admin/src/main/resources/META-INF/native-image/org.apache.pulsar/pulsar-client-admin-original/`. The admin `resource-config.json` covers the AsyncHttpClient `ahc-default.properties`/`ahc.properties` resources, and `native-image.properties` marks `AsyncHttpConnector` for runtime initialization. A `NativeImageConfigAdminTest` static validator checks these files.

tests/pom.xml
david-streamlio: Good catch — the branch was based on the Maven-era master. I've rebased onto the current Gradle master and dropped the `tests/pom.xml` change. The module is now registered in the top-level `settings.gradle.kts` alongside the other test modules.

tests/pulsar-client-native-image/pom.xml
david-streamlio: Done. Replaced the Maven `pom.xml` with a `build.gradle.kts` modeled on `tests/pulsar-client-all-shade-test/build.gradle.kts`, using the GraalVM `native-build-tools` plugin (`org.graalvm.buildtools.native`, declared in `gradle/libs.versions.toml`). The plugin's `nativeCompile` builds `NativeImageTesterApp` into a binary (consuming the embedded `META-INF/native-image` metadata), and the TestNG `NativeImageSmokeTest` runs on the JVM and drives that binary via `ProcessBuilder`.

.github/workflows/ci-pulsar-native-image.yaml
david-streamlio: Switched to the integration-test matrix as suggested. Deleted the standalone `ci-pulsar-native-image.yaml` and added a `NATIVE_IMAGE` entry to the `integration-tests` matrix in `pulsar-ci.yaml` (with a conditional `graalvm/setup-graalvm` step and a 40-min timeout), backed by a `test_group_native_image` function in `pulsar-build/run_integration_group_gradle.sh`. Quick build mode (below) should keep it well under the 20-min concern; I'll confirm the actual wall-clock on the first CI run.

david-streamlio: Added. The `build.gradle.kts` configures `graalvmNative { binaries.all { quickBuild = true } }` exactly as suggested.

.github/workflows/pulsar-ci.yaml
```diff
@@ -533,6 +539,16 @@ jobs:
             ${{ matrix.runtime_jdk != env.CI_JDK_MAJOR_VERSION && matrix.runtime_jdk || '' }}
             ${{ env.CI_JDK_MAJOR_VERSION }}
 
+      # The NATIVE_IMAGE group compiles a native binary with GraalVM native-build-tools,
+      # which requires a GraalVM JDK. This overrides JAVA_HOME for that matrix entry only.
+      - name: Set up GraalVM
+        if: ${{ matrix.group == 'NATIVE_IMAGE' }}
+        uses: graalvm/setup-graalvm@v1
```
lhotari: For ASF projects, GitHub Actions must use a fixed hash. The approved hashes are available in https://github.com/apache/infrastructure-actions/blob/main/actions.yml#L491

```suggestion
        uses: graalvm/setup-graalvm@329c42c5f4c343bceb505f0b28cc8499bc2bf174 # v1.5.4
```

.github/workflows/pulsar-ci.yaml
```diff
@@ -533,6 +539,16 @@ jobs:
...
+        with:
+          distribution: 'graalvm-community'
+          java-version: ${{ env.CI_JDK_MAJOR_VERSION }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
lhotari: There shouldn't be a need to pass the github-token although graalvm's documentation claims that it should be done. The default value is `${{ github.token }}` which is the same as `${{ secrets.GITHUB_TOKEN }}` unless `GITHUB_TOKEN` secret is explicitly provided. Let's remove this.
```suggestion
```

gradle/libs.versions.toml
```diff
@@ -145,6 +145,7 @@ jakarta-servlet = "6.0.0"
 oxia = "0.7.4"
 # Build plugins
 lightproto = "0.7.3"
+graalvm-buildtools = "0.10.6"
```
lhotari: Latest version is `1.1.1`
```suggestion
graalvm-buildtools = "1.1.1"
```

lhotari (approved):
LGTM

PR #25884, opened 2026-05-28
[improve][broker] PIP-473 P5.2: scalable-topics TC timeout + GC sweeps

merlimat:
## Summary

Adds the periodic background sweeps for the scalable-topics transaction coordinator landed in [#25863](https://github.com/apache/pulsar/pull/25863) (P5.1). Both run on a dedicated single-thread scheduler started by `PulsarService` and gated on assign-partition-0 ownership — only the elected broker sweeps each cycle. Concurrent sweeps from a stale owner remain safe because every state transition is a header CAS; the election is purely an efficiency measure (per-partition scoping comes with the partitioned TC in a later phase).

### Timeout sweep

Default cadence **60s**. Scans the by-deadline index up to `now` and drives each expired open txn through `endTransaction(ABORT)`, which re-reads and CAS-guards the header — so a txn the client commits in the same window is left alone (the resulting `InvalidTxnStatusException` / `BadVersionException` is treated as a benign race and logged at debug).

### GC sweep

Default cadence **300s**, retention **900s**. For each terminal state, scans the by-final-state index up to `now - retention`. For each candidate:

- If leftover `/txn/op` records remain — some participant hasn't applied the outcome yet, or never received the event (e.g. the TC crashed between the header CAS and the fan-out) — re-drive `fanOutEvents` and **leave the header in place** so the participant can re-read the true outcome. It removes its op records once it applies them, and a later GC pass — seeing no op records — deletes the header.
- If no op records remain, delete the header.

This ordering closes the fan-out-durability gap [lhotari raised on #25863](https://github.com/apache/pulsar/pull/25863#discussion_r3298435980) without ever stranding a committed txn's data: we never delete a header while a participant might still re-read it (which would default the outcome to ABORTED).

### Config

| Key | Default |
|---|---|
| `transactionCoordinatorScalableTopicsTimeoutSweepIntervalSeconds` | 60 |
| `transactionCoordinatorScalableTopicsGcIntervalSeconds` | 300 |
| `transactionCoordinatorScalableTopicsGcRetentionSeconds` | 900 |

All only meaningful when `transactionCoordinatorScalableTopicsEnabled = true` (still off by default).

### Drive-by

Refactored `fanOutEvents` to use `FutureUtil.waitForAll(List<CompletableFuture<Void>>)` — matches the new sweep methods and addresses the same comment lhotari left on P5.1.

## Test plan

- [x] `pulsar-broker:test --tests TransactionCoordinatorV5Test` — 5 new sweep cases plus all P5.1 cases:
  - `sweepTimeouts_abortsExpiredOpenTxnAndFansOut`
  - `sweepTimeouts_leavesUnexpiredOpenTxnAlone`
  - `sweepGc_deletesHeaderWhenNoOpsRemain`
  - `sweepGc_repairsAndRetainsHeaderWhenOpsRemain` (the fan-out-durability scenario)
  - `sweeps_skipWhenNotElected`
- [x] `pulsar-broker:test --tests TxnMetadataStoreTest` / `MetadataTransactionBufferTest` / `MetadataPendingAckStoreTest` — green.
- [x] Checkstyle clean (main + test).

## Deferred / follow-ups

- **Per-partition sweep scoping** lands with the partitioned TC (P5.3), replacing the single-elected-sweeper interim.
- **Pure metadata-store leader election** also belongs to P5.3.
- A leftover op record from a permanently-gone participant (segment deleted) currently keeps its header alive forever — the GC sweep keeps re-publishing harmlessly. A future phase can add a liveness check to force-cleanup, but doing so safely needs the participant-liveness signal that doesn't exist yet.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
@@ -301,16 +301,16 @@ public class TransactionCoordinatorV5 {
             }
         }).thenCompose(__ -> {
             TxnEvent event = new TxnEvent(txnIdKey, decision);
-            CompletableFuture<?>[] publishes = new CompletableFuture<?>[
-                    writeSegments.size() + ackParticipants.size()];
-            int i = 0;
+            List<CompletableFuture<Void>> publishes = new ArrayList<>(
+                    writeSegments.size() + ackParticipants.size());
             for (String segment : writeSegments) {
-                publishes[i++] = txnStore.publishSegmentEvent(segment, event);
+                publishes.add(txnStore.publishSegmentEvent(segment, event).thenApply(s -> null));
             }
             for (AckParticipant p : ackParticipants) {
-                publishes[i++] = txnStore.publishSubscriptionEvent(p.segment(), p.subscription(), event);
+                publishes.add(txnStore.publishSubscriptionEvent(p.segment(), p.subscription(), event)
+                        .thenApply(s -> null));
             }
-            return CompletableFuture.allOf(publishes);
+            return FutureUtil.waitForAll(publishes);
         });
     }
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
@@ -148,7 +148,15 @@ public class TransactionCoordinatorV5 {
         }
         try {
             sweep.get().get();
+        } catch (InterruptedException ie) {
+            // shutdownNow() interrupted the sweep thread mid-wait — restore the flag and exit
+            // quietly; this is the expected shutdown signal, not a failure.
+            Thread.currentThread().interrupt();
         } catch (Throwable t) {
+            if (closed) {
+                // close() raced with an in-flight async chain; not worth a WARN.
+                return;
+            }
             log.warn().attr("sweep", name).exception(t).log("v5 TC sweep cycle failed; will retry");
         }
     }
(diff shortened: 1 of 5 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
```diff
@@ -232,19 +301,159 @@ public void onCompleted() {
...
+        });
+    }
+
+    /**
+     * Run {@code action} only on the elected sweeper — the broker that owns partition 0 of
+     * {@code transaction_coordinator_assign}. Not owning it (or any error checking ownership) means
+     * "skip this cycle". Correctness doesn't depend on the election: every transition is a header
+     * CAS, so a stale owner sweeping concurrently is harmless.
+     */
+    private CompletableFuture<Void> ifElectedSweeper(Supplier<CompletableFuture<Void>> action) {
```
lhotari, line 467: is there a need to add a check for `closed` so that operations don't continue after shutdown?

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
```diff
@@ -70,18 +78,79 @@
...
+     */
+    private void runSweep(String name, Supplier<CompletableFuture<Void>> sweep) {
+        if (closed) {
+            return;
+        }
+        try {
+            sweep.get().get();
+        } catch (Throwable t) {
+            log.warn().attr("sweep", name).exception(t).log("v5 TC sweep cycle failed; will retry");
+        }
```
lhotari, line 171: runSweep logs at WARN on shutdown interruption — runSweep close() sets closed=true then shutdownNow(), interrupting an in-flight sweep.get().get(). The resulting InterruptedException is caught as Throwable and logged "v5 TC sweep cycle failed; will retry" at WARN — spurious noise on every shutdown mid-sweep. Suggest checking closed (or Thread.interrupted()) in the catch and downgrading/suppressing in that case.

lhotari (approved):
LGTM, just a minor comment about handling "closed" state for GC operations.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/coordinator/v5/TransactionCoordinatorV5.java
merlimat: Fixed

PR #25886, opened 2026-05-29
[feat] Upgrade BookKeeper dependency to 4.18.0

hangc0276:
### Motivation

Apache BookKeeper [4.18.0 has been released](https://github.com/apache/bookkeeper/releases/tag/release-4.18.0). This PR moves Pulsar from BookKeeper 4.17.3 to 4.18.0, which brings in particular:

* **Netty 4.2** — including the io_uring transport promoted out of the incubator and into mainline Netty.
* **LightProto-generated metadata formats** — BookKeeper 4.18 replaced the protobuf-generated `DataFormats` classes with LightProto-generated ones, so Pulsar code that reads/writes these formats needed to be adapted (the wire and text formats are unchanged, only the generated Java API differs).
* **A Jetty-12-compatible Prometheus metrics provider** — this lets us delete the temporary fork of the BookKeeper Prometheus stats provider that Pulsar has been carrying under `jetty-upgrade/` since the Jetty 12 migration, and go back to the upstream artifact.

Since BookKeeper 4.18 is built against Netty 4.2.x and gRPC 1.79.x, Pulsar's own versions of those libraries are aligned in the same change. BookKeeper 4.18 also stopped pulling the `io.grpc:grpc-all` aggregator into the stream-storage modules ([apache/bookkeeper#4784](https://github.com/apache/bookkeeper/pull/4784)), which makes it possible to remove `grpc-all` from Pulsar's own modules as well and slim the gRPC footprint of the server distribution in the same change.

### Modifications

**Dependency upgrades** (`gradle/libs.versions.toml`):
* BookKeeper 4.17.3 → 4.18.0 (including distributedlog and stream-storage artifacts)
* Netty 4.1.135.Final → 4.2.15.Final; the io_uring transport moves from `io.netty.incubator:netty-incubator-transport-*:0.0.26.Final` to the mainline `io.netty:netty-transport-*-io_uring` artifacts versioned with Netty itself
* gRPC 1.75.0 → 1.79.0
* RocksDB 7.9.2 → 9.9.3 (matching BookKeeper 4.18)
* slog 0.9.7 → 0.9.9
* perfmark 0.26.0 → 0.27.0 (gRPC 1.79's tested version)

**Netty 4.2 adaptations:**
* `EventLoopUtil`: the incubator `IOUringEventLoopGroup` no longer exists; io_uring event loops are now created with `MultiThreadIoEventLoopGroup` + `IoUringIoHandler`, and io_uring detection uses `IoEventLoopGroup.isIoType(IoUringIoHandler.class)` instead of an `instanceof` check. Channel classes renamed from `io.netty.incubator.channel.uring.*` to `io.netty.channel.uring.*`.
* `DefaultPulsarSslFactory`: Netty 4.2's `SslContext.newEngine(alloc, peerHost, peerPort)` now defaults the endpoint identification algorithm to `HTTPS` (it was unset in 4.1). Pulsar applies hostname verification separately (and only when `tlsHostnameVerificationEnable` is set), so the engine-level algorithm is explicitly cleared to avoid unintended hostname verification failures (e.g. internal `localhost` connections).
* `BitSetRecyclable`/`ConcurrentBitSetRecyclable` tests no longer assert that the just-recycled instance is returned by the next `create()`: the Netty 4.2 `Recycler` rework dropped that LIFO thread-local behavior, which was an implementation detail rather than a contract.
* Shading rules for the client and localrun shaded jars drop the `io.netty.incubator` includes.

**BookKeeper 4.18 LightProto API migration:**
* `OffloadUtils`, `OffloadIndexBlockImpl`, `BookieServiceInfoSerde`, and `PulsarLedgerUnderreplicationManager` (plus related tests) are migrated from the protobuf builder API (`DataFormats.XxxFormat.newBuilder()...build()`, `TextFormat.merge()`) to the LightProto API (`new XxxFormat()`, `parseFrom()`, `toByteArray()`, `toTextFormat()`/`parseFromTextFormat()`). Serialized data is fully compatible; only the generated classes changed.
* The stream-storage client configuration objects (`StreamConfiguration`, `NamespaceConfiguration`) underwent the same builder → mutable-object change (`BKStateStoreProviderImpl`, `LocalBookkeeperEnsemble`).
* `MathUtils` import updated to its new location `org.apache.bookkeeper.common.util.MathUtils`.

**gRPC dependency slimming** (enabled by BookKeeper 4.18 dropping `grpc-all`):
* The functions modules (`proto`, `instance`, `runtime`, `localrun`) only use a basic unary gRPC client/server (`ManagedChannel`/`Server`/`StreamObserver` over Netty), so `io.grpc:grpc-all` is replaced with `grpc-netty-shaded` + `grpc-stub`. The lightproto-generated stub uses its own marshaller, so `grpc-protobuf` is not required; `pulsar-functions-proto` now exposes `netty-buffer`/`grpc-stub` as `api` since they appear in its public signatures.
* `pulsar-functions-instance` no longer needs `protobuf-java` at compile time: BookKeeper 4.18's stream proto types (`StreamConfiguration`, `NamespaceConfiguration`) are LightProto-generated, and the remaining `com.google.protobuf` references in main code are reflective (`Class.forName`) schema detection for user functions. It is now `testImplementation` only (tests exercise that path with concrete protobuf types).
* The server distribution drops the explicit `grpc-all` dependency and all `io.grpc` excludes — the gRPC modules that are actually needed (`grpc-netty-shaded`/`grpc-stub` via the functions modules, `grpc-protobuf`/`grpc-services` via the Oxia client) now flow transitively, and the heavy `grpc-all` tail (`grpc-xds`, `grpc-alts`, `grpc-grpclb`, `grpc-rls`, `grpc-opentelemetry`, `grpc-gcp-csm-observability`, `grpc-servlet`, `grpc-auth`, and the transitive `opentelemetry-gcp-resources`) is no longer bundled.
* The version catalog drops the obsolete "match Maven" transitive-dependency pins and their unused aliases (`opencensus`, `google-http-client`, `j2objc-annotations`, `google-auth`) that only downgraded gRPC's own tested transitive versions; those now flow through at the versions gRPC ships with.

**Prometheus metrics provider:**
* Deleted the `jetty-upgrade/bookkeeper-prometheus-metrics-provider` module (a fork of BookKeeper's provider made Jetty-12-compatible during the Jetty upgrade). The server distribution now ships the upstream `org.apache.bookkeeper.stats:prometheus-metrics-provider:4.18.0`, and `conf/bookkeeper.conf` points `statsProviderClass` back to `org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider`. (The forked class name was never part of a Pulsar release.)

**Bookie startup / test environment fixes:**
* Test JVMs now run with `-Djava.net.preferIPv4Stack=true`, matching `bin/pulsar` / `bin/bookkeeper`. BookKeeper's `BookieId` validation rejects IPv6 zone identifiers (e.g. `fe80::1%lo0`), so without this, bookies bound to loopback fail to start on hosts where loopback resolves to an IPv6 link-local address (notably macOS).
* `LocalBookkeeperEnsemble.startStandalone()` sets `allowLoopback=true` on the bookie configuration, mirroring what the `start()` path already does: the embedded ensemble advertises a loopback address, which BookKeeper otherwise refuses to bind.
* `pulsar-broker` tests declare `re2j` explicitly (it was previously only available transitively; the dependency graph changed with the upgrades above).

**Licensing:**
* `LICENSE.bin.txt` for the server and shell distributions updated for the new artifact versions and the slimmed gRPC jar set, verified with `checkBinaryLicense`.

### Verifying this change

This change is already covered by existing tests: the full test suite exercises the upgraded BookKeeper client/server (embedded bookies in broker tests), the ledger metadata and offload metadata round-trips, and the TLS code paths affected by the Netty 4.2 behavior change.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

The io_uring native transport now comes from the mainline `io.netty:netty-transport-native-io_uring` artifacts instead of the Netty incubator ones, and the bookie's Prometheus stats provider is the upstream BookKeeper one again (`statsProviderClass` in `conf/bookkeeper.conf`).


dao-jun (approved):
LGTM

lhotari:
marking as draft until BK 4.18 has been released and this PR has been updated to use the BK 4.18 release

lhotari:
#25965 should be merged before this PR since this PR also includes those changes for testing reasons

merlimat (approved):

lhotari (approved):

dao-jun:
@hangc0276 @lhotari Can we cherrypick this PR to branch 4.0?

merlimat:
> @hangc0276 @lhotari Can we cherrypick this PR to branch 4.0?

No no, we shouldn't backport this to LTS version. It's a big change and above all is not a bugfix.

merlimat:
@dao-jun I think this should not be backported to 4.2 either.

PR #25887, opened 2026-05-29
[improve][test] Migrate pulsar-perf to the V5 client API

merlimat:
## Motivation

PIP-475 added regular-to-scalable topic migration on the broker and the V5 client SDK. The V5 client transparently routes against both regular and scalable topics. Pulsar's perf CLI (`pulsar-perf produce / consume / read / transaction`) still used the v4 client API and therefore could not benchmark scalable topics at all.

This PR migrates all four `pulsar-perf` commands to the V5 client API so the same binary now works against regular and scalable topics out of the box.

A small V5 client-side fix is bundled in because it is required for the perf migration: `PulsarClientBuilderV5#tlsPolicy` was a stub that dropped every TlsPolicy field on the floor (only setting `useTls=true`). The perf commands take `--trust-cert-file` and the related flags and expect them to land, so the policy is now wired through to `ClientConfigurationData` end-to-end.

## Modifications

**`pulsar-client-v5`**

- `PulsarClientBuilderV5#tlsPolicy` now wires `trustCertsFilePath`, `keyFilePath`, `certificateFilePath`, `allowInsecureConnection`, `enableHostnameVerification` through to the underlying v4 conf. Two new unit tests in `PulsarClientBuilderV5Test` cover field propagation and the `TlsPolicy.ofInsecure()` shortcut.

**`pulsar-testclient`** — depends on `pulsar-client-v5` / `pulsar-client-api-v5`.

- `PerfClientUtils` gets `createV5ClientBuilderFromArguments()` mirroring the v4 helper. Important detail: the TLS predicate only enables TLS when the user genuinely wants it (URL is `pulsar+ssl://`, a trust cert path is supplied, or a flag is explicitly TRUE). Earlier wiring that treated non-null Booleans as intent caused TLS to flip on against plaintext brokers because picocli / default-value resolution coerces those flags to `Boolean.FALSE` even when not passed.

- `PerformanceProducer`: V5 `Producer` / `AsyncProducer.async().send()`. `BatchingPolicy` / `ChunkingPolicy` / `CompressionPolicy` / `ProducerAccessMode`. `ProducerEncryptionPolicy` backed by `PemFileKeyProvider` for `--encryption-key-name` + `--encryption-key-value-file`. V5 client-level `TransactionPolicy` + `Transaction.async().commit()`.

- `PerformanceConsumer`: V5 `QueueConsumer`. All subscription types fold into `QueueConsumer` semantics (work distribution); `Exclusive` / `Failover` trigger a runtime warning. V5 has no `MessageListener`, so each consumer gets one dedicated poll thread driving `receive(Duration)` and invoking the same handler the v4 listener did. V5 `acknowledge(...)` is synchronous void — wrapped in try / catch with the existing failure counters.

- `PerformanceReader` → V5 `CheckpointConsumer`. Same poll-thread emulation as `PerformanceConsumer` replaces `ReaderListener`. `--start-message-id` now only accepts `earliest` / `latest` — V5 `Checkpoint` has no `lid:eid` factory and rejecting that form explicitly is cleaner than silently falling back to `earliest`.

- `PerformanceTransaction`: combined producer+consumer over V5 `Transaction`. V5 puts `.transaction(txn)` on the message builder rather than `producer.newMessage(txn)`. Same semantics, different ergonomics.

### V5 features without a direct v4 equivalent

The following CLI flags survive for back-compat but are logged as no-ops at start-up:

`--max-outstanding`, `--max-outstanding-across-partitions`, `--stats-interval-seconds`, `--max-lookup-request`, `--ssl-factory-plugin*`, `--busy-wait`, `--auto-scaled-receiver-queue-size`, `--batch-index-ack`, `--pool-messages`, chunked-message knobs (`--max_chunked_msg`, `--expire_time_incomplete_chunked_messages`, `--auto_ack_chunk_q_full`), `--receiver-queue-size-across-partitions`, `--replicated`, `--use-tls`.

Per-transaction `--txn-timeout` moves to client-level `TransactionPolicy`.

## Verifying this change

- New unit tests in `PulsarClientBuilderV5Test` cover the TlsPolicy wiring.

This change is already covered by existing tests, so it does not need additional tests.

- The public API: yes — the four `pulsar-perf` subcommands. CLI flag UX is preserved end-to-end; behavior for flags listed above is logged as a no-op.
- The schema: no
- The default values of configurations: no
- The wire protocol: no
- The rest endpoints: no
- The admin cli options: no
- Anything that affects deployment: no

### DIFF
--- a/pulsar-testclient/src/main/java/org/apache/pulsar/testclient/PerformanceProducer.java
+++ b/pulsar-testclient/src/main/java/org/apache/pulsar/testclient/PerformanceProducer.java
@@ -590,6 +590,11 @@ public class PerformanceProducer extends PerformanceTopicListArguments{
             AtomicLong totalSent = new AtomicLong(0);
             AtomicLong numMessageSend = new AtomicLong(0);
             Semaphore numMsgPerTxnLimit = new Semaphore(this.numMessagesPerTransaction);
+            // Send futures of the in-flight transaction. V5 transaction-aware sends are queued
+            // onto an internal dispatch chain, so the v4-side txn-coordinator registration can
+            // lag the local counter; we await these before committing so commit never races
+            // ahead of the sends (otherwise the broker rejects with InvalidTxnStatusException).
+            final List<java.util.concurrent.CompletableFuture<?>> pendingTxnSends = new ArrayList<>();
             while (!Thread.currentThread().isInterrupted()) {
                 if (produceEnough) {
                     break;
@@ -664,7 +669,7 @@ public class PerformanceProducer extends PerformanceTopicListArguments{
                         messageBuilder.key(String.valueOf(totalSent.get()));
                     }
                     PulsarClient pulsarClient = client;
-                    messageBuilder.send().thenRun(() -> {
+                    var sendFuture = messageBuilder.send().thenRun(() -> {
                         bytesSent.add(payloadData.length);
                         messagesSent.increment();
                         totalSent.incrementAndGet();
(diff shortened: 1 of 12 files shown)


lhotari (approved):
LGTM

dao-jun (approved):

pulsar-testclient/src/main/java/org/apache/pulsar/testclient/PerformanceProducer.java
```diff
@@ -693,11 +702,9 @@ private void runProducer(int producerId,
                     if (this.isEnableTransaction
                             && numMessageSend.incrementAndGet() == this.numMessagesPerTransaction) {
                         if (!this.isAbortTransaction) {
-                            transaction.commit()
+                            transaction.async().commit()
```
void-ptr974, line 726: messageBuilder.send() is asynchronous, but this commits based only on the local send counter. In V5 the actual send is dispatched through the producer dispatch chain, so commit can race before all sends are registered in the transaction. Please collect the per-transaction send futures and wait for them before committing, like PerformanceTransaction does.

merlimat: Good catch — fixed in 70d897584d8. `PerformanceProducer` now collects the in-flight transaction's send futures and awaits them with `allOf(...).join()` before `commit()`, mirroring the `PerformanceTransaction` fix you referenced. The send chain already swallows per-send failures, so the join never throws on a send error.

void-ptr974 (approved):
Thanks for the update. This makes V5 more practical to try and benchmark, especially with scalable topics.

Looking forward to the broader V5 client work.

LGTM.

PR #25894, opened 2026-05-30
[fix][client] Fix incorrect topic in OTel spans for multi-topic consumers

### LINKED ISSUE #25895: [Bug] OTel consumer spans use incorrect topic for multi-topic consumers
### Search before reporting

- [x] I searched in the issues and found nothing similar.

### Read release policy

### User environment

- Pulsar Java client with OpenTelemetry tracing enabled.
- Observed from the current master branch code path.
- This issue is not tied to a specific broker OS or deployment environment.

### Issue Description

When OpenTelemetry tracing is enabled, consumer spans for multi-topic consumers can use the internal consumer topic name instead of the actual topic of the consumed message.

For example, the span name and `messaging.destination.name` can contain:

```text
MultiTopicsConsumer-*
```

instead of the real Pulsar topic or partition topic that the message came from.

This makes traces misleading because users cannot see the actual message source topic from the consumer span.

### Error messages

No error messages or stack traces are emitted. The issue is incorrect OpenTelemetry span metadata.

### Reproducing the issue

1. Enable OpenTelemetry tracing in the Pulsar Java client.
2. Create a consumer subscribed to multiple topics.
3. Receive and acknowledge a message.
4. Inspect the exported consumer span.

Actual result:

```text
span name = process MultiTopicsConsumer-*
messaging.destination.name = MultiTopicsConsumer-*
```

Expected result:

```text
span name = process <actual-message-topic>
messaging.destination.name = <actual-message-topic>
```

### Additional information

A fix is proposed in https://github.com/apache/pulsar/pull/25894.

void-ptr974:
Fixes #25895

### Motivation

OTel consumer spans for multi-topic consumers used the internal dummy consumer topic, so traces could show span names and destination attributes like `MultiTopicsConsumer-*` instead of the actual message topic.

### Modifications

Use the message topic when setting the consumer span destination, and keep the consumer topic only as a fallback for messages or callbacks that do not carry a topic.

### Verifications

- `./gradlew :pulsar-client-original:compileJava`
- `./gradlew :pulsar-broker:test --tests org.apache.pulsar.broker.service.OpenTelemetryTracingIntegrationTest.testMultiTopicConsumerTracing`


void-ptr974:
@Technoboy- Could you please take a look at this small OpenTelemetry tracing fix when you have time? Thanks!

Technoboy- (approved):

PR #25892, opened 2026-05-30
[fix][test] Fix flaky SameAuthParamsLookupAutoClusterFailoverTest.testAutoClusterFailover() test

oneby-wang:
### Motivation

`SameAuthParamsLookupAutoClusterFailoverTest.testAutoClusterFailover` can fail intermittently when run repeatedly with `invocationCount = 100`.

The failure reproduced as:

```text
Gradle suite > Gradle test > org.apache.pulsar.broker.SameAuthParamsLookupAutoClusterFailoverTest > testAutoClusterFailover[42](false) FAILED
    org.awaitility.core.ConditionTimeoutException: Assertion condition Arrays differ at element [1]: Healthy != PreFail expected [Healthy] but found [PreFail] within 3 minutes.
        at org.apache.pulsar.broker.SameAuthParamsLookupAutoClusterFailoverTest.awaitStatesAndIndex(SameAuthParamsLookupAutoClusterFailoverTest.java:154)
        at org.apache.pulsar.broker.SameAuthParamsLookupAutoClusterFailoverTest.testAutoClusterFailover(SameAuthParamsLookupAutoClusterFailoverTest.java:134)
```

The key flaky signal is that the same failover provider was driven by two scheduled check threads:

```text
2026-05-29T22:55:15,723 - INFO - [broker-service-url-check-2796-1:SameAuthParamsLookupAutoClusterFailover] - Failover to low priority pulsar service [0] pulsar://localhost:53673 --> [2] pulsar://localhost:53683. States: [Failed, Failed, Healthy], Counters: [0, 0, 0] {}
2026-05-29T22:55:15,723 - INFO - [broker-service-url-check-2795-1:SameAuthParamsLookupAutoClusterFailover] - Failover to low priority pulsar service [0] pulsar://localhost:53673 --> [2] pulsar://localhost:53683. States: [Failed, Failed, Healthy], Counters: [0, 0, 0] {}
2026-05-29T22:55:21,734 - INFO - [broker-service-url-check-2796-1:SameAuthParamsLookupAutoClusterFailover] - Recover to high priority pulsar service [2] pulsar://localhost:53683 --> [1] pulsar://localhost:53683. States: [Failed, Healthy, Healthy], Counters: [0, 0, 0] {}
[...]

PR #25899, opened 2026-05-30
[fix][client] Prevent duplicate ServiceUrlProvider initialization

oneby-wang:
### Motivation

apache/pulsar#25892 fixed a flaky `SameAuthParamsLookupAutoClusterFailoverTest` by removing an extra manual `failover.initialize(client)` call from the test.

The root cause of that flakiness was duplicate initialization. `PulsarClientBuilder.build()` already initializes the configured `ServiceUrlProvider` through `PulsarClientImpl`, so calling `initialize(client)` again starts duplicate background checks for the same provider instance.

This is especially problematic for `SameAuthParamsLookupAutoClusterFailover`, because each `initialize` call creates a new `broker-service-url-check` `EventLoopGroup`. Multiple checker threads can then mutate the same failover state and produce subtle race conditions that are difficult to diagnose. `AutoClusterFailover` and `ControlledClusterFailover` have the same lifecycle risk: duplicate initialization can register duplicate scheduled tasks, and `ControlledClusterFailover` can also recreate its HTTP client without closing the previous one.

### Modifications

- Make AutoClusterFailover, ControlledClusterFailover, and SameAuthParamsLookupAutoClusterFailover fail fast when initialize(PulsarClient) is called more than once.
- Remove the duplicate manual initialize call from SameAuthParamsLookupAutoClusterFailoverTest.testAutoClusterFailover because the provider is already initialized by PulsarClientBuilder.build().
- Add dedicated tests that build a PulsarClient with each provider and then verify a second initialize(client) call throws IllegalStateException.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

The threading model is affected only by preventing duplicate background failover check tasks from being registered for the same ServiceUrlProvider instance.

### DIFF
--- a/pulsar-broker/src/test/java/org/apache/pulsar/broker/SameAuthParamsLookupAutoClusterFailoverTest.java
+++ b/pulsar-broker/src/test/java/org/apache/pulsar/broker/SameAuthParamsLookupAutoClusterFailoverTest.java
@@ -99,7 +99,6 @@ public class SameAuthParamsLookupAutoClusterFailoverTest extends OneWayReplicato
                 .tlsTrustCertsFilePath(CA_CERT_FILE_PATH);
         }
         final PulsarClient client = clientBuilder.build();
-        failover.initialize(client);
         final EventLoopGroup executor = WhiteboxImpl.getInternalState(failover, "executor");
         final PulsarServiceState[] stateArray =
                 WhiteboxImpl.getInternalState(failover, "pulsarServiceStateArray");
@@ -139,6 +138,20 @@ public class SameAuthParamsLookupAutoClusterFailoverTest extends OneWayReplicato
         producer.close();
         client.close();
         dummyServer.close();
+    }
+
+    @Test
+    public void testInitializeCanOnlyBeCalledOnce() throws Exception {
+        setup();
+        final SameAuthParamsLookupAutoClusterFailover failover = SameAuthParamsLookupAutoClusterFailover.builder()
+                .pulsarServiceUrlArray(new String[]{pulsar1.getBrokerServiceUrl()})
+                .checkHealthyIntervalMs(1000)
+                .build();
+
+        try (PulsarClient client = PulsarClient.builder().serviceUrlProvider(failover).build()) {
+            Throwable error = Assert.expectThrows(IllegalStateException.class, () -> failover.initialize(client));
+            Assert.assertEquals(error.getMessage(), "ServiceUrlProvider has already been initialized");
+        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


void-ptr974 (changes requested):
I think this needs a lifecycle fix. If the same ServiceUrlProvider instance is reused to build a second client, the second build now fails with IllegalStateException, but the constructor failure path calls shutdown(), which unconditionally closes conf.getServiceUrlProvider(). That can close the provider still used by the first live client.

Example:

```
ServiceUrlProvider provider = AutoClusterFailover.builder()
        .primary(primary)
        .secondary(List.of(secondary))
        .failoverDelay(1, TimeUnit.SECONDS)
        .switchBackDelay(1, TimeUnit.SECONDS)
        .build();

PulsarClient client1 = PulsarClient.builder()
        .serviceUrlProvider(provider)
        .build();

PulsarClient.builder()
        .serviceUrlProvider(provider)
        .build(); // fails, then closes provider used by client1
```

Could we only close the provider on constructor failure if this PulsarClientImpl successfully initialized it?

oneby-wang:
> Could we only close the provider on constructor failure if this PulsarClientImpl successfully initialized it?

@void-ptr974 Nice catch. I agree with this approach. @lhotari WDYT?

lhotari (approved):
LGTM

lhotari:
> > Could we only close the provider on constructor failure if this PulsarClientImpl successfully initialized it?
> 
> @void-ptr974 Nice catch. I agree with this approach. @lhotari WDYT?

@oneby-wang It would be an improvement to handle this case. In addition, the javadoc of `org.apache.pulsar.client.api.ServiceUrlProvider` should be improved. There should be explanation that each instance lifecycle is tied to one PulsarClient instance. Similar javadoc should be added to the implementation where this is now enforced.

oneby-wang:
> It would be an improvement to handle this case.

@lhotari I see, let's get this merged first. I'll create another PR to address the improvement.

lhotari:
> > It would be an improvement to handle this case.
> 
> @lhotari I see, let's get this merged first. I'll create another PR to address the improvement.

@oneby-wang I think that the javadoc improvement belongs to this PR. This PR makes the implicit contract of the ServiceUrlProvider interface explicit and it's useful to document that in javadoc.

oneby-wang:
> This PR makes the implicit contract of the ServiceUrlProvider interface explicit and it's useful to document that in javadoc.

@lhotari Addressed.

lhotari (approved):
LGTM, good work @oneby-wang

PR #25900, opened 2026-05-30
[fix][test] Stabilize testSecondaryIsolationGroupsBookiesNegative() test

oneby-wang:
### Motivation

`testSecondaryIsolationGroupsBookiesNegative()` is flaky because the test calls newEnsemble immediately after initializing IsolatedBookieEnsemblePlacementPolicy. The policy loads the bookie rack metadata cache asynchronously. If the first newEnsemble call races with the initial cache load, cachedRackConfiguration can still be empty and the isolation filtering is skipped, so the ensemble may be created successfully instead of throwing BKNotEnoughBookiesException.

The test is intended to verify the negative fallback path for secondary isolation groups: when the primary group does not have enough available bookies and the configured secondary group is absent from rack metadata, the policy must not fall back to default-group bookies, so creating an ensemble of size 3 should fail.

### Modifications

Add an Awaitility guard before calling newEnsemble so the test waits until the async bookie rack metadata cache load triggered by initialize() is visible. This matches the pattern used by nearby tests and keeps the production placement policy behavior unchanged.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


lhotari (approved):
LGTM

PR #25908, opened 2026-06-01
[feat][pip] Retain Ended Transaction Status for Precise Error Reporting

poorbarcode:
### Motivation

This PIP is a proposal for Retain Ended Transaction Status for Precise Error Reporting

Discuss thread: https://lists.apache.org/thread/3fmnpzzqo4040mrqtbptnf7kr48r8jb1

Vote thread: 

Implementation: https://github.com/apache/pulsar/pull/25909

### DIFF
--- a/pip/pip-476.md
+++ b/pip/pip-476.md
@@ -1,0 +1,140 @@
+# PIP-476: Retain Ended Transaction Status for Precise Error Reporting
+
+# Background knowledge
+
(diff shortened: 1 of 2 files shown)


gaoran10:
Because the broker retains the aborted transaction index, and the `leastSigBits` of the transaction ID is monotonically increasing, could we use them to determine whether a transaction is aborted or committed?

PR #25913, opened 2026-06-01
[fix][meta] Fix PulsarZooKeeperClient async addWatch callback retry behavior

oneby-wang:
### Motivation

The async ZooKeeper addWatch wrapper in PulsarZooKeeperClient defined a retry callback but delegated the operation using the original callback and context. If addWatch returned a recoverable error, the caller could observe the transient failure directly instead of retrying. The wrapper's completion branch also recursively invoked itself, which would be incorrect if the wrapper callback were used.

This affects the SessionReestablished path in ZKMetadataStore, where Pulsar recreates the persistent recursive watch before notifying session listeners.

### Modifications

- Route async addWatch calls through the retry wrapper callback and ZooWorker context.
- Complete the caller-provided callback with the final result after retry handling.
- Fix the addWatch retry operation description.
- Add a regression test that verifies a recoverable CONNECTIONLOSS is retried and the caller receives the final OK result.

### Verifying this change

  - Added MetadataStoreTest.testAsyncAddWatchRetriesWithWrapperCallback to cover addWatch retry callback behavior.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/PulsarZooKeeperClient.java
+++ b/pulsar-metadata/src/main/java/org/apache/pulsar/metadata/impl/PulsarZooKeeperClient.java
@@ -1174,7 +1174,7 @@ public class PulsarZooKeeperClient extends ZooKeeper implements Watcher, AutoClo
     }
 
     @Override
-    public void addWatch(String basePath, Watcher watcher, AddWatchMode mode, VoidCallback cb, Object ctx) {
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM, good work @oneby-wang

PR #25914, opened 2026-06-01
[fix][broker] Prevent backlog quota cursor updates during topic close

Denovo1998:
Follow-up to #25684

### Motivation

`BacklogQuotaManager` already skips backlog quota handling when a topic is fenced or closing/deleting. However, checking the topic state only before entering the eviction path is not enough: a topic can begin close/delete after backlog quota eviction has started but before eviction mutates the slowest cursor.

This can let backlog quota eviction call `skipEntries` or `markDelete` while the topic teardown path is starting, which is the race this PR is intended to harden.

### Modifications

- Keep the existing entry guard for fenced or closing/deleting topics.
- Add a topic close/delete read-lock helper in `PersistentTopic`.
- Run direct backlog quota cursor mutations under that read lock, after checking `isClosingOrDeleting()`.
- Apply the guarded mutation path to size-based eviction before `ManagedCursor#skipEntries`.
- Apply the guarded mutation path to non-precise time-based eviction before `ManagedCursor#markDelete`.
- Keep the pending-ack cleanup hook after cursor advancement by continuing to call `markDeletePositionMoveForward()` inside the guarded mutation.
- Add real broker/topic race tests that start topic close after backlog quota eviction reaches the cursor mutation point and verify the cursor mark-delete position does not move.

### Verifying this change

- Added `BacklogQuotaManagerTest.testSizeBacklogEvictionRaceWithTopicCloseDoesNotSkipEntries`
- Added `BacklogQuotaManagerTest.testTimeBacklogEvictionRaceWithTopicCloseDoesNotMarkDelete`

Verified locally with:

```bash
./gradlew :pulsar-broker:test --rerun-tasks --tests "org.apache.pulsar.broker.service.BacklogQuotaManagerTest.testSizeBacklogEvictionRaceWithTopicCloseDoesNotSkipEntries" --tests "org.apache.pulsar.broker.service.BacklogQuotaManagerTest.testTimeBacklogEvictionRaceWithTopicCloseDoesNotMarkDelete"
./gradlew :pulsar-broker:spotlessCheck :pulsar-broker:checkstyleMain :pulsar-broker:checkstyleTest
git diff --check
```

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java
@@ -23,6 +23,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import lombok.CustomLog;
 import lombok.Getter;
@@ -195,18 +196,20 @@ public class BacklogQuotaManager {
                     break;
                 }
                 beforeBacklogQuotaCursorMutation(persistentTopic);
-                if (shouldStopEvictionOnTopicClose(persistentTopic)) {
+                if (!runCursorMutationIfTopicNotClosingOrDeleting(persistentTopic, () -> {
+                    // Skip messages on the slowest consumer
+                    log.debug()
+                            .attr("topic", persistentTopic.getName())
+                            .attr("messagesToSkip", messagesToSkip)
+                            .attr("consumer", slowestConsumer.getName())
+                            .attr("entriesInBacklog", entriesInBacklog)
+                            .log("Skipping messages on slowest consumer having backlog entries");
+                    slowestConsumer.skipEntries(messagesToSkip, IndividualDeletedEntries.Include);
+                    markDeletePositionMoveForward(persistentTopic, slowestConsumer);
+                    return null;
+                })) {
                     break;
                 }
-                // Skip messages on the slowest consumer
-                log.debug()
-                        .attr("topic", persistentTopic.getName())
-                        .attr("messagesToSkip", messagesToSkip)
-                        .attr("consumer", slowestConsumer.getName())
-                        .attr("entriesInBacklog", entriesInBacklog)
-                        .log("Skipping messages on slowest consumer having backlog entries");
-                slowestConsumer.skipEntries(messagesToSkip, IndividualDeletedEntries.Include);
-                markDeletePositionMoveForward(persistentTopic, slowestConsumer);
             } catch (Exception e) {
                 log.error()
                         .attr("topic", persistentTopic.getName())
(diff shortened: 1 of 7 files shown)


Denovo1998:
@lhotari @merlimat @dao-jun @nodece @Technoboy- PTAL.

void-ptr974:
Thanks for the follow-up. I think adding the second check before `skipEntries` / `markDelete` is a good improvement. The original entry check only handles the case where the topic is already closing before backlog eviction starts. This PR also handles the more important race where backlog eviction has already started, and then the topic begins closing before the cursor is mutated.

One thing I think we should decide is whether this PR is intended to be a best-effort mitigation or a strict guarantee.

With the current code, there is still a small window:

1. `shouldStopEvictionOnTopicClose()` returns false
2. topic close/delete starts immediately after that
3. eviction still continues to call `skipEntries` / `markDelete`

So this change can reduce the race window a lot, but it cannot completely guarantee that cursor mutation never happens during topic close/delete.

If best-effort mitigation is enough for this issue, I think the current direction is reasonable. If the goal is to strictly prevent cursor mutation during teardown, then we may need stronger coordination with the topic close/delete path, instead of only checking `isClosingOrDeleting()` before mutation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BacklogQuotaManager.java
```diff
@@ -196,6 +194,10 @@ private void dropBacklogForSizeLimit(PersistentTopic persistentTopic, BacklogQuo
                     log.debug().attr("slowestConsumer", slowestConsumer).log("no messages to skip for");
                     break;
                 }
+                beforeBacklogQuotaCursorMutation(persistentTopic);
+                if (shouldStopEvictionOnTopicClose(persistentTopic)) {
```
void-ptr974: This check is useful because it is close to the actual cursor mutation point.

The remaining subtle case is:

1. this check returns false
2. topic close/delete starts right after that
3. eviction still proceeds to `skipEntries` / `markDelete`

So I think we should clarify whether this is intended as a best-effort race reduction, or whether we need stronger coordination to strictly avoid cursor mutation once topic close/delete starts.

Denovo1998:
@void-ptr974 

Thanks, that is a good point. The previous version was best-effort only.

I updated the code to use the existing `PersistentTopic` close/delete read-write lock. Backlog quota eviction now checks `isClosingOrDeleting` and performs `skipEntries` / `markDelete` while holding the read lock; close/delete takes the write lock before setting the closing/deleting state. This removes the window between the state check and the cursor mutation with respect to the topic teardown state transition.

void-ptr974 (commented):
The current guard only protects the synchronous part that triggers eviction, but some cursor mutations happen later from async callbacks.

For the precise time-based eviction path, dropBacklogForTimeLimit() calls subscription.getExpiryMonitor().expireMessages(target), while the actual cursor advancement is later issued in PersistentMessageExpiryMonitor.findEntryComplete() via cursor.asyncMarkDelete(...). That later asyncMarkDelete(...) is outside the topic close/delete read-lock guard, so topic close/delete may still start between the outer expireMessages(...) call and the actual cursor mutation.

Suggested minimal fix: place the guard at the actual mutation point. In PersistentMessageExpiryMonitor.findEntryComplete(), acquire the topic close read lock, re-check closing/deleting state, and only then call cursor.asyncMarkDelete(...). If the topic is already closing/deleting, reset expirationCheckInProgress and skip the mark-delete.

Denovo1998:
@void-ptr974 
I updated the precise time-based eviction path as suggested.

The actual cursor mutation in `PersistentMessageExpiryMonitor.findEntryComplete()` now runs under the existing topic close/delete read lock. If the topic is already closing or deleting, the expiry path skips `cursor.asyncMarkDelete(...)`, resets `expirationCheckInProgress`, and returns without advancing the cursor.

I also added a broker-level race test that uses a real topic/subscription/cursor: it blocks just before the expiry mark-delete, starts topic close, waits until the topic is closing/deleting, then verifies the cursor mark-delete position is not advanced.

void-ptr974 (approved):
LGTM. The main backlog quota cursor mutation paths are now guarded close to the actual mutation points, including the precise expiry `asyncMarkDelete(...)` path. The added race tests cover the intended close/delete interleavings for the consumer backlog eviction cases.

PR #25915, opened 2026-06-01
[fix][broker]Do not trigger topic GC if replication is still active

poorbarcode:
### Motivation

Replication is stuck due to the Topic GC
- Enable binary way replication between the `primary cluster` and `backup cluster`
- `primary cluster`: Publishing messages into `primary cluster`
  - `backup cluster`: No consumer/producer is registered.
- `backup cluster`: Check topic GC
  - No subscriptions
  - No producers except `remote producer`
  - The topic should be GC
    - Disable replication
    - The topic will not be deleted since the remote-side producer is still registered 
- `backup cluster`: the topic GC progress is waiting for the remote-producer to be disconnected.
  - It will not be executed since no one wants to delete the topic. 
- `backup cluster`: backlog increases because the replicator was closed
  - Although the messages copied from the remote end will not be copied back repeatedly, the replicator still needs to perform a check and then mark delete.
 
### Modifications

- Topic GC will only be triggered if there is no producer(includes remote producer) and no replicator
- Replicator producer will be closed after there are no messages to be replicated anymore.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -826,6 +826,17 @@ public class ServiceConfiguration implements PulsarConfiguration {
     private Integer brokerDeleteInactiveTopicsMaxInactiveDurationSeconds = null;
 
     @FieldContext(
+        category = CATEGORY_POLICIES,
+        dynamic = true,
+            doc = "Time in seconds since a replicator's producer last sent a message, after which the replicator"
+                    + " is considered inactive. When inactive-topic deletion is enabled (via"
+                    + " brokerDeleteInactiveTopicsEnabled, or enabled by namespace/topic policy), a replicated"
+                    + " topic is eligible for deletion only once all its replicators have been inactive for this"
+                    + " duration."
+    )
+    private Integer brokerReplicationInactiveThresholdSeconds = 24 * 3600;
+
+    @FieldContext(
             category = CATEGORY_POLICIES,
             dynamic = true,
             doc = "Allow forced deletion of tenants. Default is false."
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -839,7 +839,7 @@ public class ServiceConfiguration implements PulsarConfiguration {
                 + " runs with the inactive-topic monitor, whose interval is"
                 + " brokerDeleteInactiveTopicsFrequencySeconds. The default is 86400 seconds (24 hours)."
     )
-    private Integer brokerReplicationInactiveThresholdSeconds = 24 * 3600;
+    private int brokerReplicationInactiveThresholdSeconds = 24 * 3600;
 
     @FieldContext(
             category = CATEGORY_POLICIES,
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -828,16 +828,14 @@ public class ServiceConfiguration implements PulsarConfiguration {
     @FieldContext(
         category = CATEGORY_POLICIES,
         dynamic = true,
-        doc = "Time in seconds that a topic may have no local producers before the broker considers its outbound"
-                + " replication producers idle. The timer starts when the inactive-replication check first observes"
-                + " that the topic has no local producers; remote producers created by replication from another"
-                + " cluster do not reset this timer. When the threshold is exceeded and the replication backlog is"
-                + " clear, the broker disconnects the topic's replication producers to release idle replication"
-                + " resources. A connected remote producer still makes the topic active for inactive-topic GC, so"
-                + " the topic is not deleted only because local producers are absent; deletion is still controlled"
-                + " by brokerDeleteInactiveTopicsEnabled and the namespace/topic inactive-topic policies. The check"
-                + " runs with the inactive-topic monitor, whose interval is"
-                + " brokerDeleteInactiveTopicsFrequencySeconds. The default is 86400 seconds (24 hours)."
+        doc = "Time in seconds that a persistent geo-replication replicator may stay idle before the broker"
+                + " disconnects its replication producer. A replicator is eligible only when it has no backlog and"
+                + " has not read entries for replication processing for longer than this threshold. Disconnecting"
+                + " only releases the idle producer; the replicator and its cursor remain available, and the"
+                + " producer is recreated automatically when new messages need to be replicated. Set this value to"
+                + " 0 or a negative value to disable idle-replicator disconnection. The check runs with the"
+                + " inactive-topic monitor, whose interval is brokerDeleteInactiveTopicsFrequencySeconds, and only"
+                + " when brokerDeleteInactiveTopicsEnabled is true. The default is 86400 seconds (24 hours)."
     )
     private int brokerReplicationInactiveThresholdSeconds = 24 * 3600;
 
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -828,11 +828,16 @@ public class ServiceConfiguration implements PulsarConfiguration {
     @FieldContext(
         category = CATEGORY_POLICIES,
         dynamic = true,
-            doc = "Time in seconds since a replicator's producer last sent a message, after which the replicator"
-                    + " is considered inactive. When inactive-topic deletion is enabled (via"
-                    + " brokerDeleteInactiveTopicsEnabled, or enabled by namespace/topic policy), a replicated"
-                    + " topic is eligible for deletion only once all its replicators have been inactive for this"
-                    + " duration."
+        doc = "Time in seconds that a topic may have no local producers before the broker considers its outbound"
+                + " replication producers idle. The timer starts when the inactive-replication check first observes"
+                + " that the topic has no local producers; remote producers created by replication from another"
+                + " cluster do not reset this timer. When the threshold is exceeded and the replication backlog is"
+                + " clear, the broker disconnects the topic's replication producers to release idle replication"
+                + " resources. A connected remote producer still makes the topic active for inactive-topic GC, so"
+                + " the topic is not deleted only because local producers are absent; deletion is still controlled"
+                + " by brokerDeleteInactiveTopicsEnabled and the namespace/topic inactive-topic policies. The check"
+                + " runs with the inactive-topic monitor, whose interval is"
+                + " brokerDeleteInactiveTopicsFrequencySeconds. The default is 86400 seconds (24 hours)."
     )
     private Integer brokerReplicationInactiveThresholdSeconds = 24 * 3600;
 
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -856,8 +856,12 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
                             log.warn("Attempting to add producer to a terminated topic");
                             throw new TopicTerminatedException("Topic was already terminated");
                         }
-                        return internalAddProducer(producer).thenCompose(__ -> this.startReplProducers())
-                                .thenApply(ignore -> {
+                        CompletableFuture<Void> internalAppProducerFuture = internalAddProducer(producer);
+                        internalAppProducerFuture.thenApply(__ -> this.startReplProducers()).exceptionally(ex -> {
+                            log.error("Failed to start replication producers.");
+                            return null;
+                        });
+                        return internalAppProducerFuture.thenApply(ignore -> {
                             USAGE_COUNT_UPDATER.incrementAndGet(this);
                             log.debug()
                                     .attr("producerName", producer.getProducerName())
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -678,17 +678,21 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
         if (cachedTime == null) {
             return;
         }
-        long threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds() * 1000;
-        if (System.currentTimeMillis() - cachedTime > threshold) {
+        int threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds();
+        if (System.currentTimeMillis() - cachedTime > threshold * 1000L) {
+            log.info().attr("brokerReplicationInactiveThresholdSeconds", threshold)
+                    .log("Disconnecting replication producers since no producer is active for a long time.");
             closeReplProducersIfNoBacklog().whenCompleteAsync((__, ex) -> {
                 // Here, "!=" is used instead of the "!equals()" to avoid the problem of "inability to compare due
                 // to multiple changes in this attribute within a short period of time".
                 if (cachedTime != localProducersEmptyTime) {
-                    log.warn().log("Restart replication producers since new producers were registered concurently"
+                    log.warn().log("Restart replication producers since new producers were registered concurrently"
                             + " when closing producers.");
                     startReplProducers().whenComplete((ignore, ex2) -> {
-                        log.error().exception(ex2).log("Failed to start replication producers after registered"
-                            + " new producers");
+                        if (ex2 != null) {
+                            log.error().exception(ex2).log("Failed to start replication producers after registered"
+                                + " new producers");
+                        }
                     });
                 }
             });
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -675,10 +675,16 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
         updateLocalProducersEmptyTime();
 
         final Long cachedTime = localProducersEmptyTime;
+        // Still active.
         if (cachedTime == null) {
             return;
         }
+        // Disabled the feature.
         int threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds();
+        if (threshold <= 0) {
+            return;
+        }
+        // Check and close replication producers.
         if (System.currentTimeMillis() - cachedTime > threshold * 1000L) {
             log.info().attr("brokerReplicationInactiveThresholdSeconds", threshold)
                     .log("Disconnecting replication producers since no producer is active for a long time.");
@@ -850,7 +856,8 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
                             log.warn("Attempting to add producer to a terminated topic");
                             throw new TopicTerminatedException("Topic was already terminated");
                         }
-                        return internalAddProducer(producer).thenApply(ignore -> {
+                        return internalAddProducer(producer).thenCompose(__ -> this.startReplProducers())
+                                .thenApply(ignore -> {
                             USAGE_COUNT_UPDATER.incrementAndGet(this);
                             log.debug()
                                     .attr("producerName", producer.getProducerName())
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -697,6 +697,11 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
                 persistentReplicator.disconnectIfNoTrafficAndBacklog();
             }
         }
+        for (Replicator replicator : getShadowReplicators().values()) {
+            if (replicator instanceof PersistentReplicator persistentReplicator) {
+                persistentReplicator.disconnectIfNoTrafficAndBacklog();
+            }
+        }
     }
 
     @Override
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -67,6 +67,7 @@ import org.apache.pulsar.broker.service.BrokerServiceException.ProducerBusyExcep
 import org.apache.pulsar.broker.service.BrokerServiceException.ProducerFencedException;
 import org.apache.pulsar.broker.service.BrokerServiceException.TopicMigratedException;
 import org.apache.pulsar.broker.service.BrokerServiceException.TopicTerminatedException;
+import org.apache.pulsar.broker.service.persistent.PersistentReplicator;
 import org.apache.pulsar.broker.service.persistent.PersistentTopic;
 import org.apache.pulsar.broker.service.plugin.EntryFilter;
 import org.apache.pulsar.broker.service.schema.SchemaRegistryService;
@@ -130,8 +131,6 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
 
     // Timestamp of when this topic was last seen active
     protected volatile long lastActive;
-    // Timestamp of when the latest producer was disconnected.
-    protected volatile Long localProducersEmptyTime;
 
     // Flag to signal that producer of this topic has published batch-message so, broker should not allow consumer which
     // doesn't support batch-message
@@ -667,60 +666,37 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
         return null;
     }
 
-    public abstract CompletableFuture<Void> closeReplProducersIfNoBacklog();
-
-    public abstract CompletableFuture<Void> startReplProducers();
-
-    public void disconnectReplicatorIfNoTrafficForLongTime() {
-        updateLocalProducersEmptyTime();
-
-        final Long cachedTime = localProducersEmptyTime;
-        // Still active.
-        if (cachedTime == null) {
-            return;
-        }
-        // Disabled the feature.
-        int threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds();
-        if (threshold <= 0) {
-            return;
-        }
-        // Check and close replication producers.
-        if (System.currentTimeMillis() - cachedTime > threshold * 1000L) {
-            log.info().attr("brokerReplicationInactiveThresholdSeconds", threshold)
-                    .log("Disconnecting replication producers since no producer is active for a long time.");
-            closeReplProducersIfNoBacklog().whenCompleteAsync((__, ex) -> {
-                // Here, "!=" is used instead of the "!equals()" to avoid the problem of "inability to compare due
-                // to multiple changes in this attribute within a short period of time".
-                if (cachedTime != localProducersEmptyTime) {
-                    log.warn().log("Restart replication producers since new producers were registered concurrently"
-                            + " when closing producers.");
-                    startReplProducers().whenComplete((ignore, ex2) -> {
-                        if (ex2 != null) {
-                            log.error().exception(ex2).log("Failed to start replication producers after registered"
-                                + " new producers");
-                        }
-                    });
-                }
-            });
-        }
-    }
-
-    protected void updateLocalProducersEmptyTime() {
-        // The timestamp has been set.
-        if (localProducersEmptyTime != null) {
-            return;
-        }
-        // Only contains remote producer | producers is empty: update the time.
-        for (Producer producer : producers.values()) {
-            if (!producer.isRemote()) {
-                return;
-            }
-        }
-        localProducersEmptyTime = System.currentTimeMillis();
-    }
-
     protected boolean hasProducersActive() {
         return !producers.isEmpty();
+    }
+
+    protected boolean hasActiveReplicators() {
+        for (Replicator replicator : getReplicators().values()) {
+            if (replicator.isConnected()) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    protected boolean hasLocalProducers() {
+        if (producers.isEmpty()) {
+            return false;
+        }
+        for (Producer producer : producers.values()) {
+            if (!producer.isRemote()) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public void disconnectReplicatorsIfNoTrafficAndBacklog() {
+        for (Replicator replicator : getReplicators().values()) {
+            if (replicator instanceof PersistentReplicator persistentReplicator) {
+                persistentReplicator.disconnectIfNoTrafficAndBacklog();
+            }
+        }
     }
 
     @Override
@@ -856,12 +832,7 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
                             log.warn("Attempting to add producer to a terminated topic");
                             throw new TopicTerminatedException("Topic was already terminated");
                         }
-                        CompletableFuture<Void> internalAppProducerFuture = internalAddProducer(producer);
-                        internalAppProducerFuture.thenApply(__ -> this.startReplProducers()).exceptionally(ex -> {
-                            log.error("Failed to start replication producers.");
-                            return null;
-                        });
-                        return internalAppProducerFuture.thenApply(ignore -> {
+                        return internalAddProducer(producer).thenApply(ignore -> {
                             USAGE_COUNT_UPDATER.incrementAndGet(this);
                             log.debug()
                                     .attr("producerName", producer.getProducerName())
@@ -1065,7 +1036,6 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
         if (existProducer != null) {
             return tryOverwriteOldProducer(existProducer, producer);
         } else if (!producer.isRemote()) {
-            localProducersEmptyTime = null;
             USER_CREATED_PRODUCER_COUNTER_UPDATER.incrementAndGet(this);
         }
         return CompletableFuture.completedFuture(null);
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
@@ -130,6 +130,8 @@ public abstract class AbstractTopic implements Topic, TopicPolicyListener {
 
     // Timestamp of when this topic was last seen active
     protected volatile long lastActive;
+    // Timestamp of when the latest producer was disconnected.
+    protected volatile Long localProducersEmptyTime;
 
     // Flag to signal that producer of this topic has published batch-message so, broker should not allow consumer which
     // doesn't support batch-message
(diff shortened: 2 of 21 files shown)


pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
```diff
@@ -812,6 +812,14 @@ The max allowed delay for delayed delivery (in milliseconds). If the broker rece
     )
     private Integer brokerDeleteInactiveTopicsMaxInactiveDurationSeconds = null;
 
+    @FieldContext(
+        category = CATEGORY_POLICIES,
+        dynamic = true,
+        doc = "Time in seconds after which a replicator is considered inactive when no replication activity is"
+                + " observed."
```
lhotari: Please improve the documentation of this configuration setting. Make it explicit that `brokerReplicationInactiveThresholdSeconds` is related to `brokerDeleteInactiveTopicsEnabled`.
For example:
```java
        doc = "Time in seconds since a replicator's producer last sent a message, after which the replicator"
                + " is considered inactive. When inactive-topic deletion is enabled (via"
                + " brokerDeleteInactiveTopicsEnabled, or enabled by namespace/topic policy), a replicated"
                + " topic is eligible for deletion only once all its replicators have been inactive for this"
                + " duration."
```

codelipenghui (commented):
@poorbarcode I'm thinking another solution

1. Close the replicator producer if the producer is idle for a while (e.g. 10 mins)
2. Check all the producers for detecting the inactive topic (now it's more complicated to skip the replicator producer)
3. Only delete the inactive topic if there is no producers (including the replicator producer)

Now, your solution added 7 days delay for inactive topic deletion if geo-replication enable. But if there is no messages from last 7 days, the issue can still happen, right?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
```diff
@@ -665,13 +665,19 @@ protected Consumer getActiveConsumer(Subscription subscription) {
...
         if (producers.isEmpty()) {
             return false;
         }
         for (Producer producer : producers.values()) {
             if (!producer.isRemote()) {
                 return true;
+            } else {
+                // If the remote producer still replicating messages, do not delete the topic.
+                if (System.currentTimeMillis() - producer.getLatestPublishTime() < (brokerService.getPulsar()
+                        .getConfig().getBrokerReplicationInactiveThresholdSeconds() * 1000)) {
```
void-ptr974: Minor: should this use `1000L` or `TimeUnit.SECONDS.toMillis(...)`?

Since `brokerReplicationInactiveThresholdSeconds` is an `Integer`, `* 1000` is evaluated as int arithmetic before the comparison. For larger configured values, the millisecond threshold could overflow and make this active check incorrect.

void-ptr974:
> @poorbarcode I'm thinking another solution
> 
> 1. Close the replicator producer if the producer is idle for a while (e.g. 10 mins)
> 2. Check all the producers for detecting the inactive topic (now it's more complicated to skip the replicator producer)
> 3. Only delete the inactive topic if there is no producers (including the replicator producer)
> 
> Now, your solution added 7 days delay for inactive topic deletion if geo-replication enable. But if there is no messages from last 7 days, the issue can still happen, right?

I agree with this concern. A replicated topic can be quiet for longer than the threshold while the replication relationship is still valid.

Using `latestPublishTime` here makes topic deletion depend on the traffic pattern instead of the producer lifecycle. It seems cleaner to let the replicator close its producer explicitly when it is really idle, and let topic GC only check whether producers still exist.

pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
poorbarcode: Changed the solution as @codelipenghui suggested here:https://github.com/apache/pulsar/pull/25915#pullrequestreview-4405441793,  and Improved the doc

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
poorbarcode: Thanks , correct the code

poorbarcode:
@codelipenghui @void-ptr974 Changed the solution as @codelipenghui suggested, please review again

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
```diff
@@ -741,6 +741,10 @@ protected void startInactivityMonitor() {
             int interval = pulsar().getConfiguration().getBrokerDeleteInactiveTopicsFrequencySeconds();
             inactivityMonitor.scheduleAtFixedRateNonConcurrently(() -> checkGC(), interval, interval,
                     TimeUnit.SECONDS);
+            if (pulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds() > 0) {
```
lhotari, line 811: `brokerReplicationInactiveThresholdSeconds` is type of `Integer` in `ServiceConfiguration`. Technically it can be set to `null`. Therefore a null check could be useful.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
```diff
@@ -665,16 +667,54 @@ protected Consumer getActiveConsumer(Subscription subscription) {
...
+
+    public void disconnectReplicatorIfNoTrafficForLongTime() {
+        updateLocalProducersEmptyTime();
+
+        final Long cachedTime = localProducersEmptyTime;
+        if (cachedTime == null) {
+            return;
         }
+        int threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds();
+        if (System.currentTimeMillis() - cachedTime > threshold * 1000L) {
```
lhotari: since `brokerReplicationInactiveThresholdSeconds` is a dynamic configuration, it's possible to change the value dynamically. It would be expected that setting it to `0` would disable the check. This condition would evaluate to true for all replicators if that would be performed.
One possible way to address the problem is to make the setting non-dynamic with `dynamic = false`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
lhotari: Another detail is that the field has `dynamic = true`. The scheduled task wouldn't get cancelled and rescheduled with the updated value. Again, setting to `dynamic = false` would be a way to address this without adding support for dynamic configuration of `brokerReplicationInactiveThresholdSeconds`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/nonpersistent/NonPersistentTopic.java
```diff
@@ -598,12 +598,18 @@ public CompletableFuture<Void> close(
         return closeFuture;
     }
 
-    public CompletableFuture<Void> stopReplProducers() {
+    public CompletableFuture<Void> closeReplProducersIfNoBacklog() {
```
lhotari: Local Claude Code review comment:

> ### `NonPersistentTopic` idle-disconnect uses `terminate()`, which can never be restarted
> 
> For `PersistentTopic`, `closeReplProducersIfNoBacklog()` calls `replicator.disconnect()` → `State.Disconnected`, and `startProducer()` revives from `Disconnected` (`AbstractReplicator#startProducer`). So the disconnect→reconnect cycle this PR relies on works.
> 
> For `NonPersistentTopic`, the renamed `closeReplProducersIfNoBacklog()` still calls `replicator.terminate()` → `State.Terminated`. That breaks the reconnect path in two ways:
> 
> - `startProducer()` is a **no-op** from `Terminated` (`AbstractReplicator#startProducer`: *"Skip the producer creation since the replicator is terminating"*), so neither the `startReplProducers()` safety net nor `addProducer()` can revive it.
> - `terminate()` does **not** remove the replicator from the `replicators` map (only `removeReplicator()` does). `checkReplication()` only recreates *missing* replicators (`!replicators.containsKey(cluster)`), so it won't recreate the terminated-but-present one either.
> 
> Net effect: once a non-persistent replicated topic that has a remote producer (but no local producers) goes idle past `brokerReplicationInactiveThresholdSeconds`, `disconnectReplicatorIfNoTrafficForLongTime()` terminates its outbound replicators **permanently** — replication stays broken until the topic is unloaded/reloaded, even if a local producer reconnects later. This violates the disconnect→reconnect design the PR is built on.
> 
> The previous `terminate()` usage was always immediately followed by `delete()`, so the permanent state didn't matter; this PR introduces a terminate-without-delete path where it does.
> 
> The two new tests cover only the persistent topic; the non-persistent path is untested.
> 
> **Suggestion:** make `NonPersistentTopic.closeReplProducersIfNoBacklog()` use reconnectable (`disconnect()`) semantics consistent with `PersistentTopic`, or remove+recreate the replicator, and add a non-persistent test for the idle disconnect→reconnect cycle.
>

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/BrokerService.java
poorbarcode: > brokerReplicationInactiveThresholdSeconds is type of Integer in ServiceConfiguration. Technically it can be set to null. Therefore a null check could be useful.

Changed the type to `int`

> Another detail is that the field has dynamic = true. The scheduled task wouldn't get cancelled and rescheduled with the updated value. Again, setting to dynamic = false would be a way to address this without adding support for dynamic configuration of brokerReplicationInactiveThresholdSeconds.

The scheduled interval depends on another configuration `brokerDeleteInactiveTopicsFrequencySeconds`, Once the value of `brokerReplicationInactiveThresholdSeconds`, the behavior will change, the test `testDisconnectAndReconnectInactiveReplicator` proved the behavior

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
poorbarcode: Good pointer, added a check: it only works when threshold is larger than `0`

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/nonpersistent/NonPersistentTopic.java
poorbarcode: Good pointer, the function of non-persistentTopic was indeed prone to being overlooked, but it has now been fixed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
```diff
@@ -810,7 +856,8 @@ public CompletableFuture<Optional<Long>> addProducer(Producer producer,
                             log.warn("Attempting to add producer to a terminated topic");
                             throw new TopicTerminatedException("Topic was already terminated");
                         }
-                        return internalAddProducer(producer).thenApply(ignore -> {
+                        return internalAddProducer(producer).thenCompose(__ -> this.startReplProducers())
```
void-ptr974: This introduces a race in producer registration.

After `internalAddProducer()` succeeds, the producer is already in `producers`, but `usageCount` is only incremented after `startReplProducers()` completes. For persistent topics, that call can wait on an async policy lookup.

During this window, `PersistentTopic.delete()` can still see `currentUsageCount() == 0` and treat the topic as idle, even though a producer has already been registered.

Please keep the previous ordering and increment `usageCount` immediately after `internalAddProducer()` succeeds, before starting repl producers.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
```diff
@@ -810,7 +856,12 @@ public CompletableFuture<Optional<Long>> addProducer(Producer producer,
                             log.warn("Attempting to add producer to a terminated topic");
                             throw new TopicTerminatedException("Topic was already terminated");
                         }
-                        return internalAddProducer(producer).thenApply(ignore -> {
+                        CompletableFuture<Void> internalAppProducerFuture = internalAddProducer(producer);
+                        internalAppProducerFuture.thenApply(__ -> this.startReplProducers()).exceptionally(ex -> {
```
void-ptr974: Would it be possible to narrow this trigger a bit?

This runs for every producer registration, including remote producers and additional local producers when the replicators may already be running. For persistent topics, startReplProducers() also does an async namespace policy lookup, so this adds work to the producer creation path.

Maybe we can check whether this is a non-remote producer and whether localProducersEmptyTime was non-null before resetting it. In that case, startReplProducers() would only run when a local producer brings the topic back from the no-local-producer state.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractTopic.java
```diff
@@ -665,16 +667,60 @@ protected Consumer getActiveConsumer(Subscription subscription) {
...
+        // Disabled the feature.
+        int threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds();
+        if (threshold <= 0) {
+            return;
+        }
+        // Check and close replication producers.
+        if (System.currentTimeMillis() - cachedTime > threshold * 1000L) {
+            log.info().attr("brokerReplicationInactiveThresholdSeconds", threshold)
+                    .log("Disconnecting replication producers since no producer is active for a long time.");
+            closeReplProducersIfNoBacklog().whenCompleteAsync((__, ex) -> {
```
void-ptr974: One concern here is the case where a replicated topic only has remote producers.

For example, cluster A may keep replicating messages to cluster B, while B has no local producers. In that state, Topic GC now considers B's topic active because the remote producer is connected, but this idle check can still disconnect B's outbound replicator after the threshold. If no local producer later connects to B, there may not be a clear event to start that replicator again, so the replication cursor could stop advancing and backlog could grow.

poorbarcode: The previous solution solved the issue you mentioned; your suggestion here https://github.com/apache/pulsar/pull/25915#issuecomment-4597972779 introduced a regression

void-ptr974: Thanks for clarifying. I see your point, and I agree the previous traffic-based approach can cover the case where remote traffic keeps arriving continuously.

I think we are mainly discussing which signal should be used as the invariant here. The traffic-based signal and the producer-lifecycle signal cover slightly different cases:

- If remote messages keep arriving, traffic-based detection works well.
- If the remote producer connection is still valid but the topic is quiet for a long time, traffic-based detection may still treat the topic as idle.

So I think moving toward producer-lifecycle based detection is reasonable and closer to the original issue.

One detail I want to double-check is that the PR description says the replicator producer is closed after there is “no producer registered” for a long time. That sounds like it includes both local and remote producers. The current implementation uses “no local producers” as the timer condition. In the backup-side scenario, a topic can have no local producer but still have a connected remote producer, so the current condition seems a little broader than the description.

Would it be cleaner to align the implementation with the PR description?

- Topic GC treats any producer, including remote producers, as active.
- The inactive-replication timer starts only when there are no producers at all.
- startReplProducers() only runs when the topic transitions from no producers to having a producer again.

Concretely, localProducersEmptyTime could become something like producersEmptyTime. The timer would be set only when producers.isEmpty(), cleared when a producer is added, and startReplProducers() would only run when a producer is added after the timer had been set.

This keeps the current producer-lifecycle direction, avoids relying only on recent message traffic, and also keeps the idle-replication condition consistent with the Topic GC condition.

poorbarcode:
@codelipenghui @lhotari @void-ptr974 

Please review the PR again. I have changed the solution: 
- Topic GC will only be triggered if there is no producer(includes remote producer) and no replicator
- Replicator producer will be closed after there are no messages to be replicated, it does not rely on the Persistent Topic anymore

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -3355,7 +3313,7 @@ public boolean isActive(InactiveTopicDeleteMode deleteMode) {
                 break;
         }
         // no local producers
-        return hasLocalProducers();
+        return hasProducersActive() || hasActiveReplicators();
```
void-ptr974, line 3335: I think this still needs to account for replication backlog, not only connected replicators.

Trigger scenario:
1. A replicator is idle-disconnected when its backlog is 0.
2. New messages are written later, so the replication cursor gets backlog.
3. The replicator is still `Disconnected`, so `hasActiveReplicators()` returns false.
4. The producer closes and the topic has no subscriptions/producers/connected replicators.
5. `checkGC()` can enter the delete path.

Impact:
This can delete the topic ledger while messages are still pending replication to the remote cluster, which is a data-loss risk.

Suggested fix:
Please make inactive-topic GC explicitly protect replication backlog. For example, this activity check could include `isReplicationBacklogExist()`, and the GC delete path should also have a final guard before deleting so the topic cannot be removed while any replication cursor has backlog.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -360,20 +378,64 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
+            }, 100, TimeUnit.MILLISECONDS);
+        };
+
+        // Retry.
+        if (state == Disconnecting ||  state == Starting) {
+            retryReplicateEntries.run();
+            return;
+        }
+        // Start producer and retry.
+        if (state == Disconnected) {
```
void-ptr974, line 400: This recovery path depends on reaching `replicateEntries()`, but a disconnected replicator may not always have a pending read callback that gets here.

Trigger scenario:
1. `disconnectIfNoTrafficAndBacklog()` disconnects the producer while backlog is 0.
2. At that moment there is no pending `asyncReadEntriesOrWait`, or reads are blocked by rate limits, permits, schema fetch, backoff, etc.
3. New messages are written and replication backlog appears.
4. Since the state is `Disconnected`, `readMoreEntries()` may not create a normal read task, and `PersistentTopic.addProducer()` no longer starts replication producers.
5. The replicator can stay `Disconnected`.

Impact:
Replication can stall and backlog can keep growing until some external event reloads/recreates the topic or otherwise restarts the replicator. Combined with inactive-topic GC, this also increases the risk of deleting unreplicated data.

Suggested fix:
Please add a deterministic resume path outside this read callback. For example, the periodic inactive-replication check could restart a disconnected replicator when backlog appears:

```java
if (state == Disconnected && getNumberOfEntriesInBacklog() > 0) {
    startProducer();
    return;
}

File: pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java (hunk 9 above)
Comment (poorbarcode) [reply pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java:3335]: > New messages are written later, so the replication cursor gets backlog.

Since new messages were written, a new producer will be registered and wakeup replicators

File: pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java (hunk 10 above)
Comment (poorbarcode) [reply pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java:400]: > At that moment there is no pending asyncReadEntriesOrWait, or reads are blocked by rate limits, permits, schema fetch, backoff, etc.

This situation does not exist. When there is no backlog, there will definitely be a `asyncReadEntriesOrWait`

File: pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java [hunk 11]
```diff
@@ -360,20 +378,64 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
+            startProducer();
+            retryReplicateEntries.run();
+            return;
```
lhotari, line 403: **[BUG] Auto-resume can replicate newer positions before older ones → reordering, and silent message loss when dedup is enabled**

After an idle disconnect, this resume path can replicate a *newer* batch before the older one it is holding, which breaks replication ordering and — with deduplication enabled — silently drops the older message.

Sequence: when a wait-read completes while `state == Disconnected`, this branch holds that batch (call it R1, the older positions) and reschedules it ~100ms later, while also calling `startProducer()`. `startProducer()` → `setProducerAndTriggerReadEntries()` flips the state to `Started` and immediately calls `readMoreEntries()`:

https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L156-L180

Because R1's `InFlightTask` already has `entries != null`, it is no longer counted as a pending read, so a fresh read R2 (strictly newer positions — the cursor read position already advanced past R1 and is no longer rewound) is issued and can be sent by `doReplicateEntries(R2)` before R1's 100ms-deferred retry fires.

Two consequences:

- Replication order across the reconnect can be violated.
- The receiving cluster dedups replicated messages by *source position*, not wire order:

https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/MessageDeduplication.java#L389-L396

  Once R2's higher position is recorded, R1 (lower position) is classified a duplicate and discarded on the remote, while the source side mark-deletes R1 as delivered on the (successful) send callback:

https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L463-L479

  i.e. permanent silent message loss across the reconnect. (Loss requires topic dedup enabled; without it the symptom is out-of-order replication, which is still a correctness issue for ordered consumers.)

The previous design avoided this: `cursor.rewind()` on (re)connect plus pending-read cancellation guaranteed in-order resume from `markDelete + 1`. Could we hold the new read until the held batch has been replicated (or rewind / cancel the pending read on disconnect) so resume is strictly in order?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -360,20 +378,64 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
+
+        // Retry to replicate messages if it is not started.
+        ManagedLedgerImpl ml = (ManagedLedgerImpl) cursor.getManagedLedger();
+        Runnable retryReplicateEntries = () -> {
+            ml.getScheduledExecutor().schedule(() -> {
+                ml.getExecutor().execute(() -> {
+                    replicateEntries(entries, inFlightTask, skippedReadAfterSent);
+                });
+            }, 100, TimeUnit.MILLISECONDS);
+        };
```
lhotari, line 392: **[PERF] The 100ms retry loop defeats the connection backoff and pins a read batch in memory during a remote outage**

This retry primitive reschedules the same `entries` batch every 100ms with no backoff and no cap, and the batch is held (not released) for the whole time the replicator is not `Started`. Two concerns when the remote cluster is unreachable:

1. **The connection backoff is defeated.** On the `Disconnected` branch just below, every 100ms tick also calls `startProducer()`. Its own failure path returns the state to `Disconnected` with an exponential connection backoff (100ms → 1min):

https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/AbstractReplicator.java#L219-L228

   But this 100ms loop re-triggers `startProducer()` as soon as the state is `Disconnected` again, so a down remote is hit at roughly 10 connection attempts/sec per replicator instead of throttling toward ~1/min.

2. **Entry pinning.** The held batch is `ByteBuf`-backed (up to `dispatcherMaxReadBatchSize`) and those positions stay in the backlog for the entire outage; across many replicators/topics this is sustained direct-memory pressure. The previous design released the entries / rewound the cursor on disconnect rather than holding the batch.

Could the retry go through the existing `Backoff`, and release/return the batch when it can't be sent promptly?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -283,6 +277,30 @@ private AvailablePermits getRateLimiterAvailablePermits(int availablePermits) {
...
+        // Disabled the feature.
+        int threshold = brokerService.getPulsar().getConfig().getBrokerReplicationInactiveThresholdSeconds();
+        if (threshold <= 0) {
+            return;
+        }
+        // Has backlog.
+        long backlog = getNumberOfEntriesInBacklog();
+        if (backlog > 0) {
+            return;
+        }
```
lhotari, line 290: **[QUALITY] Loss-safety now rests entirely on the `backlog == 0` precondition — worth documenting the invariant**

It would help to document (a comment, or an assertion) that this `backlog == 0` precondition is now load-bearing for correctness. Since both the `cursor.rewind()` on reconnect and the `afterDisconnected()` rewind were removed, nothing re-reads entries that are read-but-not-yet-acked at producer restart. No-loss safety now rests entirely on (a) this `backlog == 0` check and (b) `beforeDisconnect()` rejecting any in-flight task below the last confirmed entry:

https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L988-L999

There's no construct of a loss case under these guards currently, so this is fine as written — but the invariant is implicit and easy to break. A future disconnect trigger that doesn't enforce `backlog == 0`, or a regression in `beforeDisconnect()`, would silently drop those entries, with no rewind left as a safety net. A short comment stating the requirement would make that explicit.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
```diff
@@ -159,9 +159,6 @@ private void receivedSnapshotRequest(ReplicatedSubscriptionsSnapshotRequest requ
                             + "configured for that cluster. Ignoring the request.");
             return;
         }
-        if (!replicator.isConnected()) {
-            topic.startReplProducers();
```
lhotari, line 163: **[QUALITY] Stale comment: it still claims the replicator producer is restarted here, but that code was removed**

With this block removed, the comment a few lines up in `receivedSnapshotRequest` is now stale — it still claims the producer is restarted here:

https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java#L151-L153

The behavior is fine — writing the snapshot-response marker self-heals the producer via the `Disconnected` branch in `replicateEntries()` — but the comment should be updated to describe that self-healing path instead of an explicit restart that no longer exists.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
```diff
@@ -235,6 +246,207 @@ public void testDeleteTopicWhenReplicating() throws Exception {
...
+                assertFalse(persistentTopic1.getReplicators().isEmpty());
+                PersistentReplicator persistentReplicator =
+                        (PersistentReplicator) persistentTopic1.getReplicators().get(cluster2);
+                assertTrue(persistentReplicator.isConnected());
+                assertEquals(persistentReplicator.getState(), AbstractReplicator.State.Started);
+                Thread.sleep(1000);
+            }
+        } else {
+            // TODO terminate 的判断逻辑搬运到 replicator 中。
+            //  resume 的判断逻辑可言在 scheduled task 中，也可以在 cursor pending read 中。
```
lhotari: **[QUALITY] Leftover non-English TODOs and commented-out scaffolding in the new test**

These look like leftover development notes — both are non-English (the repo convention is English comments) and read as unfinished TODOs. Could they be translated or removed before merge?

The same method carries other leftover scaffolding worth cleaning up too: commented-out assertions referencing a `getLatestPublishTime()` accessor that doesn't exist anywhere in the tree (lines 349/351/364/366), and a commented-out "Verify: resume" block (lines 392-398). If the per-producer idle-time checks were intended they'd directly cover the `latestPublishTime` logic this PR adds, so it'd be worth implementing them; otherwise they should be dropped.

void-ptr974 (approved):
Thanks for the discussion and updates. The current approach looks reasonable to me.

pulsar-broker/src/test/java/org/apache/pulsar/broker/service/OneWayReplicatorTest.java
poorbarcode: Removed

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java
poorbarcode: Removed java doc outdated

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
lhotari: There's also a replicator dedup related PR in #25860 by @void-ptr974

poorbarcode: > Because R1's InFlightTask already has entries != null, it is no longer counted as a pending read, so a fresh read R2 (strictly newer positions — the cursor read position already advanced past R1 and is no longer rewound) is issued and can be sent by doReplicateEntries(R2) before R1's 100ms-deferred retry fires.

Could you point why the `entries` is not `null`?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
poorbarcode: Improved the case

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
poorbarcode: I can not understand what concern is it

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
lhotari: @poorbarcode — `entries` is set as the very first action of the read callback, before any state is checked. In [`readEntriesComplete`](https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L360-L365):

```java
InFlightTask inFlightTask = (InFlightTask) ctx;
inFlightTask.setEntries(entries);   // entries becomes non-null here
```

So by the time we get into `replicateEntries` and hit the [`Disconnected` branch](https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L423-L427) (`startProducer(); retryReplicateEntries.run(); return;`), `inFlightTask.entries` is already populated, and that branch reschedules R1 ~100 ms later **without resetting `entries` to null**. R1 also can't be recycled while it's parked (`isDone()` is false: `completedEntries == 0 < entries.size()`), so it stays in `inFlightTasks` with non-null `entries`.

That's what removes the gate: the "pending read" test is exactly `readPos != null && entries == null` in both [`getPermitsIfNoPendingRead` (L961)](https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L958-L969) and [`hasPendingRead` (L1054)](https://github.com/apache/pulsar/blob/7128f360d300bb1f855690625f06ae4681ce4699/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java#L1049-L1061). With `entries != null`, R1 no longer counts as a pending read, so when `startProducer()` flips the state to `Started` and `setProducerAndTriggerReadEntries()` calls `readMoreEntries()`, a fresh read R2 is issued at the advanced `cursor.getReadPosition()` — and R2 can be sent before R1's deferred retry fires.

Worth noting: the GC/idle `disconnect()` path goes through `beforeDisconnect()` (backlog check only), so unlike `beforeTerminateOrCursorRewinding()` it does **not** `cancelPendingReadRequest()` or rewind the cursor — which is also why R1's wait-read survives the producer close and completes while `Disconnected` in the first place. This is essentially what the `cursor.rewind()` + pending-read cancellation that this PR removed from `setProducerAndTriggerReadEntries` used to prevent.

On the consequence: repl dedup is a high-water-mark keyed by source position (`isDuplicateReplV2`), so R2 (newer) being recorded before R1 (older) is enough — once the mark is past R1, R1 comes back as a duplicate. If R2 is already persisted, R1 gets `Dup`, which `PersistentTopic` acks as success, so the source mark-deletes it → silent loss; if R2 is only pushed-not-persisted, R1 gets `Unknown` and the source retries (recoverable). With dedup off it's out-of-order delivery. The simplest fix is to keep resume strictly in order — don't let the deferred retry leave a held batch counting as "not a pending read" while a fresh read is issued ahead of it (e.g. gate the new read on the held task, or rewind/cancel on disconnect as before).

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
```diff
@@ -360,20 +378,71 @@ public void readEntriesComplete(List<Entry> entries, Object ctx) {
...
+
+        // Retry to replicate messages if it is not started.
+        ManagedLedgerImpl ml = (ManagedLedgerImpl) cursor.getManagedLedger();
+        Runnable retryReplicateEntries = () -> {
+            long estimatedTimeStampProducerConnected = this.estimatedTimeStampProducerConnected;
+            long delayMillis;
+            if (estimatedTimeStampProducerConnected > System.currentTimeMillis()) {
+                delayMillis = (estimatedTimeStampProducerConnected - System.currentTimeMillis()) + 100;
```
lhotari, line 383: could delayMillis be negative?

poorbarcode: line-412 prevents the situation you mentioned

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
poorbarcode: Thanks for the explanation, I should let the check to be in `readCompete`, rather than let it to be in `replicatEntries`.

Fixed

---

> There's also a replicator dedup related PR in https://github.com/apache/pulsar/pull/25860 by @void-ptr974

I will review it tomorrow ❤️

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentReplicator.java
poorbarcode: It is line-382 now: https://github.com/apache/pulsar/pull/25915/changes#diff-16c47c755ed17a62606a6b9a6b1dc65d5fe0baed21f6e3541df264bec81b83f6R382

gaoran10:
Please check the code style.

gaoran10 (approved):

PR #25916, opened 2026-06-02
[fix][client] Clean up unacked messages when unsubscribing a topic with ack timeout backoff

Dream95:
### Motivation
When a multi-topic consumer unsubscribes from a single topic, the client should drop that topic's unacked messages from the ack-timeout tracker so they are not redelivered later.
`MultiTopicsConsumerImpl` already called `removeTopicMessages` for `UnAckedTopicMessageTracker`, but not for `UnAckedTopicMessageRedeliveryTracker`, which is used when `ackTimeoutRedeliveryBackoff` is configured . Unsubscribed topics could therefore leave stale entries in the redelivery partition map and in `ackTimeoutMessages`.
Additionally, `UnAckedTopicMessageRedeliveryTracker.removeTopicMessages` used the partition-map iterator when iterating `ackTimeoutMessages` (`iterator.hasNext()` / `iterator.remove()` instead of `iteratorAckTimeOut`), so ack-timeout entries were not removed correctly and internal tracker state could be corrupted.
### Modifications
- Call `removeTopicMessages` on `UnAckedTopicMessageRedeliveryTracker` when a topic is unsubscribed from a multi-topic consumer.
- Fix ack-timeout iteration in `UnAckedTopicMessageRedeliveryTracker.removeTopicMessages` to use `iteratorAckTimeOut`.
- Add `UnAckedTopicMessageRedeliveryTrackerTest` to verify `removeTopicMessages` clears both the partition map and ack-timeout map.
### Verifying this change
- [x] `./gradlew :pulsar-client-original:test --tests org.apache.pulsar.client.impl.UnAckedTopicMessageRedeliveryTrackerTest`

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

PR in forked repository: https://github.com/Dream95/pulsar/pull/12

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java
@@ -1302,7 +1302,7 @@ public class MultiTopicsConsumerImpl<T> extends ConsumerBase<T> {
                     removeTopic(topicName);
                     if (unAckedMessageTracker instanceof UnAckedTopicMessageTracker tracker) {
                         tracker.removeTopicMessages(topicName);
-                    }else if (unAckedMessageTracker instanceof UnAckedTopicMessageRedeliveryTracker tracker){
+                    } else if (unAckedMessageTracker instanceof UnAckedTopicMessageRedeliveryTracker tracker){
                         tracker.removeTopicMessages(topicName);
                     }
 
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java
@@ -1300,8 +1300,10 @@ public class MultiTopicsConsumerImpl<T> extends ConsumerBase<T> {
                     });
 
                     removeTopic(topicName);
-                    if (unAckedMessageTracker instanceof UnAckedTopicMessageTracker) {
-                        ((UnAckedTopicMessageTracker) unAckedMessageTracker).removeTopicMessages(topicName);
+                    if (unAckedMessageTracker instanceof UnAckedTopicMessageTracker tracker) {
+                        tracker.removeTopicMessages(topicName);
+                    }else if (unAckedMessageTracker instanceof UnAckedTopicMessageRedeliveryTracker tracker){
+                        tracker.removeTopicMessages(topicName);
                     }
 
                     unsubscribeFuture.complete(null);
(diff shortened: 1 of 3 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java
```diff
@@ -1300,8 +1300,10 @@ public CompletableFuture<Void> unsubscribeAsync(String topicName) {
                     });
 
                     removeTopic(topicName);
-                    if (unAckedMessageTracker instanceof UnAckedTopicMessageTracker) {
-                        ((UnAckedTopicMessageTracker) unAckedMessageTracker).removeTopicMessages(topicName);
+                    if (unAckedMessageTracker instanceof UnAckedTopicMessageTracker tracker) {
+                        tracker.removeTopicMessages(topicName);
+                    } else if (unAckedMessageTracker instanceof UnAckedTopicMessageRedeliveryTracker tracker){
+                        tracker.removeTopicMessages(topicName);
                     }
```
lhotari, line 1307: `UnAckedTopicMessageTracker extends UnAckedMessageTracker`, therefore it seems that this change is not necessary.

lhotari: oh sorry, `UnAckedTopicMessageRedeliveryTracker` is different.

Dream95: They share the same base class , but removeTopicMessages is not declared on that base type.

lhotari (approved):
LGTM, good work @Dream95

void-ptr974 (approved):
LGTM. Thanks for the fix!

The missing cleanup for UnAckedTopicMessageRedeliveryTracker looks correct, and the iterator fix in removeTopicMessages is important.

One non-blocking note: the current getOwnerTopic().contains(topicName) matching seems a bit loose and may match similar topic names, such as my-topic and my-topic-v2. This is pre-existing behavior, so it shouldn’t block this PR.

Would you be interested in addressing this in a follow-up PR?

Dream95:
> LGTM. Thanks for the fix!
> 
> The missing cleanup for UnAckedTopicMessageRedeliveryTracker looks correct, and the iterator fix in removeTopicMessages is important.
> 
> One non-blocking note: the current getOwnerTopic().contains(topicName) matching seems a bit loose and may match similar topic names, such as my-topic and my-topic-v2. This is pre-existing behavior, so it shouldn’t block this PR.
> 
> Would you be interested in addressing this in a follow-up PR?

Thanks for the review. I'll fix that in a follow-up PR.

PR #25921, opened 2026-06-03
[fix][client] Match logical topic when removing unacked messages

Dream95:
### Motivation
When a multi-topic consumer unsubscribes from a topic, it removes unacked messages that belong to the unsubscribed topic.
The previous matching logic used substring matching on `TopicMessageId#getOwnerTopic()`. This could incorrectly remove messages from topics with similar prefixes, such as matching `my-topic-v2` when unsubscribing `my-topic`.
### Modifications
- Normalize both the message owner topic and the target topic to their partitioned topic name before comparing them.
- Remove unacked messages for all partitions of the unsubscribed logical topic.
- Add test coverage to verify that similarly prefixed topics are not removed by mistake.

### Verifying this change

`./gradlew :pulsar-client-original:test --tests org.apache.pulsar.client.impl.UnAckedTopicMessageRedeliveryTrackerTest`

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TopicMessageId.java
+++ b/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TopicMessageId.java
@@ -41,6 +41,19 @@ public interface TopicMessageId extends MessageId {
      */
     String getOwnerTopic();
 
+    /**
+     * Checks if this message's owner topic and the given topic refer to the same base
+     * partitioned topic by comparing their base partitioned topic names.
+     *
+     * <p>For example, {@code persistent://public/default/my-topic-partition-0} matches
+     * {@code persistent://public/default/my-topic} or any other partition of that topic.
+     * Topics sharing only a name prefix (e.g., {@code my-topic} vs {@code my-topic-v2}) do not match.
+     *
+     * @param topicName a full topic name (non-partitioned, partitioned, or specific partition)
+     * @return {@code true} if both topics resolve to the same base partitioned topic name
+     */
+    boolean hasSameBasePartitionedTopic(String topicName);
+
     static TopicMessageId create(String topic, MessageId messageId) {
         if (messageId instanceof TopicMessageId) {
             return (TopicMessageId) messageId;
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TopicMessageIdImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TopicMessageIdImpl.java
@@ -23,6 +23,7 @@ import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageIdAdv;
 import org.apache.pulsar.client.api.TopicMessageId;
 import org.apache.pulsar.client.api.TraceableMessageId;
+import org.apache.pulsar.common.naming.TopicName;
 
 public class TopicMessageIdImpl implements MessageIdAdv, TopicMessageId, TraceableMessageId {
     private static final long serialVersionUID = 1L;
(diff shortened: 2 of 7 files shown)


pulsar-client/src/main/java/org/apache/pulsar/client/impl/UnAckedMessageTracker.java
```diff
@@ -304,4 +306,12 @@ private void stop() {
     public void close() {
         stop();
     }
+
+    protected static boolean messageBelongsToTopic(MessageId messageId, String topicName) {
```
lhotari: the name of this method is vague. It would be useful to add a javadoc to explain the contract.
Perhaps moving the method as an instance method of `TopicMessageId` would make more sense.
Naming the method continues to be a challenge also in that case, but it could make it slightly easier since there's proper context.

lhotari: One reason why naming is hard is the fact that the concepts of "owner topic" and "partitioned topic name" aren't defined properly. For example, the description `getOwnerTopic` in `TopicMessageId` is "Return the owner topic name of a message". The concept of "owner topic" isn't described.
The situtation with "partitioned topic name" is better. It's defined as "For partitions in a topic, return the base partitioned topic name."

Perhaps this gives some ideas for the instance method naming. 
For example `boolean hasSameBasePartitionedTopic(String topicName)`.

pulsar-client-api/src/main/java/org/apache/pulsar/client/api/TopicMessageId.java
```diff
@@ -41,6 +41,19 @@ public interface TopicMessageId extends MessageId {
...
+     * partitioned topic by comparing their base partitioned topic names.
+     *
+     * <p>For example, {@code persistent://public/default/my-topic-partition-0} matches
+     * {@code persistent://public/default/my-topic} or any other partition of that topic.
+     * Topics sharing only a name prefix (e.g., {@code my-topic} vs {@code my-topic-v2}) do not match.
+     *
+     * @param topicName a full topic name (non-partitioned, partitioned, or specific partition)
+     * @return {@code true} if both topics resolve to the same base partitioned topic name
+     */
+    boolean hasSameBasePartitionedTopic(String topicName);
```
lhotari, line 55: add the current implementation in `TopicMessageIdImpl` as the default implementation of this method.

pulsar-client/src/main/java/org/apache/pulsar/client/impl/TopicMessageIdImpl.java
```diff
@@ -92,6 +93,12 @@ public String getOwnerTopic() {
         return ownerTopic;
     }
 
+    @Override
+    public boolean hasSameBasePartitionedTopic(String topicName) {
+        return TopicName.get(getOwnerTopic()).getPartitionedTopicName()
+                .equals(TopicName.get(topicName).getPartitionedTopicName());
+    }
```
lhotari, line 100: move to TopicMessageId as the default implementation

Dream95: > move to TopicMessageId as the default implementation

Since TopicMessageId lives in the pulsar-client-api module while TopicName lives in pulsar-common, and pulsar-common already depends on pulsar-client-api, adding a dependency on pulsar-common to pulsar-client-api would create a circular compile dependency.

void-ptr974: Since this is only used by client internals, would an internal static/package-private helper in `pulsar-client` be a better fit than exposing it on `TopicMessageId`?

For example:

static boolean hasSameBasePartitionedTopic(TopicMessageId messageId, String topicName) {
    return TopicName.get(messageId.getOwnerTopic()).getPartitionedTopicName()
            .equals(TopicName.get(topicName).getPartitionedTopicName());
}

This also avoids adding a `pulsar-client-api -> pulsar-common` dependency.

This is mainly an API design discussion point from my side. It may be better to wait for lhotari/maintainer input before deciding the final shape.

lhotari (approved):
LGTM

PR #25922, opened 2026-06-03
[fix][broker] Delete orphan cursor after failed subscribe

BewareMyPower:
### Motivation

When a durable subscription fails due to an existing non-durable subscription, the durable cursor will still be created but not deleted, this could cause the durable cursor unexpectedly recovered after the topic is loaded again.

### Modifications

Delete the cursor in this case. Improve `testSameSubscriptionNameForDurableAndNonDurableSubscription` to avoid the regression.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
@@ -1193,23 +1193,28 @@ public class PersistentTopic extends AbstractTopic implements Topic, AddEntryCal
 
                 PersistentSubscription subscription = subscriptions.get(subscriptionName);
                 if (subscription == null) {
-                    PersistentSubscription newSubscription;
-                    try {
-                        newSubscription = createPersistentSubscription(subscriptionName, cursor, replicated,
-                                subscriptionProperties);
-                    } catch (Exception e) {
-                        deleteOpenedDurableCursor(cursor, subscriptionName, subscriptionFuture, e);
+                    subscription = subscriptions.computeIfAbsent(subscriptionName,
+                                  name -> createPersistentSubscription(subscriptionName, cursor,
+                                          replicated, subscriptionProperties));
+                } else {
+                    // if subscription exists, check if it's a non-durable subscription
+                    if (subscription.getCursor() != null && !subscription.getCursor().isDurable()) {
+                        ledger.asyncDeleteCursor(encodedSubscriptionName, new DeleteCursorCallback() {
+                            @Override
+                            public void deleteCursorComplete(Object ctx) {
+                            }
+
+                            @Override
+                            public void deleteCursorFailed(ManagedLedgerException exception, Object ctx) {
+                                log.warn().attr("subscription", subscriptionName)
+                                        .exceptionMessage(exception)
+                                        .log("Failed to delete cursor for conflicts");
+                            }
+                        }, null);
+                        subscriptionFuture.completeExceptionally(
+                                new NotAllowedException("NonDurable subscription with the same name already exists."));
                         return;
                     }
-                    PersistentSubscription existingSubscription =
-                            subscriptions.putIfAbsent(subscriptionName, newSubscription);
-                    subscription = existingSubscription != null ? existingSubscription : newSubscription;
-                }
-                // If subscription exists, check if it's a non-durable subscription.
-                if (subscription.getCursor() != null && !subscription.getCursor().isDurable()) {
-                    deleteOpenedDurableCursor(cursor, subscriptionName, subscriptionFuture,
-                            new NotAllowedException("NonDurable subscription with the same name already exists."));
-                    return;
                 }
                 if (replicated != null && replicated && !subscription.isReplicated()) {
                     // Flip the subscription state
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1198,6 +1199,18 @@ public void openCursorComplete(ManagedCursor cursor, Object ctx) {
                 } else {
                     // if subscription exists, check if it's a non-durable subscription
                     if (subscription.getCursor() != null && !subscription.getCursor().isDurable()) {
```
lhotari, line 1201: just wondering whether the check for `!subscription.getCursor().isDurable()` could be performed before calling `asyncOpenCursor`.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1198,6 +1199,18 @@ public void openCursorComplete(ManagedCursor cursor, Object ctx) {
...
+                            public void deleteCursorComplete(Object ctx) {
+                            }
+
+                            @Override
+                            public void deleteCursorFailed(ManagedLedgerException exception, Object ctx) {
+                                log.warn().attr("subscription", subscriptionName)
+                                        .exceptionMessage(exception)
+                                        .log("Failed to delete cursor for conflicts");
+                            }
+                        }, null);
```
lhotari, line 1213: Just wondering if could result in the non-durable cursor getting removed.
https://github.com/apache/pulsar/blob/dd9462646ace229237d40e1d6036b9aa3c917599/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L1086-L1096

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
```diff
@@ -1184,7 +1184,8 @@ private CompletableFuture<Subscription> getDurableSubscription(String subscripti
         }
 
         Map<String, Long> properties = PersistentSubscription.getBaseCursorProperties(replicated);
-        ledger.asyncOpenCursor(Codec.encode(subscriptionName), initialPosition, properties, subscriptionProperties,
+        String encodedSubscriptionName = Codec.encode(subscriptionName);
+        ledger.asyncOpenCursor(encodedSubscriptionName, initialPosition, properties, subscriptionProperties,
```
lhotari, line 1188: The default implementation seems to return the non-durable cursor and not create a new one:
https://github.com/apache/pulsar/blob/dd9462646ace229237d40e1d6036b9aa3c917599/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L1031-L1036

However, it's possible that there's a race where this would break.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
lhotari, line 1192: The cursor returned here could be non-durable, at least in the default ManagedLedgerImpl. Checking `cursor.isDurable()` could be useful too.

BewareMyPower:
@lhotari In my use case, checking if the created cursor is non-durable might not work.

In the downstream protocol handler implementation, an internal subscription could be created like

```java
        final var subscription = topic.getSubscriptions().computeIfAbsent(subName,
            __ -> new MockStatsPersistentSubscription(topic, subName));
```

```java
    public MockStatsPersistentSubscription(PersistentTopic topic, String subscriptionName) {
        super(topic, subscriptionName, new MockManagedCursor(), false);
```

where many methods are overwritten for special use. The `MockManagedCursor` is also durable to prevent a new subscription with the same name from being created.

In this case, the created cursor will still be durable but I also want to prevent the subscription creation.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
BewareMyPower: `asyncOpenCursor` is a synchronized method of `ManagedLedgerImpl`, which means `openCursorComplete` is also safely synchronized.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
BewareMyPower: Ah, that's the concern, let me rethink

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java
BewareMyPower: The race should not exist because this method is synchronized. The only concern is that the `getCursors()` method exposes `cursors` so that `cursors` could be modified without synchronizing the whole managed ledger.

BewareMyPower:
I just rethink for a while, it should not be an issue from Pulsar side because the `asyncOpenCursor` won't have a chance to create a new durable cursor.

My modification on the test is wrong because the internal stats should include non-durable cursors as well.

image

After the change, the non-durable cursor will be cleared, which is incorrect.

I will close this PR. The issue of my use case should be fixed from downstream side.

PR #25935, opened 2026-06-04
[cleanup][broker] Standardize java.time.Clock usage in broker service

### LINKED ISSUE #25660: Usage of java.time.Clock for timestamps isn't consistently performed in Pulsar code base
In some of the existing code locations, PulsarService's `java.time.Clock clock` field is used for timestamps, however it's not consistently applied in the code base. Would it be useful to use `java.time.Clock` for timestamps?

_Originally posted by @lhotari in https://github.com/apache/pulsar/pull/25658#discussion_r3178499222_

Praveenkumar76:
Fixes [apache#25660](https://github.com/apache/pulsar/issues/25660)

### Motivation

The usage of `java.time.Clock` is currently inconsistent across the broker module, with many classes directly relying on `System.currentTimeMillis()` for timestamp generation.

Standardizing timestamp handling through the centralized `PulsarService` clock improves consistency across the codebase and makes time-dependent behavior easier to control during testing. This reduces test flakiness by enabling deterministic time mocking in unit and integration tests.

Because there are hundreds of occurrences of `System.currentTimeMillis()` across the repository, this PR intentionally scopes the change to a tightly focused set of classes. It specifically targets classes where the `Clock` dependency can be cleanly injected via the constructor, ensuring a clear separation of concerns and keeping the review safe and manageable.

### Modifications

Replaced direct usages of System.currentTimeMillis() with this.clock.millis() by utilizing field injection for the Clock dependency in the following broker module classes:

- `MessageDeduplication.java`
- `SchemaRegistryStats.java`
- `PulsarStats.java`

Additional changes:
- Left existing `System.nanoTime()` usages unchanged, as they are used for monotonic latency measurements rather than wall-clock timestamps.

### Verifying this change

This change can be verified as follows:

- Confirmed successful compilation and that existing broker tests continue to pass.
- Verified timestamp-related logic in the modified classes now consistently uses the centralized broker clock via constructor injection.
- Confirmed no behavioral changes in latency-related code paths using `System.nanoTime()`.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PulsarStats.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/PulsarStats.java
@@ -61,6 +61,7 @@ public class PulsarStats implements Closeable {
     private final boolean exposePublisherStats;
 
     private final ReentrantReadWriteLock bufferLock = new ReentrantReadWriteLock();
+    private final PulsarService pulsar;
 
     @Getter
     private long updatedAt;
@@ -81,6 +82,7 @@ public class PulsarStats implements Closeable {
 
         this.exposePublisherStats = pulsar.getConfiguration().isExposePublisherStats();
         this.updatedAt = 0;
+        this.pulsar = pulsar;
 
     }
 
@@ -230,7 +232,7 @@ public class PulsarStats implements Closeable {
         } finally {
             bufferLock.writeLock().unlock();
         }
-        updatedAt = System.currentTimeMillis();
+        updatedAt = this.pulsar.getClock().millis();
     }
 
     public synchronized NamespaceBundleStats invalidBundleStats(String bundleName) {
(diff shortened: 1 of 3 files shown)


Praveenkumar76:
Hi @lhotari, I've completed the initial research! I dug through the codebase to map out exactly how time is being used, identified which specific areas are hurting testability (like cache TTLs and connection timeouts), and put together the architectural proposal we discussed.

You can read the full breakdown here: [check this out
](https://gist.github.com/Praveenkumar76/bafd70c30710cf26951e8022f12929b8) Let me know if you agree with the categorization between the 'Diagnostics' bucket and the 'Business Logic' bucket, and if the constructor injection approach sounds like the right path forward for future PRs!

PR #25938, opened 2026-06-04
[improve][broker] PIP-483: Scalable Topic Auto Split/Merge

merlimat:
This PIP proposes automatic segment split/merge for scalable topics, a sub-PIP of [PIP-468: Scalable Topic Controller](https://github.com/apache/pulsar/blob/master/pip/pip-468.md).

## Motivation

PIP-468 gives the controller the ability to split and merge segments, but only on explicit operator request. This PIP makes the controller scale a topic automatically, eliminating the partition-count-style operational toil scalable topics were meant to remove.

## Key design points

- **Splits are fast; merges are lazy.** Splits fire quickly with only a short `splitCooldown` (1m) to coalesce bursts; merges are gated by a longer `mergeCooldown` (5m) + cold `mergeWindow`.
- **The controller reacts, it does not poll.** New stream/checkpoint consumers register directly with the controller, so consumer-count scale-up is event-driven (within seconds). Traffic load is *pushed* into the metadata store by each segment's owning broker (only on a significant change, default ±25%) and read by the controller on a 60s tick — no per-tick RPC fan-out.
- **Pure decision function.** `AutoScalePolicyEvaluator.decide(layout, load, consumerCounts, policy, now)` → `Split | Merge | None`, unit-testable in isolation.
- **Bounded churn.** `maxSegments` (64) caps growth; `maxDagDepth` (10) caps merges-in-lineage to stop split↔merge flip-flopping (splits always remain available); threshold hysteresis between split and merge.
- **Default-on cluster-wide** with per-namespace (`Policies`) and per-topic (`ScalableTopicMetadata`) overrides, following the existing `autoTopicCreationOverride` pattern.

The PIP also specifies the new `SegmentLoadStats` metadata record, the `broker.conf` knobs, the admin client API, and the observability surface.

### Links

- Parent PIP: [PIP-468](https://github.com/apache/pulsar/blob/master/pip/pip-468.md)
- Grand-parent PIP: [PIP-460](https://github.com/apache/pulsar/blob/master/pip/pip-460.md)

### DIFF
--- a/pip/pip-483.md
+++ b/pip/pip-483.md
@@ -1,0 +1,285 @@
+# PIP-483: Scalable Topic Auto Split/Merge
+
+*Sub-PIP of [PIP-468: Scalable Topic Controller](pip-468.md)*
+
+## Motivation
+
+[PIP-468](pip-468.md) gives the Scalable Topic Controller the ability to **split** and **merge** segments, but only on explicit operator request via the admin API. An operator has to watch the topic, decide it is hot, and issue a split — and later notice it has gone cold and issue a merge. This is the same operational toil that partition-count management imposes on classic Pulsar topics, which scalable topics were meant to eliminate.
+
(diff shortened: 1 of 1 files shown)


lhotari (approved):
LGTM

pdolif (approved):

PR #25931, opened 2026-06-04
[improve][client] Handle synchronous failures in RetryUtil retry

Radiancebobo:
### Motivation

`RetryUtil.retryAsynchronously` accepts a `Supplier<CompletableFuture<T>>` and retries when the returned future completes exceptionally. However, if the supplier throws before returning a future, the exception escapes from the executor task before `whenComplete` is registered.

In that case, the retry loop does not schedule another attempt and the caller-provided callback is never completed. A transient synchronous failure while building the async operation can therefore become a permanently pending callback.

### Modifications

- Wrap the supplier invocation in `RetryUtil` so synchronous exceptions are converted into an exceptionally completed future.
- Treat a `null` future from the supplier as an explicit failure instead of allowing an uncontrolled `NullPointerException`.
- Add a regression test that verifies a synchronous failure on the first attempt is retried and the callback completes successfully on the second attempt.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

This change added tests and can be verified as follows:

- Added `RetryUtilTest.testSyncFailureIsRetried` to cover supplier exceptions thrown before a future is returned.
- Verified the new regression test:

```bash
./gradlew :pulsar-broker:test --tests org.apache.pulsar.client.impl.RetryUtilTest.testSyncFailureIsRetried -PtestGroups=utils -PtestRetryCount=0
```

- Verified the adjacent test class:

```bash
./gradlew :pulsar-broker:test --tests org.apache.pulsar.client.impl.RetryUtilTest -PtestGroups=utils -PtestRetryCount=0
```

### Does this pull request potentially affect one of the following parts:

*If the box was checked, please highlight the changes*

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The schema
- [ ] The default values of configurations
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics
- [ ] Anything that affects deployment

PR #25939, opened 2026-06-05
[fix][common] Handle synchronous failures from future suppliers

Radiancebobo:
### Motivation

Some asynchronous helper methods accept a `Supplier<CompletableFuture<T>>` and assume the supplier always returns a future. However, a supplier can also fail synchronously before creating the future.

Before this change, those synchronous failures could bypass the expected asynchronous error handling path:

- `FutureUtil.Sequencer.sequential` could throw directly instead of returning a failed future.
- `FutureUtil.composeAsync` could throw inside the executor task and leave the returned future incomplete.
- `Futures.executeWithRetry` in managed-ledger could throw before entering the retry state machine, so a retryable transient failure from the operation setup path would not be retried.

These helpers should preserve the `CompletableFuture` contract: callers should receive a completed or exceptionally completed future, rather than having synchronous exceptions escape or leave the returned future hanging.

### Modifications

This change updates the future supplier handling paths to convert synchronous supplier failures into failed futures:

- Added a private helper in `FutureUtil` to safely invoke `Supplier<CompletableFuture<T>>`.
- Updated `FutureUtil.Sequencer.sequential` to use the safe supplier invocation path.
- Updated `FutureUtil.composeAsync` to complete the returned future exceptionally when the supplier fails synchronously.
- Updated managed-ledger `Futures.executeWithRetry` to route synchronous supplier failures through the existing retry logic.
- Added null-future handling so suppliers that return `null` complete exceptionally with `NullPointerException` instead of causing an unchecked synchronous failure.

### Verifying this change

- Added `FutureUtilTest.testSequencerReturnsFailedFutureWhenTaskThrowsSynchronously`
- Added `FutureUtilTest.testComposeAsyncReturnsFailedFutureWhenSupplierThrowsSynchronously`
- Added `FuturesTest.testExecuteWithRetryHandlesSynchronousFailure`

Local verification:

- `./gradlew :pulsar-common:test --tests org.apache.pulsar.common.util.FutureUtilTest -PtestRetryCount=0 --rerun-tasks`
- `./gradlew :managed-ledger:test --tests org.apache.bookkeeper.mledger.util.FuturesTest -PtestRetryCount=0 --rerun-tasks`

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics


Radiancebobo:
There's a similar PR https://github.com/apache/pulsar/pull/25931

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -200,6 +200,20 @@ public static <T> CompletableFuture<T> failedFuture(Throwable t) {
         return future;
     }
 
+    private static <T> CompletableFuture<T> getFutureSafely(Supplier<CompletableFuture<T>> supplier,
+                                                            String nullFutureMessage) {
```
lhotari: This method could be made public after renaming and simplifying the signature. It's an overkill to add a separate message for the NullPointerException.

```suggestion
    public static <T> CompletableFuture<T> supplySafely(Supplier<CompletableFuture<T>> supplier) {
```

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -200,6 +200,20 @@ public static <T> CompletableFuture<T> failedFuture(Throwable t) {
...
+        CompletableFuture<T> future;
+        try {
+            future = supplier.get();
+        } catch (Throwable t) {
+            return failedFuture(t);
+        }
+        if (future == null) {
+            return failedFuture(new NullPointerException(nullFutureMessage));
```
lhotari: Instead of using an explicity NPE message, something like this would be cover what is needed.
```suggestion
            return failedFuture(new NullPointerException("The given supplier returned null, supplier=" + supplier));
```

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
lhotari: while making this `public`, add proper javadoc

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -238,11 +251,13 @@ public synchronized CompletableFuture<T> sequential(Supplier<CompletableFuture<T
                 if (sequencerFuture.isCompletedExceptionally() && allowExceptionBreakChain) {
                     return sequencerFuture;
                 }
-                return sequencerFuture = newTask.get();
+                return sequencerFuture = getFutureSafely(newTask, "Expected Supplier should not return null");
```
lhotari: fix the compilation issues after the renaming

lhotari: We'd need to update the AGENTS.md/CONTRIBUTING.md to make it more explicit that this step is necessary before more changes are pushed to a PR:
```
# Lint / verify (license headers, formatting, checkstyle) — run before pushing
./gradlew rat spotlessCheck checkstyleMain checkstyleTest
```
https://github.com/apache/pulsar/blob/master/CONTRIBUTING.md#building

Radiancebobo: > fix the compilation issues after the renaming

Sorry about that! I applied the changes directly on GitHub and missed the other parts that needed to be updated in sync. My local environment isn't updated yet, but I'll get on it and fix everything later today. I'll be more careful next time. Thanks for your patience!

Radiancebobo:
I have executed these and made sure there are no issues.

```bash
./gradlew :pulsar-common:test :managed-ledger:test --rerun-tasks -PtestRetryCount=0 \
  --tests org.apache.pulsar.common.util.FutureUtilTest \
  --tests org.apache.bookkeeper.mledger.util.FuturesTest

./gradlew :pulsar-common:spotlessCheck :pulsar-common:checkstyleMain :pulsar-common:checkstyleTest

```

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -200,6 +200,30 @@ public static <T> CompletableFuture<T> failedFuture(Throwable t) {
         return future;
     }
 
+    /**
+     * Invokes a supplier that is expected to return a {@link CompletableFuture}, converting synchronous failures into
+     * a failed future.
+     *
+     * If the supplier itself is {@code null}, the returned future is completed exceptionally with a
```
lhotari: is null or returns null

pulsar-common/src/main/java/org/apache/pulsar/common/util/FutureUtil.java
```diff
@@ -298,13 +322,16 @@ public static <T> CompletableFuture<T> addTimeoutHandling(CompletableFuture<T> f
...
-            executor.execute(() -> futureSupplier.get().whenComplete((result, error) -> {
-                if (error != null) {
-                    future.completeExceptionally(error);
-                    return;
-                }
-                future.complete(result);
-            }));
+            executor.execute(() -> {
+                CompletableFuture<T> supplierFuture = supplySafely(futureSupplier);
+                supplierFuture.whenComplete((result, error) -> {
```
lhotari: this reads better if it would be simply `supplySafely(futureSupplier).whenComplete`  without the intermediate variable

Radiancebobo:
@lhotari Could you please do another review for me?

pulsar-common/src/test/java/org/apache/pulsar/common/util/FutureUtilTest.java
```diff
@@ -252,4 +254,88 @@ public void testSequencer() {
             Assert.assertEquals(list3.get(i), (Integer) i);
         }
     }
-}
\ No newline at end of file
+
+    @Test
+    public void testSequencerReturnsFailedFutureWhenTaskThrowsSynchronously() throws Exception {
```
lhotari, line 260: **[QUALITY] the only test covering the two sequencer branches this PR rewrites has no @Test and never runs**

This new test is good, but it only reaches `sequential()`'s `isDone()` branch — a fresh sequencer, so `sequencerFuture` is already complete and the call goes to `FutureUtil.java:267`.

The two branches you also rewrote — `thenCompose(__ -> supplySafely(newTask))` at `FutureUtil.java:270-271` — are covered only by `testSequencer` at `FutureUtilTest.java:188`, and **that method never runs**: it has no `@Test`, and the class has no class-level annotation (every other method here is annotated individually — 48, 62, 79, 89, 106, 147, 258, 275, 284, 296, 312, 324). There is no `@Factory` in the file and no TestNG suite XML selecting it by name, and the Gradle setup uses `useTestNG` with no suite file, so only annotated methods execute.

It is dormant on master too (line 186 there), so this is not something you introduced. It matters here because it is also the only coverage of `allowExceptionBreakChain` anywhere, which means the green CI on this PR says nothing about either branch you changed.

Adding `@Test` is one line, in a file this PR already touches. Its suppliers never throw and never return `null`, so `supplySafely` is transparent to it and it should pass as-is — worth running once locally to confirm before pushing.

pulsar-common/src/test/java/org/apache/pulsar/common/util/FutureUtilTest.java
```diff
@@ -252,4 +254,88 @@ public void testSequencer() {
...
+        FutureUtil.Sequencer<String> sequencer = FutureUtil.Sequencer.create();
+
+        CompletableFuture<String> future = sequencer.sequential(() -> {
+            throw new IllegalStateException("sync fail");
+        });
+
+        try {
```
lhotari, line 267: **[QUALITY] optional: the new exception-path tests could use AssertJ per CODING.md**

Entirely optional, and take it or leave it. `CODING.md:103` says to prefer AssertJ assertions over TestNG ones, and the five new exception-path tests use the `try { … fail(…) } catch (ExecutionException e) { assertTrue(…) }` shape. The AssertJ equivalent is a line shorter and cannot be written in a way that silently passes:

```java
assertThatThrownBy(() -> future.get(2, TimeUnit.SECONDS))
        .hasCauseInstanceOf(IllegalStateException.class)
        .hasRootCauseMessage("sync fail");
```

Against that: this file is uniformly TestNG-style already, so matching its neighbours is a perfectly reasonable choice, and every one of your blocks does include the `fail(...)`, so none of them can pass by accident. Not a blocker either way.

lhotari (approved):
Sorry for the long wait on your re-review request — that was five weeks, and the delay was mine.

I checked all five of my earlier points against `99405831` and they are genuinely addressed: `supplySafely` is public with the exact signature I suggested and has javadoc; the null-return message is the one I proposed, verbatim; the javadoc says "is null, throws, or returns null"; and `composeAsync` now reads `supplySafely(futureSupplier).whenComplete(...)` with no intermediate variable. CI is green, and I confirmed your new tests actually execute (the `OTHER` unit group runs `:pulsar-common:test` and `:managed-ledger:test` without a group filter).

Worth stating plainly what this fixes, because it is more than tidying. In `composeAsync`, a supplier that threw — or returned `null` — did so *inside* the executor task, so the exception went to the executor's uncaught handler and the returned future was **never completed**. Its only two callers are `LeaderElectionImpl:308` and `LockManagerImpl:121`, both wrapped in `sequencer.sequential(...)`, so a stranded future there would also wedge that sequencer's chain — leader election and distributed lock handling. `Sequencer.sequential` could likewise throw synchronously out of a `CompletableFuture`-returning method, which the project rules forbid. Both are real, and the new tests pin them.

I also verified the `CODING.md` edit is a genuine correction rather than a loss: `checkArgumentAsync` exists **nowhere** in the repository — on master it appears only in that one doc line. Replacing it with `supplySafely` fixes a dangling reference.

One comment below is worth acting on and is a one-line change in a file you already touch. The other is an optional style nit.

Two things deliberately **not** raised as problems with this PR, recorded so they are not lost:

- `FutureUtil.unwrapCompletionException` returns `null` for a `CompletionException`/`ExecutionException` whose cause is `null`, and `Futures.executeWithRetry:83` would then NPE inside its callback and leave `resultFuture` pending forever. Real mechanism, but I could not find any in-tree way to produce such a wrapper — every construction site wraps a caught exception or an `ExecutionException.getCause()` from `CompletableFuture.get`, and the JDK never produces a null cause there. A one-line null-guard in `unwrapCompletionException` would close the corner globally, but that is hardening of an existing helper, not a defect of this change.
- The helper is a good one, and there are more call sites with exactly the shape it fixes — a bare `supplier.get()` invoked inside a callback or scheduled task, where a synchronous throw strands the outer future. Concretely: `NamespaceResources.runWithMarkDeleteAsync:395` (partitioned-topic delete — a throw there also leaves the topic marked-deleted in metadata), `ConsumerImpl:2502` (retry/DLQ producer creation), and `MetadataCacheImpl:388` (the BadVersion retry path). All pre-existing and out of scope here; a follow-up sweep would be worthwhile.

lhotari:
Sorry for the wait — done now, and the delay was on me rather than anything missing from your side. All five of my earlier points check out at `99405831`, and I have left one comment worth acting on (a one-line `@Test` on the neighbouring `testSequencer`, which turns out never to run) plus one optional style note.

pulsar-common/src/test/java/org/apache/pulsar/common/util/FutureUtilTest.java
Radiancebobo: Thank you for the thorough review and for catching this missing @Test annotation.

I added @Test to testSequencer, so the thenCompose(...supplySafely(...))
branch and allowExceptionBreakChain behavior are now actually exercised.

Verified locally with:

./gradlew :pulsar-common:test --tests org.apache.pulsar.common.util.FutureUtilTest

All 13 tests passed.

pulsar-common/src/test/java/org/apache/pulsar/common/util/FutureUtilTest.java
Radiancebobo: Thanks for the suggestion. I kept the TestNG assertions for consistency with the existing tests in FutureUtilTest. The fail(...) calls ensure that the tests cannot pass silently when no exception is thrown.

Radiancebobo:
> Sorry for the wait — done now, and the delay was on me rather than anything missing from your side. All five of my earlier points check out at `99405831`, and I have left one comment worth acting on (a one-line `@Test` on the neighbouring `testSequencer`, which turns out never to run) plus one optional style note.

Thank you for the thorough review and for catching the dormant test.

I’ve added @Test to testSequencer(), so it now executes the
thenCompose(...supplySafely(...)) branch and the allowExceptionBreakChain
behavior. I also verified locally that all 13 tests in FutureUtilTest pass.

Could you please take another look when you have a chance?

Thanks again!

lhotari (approved):
LGTM. The missing `@Test` annotation is now in place, and the focused `FutureUtilTest` and `FuturesTest` runs pass at the reviewed head.

PR #25945, opened 2026-06-05
[improve][broker] PIP-473 P5.4: v4/v5 transaction coordinator coexistence + enable v5 by default

merlimat:
## Motivation

The metadata-driven (v5, PIP-473) transaction coordinator and the legacy (v4) coordinator must serve their own clients on the **same cluster**, and the v5 coordinator should be enabled by default. Before this change, the broker chose the coordinator from a global config flag, so enabling the v5 TC routed **every** client — including v4 SDK clients on `persistent://` topics — to the v5 coordinator. That breaks v4 transactions: the v5 TC notifies participants via metadata-store events, which the legacy transaction buffer / pending-ack store don't consume.

This PR routes coordination by **client/SDK kind** instead of broker capability, then flips the defaults so v5 is on out of the box while v4 keeps working.

## Modifications

**Routing by SDK kind**
- New internal `ClientConfigurationData.scalableTransactions` flag — *not* on the public `ClientBuilder`. `PulsarClientBuilderV5` sets it `true`; the v4 builder never does.
- `TransactionCoordinatorClientImpl.selectDiscovery()` routes by that flag: v5 SDK → metadata-store coordinator (assignment watch), v4 SDK → legacy assign-topic coordinator. A v5 SDK on an `http://` service URL now fails clearly rather than silently downgrading.
- Wire: an optional `scalable` flag on `CommandTcClientConnectRequest`, `CommandNewTxn`, `CommandEndTxn`, `CommandAddPartitionToTxn`, `CommandAddSubscriptionToTxn` (additive, defaults false). `ServerCnx` routes each command to the v5 TC when `scalable=true` (rejecting it if the v5 TC isn't enabled), else to the legacy TC; `verifyTxnOwnership` is parameterized by the same flag.

No shared `TxnID` namespace is needed: the two coordinators keep separate state, and a client only ever talks to one kind, so a bare `TxnID` is never looked up across both.

**Default flip (P5.4)**
- `transactionCoordinatorScalableTopicsEnabled` → `true`
- `transactionBufferProviderClassName` → `DispatchingTransactionBufferProvider`
- `transactionPendingAckStoreProviderClassName` → `DispatchingTransactionPendingAckStoreProvider`

The dispatching providers route `segment://` topics to the metadata-driven implementations and `persistent://` / `topic://` to the legacy ones by `TopicName.isSegment()`, so v4 topics get the unchanged legacy buffer/ack-store.

## Verification

- Unit: `CommandsScalableTxnFlagTest` round-trips the `scalable` flag on all five commands (present + default-false).
- Integration (`TcMetadataDiscoveryTest`, verified locally against a freshly built docker image — **all pass**): on a v5-enabled cluster, a **v5 SDK** client drives the transaction lifecycle and survives a leader-broker kill via the metadata coordinator, while a **v4 SDK** client runs a transaction on a `persistent://` topic via the legacy coordinator — concurrently. Asserts the v4 client does **not** use metadata-store discovery.

## Notes for reviewers

- This adds client-broker wire-protocol surface (the additive `scalable` flag); flagging for protocol review.
- The enable-by-default flip means transaction-enabled broker tests now also start the v5 TC; CI exercises that breadth.

### DIFF
--- a/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
+++ b/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java
@@ -3798,13 +3798,14 @@ public class ServiceConfiguration implements PulsarConfiguration {
     @FieldContext(
             category = CATEGORY_TRANSACTION,
             doc = "Enable the metadata-driven transaction coordinator used by scalable topics."
-                    + " When true, wire commands (NEW_TXN / END_TXN / etc.) are served by the"
-                    + " metadata-store-backed coordinator instead of the legacy"
-                    + " TransactionMetadataStoreService. Requires transactionCoordinatorEnabled"
-                    + " = true, and must be enabled together with the scalable-topic transaction"
-                    + " buffer and pending-ack store providers."
+                    + " When true, transaction wire commands flagged as scalable (sent by v5 SDK"
+                    + " clients) are served by the metadata-store-backed coordinator, while legacy"
+                    + " (v4) clients continue to be served by TransactionMetadataStoreService — the"
+                    + " two coexist on the same cluster. Requires transactionCoordinatorEnabled"
+                    + " = true. Enabled by default together with the dispatching transaction buffer"
+                    + " and pending-ack store providers."
     )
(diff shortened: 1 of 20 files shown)


lhotari (approved):
LGTM

PR #25946, opened 2026-06-05
[fix][broker] Fix tableview divergence in ServiceUnitStateTableViewSyncer causing flaky tests

### LINKED ISSUE #24357: Flaky-test: ExtensibleLoadManagerImplTest.testLoadBalancerServiceUnitTableViewSyncer
### Search before reporting

### Example failure

https://github.com/apache/pulsar/actions/runs/15293178018/job/43016333195?pr=24354#step:9:2457

### Exception stacktrace

```text
Error:  Tests run: 99, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 205.124 s <<< FAILURE! - in org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImplTest
  Error:  org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImplTest.testLoadBalancerServiceUnitTableViewSyncer  Time elapsed: 8.084 s  <<< FAILURE!
  java.lang.AssertionError: did not expect [pulsar://localhost:33525] but found [pulsar://localhost:33525]
  	at org.testng.Assert.failEquals(Assert.java:1581)
  	at org.testng.Assert.assertNotEqualsImpl(Assert.java:156)
  	at org.testng.Assert.assertNotEquals(Assert.java:2197)
  	at org.testng.Assert.assertNotEquals(Assert.java:2219)
  	at org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImplTest.testLoadBalancerServiceUnitTableViewSyncer(ExtensibleLoadManagerImplTest.java:1426)
```

lhotari:
Fixes #24357

### Motivation

`ExtensibleLoadManagerImplTest.testLoadBalancerServiceUnitTableViewSyncer` hung for the full 300s TestNG suite-default timeout in both attempts of CI run [27011985664](https://github.com/apache/pulsar/actions/runs/27011985664/job/79723694452), then cascaded into a later test's `@BeforeMethod initializeState()` failing with HTTP 500 (`The producer ... can not send message to the topic persistent://pulsar/system/loadbalancer-service-unit-state within given timeout`), skipping `testHandleNoChannelOwner`.

Investigation (details in [this comment on #24357](https://github.com/apache/pulsar/issues/24357#issuecomment-4634015476)) found two production bugs in `ServiceUnitStateTableViewSyncer` that make `waitUntilSynced()` spin forever on entrySet-size divergence:

1. **Tombstone asymmetry** — a deletion is delivered to the syncer's put-based tail listeners as a `null` value. `ServiceUnitStateMetadataStoreTableViewImpl.put()` is `@NonNull` on the value, so the resulting NPE is silently swallowed by the table-view dispatchers and the deletion never propagates to the metadata-store view. (The reverse direction works only by accident: the system-topic view's `delete(key)` *is* `put(key, null)`.)

2. **Existing-vs-tail listener gap** — channel updates that land between the existing-items copy and the registration of the tail listeners in `syncTailItems()` are replayed to the freshly started views as *existing* items, which are wired to a no-op listener, so they never propagate. This is near-deterministic in the `MetadataStoreToSystemTopicSyncer` direction because closing the previous reader triggers a re-assignment of the channel-topic bundle itself into that gap (reproduced locally with the same sizes-off-by-one signature as CI, e.g. `MetadataStoreTableView.size: 44, SystemTopicTableView.size: 43`).

Additionally, the leader-election-churning tests (`testRoleChange`, `testRoleChangeIdempotency`, `testHandleNoChannelOwner`) can leave the `pulsar/system/0x00000000_0xffffffff` bundle unserved (`Namespace bundle ... not served by this instance`), so the next test's namespace unload cannot publish its `Releasing` event and fails with HTTP 500 or hangs server-side. Nothing heals that state until the background monitor task (120s interval) reconciles the brokers' roles with the channel ownership — passive waiting/retrying alone cannot recover (reproduced locally and in personal CI).

### Modifications

Production — `ServiceUnitStateTableViewSyncer` (a migration feature; no interface, call-site, or config changes):
- Route `null` tail items to `delete()` instead of `put()`, and log failed sync futures at WARN (previously silently discarded by the dispatchers).
- Make `waitUntilSynced()` reconcile periodically while diverged: copy source-only items to the destination (direction-aware via `dst.isMetadataStoreBased()`; raw metadata-store writes to avoid items being conflicted out, matching `syncExistingItemsToMetadataStore`) and remove destination-only items that predate the sync phase and are re-confirmed absent from the source (so a concurrent fresh write to the destination is never discarded). Batched at `MAX_CONCURRENT_SYNC_COUNT`; failures are logged and retried on the next pass; `InterruptedException` propagates.
- Make the sync-wait budget an overridable field with a `@VisibleForTesting` setter (default unchanged at 300s) and add throttled DEBUG progress logging.
- Code cleanup: shared `NOOP_CONSUMER`, `maybeWaitCompletion()` and `writeToMetadataStore()` helpers.

Tests:
- `testLoadBalancerServiceUnitTableViewSyncer`: cap at 240s (below the 300s suite default so a hang fails fast and is retried instead of consuming the whole slot), shrink the sync-wait budget to 30s before the first `start()`, drive `monitor()` inside the activation await, and tear the syncer down in a `finally` block so a failure cannot poison later tests.
- `awaitChannelOwnerStable()` helper: after the leader-election-churning tests, drive `monitor()` to reconcile roles with channel ownership (the eager equivalent of the 120s background monitor task) and probe that the channel topic is looked up and served before yielding to the next test. `testHandleNoChannelOwner` additionally forces a deterministic leader in its cleanup, since a body failure mid-churn can otherwise leave leadership flapping between the brokers. Method timeouts for the churning tests raised from 30s to 60s.
- `initializeState()`: drive `monitor()` and retry a bounded `unloadAsync` (15s per attempt) for up to 120s before failing loudly, healing a transiently unserved channel-topic bundle instead of failing the whole suite.

### Verifying this change

This change is already covered by existing tests, such as `ExtensibleLoadManagerImplTest` (both `@Factory` parametrizations exercise both syncer directions in `testLoadBalancerServiceUnitTableViewSyncer`).

Verification performed:
- Before the fix, the full class failed deterministically in the `MetadataStoreToSystemTopicSyncer` direction locally (twice, same signature as both CI attempts). After the fix, the syncer test completes in seconds and repeated full-class runs pass (102 tests, 0 failures), including runs where the residual pre-existing first-attempt flakes in `testRoleChange`/`testHandleNoChannelOwner` fired and were contained by the retry without cascading.
- Personal CI on the fork: `Pulsar CI Flaky` (BROKER_FLAKY group) passed twice consecutively (lhotari/pulsar#225); the only full-pipeline failure was an unrelated `GeoReplicationTest` integration flake.

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
@@ -23,6 +23,7 @@ import static org.apache.pulsar.broker.ServiceConfiguration.ServiceUnitTableView
 import static org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImpl.COMPACTION_THRESHOLD;
 import static org.apache.pulsar.broker.loadbalance.extensions.ExtensibleLoadManagerImpl.configureSystemTopics;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.Closeable;
 import java.io.IOException;
@@ -41,6 +42,7 @@ import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.common.util.ObjectMapperFactory;
+import org.jspecify.annotations.NonNull;
 
 /**
  * Defines ServiceUnitTableViewSyncer.
@@ -59,6 +61,7 @@ public class ServiceUnitStateTableViewSyncer implements Closeable {
     private volatile ServiceUnitStateTableView systemTopicTableView;
     private volatile ServiceUnitStateTableView metadataStoreTableView;
     private volatile boolean isActive = false;
+    private final ObjectWriter jsonWriter = ObjectMapperFactory.getMapper().writer();
 
 
     public void start(PulsarService pulsar)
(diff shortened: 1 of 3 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
```diff
@@ -187,62 +224,132 @@ private void syncTailItems() throws InterruptedException, IOException, TimeoutEx
...
+                    }
+                    maybeWaitCompletion(futures, false);
+                }
+            }
+            for (var e : staleDstItems.entrySet()) {
+                // Only remove items written before this sync phase began and re-confirmed absent
+                // from the source: a fresh destination write (e.g. from a broker already switched
+                // to the destination implementation) is propagated to the source by the tail
+                // listener instead of being deleted here.
+                if (e.getValue().timestamp() < started && src.get(e.getKey()) == null) {
```
void-ptr974, line 338: I think this path still needs a convergence action for destination-only fresh writes. Right now, a fresh `dst-only` item is preserved, but it is not copied back to `src`, so the two views can remain with different key sets and `waitUntilSynced()` can still time out.

The shape I would expect here is:

1. Capture a local pre-clean destination snapshot, preferably lightweight, such as `key -> version/fingerprint`.
2. During reconciliation:
   - `src-only`: copy `src -> dst`.
   - `dst-only`: if the key existed in the pre-clean destination snapshot and the current destination value still matches that snapshot, treat it as stale pre-existing destination data and delete it from `dst`; otherwise treat it as a real migration/startup-time destination write and copy it back `dst -> src`.

This gives every diff a convergence action. Fresh destination-side writes are preserved and propagated back to the source, while only confirmed unchanged pre-clean leftovers are removed.

I would also avoid using `ServiceUnitStateData.timestamp() < started` for this distinction. The timestamp is produced by the writer broker, while `started` is this syncer broker's local wall clock, so clock skew can make a fresh destination write look stale or a stale pre-existing item look fresh.

lhotari: quick comment: the migration is one way, src -> dst. there isn't an intention to copy items back. I agree that timestamps are brittle here.

It seems that making this migration tool perfect would require some redesign. It looks like the initial design is best effort type.
the use case of this migration solution is rare. it's meant for online migration from system topic based ELB to metadata store based ELB.

heesung-sohn (approved):
Lgtm .

pulsar-broker/src/main/java/org/apache/pulsar/broker/loadbalance/extensions/channel/ServiceUnitStateTableViewSyncer.java
heesung-sohn: Yes, this tool is best-effort basis, in case someone needs a quick live migration with many happy-case only assumptions.

void-ptr974: Got it, thanks for clarifying. If this is meant to be a best-effort one-way migration helper, then my dst-only concern shouldn’t block this PR.

void-ptr974 (approved):

void-ptr974:
I saw CI is still failing, but it looks unrelated to this PR. I created #25951 and #25950 to track/fix it. LGTM.

PR #25947, opened 2026-06-06
[fix][client] Avoid closing reused ServiceUrlProvider in another PulsarClientImpl instance

oneby-wang:
### Motivation

When the same `ServiceUrlProvider` instance is reused to build a second `PulsarClientImpl`, the provider rejects the second initialization because each provider instance is tied to one client lifecycle.

The second client constructor currently enters the failure cleanup path and calls `shutdown()`, which closes `conf.getServiceUrlProvider()` unconditionally. This can close a provider that is still owned by the first live client.

### Modifications

- Track whether the current `PulsarClientImpl` successfully initialized its `ServiceUrlProvider`.
- Close the provider during `shutdown()` only when this client successfully initialized it.
- Add a regression test that verifies a failed second client build does not close the provider used by the first client.

### Verifying this change

- [ ] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
+++ b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java
@@ -121,6 +121,7 @@ public class PulsarClientImpl implements PulsarClient {
     @Getter
     private final Timer timer;
     private boolean needStopTimer;
+    private boolean serviceUrlProviderInitialized;
     private final ExecutorProvider externalExecutorProvider;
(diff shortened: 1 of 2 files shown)


void-ptr974 (approved):
LGTM. The fix correctly ties `ServiceUrlProvider.close()` to successful initialization by the current client, and the added regression test covers the reused-provider failure path.

merlimat (approved):

PR #25955, opened 2026-06-06
[fix][offload] Fix final ledger not being offloaded for terminated topics

### LINKED ISSUE #25956: [Bug] Final managed-ledger ledger cannot be offloaded after topic termination
### Search before reporting

- [x] I searched in the issues and found nothing similar.

### Read release policy

### User environment

Reproduced on current master.

### Issue Description

After a persistent topic is terminated, the final managed-ledger ledger cannot be offloaded.

Earlier closed ledgers can be offloaded normally, but the last ledger remains in BookKeeper. Since topic termination closes the topic for further writes, the final ledger is no longer active and should be eligible for offload when tiered storage is enabled.

This causes terminated topic data to be only partially offloaded, with the final ledger continuing to consume BookKeeper storage.

Expected behavior:

The final ledger of a terminated topic should be offloadable when it matches the configured offload policy.

Actual behavior:

The final ledger remains not offloaded after topic termination.

### Error messages

No error message is required to reproduce this issue.

### Reproducing the issue

1. Configure tiered-storage offload for a namespace or topic.
2. Produce messages to a persistent topic.
3. Terminate the topic.
4. Trigger offload, or configure automatic offload thresholds so the topic data should be eligible for offload.
5. Check the managed-ledger offload metadata or BookKeeper storage.

The final managed-ledger ledger of the terminated topic remains not offloaded.

### Additional information

Related PR: #25955

void-ptr974:
### Motivation

Terminated topics can leave their final ledger in BookKeeper even when tiered-storage offload is enabled and the ledger is eligible for offload.

Termination closes the current ledger outside the normal ledger rollover path. That means the final ledger is complete, but it is not handled like other closed ledgers for offload purposes. Its closed-ledger metadata is not refreshed during termination, automatic offload is not triggered from that path, and manual `offloadPrefix()` still excludes the last ledger as if it were active.

As a result, terminated topic data can be only partially offloaded, with the final ledger continuing to consume BookKeeper storage.

Fixes #25956

### Modifications

- Record closed-ledger metadata for the final ledger when a managed ledger is terminated.
- Trigger automatic offload after the terminated state is persisted.
- Allow manual offload to include the last ledger when the managed ledger is terminated and the requested position covers the terminated position.
- Add managed-ledger tests for manual, automatic, bounded, and idempotent offload of the final ledger after termination.

### Verifying this change

Ran:

```bash
./gradlew :managed-ledger:test   --tests org.apache.bookkeeper.mledger.impl.OffloadPrefixTest   --tests org.apache.bookkeeper.mledger.impl.ManagedLedgerTerminationTest   --no-daemon
```

Result: `BUILD SUCCESSFUL`.


managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
```diff
@@ -3682,23 +3692,27 @@ public void asyncOffloadPrefix(Position pos, OffloadCallback callback, Object ct
             }
 
             long current = ledgers.lastKey();
+            boolean includeLastLedger = STATE_UPDATER.get(this) == State.Terminated
```
Denovo1998, line 3751: I think this needs one more guard for the in-progress termination window. asyncTerminate() sets state = State.Terminated before the BookKeeper close callback refreshes the final LedgerInfo. If asyncOffloadPrefix() runs in that window, includeLastLedger becomes true, but the current ledger still has the active-ledger placeholder metadata (entries/size are still 0 and timestamp is not refreshed). The loop can then find no ledgers to offload and return lastConfirmedEntry.getNext(), which reports the prefix as fully offloaded even though the final ledger was never offloaded.

Could we gate includeLastLedger on the final LedgerInfo actually being closed/refreshed, or introduce a separate terminating state so manual offload does not treat an in-flight termination as completed? A regression test can block the BK close with PulsarMockBookKeeper.promiseAfter(0), call asyncTerminate(), call offloadPrefix() before releasing the close, and assert it does not report lastConfirmedEntry.getNext() without offloading the final ledger.

Denovo1998 (changes requested):

managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java
dao-jun: Recommended fix: Gate includeLastLedger on the final ledger actually being closed (metadata refreshed). The simplest reliable signal is whether the last ledger's LedgerInfo has been populated:

  LedgerInfo lastLedgerInfo = ledgers.get(current);
  boolean lastLedgerClosed = lastLedgerInfo != null && lastLedgerInfo.getEntries() > 0;
  boolean includeLastLedger = STATE_UPDATER.get(this) == State.Terminated
          && requestOffloadTo.compareTo(lastConfirmedEntry) >= 0
          && lastLedgerClosed;

lhotari:
Please merge origin/master and resolve conflicts.

void-ptr974:
Hi all, thanks for the feedback.

I've merged the latest `master` and resolved the conflict.

The final ledger takes a different path when a topic is terminated: it is closed outside the normal rollover flow, so it could miss the usual offload handling and remain in BookKeeper.

This update refreshes its metadata during termination, persists the terminated state, and then lets automatic or manual offload pick it up.

I also tightened two edge cases found while updating it: termination metadata updates now wait their turn with other metadata writes, and manual offload waits until the final ledger has actually finished closing before including it.

PR #25967, opened 2026-06-08
[improve][pip] PIP-484: Expose Window interface for WindowFunction processing

Dream95:
### Motivation

Pulsar Window Functions currently invoke `WindowFunction.process(Collection<Record<X>>, ...)` with **all messages in the window** on every trigger. Internally, `WindowManager` already classifies events into full, newly added, and expired lists on each activation, but `WindowFunctionExecutor` drops `getNew()` and `getExpired()` before calling the user function.
This makes incremental computation inefficient for sliding windows  and forces users to manually diff full collections.

### Modifications

This PR adds **PIP-484**, which proposes:
- Promote the existing internal `Window<T>` interface to the public API (`get()`, `getNew()`, `getExpired()`, timestamps).
- Remove the internal `org.apache.pulsar.functions.windowing.Window` type; instance code imports `org.apache.pulsar.functions.api.Window` directly.
- Add a new **default** `process(Window<Record<X>>, WindowContext)` overload on the existing `WindowFunction<X, T>` interface. The default delegates to `process(window.get(), context)`, so lambdas and existing class implementations remain unchanged.
- Update `WindowFunctionExecutor` to pass the full `Window` object to `WindowFunction.process(window, context)` instead of only `inputWindow.get()`.

- [x] The public API
- [ ] The schema
- [ ] The threading model
- [ ] The binary protocol
- [ ] The REST endpoints
- [ ] The admin CLI options
- [ ] The metrics

### DIFF
--- a/pip/pip-484.md
+++ b/pip/pip-484.md
@@ -545,6 +545,6 @@ There is no wire-protocol change between Functions Workers. No special geo-repli
 # Links
 
 * Mailing List discussion thread: https://lists.apache.org/thread/gwfpt27j57vkfgcscrj7yyfn0jdg1oq4
-* Mailing List voting thread: TBD
+* Mailing List voting thread: https://lists.apache.org/thread/mcnl2tf5lppz52vkk8qrby2xzn55gfz1
 * Related: [PIP-15: Pulsar Functions](pip-15.md)
 * Related: [PIP-396: Align WindowFunction's WindowContext with BaseContext](pip-396.md)
--- a/pip/pip-484.md
+++ b/pip/pip-484.md
@@ -125,11 +125,14 @@ default T process(Window<Record<X>> window, WindowContext context) throws Except
 }
 ```
 
-This keeps `WindowFunction` a valid `@FunctionalInterface`. Existing lambdas continue to implement the collection-based abstract method and cannot override the default overload, so lambda behavior remains unchanged. Users that need `getNew()` / `getExpired()` implement `WindowFunction` as a class and override the new default overload.
+This keeps `WindowFunction` a valid `@FunctionalInterface`. Existing lambdas bind to the abstract collection-based method and keep today's behavior.
 
-`WindowFunctionExecutor` passes the full `Window` object to `WindowFunction.process(window, context)`. For existing implementations, the default method delegates to `process(window.get(), context)`.
+`WindowFunctionExecutor` always calls `process(Window<Record<X>>, WindowContext)`. The default overload delegates to `process(window.get(), context)` unless a class overrides the window overload.
 
-If a class-based implementation overrides both methods, the `Window<Record<X>>` overload is the runtime entry point because the executor calls `process(window, context)`. The collection-based method remains available for shared logic or fallback behavior; for example, the window overload may explicitly call `process(window.get(), context)` after using `getNew()` / `getExpired()`.
+### Design tradeoff
+
+- **Lambdas cannot access incremental views** — they cannot override the default `process(Window, …)` method, so `getNew()` / `getExpired()` require a class-based implementation.
+- **Class override is the runtime entry point** — when a class overrides `process(Window, …)`, the runtime does not invoke any `process(Collection, …)` override unless the window overload delegates to it explicitly.
 
 Data flow (after the change):
 
@@ -204,10 +207,13 @@ public interface Window<T> {
     /**
      * The window end timestamp for this activation.
      * <p>
-     * For event-time windows, this is based on the watermark. For processing-time
-     * windows, this is based on processing time. Built-in trigger policies provide
-     * this reference timestamp for activated windows. If no reference timestamp is
-     * available, this method returns {@code null}.
+     * Populated by {@code WindowFunctionExecutor} when it constructs the
+     * {@code WindowImpl} for each trigger — {@code WindowImpl} stores the value
+     * as supplied; it does not derive timestamps itself. For event-time windows,
+     * this is the reference time from the trigger policy (watermark-based). For
+     * processing-time windows, this is the processing-time reference from the
+     * trigger policy. Returns {@code null} if the trigger provides no reference
+     * timestamp.
      * </p>
      */
     Long getEndTimestamp();
@@ -215,9 +221,17 @@ public interface Window<T> {
     /**
      * The window start timestamp for this activation.
      * <p>
-     * This is computed from {@link #getEndTimestamp()} and the configured
-     * time-based window length. Returns {@code null} if the window length is not
-     * time-based or if no end timestamp is available.
+     * Populated by {@code WindowFunctionExecutor} when it constructs the
+     * {@code WindowImpl} for each trigger. When a time-based window length
+     * ({@code windowLengthDurationMs}) is configured, the executor sets this to
+     * {@code getEndTimestamp() - windowLengthDurationMs}. Returns {@code null}
+     * for count-based windows (no time-based window length) or when
+     * {@link #getEndTimestamp()} is {@code null}.
+     * </p>
+     * <p>
+     * For sliding windows, start/end bound the <em>current activation's</em>
+     * window interval (window length), not the sliding step size
+     * ({@code slidingInterval*}).
      * </p>
      */
     Long getStartTimestamp();
@@ -306,37 +320,32 @@ Because the new method is `default`, adding it is source-compatible and binary-c
 
 #### Example: sliding-window sum
 
+Both overloads are shown below. The runtime invokes only `process(Window, …)`; the `process(Collection, …)` override is **not called** unless the window overload delegates to it explicitly (for example `return process(window.get(), context)`). Running sum is held in an instance field (not checkpointed across restarts).
+
 ```java
 /**
  * Maintains the sum of integer values in the current sliding window incrementally.
  */
 public class SlidingWindowSumFunction implements WindowFunction<Integer, Integer> {
 
-    private static final String RUNNING_SUM_KEY = "running-sum";
+    private long runningSum;
 
     @Override
     public Integer process(Collection<Record<Integer>> input, WindowContext context) {
-        // Note: When both overloads are overridden, the runtime invokes
-        // process(Window, WindowContext). This method remains supported for
-        // shared logic or explicit delegation (e.g. process(window.get(), context)).
+        // Not invoked at runtime when process(Window, …) is overridden.
+        // Kept for shared logic or explicit delegation from the window overload.
         return input.stream().mapToInt(Record::getValue).sum();
     }
 
     @Override
     public Integer process(Window<Record<Integer>> window, WindowContext context) throws Exception {
-        long newEventsSum = 0;
         for (Record<Integer> record : window.getNew()) {
-            newEventsSum += record.getValue();
+            runningSum += record.getValue();
         }
-        long expiredSum = 0;
         for (Record<Integer> record : window.getExpired()) {
-            expiredSum += record.getValue();
+            runningSum -= record.getValue();
         }
-        long netDelta = newEventsSum - expiredSum;
-        if (netDelta != 0) {
-            context.incrCounter(RUNNING_SUM_KEY, netDelta);
-        }
-        return (int) context.getCounter(RUNNING_SUM_KEY);
+        return (int) runningSum;
     }
 }
 ```
@@ -347,7 +356,7 @@ public class SlidingWindowSumFunction implements WindowFunction<Integer, Integer
 
 No new field or user-function detection path is needed. `initializeUserFunction()` continues to accept `WindowFunction` exactly as it does today.
 
-Update `process(Window<Record<X>>, WindowContext)` to pass the full window view to `WindowFunction`. Existing class implementations and lambdas use the default overload, which delegates to `process(inputWindow.get(), context)`. Class implementations that override the new overload receive access to `getNew()` / `getExpired()`. If both overloads are implemented, the new `Window<Record<X>>` overload is called by the runtime.
+Update `process(Window<Record<X>>, WindowContext)` to pass the full window view to `WindowFunction` (see **Design tradeoff** above).
 
 ```java
 public T process(Window<Record<X>> inputWindow, WindowContext context) throws Exception {
@@ -376,12 +385,12 @@ Promoted from the internal package. The internal `org.apache.pulsar.functions.wi
 | `List<T> get()` | All events in the current window. Returns a mutable list that is a **read-only snapshot** for this activation; callers must not modify it or retain it past `process()`. |
 | `List<T> getNew()` | Events added since the last trigger. Same lifetime and mutability contract as `get()`. |
 | `List<T> getExpired()` | Events removed since the last trigger. Same lifetime and mutability contract as `get()`. |
-| `Long getStartTimestamp()` | Window start time, computed from `getEndTimestamp()` and the configured time-based window length. Returns `null` if the window length is not time-based or if no end timestamp is available. |
-| `Long getEndTimestamp()` | Window end time for this activation: watermark-based for event-time windows, processing-time based for processing-time windows. Built-in trigger policies provide this timestamp; returns `null` only if no reference timestamp is available. |
+| `Long getStartTimestamp()` | Window start time for this activation. Set by `WindowFunctionExecutor` to `getEndTimestamp() - windowLengthDurationMs` when a time-based window length is configured; `null` for count-based windows or when `getEndTimestamp()` is `null`. For sliding windows, this bounds the current activation's window interval (window length), not the sliding step. |
+| `Long getEndTimestamp()` | Window end time for this activation. Set by `WindowFunctionExecutor` from the trigger policy's reference time (watermark-based for event-time, processing-time for processing-time). `WindowImpl` stores the value as supplied; returns `null` if the trigger provides no reference timestamp. |
 
 #### API contracts (`Window` lifetime and list mutability)
 
-On each window trigger, `WindowFunctionExecutor` builds a new `WindowImpl` with freshly collected lists and passes it to `WindowFunction.process(Window<Record<X>>, WindowContext)`. The runtime does not retain that object after `process()` returns.
+On each window trigger, `WindowFunctionExecutor` builds a new `WindowImpl` with freshly collected lists, injects `getEndTimestamp()` from the trigger's reference time and `getStartTimestamp()` from `referenceTime - windowLengthDurationMs` (when applicable), and passes the view to `WindowFunction.process(Window<Record<X>>, WindowContext)`. The runtime does not retain that object after `process()` returns.
 
 | Contract | Rule |
 |----------|------|
@@ -397,7 +406,7 @@ The existing public interface gains a new default overload.
 | Method | Description |
 |--------|-------------|
 | `T process(Collection<Record<X>> input, WindowContext context)` | Existing abstract method. Existing implementations and lambdas continue to target this method. |
-| `default T process(Window<Record<X>> window, WindowContext context)` | New default method. Delegates to `process(window.get(), context)` unless a class-based implementation overrides it to access incremental and expired events. |
+| `default T process(Window<Record<X>> window, WindowContext context)` | New default method. Delegates to `process(window.get(), context)` unless a class override provides incremental access (see **Design tradeoff**). |
 
 ### Configuration
 
--- a/pip/pip-484.md
+++ b/pip/pip-484.md
@@ -544,7 +544,7 @@ There is no wire-protocol change between Functions Workers. No special geo-repli
 
 # Links
 
-* Mailing List discussion thread: TBD
+* Mailing List discussion thread: https://lists.apache.org/thread/gwfpt27j57vkfgcscrj7yyfn0jdg1oq4
 * Mailing List voting thread: TBD
 * Related: [PIP-15: Pulsar Functions](pip-15.md)
 * Related: [PIP-396: Align WindowFunction's WindowContext with BaseContext](pip-396.md)
--- a/pip/pip-484.md
+++ b/pip/pip-484.md
@@ -1,0 +1,336 @@
+# PIP-484: Expose Incremental Window Events via IncrementalWindowFunction
+
+# Background knowledge
+
+## Pulsar Window Functions
+
+Pulsar Window Functions are a specialized form of Pulsar Function that group incoming messages into windows based on time or message count, and invoke the user function with a batch of messages each time a window fires.
+
+Window types:
+
+- **Tumbling window**: adjacent windows do not overlap; each message belongs to exactly one window.
+- **Sliding window**: adjacent windows may overlap; a message can belong to multiple windows.
+
+Time semantics:
+
+- **Processing time**: windows are driven by the clock at which messages enter the system.
+- **Event time**: windows are driven by timestamps embedded in messages, with watermarks used to track event-time progress.
+
+## Existing public API
+
+```
+pulsar-functions/api-java
+└── org.apache.pulsar.functions.api
+    ├── WindowFunction<X, T>        // user-implemented window function interface
+    └── WindowContext               // context interface for window functions
+```
+
+`WindowFunction` signature:
+
+```java
+@FunctionalInterface
+public interface WindowFunction<X, T> {
+    T process(Collection<Record<X>> input, WindowContext context) throws Exception;
+}
+```
+
+On each trigger, the user function receives a `Collection<Record<X>>` containing **all messages in the current window**.
+
+## Internal runtime pipeline
+
+```
+pulsar-functions/instance
+└── org.apache.pulsar.functions.windowing
+    ├── Window<T>                   // window view interface (internal package today)
+    ├── WindowImpl<T>               // Window implementation holding three event lists
+    ├── WindowManager<T>            // window manager; classifies events
+    ├── WindowFunctionExecutor<T,X> // executor bridging runtime and user function
+    ├── WindowLifecycleListener<T>  // window lifecycle callbacks
+    ├── EvictionPolicy<T>           // eviction policy (decides when events expire)
+    └── TriggerPolicy<T>            // trigger policy (decides when to fire a window)
+```
+
+The internal `Window<T>` interface already exposes incremental views:
+
+```java
+public interface Window<T> {
+    List<T> get();             // all events in the current window
+    List<T> getNew();          // events added since the last trigger
+    List<T> getExpired();      // events removed since the last trigger
+    Long getStartTimestamp();  // window start timestamp
+    Long getEndTimestamp();    // window end timestamp (reference time)
+}
+```
+
+# Motivation
+
+## Problem: `getNew()` / `getExpired()` data is discarded at the public API layer
+
+`WindowManager.onTrigger()` already classifies events into three categories on every window activation:
+
+| Category | Meaning |
+|----------|---------|
+| `tuples` | all events currently in the window |
+| `newTuples` | events newly added since the last trigger |
+| `expiredTuples` | events removed since the last trigger |
+
+These three lists are passed into `WindowImpl` and delivered to the executor via `WindowLifecycleListener.onActivation()`. However, `WindowFunctionExecutor.process(Window, WindowContext)` only passes `inputWindow.get()` to the user function; newly added and expired events are discarded:
+
+```java
+// WindowFunctionExecutor.java (current implementation)
+public X process(Window<Record<T>> inputWindow, WindowContext context) throws Exception {
+    // ...
+    return this.windowFunction.process(inputWindow.get(), context); // full window only; getNew()/getExpired() dropped
+}
+```
+
+## Impact
+
+Users cannot perform efficient incremental computation. Typical affected scenarios:
+
+1. **Incremental aggregation** (sliding-window statistics): on each trigger most messages in the window are unchanged; re-scanning the full collection is wasteful.
+2. **State maintenance**: when external state must track which messages entered or left the window, users must diff full collections manually — inefficient and error-prone.
+3. **Expired-event handling**: side effects such as resource release or counter decrements when messages leave the window.
+
+# Goals
+
+## In Scope
+
+- Expose the `Window<T>` interface in the public API (including `getNew()`, `getExpired()`, and timestamp methods).
+- Add a new public `IncrementalWindowFunction<X, T>` interface so users can receive the full `Window<Record<X>>` view.
+- Have `WindowFunctionExecutor` transparently support the new interface without requiring configuration or deployment changes.
+- Update Functions deployment validation (`FunctionConfigUtils.doJavaChecks`) so `IncrementalWindowFunction` implementations pass the same Java class checks as `WindowFunction`.
+- Preserve all existing behavior for current `WindowFunction` users.
+
+## Out of Scope
+
+- Incremental support for `java.util.function.Function` (bare window functions).
+- Equivalent capability for Python / Go Functions.
+- Changes to window state snapshot / checkpoint mechanisms.
+
+# High Level Design
+
+Introduce a new public interface `IncrementalWindowFunction<X, T>` whose `process` method accepts `Window<Record<X>>` instead of `Collection<Record<X>>`, giving users access to:
+
+- `window.get()` — all messages in the current window
+- `window.getNew()` — messages added since the last trigger
+- `window.getExpired()` — messages removed since the last trigger
+- `window.getStartTimestamp()` / `window.getEndTimestamp()` — window time boundaries
+
+`WindowFunctionExecutor` detects at initialization whether the user class implements `IncrementalWindowFunction`. If so, it passes the `Window` object directly; otherwise it follows the existing code path.
+
+Data flow (after the change):
+
+```mermaid
+flowchart TD
+    WM["WindowManager.onTrigger()"]
+    WL["onActivation(tuples, newTuples, expiredTuples)"]
+    WE["WindowFunctionExecutor.processWindow()"]
+    WI["WindowImpl(tuples, newTuples, expiredTuples)"]
+
+    subgraph executor ["process(Window, WindowContext) dispatch"]
+        P1["IncrementalWindowFunction\n→ process(window, context)"]
+        P2["WindowFunction\n→ process(window.get(), context)"]
+        P3["bareWindowFunction\n→ apply(values)"]
+    end
+
+    WM --> WL --> WE --> WI --> executor
+```
+
+# Detailed Design
+
+## Design & Implementation Details
+
+### Change 1: Move `Window<T>` to `api-java`
+
+**Current path**: `pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/Window.java`
+
+**New path**: `pulsar-functions/api-java/src/main/java/org/apache/pulsar/functions/api/Window.java`
+
+The interface methods remain unchanged; only the package declaration and license header are updated:
+
+```java
+// pulsar-functions/api-java/.../api/Window.java
+public interface Window<T> {
+    List<T> get();
+    List<T> getNew();
+    List<T> getExpired();
+    Long getEndTimestamp();
+    Long getStartTimestamp();
+}
+```
+
+The existing internal `Window.java` is replaced by a reference to the `api-java` interface (or removed entirely, with `WindowImpl` implementing the new public interface directly).
+
+### Change 2: Add `IncrementalWindowFunction<X, T>` interface
+
+**Path**: `pulsar-functions/api-java/src/main/java/org/apache/pulsar/functions/api/IncrementalWindowFunction.java`
+
+```java
+@FunctionalInterface
+public interface IncrementalWindowFunction<X, T> {
+    /**
+     * Process the triggered window.
+     *
+     * @param inputWindow the window view for this activation, providing access to
+     *                    all current events ({@link Window#get()}),
+     *                    newly added events ({@link Window#getNew()}), and
+     *                    expired events ({@link Window#getExpired()}).
+     * @param context     the window function context
+     * @return the output, or {@code null} to suppress output
+     */
+    T process(Window<Record<X>> inputWindow, WindowContext context) throws Exception;
+}
+```
+
+#### Example: sliding-window sum
+
+```java
+/**
+ * Maintains the sum of integer values in the current sliding window incrementally.
+ */
+public class SlidingWindowSumFunction implements IncrementalWindowFunction<Integer, Integer> {
+
+    private static final String RUNNING_SUM_KEY = "running-sum";
+
+    @Override
+    public Integer process(Window<Record<Integer>> window, WindowContext context) throws Exception {
+        long newEventsSum = 0;
+        for (Record<Integer> record : window.getNew()) {
+            newEventsSum += record.getValue();
+        }
+        long expiredSum = 0;
+        for (Record<Integer> record : window.getExpired()) {
+            expiredSum += record.getValue();
+        }
+        long netDelta = newEventsSum - expiredSum;
+        if (netDelta != 0) {
+            context.incrCounter(RUNNING_SUM_KEY, netDelta);
+        }
+        return (int) context.getCounter(RUNNING_SUM_KEY);
+    }
+}
+```
+
+### Change 3: Update `WindowFunctionExecutor`
+
+**Path**: `pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java`
+
+#### 3a. Add field
+
+```java
+protected IncrementalWindowFunction<T, X> incrementalWindowFunction;
+```
+
+#### 3b. Extend `initializeUserFunction()`
+
+Detect `IncrementalWindowFunction` via `instanceof` in `initializeUserFunction()`, following the same pattern used for `WindowFunction` today:
+
+```java
+@SuppressWarnings("unchecked")
+private void initializeUserFunction(WindowConfig windowConfig) {
+    // ...
+    if (userClassObject instanceof java.util.function.Function) {
+        // existing logic, unchanged
+        bareWindowFunction = ...;
+    } else if (userClassObject instanceof IncrementalWindowFunction) {
+        incrementalWindowFunction = (IncrementalWindowFunction<T, X>) userClassObject;
+    } else if (userClassObject instanceof WindowFunction) {
+        // existing logic, unchanged
+        windowFunction = (WindowFunction<T, X>) userClassObject;
+    } else {
+        throw new IllegalArgumentException("Window function does not implement the correct interface");
+    }
+}
+```
+
+#### 3c. Update `process(Window<Record<T>>, WindowContext)`
+
+```java
+public X process(Window<Record<T>> inputWindow, WindowContext context) throws Exception {
+    if (this.bareWindowFunction != null) {
+        Collection<T> values = inputWindow.get().stream()
+                .map(Record::getValue).collect(Collectors.toList());
+        return this.bareWindowFunction.apply(values);
+    } else if (this.incrementalWindowFunction != null) {
+        // pass the full Window view; user can access getNew() / getExpired()
+        return this.incrementalWindowFunction.process(inputWindow, context);
+    } else {
+        // existing behavior: pass full message collection only
+        return this.windowFunction.process(inputWindow.get(), context);
+    }
+}
+```
+
+### Change 4: Update deployment validation (`functions/utils`)
+
+Submit-time validation must accept `IncrementalWindowFunction` the same way it already accepts `WindowFunction`.
+
+| File | Change |
+|------|--------|
+| `FunctionConfigUtils.doJavaChecks()` | Add `IncrementalWindowFunction` to the allowed user-class interfaces. |
+| `FunctionCommon.getFunctionClassParent()` | When `windowConfig` is set, resolve `IncrementalWindowFunction` before `WindowFunction` so input/output type inference for SerDe and schema checks stays correct. |
+
+
+## Public-facing Changes
+
+### Public API
+
+#### New interface: `org.apache.pulsar.functions.api.Window<T>`
+
+Promoted from the internal package.
+
+| Method | Description |
+|--------|-------------|
+| `List<T> get()` | All events in the current window |
+| `List<T> getNew()` | Events added since the last trigger |
+| `List<T> getExpired()` | Events removed since the last trigger |
+| `Long getStartTimestamp()` | Window start time (non-null for time-based windows, otherwise `null`) |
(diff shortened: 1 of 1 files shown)


dao-jun:
overall LGTM, just 2 points:
  1. The FunctionCommon.getFunctionClassParent() change is more critical than the PIP describes — it's not about "keeping type inference correct," it's that without it, deployment will NPE (assuming doJavaChecks is fixed first). The call chain is getFunctionTypes → getFunctionClassParent → resolveInterfaceTypeArguments.
  When the parent interface is misidentified as java.util.function.Function, resolveInterfaceTypeArguments traverses the interface hierarchy and never finds Function (since IncrementalWindowFunction doesn't extend it), returns null, and typeArgsList.get(0) throws NPE. This should be documented as a hard dependency, not a
  correctness improvement.
  2. Window.get() / getNew() / getExpired() return mutable List<T>. While each trigger creates fresh lists so mutations can't corrupt internal state, once this becomes a public API the Javadoc should explicitly state whether callers are allowed to modify the returned lists.

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Downgrade / Rollback
+
+To roll back to a version without this feature:
+
+- User functions that implement `IncrementalWindowFunction` must be rewritten to implement `WindowFunction`, replacing `getNew()` / `getExpired()` logic with manual diffing over the full message collection, before they can be deployed on the older version.
+
+## Pulsar Geo-Replication Upgrade & Downgrade/Rollback Considerations
+
+There is no wire-protocol change between Functions Workers. No special geo-replication considerations apply.
```
david-streamlio, line 537: The template requires three sections that are currently missing — could you add them, even if brief?

- **Security Considerations** — this is a pure API addition with no new endpoints, so a sentence confirming "no new REST/protocol surface, no new auth or multi-tenancy implications" is enough.
- **Monitoring / Metrics** — please state explicitly "no new metrics; runtime behavior is unchanged."
- **Alternatives** — the most important one. Why a *new* interface rather than (a) `default` methods on `WindowFunction`, (b) an overloaded `process(Window, ...)`, or (c) a config flag? Documenting why these were rejected will pre-empt the obvious review questions. It's also the right place to defend the name `IncrementalWindowFunction`, since it exposes expired events too, not just increments.

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+public interface Window<T> {
+    List<T> get();
+    List<T> getNew();
+    List<T> getExpired();
+    Long getEndTimestamp();
+    Long getStartTimestamp();
+}
+```
+
+The existing internal `Window.java` is replaced by a reference to the `api-java` interface (or removed entirely, with `WindowImpl` implementing the new public interface directly).
```
david-streamlio: Promoting an internal type to public API is exactly the surface the PIP process exists to scrutinize, so this shouldn't be left as an either/or ("replaced by a reference … or removed entirely"). Please commit to one approach and spell out what happens to any existing references to the old `org.apache.pulsar.functions.windowing.Window` (even though it's an internal package today).

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+```java
+@FunctionalInterface
+public interface IncrementalWindowFunction<X, T> {
+    /**
+     * Process the triggered window.
+     *
+     * @param inputWindow the window view for this activation, providing access to
+     *                    all current events ({@link Window#get()}),
+     *                    newly added events ({@link Window#getNew()}), and
+     *                    expired events ({@link Window#getExpired()}).
```
david-streamlio: In addition to the list-mutability question already raised on this PR, please document the **lifetime** of the `Window` reference: is it valid only during the `process()` call, or may a user retain it across triggers? Lifetime/ownership contracts matter once this interface is public.

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+```
+
+### Change 3: Update `WindowFunctionExecutor`
+
+**Path**: `pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java`
+
+#### 3a. Add field
+
+```java
+protected IncrementalWindowFunction<T, X> incrementalWindowFunction;
```
david-streamlio: The public interface is declared `IncrementalWindowFunction<X, T>` (X=input, T=output), but this executor field is `<T, X>`. This matches the internal `WindowFunction<T,X>` convention, so it's defensible — but the doc shows both orderings without comment, which will trip readers. A one-line note clarifying the convention would help.

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+Detect `IncrementalWindowFunction` via `instanceof` in `initializeUserFunction()`, following the same pattern used for `WindowFunction` today:
+
+```java
+@SuppressWarnings("unchecked")
+private void initializeUserFunction(WindowConfig windowConfig) {
+    // ...
+    if (userClassObject instanceof java.util.function.Function) {
+        // existing logic, unchanged
+        bareWindowFunction = ...;
+    } else if (userClassObject instanceof IncrementalWindowFunction) {
```
david-streamlio: The dispatch order here is `Function` → `IncrementalWindowFunction` → `WindowFunction`. A user class implementing both `IncrementalWindowFunction` and `WindowFunction` (or both `Function` and `IncrementalWindowFunction`) resolves by this precedence. Since that becomes an observable public-API contract, please state the ordering explicitly and confirm it's intentional.

david-streamlio (commented):
Review of PIP-484. Overall this is a clear, well-scoped proposal — good background, an honest backward-compat analysis, a concrete example, and a diagram; it passes the "can a reader understand it without hours of code reading" health check. The inline comments below are (1) a few required template sections that are missing and (2) some public-API design points to nail down. I've left out the two points already raised on the PR (the getFunctionClassParent NPE dependency and list mutability).

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Change 4: Update deployment validation (`functions/utils`)
+
+Submit-time validation must accept `IncrementalWindowFunction` the same way it already accepts `WindowFunction`.
+
+| File | Change |
+|------|--------|
+| `FunctionConfigUtils.doJavaChecks()` | Add `IncrementalWindowFunction` to the allowed user-class interfaces. |
+| `FunctionCommon.getFunctionClassParent()` | When `windowConfig` is set, resolve `IncrementalWindowFunction` before `WindowFunction` so input/output type inference for SerDe and schema checks stays correct. |
+
```
david-streamlio: Not required by the template, but reviewers usually ask: a sentence on intended test coverage (executor dispatch for each interface type, and deployment-validation acceptance of the new interface) would strengthen the proposal.

pip/pip-484.md
```diff
@@ -0,0 +1,336 @@
...
+Promoted from the internal package.
+
+| Method | Description |
+|--------|-------------|
+| `List<T> get()` | All events in the current window |
+| `List<T> getNew()` | Events added since the last trigger |
+| `List<T> getExpired()` | Events removed since the last trigger |
+| `Long getStartTimestamp()` | Window start time (non-null for time-based windows, otherwise `null`) |
+| `Long getEndTimestamp()` | Window end time (watermark in event-time mode, system time in processing-time mode) |
```
david-streamlio: `getStartTimestamp()` documents its null behavior, but `getEndTimestamp()`'s description implies it is never null. Please confirm and capture this in the Javadoc, since both methods are now public.

david-streamlio:
A few PIP-process items (separate from the proposal content):

- **PR title** is `[improve][fn]`, but the [PIP README](https://github.com/apache/pulsar/blob/master/pip/README.md) specifies `[improve][pip] PIP-xxx: {title}` for PIP PRs. Worth aligning (the semantic-PR CI may flag it).
- **No `PIP` label** is applied yet — it's needed for the PIP index/tracking query.
- Please **confirm PIP-484 is still the next free number** — numbers are assigned by open-PR order and can race with other in-flight PIPs.
- The **Links** section still has `TBD` for the `[DISCUSS]`/`[VOTE]` mailing-list threads — expected at this stage, just a reminder to fill them in before the vote concludes.

pip/pip-484.md
Dream95: Option A seems better than the current approach of adding new interfaces. Let me think about whether there are any compatibility issues.

Dream95:
@dao-jun  @david-streamlio Thanks for the review.
I've revised PIP-484: instead of adding IncrementalWindowFunction, we now expose the public Window interface and extend WindowFunction with a default process(Window, WindowContext) overload. Existing WindowFunction users (including lambdas) remain backward compatible, and no deployment-validation changes are required.

pip/pip-484.md
```diff
@@ -0,0 +1,541 @@
...
+
+Because the new method is `default`, adding it is source-compatible and binary-compatible for existing `WindowFunction` implementations.
+
+#### Example: sliding-window sum
+
+```java
+/**
+ * Maintains the sum of integer values in the current sliding window incrementally.
+ */
+public class SlidingWindowSumFunction implements WindowFunction<Integer, Integer> {
```
david-streamlio, line 329: This example surfaces a footgun worth calling out in the prose. The abstract `process(Collection, …)` is what every user *must* implement, but once a class also overrides the `process(Window, …)` default, the executor only ever calls the `Window` overload — so the `Collection` override becomes **dead code at runtime**. Here the two overloads even return *divergent* results (`stream().sum()` vs. the counter-based running sum), which reads as if both paths are live.

Suggest either (a) having the `Collection` method share logic with the `Window` method (e.g. the `Window` override calls `process(window.get(), context)` for the base case), or (b) dropping the second override from the example — and adding one sentence: "once you override `process(Window, …)`, the runtime no longer invokes any `process(Collection, …)` override."

pip/pip-484.md
```diff
@@ -0,0 +1,541 @@
...
+        }
+        long expiredSum = 0;
+        for (Record<Integer> record : window.getExpired()) {
+            expiredSum += record.getValue();
+        }
+        long netDelta = newEventsSum - expiredSum;
+        if (netDelta != 0) {
+            context.incrCounter(RUNNING_SUM_KEY, netDelta);
+        }
+        return (int) context.getCounter(RUNNING_SUM_KEY);
```
david-streamlio: Minor / optional: using durable counter state (`incrCounter`/`getCounter`) to hold a *window-scoped* aggregate is fragile as a reference example — the counter and the window have different lifetimes, so an instance restart drifts the result (counter state persists while the window is rebuilt, or vice-versa). For teaching incremental aggregation, a local accumulator field (with the per-instance caveat noted) or returning the net delta against a maintained value would be clearer and wouldn't imply that durable state is the intended pattern.

pip/pip-484.md
```diff
@@ -0,0 +1,541 @@
...
+
+The new overload is a default method that delegates to the existing abstract `process(Collection<Record<X>>, WindowContext)` method:
+
+```java
+default T process(Window<Record<X>> window, WindowContext context) throws Exception {
+    return process(window.get(), context);
+}
+```
+
+This keeps `WindowFunction` a valid `@FunctionalInterface`. Existing lambdas continue to implement the collection-based abstract method and cannot override the default overload, so lambda behavior remains unchanged. Users that need `getNew()` / `getExpired()` implement `WindowFunction` as a class and override the new default overload.
```
david-streamlio: This is the central tradeoff of the chosen design and I'd make it more prominent (it's currently only implied here and in Alternative 2): **lambda implementations cannot access the incremental views at all**, since they bind to the abstract method and can't override a default. Given that lambdas are a very common window-function authoring style, this is exactly the point I'd want the mailing-list discussion/vote to ratify explicitly — Alternative 2 (`WindowContext.getWindow()`) would cover lambdas too, at the cost of weaker typing. No change needed to the design; just recommend stating the lambda limitation as an accepted, voted-on tradeoff rather than a footnote.

pip/pip-484.md
```diff
@@ -0,0 +1,541 @@
...
+     * this reference timestamp for activated windows. If no reference timestamp is
+     * available, this method returns {@code null}.
+     * </p>
+     */
+    Long getEndTimestamp();
+
+    /**
+     * The window start timestamp for this activation.
+     * <p>
+     * This is computed from {@link #getEndTimestamp()} and the configured
```
david-streamlio: Please confirm this wording matches how the runtime actually populates these. `WindowImpl` just stores constructor-injected `startTimestamp`/`endTimestamp`; the "start = end − window length" computation lives in the caller (`WindowManager`). Since these are now public-API contracts, worth double-checking the sliding-window case in particular (window length vs. sliding interval) so the Javadoc doesn't over-promise.

david-streamlio (commented):
Thanks for the substantial revision — the pivot from a new `IncrementalWindowFunction` interface to a `default process(Window, WindowContext)` overload on the existing `WindowFunction` is a cleaner direction, and it resolves most of the prior round's concerns as a side effect (no new submit-time validation needed, no multi-interface dispatch precedence to document, type-param ordering now consistent). The template is now complete (Security / Monitoring / Alternatives all present) and the backward-compat + classloader analysis is thorough. I verified the executor/interface claims against current `master` and they hold.

Posting as **COMMENTED** — the remaining items are example quality plus one design tradeoff I think the mailing-list vote should explicitly ratify, not correctness defects in the proposal. Details inline.

david-streamlio:
@Dream95 The proposal reads well now and the design discussion has mostly converged here — at this point the main thing gating progress is the **mailing-list `[DISCUSS]` thread**, which is a required step in the [PIP process](https://github.com/apache/pulsar/blob/master/pip/README.md) before a vote can be called.

Could you kick that off? Roughly:

1. Send a `[DISCUSS] PIP-484: Expose Window interface for WindowFunction processing` email to `dev@pulsar.apache.org` (link back to this PR).
2. Once it's on the list, replace `Mailing List discussion thread: TBD` in the PIP with the lists.apache.org permalink.
3. Let it run ~a week / until comments settle, fold in any feedback, then start the `[VOTE]` thread and fill in that link too.

That moves the design conversation to where the binding +1s happen. The review comments I left are all polish and can be folded in during the discussion — they're not blockers. Thanks for the iterations on this!

Dream95:
@dao-jun @david-streamlio Thanks again for reviewing PIP-484 earlier.

The mailing-list `[VOTE]` thread is now open:
https://lists.apache.org/thread/mcnl2tf5lppz52vkk8qrby2xzn55gfz1

If you have a chance, could you please cast a vote there? Happy to address any remaining concerns.

david-streamlio:
@Dream95 apologies for the long silence on this — you asked in July and I let it sit.

I looked at why the vote stalled rather than just replying late, and I think the thread never had a
chance: [the VOTE thread](https://lists.apache.org/thread/mcnl2tf5lppz52vkk8qrby2xzn55gfz1) has been
open since 22 July with no replies at all, and the two people it asked — @dao-jun and me — are both
committers rather than PMC members. Per [`pip/README.md`](https://github.com/apache/pulsar/blob/master/pip/README.md):

> Everyone is welcome to vote on the proposal, though only the vote of the PMC members will be
> considered binding. The requirement is to have at least one binding +1 vote from a lazy majority if
> no binding -1 votes have been cast on the PIP.

So neither of our votes can carry it, which is almost certainly why it went quiet rather than any
objection to the proposal. The good news is the bar is one binding +1, and nobody has raised a
concern on either thread.

For what it is worth: **+1 (non-binding)**. The revision you made after the June discussion is the
right shape — exposing the public `Window` interface and adding a `default process(Window,
WindowContext)` overload keeps every existing `WindowFunction`, lambdas included, source-compatible,
and it avoids the extra deployment-validation surface the original `IncrementalWindowFunction` would
have needed.

Suggested next step: since the original thread is seven weeks old, I would start a fresh
`[VOTE] PIP-484: ...` on dev@ rather than reviving that one, link both the PIP PR and the earlier
DISCUSS and VOTE threads, and explicitly ask for a binding vote from a PMC member. The PIP guide
endorses exactly this nudge when a vote goes unanswered. I am happy to reply with my non-binding +1
on the new thread so it does not start empty.

If you would rather not restart it, replying to the existing thread asking specifically for a binding
+1 would probably also do the job.

david-streamlio:
@lhotari when you have a moment — would you be willing to cast a binding vote on this one?

PIP-484 has been sitting since 22 July. The [VOTE thread](https://lists.apache.org/thread/mcnl2tf5lppz52vkk8qrby2xzn55gfz1) has no replies at all, and I think the reason is procedural rather than substantive: the only two people it asked, @dao-jun and me, are both committers, so neither of us can supply the binding +1 the process needs. Nobody has raised an objection on either the DISCUSS or the VOTE thread.

It is a small proposal — it exposes the existing public `Window` interface and adds a `default process(Window, WindowContext)` overload to `WindowFunction`, so current implementations and lambdas stay source-compatible. `WindowManager` already tracks the new and expired events internally; this just surfaces them.

No rush, and no objection if you would rather it went through a fresh vote thread first — mainly flagging that one binding vote is the whole gap.

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+
+On each window trigger, `WindowFunctionExecutor` builds a new `WindowImpl` with freshly collected lists, injects `getEndTimestamp()` from the trigger's reference time and `getStartTimestamp()` from `referenceTime - windowLengthDurationMs` (when applicable), and passes the view to `WindowFunction.process(Window<Record<X>>, WindowContext)`. The runtime does not retain that object after `process()` returns.
+
+| Contract | Rule |
+|----------|------|
+| **`Window` lifetime** | The `inputWindow` argument is valid **only for the duration of a single `process()` call**. Implementations must not store the `Window` reference (or any list obtained from it) in a field, cache, or closure for use on a later trigger. Each activation receives a new `Window` instance. |
+| **List mutability** | `get()`, `getNew()`, and `getExpired()` return mutable `List` instances that are **snapshots for the current activation**. Callers **must not modify** them. Mutations do not update the runtime's live window state (the lists are copies), but modification is unsupported and must not be relied upon. Javadoc on each getter states this explicitly. |
```
lhotari, line 398: **[DESIGN] The list-mutability contract is wrong: `get()` hands back the exact list the runtime acks from**

"Mutations do not update the runtime's live window state (the lists are copies)" is not true for `get()`. `WindowFunctionExecutor.processWindow()` passes the same `tuples` list into `WindowImpl` and then iterates *that list* to ack the records after `process()` has returned:

https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java#L230-L249

So a `WindowFunction` that does `window.get().clear()` or `window.get().removeIf(...)` — plausible for "drain the window" style code, and the sort of thing a user does precisely *because* the list is mutable — silently stops those records from being acked under `ATLEAST_ONCE`. The records are never acked, they are redelivered, and the backlog grows with no error anywhere. In the `output != null` branch the ack loop runs inside a `thenAccept` callback after the publish completes, so the mutation is read back asynchronously, later than the user is likely to expect.

This hazard already exists today via the `Collection<Record<X>>` parameter, so the PIP is not introducing it. But the PIP is the moment it stops being an accident of an internal type and becomes a documented public contract, and the sentence as written actively reassures the reader that mutation is harmless. [`pip-484.md:271`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pip/pip-484.md#L271) then declines the one-line fix on the strength of that reasoning.

Since `Window` is proposed as `@InterfaceStability.Stable`, this is the only chance to decide it. Two options, either of which I would be happy with — the PIP just needs to pick one and say why:

1. Wrap the three getters in `Collections.unmodifiableList(...)` in `WindowImpl`. Note this *is* an observable behaviour change for any existing user who mutates the collection today (they would start getting `UnsupportedOperationException` instead of silently losing acks), so it belongs in the Backward Compatibility section rather than being introduced as a footnote.
2. Keep the lists mutable but have the executor snapshot the ack list before calling `process()`, so user mutation cannot reach the ack path. Then the "mutations only affect a copy" sentence becomes true, and the Javadoc prohibition is advisory rather than load-bearing.

What should not survive is the current combination: mutable lists, a Javadoc-only prohibition, and a normative sentence that misstates what happens if the prohibition is ignored. If the decision is to defer this, please mark the interface `@InterfaceStability.Evolving` rather than `Stable` so it can still be tightened.

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+}
+```
+
+This keeps `WindowFunction` a valid `@FunctionalInterface`. Existing lambdas bind to the abstract collection-based method and keep today's behavior.
+
+`WindowFunctionExecutor` always calls `process(Window<Record<X>>, WindowContext)`. The default overload delegates to `process(window.get(), context)` unless a class overrides the window overload.
+
+### Design tradeoff
+
+- **Lambdas cannot access incremental views** — they cannot override the default `process(Window, …)` method, so `getNew()` / `getExpired()` require a class-based implementation.
```
lhotari, line 134: **[DESIGN] The lambda tradeoff is compile-time only — a deployed window function cannot be a lambda**

This is stated as the central drawback of the chosen design, and the discussion on this PR asked the mailing list to ratify it explicitly. It is worth checking first, because I think it is largely not a real limitation.

The Java Functions runtime instantiates the user's window function **by class name, via a no-arg constructor** — [`WindowFunctionExecutor.java:69-76`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java#L69-L76) calls `Reflections.createInstance(actualWindowFunctionClassName, clsLoader)`, and [`Reflections.java:110-141`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-common/src/main/java/org/apache/pulsar/common/util/Reflections.java#L110-L141) resolves that name with `Class.forName` and invokes the declared no-arg constructor. A lambda has no such name, so it cannot be the value of `--classname`. Consistently, every `WindowFunction` in the repository is a named class (the seven examples under `pulsar-functions/java-examples/.../api/examples/window/`).

So "lambdas cannot access the incremental views" is true as a statement about the Java language and false as a statement about deployed functions: there are no deployed `WindowFunction` lambdas to lose anything. The same correction applies to Alternative 2's listed advantage "Lambdas can access incremental data without converting to a class" ([`pip-484.md:489`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pip/pip-484.md#L489)) — that is not a deployment advantage either, which strengthens the case for the design you picked.

Suggestion: replace the tradeoff bullet and the repeated "including lambdas" compatibility claims with one accurate sentence — something like "the runtime instantiates window functions by class name, so a deployed `WindowFunction` is always a class and can always override the new overload; the lambda restriction applies only to in-process/test usage." That removes a stated blocker from the vote instead of asking the list to accept it.

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+     *
+     * <p>This remains the single abstract method of the interface, so existing
+     * lambda implementations and class implementations continue to work without
+     * source changes.</p>
+     *
+     * @param input   all current records in the window
+     * @param context the window function context
+     * @return the output, or {@code null} to suppress output
+     */
+    T process(Collection<Record<X>> input, WindowContext context) throws Exception;
```
lhotari, line 293: **[DESIGN] An incremental implementation must still write a `process(Collection, …)` body that the runtime never calls — say what goes in it**

This stays the single abstract method, which is what keeps the change source- and binary-compatible — good. The consequence is that anyone who wants incremental access is *required* to write a `process(Collection, …)` body which, per the Design tradeoff at [`pip-484.md:135`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pip/pip-484.md#L135), the runtime will then never invoke. That mandatory-dead method is a permanent wart in a `Stable` interface, and it is the real cost the vote is being asked to accept in exchange for not adding a second interface (Alternative 1). I think that trade is defensible, but the PIP should state it in those terms rather than as a footnote, because it is what reviewers will re-litigate later.

More practically: the PIP should say what belongs in that body, because whatever the example shows will be copy-pasted for years. The example at [`pip-484.md:329-350`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pip/pip-484.md#L329-L350) currently gives the two overloads *divergent* semantics — `input.stream().mapToInt(...).sum()` versus a counter-based running sum — with only a comment explaining that one of them is dead. A reader skimming it will reasonably conclude both paths are live.

The pattern worth modelling is the delegating one: put the full-window computation in `process(Collection, …)` and have `process(Window, …)` call it when it cannot do the incremental thing (first activation, restart, expiry it cannot reconcile). That makes the required method genuinely useful instead of dead, and it makes the example self-consistent. Alternatively drop the `Collection` override from the example entirely and show only the minimum an incremental function must declare.

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+|------|--------|
+| `api-java/.../api/Window.java` | **Added** — canonical public interface (Javadoc, lifetime/mutability contracts) |
+| `windowing/Window.java` | **Removed** |
+| `windowing/WindowImpl.java` | `implements org.apache.pulsar.functions.api.Window<T>` |
+| `windowing/WindowFunctionExecutor.java` | import `org.apache.pulsar.functions.api.Window` |
+| `windowing/WindowFunctionExecutorTest.java` | import `org.apache.pulsar.functions.api.Window` |
+
+The old FQCN was never part of the supported public Functions API. Within this repository, only the files above referenced it; all are migrated to `org.apache.pulsar.functions.api.Window`. No deprecated shim is shipped in Apache Pulsar itself.
+
+**Optional compatibility note (forks / downstream only)**
```
lhotari, line 255: **[MINIMALISM] Drop the "Optional compatibility note (forks / downstream only)" — it is explicitly not part of the proposal**

This block, and the paragraph that repeats it at [`pip-484.md:523`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pip/pip-484.md#L523), describe a deprecated shim that the PIP itself says is "not part of the Apache Pulsar change set" and "a local fork decision". A PIP is the project's decision record; specifying what a downstream fork may do in its own tree is out of its remit, and the only effect of keeping it is that a reader can come away thinking Pulsar ships a `@Deprecated org.apache.pulsar.functions.windowing.Window`. Cutting the two blocks removes about 20 lines and one source of confusion.

The deletion itself is well justified and I verified the migration table is exactly right: the only in-tree references to the internal type are [`WindowImpl.java:26`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowImpl.java#L26), [`WindowFunctionExecutor.java:320`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java#L320) and [`WindowFunctionExecutorTest.java:58-61`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/test/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutorTest.java#L58-L61), and because all three live in `org.apache.pulsar.functions.windowing` today none of them even has an import to change — the change is "add one import" in each file, as the table says. Saying that plainly is a stronger argument for deleting the internal type than the fork note is.

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+### Change 3: Update `WindowFunctionExecutor`
+
+**Path**: `pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java`
+
+No new field or user-function detection path is needed. `initializeUserFunction()` continues to accept `WindowFunction` exactly as it does today.
+
+Update `process(Window<Record<X>>, WindowContext)` to pass the full window view to `WindowFunction` (see **Design tradeoff** above).
+
+```java
+public T process(Window<Record<X>> inputWindow, WindowContext context) throws Exception {
```
lhotari, line 362: **[MINIMALISM] The `WindowFunctionExecutor` type-parameter rename is unexplained scope and makes the Change 3 snippet not match the file**

The class is `WindowFunctionExecutor<T, X>` today ([`WindowFunctionExecutor.java:47`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java#L47)) and the method is `public X process(Window<Record<T>> inputWindow, WindowContext context)` ([`WindowFunctionExecutor.java:320`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/WindowFunctionExecutor.java#L320)). This snippet silently shows the parameters swapped, and the rename is mentioned only in a parenthetical inside a comment at [`pip-484.md:80`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pip/pip-484.md#L80) — Change 3 itself never says it is happening.

A reader diffing this snippet against the file will think one of them is wrong. Two ways out, either fine: drop the rename and keep the implementation diff to the one line that actually changes (`inputWindow.get()` → `inputWindow`), or promote it to an explicit numbered change with a sentence saying why. The first is smaller and I would prefer it — the executor's parameter names are internal and the reordering does not serve the goal of the PIP.

(An earlier thread on this PR raised the same `<X,T>` vs `<T,X>` inconsistency against the previous revision; it is still present in the current text.)

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+     * Populated by {@code WindowFunctionExecutor} when it constructs the
+     * {@code WindowImpl} for each trigger — {@code WindowImpl} stores the value
+     * as supplied; it does not derive timestamps itself. For event-time windows,
+     * this is the reference time from the trigger policy (watermark-based). For
+     * processing-time windows, this is the processing-time reference from the
+     * trigger policy. Returns {@code null} if the trigger provides no reference
+     * timestamp.
+     * </p>
+     */
+    Long getEndTimestamp();
```
lhotari, line 219: **[NIT] `getEndTimestamp()`'s documented null case does not occur; the count-window semantics that will confuse people are undocumented**

"Returns `null` if the trigger provides no reference timestamp" describes a state no current trigger policy can produce. All four set a non-null reference time: [`CountTriggerPolicy.java:54`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/triggers/CountTriggerPolicy.java#L54) uses `System.currentTimeMillis()`, [`TimeTriggerPolicy.java:115`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/triggers/TimeTriggerPolicy.java#L115), [`WatermarkCountTriggerPolicy.java:85`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/triggers/WatermarkCountTriggerPolicy.java#L85) and [`WatermarkTimeTriggerPolicy.java:89`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/triggers/WatermarkTimeTriggerPolicy.java#L89) all pass a timestamp. Documenting a defensive `null` is harmless, but it is not the thing users will trip on.

The thing they will trip on is that **for a count-based window `getEndTimestamp()` is a wall-clock reading taken at trigger time, not a window boundary**, while `getStartTimestamp()` is `null` there (`getWindowStartTs()` returns `null` when `windowLengthDurationMs` is unset). A non-null "end" with a null "start", where the end does not bound anything, is exactly the asymmetry worth stating in the Javadoc — and it is being frozen as public API, so it is worth getting right now.

The rest of this section checks out: the executor really does compute `start = referenceTime - windowLengthDurationMs`, and the sliding-window clarification (window length, not sliding step) is correct.

pip/pip-484.md
```diff
@@ -0,0 +1,550 @@
...
+
+**Current path**: `pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/windowing/Window.java`
+
+**New path**: `pulsar-functions/api-java/src/main/java/org/apache/pulsar/functions/api/Window.java`
+
+The interface methods are unchanged in signature; the package moves to the public API module :
+
+```java
+// pulsar-functions/api-java/.../api/Window.java
+@InterfaceAudience.Public
```
lhotari, line 169: **[NIT] In the `Window` snippet the annotations precede the Javadoc, so the comment would not attach to the declaration**

`@InterfaceAudience.Public` / `@InterfaceStability.Stable` are placed above the class Javadoc here. That compiles, but the doc comment is no longer the last thing before the declaration, so javadoc does not associate it with the interface and the class-level documentation — including the lifetime contract, which is the part that matters most — silently disappears from the generated API docs.

The existing [`WindowFunction.java:26-32`](https://github.com/apache/pulsar/blob/e78acd7d91c28e7fc35f38d2dd61ce3e9aa7909e/pulsar-functions/api-java/src/main/java/org/apache/pulsar/functions/api/WindowFunction.java#L26-L32) has the order the PIP should copy: Javadoc first, then the annotations, then `public interface`. Purely a snippet fix, but this snippet is what the implementation PR will be written from.

lhotari (commented):
Read the proposal end to end and checked every claim it makes about the current runtime against `pulsar-functions/instance` at this head. The June revision is the right shape: promoting the existing `Window<T>` view instead of adding a second user-function interface avoids the deployment-validation and dispatch-precedence costs that the original `IncrementalWindowFunction` design would have carried, and it surfaces data the runtime already computes rather than adding windowing logic. Every factual claim I could check holds — the migration really is three files, the parent-first classloader argument is correct (both `FunctionClassLoaders.ParentFirstClassLoader` and `NarClassLoader` delegate to the parent first), `getWindowStartTs()` really is `referenceTime - windowLengthDurationMs` in the executor, and the template sections are all present.

One substantive thing should be settled in the text before this is frozen as `@InterfaceStability.Stable`: the **list-mutability contract is wrong about the consequences of mutation**, and this is the last moment it can be changed (see the first comment). Everything else below is either minimalism — the PIP would be stronger about 40 lines shorter — or wording.

A note on the lambda tradeoff, since it is the one thing the discussion asked the list to ratify: it is a compile-time-only concern. A deployed Java window function is instantiated by class name with a no-arg constructor, so a lambda is not a deployable shape at all, and the tradeoff mostly evaporates. That is an argument *for* the chosen design, not against it.

On the stalled vote: the last substantive change to the text was 2026-06-21 (`f04f80d0f`); the only commit after it, `e78acd7d9`, adds the vote link — so the text on the list is the text here, and nothing has drifted. The 22 July thread has no replies at all. Given the age, a fresh `[DISCUSS]`-free `[VOTE]` thread with the mutability point resolved in the text looks like the shortest path to a binding +1.

pip/pip-484.md
Dream95: Thanks for catching this. I chose option 2 and snapshot the ack list before process(). This protects acks without causing existing functions that modify the input to throw UnsupportedOperationException.

lhotari (approved):
LGTM. The revised proposal addresses my previous comments, including the ack snapshot, deployment semantics, and full-window fallback example.

pip/pip-484.md
lhotari: Thanks, option 2 addresses this concern. The proposed copy is taken before user code runs, so both ack paths can use it without changing input-list mutability.

https://github.com/apache/pulsar/blob/2f0a4c0e8461b364b3acd3ba9aa81e0a5c0d643c/pip/pip-484.md#L359-L366

lhotari:
@david-streamlio Please complete your review so that we can merge after the PIP vote has been completed.

lhotari:
[A single binding vote is sufficient](https://github.com/apache/pulsar/blob/master/pip/README.md#how-does-the-pip-process-work) for the PIP vote. https://lists.apache.org/thread/mcnl2tf5lppz52vkk8qrby2xzn55gfz1 . @Dream95 you can proceed to close the PIP vote while David completes his review.

david-streamlio:
# Review: PIP-484 — two blocking concerns

The API shape looks right to me: the default-overload approach is genuinely source- and binary-compatible, the reasoning against both alternatives is convincing, and the ack-snapshot detail shows real care. I checked the doc's claims against `master` — the migration table is complete, `getWindowStartTs` matches the proposed Javadoc, and the parent-first classloader argument holds.

My concern is narrower: `getNew()`/`getExpired()` are today an internal detail that no user-visible behavior depends on, so the runtime has never had to deliver them reliably — and it doesn't. Promoting them to `@InterfaceStability.Stable` public API with an incremental-aggregation example attached needs those gaps closed or specified first.

## 1. Expired events are silently dropped whenever the window drains to empty

`WindowManager.java:130-131` drains `expiredEvents` into a local `expired`. `WindowManager.java:145-153` passes it to `onActivation` **only when `events` is non-empty** — otherwise it is discarded, and `WindowFunctionExecutor`'s `onExpiry` is a no-op (`WindowFunctionExecutor.java:206-208`). The expirations are gone.

This is routine, not a corner case: `TimeTriggerPolicy` fires `onTrigger()` unconditionally on a fixed-rate timer (`TimeTriggerPolicy.java:73-75`). Window length 10s, slide 5s, traffic stops → a trigger lands with everything expired and nothing remaining → the entire expired batch vanishes.

The repo's own test already asserts this. `WindowManagerTest.testTimeBasedWindowExpiry:320-331`: after all 10 events expire, `onActivationEvents` is empty and `onActivationExpiredEvents` is *still* empty on the following trigger — they reached only `onExpiry`. That test passes on master, so this is verified current behavior, not inference.

Impact: **the PIP's own `SlidingWindowSumFunction` example produces permanently wrong output.** `runningSum` never receives the decrement for the dropped batch and stays skewed for the life of the instance. Under today's API this is harmless — an empty window has nothing to compute — and this PIP is precisely what makes `getExpired()` load-bearing.

The PIP needs to pick one:
- Deliver `onActivation` when `events` is empty but `expired` is not. That is a behavior change worth calling out explicitly, since existing functions would start seeing `process()` invocations with an empty `get()`.
- Or specify in the `getExpired()` contract that expirations are best-effort and may be dropped — in which case the incremental example must go, because it is not implementable on that contract.

## 2. `getExpired()` can contain records that never appeared in any `getNew()`

`expiredEvents` accumulates from `compactWindow()` on every `add()` (`WindowManager.java:112`, `171-173`), independently of triggers, and `scanEvents` removes expiring events from the queue before they can reach `windowEvents`. With slide > window length — a legal config, e.g. windowLength 10s / slide 60s — records that arrive and expire between two triggers surface in `getExpired()` at the next trigger having never appeared in `get()` or `getNew()`. A decrement-on-expiry aggregate drifts negative.

This belongs in the `Window` Javadoc contract, alongside either a way for the user to detect it or the removal of the running-aggregate guidance.

## Suggestion

Since there's no implementation PR yet, the cheapest path is probably to amend the contract sections of the PIP and fix delivery in the implementation, rather than re-vote. Both findings are about what the runtime guarantees, not about the API shape — which I think is sound as proposed.

---

Verified by reading `WindowManager`, `WindowFunctionExecutor`, the trigger/eviction policies and `WindowManagerTest`, plus one green run of `:pulsar-functions:pulsar-functions-instance:test --tests WindowManagerTest -PtestRetryCount=0`.

david-streamlio:
Follow-up to my previous comment — the non-blocking half of the review. None of these should hold up the PIP on their own.

## 3. `@InterfaceStability.Stable` is too strong for a new interface

The PIP marks the new `org.apache.pulsar.functions.api.Window` as `@InterfaceStability.Stable`. Every interface in `functions.api` carrying `Stable` is years old — `Context`, `Function`, `Record`, `SerDe`, `WindowContext`, `WindowFunction`, `BaseContext`, `KVRecord`. The two newer additions in that same package, `StateStore` and `StateStoreContext`, both use `Evolving`.

Given that the two blocking findings may change the `getExpired()` contract, I'd ship `Window` as `Evolving` and promote it after a release.

## 4. Failure semantics are unspecified — and that's what incremental state breaks on

`WindowManager.onTrigger` updates `prevWindowEvents` at `WindowManager.java:146`, **before** calling `onActivation` at line 150. If the user function throws, `processWindow` rethrows as `RuntimeException`, that activation's deltas are gone — the next trigger won't re-report them as new, because `prevWindowEvents` already absorbed them — and `triggerPolicy.reset()` at line 155 is skipped.

Downstream of that, behavior diverges by trigger policy:

- `TimeTriggerPolicy` wraps the task in `catchingAndLoggingThrowables` (`TimeTriggerPolicy.java:73-75`), so the throwable is logged and swallowed and the fixed-rate timer keeps going.
- `CountTriggerPolicy.track` calls `handler.onTrigger()` inline, so the throwable propagates into the message-processing path.

So delta delivery is at-most-once, not failure-atomic, and inconsistent across policies. The "window state snapshot / checkpoint mechanisms" out-of-scope note doesn't cover this. It deserves its own Failure Semantics section: an incremental function that threw mid-update is permanently diverged, with no signal that it needs to resync.

## 5. The compatibility section understates the source break

The migration table lists only the three in-repo files. But `WindowFunctionExecutor` is public and non-final, and `process(Window<Record<T>>, WindowContext)` is a public overridable method — the repo's own `WindowFunctionExecutorTest:58-61` subclasses it and overrides exactly that method to collect `Window` objects. Anyone who did the same downstream gets a compile error on upgrade.

To be clear, I agree with deleting rather than shimming, and for a stronger reason than the PIP gives. A deprecated `windowing.Window extends api.Window` would be actively *worse*: an override typed on the old interface would become a silent overload that the runtime never calls, turning a compile error into wrong runtime behavior. That argument belongs in the PIP — it's what makes the clean break correct, and it warns downstream subclassers who currently get no heads-up.

## Nits

- The ack-copy in Change 3 fixes more than the PIP claims. `processWindow` acks inside `context.publish(...).thenAccept(...)` (`WindowFunctionExecutor.java:237-243`), so today the ack loop can race a user function still mutating the list on the function thread. Worth saying — it strengthens the case for the change.
- "Mutation cannot ... update `WindowManager`'s live event store" was already true before this change: `onActivation` hands over fresh lists built with `Collectors.toList()` (`WindowFunctionExecutor.java:216-218`). Reword so the only *new* guarantee claimed is the ack one.
- Mermaid node labels use `\n` for line breaks, which renders literally. Use `<br/>`.
- The doc alternates between the `WindowFunction<X, T>` and `WindowFunctionExecutor<T, X>` type-parameter conventions mid-page. A one-line note where `Window<Record<X>>` first appears would save readers a double-take.

lhotari:
@Dream95 do you have a chance to address the review comments and reply to David's comment? I'd like to get this into Pulsar 5.0.0 and the merge window is closing soon.

Dream95:
> # Review: PIP-484 — two blocking concerns
> The API shape looks right to me: the default-overload approach is genuinely source- and binary-compatible, the reasoning against both alternatives is convincing, and the ack-snapshot detail shows real care. I checked the doc's claims against `master` — the migration table is complete, `getWindowStartTs` matches the proposed Javadoc, and the parent-first classloader argument holds.
> 
> My concern is narrower: `getNew()`/`getExpired()` are today an internal detail that no user-visible behavior depends on, so the runtime has never had to deliver them reliably — and it doesn't. Promoting them to `@InterfaceStability.Stable` public API with an incremental-aggregation example attached needs those gaps closed or specified first.
> 
> ## 1. Expired events are silently dropped whenever the window drains to empty
> `WindowManager.java:130-131` drains `expiredEvents` into a local `expired`. `WindowManager.java:145-153` passes it to `onActivation` **only when `events` is non-empty** — otherwise it is discarded, and `WindowFunctionExecutor`'s `onExpiry` is a no-op (`WindowFunctionExecutor.java:206-208`). The expirations are gone.
> 
> This is routine, not a corner case: `TimeTriggerPolicy` fires `onTrigger()` unconditionally on a fixed-rate timer (`TimeTriggerPolicy.java:73-75`). Window length 10s, slide 5s, traffic stops → a trigger lands with everything expired and nothing remaining → the entire expired batch vanishes.
> 
> The repo's own test already asserts this. `WindowManagerTest.testTimeBasedWindowExpiry:320-331`: after all 10 events expire, `onActivationEvents` is empty and `onActivationExpiredEvents` is _still_ empty on the following trigger — they reached only `onExpiry`. That test passes on master, so this is verified current behavior, not inference.
> 
> Impact: **the PIP's own `SlidingWindowSumFunction` example produces permanently wrong output.** `runningSum` never receives the decrement for the dropped batch and stays skewed for the life of the instance. Under today's API this is harmless — an empty window has nothing to compute — and this PIP is precisely what makes `getExpired()` load-bearing.
> 
> The PIP needs to pick one:
> 
> * Deliver `onActivation` when `events` is empty but `expired` is not. That is a behavior change worth calling out explicitly, since existing functions would start seeing `process()` invocations with an empty `get()`.
> * Or specify in the `getExpired()` contract that expirations are best-effort and may be dropped — in which case the incremental example must go, because it is not implementable on that contract.
> 
> ## 2. `getExpired()` can contain records that never appeared in any `getNew()`
> `expiredEvents` accumulates from `compactWindow()` on every `add()` (`WindowManager.java:112`, `171-173`), independently of triggers, and `scanEvents` removes expiring events from the queue before they can reach `windowEvents`. With slide > window length — a legal config, e.g. windowLength 10s / slide 60s — records that arrive and expire between two triggers surface in `getExpired()` at the next trigger having never appeared in `get()` or `getNew()`. A decrement-on-expiry aggregate drifts negative.
> 
> This belongs in the `Window` Javadoc contract, alongside either a way for the user to detect it or the removal of the running-aggregate guidance.
> 
> ## Suggestion
> Since there's no implementation PR yet, the cheapest path is probably to amend the contract sections of the PIP and fix delivery in the implementation, rather than re-vote. Both findings are about what the runtime guarantees, not about the API shape — which I think is sound as proposed.
> 
> Verified by reading `WindowManager`, `WindowFunctionExecutor`, the trigger/eviction policies and `WindowManagerTest`, plus one green run of `:pulsar-functions:pulsar-functions-instance:test --tests WindowManagerTest -PtestRetryCount=0`.

@david-streamlio  For point 2, I think a better approach is to disallow slide > window length, which some other windowing engines already forbid. I’ll illustrate this in the example code first.

PR #25975, opened 2026-06-08
[improve][txn] PIP-473: GC aborted-transaction records on ML trim and segment drop

merlimat:
### Motivation

The metadata transaction buffer's durable aborted-transaction records (`/txn/segment-state/<segment>/aborted/<txnId>`) and its in-memory `abortedTxns` set grew for the segment's **entire lifetime**:

- They were never pruned when the segment's managed ledger **trimmed** the underlying data (retention / ack-based deletion within a live segment).
- They were never deleted when the whole **segment/topic was dropped** (`deleteSegmentWatermark` was dead code; `deleteScalableTopic` left the records orphaned).

The result is storage- and heap-growth proportional to the total number of aborts over the segment's life, not to the live data. This is the trim/drop GC the design always intended (each aborted record already stores its max position via `IDX_TXN_ABORTED_BY_POSITION`) but never wired up. It is a resource leak, not a data-correctness bug: trimmed positions are never dispatched, so their abort-filtering is no longer needed, and transaction ids never repeat.

### Modifications

- **`MetadataTransactionBuffer` — ML-trim GC.** Schedule a periodic task on the broker executor (cadence = `transactionCoordinatorScalableTopicsGcIntervalSeconds`; cancelled on close; null-guarded so mocked unit tests drive it directly). `pruneTrimmedAbortedTxns()` reads `ledger.getFirstPosition()` and range-deletes the durable aborted records — and evicts their `abortedTxns` entries — whose highest position in the segment is strictly below it (fully trimmed). Records for still-readable data are retained. New aborts land at the tail, so they are never pruned.
- **`TxnMetadataStore.deleteAllSegmentState(segment)`** — delete every aborted record plus the watermark for a segment (idempotent).
- **`ScalableTopicService.deleteScalableTopic`** — when transactions are enabled, clean up each segment's `/txn/segment-state` alongside deleting the segment topic.

### Verifying this change

Covered by unit tests:
- `MetadataTransactionBufferTest.pruneTrimmedAborted_dropsBelowFirstValid_retainsAbove` — a fully-trimmed aborted txn is dropped from both the in-memory set and the durable store (verified via a fresh TB's recovery); a still-readable aborted txn stays filtered.
- `TxnMetadataStoreTest.deleteAllSegmentState_removesAbortedRecordsAndWatermark`.
- `ScalableTopicServiceTest` (delete wiring) and `V5TransactionScalableTest` (e2e abort + topic cleanup) pass.

- Dependencies: no
- The public API: no
- The schema: no
- The default values of configurations: no
- The threading model: a per-segment-buffer periodic GC task is scheduled on the existing broker executor
- The binary protocol: no
- The REST endpoints: no
- The admin CLI options: no
- Anything that affects deployment: no

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
@@ -21,6 +21,7 @@ package org.apache.pulsar.broker.transaction.buffer.impl;
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.buffer.ByteBuf;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -29,6 +30,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAdder;
 import lombok.CustomLog;
 import org.apache.bookkeeper.mledger.AsyncCallbacks;
@@ -42,6 +46,7 @@ import org.apache.pulsar.broker.transaction.buffer.AbortedTxnProcessor;
 import org.apache.pulsar.broker.transaction.buffer.TransactionBuffer;
 import org.apache.pulsar.broker.transaction.buffer.TransactionBufferReader;
 import org.apache.pulsar.broker.transaction.buffer.TransactionMeta;
+import org.apache.pulsar.broker.transaction.metadata.AbortedTxnRecord;
 import org.apache.pulsar.broker.transaction.metadata.SegmentWatermark;
 import org.apache.pulsar.broker.transaction.metadata.TxnHeader;
 import org.apache.pulsar.broker.transaction.metadata.TxnIds;
@@ -132,6 +137,9 @@ public class MetadataTransactionBuffer implements TransactionBuffer {
     private final LongAdder committedCount = new LongAdder();
     private final LongAdder abortedCount = new LongAdder();
 
+    /** Periodic task that range-deletes aborted-txn records once the segment ML trims past them. */
+    private final ScheduledFuture<?> abortedGcTask;
+
     public MetadataTransactionBuffer(PersistentTopic topic, TxnMetadataStore txnStore) {
         this.topic = topic;
         this.ledger = topic.getManagedLedger();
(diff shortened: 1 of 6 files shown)


pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
```diff
@@ -140,6 +148,43 @@ public MetadataTransactionBuffer(PersistentTopic topic, TxnMetadataStore txnStor
...
+        long intervalMs = TimeUnit.SECONDS.toMillis(intervalSeconds);
+        return executor.scheduleWithFixedDelay(() -> {
+            if (closed) {
+                return;
+            }
+            pruneTrimmedAbortedTxns().exceptionally(ex -> {
+                log.warn().attr("segment", segmentName).exception(ex)
+                        .log("Aborted-txn GC sweep failed; will retry next cycle");
+                return null;
+            });
```
lhotari, line 191: Since `pruneTrimmedAbortedTxns` is an async task, it's possible that a previous task is running when a new one gets triggered. It might be useful to handle the case by either skipping the task that gets triggered while the previous one is executing or run `pruneTrimmedAbortedTxns` immediately after the previous call completes.

pulsar-broker/src/main/java/org/apache/pulsar/broker/transaction/buffer/impl/MetadataTransactionBuffer.java
```diff
@@ -140,6 +148,43 @@ public MetadataTransactionBuffer(PersistentTopic topic, TxnMetadataStore txnStor
...
+     */
+    private ScheduledFuture<?> scheduleAbortedGc() {
+        ScheduledExecutorService executor = brokerExecutor();
+        if (executor == null) {
+            return null;
+        }
+        long intervalSeconds = Math.max(1, topic.getBrokerService().getPulsar().getConfiguration()
+                .getTransactionCoordinatorScalableTopicsGcIntervalSeconds());
+        long intervalMs = TimeUnit.SECONDS.toMillis(intervalSeconds);
+        return executor.scheduleWithFixedDelay(() -> {
```
lhotari: It could be useful to wrap the Runnable with `org.apache.pulsar.common.util.Runnables#catchingAndLoggingThrowables` so that an unexpected RuntimeException doesn't discontinue the scheduling (since `scheduleWithFixedDelay` will stop if any task throws an exception). It's not expected that exceptions would be thrown, but logging would help if they happen and Runnables.catchingAndLoggingThrowables will keep the scheduler running. (exceptions would get logged by [`DefaultUncaughtExceptionHandler` set in PulsarBrokerStart](https://github.com/apache/pulsar/blob/4998cd9e4f66ade6a7972da4afcd0f1546ddfc16/pulsar-broker/src/main/java/org/apache/pulsar/PulsarBrokerStarter.java#L335-L340) in any case, so exceptions aren't silently swallowed without `catchingAndLoggingThrowables`)

lhotari (approved):
LGTM, minor comments about possible overlapped/concurrent execution and the possibility of scheduleWithFixedDelay gettings cancelled if a RuntimeException is thrown.

PR #25980, opened 2026-06-08
[improve][broker] PIP-483: scalable topic auto split/merge

merlimat:
Implements [PIP-483: Scalable Topic Auto Split/Merge](https://github.com/apache/pulsar/blob/master/pip/pip-483.md) (sub-PIP of PIP-468). The controller now scales a scalable topic's segment count automatically, default-on cluster-wide, bounded by hard caps and asymmetric cooldowns.

End-to-end flow: each broker reports per-segment load to the metadata store → the controller leader reads load + consumer counts → a pure decision function decides → it dispatches to the existing split/merge protocols.

## Increments (one commit each, each independently tested)

1. **Decision core** — `AutoScalePolicyEvaluator.decide(...)`, a pure I/O-free function: split pass (consumer-count then traffic, gated by `splitCooldown` + `maxSegments`) then merge pass (cold adjacent pair, gated by `mergeCooldown` + `mergeWindow` + `minSegments` + `maxDagDepth`). Plus `SegmentLoadStats`, `AutoScaleConfig`, `AutoScaleDecision`, `SegmentLoadSample`, and `SegmentLayout.mergeDepth()`.
2. **Broker config + resolver** — 17 `scalableTopic*` knobs (auto-scale on, `maxSegments=64`, `maxDagDepth=10`, `splitCooldown=60s`, `mergeCooldown/Window=300s`, four split + four merge rate thresholds, etc.) and `AutoScaleConfig.fromBrokerConfig()`.
3. **Load record + reporter** — `ScalableTopicResources` load get/put/delete paths and `SegmentLoadReporter`, which writes a sample only when a rate changed materially (default ±25%) or crossed zero, keeping metadata write volume bounded.
4. **Controller wiring** — periodic `AutoScaleTick` (traffic) + event-driven evaluation on stream/checkpoint consumer registration (consumer-count scale-up within seconds, no polling). Serialized by an in-flight guard; `splitSegment`/`mergeSegments` set the cooldowns so manual ops count too.
5. **Broker sweep** — `BrokerService` periodically sweeps the `segment://` topics it hosts, reads their in/out msg+byte rates, and feeds the reporter. This populates the records the controller reads.

## Design highlights

- **Splits are fast, merges are lazy.** Splits fire as soon as conditions hold (only `splitCooldown` coalesces bursts); merges require a durable cold window and a longer cooldown.
- **Max DAG depth caps merges, not splits** — bounds split↔merge flip-flopping while never blocking a throughput-driven split.
- **The controller reacts, it doesn't poll** — load is pushed to metadata on material change; the controller reads it. No cross-broker RPC fan-out.

## Test plan

- [x] `AutoScalePolicyEvaluatorTest` (20 cases) — every split/merge rule, caps, cooldowns, hysteresis, adjacency, max-depth.
- [x] `AutoScaleConfigTest`, `SegmentLoadReporterTest`, `SegmentLayoutTest` (mergeDepth).
- [x] `ScalableTopicControllerAutoScaleTest` — load-driven split, consumer-driven split (event path), cold-pair merge, disabled no-op, cooldown blocks second split (against in-memory store + mocked admin).
- [x] `V5SegmentLoadReporterTest` — end-to-end: producing across a scalable topic + running the sweep writes a load record per segment.
- [x] Full `org.apache.pulsar.broker.service.scalable.*` suite + checkstyle.

## Follow-up (separate PR)

Per-namespace and per-topic `AutoScalePolicyOverride` + `admin.scalableTopics().set/get/removeAutoScalePolicy(...)`. The feature is fully functional via the cluster config without it.

### DIFF
--- a/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/AutoScalePolicyEvaluator.java
+++ b/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/AutoScalePolicyEvaluator.java
@@ -1,0 +1,252 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pulsar.broker.service.scalable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.apache.pulsar.common.scalable.SegmentInfo;
+import org.apache.pulsar.common.scalable.SegmentLoadStats;
+
+/**
+ * Pure, side-effect-free decision function for scalable-topic auto split/merge (PIP-483).
+ *
+ * <p>Given a snapshot of the current layout, per-segment load samples, per-subscription
+ * stream/checkpoint consumer counts, the resolved policy, and the current time, it returns
+ * exactly one {@link AutoScaleDecision}. It performs no I/O and holds no state — the caller
+ * (the controller leader) collects the inputs and dispatches the result.
+ *
+ * <p>It runs two passes and emits at most one action:
+ * <ol>
+ *   <li><b>Split</b> (fast, lightly coalesced by {@code splitCooldown}): consumer-count
+ *       scale-up first, then traffic-driven scale-up.</li>
+ *   <li><b>Merge</b> (lazy, gated by {@code mergeCooldown} + {@code mergeWindow} +
+ *       {@code maxDagDepth}): only if no split fired.</li>
+ * </ol>
+ */
+public final class AutoScalePolicyEvaluator {
+
+    private AutoScalePolicyEvaluator() {
+    }
+
+    /**
+     * Decide whether to split, merge, or do nothing.
+     *
+     * @param layout              the current segment layout
+     * @param loadBySegment       per active-segment load sample; a missing entry is treated
+     *                            as zero load with no age (never merge-eligible)
+     * @param streamConsumerCount per-subscription count of STREAM/CHECKPOINT (controller-managed)
+     *                            consumers; QUEUE subscriptions are excluded by the caller
+     * @param config              the resolved policy
+     * @param nowMs               current wall-clock time, epoch millis
+     * @param lastSplitAtMs       epoch millis of the last split on this topic (manual or auto),
+     *                            or {@code Long.MIN_VALUE} if none
+     * @param lastMergeAtMs       epoch millis of the last merge on this topic (manual or auto),
+     *                            or {@code Long.MIN_VALUE} if none
+     * @return the decision
+     */
+    public static AutoScaleDecision decide(
+            SegmentLayout layout,
+            Map<Long, SegmentLoadSample> loadBySegment,
+            Map<String, Integer> streamConsumerCount,
+            AutoScaleConfig config,
+            long nowMs,
+            long lastSplitAtMs,
+            long lastMergeAtMs) {
+
+        if (!config.enabled()) {
+            return AutoScaleDecision.NONE;
+        }
+
+        List<SegmentInfo> active = new ArrayList<>(layout.getActiveSegments().values());
+
+        AutoScaleDecision split = trySplit(active, loadBySegment, streamConsumerCount,
+                config, nowMs, lastSplitAtMs);
+        if (!(split instanceof AutoScaleDecision.NoAction)) {
+            return split;
+        }
+
+        return tryMerge(active, layout, loadBySegment, config, nowMs, lastMergeAtMs);
+    }
+
+    // --- Split pass ---
+
+    private static AutoScaleDecision trySplit(
+            List<SegmentInfo> active,
+            Map<Long, SegmentLoadSample> loadBySegment,
+            Map<String, Integer> streamConsumerCount,
+            AutoScaleConfig config,
+            long nowMs,
+            long lastSplitAtMs) {
+
+        if (active.size() >= config.maxSegments()) {
+            return AutoScaleDecision.NONE;
+        }
+        if (withinCooldown(nowMs, lastSplitAtMs, config.splitCooldown().toMillis())) {
+            return AutoScaleDecision.NONE;
+        }
+
+        // (a) Consumer-driven: per-subscription max. If any managed subscription has more
+        // consumers than there are active segments, add a segment so the 1:1 assignment can
+        // give the extra consumer its own segment. Split the busiest segment by msgRateIn so
+        // the new pair lands where it relieves the most ingest.
+        int requiredConsumers = streamConsumerCount.values().stream()
+                .mapToInt(Integer::intValue).max().orElse(0);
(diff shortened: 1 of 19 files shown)


lhotari:
I performed an AI assisted (Claude Fable 5) review and these were the findings. Some might be useful to address.

### Findings

1. **[QUALITY] In-flight guard can leak permanently — `ScalableTopicController.java:322-341`**
   If anything between the successful `autoScaleInFlight.compareAndSet` and the future-chain construction throws synchronously (e.g. a misbehaving `getSegmentLoadAsync`), the flag is never reset: `runAutoScaleSafely` catches the `Throwable` but doesn't release the guard, so auto-scaling silently stops on this topic until a leadership change. Low likelihood (sync throws from async methods violate the codebase rules), but the failure mode is invisible and permanent — the feature dies with no recovery and nothing but a single warn log to show for it. A try/catch that resets the flag would harden it cheaply.

2. **[QUALITY] Load records leak in the metadata store; `forget()` is dead code**
   `deleteSegmentLoadAsync` and `SegmentLoadReporter.forget()` have no production callers (tests only). The GC tick prunes segments from the layout and deletes backing topics (`ScalableTopicController.java:1126`) but leaves the `.../segments/{id}/load` znode behind forever. Likewise the reporter's `lastWritten` map grows unboundedly with segment churn on a long-lived broker, and its documented contract — "call when this broker stops owning the segment topic" — is never wired into unload/seal/delete. Both APIs were built for a purpose this PR doesn't fulfill; either wire them in (GC tick for the znode, topic-close path for `forget`) or note it as a known follow-up.

3. **[QUALITY] Materiality band is anchored at the last-written value, not at the thresholds — confirm intended**
   The reporter skips writes when every rate changed <25% relative to the *last written* value (`SegmentLoadReporter.java:95-102`), with only a zero-crossing exception. The effect approximates hysteresis but with two quirks: it's **path-dependent** (a segment at 10,900 msg/s splits if the last record was 5,000, but not if it was 8,800 — same load, opposite outcomes), and the suppression is **permanent rather than a delay** (a true rate that settles inside `(stored, stored × 1.25)` never produces a new record, so a segment can run sustained at up to 25% over the split threshold — or sit 25–33% below the merge threshold — indefinitely without action). The impact is bounded by the tunable 0.25 knob and is one-directional (delays action, never causes spurious action), and adding threshold-crossing writes would interact with the merge window (each crossing resets the cold clock — arguably correct, but a behavior change). The PR frames the knob purely as metadata-write economy, so the right ask is: if the blind spot is accepted, document it on `scalableTopicLoadReportRateChangeThreshold`; if not, the reporter or evaluator needs threshold awareness — complicated by the reporter being deliberately policy-agnostic.

4. **[QUALITY] Topic ownership moves reset the merge window**
   `reportSegmentLoadAsync` uses `readModifyUpdateOrCreate(existing -> stats)`, which writes unconditionally; a newly-owning broker has an empty `lastWritten` cache, so after every unload/rebalance an *identical* record is rewritten and the znode mtime — which `coldEnough` uses as "cold since" — resets. Under frequent rebalancing, merges can be starved cluster-wide, since load-manager shedding is routine in a busy cluster. Returning `existing` from the RMW lambda when the values are equal would fix it. Relatedly, `coldEnough` compares the controller's `clock.millis()` against the metadata store's server-side mtime, so broker↔store clock skew shifts the window — acceptable for a heuristic, worth a code comment.

5. **[QUALITY] No validation of the policy config**
   Nothing enforces `minSegments ≤ maxSegments`, positive cooldowns, or the invariant the `AutoScaleConfig` javadoc itself states as a requirement — split thresholds strictly above merge thresholds. A threshold configured as 0 makes the evaluator divide by zero: a positive rate scores `Infinity` (permanent split pressure), while `0/0` is NaN and silently ignored. A validation pass in `fromBrokerConfig` would be cheap and catches operator error at the right layer.

6. **[QUALITY] Cooldowns don't survive leader failover, and a failed split consumes one**
   `lastSplitAtMs`/`lastMergeAtMs` are in-memory volatiles that reset to `MIN_VALUE` on a new leader, so an auto merge can fire immediately after failover even if one ran seconds earlier. Both timestamps are also set *before* the operation runs (`ScalableTopicController.java:557,608`), so a split that fails on a metadata conflict still burns the 60 s cooldown. Both are bounded in impact; worth a comment if intentional.

7. **[QUALITY] Several `dynamic = true` knobs aren't actually dynamic**
   `scalableTopicLoadReportIntervalSeconds` and `scalableTopicLoadReportRateChangeThreshold` are read once in `BrokerService.startSegmentLoadReporter()` at broker start; `scalableTopicAutoScaleIntervalSeconds` once at leader election; and turning `scalableTopicAutoScaleEnabled` *on* after it was off at election time has no effect until the next leadership cycle (turning it off works, since `evaluateAndAct` rechecks every evaluation). The rate thresholds and cooldowns *are* genuinely dynamic via `AutoScaleConfig.fromBrokerConfig` per evaluation. Either honor changes for the others or drop their `dynamic` flag so `update-dynamic-config` users aren't misled.

8. **[INTENT MISMATCH] "Consumer-count scale-up within seconds" overstates the burst case**
   Only the *first* split fires within seconds of consumer registration; a burst of N consumers scales one split per `max(splitCooldown, tick)`, so going from 1 segment to N takes roughly (N−1)×60 s. A consumer-change event arriving while an evaluation is in-flight is also silently dropped (CAS fails) and only recovered by the next tick. Both behaviors follow from the cooldown design the PR body itself describes — calibrate the headline claim, or consider re-triggering an evaluation after a consumer-driven split completes so a burst converges in one cooldown chain rather than tick-by-tick.

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/AutoScalePolicyEvaluator.java
```diff
@@ -0,0 +1,252 @@
...
+        String hottestReason = null;
+        for (SegmentInfo segment : active) {
+            SegmentLoadStats stats = statsOf(segment.segmentId(), loadBySegment);
+            double score = 0.0;
+            String reason = null;
+            double[] ratios = {
+                    stats.msgRateIn() / config.splitMsgRateIn(),
+                    stats.bytesRateIn() / config.splitBytesRateIn(),
+                    stats.msgRateOut() / config.splitMsgRateOut(),
+                    stats.bytesRateOut() / config.splitBytesRateOut(),
```
lhotari, line 131: does divide by 0 (NaN for double) behave in a correct way?

pulsar-broker/src/main/java/org/apache/pulsar/broker/service/scalable/ScalableTopicController.java
```diff
@@ -237,6 +257,142 @@ private void runGcTickSafely() {
...
+    private CompletableFuture<Void> evaluateAndAct(String trigger) {
+        ServiceConfiguration brokerConfig = brokerConfig();
+        if (brokerConfig == null) {
+            return CompletableFuture.completedFuture(null);
+        }
+        AutoScaleConfig config = AutoScaleConfig.fromBrokerConfig(brokerConfig);
+        if (!config.enabled()) {
+            return CompletableFuture.completedFuture(null);
+        }
+        if (!autoScaleInFlight.compareAndSet(false, true)) {
```
lhotari, line 366: How does the concurrency of auto scaling logic work in the case that the user performs "manual" split or merge
with org.apache.pulsar.broker.admin.v2.ScalableTopics#splitSegment (`/{tenant}/{namespace}/{topic}/split/{segmentId}`) and org.apache.pulsar.broker.admin.v2.ScalableTopics#mergeSegments (`/{tenant}/{namespace}/{topic}/merge/{segmentId1}/{segmentId2}`) REST APIs?

Is it planned as part of PIP-483 to be able to instead of splitting and merging manually with the REST API to change `maxSegments`, `minSegments` and `maxDagDepth` per scalable topic? It could be useful to have some setting to control whether the change is made immediately or only when the auto scaling rules take place. I guess scaling immediately after increasing `minSegments` would have some separate challenges for picking the segments which are worth splitting. I guess it could be done based on load stats also in that case. Decreasing `maxSegments` would be a similar case but in the other direction.

lhotari (approved):
LGTM, just check the local Claude Fable 5 review findings and the comments from my actual review. Feel free to mark them resolved without waiting for a new review.

